PR #9, opened 2026-07-17
docs: comprehensive gemi framework documentation

nstfkc:
## Summary

Adds a complete, source-verified documentation set for the gemi framework under `docs/`, an `llms.txt` bundle for LLM consumption, and a GitHub Pages deployment.

Written against the `server2-migration` source, the `saas-starter` template, and real usage in the `folio` production app. Every documented API was checked against the framework source rather than assumed.

## Documentation (20 pages + index)

- **Foundations:** getting-started, project-structure (Kernel + service providers), configuration, cli
- **HTTP layer:** routing, controllers, middleware
- **Views & client:** views-and-layouts, data-fetching, forms, navigation
- **Auth:** authentication, authorization
- **Services & facades:** facades, file-storage, email, jobs-and-queues, cron, broadcasting, i18n
- **README.md** — overview, when-to-use, doc map, core-concepts table, and a "things that will bite you" section

Old drafts (`Routing.md`, `ViewRoutes.md`, `Introduction.md`, etc.) are removed. Filenames are normalized to lowercase (the old capitalized names case-collided on macOS; committed cleanly so links resolve on case-sensitive systems).

## llms.txt (llmstxt.org convention)

- `docs/llms.txt` — curated index linking each page.
- `docs/llms-full.txt` — every page concatenated into one file for one-shot LLM context.

## GitHub Pages

- `.github/workflows/docs.yml` deploys `docs/` verbatim (`.nojekyll`), so all `.md` files and both `llms.txt`/`llms-full.txt` are served at stable URLs with relative links intact.
- `docs/index.html` — a static landing page listing every doc.
- Pages is enabled (build type: GitHub Actions); the `github-pages` environment now allows `main` and `server2-migration`. **The site publishes automatically on merge** to https://nstfkc.github.io/gemi/

## Discrepancies surfaced while writing (worth a look)

- **`useGet` doesn't exist** — the only query hook gemi exports is `useQuery`. `apps/web/CLAUDE.md` in the app repo references `useGet` and could be corrected.
- **Cron shortcuts** — `folio` uses `CronJob.exp("@at_9:00")`-style shortcuts, but on `server2-migration` `CronJob` now uses `Bun.cron`, which only understands standard 5-field cron + `@daily`/`@hourly` nicknames. The old forms come from the previous parser (still in `dist/`) and would fail against current source. `cron.md` documents the real grammar with a translation table.
- **Router field is `middlewares` (plural)** in source, though some notes use the `middleware` shorthand.

https://claude.ai/code/session_01LUQX44wh8E54X2mY7HkwRX


docs/authorization.md
nstfkc, line 1: Remove the policies section, it will be deprecated.

docs/broadcasting.md
nstfkc, line 1: remove BroadcastingServiceProvider will be deprecated

docs/data-fetching.md
```diff
@@ -0,0 +1,288 @@
...
+`Query` facade from `gemi/facades`. Call it inside a view or layout handler
+([Views and Layouts](./views-and-layouts.md)); when the matching `useQuery` mounts on
+the client, its data is already there as `fallbackData`.
+
+```typescript
+import { Query } from "gemi/facades";
+
+"/dashboard": this.view("Dashboard", async () => {
+  // Fetch eagerly and await it (blocks render until ready)
+  await Query.instant("/todos");
```
nstfkc: this is for using the data here. not assigning it to a variable makes it useless. This is for not repeating the logic in the api route and reusing it here.

docs/data-fetching.md
```diff
@@ -0,0 +1,288 @@
...
+| `data` | The response body after success. |
+| `error` | A `MutationError` on failure (see below). |
+| `loading` | `true` while in flight. |
+| `cancel()` | Abort the in-flight request. |
+| `formData` | A mutable `FormData` accumulator used when `trigger()` is called with no input. |
+
+`useMutation` takes the method explicitly:
+
+```tsx
+import { useMutation } from "gemi/client";
```
nstfkc: `useMutation` hook should not be used, usePost, usePut etc should be preferred. But ultimately, Form is the best way to send mutations to the backend.

docs/data-fetching.md
```diff
@@ -0,0 +1,288 @@
...
+const mutate = useMutate();
+
+// After creating a todo:
+mutate({ path: "/todos" }, (todos) => [newTodo]);
+```
+
+The signature is `mutate({ path, params?, search? }, fn?)`, with the same
+merge/append semantics as `useQuery`'s `mutate`.
+
+## Type safety: the generated `gemi.d.ts`
```
nstfkc: This doesn't need to be explained. Just mention that type-safety between routes and client is established by the framework.

docs/getting-started.md
```diff
@@ -0,0 +1,138 @@
...
+
+A single gemi project serves both your API and your React front end from one Bun process. Views are rendered on the server and hydrated on the client, and the network layer between them is fully typed end to end.
+
+### When should you use gemi?
+
+gemi is a good fit for **B2B applications with many API endpoints and non-trivial business logic** — dashboards, internal tools, and SaaS products where you control the whole stack.
+
+> **Note:** If you are building a mostly static site, a blog, or a content-heavy e-commerce storefront where cold-start load time is the top priority, a static-first framework like Next.js or Astro may serve you better. gemi optimizes for application development, not static content delivery.
+
+## Requirements
```
nstfkc, line 17: @vitejs/plugin-react also required

docs/i18n.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+An instance dictionary also offers **server-only** helpers:
+
+- `dictionary.render(key, { locale?, params? })` — resolve a single key to a string in the given (or current) locale, with interpolation.
+- `dictionary.reference(key)` — return the full per-locale map for a key.
+
+> **Note:** `render` and `reference` throw if called in the browser — they are server-only. On the client, use `useTranslator`.
+
+### App-level translation services (`Translation.transform`)
```
nstfkc: This doesn't belong to the gemi.

docs/i18n.md
```diff
@@ -0,0 +1,248 @@
...
+
+```tsx
+import { I18n } from "gemi/facades";
+
+// When dispatching the email
+sendPinEmail({ locale: I18n.locale(), pin });
+
+// Inside the email component
+function PinEmail({ locale, pin }: { locale: string; pin: string }) {
+  const heading = locale === "tr-TR" ? "Giriş kodunuz" : "Your login code";
```
nstfkc: Emails can use `Dictionary.text` as well.

docs/jobs-and-queues.md
nstfkc, line 1: Provide a more generic example e.g video processing

nstfkc (commented):
Looks good, just a few comments.

nstfkc:
Addressed all comments in `889f451`:

- **authorization.md** — removed the **Policies** section entirely (being deprecated) and all lingering "policy" references across the docs.
- **broadcasting.md** — dropped the `BroadcastingServiceProvider` registration block (being deprecated); kept the topic↔channel pattern-matching explanation.
- **data-fetching.md**
  - `Query.instant` example now assigns and returns the data — framed as **reusing an API route handler on the server** instead of duplicating it.
  - Mutations section now recommends **`Form`** as the primary way and the typed **`usePost`/`usePut`/…** hooks for imperative use; removed the `useMutation` example.
  - Condensed the `gemi.d.ts` section to a short "type-safety is established by the framework" mention (kept only the don't-edit / regenerate note).
- **getting-started.md** — added **`@vitejs/plugin-react`** to Requirements.
- **i18n.md** — removed the app-specific `Translation.transform` section; the email example now uses **`Dictionary.text({...}, { locale })`**.
- **jobs-and-queues.md** — replaced the folio-specific `TranslationJob` with a generic **`ProcessVideoJob`** example.

Also regenerated `llms-full.txt` and re-verified all internal links resolve.

Issue #2, opened 2024-07-11
bun: command not found: open


draylegend:
```js
dl@dl:~$ bun upgrade
Bun v1.1.18 is out! You're on v1.1.17
[4.63s] Upgraded.

Welcome to Bun v1.1.18!

What's new in Bun v1.1.18:

    https://bun.sh/blog/release-notes/bun-v1.1.18

Report any bugs:

    https://github.com/oven-sh/bun/issues

Commit log:

    https://github.com/oven-sh/bun/compare/bun-v1.1.17...bun-v1.1.18
dl@dl:~$ bunx create-gemi-app
✔ Enter project name: … gemi
Extracting to /home/dl/gemi
Downloading template...
Installing dependencies...
bun install v1.1.18 (5a0b9352)
Resolving dependencies
Resolved, downloaded and extracted [2714]
warn: incorrect peer dependency "eslint@8.45.0"

warn: incorrect peer dependency "eslint@8.45.0"

warn: incorrect peer dependency "eslint@8.45.0"

warn: incorrect peer dependency "eslint@8.45.0"
Saved lockfile

+ @types/react@18.3.3
+ @types/react-dom@18.3.0
+ @typescript-eslint/eslint-plugin@7.16.0
+ @typescript-eslint/parser@7.16.0
+ @vitejs/plugin-react-swc@3.7.0
+ autoprefixer@10.4.19
+ eslint@8.45.0 (v9.6.0 available)
+ eslint-plugin-react@7.34.3
+ postcss@8.4.39
+ prettier@3.3.2
+ prisma@5.16.2
+ tailwindcss@3.4.4
+ typescript@5.5.3
+ vite@5.3.3
+ vitest@1.6.0 (v2.0.2 available)
+ @libsql/client@0.6.2 (v0.7.0 available)
+ @prisma/adapter-libsql@5.16.2
+ @prisma/client@5.14.0 (v5.16.2 available)
+ @react-email/components@0.0.19 (v0.0.21 available)
+ gemi@0.4.9
+ react@19.0.0-rc-dfd30974ab-20240613
+ react-dom@19.0.0-rc-dfd30974ab-20240613
+ sharp@0.33.4

695 packages installed [37.29s]
Initialized empty Git repository in /home/dl/gemi/.git/
Happy coding
visit https://github.com/nstfkc/gemi for documentation



run cd `gemi` and run `bun dev` to start the development server
dl@dl:~$ cd gemi
dl@dl:~/gemi$ bun dev
[...]

PR #22, opened 2026-07-21
Fix mutate(fn): replace instead of merge, and stop no-op'ing on unresolved queries

### LINKED ISSUE #20: useQuery mutate(fn) merges instead of replacing, and no-ops on unresolved/lazy queries
## Summary

`useQuery(...).mutate(fn)` has two surprising behaviors that together make it easy to write call sites that are silently wrong. The function form does **not replace** the cached data with what the callback returns — it **merges** it — and on an unresolved/lazy query it silently does **nothing**. Both are undocumented and bite in opposite directions.

Source: `packages/gemi/client/useQuery.ts` (the `mutate` wrapper) and `packages/gemi/client/QueryResource.ts` (`QueryResource#mutate`).

## Behavior 1 — `mutate(fn)` merges instead of replaces

The `mutate` wrapper in `useQuery.ts` takes the callback's return value and folds it back into the existing cache rather than replacing it:

```ts
const updatedData = typeof fn === "function" ? fn(data) : fn;
if (isPlainObject(data)) {
  if (isPlainObject(updatedData)) return { ...data, ...updatedData };
  throw new Error("Mutate function must return an object when the current data is an object.");
}
if (Array.isArray(data)) {
  if (Array.isArray(updatedData)) return [...data, ...updatedData]; // <-- appends
  throw new Error("Mutate function must return an array when the current data is an array.");
}
```

Consequences for the natural ways people write optimistic updates:

- **Identity refresh** `mutate((x) => x)` on a loaded array → cache becomes `[...data, ...data]`, so every row renders twice until the follow-up refetch lands. The intent ("refresh this list") is not what the code expresses, and the intermediate state is wrong.
- **Optimistic append** `mutate((list) => [...list, newItem])` → `[...list, ...list, newItem]`, duplicating every existing row before adding the new one.
- **Optimistic delete / update** `mutate((list) => list.filter(...))` or `.map(...)` → the returned array is *appended* to the cached one, so the removed row stays and the survivors duplicate. There is currently **no way to shrink or wholesale-replace** an array (or replace, rather than shallow-merge, an object) through `mutate`.

The array/append shape is self-correcting because a refetch follows the cache write, but it round-trips through a visibly wrong intermediate render for no benefit, and the delete/update case can't be expressed correctly at all.

## Behavior 2 — `mutate(fn)` is a no-op on an unresolved or lazy query

`QueryResource#mutate` guards on existing data and returns early:

```ts
const state = store.get(variantKey);
if (!state || !state.data) return; // no cache write, no refetch
```

So when the query hasn't resolved yet — most importantly a `lazy` query, which never populates the cache on mount — `mutate(fn)` returns before doing anything: no cache write **and no request**. Any call site that used the function form to mean "go refresh this" is a permanent no-op unless some other eager consumer happens to have filled the same key.

Real example this bit: a poll that fires `mutate((x) => x)` every few minutes on a `lazy` pending-order-count query never reached the server — every tick hit this guard and returned.

(The argument-less `mutate()` is fine — it calls `refetch` directly and bypasses both behaviors.)

## Why this is a footgun

Both behaviors are undocumented, and the two forms of `mutate` do very different things:

- `mutate()` → refetch from server.
- `mutate(fn)` → merge `fn(data)` into cache (append for arrays, shallow-spread for objects), then refetch — **but only if data is already loaded**, otherwise silently nothing.

The identity/append/filter shapes all *look* correct and mostly self-heal via the trailing refetch, so the bug hides. The lazy-query case does not self-heal.

## Suggested directions (pick per intent)

1. **Make replace the default, or offer an explicit replace.** The callback returning the new data should set the cache to that data. If append/merge is a wanted behavior, expose it explicitly (e.g. a separate helper or option) rather than making it the implicit meaning of any returned array/object.
2. **Don't silently drop `mutate(fn)` on unresolved data.** Either fall through to a refetch when there's nothing to mutate, or surface a warning, so a lazy-query poll isn't a silent no-op.
3. **Document the two forms** and the merge semantics either way.

## Context

Surfaced while auditing `useQuery` call sites in a downstream app — folioai.com PR #355 replaces ~26 identity/append `mutate((x) => x)` and `[...list, item]` call sites with argument-less `mutate()` as a workaround. That fixes the app but the sharp edges live in gemi.

gemi version observed: `0.41.2`.

nstfkc:
Fixes #20.

`useQuery(...).mutate(fn)` (and `useMutate`'s equivalent) had two undocumented footguns that made call sites silently wrong in opposite directions.

## Behavior 1 — merge/append instead of replace

The callback's return value was folded back into the cache rather than replacing it: objects were shallow-spread and **arrays were appended**. This meant:

- `mutate((x) => x)` on a loaded array → `[...data, ...data]` (every row rendered twice)
- `mutate((list) => [...list, item])` → duplicated every existing row
- Optimistic delete/update via `filter`/`map` was **impossible** — the returned array was appended, so removals stayed and survivors duplicated

**Fix:** the callback's return value now **replaces** the cached data. The type checks are kept only to guard the shape (object stays object, array stays array). Spread the existing data yourself when you want to keep it:

```tsx
mutate((todos) => [...todos, newItem]);          // append
mutate((todos) => todos.filter((t) => t.id !== id)); // delete
mutate((todos) => todos.map((t) => t.id === id ? next : t)); // update
```

## Behavior 2 — silent no-op on unresolved / lazy queries

`QueryResource#mutate` returned early when nothing was cached (`!state || !state.data`), so `mutate(fn)` on a `lazy` query — which never populates the cache on mount — did nothing: no cache write **and no request**. A poll firing `mutate((x) => x)` on a lazy query never reached the server.

**Fix:** when there's nothing to update optimistically, fall through to a refetch, so `mutate(fn)` still means "go get the latest data".

## Also

- Updated the `mutate` type signatures (dropped `Partial<...>`, since the callback now returns the full next value) in both `useQuery` and `useMutate`.
- Documented both forms of `mutate` and the replace-then-refetch semantics in `docs/data-fetching.md`.

## Changed files

- `packages/gemi/client/QueryResource.ts` — fall through to refetch instead of no-op
- `packages/gemi/client/useQuery.ts` — replace instead of merge/append; type signature
- `packages/gemi/client/useMutate.ts` — same replace fix; type signature
- `docs/data-fetching.md` — document the two forms

### DIFF
--- a/docs/data-fetching.md
+++ b/docs/data-fetching.md
@@ -94,19 +94,31 @@ const { data } = useQuery("/feed", {}, {
 
 ### Optimistic updates with `mutate`
 
-The `mutate` returned by `useQuery` updates the cached value in place:
+The `mutate` returned by `useQuery` has two forms:
 
 ```tsx
 const { data, mutate } = useQuery("/todos");
 
-// Merge/append into the current data
-mutate((todos) => [{ id: "tmp", title: "New" }]);
+// 1. Optimistically REPLACE the cached data with what the callback returns,
+//    then refetch from the server in the background.
+mutate((todos) => [...todos, { id: "tmp", title: "New" }]); // append an item
+mutate((todos) => todos.filter((t) => t.id !== id));        // remove an item
+mutate((todos) => todos.map((t) => (t.id === id ? next : t))); // update an item
 
-// Or refetch by calling with no args
+// 2. Refetch from the server (no optimistic update) by calling with no args.
 mutate();
 ```
 
-For objects, the returned partial is shallow-merged; for arrays, it is appended.
+The callback's return value **replaces** the cached value — it is not merged or
+appended, so you return the full next value (spread the existing data yourself when
+you want to keep it). It must keep the same shape as the current data: return an
+object when the data is an object, an array when it's an array. After the optimistic
+write, `mutate` always refetches so the cache reconciles with the server.
+
+`mutate(fn)` on a query whose data hasn't loaded yet (including a `lazy` query) has
+nothing to update optimistically, so it falls through to a refetch rather than doing
+nothing.
+
 To update a query from **outside** the component that owns it, use `useMutate`.
 
 ## Reusing endpoints on the server: the `Query` facade
@@ -232,7 +244,7 @@ mutate({ path: "/todos" }, (todos) => [newTodo]);
 ```
 
 The signature is `mutate({ path, params?, search? }, fn?)`, with the same
-merge/append semantics as `useQuery`'s `mutate`.
+replace-then-refetch semantics as `useQuery`'s `mutate`.
 
 ## Type safety
 
--- a/packages/gemi/client/QueryResource.ts
+++ b/packages/gemi/client/QueryResource.ts
@@ -80,6 +80,10 @@ export class QueryResource {
     const store = this.store.getValue();
     const state = store.get(variantKey);
     if (!state || !state.data) {
+      // Nothing is cached yet to update optimistically — e.g. a lazy query, or
+      // one that hasn't resolved. Fall through to a refetch so `mutate(fn)` is
+      // not a silent no-op (it still means "go get the latest data").
+      this.resolveVariant(variantKey, false, false);
       return;
     }
     const data = fn(state.data);
--- a/packages/gemi/client/useMutate.ts
+++ b/packages/gemi/client/useMutate.ts
@@ -29,8 +29,8 @@ export function useMutate() {
       search?: Record<string, any>;
     },
     fn?:
-      | ((data: NestedPrettify<Data<T>>) => Partial<NestedPrettify<Data<T>>>)
-      | Partial<NestedPrettify<Data<T>>>,
+      | ((data: NestedPrettify<Data<T>>) => NestedPrettify<Data<T>>)
+      | NestedPrettify<Data<T>>,
   ) {
     const { path, params = {}, search = {} } = options ?? {};
     const normalPath = applyParams(path, params);
@@ -41,18 +41,20 @@ export function useMutate() {
     return resource.mutate.call(resource, variantKey, (data: any) => {
       if (data === undefined || data === null) {
         console.warn("Mutate function called before the query.");
-        return;
+        return data;
       }
 
       if (!fn) {
         return data;
       }
 
+      // The callback's return value *replaces* the cached data. The type checks
+      // below only ensure the shape matches; they do not merge or append.
       const updatedData = typeof fn === "function" ? fn(data) : fn;
 
       if (isPlainObject(data)) {
         if (isPlainObject(updatedData)) {
-          return { ...data, ...updatedData };
+          return updatedData;
         }
         throw new Error(
           "Mutate function must return an object when the current data is an object.",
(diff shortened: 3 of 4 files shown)


nstfkc:
## Review: replace-instead-of-merge + stop no-op'ing on unresolved queries

### Overview
Fixes two real footguns in `useQuery(...).mutate` / `useMutate`:
1. **Merge/append → replace:** the callback's return value now *replaces* the cached data instead of being shallow-merged (objects) or appended (arrays), making `filter`/`map`-based optimistic delete/update actually possible.
2. **No-op on unresolved/lazy queries → refetch:** `QueryResource.mutate` previously returned early when nothing was cached; it now falls through to a refetch so `mutate(fn)` always at least fetches.

Type signatures are updated (dropping `Partial<>`) and `docs/data-fetching.md` is revised. Well-scoped, clean diff, and an excellent PR description that documents the rationale and migration path.

### Correctness — looks right
- Replace logic is correct: `isPlainObject`/`Array.isArray` guards are retained purely to enforce shape, and the merge/append is replaced with `updatedData`. Consistent across both `useQuery.ts` and `useMutate.ts`.
- The fall-through refetch (`resolveVariant(variantKey, false, false)`) uses `silent=false` (shows loading) and `cache=false` (network reload) — appropriate for "go get the latest."

### Issues & risks

**1. Breaking change — call out the semver impact.**
Switching object-merge → replace silently changes existing behavior. A previously-common pattern like:
```tsx
mutate((post) => ({ likes: post.likes + 1 })) // used to shallow-merge, keeping other fields
```
now **replaces the whole object**, dropping every other field. This is intended (the fix for #20) and the docs cover it, but there's no test and nothing forces consumers to notice at compile time — the merge case that used to "work" now silently drops data at runtime. Recommend a version bump per your convention and a CHANGELOG/migration note beyond the docs.

**2. Possible redundant fetch when mutating an in-flight query (minor).**
The new early branch triggers on `!state || !state.data`. During the initial load a variant exists with `loading: true` but `data` still undefined, so calling `mutate` mid-flight now kicks off a *second* concurrent `resolveVariant` for the same endpoint (previously a no-op). Same URL, so the final data is consistent, but it's a wasted request and a possible loading flicker. Consider guarding:
```ts
if (!state || !state.data) {
  if (!state?.loading) this.resolveVariant(variantKey, false, false);
  return;
}
```

**3. Dead-code null guard (nit, pre-existing).**
The `if (data === undefined || data === null) { ...; return data; }` block inside the callbacks is unreachable: `QueryResource.mutate` only calls `fn(state.data)` after the `!state.data` guard, so `fn` never receives null/undefined. The `return;` → `return data;` change is harmless but doesn't fix a real path. Could be removed for clarity in a follow-up.

### Style & conventions
- Added comments are clear and match surrounding style.
- Type overloads (`useQuery.ts`) and the union type (`useMutate.ts`) are updated consistently.
- Docs rewrite is accurate and matches the new runtime behavior.

### Test coverage
No tests. The `client/` directory has **no existing test files** (tests live only under `app/` and `services/router/`), so the omission is consistent with repo convention. That said, this is a behavioral fix that flipped a documented contract — a small unit test around `QueryResource.mutate` (replace semantics + fall-through-refetch on empty cache) would be high-value insurance against future regressions, even if it's the first client test.

### Security / performance
- No security surface touched.
- Replace avoids the old accidental array duplication (a real render-time win). Only cost is the potential redundant fetch in issue #2.

### Verdict
**Approve with minor suggestions.** Core fix is correct and a net improvement. Before merging: (a) confirm version-bump/changelog treatment given the breaking merge→replace change, and (b) optionally add the in-flight guard from issue #2.

PR #30, opened 2026-07-23
Adopt Laravel's real container/provider architecture

### LINKED ISSUE #29: Adopt Laravel's real container/provider architecture
# Adopt Laravel's real container/provider architecture

## Context

gemi borrows Laravel's vocabulary but inverts two of its core meanings, and the
mismatch is now baked into the public API that every app writes by hand.

**`XServiceContainer` is not a container.** In Laravel there is exactly *one*
container (`Illuminate\Container\Container`) and it resolves *many* services.
gemi has 16 classes named `*ServiceContainer`, each wrapping exactly one service
and exposing a static `.use()` that reads an `AsyncLocalStorage` store
(`packages/gemi/services/ServiceContainer.ts:10`). The actual container is the
untyped `Kernel.services: Record<string, ServiceContainer>` map
(`packages/gemi/kernel/Kernel.ts:52`), keyed by hand-maintained `_name` strings.
So the one real container is anonymous, and 16 non-containers carry the name.

**`XServiceProvider` is not a service provider.** A Laravel provider's defining
job is `register()` — binding things *into* the container. gemi's providers are
property bags that get *handed to* a container
(`new EmailServiceContainer(new this.emailServiceProvider())`,
`Kernel.ts:58`). What they hold is settings and hooks — `driver`, `headers`,
`filterRecipients()` for email; `maxFileSize`, `onLogCreated()` for logging.
That is `config/mail.php` and `config/logging.php`, not a provider. Telling
evidence: `ServiceProvider.boot()` is declared abstract and implemented as an
empty method in every subsystem, and **the Kernel never calls it** — the only
`.boot()` call site in the package is `LoggingServiceContainer.ts:32` calling
its own method. Laravel's two-phase register/boot lifecycle simply doesn't
exist here, so the base class's one abstract member is dead weight.

Two things are already right and should be built on, not replaced:
`packages/gemi/facades/` is a genuine, correct facade layer (`Auth`, `Log`,
`Redis`, `Broadcast`), and `ServiceContainer.use()` over `AsyncLocalStorage` is
a sound request-scoped resolution mechanism — it just needs to become
`Container.make()`.

**Decisions taken:** go to the real architecture (a `Container` with
`bind`/`singleton`/`make` and providers with `register`/`boot`), name services
after Laravel's actual classes rather than a uniform suffix, and break cleanly
at 0.43 with a `gemi migrate` codemod rather than carrying deprecation aliases.

Outcome: app authors write `app/config/*.ts` for settings and
`app/providers/*.ts` for wiring, exactly as in Laravel; the framework gains
lazy service resolution and a real boot lifecycle; and every borrowed word means
what it means upstream.

## Target architecture

| Today | Becomes | Laravel counterpart |
|---|---|---|
| `Kernel.services` record | `Container` class | `Illuminate\Container\Container` |
| (none) | `Application extends Container` | `Illuminate\Foundation\Application` |
| `ServiceProvider` (config bag) | `ServiceProvider` w/ `register()`/`boot()` | `Illuminate\Support\ServiceProvider` |
| `app/kernel/providers/*.ts` | `app/config/*.ts` | `config/*.php` |
| `XServiceContainer` | Manager / service (table below) | `AuthManager`, `Router`, … |
| `ServiceContainer.use()` | `app().make(X)` | `app()` / `resolve()` |
| `Singleton` + `SingletonService*` | deleted → `Container.singleton()` | container bindings |
| `facades/*` | unchanged mechanism, `Facade` base added | `Illuminate\Support\Facades\*` |

Service renames (`packages/gemi/**/*ServiceContainer.ts`):

```
AuthenticationServiceContainer      → AuthManager
EmailServiceContainer               → MailManager
LoggingServiceContainer             → LogManager
FileStorageServiceContainer         → FilesystemManager
QueueServiceContainer               → QueueManager
RedisServiceContainer               → RedisManager
BroadcastingServiceContainer        → BroadcastManager
ImageOptimizationServiceContainer   → ImageManager
ApiRouterServiceContainer           → ApiRouter
ViewRouterServiceContainer          → ViewRouter
I18nServiceContainer                → Translator
RateLimiterServiceContainer         → RateLimiter
CronServiceContainer                → Scheduler
MiddlewareServiceContainer          → MiddlewareRegistry
KernelIdServiceContainer            → KernelId
SingletonServiceContainer           → (deleted)
```

`Manager` is used only where Laravel uses it — driver resolution. `Router`,
`Translator`, `RateLimiter`, `Scheduler` resolve no drivers and take Laravel's
own names for those classes.

Target app surface:

```ts
// app/config/mail.ts
import { defineMailConfig, ResendDriver } from "gemi/services";
export default defineMailConfig({
  driver: new ResendDriver(),
  headers: {},
  filterRecipients: (emails) => emails,
});

// app/kernel/Kernel.ts  — cf. Laravel 11 bootstrap/providers.php
import { Kernel } from "gemi/kernel";
import AppServiceProvider from "../providers/AppServiceProvider";
export default class extends Kernel {
  providers = [AppServiceProvider];   // framework providers registered by default
}

// app/providers/AppServiceProvider.ts
export default class extends ServiceProvider {
  register() { this.app.singleton(Billing, () => new Billing()); }
  boot() { /* gates, macros, listeners */ }
}
```

**Hooks stay in config as callbacks.** Laravel would push `filterRecipients` /
`onLogCreated` into `AppServiceProvider::boot()` via `Mail::alwaysTo(…)`-style
macros. TS config objects hold functions natively, so keeping them as config
keys avoids forcing every app to author a provider for a one-line hook. This is
a deliberate, documented divergence.

## Implementation

Sequence so the repo compiles at every step.

**1. Container + Application.** New `packages/gemi/container/Container.ts`:
`bind`, `singleton`, `instance`, `make`, `bound`. Tokens are class constructors
(TS-idiomatic, gives `make(MailManager): MailManager` for free) — this retires
the hand-maintained `static _name` strings. `singleton` factories run lazily on
first `make()`, replacing today's eager construction of all 16 services at boot.
`make()` on an unbound token throws a `BindingResolutionError` — today
`ServiceContainer.use()` returns `undefined` with a commented-out error
(`ServiceContainer.ts:16-18`) and callers `?.` past it.
New `packages/gemi/foundation/Application.ts extends Container`, owning the
provider list and the two-phase lifecycle (all `register()`, then all `boot()`).
Reuse `kernelContext` (`packages/gemi/kernel/context.ts`) as the ALS the
`app()` helper reads; `KernelContext.getStore()` and `Kernel.run()` keep working.

**2. Rename services.** Mechanical rename per the table across
`packages/gemi/**` (~39 files reference `ServiceContainer`). Constructors take
their config object instead of a `ServiceProvider` instance. Fix the two
inconsistencies while here: `AuthenticationServiceContainer` names its
constructor param `provider` where every other container uses `service`
(`auth/AuthenticationServiceContainer.ts:10`), and `Kernel` has
`broadcastingsServiceProvider` (stray plural, `Kernel.ts:44`) and
`imageServiceProvider` for `ImageOptimizationServiceProvider` (`Kernel.ts:47`) —
both vanish with the provider-array rewrite.

**3. Config repository.** `packages/gemi/config/Repository.ts` with dot-path
`get`/`set`, mirroring `Illuminate\Config\Repository`, bound into the container
and loaded from `app/config/*.ts`. Add a `define*Config()` helper per subsystem
next to each service, exported from `gemi/services` — same identity-function
pattern as the existing `defineConfig` in `packages/gemi/config/index.ts`.
**Naming collision to avoid:** `gemi/config` is already the *build* config
(`gemi.config.ts`, `GemiViteConfig`/`GemiBunConfig`). Put the runtime repository
under a distinct export (`gemi/support` or `gemi/foundation`) and leave
`gemi/config` alone.

**4. Framework providers.** One real provider per subsystem
(`MailServiceProvider`, `AuthServiceProvider`, `RouteServiceProvider`,
`LogServiceProvider`, `QueueServiceProvider`, `RedisServiceProvider`,
`FilesystemServiceProvider`, `BroadcastServiceProvider`,
`TranslationServiceProvider`, `ScheduleServiceProvider`,
`RateLimiterServiceProvider`, `ImageServiceProvider`), each `register()`ing its
service as a singleton over its config slice. These are the *default* provider
list; `Kernel.providers` appends app providers. `CronServiceProvider` currently
takes the Kernel in its constructor (`Kernel.ts:96`) — it takes `this.app`
instead, like every other provider. Delete `services/Singleton.ts` and
`services/singleton/` — `Singleton.use()` is a class-name-keyed instance map
that `Container.singleton()` subsumes.

**5. Facades.** Add a `Facade` base with `getFacadeAccessor()` and retarget the
11 files in `packages/gemi/facades/` from `XServiceContainer.use()` to
`app().make(X)`. Rename to Laravel's facade names: `FileStorage` → `Storage`,
`I18n` → `Lang`. `Auth`, `Log`, `Redis`, `Broadcast`, `Cookie`, `Url` already
match.

**6. Template + docs.** Delete `templates/saas-starter/app/kernel/providers/`
(11 files); add `app/config/*.ts` and `app/providers/AppServiceProvider.ts`;
rewrite `app/kernel/Kernel.ts` to the providers array. Update the 13 docs files
that mention the old names, `docs/project-structure.md` for the new layout, and
regenerate `docs/llms.txt` / `docs/llms-full.txt`.

**7. Codemod.** New `migrate` command in `packages/gemi/bin/gemi.ts` (existing
`commander` program): moves `app/kernel/providers/*ServiceProvider.ts` →
`app/config/*.ts` converting property overrides to config keys and method
overrides to callback keys; rewrites `app/kernel/Kernel.ts`; rewrites
`gemi/services` and `gemi/facades` imports. Anything it cannot translate
mechanically is left in place with a `// TODO(gemi-migrate):` comment and
reported in the summary — silent partial migration is worse than a loud one.
Ship `UPGRADE.md` covering the manual cases regardless.

## Verification

- `bun test` in `packages/gemi` (vitest) and `bun run build` — the build has
  five stages (`build:core`, `bin`, `types`, `client`, `plugin`); type errors in
  the rename land in `build:types`.
- Boot the template: `cd templates/saas-starter && bun run dev`, exercise a view
  route, an API route, login (auth), a queued job, and a broadcast channel —
  these cover the ALS-scoped services whose resolution path changed most.
- Assert the lifecycle that doesn't exist today: a test where two providers'
  `register()` both run before either `boot()`, and a test that a service bound
  but never resolved is never constructed (lazy `singleton`).
- `make()` on an unbound token throws rather than returning `undefined`.
- Codemod: run `gemi migrate` against a pristine copy of the pre-change
  `templates/saas-starter`, diff against the hand-written post-change template —
  they should match, and any divergence is either a codemod bug or a case that
  must be documented in `UPGRADE.md`.

nstfkc:
Closes #29.

## Why

gemi borrows Laravel's vocabulary but inverted two of its core meanings, and the mismatch was baked into the public API that every app writes by hand.

**`XServiceContainer` was not a container.** Laravel has exactly one `Container` that resolves many services. gemi had 16 classes each wrapping exactly one service behind a static `.use()`, while the actual container was the untyped `Kernel.services` record keyed by hand-maintained `_name` strings. The one real container was anonymous; 16 non-containers carried the name.

**`XServiceProvider` was not a service provider.** A Laravel provider's defining job is `register()` — binding *into* the container. gemi's were property bags handed *to* one. The tell: `ServiceProvider.boot()` was declared abstract, implemented empty in every subsystem, and **never called by the Kernel**. Laravel's two-phase lifecycle simply didn't exist.

Two things were already right and are kept: `facades/` was a genuine facade layer, and `.use()` over `AsyncLocalStorage` was sound request-scoped resolution — it just needed to become `Container.make()`.

## What changed

| Before | After |
|---|---|
| `Kernel.services` record | `container/Container.ts` |
| — | `foundation/Application.ts` (two-phase register/boot) |
| `ServiceProvider` (config bag) | `support/ServiceProvider.ts` (`register`/`boot`) |
| `app/kernel/providers/*.ts` | `app/config/*.ts` |
| `XServiceContainer` | `AuthManager`, `MailManager`, `Translator`, `Scheduler`, … |
| `XServiceContainer.use()` | `app().make(X)` |
| `Singleton` + `SingletonService*` | deleted — `Container.singleton()` subsumes it |

Also: `FileStorage` → `Storage`, `I18n` → `Lang`; a `gemi migrate` codemod; `UPGRADE.md`; version 0.43.0.

## Two deliberate divergences

**Container tokens keep a stable `static token` string.** The issue said to retire the `_name` strings in favor of class-constructor tokens. That would have broken the build tooling: `dist/bin/gemi.js` bundles its own copy of gemi, so class objects are not identical across the two copies — and `tsc` cannot catch it. The class stays the typed handle (`app().make(MailManager)` is `MailManager`, no cast); the string carries identity. Laravel binds both `'mail'` and `MailManager::class`, so this is faithful rather than a compromise.

**Router services are `ApiRouteDispatcher` / `ViewRouteDispatcher`.** `ApiRouter`/`ViewRouter` are already public classes apps subclass to declare routes. Laravel has one `Router` and gemi has two, so the naming convention underdetermines this; the runtime services take descriptive names and the public classes are untouched.

**Hooks stay as config callbacks** (`filterRecipients`, `onLogCreated`, …) rather than provider `boot()` macros, since TS config objects hold functions natively.

## Behavioral changes worth reviewing

- **Services are now lazy.** `Route`, `Log`, and `KernelId` deliberately opt back into eager construction — without it, a route under `/__gemi__` would have become a 500 on the first request instead of a boot failure. Redis and Cron stay lazy (`new RedisClient(url)` doesn't connect).
- **`ViewRouteConfig.onRequestStart` was dead code and now fires.**
- **`Singleton` is removed from the public API** — a hard break, per the 0.43 policy.

## Verification

- `bun run build:types` — green.
- `bunx vitest` — failure set **identical to `main`**, confirmed by running both in a `git worktree` of `main`. The 7 failures are pre-existing; passing tests went 16 → 45.
- Codemod run against a materialized pre-change template: emits a **byte-identical `Kernel.ts`** and leaves **zero unhandled TODOs**. Config output differs from the hand-written template only cosmetically, except `auth.ts`, where the codemod faithfully preserves a `TS2322` that already exists on `main` and the template hand-fixed.
- New unit tests assert the two properties that didn't exist before: all `register()` run before any `boot()`, and a bound-but-unresolved singleton is never constructed.

Not verified: docs accuracy across the 19 changed doc files.

https://claude.ai/code/session_01PD1svYQLSDL7duJagM5Hqu


packages/gemi/app/App.ts
```diff
@@ -24,54 +20,49 @@ export class App {
...
       if (url.pathname.startsWith("/api")) {
-        return await ApiRouterServiceContainer.use().handleApiRequest(req);
+        return await this.kernel.apiRoutes().handleApiRequest(req);
       }
-      return await ViewRouterServiceContainer.use().handleViewRequest(req);
+      return await this.kernel.viewRoutes().handleViewRequest(req);
     });
   }
 
   public websocket: WebSocketHandler<{ headers: Headers }> = {
```
nstfkc, line 59: Remove websocket related things + BroadcastingServiceContainer.

templates/saas-starter/app/config/translation.ts
```diff
@@ -0,0 +1,15 @@
+import { defineTranslationConfig } from "gemi/i18n";
+
+import components from "@/app/i18n";
+
+export default defineTranslationConfig({
+  defaultLocale: "en-US",
+  supportedLocales: ["en-US", "tr-TR"],
+  prefetch: {
```
nstfkc, line 8: follow up issue, I want this prefetch to be handled in route handler not here.

nstfkc (commented):
a small comment.

nstfkc:
Closing as **already shipped**, not as abandoned.

This PR's content is on `release/v0.50.0-rc.2` and has been since 2026-07-28:

```
d51f8a4 Release 0.50.0-rc.2
```

`git merge-base --is-ancestor` confirms this branch is contained in `rc.2`, and in neither case in `main`. The PR stayed open afterwards, which is why it still appears as work outstanding — and, for #33, why it reports a conflict against a base whose resolution already exists one branch over.

Nothing is lost by closing: the branch is untouched and the commits are reachable from `rc.2`. What changes is that the board stops showing two open PRs describing history.

**Consequence, stated plainly:** with #30 and #33 closed, PR #45 (`feat/database-layer <- feat/orm`) proposes into a branch that no longer has a PR pointed anywhere. Merging it as-is would land the ORM on `feat/database-layer` and nowhere else — the pooling failure `plans/orm/README.md` documents. #45 needs a new base before it merges; `release/v0.50.0-rc.2` costs 2 conflicts (`bun.lock`, `packages/gemi/package.json`) against `main`'s 10. See #268, updated.

PR #33, opened 2026-07-23
Database layer on Bun's SQL client, with dialect inferred from DATABASE_URL

nstfkc:
> **Stacked on #30.** Base is `refactor/laravel-container-architecture`, not `main` — this builds on the config/provider/facade machinery from that PR. Review #30 first; this diff is 14 files.

First step toward deprecating `packages/gemi/auth/adapters` in favour of Bun's built-in database client.

## The part that isn't obvious

Bun ships **one** client that speaks SQLite, Postgres, MySQL and MariaDB, so gemi doesn't need a driver per database — and `new SQL(url)` already infers the *connection* from the URL protocol. That part is nearly free.

What Bun does **not** do is tell you which dialect it picked, and it **defaults to postgres when it can't tell**. So an unrecognised URL connects cleanly and then fails on the first query with a confusing dialect error, well away from the actual mistake.

So gemi infers the dialect a second time, independently, and `inferDialect()` **throws rather than guessing**. This is needed regardless of connection: `RETURNING`, upsert syntax, autoincrement, and boolean/timestamp types all differ per dialect, and the 22-method `IAuthenticationAdapter` uses all of them.

## What's here

| File | |
|---|---|
| `database/dialect.ts` | Inference over every URL form Bun accepts, including bare SQLite paths (`./dev.db`, `:memory:`). `isSqlite`/`isMysqlFamily` — MariaDB is wire-compatible with MySQL, so query generation collapses the two. |
| `database/DatabaseManager.ts` | `static token = "database"`; exposes `sql`, `dialect`, `url`. Bound as a **lazy singleton**, so an app that never queries never has to set `DATABASE_URL`. |
| `database/config.ts` | `defineDatabaseConfig`, `url` defaulting to `DATABASE_URL`, plus a `dialect` override for URLs behind a pooler on a custom scheme. |
| `database/DatabaseServiceProvider.ts` | Registered in `frameworkProviders` (now 15 providers / 16 services). |
| `facades/DB.ts` | `DB.sql`, `DB.dialect`, `DB.transaction()`. |
| `templates/saas-starter/app/config/database.ts` | |

```ts
const users = await DB.sql`select * from users where email = ${email}`
```

## Verification

Run against a **real SQLite database**, not just typechecked: schema creation, insert, parameterised select, and confirmed that interpolated values bind as parameters — an injection attempt inserted a literal string rather than dropping the table.

- 21 unit tests on dialect inference, including that it refuses to guess for `redis://`, `http://`, empty, and nonsense URLs.
- `bun run build:types` green.
- Full suite: same 7 pre-existing failures as `main`; passing tests 45 → 66.

## Next, not in this PR

- `SqlUserProvider` implementing the existing `IAuthenticationAdapter` with dialect-aware SQL.
- `gemi db:setup` to create the 8 auth tables (versioned migrations deferred by decision).

Together those are what let `auth/adapters/prisma.ts` be deprecated. Keeping the interface as the contract means Prisma keeps working as a custom provider throughout — the same shape Laravel uses, where `UserProvider` is a contract with `Eloquent` and `Database` implementations shipped.

**Worth deciding before that lands:** `gemi migrate` is currently the upgrade codemod from #30. Versioned migrations are deferred, but every Laravel user will expect `migrate` to mean schema. Renaming the codemod to `gemi upgrade` is free now and a breaking change once 0.43 ships.

https://claude.ai/code/session_01PD1svYQLSDL7duJagM5Hqu

### DIFF
--- a/packages/gemi/database/DatabaseManager.ts
+++ b/packages/gemi/database/DatabaseManager.ts
@@ -1,0 +1,45 @@
+import { SQL } from "bun";
+import type { DatabaseConfig } from "./config";
+import { MissingDatabaseUrlError, inferDialect, type Dialect } from "./dialect";
+
+// Wraps Bun's `SQL` client. Bun ships one client that speaks SQLite, Postgres,
+// MySQL and MariaDB, so gemi does not need a driver per database — it needs to
+// know *which* one is in use, which is what `dialect` carries.
+//
+// Like `RedisManager`, this is safe to build eagerly: Bun's client connects on
+// the first query, not at construction. It is still bound as a lazy singleton,
+// so an app that never touches the database never resolves it and never has to
+// have DATABASE_URL set.
+export class DatabaseManager {
+  static token = "database";
+
+  public readonly sql: SQL;
+  public readonly dialect: Dialect;
+  public readonly url: string;
+
+  constructor(public config: DatabaseConfig = {}) {
+    const url = config.url;
+    if (!url) {
+      throw new MissingDatabaseUrlError();
+    }
+
+    this.url = url;
+    // An explicit `dialect` wins, for URLs whose protocol we can't read (a
+    // pooler on a custom scheme). Otherwise infer, which throws rather than
+    // guessing — see the note in dialect.ts.
+    this.dialect = config.dialect ?? inferDialect(url);
+    this.sql = config.options
+      ? new SQL(url, config.options as any)
+      : new SQL(url);
+  }
+
+  // Escape hatch matching Bun's own API, so `db.query` reads like `sql` does in
+  // Bun's docs: db.query`select * from users where id = ${id}`
+  get query(): SQL {
+    return this.sql;
+  }
+
+  async close(): Promise<void> {
+    await this.sql.close();
+  }
+}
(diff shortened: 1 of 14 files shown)


nstfkc:
Closing as **already shipped**, not as abandoned.

This PR's content is on `release/v0.50.0-rc.2` and has been since 2026-07-28:

```
d51f8a4 Release 0.50.0-rc.2
```

`git merge-base --is-ancestor` confirms this branch is contained in `rc.2`, and in neither case in `main`. The PR stayed open afterwards, which is why it still appears as work outstanding — and, for #33, why it reports a conflict against a base whose resolution already exists one branch over.

Nothing is lost by closing: the branch is untouched and the commits are reachable from `rc.2`. What changes is that the board stops showing two open PRs describing history.

**Consequence, stated plainly:** with #30 and #33 closed, PR #45 (`feat/database-layer <- feat/orm`) proposes into a branch that no longer has a PR pointed anywhere. Merging it as-is would land the ORM on `feat/database-layer` and nowhere else — the pooling failure `plans/orm/README.md` documents. #45 needs a new base before it merges; `release/v0.50.0-rc.2` costs 2 conflicts (`bun.lock`, `packages/gemi/package.json`) against `main`'s 10. See #268, updated.

PR #44, opened 2026-07-27
ORM iteration 1 — skateboard: generate, and read one row

nstfkc:
Implements [`plans/orm/01-skateboard-generate-and-read.md`](https://github.com/nstfkc/gemi/blob/feat/orm/plans/orm/01-skateboard-generate-and-read.md).

`User.findMany({ where: { email } })` returns correctly typed POJOs from the
template's SQLite database, through the full pipeline, with no Prisma client at
runtime.

This is worth almost nothing as a feature and almost everything as a skeleton:
every later iteration fills in a stage this one puts in place. All six invariants
from the [plan README](https://github.com/nstfkc/gemi/blob/feat/orm/plans/orm/README.md)
are established here.

## The pipeline

```
packages/gemi/orm/
  schema.ts    runtime model metadata — the only source of identifiers
  plan.ts      QueryPlan, the plan cache, structural arg-shape hashing
  compile/     arg tree -> SQL fragments (findMany, scalar equality only)
  dialect/     the SqlDialect strategy, with a SQLite implementation
  shape.ts     precompiled per-plan row shaper
  registry.ts  lazy, name-keyed model registry
  Model.ts     the $exec choke point
  errors.ts    UnsupportedQueryError and friends
```

**Compile is split from bind and is pure.** A `Fragment` carries `binders` that
close over a *path* into the argument tree, never a value, so the same shape with
different values is byte-identical SQL and one cache entry. Placeholders use a
sentinel resolved at assembly time, so Postgres's `$1`/`$2` numbering drops in
without a shared mutable counter.

**Identifiers come only from the generated schema; values are always parameters.**
A `where` key with no matching field is an `UnknownFieldError`, never a
passthrough. Never `SELECT *` — always an explicit column list.

**Anything unimplemented throws**, naming the key: `gemi ORM does not support
'orderBy' yet (User.findMany).` Signatures accept the full Prisma argument type
from day one; capability grows underneath them.

## Generation

A Prisma generator block, not a `gemi` subcommand — gemi must not shadow the
Prisma CLI:

```prisma
generator gemi {
  provider = "gemi-orm-generator"
  output   = "../app/models/generated"
}
```

`packages/gemi/bin/orm-generator.ts` is a thin stdio process on
`@prisma/generator-helper` (a new devDependency); the emitters in `bin/orm/emit.ts`
are pure functions of the DMMF, so the whole generator is unit-testable with no
Prisma CLI and no database. Output is committed and byte-stable across runs.

The runtime imports nothing from `@prisma/*` — asserted by a test over the whole
`orm/` directory, not by convention.

## Two things the plan did not anticipate

Both found by probing Bun rather than assuming, as the plan's notes asked.

**Bun's SQLite driver binds a `Date` parameter to `NULL`.** `where: { createdAt:
date }` would have returned an empty array — a wrong answer rather than an error,
which is exactly the failure mode this iteration is built to prevent. The dialect
gained an `encode` mirror of the planned `decode`. Which encoder runs is still
fixed at compile time from the field's schema, so compile stays pure; only the
value comes from the call. (Verified alongside it: `?` positional placeholders are
correct for SQLite, and `true`/`false` already bind to `1`/`0`.)

**vitest runs under node in `packages/gemi`,** where `import { SQL } from "bun"`
cannot resolve, and switching that package's whole suite to Bun breaks 7
pre-existing tests. So the compiler, plan-cache, dialect and shaper tests are pure
and stay in `packages/gemi`; the end-to-end database test lives in the template,
whose `test` script now runs vitest under Bun (its own suite passes there
unchanged). Iteration 2's differential harness against Prisma will want a
Bun-runtime lane in `packages/gemi` — worth deciding then rather than guessing now.

## Acceptance criteria

| # | Criterion | Where |
| --- | --- | --- |
| 1 | `prisma generate` twice, zero-line diff | `templates/saas-starter/app/models/generated.test.ts`, plus emitter-level determinism in `bin/orm/emit.test.ts` |
| 2 | Reads a row with `createdAt` a `Date` and `deletedAt` `null` | `templates/saas-starter/app/models/User.test.ts` |
| 3 | Exact SQL string and exact parameter array, no database | `orm/compile/findMany.test.ts` |
| 4 | Two calls, same shape, different values -> one compile | `orm/plan.test.ts`, `User.test.ts` |
| 5 | `orderBy` throws `UnsupportedQueryError` naming it | `orm/compile/findMany.test.ts`, `User.test.ts` |
| 6 | `bun run lint` and `bun run test` | see below |
| 7 | Zero `@prisma/*` under `orm/`; type-only in generated code | `orm/runtime-isolation.test.ts`, `bin/orm/emit.test.ts` |

The acceptance query compiles to exactly:

```sql
select "id", "publicId", "name", "email", "emailVerifiedAt", "verificationToken", "locale", "globalRole", "password", "organizationId", "createdAt", "updatedAt", "deletedAt" from "User" where "email" = ?
```

**On criterion 6.** `oxlint` is clean: 0 errors, and 0 warnings from new code (the
88 remaining are pre-existing). Tests: 114 new tests in `packages/gemi` and 14 in
the template, all passing. `bun run test` in `packages/gemi` does **not** pass —
but it does not pass on the base branch either: 8 test files / 7 tests fail
identically before and after this change (router manifest tests, plus suites that
cannot import `bun` under node). This branch adds no new failures. Type checks are
also clean for new code in both packages, against a similarly pre-broken baseline.

## Notes for review

- `packages/gemi/orm/fixtures.ts` is test data, not public surface — nothing in
  `orm/index.ts` imports it, so it never reaches a bundle.
- The generator resolves by bin name from `node_modules/.bin`, which in this repo
  symlinks to built output. `bun run build:bin` in `packages/gemi` is a
  prerequisite for `prisma generate` here; the idempotence test says so explicitly
  when the link is missing.
- `prepare-bin.ts` now handles both bin entries and is idempotent about the
  shebang.
- `templates/saas-starter/prisma/dev.db` is committed and is never written to —
  the end-to-end test copies it to a temp file and seeds its own row.
- Out of scope, per the plan: every other `where` operator, `orderBy`, pagination,
  `select`, `include`, the other eleven operations, Postgres, writes, transactions,
  policies, benchmarks.


packages/gemi/orm/dialect/sqlite.ts
```diff
@@ -0,0 +1,101 @@
...
+        return typeof value === "string" ? value : JSON.stringify(value);
+      default:
+        // BigInt passes through. Bun's driver truncates integers above 2^53 on
+        // the way in, which affects a raw `db.sql` query identically — it is
+        // not something the compiler can correct at the parameter level.
+        return value;
+    }
+  }
+
+  needsDecode(field: FieldSchema): boolean {
```
nstfkc, line 90: **`Decimal` silently bypasses the encode/decode mirror.**

`needsDecode` returns `false` for `Decimal`, and `encode`'s switch above has no `Decimal` case either — but `emit.ts` maps `decimal` into `SCALAR_TYPES` without complaint, so a model with a `Decimal` field generates cleanly and then returns the driver's raw number typed as `Prisma.Decimal`. A `Prisma.Decimal` passed as a filter value likewise reaches Bun unconverted.

That is exactly the failure mode this file's own comment names on the `DateTime` case: a wrong answer rather than an error. The template schema has no `Decimal`, so nothing here catches it.

The generator already refuses scalar lists with `UnsupportedSchemaError` for precisely this reason. `Decimal` deserves the same refusal until a dialect can decode it. `Bytes` has a weaker version of the same problem — Prisma returns a `Uint8Array` and Bun's BLOB handling is close enough that it may be fine, but it is unasserted either way.

nstfkc (commented):
Reviewed against the checked-out branch and ran the suites rather than taking the description's word for them.

**Verified**

- `packages/gemi` ORM + emit tests: 114 pass across 6 files.
- Template `app/models`: 11 pass, including the real `prisma generate` idempotence run — the bin is linked and built here, so that test genuinely executed rather than short-circuiting on the missing-link assertion.
- Full `packages/gemi` suite: 8 files / 7 tests fail, all in `app/create*`, `services/router/*`, and the redis / `parseTranslation` bun-import suites. None is touched by this PR. The disclosure on criterion 6 is accurate.

**On the design**

Compile-is-pure / bind-sees-values is the right cut, and the `PARAM_MARKER` sentinel avoids the shared-counter trap that bites Postgres ports later. "Anything unimplemented throws, naming the key" is the correct default for an ORM that is one operation wide. Asserting the `@prisma/*` boundary over the whole directory rather than per-file is the right shape for that invariant. The comments explain why rather than what.

The injection story holds: identifiers come only from `schema.fields` / `schema.table` through `quoteIdent`, an unmatched `where` key raises `UnknownFieldError` rather than passing through, values are always parameters, and `render` asserts placeholder count against binder count.

**Worth acting on**

One finding is worth fixing before merge — `Decimal` silently bypasses the encode/decode mirror, which is the same wrong-answer-rather-than-an-error class as the `Date` -> `NULL` discovery the iteration is built around. The fix is a `throw` in the generator. Two others (array shapes in the plan cache key, and the four non-ORM exports) are worth deciding now even if the change lands later. Details inline.

packages/gemi/bin/orm/emit.ts
```diff
@@ -0,0 +1,330 @@
...
+export const GENERATED_FILES = ["schema.ts", "models.ts", "index.ts"] as const;
+export type GeneratedFile = (typeof GENERATED_FILES)[number];
+
+const HEADER = `// Generated by the gemi ORM generator. Do not edit.
+//
+// Re-run \`prisma generate\` (or \`prisma migrate dev\`) to refresh this file.
+// It is committed on purpose: diffs stay reviewable and CI needs no codegen step.
+`;
+
+const SCALAR_TYPES: Record<string, ScalarType> = {
```
nstfkc, line 29: This table accepts `decimal` and `bytes`, but `SqliteDialect` handles neither in `encode`/`decode`/`needsDecode` — see the note there. The `isList` check further down already establishes the pattern for this: refuse in the generator rather than let the runtime return a value that diverges from Prisma's result shape.

A `Decimal` entry that throws `UnsupportedSchemaError` would keep the promise the iteration makes everywhere else.

packages/gemi/orm/dialect/sqlite.ts
```diff
@@ -0,0 +1,101 @@
...
+    // The driver reports a missing value as `null`; Prisma does too, so there
+    // is nothing to convert. `undefined` can only appear if a column was not
+    // selected, which the explicit column list makes impossible.
+    if (value === null || value === undefined) return null;
+
+    switch (field.type) {
+      case "DateTime":
+        return toDate(value);
+      case "Boolean":
+        return typeof value === "boolean" ? value : value !== 0;
```
nstfkc: `value !== 0` is strict, so a bigint zero decodes to `true` (`0n !== 0` is `true`). Latent rather than live — Bun returns numbers for SQLite integers — but `Number(value) !== 0` closes it for free.

Two neighbours in the same switch: `BigInt(value as string)` throws on a float, and `toDate` returns a silent `Invalid Date` for a string that matches neither the integer path nor `SQLITE_TEXT_TIMESTAMP`.

packages/gemi/orm/plan.ts
```diff
@@ -0,0 +1,112 @@
...
+export function canonicalShape(value: unknown): string {
+  if (value === null) return "null";
+  if (value === undefined) return "undefined";
+
+  const type = typeof value;
+  if (type !== "object") return type;
+
+  if (value instanceof Date) return "date";
+  if (Array.isArray(value)) {
+    return `[${value.map(canonicalShape).join(",")}]`;
```
nstfkc, line 72: **Fine today, unbounded in iteration 2.**

Descending into arrays element-wise makes `[number,number]` and `[number,number,number]` distinct cache keys. Right now nothing reaches the cache with an array — unknown operators throw before `cache.set`, which `plan.test.ts` correctly asserts — so the key space is bounded by field subsets.

The moment `in: [...]` lands, every distinct array length from a user-supplied filter list becomes a permanent entry in a module-global `Map` with no eviction, keyed off request input. Worth collapsing a homogeneous array to a length-independent token, or capping the cache, before that iteration rather than after it ships.

packages/gemi/package.json
```diff
@@ -33,6 +34,7 @@
     "./container": "./container/index.ts",
     "./foundation": "./foundation/index.ts",
     "./database": "./database/index.ts",
+    "./orm": "./orm/index.ts",
     "./support": "./support/index.ts",
```
nstfkc, line 38: **New public surface beyond what the ORM needs.**

Only `./orm` is required by generated code. `./container`, `./foundation`, `./database` and `./support` are load-bearing for `User.test.ts` alone.

Publishing is safe — I checked `build-publish.ts`, and `toPublishExports` rewrites all five mechanically with no override needed. The concern is the surface decision itself: four new entry points are much harder to withdraw than to add, and they arrive inside an ORM PR where a reviewer isn't looking for them. Worth being deliberate about rather than incidental.

packages/gemi/package.json
```diff
@@ -70,6 +72,7 @@
   },
   "devDependencies": {
     "@babel/parser": "^7.25.4",
+    "@prisma/generator-helper": "^6.5.0",
```
nstfkc, line 75: Correct as a `devDependency` given `build:bin` bundles it — verified: 56 KB `dist/bin/orm-generator.js`, and `prisma generate` runs against it here.

The consequence is that the DMMF reader version is frozen into the published artifact, so an app on a much newer Prisma gets a mismatched helper with no signal at all. A documented supported-Prisma range, or a `peerDependency` on `prisma`, would make that legible.

packages/gemi/orm/compile/findMany.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+const SUPPORTED_ARGS = new Set(["where"]);
+
+export function compileFindMany(
+  schema: ModelSchema,
+  args: any,
+  dialect: SqlDialect,
+): QueryPlan {
+  if (args !== undefined && args !== null) {
+    for (const key of Object.keys(args).sort()) {
```
nstfkc, line 34: Minor: `Object.keys` without a plain-object check means `findMany("x" as any)` reports `does not support '0'`. Types make it unlikely; the error is just misleading if it happens.

packages/gemi/orm/dialect/sqlite.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export class SqliteDialect implements SqlDialect {
+  readonly name: Dialect = "sqlite";
+
+  quoteIdent(name: string): string {
+    // Identifiers only ever come from the generated schema, so an embedded
+    // quote is not an attack vector — but escaping it costs nothing and keeps
+    // the invariant "the compiler cannot emit broken SQL" unconditional.
+    return `"${name.replace(/"/g, '""')}"`;
```
nstfkc, line 59: The comment argues NUL can't appear here, and it's right — but NUL is also the parameter sentinel in `fragment.ts`, so this is the one identifier character that could corrupt parameter alignment rather than just producing broken SQL. A `throw` on `name.includes("\u0000")` would make the invariant unconditional instead of argued, which is the standard the rest of this file holds itself to.

packages/gemi/orm/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  ModelNotRegisteredError,
+  UnknownFieldError,
+  UnsupportedQueryError,
+} from "./errors";
+
+export {
+  canonicalShape,
+  clearPlanCache,
+  getOrCompile,
+  planCacheStats,
```
nstfkc, line 27: `clearPlanCache` and `planCacheStats` are test seams on the public surface. `User.test.ts` needs them across the package boundary, so there may be no better option today — but they'll be permanent once apps can import them.

packages/gemi/bin/orm/emit.ts
```diff
@@ -0,0 +1,330 @@
...
+  byRelationName: Map<string, { model: string; field: DMMF.Field }[]>,
+): RelationSchema["joinTable"] | undefined {
+  if (!field.isList || !field.relationName) return undefined;
+  if ((field.relationFromFields ?? []).length > 0) return undefined;
+
+  const sides = byRelationName.get(field.relationName) ?? [];
+  if (sides.length !== 2) return undefined;
+  if (!sides.every((side) => side.field.isList)) return undefined;
+
+  const [first, second] = [model.name, field.type].sort();
```
nstfkc, line 165: For a self-referential implicit m2m, `model.name` and `field.type` are the same string, so `a === b` and the two sides become indistinguishable. Unused until iteration 3 — worth a note there rather than a fix here.

packages/gemi/orm/dialect/sqlite.ts
nstfkc: Fixed in 671f26a — and you were right that this is the same class as the `Date` -> NULL case.

Probed it rather than reasoning about it: SQLite stores `Decimal` as a REAL and Bun returns a JS `number`, so a `Decimal` field returned `1.005` typed as `Prisma.Decimal`. A float pretending to be an arbitrary-precision decimal, losing exactly the precision the type exists to keep. The generator now refuses it with `UnsupportedSchemaError`.

`Bytes` went the other way. Probed that too: Bun returns a `Uint8Array` for a BLOB, which is what Prisma 6 returns, and equality filtering round-trips correctly. So it is genuinely fine — but you were right that it was unasserted, so it now has a test instead of an assumption.

packages/gemi/bin/orm/emit.ts
nstfkc: Fixed in 671f26a. Kept `decimal` in `SCALAR_TYPES` and added an `UNSUPPORTED_SCALARS` table carrying the reason, so the error explains the precision problem rather than falling through to the generic "does not know how to map" message.

`bytes` stays in the table — verified against the driver, see the reply on sqlite.ts.

packages/gemi/orm/dialect/sqlite.ts
nstfkc: All three fixed in 671f26a.

`Number(value) !== 0` for the bigint zero, as you suggested. The other two were worse than the report implied once I looked: `BigInt(3.5)` raised a RangeError naming neither field nor column, and malformed JSON propagated a bare SyntaxError. `toDate` was the real one — an `Invalid Date` whose every field reads NaN would have travelled a long way before anyone noticed.

All four now raise a `DecodeError` naming the column, the value and the expected type. It lives in `errors.ts` rather than beside SQLite, since Postgres will need it, and is exported.

packages/gemi/orm/plan.ts
nstfkc: The diagnosis is right; the suggested fix would break, so I did the other one.

Collapsing a homogeneous array to a length-independent token is wrong precisely when `in` lands: `in: [a, b]` compiles to `in (?, ?)` and `in: [a, b, c]` to `in (?, ?, ?)`. A collapsed key would hand a cached plan the wrong number of placeholders — a worse bug than the leak. Length has to stay in the shape.

So the real exposure is the one you named second: unbounded growth in a module-global `Map` keyed off request input. That needs an eviction policy, not a coarser key. Recorded both halves as a warning at the site in 671f26a so iteration 2 cannot reach for the collapse instead.

packages/gemi/package.json
nstfkc: Checked this against the diff — only `./orm` is new. `./container`, `./foundation`, `./database` and `./support` all pre-date this branch; they appear in the diff as context around the added `./orm` line.

`git diff 990b719..HEAD -- packages/gemi/package.json` shows the one addition. So the surface decision here is a single entry point, which generated code does require.

packages/gemi/package.json
nstfkc: Good point — the freeze is real and was invisible. Added a Prisma major-version check in 671f26a: the generator compares `options.version` (what the CLI reports) against the major the bundled helper was built for, and warns on a mismatch.

Warn rather than throw, deliberately: a mismatch is usually harmless, and refusing to generate would be a worse failure than a noisy one. A `peerDependency` on `prisma` would be stronger but also constrains apps that never use the ORM, which is most of them today.

packages/gemi/orm/compile/findMany.ts
nstfkc: Fixed in 671f26a — plain-object guard, with a test asserting the error does *not* mention `'0'.

packages/gemi/orm/dialect/sqlite.ts
nstfkc: Fixed in 671f26a. You are right that this is the one identifier character with consequences beyond broken SQL, and that arguing the invariant is a weaker standard than asserting it. `quoteIdent` now throws on an embedded NUL.

packages/gemi/orm/index.ts
nstfkc: Kept, but documented rather than left implicit — 671f26a.

On reflection `planCacheStats` is not only a test seam: `compiles` climbing in step with `hits` means some caller is varying its argument shape per request, which is the cheapest possible signal for that class of bug and worth graphing. `clearPlanCache` is needed by anything reloading the generated schema in a running process, not just by tests. Both now say so in their doc comments, so the next reader does not have to guess whether they are safe to depend on.

packages/gemi/bin/orm/emit.ts
nstfkc: Agreed, and noted at the site in 671f26a rather than fixed: both names are the same string on a self-relation, so `a === b` and the record cannot say which column holds which end. Prisma disambiguates by field order, not by model name — the note says that, so iteration 3 has the pointer rather than the puzzle.

nstfkc:
Pushed 671f26a. Summary of what changed and what did not.

**Fixed**

- **`Decimal`** — refused in the generator with `UnsupportedSchemaError`. Probed it first: SQLite stores it as a REAL, Bun returns a JS number, so the field returned a float typed as `Prisma.Decimal`. Same class as the `Date` -> NULL case, as you said.
- **`Bytes`** — probed and kept. Bun returns a `Uint8Array` for a BLOB, which is what Prisma 6 returns, and equality round-trips. Now asserted rather than assumed.
- **Decode hardening** — bigint zero decoding as `true`, `BigInt(3.5)`, malformed JSON, and the `Invalid Date` all now raise a `DecodeError` naming the column, value and expected type. Moved to `errors.ts` and exported, since Postgres will need it.
- **`quoteIdent`** — throws on an embedded NUL, so the sentinel invariant is asserted rather than argued.
- **`findMany` args** — plain-object guard, with a test that the error does not blame a character index.
- **Prisma version drift** — the generator warns when the CLI's major differs from the one the bundled helper was built against.

**Decided, not changed**

- **Plan cache** — the diagnosis is right but the suggested fix would break. Collapsing a homogeneous array to a length-independent token is wrong exactly when `in` lands, since `in (?, ?)` and `in (?, ?, ?)` are different SQL and a collapsed key would hand a plan the wrong placeholder count. The real exposure is unbounded growth keyed off request input, which needs eviction rather than a coarser key. Both halves recorded at the site so iteration 2 cannot reach for the wrong one.
- **`clearPlanCache` / `planCacheStats`** — kept and documented. `compiles` climbing with `hits` is a real observability signal, not only a test seam.
- **Self-referential implicit m2m** — noted at the site for iteration 3, including that Prisma disambiguates by field order rather than model name.

**One correction**

The four extra exports are not new. Only `./orm` is added; `./container`, `./foundation`, `./database` and `./support` pre-date this branch and appear in the diff as context. `git diff 990b719..HEAD -- packages/gemi/package.json` shows the single addition.

**Verification**

`packages/gemi` ORM + emit: 123 pass across 6 files, up from 114. Template: 14 pass, including the real `prisma generate` idempotence run. The regenerated artifacts are byte-identical after the generator rebuild, as expected — the template schema has no `Decimal`. Full `packages/gemi` suite still fails the same 8 files / 7 tests as the base branch, with 189 passing (was 167). Lint 0 errors, 0 warnings from new code. Type checks clean for new code in both packages.

PR #45, opened 2026-07-27
Feat/orm

nstfkc:


templates/saas-starter/app/models/User.ts
```diff
@@ -0,0 +1,27 @@
...
+// the registry by name, so whatever is registered under "User" is the class that
+// runs for every nested `include` — and if that is the generated base while your
+// policy lives here, the policy applies to root queries and is skipped inside
+// includes. Scoped one way, unscoped the other, with nothing to notice it.
+//
+// `Model.$exec` raises `UnregisteredPolicyClassError` if a class carrying
+// policies is not the registered one, so forgetting this line fails loudly
+// rather than leaking — but the line belongs next to every subclass regardless,
+// because the same reasoning will apply to observers and hooks.
+register("User", User);
```
nstfkc, line 45: Can the framework auto register the models if the models directory is guaranteed?

nstfkc:
## Review — `feat/orm`

Reviewed the whole diff at file level and read the following closely: `compile/{fragment,where,read,select,orderBy,nested-writes,lateral,plan-relations}.ts`, `dialect/{index,postgres}.ts`, `plan.ts`, `context.ts`, `policy.ts`, `Model.ts`, plus the generator and auth adapter at a scan. Ran `bun test packages/gemi/orm packages/gemi/bin/orm` (740 pass) and `templates/saas-starter/app/models/policies.test.ts` on SQLite (45 pass, 1 skipped).

### What it is

A Prisma-compatible ORM built on Bun's SQL client: schema generation from `schema.prisma`, a pure argument-shape → SQL compiler with an LRU plan cache, two dialect implementations behind a `SqlDialect` strategy, relation loading under two interchangeable strategies (batched / lateral+`json_agg`), ambient transactions on `AsyncLocalStorage`, nested writes, row provenance + `save()`, and a per-model policy layer that fires on nested reads as well as root ones.

The architecture holds up under reading. Three invariants are actually enforced rather than aspirational: `$exec` really is the only door to the database; compilation really is a function of argument *shape* with every value bound through a `Binder`; and the dialect seam has no `if (dialect === "postgres")` leaking into the compiler. I looked specifically for SQL text built from user-controlled data and did not find any — every interpolation is a `quoteIdent`ed schema identifier or a keyword from a closed set, `render()` is the single choke point for the parameter ceiling, and NUL is rejected at `quoteIdent` because it is the placeholder sentinel. The comment density is unusual and mostly earns itself: the notes on `Buffer` vs `Uint8Array` for `Bytes`, on `bigint` losing its low bit through `json_build_object`, and on the Postgres simple-vs-extended protocol `timestamp` divergence are the kind of thing that is only ever learned the expensive way.

---

## Findings

### 1. `create` / `createMany` with an `include` of a policied relation — confirmed against SQLite

`Model.$exec` calls `applyNestedPolicies(schema, effective, op, …)` for every operation, and the walk passes the **root** operation down to each nested node (`policy.ts:1000`). A nested `include`/`select` node is always a *read*, but under a root `create` it is scoped as if it were an insert, and `applyPolicies` takes its `INSERTING` branch (`policy.ts:473`). Two outcomes, both wrong:

**(a) Child policy has `scope` and no `onCreate` → hard failure.**

```ts
// Account has { scope: ctx => ({ organizationId: ctx.user.organizationId }) }
await Model.asUser(alice, () =>
  User.create({ data: { email }, include: { accounts: true } })
)
```

```
UnsupportedQueryError: gemi ORM does not support 'create' yet (Account.create).
Account has a policy that scopes reads but no 'onCreate', and create has no
where clause for a scope to narrow. …
  at assertCreateCovered  packages/gemi/orm/policy.ts:701
  at applyPolicies        packages/gemi/orm/policy.ts:474
  at applyNestedPolicies  packages/gemi/orm/policy.ts:998
  at $exec                packages/gemi/orm/Model.ts:505
```

Nothing is being created on `Account` — it is being read back on the new row — and the error names an operation the caller never asked for.

**(b) Child policy has `scope` + `onCreate` → the child's read scope is silently dropped.**

The `INSERTING` branch `continue`s past `withScope`, then the `CREATING` pass runs `onCreate` over the include node, so the node comes out carrying a `data` key instead of a `where`:

```
findMany include: {"accounts":{"where":{"organizationId":7}}}
create   include: {"accounts":{"data":{}}}              // scope gone
softDeletes create include: {"accounts":{"data":{}}}    // deletedAt: null gone
```

The `create` case raises no error (verified end to end), so this is the quiet half. The blast radius is small today because the included children belong to the row just created — but a `softDeletes` scope disappearing from a nested read is exactly the class of thing this layer exists to make impossible, and the same code path is what would carry a real scope if `include` on writes ever widens.

Suggested shape of the fix: nested `include` / `select` / `_count` / `orderBy` / relation-filter nodes are reads by construction, so `applyNestedPolicies` should scope them under a read operation regardless of the root op — pass e.g. `"findMany"` into the nested `policyContext` rather than the caller's `op`. That also makes a policy branching on `context.operation` see the truth (`Model.ts` already takes the honest-reading position for the `upsert` → find/create/update split and for `delete`'s pre-read; this is the same argument). Whichever way it goes, it wants a test: `policy.test.ts`'s `applyNestedPolicies` block currently only exercises read operations, which is why this got through.

### 2. Lateral strategy appears to drop a nested `_count` — not verified, no Postgres here

`resolveSelection` accepts `_count` inside a relation node's `select` (`select.ts:80`, counted as a selection so the node is not "empty"). `lateral.ts`'s `decline()` refuses a node with a nested `include`, or a `select` naming a *relation* — but `_count` is neither, so the node folds. `jsonObject()` then projects only the child's scalar `FieldSchema`s, so the folded JSON has no `_count` key and `buildDecoder` cannot invent one. Under the batched strategy the same node goes through the child's own `$exec`, where `planRelationCounts` runs and `_count` is present.

If that reading is right, `include: { accounts: { select: { id: true, _count: { select: { … } } } } }` returns rows with `_count` on SQLite and without it on Postgres — a silent per-dialect divergence, and one the differential harness would not necessarily catch unless the matrix has this exact shape. Cheapest fix is one more line in `decline()` (`if (select[COUNT_KEY]) return "nested _count"`), which costs a round trip on a rare shape and cannot be wrong. Worth confirming on a real Postgres before acting.

### 3. `take` / `skip` are not validated as numbers — confirmed

`pagination()` decides the sign at compile time with `typeof take === "number"` (`read.ts:302`) but binds with `Math.abs(Number(callArgs?.take))` (`read.ts:310`). A non-number that survives to runtime silently changes the query's meaning:

```
take: -5   -> … order by "id" desc limit ?   [5]     // last 5, correct
take: "-5" -> … order by "id" asc  limit ?   [5]     // FIRST 5
take: "abc"-> … order by "id" asc  limit ?   [NaN]
```

The types stop this statically, and the same argument the `contains`-must-be-a-string guard makes (`where.ts:582` — "the types catch the static case but not a value that arrives dynamically") applies here verbatim: `take` is the single most likely argument to arrive off a query string. A `typeof`/`Number.isInteger` check in `assertArgs` would close it.

### Smaller things

- `where.ts:195` — `const path = …` in `compileRelationFilter` is unused; `readOperators` has its own.
- `lateral.ts:220` — `` `'${field.name}', …` `` is the one place a schema string becomes a SQL *literal* rather than a quoted identifier. Not reachable today (Prisma identifiers can't contain a quote), but the generator does not assert it either — an `UnsupportedSchemaError` on a name that isn't `/^[A-Za-z][A-Za-z0-9_]*$/` in `bin/orm/emit.ts` would make that a property rather than a coincidence.
- PR body is empty for a 117-file, ~39k-line change, and the base is `feat/database-layer` rather than `main` — worth stating the merge order and pointing at `plans/orm/README.md` so a reader lands somewhere. The ten `orm`-labelled issues (#62–#71) covering the deferrals are a good pattern; the README's two open decisions (MySQL, and the `omit`/`distinct`/`cursor` group) are the ones I'd want resolved before this is called done.

### What I did not review in depth

`compile/write.ts` (1316 lines), `compile/plan-relations.ts` (1189), `bin/orm/emit.ts`, `provenance.ts`, `shape.ts` and the auth adapter got a scan for injection and obvious structure only, not a line-by-line read. The differential harness and the Postgres-only paths were not exercised — everything above is SQLite plus unit tests.

nstfkc:
### Follow-up: finding 2 is confirmed, and it is worse than I flagged it

I said "not verified, no Postgres here" and left it as a code-reading. I've since run it against Postgres 16 via `app/models/postgres.sh`. It reproduces, and **lateral is the default on Postgres** (`compile/strategy.ts:46`), so this is the default path in production rather than an opt-in one.

```ts
const query = {
  include: {
    users: { select: { id: true, _count: { select: { accounts: true } } } },
  },
}

await Organization.findMany(query, { strategy: "batched" })
// users: [{"id":1,"_count":{"accounts":2}}]      ✓

await Organization.findMany(query, { strategy: "lateral" })
// users: [{"id":1}]                              ✗ no _count key at all
```

Not a wrong number — the key is absent, so `row._count.accounts` is a `TypeError` on `undefined`. And because SQLite always batches, it works in development and fails in production: the same dev/prod asymmetry the `Buffer`/`Uint8Array` note in `PostgresDialect.decode` and the `timestamp` protocol note both call out as the failure mode worth spending effort on.

The mechanism is as described: `decline()` (`compile/lateral.ts:159`) checks a node's `select` for keys in `child.relations`, and `_count` is not one, so the node folds; `jsonObject` then projects only scalar `FieldSchema`s and there is nothing left to carry the count. `resolveSelection` explicitly allows `_count` at that position (`compile/select.ts:80`), so nothing upstream rejects it either.

Two notes on scope:

- `grep -n _count` finds nothing in `app/models/differential.ts` or `app/models/lateral.test.ts`. So the claim in `strategy.ts` that "the **full** nested differential matrix runs against it" is true of what the matrix contains, and the matrix has no `_count` inside a relation — which is exactly why this shape got through. A row in the matrix for it would be worth more than the one-line fix.
- The fix itself is still one line in `decline()` — `if (select[COUNT_KEY]) return "nested _count"` — costing a round trip on a rare shape and unable to be wrong. Folding it properly means a second aggregate inside the lateral subquery, which is a real feature and probably belongs with #62 rather than here.

I also did the line-by-line read of `compile/write.ts` and `compile/plan-relations.ts` that I'd listed as "scanned only". Nothing further to report from either — the two shapes I went looking for are both already closed and commented: `createManyColumns` builds its grid in schema order rather than first-row key order (a per-row key-order mismatch would bind values into the wrong columns silently), and `withConflictKeyCheck` compares the upsert's conflict key at *bind* time rather than compile time, with `sameEncoded` handling the `Date`/`Uint8Array` cases where `===` would refuse a correct call. `attachRelations` skipping any plan carrying `root` is the thing that keeps the folded path from also issuing the batched query, and `lateral.ts`'s `load` throws rather than falling back if that skip ever regresses, which is the right way round.

So findings 1 and 3 stand as posted, and 2 is now confirmed rather than suspected.

nstfkc:
### Finding 4: a model with a composite `@@id` cannot be read, updated or deleted at all

Continuing through the files I'd listed as unreviewed — `bin/orm/emit.ts`, `shape.ts`, `provenance.ts`. The first turned this up.

`compile/unique.ts`'s `uniqueKeys()` unions the primary key with `schema.uniques` (`unique.ts:56`), so `matchUniqueKey` accepts a compound `@@id` under Prisma's `a_b` name. `compile/where.ts`'s `compileCompoundKey` searches **`schema.uniques` alone** (`where.ts:367`) — and the generator does not put an `@@id` group there. Verified by adding one model to `templates/saas-starter/prisma/schema.prisma` and running the real generator:

```prisma
model Membership {
  userId Int
  orgId  Int
  role   String
  @@id([userId, orgId])
}
```

```jsonc
// app/models/generated/schema.ts
"primaryKey": ["userId", "orgId"],
"uniques":    []                     // <- nothing for compileCompoundKey to find
```

Compiling against that generated artifact, **both** spellings fail, and each error points at the other:

```
where: { userId_orgId: { userId: 1, orgId: 2 } }
  findUnique → 'userId_orgId' is not a field on model Membership.
               Known fields: userId, orgId, role.
  update     → same
  delete     → same

where: { userId: 1, orgId: 2 }
  findUnique → findUnique needs a unique field.
               Membership declares: userId_orgId.        <- which throws the error above
```

So there is no working form. `Model.save()` goes the same way, since it builds `where` from `record.key`, which is the flat primary key. `findFirst` / `updateMany` / `deleteMany` still work with the flat form and are the only way to touch such a row today — at the cost of the single-row guarantee and the P2025 that `update` and `delete` raise.

The contrast that isolates it: a composite **`@@unique`** works correctly, because the generator does emit those into `uniques`. Same generated file, same code path:

```
SocialAccount  primaryKey: ["id"]  uniques: [["username","provider"]]
findUnique({ where: { username_provider: {...} } })
  → select … where ("username" = ? and "provider" = ?) limit ?     ✓
```

Two notes on scope. `plans/orm/02-scooter-query-surface.md:27` says compound keys "are what `findUnique` must accept", and `docs/orm.md`'s **Not in scope** list does not mention composite primary keys — so this reads as intended-supported rather than a deferral. And the shape is not exotic: a join model with `@@id([aId, bId])` is the standard way to write an explicit many-to-many, which is the thing #66 tells people to use while implicit m-n writes are unimplemented. The template schema has no `@@id`, which is why nothing catches it.

The fix looks like one line — `compileCompoundKey` should search `uniqueKeys(schema)` rather than `schema.uniques`, i.e. the same union `matchUniqueKey` already consults, which is what makes the two agree by construction instead of by coincidence. The flat form staying rejected is correct: Prisma's `WhereUniqueInput` requires the compound name there too.

Nothing to report from `shape.ts` or `provenance.ts`. One dead line worth a glance: `assertNothingUnfetched` (`provenance.ts:147`) checks `if (name in schema.relations) continue` after already requiring `name in schema.fields`, and a name cannot be in both — the relation keys it means to skip are excluded by the `schema.fields` check one line earlier. Harmless, but it reads as a guard that is doing something.

That closes my pass over this PR: everything in `packages/gemi/orm/` and `packages/gemi/bin/orm/` has now been read rather than scanned.

nstfkc:
## `feat/orm` at `00d5075` — state of the branch, and six guards that do not fire

The review stack has landed. I re-ran everything against the merged result rather than trusting the per-branch numbers, since two of the three breaks in this series existed only in the combination.

**The branch is green.**

```
tsc --noEmit                     clean          (#159 closed by #160)
vitest run orm database          1367 passed    (#166 closed by #167)
whole package                    7 failed | 1425 passed
                                   the six pre-existing router/translation files
template  app/models             614 passed | 32 skipped, 0 failed
```

Nothing else came out of merging ~20 branches at once, which is the better outcome than two instances in a row suggested.

### Six findings merged unaddressed

Each raised on its own PR, each re-verified here by breaking the thing the guard exists to catch and running the full suite. All six leave **1367 passing**:

| what I broke | should fail | does |
| --- | --- | --- |
| `mode` removed from `LITERAL_KEYS` | plan-key invariant 3 | passes |
| `_sum` removed from `LITERAL_KEYS` | same | passes |
| `OPS[dialect.name]` deciding the operator | the dialect seam check | passes |
| `compile/select.ts` re-exporting `../policy` | the transitive import check | passes |
| the `_count.select` null guard inverted | the malformed-passthrough tests | passes |
| `User.ts`/`Post.ts` in `generated/`, importing each other | the artifact's no-import-graph check | passes |

None is a live defect — the shipped behaviour is right in every case. They are guards that would not fire, which matters because each was added specifically to catch its case, and three of them are the last line on a hazard the code comments describe as severe: a plan shared between `ilike` and `like`, a `where` scope dropped, the ESM bootstrap cycle.

The two cheapest are worth doing before this merges:

- **`mode`** — its invariant-3 pair is `{contains}` against `{contains, mode}`, distinguished by key *presence*, so it passes with `mode` non-literal. `mode: "insensitive"` against `mode: "default"` is the pair that bites: same key, different SQL, one plan entry.
- **`_count: { select: null }`** — `Object.keys(null)` throws where `Object.keys(5)` returns `[]`, so the null disjunct of that guard is the one not subsumed by the loop below it. One case covers it.

The other four each need a few lines and are written up on #123, #127, #129 and #131 with the reproduction and a verified fix.

### One thing the merge fixed that I should record

`#165`'s dead-export check caught a probe of mine before the check I was actually testing did — I had exported a helper to trigger the dialect-seam evasion, and the new test named the export instead. That is the check working, and it cost me a wrong conclusion for one iteration: I read the failure as the seam check catching the lookup, and it was not. Re-run without the export, the seam evasion passes.

nstfkc:
## Update at `b47a9c0` — CI has landed, and it does not catch the six

The CI stack (#169, #171, #173, #175) merged. Re-ran everything on the merged result, since the last two breaks on this branch existed only in a combination.

**Green, and the merge introduced nothing:**

```
tsc --noEmit                     clean
vitest run orm database          1368 passed        (what CI runs)
oxlint orm --deny-warnings       exit 0
template  app/models             614 passed | 32 skipped, 0 failed
```

Third clean four-PR merge in a row, after two that broke — and #159's and #166's class is now the thing CI exists to catch, so it should stay that way.

**The six findings are unchanged, and CI does not close them.** Each probe below is the guard's own subject, and each passes the exact command the workflow runs:

```
mode removed from LITERAL_KEYS                    1368 passed
_count.select null guard inverted                 1368 passed
compile/select.ts re-exporting ../policy          1368 passed
OPS[dialect.name] deciding the operator           1368 passed
```

plus the two template-side ones — per-model modules in `generated/` importing each other, and the two Postgres failure rows that cannot fail.

That is the thing worth being explicit about now that CI is in: the workflow makes a *red* branch impossible to miss, which is what it was for. It does not make a guard that never fires fire. The six are the same six, and they are the last line on the hazards the surrounding comments call severe — a plan shared between `ilike` and `like`, a dropped `where` scope, the ESM bootstrap cycle.

The two cheapest remain `mode` and `_count: { select: null }`; both are one corpus line or one test case, both written up with a verified fix on #123 and #139.

**One correction of my own carried forward.** I argued on #161 and #170 that CI's `prisma generate` step should be removed because it turns the artifact diff test into a no-op. Measuring #178 showed that step is also what makes the differential sensitive to a wrong generator at all — removing it restores one check and removes another. `git diff --exit-code` on `app/models/generated` immediately after the step is the version that does not trade one for the other.

nstfkc:
## Update at `4037fb7` — #182's fix merged with two stale counts still in source, and one is my miss

Branch is healthy: `tsc --noEmit` clean, `vitest run orm database` **1373 passed**.

#182 closed with the count corrected everywhere it enumerated. Re-swept the merged branch — this time across **all** of `packages/gemi`, not just `orm/`:

```
fourteen    0
fifteen     2   docs/orm.md, docs/llms-full.txt          <- correct
twelve      4   plans/ ×3                                <- deliberate
                packages/gemi/bin/orm/emit.ts:439        <- not deliberate
thirteen    2   plans/ ×1                                <- deliberate
                packages/gemi/orm/compile/unique.ts:13   <- not deliberate
```

```
bin/orm/emit.ts:439      * not have. These classes are constructed by `wrap` and by nothing else — the
                         * twelve operations return plain objects — so the lie is confined to a

orm/compile/unique.ts:13 * that is not one of the thirteen. `UnsupportedQueryError` documents those
```

**The second one is mine.** My sweep on #182 covered `packages/gemi/orm`, `docs/` and `plans/orm/` — I never looked in `packages/gemi/bin/`, so I reported "two the table misses" when there were three. That is the same error I have now flagged in three PRs and made twice myself on this branch: a boundary picked without checking what it excludes.

`bin/` is not an obscure corner for this — it is where the generator lives, and #178 and #185 both landed changes there in the last few iterations.

Neither is caught by anything. #182's guard checks the documented **list** against the compiler's set, which is the right design and the reason the docs are now correct and will stay so; it does not read source comments, and nothing else does either.

The fix is the phrasing the PR already chose everywhere else — "the operations", not a number — applied to two more lines. Worth doing on this branch rather than as a follow-up, since the issue that named the problem is closed and these are the last two instances.

Running total on this branch is now eight: the six guards that merged without firing, plus these two.

nstfkc:
## Update at `f307597` — #185 merged, and the correction now contradicts the file it describes

Branch is healthy: `tsc --noEmit` clean, **1376 passed**, `oxlint orm --deny-warnings` exit 0.

#185's docs fix landed. The stale copy in the template did not, so the two now disagree:

```
docs/orm.md:34
  Add the gemi generator to `schema.prisma`. Conventionally after the `client` block,
  though the order does not matter …

templates/saas-starter/prisma/schema.prisma:10
  // Declared after `client` on purpose: the emitted bases type-import
  // `@prisma/client`, and Prisma runs generators in declaration order.
```

Before the merge both said the same wrong thing. Now one says the order does not matter and the other says it was chosen on purpose for a reason that was measured to be false — and the second is the file a reader has open while doing what the first tells them to do. That is a narrow way in which correcting one copy has left the repository worse than it was: a single wrong answer became two contradictory ones.

`packages/gemi/bin/orm-generator.ts:23` is the third account — *"the client generator has to have run"* — against the new docs' *"only has to resolve when you typecheck"*.

Nothing catches any of it. #185's guard asserts the generator's **imports**, which is the right property and stays true; the prose is not covered, and #180's verbatim check only spans `orm.md` ↔ `llms-full.txt`.

### Ledger on this branch

Nine items, each verified on this head:

| | |
| --- | --- |
| six guards that merged without firing | `mode` and `_sum` out of `LITERAL_KEYS`, the dialect-seam lookup, the re-export import edge, the `_count.select` null guard, per-model modules in `generated/` |
| two stale counts | `orm/compile/unique.ts:13` "thirteen", `bin/orm/emit.ts:439` "twelve" |
| the ordering comment | `templates/saas-starter/prisma/schema.prisma:10`, now contradicting `docs/orm.md` |

Seven of the nine are one line each. The two that are not — `mode` and `_count: { select: null }` — are one corpus line and one test case, written up with verified fixes on #123 and #139.

They share a shape worth naming now that there are nine of them: every one is a place where the *claim* is right and the thing that would notice it going wrong is absent or scoped past. That is the failure mode this whole series has been documenting, and the branch is about to merge carrying nine instances of it.

nstfkc:
## The three open PRs compose cleanly — checked, because this branch has been broken twice by a combination

#191, #193 and #195 each state they are independent of the others. Verified rather than taken, since #159 and #166 were both defects that existed only after a merge and neither PR could see.

```
merge-tree against feat/orm     all three merge with no conflict
all three applied together      tsc --noEmit clean
                                vitest run orm database   1410 passed
                                template app/models        616 passed | 32 skipped
                                oxlint orm --deny-warnings exit 0
```

So they can land in any order.

**One thing worth recording about how I nearly got this wrong.** My first composition copied each PR's changed files wholesale onto `feat/orm`, and produced two failures — #189's benchmark-figure guard reporting `+40% on Postgres` again. That looked like a real interaction: #191 branched before #189 merged, so its copy of `orm-rows-and-entities.md` still holds the old sentence.

It is not one. The two PRs edit different hunks of that file — line 52 and line 107 — so a real merge keeps both, and `git merge-tree` confirms the merged file carries #189's correction. Overwriting whole files is not a merge, and the difference is exactly the case where two changes to one file are compatible.

Fifth time this session my own measurement method has been the thing that was wrong, and the same shape as the other four: a boundary or a mechanism chosen for convenience, not checked against what it excludes. Worth stating alongside the result, since a "these conflict" report would have cost someone a rebase they did not need.

### Ledger unchanged

Nine items still open on the branch, listed with reproductions in my earlier comments. Two of them — `orm/compile/unique.ts:12` and `plans/orm/README.md:186` — are addressed by neither #195 nor anything else in flight; both are written up on #195 with a verified pattern that catches them and nothing else.

nstfkc:
## Update at `8d8acd9` — three benchmark guards merged, all three keyed on `×`

Branch is healthy: `tsc --noEmit` clean, **1403 passed**.

#193 landed, so all three prose guards from #187, #189 and #193 are now on the branch. All three still match only `µs`, `×`, `%`:

```
benchmarks-report.test.ts:96    /\d+(\.\d+)?\s*(µs|×|%)|\bwas measured\b/
benchmarks-report.test.ts:139   /\b0\.\d{2}×/g
docs-benchmark-figures.test.ts:56  /…\s*(%|µs|×)\s+on\s+(SQLite|Postgres)/gi
```

One line added to `plans/orm/benchmarks.md`, stating two Postgres figures the run never produced:

```
- Postgres point reads come out at 0.82x, and the depth-2 include at 3.4 ms.

  1403 passed
```

Nothing fires. Written with `×` and `µs` the same sentence trips two of the three.

Adding the equivalent to `docs/orm-rows-and-entities.md` does produce one failure — but it is #179's *verbatim-embedding* check noticing the file no longer matches `llms-full.txt`, not any guard reading the claim. A re-sync would silence it, and `plans/` has no embedding guard at all, so the report itself is uncovered in the direction that matters.

I have raised this on each of the four as it went past. Recording it here because it is now a property of the branch rather than of a PR: three guards, one character class, and the character that is hard to type is the one they require. The fix is `(%|µs|us|ms|×|x)` in all three — I verified on #189 that widening the units alone introduces no false positive on the corrected pages, and that the word-order half is the part needing care.

### Ledger

Eleven items now, each verified on this head:

| | |
| --- | --- |
| six guards that merged without firing | `mode` / `_sum` out of `LITERAL_KEYS`, the dialect-seam lookup, the re-export import edge, the `_count.select` null guard, per-model generated modules |
| two stale counts | `orm/compile/unique.ts:12`, `plans/orm/README.md:186` — both written up with a verified pattern on #195 |
| the ordering comment | `templates/saas-starter/prisma/schema.prisma:10`, contradicting `docs/orm.md` since #185 merged |
| the benchmark unit set | three guards, above |
| `bin/orm-generator.ts:23` | "has to have run" against the corrected docs' "only has to resolve when you typecheck" |

The three open PRs — #195, #197, #199 — compose cleanly with each other and with this head; I checked that last tick and again after #193 landed.

nstfkc:
## Correction and an update at `0bda47e` — both stale counts are still there, and my last report said one was fixed

Branch is healthy: `tsc --noEmit` clean, **1419 passed**.

**First, a correction to myself.** Ten minutes ago I reported `plans/orm/README.md:186` as fixed by #195's merge. It is not. I checked with a single-line grep for `"twelve public operations"`, and the sentence wraps:

```
186| `Model.$exec(op, args)` is the only place that touches the database. The twelve
187| public operations are one-line delegations to it.
```

The phrase never appears on one line, so the grep found nothing and I read that as absence. Same class of error as everything else in this thread — a pattern narrower than the prose — and the sixth time this session my own method has been the thing that was wrong. Worth saying, because "fixed" is the report that stops anyone looking.

**Both are still on the branch:**

```
plans/orm/README.md:186          "The twelve public operations …"     invariant 1, "Every iteration must preserve these"
orm/compile/unique.ts:13         "… not one of the thirteen."
```

`unique.ts:13` has now survived six passes: #183's enumerated fix, #195's pattern sweep, and four flags. Not because anyone disagreed — because every pattern used has been keyed on `<count> operations`, and neither of these is that. One puts an adjective in the middle; the other puts the number after the noun, in the singular.

### The fix, applied to this head and verified

The merged guard is one line:

```js
const COUNT = new RegExp(`\\b(${NUMBER_WORDS.join("|")})\\s+operations\\b`, "gi");
```

Widening it for the two phrasings, on `0bda47e`:

```js
const BIG = NUMBER_WORDS.slice(11).join("|");
const COUNT = new RegExp(
  `\\b(${NUMBER_WORDS.join("|")})\\s+(?:\\w+\\s+)?operations\\b`
    + `|\\bone of the\\s+(${BIG})\\b`,
  "gi",
);
// …and `word: (match[1] ?? match[2]).toLowerCase()`
```
```
× packages/gemi/orm/compile/unique.ts: "one of the thirteen"
× plans/orm/README.md: "twelve public operations"
```

Both named, no false positives — I checked across `packages/gemi`, `docs/` and `plans/orm/README.md`. Restricting the second alternative to eleven-and-up is what keeps *"not one of the operations"* — the drift-proof wording #185 introduced — out of it, and allowing only **one** intervening word is what stops *"one of the operations"* matching as a count.

Ledger otherwise unchanged at eleven; this resolves the description of two of them rather than adding any.

nstfkc:
## Three of the merged guards are case-sensitive, and #205 was a case-sensitivity bug

Having raised "the pattern is narrower than the text it guards" on five PRs now, I swept every guard on `0c01841` that reads a text artifact rather than code behaviour, and tested the ones where case could plausibly vary. Two more evade, both on the branch today:

**`plan-record.test.ts:46`** — `/a future \`([A-Za-z_][\w.]*)\`/g`

```
plans/orm/README.md += "A future `ActiveRecordModel` will make this simpler."
  → 2 passed
```

A sentence-initial capital is not an edge case; it is how the sentence gets written if it starts a paragraph. The guard from #197 exists to catch exactly this claim.

**`benchmarks-report.test.ts:65`** — `/^\|\s*(sqlite|postgres)\b/`

```
plans/orm/benchmarks.md += "| Postgres | 1. point read by pk | 1 | 15.0 | 28.2 | 1.88x |"
  → 5 passed
```

A table row asserting a dialect the run never measured. The report's own prose writes "Postgres" capitalised throughout, so a hand-written row plausibly would too.

With #207's `CREATE TABLE` pattern — which I raised there — that is **three** guards in this series that compare case-sensitively against text a human writes. And #205, one commit below #207, is a defect that existed *because* `-t` compares case-sensitively and eight tests were named `Postgres`.

Adding `i` to both closes them, verified in all three directions:

```
patterns unchanged, artifacts unchanged     7 passed
+ `i`, artifacts unchanged                  7 passed    (no false positive)
+ `i`, "A future `ActiveRecordModel`"       1 failed
+ `i`, "| Postgres | …"                     1 failed
```

Worth doing as one change rather than three, because the property is shared: **a guard that reads prose should compare the way prose varies.** The exception is the one #205 established — `-t` genuinely is case-sensitive, so the check on suite names must be too, and #205's comment says so explicitly. That is the only one of the four where case-sensitivity is the correct answer, and it is the one where it was chosen deliberately.

Ledger goes to thirteen; these two are one character each.

nstfkc:
## At `1fb8305` — three planted falsehoods, three guards written for them, suite green

Branch is healthy: `tsc --noEmit` clean, **52 files / 1436 passed**.

#207 merged with its `CREATE TABLE` pattern unchanged, so all three case-sensitive guards are now on the branch together. I planted one artifact for each — every one exactly the class its guard exists to catch:

```
templates/.../shadow.test.ts   create table "Ledger" ("id", "secret")
                               a fixture shadowing a real model with a different shape

plans/orm/README.md            A future `ActiveRecordModel` will make this simpler.
                               shipped API described as forthcoming

plans/orm/benchmarks.md        | Postgres | 1. point read by pk | 1 | 15.0 | 28.2 | 1.88x |
                               a table row for a dialect with no data
```

```
1436 passed
```

Nothing fires. Each differs from a caught form only by case.

**One character in each closes all three**, verified in both directions on this head:

```
patterns + `i`, artifacts unchanged     1436 passed        (no false positive)
patterns + `i`, the three planted       4 failed
  × shadow.test.ts: Ledger
  × the shadowing set is the known one
  × no table row names a dialect the data does not have
  × A future `ActiveRecordModel` is not something the ORM already ships
```

```
fixture-shape.test.ts       …\)\s*`/gs   → /gis
plan-record.test.ts         …`/g          → /gi
benchmarks-report.test.ts   …postgres)\b/ → /i
```

The one to leave alone is `postgres-suite-selection.test.ts`: `-t` genuinely is case-sensitive, that is the defect #205 fixed, and its comment says so. That is the only one of the four where the current behaviour is the correct answer — which is worth noting, because "make the guards case-insensitive" applied uniformly would undo the fix that started this.

### Ledger

Thirteen items, all verified on this head. The three above are one character each; the six guards that merged without firing and the two stale counts are written up with reproductions and verified fixes on earlier comments here.

Worth a note on the shape of the last twenty PRs: nine of them added a guard over prose or configuration, and five of those shipped with a pattern narrower than the text it reads — three by case, two by phrasing. Every one was written against the instance that was wrong, and every one passes on the next spelling of the same mistake. That is a more useful thing to know than any of the thirteen individually.

nstfkc:
## The branch carries a silent wrong-results bug, and the issue that held it is now closed

Recording it here because #224 was closed as completed and `plan.ts` still has no sentinel handling, so nothing tracks this any more.

`canonicalShape` renders both Prisma sentinels — and any other object with no enumerable properties — as `{}`. That was harmless while all three compiled to the same SQL and differed only in the bound parameter, which is exactly what `VALUE_KEYS` is for. #225 changed it: `DbNull` now compiles to `is null` with **no parameter**, while `JsonNull` and an ordinary `{}` stay `= ?`. One cache key, three different statements.

Reproduced on `95e6212`, the current head of this branch, against a table holding one row of each kind:

```
one warm cache
  equals DbNull               -> sqlnull
  equals JsonNull             -> sqlnull      wrong
  equals {}                   -> sqlnull      wrong

refusal bypass
  settings: {} (empty filter) -> sqlnull,jsonnull,obj
  bare DbNull                 -> sqlnull,jsonnull,obj    the InvalidArgumentError never fires
```

Order decides the direction. Compile `JsonNull` first and `equals: DbNull` returns nothing at all — #224's exact title, restored by the cache rather than by the compiler.

Nothing in the suite can see it: `expectSame` opens with `clearPlanCache()`, so no two comparisons ever share an entry, and the six filter cases #225 added each run against a cold cache.

The shape of the fix has not changed — make the sentinel part of the key, since `jsonNullKind` already identifies it without importing the Prisma client, and cover it with two queries under one warm cache. `plan.discrimination.test.ts` is where the sibling properties live, and `planKey`'s own comment about the relation strategy already describes this failure class: *"different strategies emit different SQL for the same arguments … silent wrong results from a caching bug."*

nstfkc:
## The three open docs PRs compose, in any order

No new PR this iteration, so I checked the thing none of #231, #233 or #235 can see from inside itself: they all edit `docs/orm.md` **and** `docs/llms-full.txt`, and each re-synced the second against its own copy of the first. Three independently-synced copies of a large generated file merging textually is the case where a clean `git merge` and a passing guard are different questions.

They are fine. Four orders tried, all clean, and in each the verbatim property survives:

```
231 → 233 → 235   clean, orm.md still embedded verbatim
235 → 233 → 231   clean, verbatim
233 → 235 → 231   clean, verbatim
235 → 231 → 233   clean, verbatim
```

On the composed tree, `packages/gemi/orm` is green — **55 files, 1471 passed** — which covers #231's sentinel assertions, #235's gap guard, and #181's verbatim check.

Each PR's own extraction still finds what it is looking for on the merged page, which is the part a textual merge could have broken quietly:

```
#233  differential corpus — **48** of them today     matches
#231  Prisma.DbNull / Prisma.JsonNull named          both present
#231  no bare-null offer in the Json section         holds
#235  no `Known gap (#N)` naming a landed issue      holds
```

One gap in that: #233's assertion lives in `templates/saas-starter/app/models/differential.test.ts`, which I cannot run — my Prisma client predates #219's `Organization.plan`, so that file's `beforeAll` fails here. I checked its regex against the composed page directly instead, which confirms the sentence survived the merge but not that `RELATION_CASES.length` is still 48. That half is unverified by me.

Nothing to act on. Recording it so the merge order does not have to be argued about, and so the one unchecked assertion is named rather than assumed.

PR #46, opened 2026-07-27
ORM iteration 2 — scooter: the full read surface

nstfkc:
Implements [`plans/orm/02-scooter-query-surface.md`](https://github.com/nstfkc/gemi/blob/feat/orm/plans/orm/02-scooter-query-surface.md).

Stacked on iteration 1 (#44), which is still open — this targets
`feat/orm-01-skateboard`, not `feat/orm`.

Every read operation that does not involve a relation, on SQLite and Postgres,
with a differential harness that runs each query through both Prisma and gemi
and deep-equals the results.

## Semantics came from running Prisma, not from reading its docs

The plan's open questions were settled empirically: Prisma was pointed at a
throwaway copy of the template database with query logging on, and the SQL and
bound parameters it emits were read directly. That turned up several behaviours
worth naming, all now matched and covered:

| Behaviour | What Prisma actually does |
| --- | --- |
| `contains` escaping | **Does not escape `%` or `_`** — `contains: "50%_x"` binds `"%50%_x%"` |
| pagination without `orderBy` | injects `order by <primary key> asc` |
| negative `take` | flips every ordering term, binds the magnitude |
| `{ AND: [] }` | vacuously true |
| `{ OR: [] }`, `{ in: [] }` | match nothing |
| `mode: "insensitive"` | rejected outright on SQLite |

**The escaping one contradicts the plan**, which said "escaping `%` and `_`
inside user-supplied values is a real correctness issue and Prisma handles it".
It does not. A wildcard in user input really is a wildcard. Since the contract
this ORM signs is differential equality with Prisma, escaping here would break
every such query — so gemi matches Prisma and says so at the call site. Worth
knowing about; worth not fixing silently.

**`mode` is the other one to be deliberate about.** SQLite's `LIKE` is already
case-insensitive for ASCII and Postgres's is not, so the same `contains` returns
different rows on the two dialects. That is Prisma's behaviour too. gemi matches
Prisma per dialect rather than inventing a uniformity Prisma does not have, and
both dialect files say why.

## Two bugs caught by tests rather than by review

- **`count({ take })` reported the unpaginated total.** `limit` was appended to
  the aggregate, capping the count's own single result row instead of the rows
  being counted. The differential harness caught this on its very first run.
  Now wrapped in a subquery, the way Prisma does it.
- **The plan cache collided `asc` with `desc`.** `canonicalShape` reduced values
  to their *types*, which is right for a `where` value and wrong for anything
  structural — `orderBy` directions, `select` booleans, `mode`. Two different
  queries shared one entry, so the second caller got the first one's SQL. That
  is the silently-wrong-query failure the plan calls the worst class of bug in
  this codebase. Keys whose values are structural are now recorded verbatim, and
  `take`'s sign is keyed separately from its magnitude (the sign flips the
  ordering; the magnitude is just a parameter).

## The dialect seam held

`inList`, `like` and `paginate` were added to `SqlDialect` rather than letting
Postgres branch inline. Each is a place where the two dialects disagree about
*structure*, not spelling:

- **`inList`** — SQLite expands one placeholder per element, so `in: [1,2]` and
  `in: [1,2,3]` are different SQL and different plans. Postgres binds the whole
  array to one parameter with `= any($1)`, so every length shares a single plan
  and a single prepared statement. There is a test asserting exactly that
  asymmetry.
- **`like`** — Postgres has `ilike`; SQLite has no equivalent.
- **`paginate`** — SQLite cannot parse `offset` without a preceding `limit`, so
  a bare `skip` gets a limit bound to `-1` rather than an inlined literal.

`$1` numbering needed nothing but the interface: `bind` already returned an
ordered array, and the parameter sentinel in `Fragment` already deferred
placeholder rendering to assembly time.

## Acceptance criteria

| # | Criterion | Where |
| --- | --- | --- |
| 1 | Six operations on both dialects | `orm/compile/read.ts`, `read.test.ts` |
| 2 | Differential harness green on both | `app/models/differential.test.ts` — 72 cases |
| 3 | Exact SQL + parameters per dialect | `orm/compile/read.test.ts` — 62 cases |
| 4 | `findUnique` names the required fields | `read.test.ts`, `User.test.ts` |
| 5 | Plan-cache discrimination incl. `in`-length | `orm/plan.discrimination.test.ts` |
| 6 | `select` on a relation throws | `orm/compile/select.ts`, `read.test.ts` |
| 7 | No value inlined anywhere | `plan.discrimination.test.ts` — strips quoted identifiers, asserts no digit survives, both dialects |
| 8 | `bun run lint` / `bun run test` | see below |

**Postgres is env-gated and the skip is loud.** With no `TEST_POSTGRES_URL` the
suite prints a warning naming the dialect that did not run, rather than a silent
`skipped`. The Postgres dialect is fully covered by compiler unit tests
regardless; what the env var adds is the differential comparison. I have not run
it against a live Postgres — that needs a scratch database I do not have here,
so treat Postgres as compiler-verified but not yet differentially verified.

**On criterion 8.** Lint is clean: 0 errors, 0 warnings from new code. 193 tests
pass in `packages/gemi` (was 123) and 91 in the template (was 14). Full
`packages/gemi` suite still fails the same 8 files / 7 tests as the base branch —
unchanged, unrelated, pre-existing. Type checks clean for new code in both
packages.

## One thing to know about the harness

It builds its SQLite fixture by replaying the committed migrations into a fresh
temp file, rather than calling `prisma db push --force-reset`. The Prisma CLI
refuses to run that unattended, and it is right to: it is destructive by design.
Replaying migrations into a file that did not exist a moment ago cannot harm
anything, needs no consent prompt, and is faster.

The Postgres path assumes `TEST_POSTGRES_URL` points at a scratch database with
the schema already applied, and clears only the `User` table it seeds.

## Out of scope, per the plan

`include` and relation `select` (iteration 3), writes (4), transactions (5),
policies (6). `cursor` and `distinct` throw `UnsupportedQueryError`; `aggregate`
and `groupBy` are not emitted onto the generated bases at all.


packages/gemi/orm/compile/read.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  if (paginating && terms.length === 0 && !single) {
+    terms = schema.primaryKey.map((name) => ({
+      column: schema.fields[name]?.column ?? name,
+      direction: "asc" as const,
+    }));
+  }
+
+  const negative = typeof take === "number" && take < 0;
+  if (negative) terms = reverse(terms);
```
nstfkc, line 201: **Blocking: negative `take` returns rows in the wrong order.**

This flips the ordering terms, but `shape` (below, L133-143) hands the rows back as-is. Prisma flips the SQL order *and then reverses the result set*, so it returns rows in the order the caller's `orderBy` asked for. Verified against Prisma directly:

```
prisma take:-3 orderBy id asc => [ "c", "d", "e" ]
gemi   take:-3 orderBy id asc => [ "e", "d", "c" ]
```

Adding the case to this PR's own harness fails both with and without an explicit `orderBy`:

```
User.findMany({"take":-3,"orderBy":{"id":"asc"},"select":{"name":true}})
- Expected  n3, n4, n5
+ Received  n5, n4, n3
```

`take: -1` is order-insensitive, which is why the compiler test at `read.test.ts:224` passes.

Fix: have `pagination` report that it reversed, and reverse the shaped array in `shape` before returning. Then add both cases to `CASES`.

packages/gemi/orm/compile/read.ts
```diff
@@ -0,0 +1,301 @@
...
+export function compileRead(
+  schema: ModelSchema,
+  op: Operation,
+  args: any,
+  dialect: SqlDialect,
+): QueryPlan {
+  assertArgs(schema, op, args);
+
+  const context = { dialect, operation: op };
+  const where = compileWhere(schema, args?.where, context, (a) => a?.where);
```
nstfkc, line 53: **Blocking: `findUnique` with a compound unique key does not work.**

`compileWhere` runs here, *before* `assertUniqueWhere` on L51. `assertUniqueWhere` accepts Prisma's joined form `{ username_provider: { ... } }` (L269-273), but `compileWhere` has no branch for it — the joined name is not in `schema.fields`, so it throws first:

```
THREW: UnknownFieldError - 'isArchived_occurredAt' is not a field on model SocialAccount.
```

The template's `SocialAccount` declares `@@unique([username, provider])`, so this is reachable today.

Either destructure the compound key in `compileWhere` into an AND over its members, or drop the compound branch from `assertUniqueWhere` and reject it with a message that says so.

packages/gemi/orm/compile/read.test.ts
```diff
@@ -0,0 +1,513 @@
...
+
+  test("accepts Prisma's joined compound form", () => {
+    expect(() =>
+      compileRead(
+        composite,
+        "findUnique",
+        { where: { isArchived_occurredAt: { isArchived: true } } },
+        sqlite,
+      ),
+    ).not.toThrow(/needs a unique field/);
```
nstfkc: **This assertion cannot fail.** The call *does* throw — `UnknownFieldError: 'isArchived_occurredAt' is not a field on model SocialAccount` — just with a different message, so the negative-regex match passes and the compound-unique bug went unnoticed.

`expect(...).not.toThrow(/re/)` is worth banning outright: it passes both when nothing throws and when the wrong thing throws. Use bare `.not.toThrow()` (as L300 correctly does) or assert the compiled SQL.

packages/gemi/orm/compile/read.ts
```diff
@@ -0,0 +1,301 @@
...
+    "where",
+    "orderBy",
+    "skip",
+    "take",
+    "select",
+    "include",
+  ]),
+  findUnique: new Set(["where", "select", "include"]),
+  findUniqueOrThrow: new Set(["where", "select", "include"]),
+  count: new Set(["where", "orderBy", "skip", "take", "select"]),
```
nstfkc: `count` permits `select` here and the compiler never reads it — `count({ select: { id: true } })` compiles to a plain `count(*)` and silently returns the unfiltered total instead of Prisma's per-field count object.

`emit.ts:352` already `Omit`s `select` from the generated type, so the runtime is strictly more permissive than the type for no benefit. Drop `"select"` from this set and let `assertArgs` reject it.

packages/gemi/orm/compile/read.ts
```diff
@@ -0,0 +1,301 @@
...
+    let statement: Fragment;
+    if (!paginated) {
+      statement = concat(counted, from, whereClause);
+    } else {
+      const { clause, terms } = pagination(
+        schema,
+        op,
+        args,
+        dialect,
+        parseOrderBy(schema, args?.orderBy, op),
```
nstfkc: `parseOrderBy` is only reached inside the `paginated` branch, so `orderBy` on a `count` is validated only when `take`/`skip` are also present:

```
count({ orderBy: { nonexistent: "asc" } })            => compiles clean
count({ orderBy: { nonexistent: "asc" }, take: 1 })   => UnknownFieldError
```

Same args, different validation. Parse unconditionally and discard the terms when unpaginated.

packages/gemi/orm/plan.ts
```diff
@@ -63,22 +85,34 @@ export function canonicalShape(value: unknown): string {
...
-    // a request supplies becomes a permanent entry in a module-global Map keyed
-    // off user input. The cache needs a bound — an eviction policy, not a
-    // coarser key — before that ships.
-    return `[${value.map(canonicalShape).join(",")}]`;
+    // WARNING: `in` has now landed, so this is live. Every distinct
+    // filter-list length a request supplies becomes a permanent entry in a
+    // module-global Map keyed off user input. On Postgres the array binds to a
+    // single parameter and all lengths share one text, so only SQLite is
+    // exposed — but the cache still needs a bound. An eviction policy, not a
+    // coarser key: collapsing lengths would hand a plan the wrong number of
```
nstfkc: This comment names the problem honestly, but the PR ships without the bound it calls for.

On SQLite every distinct `in`-list length becomes a permanent entry in a module-global `Map`, and list length is frequently request-derived (`ids: string[]` off a query param). That is a slow memory leak reachable from untrusted input — not theoretical now that `in` has landed.

An LRU with a cap is a small change, and landing it here beats deferring it to an iteration with more moving parts. Agreed on the reasoning that the fix is eviction, not a coarser key.

packages/gemi/orm/plan.ts
```diff
@@ -63,22 +85,34 @@ export function canonicalShape(value: unknown): string {
...
 
+function shapeOfMember(key: string, value: unknown, literal: boolean): string {
+  // `take` is a parameter, but its *sign* is not: a negative take means "the
+  // last N", which flips every ordering term and so changes the SQL text. The
+  // magnitude stays out of the key, so `take: 10` and `take: 20` still share a
+  // plan the way every other value does.
+  if (key === "take" && typeof value === "number") {
+    return `number:${Math.sign(value)}`;
+  }
+  return canonicalShape(value, literal || LITERAL_KEYS.has(key));
```
nstfkc, line 143: `literal` propagates down the entire subtree once set, and `LITERAL_KEYS` contains `include`/`select`. Inert today because `include` throws — but iteration 3 turns it on:

```
canonicalShape({ include: { posts: { where: { title: "secret-user-value" } } } })
=> {include:{posts:{where:{title:"secret-user-value"}}}}
```

Every distinct filter value then becomes its own cache entry, and user data ends up in a long-lived global keyed structure — the unbounded-growth problem again, but keyed on values rather than lengths.

`literal` should stop propagating at a `where` boundary. Cheaper to fix now than to discover it alongside relation compilation.

packages/gemi/orm/compile/where.ts
```diff
@@ -73,43 +103,307 @@ export function compileWhere(
...
+ * That is a footgun, but the contract this ORM signs is differential equality
+ * with Prisma, and escaping here would break it on every such query. Matching
+ * Prisma and saying so is the honest option; diverging quietly is not.
+ */
+function likePattern(
+  operator: "contains" | "startsWith" | "endsWith",
+  locate: (args: any) => any,
+): Binder {
+  return (args) => {
+    const value = String(locate(args));
```
nstfkc, line 466: `String(...)` is unconditional, so `{ email: { contains: null } }` compiles to `email like '%null%'` — verified, it binds `["%null%"]`. Prisma raises a validation error. Same shape for `contains` on a non-string field, where the emitted `like` on an Int column is meaningless rather than rejected.

Types catch the static case, but not a value that arrives dynamically. A `typeof value !== "string"` throw costs one line and keeps this in the errors column rather than the wrong-rows column.

templates/saas-starter/app/models/differential.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  ["findFirstOrThrow missing", "findFirstOrThrow", { where: { name: "Nobody" } }],
+  ["findUnique by id", "findUnique", { where: { id: 1 } }],
+  ["findUnique by unique", "findUnique", { where: { email: "ada@example.dev" } }],
+  ["findUnique no match", "findUnique", { where: { email: "nobody@example.dev" } }],
+  ["findUniqueOrThrow", "findUniqueOrThrow", { where: { email: "ada@example.dev" } }],
+  ["findUniqueOrThrow missing", "findUniqueOrThrow", { where: { email: "nope@x.dev" } }],
+
+  // --- count ------------------------------------------------------------
+  ["count", "count", undefined],
+  ["count with where", "count", { where: { deletedAt: null } }],
```
nstfkc, line 199: This case list is the strongest part of the PR — and the gaps are exactly where both confirmed bugs live.

Missing, in rough priority order:

- **negative `take`** (with and without an explicit `orderBy`) — catches the ordering bug
- **compound-unique `findUnique`** — catches the `UnknownFieldError` bug
- `mode: "insensitive"` on both dialects — the PR describes the split behaviour but nothing pins it
- `NOT` and `notIn` over a nullable column — the three-valued-logic cases the `compileNot` comment reasons about
- `count` with `skip` (I checked this one — it does pass)

General rule worth adopting: anything the description claims was "verified against Prisma" gets a differential case, not only a compiler-text case.

nstfkc (commented):
Reviewed the read surface and verified the suspicious paths by running them — including against real Prisma, using this PR's own differential harness.

The architecture is right: the dialect seam is drawn at the three places the dialects genuinely differ structurally, no value is ever inlined, and the plan-cache `asc`/`desc` collision fix is a real catch. 193 tests pass locally as claimed.

Two confirmed bugs, both in the silently-wrong-result class this PR names as the worst kind:

1. **Negative `take` returns rows in the wrong order.** Prisma flips the SQL order *and then reverses the result set*; gemi only does the first half. Fails in your own harness.
2. **`findUnique` with a compound unique key throws `UnknownFieldError`.** `compileWhere` runs before `assertUniqueWhere` and has no branch for Prisma's joined form, so the feature `assertUniqueWhere` advertises does not work. The test covering it cannot fail.

Both live in exactly the gaps in the differential case list — which is worth a rule: any behaviour the description claims was "verified against Prisma" should have a differential case, not just a compiler-text case. Asserting SQL text proves you emitted what you intended, which is precisely what was true for the negative-`take` bug.

Inline comments have the details, plus five smaller items and two noted risks. Nothing here argues against the approach.

packages/gemi/orm/compile/read.ts
nstfkc: Confirmed and fixed in 3fbb28f. Ran it against real Prisma before touching anything:

```
take -3, orderBy id asc  -> c, d, e
take -3, orderBy id desc -> c, b, a
take -2, skip 1, id asc  -> c, d
```

So: flip the SQL order, take abs(n), then flip the page back. `pagination` now returns `reversed` and `shape` reverses before returning.

Five differential cases added — plain, asc, desc, with-skip, and `take: -1`. I then broke the fix deliberately and re-ran to confirm four of them fail without it, because the whole point of this finding is that a test which cannot fail is worse than no test.

packages/gemi/orm/compile/read.ts
nstfkc: Confirmed and fixed in 3fbb28f. `compileWhere` now destructures the compound key into an AND over its members before the unknown-field check, and requires every member — a partial composite key is not unique, so accepting one would be a non-unique lookup wearing a `findUnique` label.

Verified Prisma accepts and resolves the joined form on the template's `SocialAccount`, and the differential harness now covers both the hit and the miss.

packages/gemi/orm/compile/read.test.ts
nstfkc: Fixed in 3fbb28f, and you are right that the form is worth banning outright — `.not.toThrow(/re/)` passes when nothing throws *and* when the wrong thing throws, which is exactly how the compound-unique bug shipped with a green test next to it.

Replaced with an assertion on the compiled SQL and parameters. I checked the rest of the suite for the same pattern; this was the only instance.

packages/gemi/orm/compile/read.ts
nstfkc: Fixed in 3fbb28f — `select` dropped from the `count` set, so `assertArgs` rejects it. You are right that the runtime being more permissive than the generated type is pure downside.

packages/gemi/orm/compile/read.ts
nstfkc: Fixed in 3fbb28f. `parseOrderBy` now runs unconditionally and the terms are discarded when the count does not paginate. Validation that depends on whether a *different* argument happens to be present is the kind of inconsistency nobody can hold in their head.

packages/gemi/orm/plan.ts
nstfkc: Fixed in 3fbb28f rather than deferred — agreed that landing it here beats an iteration with more moving parts.

LRU with a cap of 1000, implemented on `Map` insertion order: a hit re-inserts to move the key to the end, so `keys().next()` is always the coldest. `planCacheStats` now reports `evictions` and `capacity`, so a non-zero eviction count is a visible signal that an app has outgrown the cap rather than something silent.

Two tests: one floods distinct `in`-lengths and asserts the cache stays bounded, one asserts a repeatedly-used shape survives the flood — a cache that evicted its hot entries under exactly the traffic that fills it would be worse than none.

packages/gemi/orm/plan.ts
nstfkc: Fixed in 3fbb28f. `where`, `cursor` and `data` now end the structural subtree, so `literal` stops propagating at a value boundary.

Worth noting how bad this would have been: not just cache growth, but user-supplied filter values sitting in a module-global map for the process lifetime. Much better caught now than during relation compilation.

packages/gemi/orm/compile/where.ts
nstfkc: Fixed in 3fbb28f. Confirmed Prisma raises `Argument `contains` must not be null`.

Went slightly further than the suggestion: non-string *operands* and non-string *columns* are both rejected now, since `like` on an Int column is the same class of meaningless-but-running query.

templates/saas-starter/app/models/differential.test.ts
nstfkc: Agreed, and adopted. Added: negative `take` (four variants), compound-unique `findUnique` (hit and miss), `mode: "insensitive"` on both dialects, `NOT`/`notIn`/`not: { in }` over a nullable column, `NOT` over a null literal, and `count` with skip / take+skip / orderBy.

`mode` could not go in the shared table since the dialects legitimately disagree, so there are now `SQLITE_ONLY` and `POSTGRES_ONLY` lists selected by which suite is running — Prisma rejecting it on SQLite is itself the pinned behaviour.

86 differential cases now, up from 72. Taking the general rule as standing for the later iterations.

nstfkc:
Pushed 3fbb28f. Both blocking findings confirmed against real Prisma before fixing, then all seven smaller items.

**The two bugs**

- **Negative `take` ordering.** Prisma flips the SQL order *and* reverses the result set; gemi did only the first half. `pagination` now reports that it reversed and `shape` flips the page back. I broke the fix deliberately and re-ran to confirm four of the new differential cases fail without it.
- **Compound-unique `findUnique`.** `compileWhere` now destructures Prisma's joined `{ username_provider: { ... } }` form into an AND over its members, and requires every member — a partial composite key is a non-unique lookup wearing a `findUnique` label.

**The rest**

- `.not.toThrow(/regex/)` replaced with an assertion on compiled SQL. Checked the suite; it was the only instance. Agreed the form is worth avoiding outright.
- `count` no longer accepts `select`, and validates `orderBy` whether or not it paginates.
- `contains`/`startsWith`/`endsWith` reject non-string operands *and* non-string columns. Confirmed Prisma raises on null.
- Plan cache now has an LRU cap of 1000, with `evictions` and `capacity` in `planCacheStats`. Two tests: one floods distinct `in`-lengths and asserts boundedness, one asserts a hot shape survives the flood.
- Structural keying stops at `where` / `cursor` / `data`, so iteration 3 will not put user filter values into a process-lifetime global map.

**Your general rule is adopted.** Anything the description claims was verified against Prisma now has a differential case. 86 cases, up from 72 — added negative take (4 variants), compound uniques (hit and miss), `mode: "insensitive"` on both dialects, `NOT`/`notIn`/`not: { in }` over a nullable column, and count with skip / take+skip / orderBy. `mode` needed its own per-dialect lists since the two legitimately disagree.

**Verification.** 207 tests in `packages/gemi` (was 193), 105 in the template (was 91). Lint 0 errors, 0 warnings from new code. Type checks clean in both packages. Full `packages/gemi` suite still fails the same 7 pre-existing tests as the base branch.

Postgres remains compiler-verified but not differentially verified — unchanged from the original description, and the loud skip still says so.

PR #47, opened 2026-07-27
ORM iteration 3 — motorcycle: relations through the batched planner

nstfkc:
Third in the ORM stack, on top of #46. Implements
[plans/orm/03-motorcycle-relations.md](../blob/feat/orm-03-motorcycle/plans/orm/03-motorcycle-relations.md):
`include` and nested `select` / `include` to arbitrary depth, through the
batched planner.

## The planner is a stage, not a branch

`packages/gemi/orm/compile/plan-relations.ts` holds a `RelationStrategy`: given
a parent schema and one relation node, produce the parent field the stitch keys
on plus a `load` that fetches the children and attaches them. `batchedStrategy`
is the only implementation. Iteration 7's `LATERAL` + `json_agg` has to be a
sibling implementation of the same interface — that is the whole point of the
seam (invariant 4).

## One query per node, through the child's own `$exec`

For `include: { accounts: true }` over 100 users: run the root query, collect
the parent keys, run **one** child query with `where userId in (<keys>)`, stitch
by key. A depth-2 tree with two branches is 4 queries over 5 rows and 4 queries
over 1 row — asserted, because an accidental N+1 is invisible in the results and
only shows up in production.

Each child query is `$exec` on the **related model's own class**, resolved by
name from the registry at call time — not a private helper (invariant 1). That
is what makes iteration 6's policies apply to nested reads, and there is a test
that spies on `$exec` to keep it that way. Nested trees need no special handling
for the same reason: a child's subtree is just its own call's arguments.

The `in (<parent keys>)` list has iteration 2's variable-length plan-cache
problem at higher volume, and reuses iteration 2's answer (the LRU bound) rather
than inventing a second mechanism.

## Result shape

The three lines most divergence bugs live on:

- to-one present → the object
- to-one absent → `null`, not `undefined`, not a missing key
- to-many empty → `[]`, not `null`, not absent

The precompiled shaper writes the empty value per row, so the key exists before
anything is loaded. Stitching builds one `Map` from parent key to parent rows
per child query — never a `find()` per row.

A `select` that omits the join key still has to fetch it: those columns are
selected, used, and deleted (`plan.hidden`). With `include`, or a `select` that
names the key, nothing is hidden and nothing is deleted.

**One deliberate divergence:** a child row reached by several parents is
attached by reference, so 100 users in 3 organizations share 3 objects rather
than getting 100 copies. Values are identical and the differential harness sees
nothing; identity differs from Prisma. Cloning is the copying the batched
strategy exists to avoid, and a shallow clone would still alias everything
below it. Noted in the code, worth revisiting in iteration 8.

## The two decisions the doc asked for

**`take` / `skip` on a to-many relation throws `UnsupportedQueryError`.** It
means N *per parent*, which under batching needs
`row_number() over (partition by ...)` in a wrapping subquery — not expressible
through `$exec` on the child's model class, so it would need exactly the private
query path invariant 1 exists to prevent. Applying the limit globally returns
wrong data, which is worse than not supporting it. It falls out of iteration 7's
lateral strategy. The error says so. Tested as a compiler error and end to end.

**Implicit many-to-many is implemented**, not thrown — two hops: read the
`_PostToTag` pairs, then load the rows through `Tag.$exec`. The join table is
the one query the ORM issues outside `$exec`, deliberately: it has no model, no
fields of its own, and nothing an application could scope. The hop that reads
*rows* still goes through the choke point, so a policy on `Tag` will apply to
`include: { tags: true }`. A self-referential implicit m-n throws, because the
generated artifact cannot say which of `A` / `B` is which end.

The fixture is a real database rather than a second Prisma client
(`relations.many-to-many.test.ts`): the DDL is verbatim
`prisma migrate diff --from-empty --to-schema-datamodel` output, and the
generator's `joinTable` emission was confirmed by running `prisma generate` over
that schema against real DMMF rather than assumed. It runs on both dialects.

That is weaker than the doc's first choice — a differential run against a
generated fixture client — and stronger than its stated fallback. The cost I
declined was a second `prisma generate` inside the test path. Say the word and
it can be upgraded; it is one file.

## Guards

- Depth limit of 10, with the *whole* tree walked at root compile — a guard that
  only ever sees one level catches an unbounded tree one query too late, every
  time.
- A relation naming an unregistered model reports the model, the fix
  (`prisma generate`), and what *is* registered — it can only mean a stale
  generated artifact.
- `select` + `include` together throws at every level, not just the root.
- Multi-field relations and self-referential implicit m-n throw rather than
  guessing.

## Two Postgres defects found — one fixed here, one for review

The Postgres differential had never actually run: a generated Prisma client is
provider-bound, so `TEST_POSTGRES_URL` alone fails before any query is issued.
Covering it means flipping `datasource db { provider }`, regenerating, and
running the suite — during which the SQLite suite cannot run. The harness now
says that in an error message instead of surfacing Prisma's own. (The gemi
artifacts are byte-identical under either provider, which is its own
confirmation that generated output is dialect-agnostic.)

**Fixed here, because every batched relation query depends on it:** Bun's driver
rejects a JS array bound to `= any($1)` outright — `insufficient data left in
message` for numbers, `malformed array literal` for strings. So iteration 2's
Postgres `in` never worked at runtime. The values are now serialized to a
Postgres array literal: still one bound parameter, still nothing in the SQL text,
so neither the injection story nor the plan cache changes. Verified against
Postgres 16 for `int`, `text` (quotes, commas, braces, backslashes, newlines),
`timestamp`, `boolean`, `bigint` past 2^53, `bytea` and `double precision`, with
unit tests pinning the escaping.

**Not fixed, and I do not think it belongs in this PR:** Bun decodes a zoneless
`timestamp` — which is what Prisma maps `DateTime` to — as **local** time when a
statement binds no parameters, and as UTC when it binds one. Same row, same
column, two instants, on any machine that is not already UTC:

```
select "createdAt" from "User" limit 1           -> 10:26:40Z
select "createdAt" from "User" where "id" = $1   -> 12:26:40Z
select "createdAt"::text from "User" limit 1     -> 12:26:40     (what is stored)
```

A dialect `decode` cannot correct it: the value does not say which protocol
produced it, and nothing below the plan does. It is iteration-2 surface, it
affects every `DateTime` rather than anything about relations, and the real fix
is either upstream in Bun or a change to the SQL emitted for timestamp columns —
which deserves its own review. Documented in `orm/dialect/postgres.ts` with the
reproduction. Interim mitigation: run with `TZ=UTC`, where both paths agree.

## Verification

- **SQLite:** 145 template tests, 252 package tests, green.
- **Postgres** (`TZ=UTC`, provider flipped, scratch database): all 123
  differential cases — including ~35 new relation cases — plus the m-n fixture,
  green.
- The differential matrix now covers: to-one, to-many, empty to-many, null
  to-one, `select` inside `include`, `include` inside `select`, a relation
  selected with no scalars at all, relation-level `where` / `orderBy`, depth 3,
  a cyclic-but-finite tree, relations on `findFirst` / `findUnique*`, and both
  far sides of every relation.
- `bun run lint` 0 errors; `tsc` clean in both packages.
- The 7 pre-existing router / i18n failures in `packages/gemi` are unchanged.

## Out of scope, as scheduled

Lateral / `json_agg` (iteration 7), `_count` on relations, relation filters in
`where` (`some` / `every` / `none` — they belong in the where compiler as
`EXISTS` subqueries and are worth scheduling explicitly), all writes,
transactions, policies.

### DIFF
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -1,5 +1,3 @@
-import { DatabaseManager } from "../../database/DatabaseManager";
-import { app } from "../../foundation/app";
 import type { SqlDialect } from "../dialect";
 import {
   MalformedRelationError,
@@ -12,6 +10,7 @@ import {
 import * as registry from "../registry";
 import type { FieldSchema, ModelSchema, RelationSchema } from "../schema";
 import { concat, render, sql } from "./fragment";
+import { resolveSelection } from "./select";
 
 /**
  * The relation planner: an include tree plus a root model in, a *strategy* out.
@@ -28,8 +27,16 @@ import { concat, render, sql } from "./fragment";
  * SQL that takes, and how many statements, is entirely the strategy's business.
  */
 
-/** The keys a relation node accepts. `take` / `skip` are refused; see below. */
+/**
+ * The keys a relation node accepts, which differ by kind. `take` / `skip` are
+ * refused on both; see `assertNodeArgs`.
+ *
+ * Prisma's to-one include argument is `{ where?, select?, include? }` — it
+ * takes a `where` (verified: it filters, and the relation comes back `null`
+ * when nothing matches) but not an `orderBy`, since there is at most one row.
+ */
 const RELATION_ARGS = new Set(["where", "orderBy", "select", "include"]);
+const TO_ONE_ARGS = new Set(["where", "select", "include"]);
 
 /**
  * How deep an include tree may nest. Not a modelling limit — a legal tree can
@@ -41,13 +48,38 @@ export const MAX_RELATION_DEPTH = 10;
 type Row = Record<string, unknown>;
 
 /**
- * The registry hands back model *classes*; the planner only ever needs these
- * two statics off them. Typed structurally rather than by importing `Model`,
- * which would make `compile/` depend on the class it is compiled for.
+ * The registry hands back model *classes*; planning only ever needs the schema
+ * off them. Typed structurally rather than by importing `Model`, which would
+ * make `compile/` depend on the class it is compiled for.
  */
 interface RelatedModel {
   $schema?: ModelSchema;
-  $exec(op: string, args: unknown): Promise<unknown>;
+}
+
+/**
+ * Everything a loaded relation needs from the outside world, handed in by the
+ * caller rather than reached for.
+ *
+ * This is what keeps `compile/` a pure schema-in / SQL-out directory: without
+ * it, planning a relation would put `DatabaseManager` — and so `import { SQL }
+ * from "bun"` — on the load path of every compiler unit test, which is a real
+ * cost and not a theoretical one. It also means the m-n join hop runs on
+ * whatever connection `$exec` resolved, so iteration 5's ambient transaction
+ * will cover both hops of a many-to-many read rather than only the second.
+ */
+export interface RelationExecutor {
+  /**
+   * A read on another model, through *that model's own* `$exec` (invariant 1).
+   * Named rather than passed as a class so resolution stays at call time
+   * (invariant 6).
+   */
+  exec(model: string, op: string, args: unknown): Promise<unknown>;
+  /**
+   * A statement with no model behind it. Exactly one query in the ORM is like
+   * this: Prisma's implicit many-to-many join table, which has no model, no
+   * fields of its own and nothing an application could scope.
+   */
+  query(text: string, values: unknown[]): Promise<unknown>;
 }
 
 /** One relation as it appears in an `include` / `select` argument tree. */
@@ -88,7 +120,7 @@ export interface RelationPlan {
   /** Which strategy produced this plan. Reported by tests and, later, metrics. */
   strategy: string;
   /** Fetches the children for these parents and attaches them in place. */
-  load(parents: Row[], args: any): Promise<void>;
+  load(parents: Row[], args: any, executor: RelationExecutor): Promise<void>;
 }
 
 export interface RelationStrategy {
@@ -160,6 +192,7 @@ export async function attachRelations(
   hidden: readonly string[] | undefined,
   result: unknown,
   args: any,
+  executor: RelationExecutor,
 ): Promise<void> {
   const parents: Row[] = Array.isArray(result)
     ? (result as Row[])
@@ -175,7 +208,7 @@ export async function attachRelations(
   // concurrent statements are not safe. Parallelism belongs with iteration 7,
   // which owns performance and can measure it.
   for (const relation of relations) {
-    await relation.load(parents, args);
+    await relation.load(parents, args, executor);
   }
 
   if (hidden && hidden.length > 0) {
@@ -310,9 +343,25 @@ function assertNodeArgs(
   const args = value as Record<string, unknown>;
   assertExclusive(schema, args.include, args.select, operation, node.as);
 
+  const many = node.relation.kind === "many";
+
   for (const key of Object.keys(args).sort()) {
     if (args[key] === undefined) continue;
-    if (RELATION_ARGS.has(key)) continue;
+    if (many ? RELATION_ARGS.has(key) : TO_ONE_ARGS.has(key)) continue;
+
+    // A to-one relation is at most one row, so ordering it is meaningless and
+    // Prisma rejects it outright — `Unknown argument 'orderBy'`, listing
+    // `where?` as one of the options it *does* take. Accepting it here would be
+    // a silent superset of Prisma on an argument that cannot mean anything.
+    if (key === "orderBy" && !many) {
+      throw new UnsupportedQueryError(
+        `${node.as}.orderBy`,
+        schema.name,
+        operation,
+        `Prisma does not accept 'orderBy' on a to-one relation — there is at ` +
+          `most one row to order.`,
+      );
+    }
 
     // The decision recorded in the PR: `take` / `skip` on a to-many relation
     // throw rather than being implemented under the batched strategy.
@@ -343,7 +392,16 @@ function assertNodeArgs(
   }
 }
 
-/** Walks the rest of the tree for the structural rules and the depth guard. */
+/**
+ * Walks the rest of the tree for the structural rules and the depth guard.
+ *
+ * Field names go through the same walk, not just structure: without it a typo
+ * in a `select` three levels down is an `UnknownFieldError` raised by that
+ * model's own compile — which is to say, after two queries have already run.
+ * The whole reason this walk exists is that a guard seeing one level at a time
+ * catches things a query too late, and that applies to a misspelled column as
+ * much as to an unbounded tree.
+ */
 function validateSubtree(
   schema: ModelSchema,
   node: unknown,
@@ -351,6 +409,12 @@ function validateSubtree(
   depth: number,
 ): void {
   if (typeof node !== "object" || node === null || Array.isArray(node)) return;
+
+  // Validates the scalar keys and rejects unknown ones; the relation keys in a
+  // `select` are this function's own business and it skips them.
+  if ((node as Record<string, unknown>).select !== undefined) {
+    resolveSelection(schema, node, operation);
+  }
 
   const nodes = relationNodes(schema, node, operation);
   if (nodes.length === 0) return;
@@ -590,6 +654,8 @@ export const batchedStrategy: RelationStrategy = {
     // compiled rather than after the root query has already run.
     const parentKeyField = field(request.parent, request.relation, link.parentField);
     const childKeyField = field(request.child, request.relation, link.childField);
+    assertKeyable(request.parent, request.relation, parentKeyField);
+    assertKeyable(request.child, request.relation, childKeyField);
 
     const many = request.relation.kind === "many";
 
@@ -598,22 +664,18 @@ export const batchedStrategy: RelationStrategy = {
       kind: request.relation.kind,
       parentField: link.parentField,
       strategy: BATCHED,
-      load(parents, args) {
+      load(parents, args, executor) {
+        const context = {
+          parentKeyField,
+          childKeyField,
+          many,
+          parents,
+          args,
+          executor,
+        };
         return link.join
-          ? loadThroughJoinTable(request, link, {
-              parentKeyField,
-              childKeyField,
-              many,
-              parents,
-              args,
-            })
-          : loadDirect(request, link, {
-              parentKeyField,
-              childKeyField,
-              many,
-              parents,
-              args,
-            });
+          ? loadThroughJoinTable(request, link, context)
+          : loadDirect(request, link, context);
       },
     };
   },
@@ -625,6 +687,7 @@ interface LoadContext {
   many: boolean;
   parents: Row[];
   args: any;
+  executor: RelationExecutor;
 }
 
 async function loadDirect(
@@ -640,7 +703,8 @@ async function loadDirect(
   const node = request.locate(context.args);
   const query = childQuery(node, link.childField, keys);
 
-  const rows = (await childModel(request).$exec(
+  const rows = (await context.executor.exec(
+    request.relation.model,
     "findMany",
     query.args,
   )) as Row[];
@@ -666,6 +730,12 @@ async function loadDirect(
  * holds two foreign keys. The hop that reads *rows* still goes through `$exec`
  * on the child's own model class, so a policy on `Tag` applies to
  * `include: { tags: true }` exactly as it would to `Tag.findMany`.
+ *
+ * It runs on `executor.query` rather than resolving a connection of its own,
+ * which is what keeps the two hops on the same one. Iteration 5 should keep it
+ * that way: if the ambient transaction is resolved inside `$exec` and this hop
+ * reached for `DatabaseManager` directly, a many-to-many read would straddle
+ * the transaction boundary — second hop enrolled, first not.
  */
 async function loadThroughJoinTable(
   request: RelationRequest,
@@ -682,6 +752,7 @@ async function loadThroughJoinTable(
     dialect,
     join,
     keys.map((key) => dialect.encode(key, context.parentKeyField)),
+    context.executor,
   )) as Row[];
 
   // Parents keyed by the *child* id they link to, so the second hop can stitch
@@ -716,7 +787,8 @@ async function loadThroughJoinTable(
   const node = request.locate(context.args);
   const query = childQuery(node, link.childField, childKeys);
 
-  const rows = (await childModel(request).$exec(
+  const rows = (await context.executor.exec(
+    request.relation.model,
     "findMany",
     query.args,
   )) as Row[];
@@ -747,6 +819,7 @@ async function readPairs(
   dialect: SqlDialect,
   join: NonNullable<Link["join"]>,
   values: unknown[],
+  executor: RelationExecutor,
 ): Promise<unknown> {
   const statement = concat(
     sql(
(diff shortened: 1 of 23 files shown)


nstfkc (commented):
## Review — relations through the batched planner

The shape is right. `RelationStrategy` is a genuinely narrow seam, child queries going through `Model.$exec` on the related class is what makes nested reads inherit the choke point, the stitch is `Map`-based rather than `find()`-per-row, and writing the empty value in the shaper gets the three divergence-prone result-shape lines right by construction rather than by care. The query-count assertions and the `$exec` spy cover the parts a result comparison cannot see.

Verified locally on `3d0643a`:

- `bun --bun vitest` → 252/252 in `packages/gemi`, 145/145 in the template.
- `oxlint` → 0 errors.
- The Postgres array-literal escaping tests hold up.

Six comments inline, roughly in priority order. Two of them (the layering one and the plan-key one) I reproduced rather than reasoned about; the rest are read-throughs.

One finding has no line to anchor to, since the file is not in the diff:

**`packages/gemi/package.json:51` declares `"test": "vitest"`** — plain node, no `--bun`, unlike `templates/saas-starter` which uses `bun --bun vitest`. That is why the regression in the first inline comment is invisible from `bun test`. Changing it to `bun --bun vitest` gives 9 files / 252 passed on this branch and makes the declared command agree with the number in the PR description.

Things I would explicitly keep as they are: the shared-child-reference divergence (documented at the point of the code, invisible to the harness, and cloning genuinely defeats the strategy), the `take`/`skip` refusal and its error message, leaving the Bun timestamp defect to its own PR, and the m-n fixture as a real database rather than a second generated client — the DDL provenance and the two query-count assertions buy most of what the differential run would.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
+import { DatabaseManager } from "../../database/DatabaseManager";
+import { app } from "../../foundation/app";
```
nstfkc: **This breaks `bun run test` in `packages/gemi`.** Verified, not theoretical.

These two imports put the runtime on the load path of the pure compiler: `DatabaseManager` does `import { SQL } from "bun"`, `compile/read.ts` imports this file, and `plan.ts` imports `read.ts`. Under the package's declared `"test": "vitest"` (node pool, no `--bun`):

| | base `feat/orm-02-scooter` | this branch |
|---|---|---|
| `bunx vitest run orm/` | 7 files, **207 passed** | 4 files, **89 passed**, 5 suites fail to load |

```
Error: Cannot find package 'bun' imported from database/DatabaseManager.ts
 ❯ database/DatabaseManager.ts:1:1
 ❯ orm/compile/plan-relations.ts:1:1
```

Three of the five (`plan.test.ts`, `plan.discrimination.test.ts`, `read.test.ts`) passed on base. The "252 package tests, green" in the description was measured under `bun test`, which is not the declared command.

Two fixes, and I would take both:

1. **One line:** `"test": "bun --bun vitest"` in `packages/gemi/package.json`, matching the template. I confirmed that gives 9 files / 252 passed here.
2. **Structural:** `attachRelations`, `batchedStrategy.load` and `readPairs` are runtime, not compile-time. The only thing that actually needs a database handle in this file is `readPairs`. Injecting an executor — or hanging that one hop off a `Model` static — keeps `compile/` schema-in / SQL-text-and-binders-out, which is the property the rest of the directory is built on and the one this file's own header argues for ("typed structurally rather than by importing `Model`, which would make `compile/` depend on the class it is compiled for"). The same reasoning applies one layer down.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
...
+
+/**
+ * The child query's arguments: the caller's own relation arguments, plus the
+ * key filter, plus the key itself when a `select` left it out.
+ *
+ * The `in (<parent keys>)` list has the plan-cache problem iteration 2 already
+ * has for `in`, at higher volume: on SQLite each distinct parent count is its
+ * own plan text. It is the same mechanism and the same answer — the plan
+ * cache's LRU bound — rather than a second one.
+ */
```
nstfkc: **The Postgres half of this is not true today, and iteration 3 is what makes it matter.**

`plan.ts:118-123` states it more explicitly — *"On Postgres the array binds to a single parameter and every length shares one text, so only SQLite grows entries this way."* The SQL **text** does share. The cache **key** does not, because `canonicalShape` records arrays element-wise regardless of dialect:

```
postgres:User:findMany:{where:{id:{in:[number,number]}}}
postgres:User:findMany:{where:{id:{in:[number,number,number]}}}
```

Reproduced: `compileRead` returns byte-identical `plan.text` for those two, and `planKey` returns two different keys. So Postgres mints one entry per distinct list length too, each holding SQL identical to its neighbours.

Iteration 2 could live with that. Iteration 3 is less comfortable: every batched relation query's `in` length **is the parent row count**, so a variably-sized parent set now churns the 1000-entry LRU on both dialects rather than one.

`planKey` already receives the dialect. Collapsing array length to a single token when the dialect binds arrays as one parameter is a small change and removes the churn outright. At minimum `plan.ts:118-123` needs correcting — as written it tells the next reader the problem is already handled.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
...
+ * a rewrite of everything that reads relations.
+ *
+ * The strategy interface is deliberately narrow: given a parent schema and one
+ * relation node, produce (a) the parent field the stitch keys on, and (b) a
+ * `load` that fetches the children and attaches them to the parent rows. What
+ * SQL that takes, and how many statements, is entirely the strategy's business.
+ */
+
+/** The keys a relation node accepts. `take` / `skip` are refused; see below. */
+const RELATION_ARGS = new Set(["where", "orderBy", "select", "include"]);
```
nstfkc, line 38: **`where` and `orderBy` should be kind-gated the way `take`/`skip` are.**

This set is flat, but the check at L328-340 is kind-aware and says so: *"Prisma does not accept `'${key}'` on a to-one relation."* The same is true of `where` and `orderBy` — Prisma's to-one include argument is `{ select?, include? }` and nothing else. Confirmed this compiles clean today:

```ts
compileRead(user, "findMany", {
  include: { organization: { where: { name: "x" }, orderBy: { id: "asc" } } },
}, pg)  // → 1 relation plan, no error
```

`where` is the one that matters: gemi silently returns `null` for the relation where Prisma raises a validation error. The generated types catch it at compile time, but anything forwarding an argument tree from a request body reaches it at runtime — which is the threat model `MAX_RELATION_DEPTH` exists for.

Gating on `relation.kind === "many"` puts them in the same branch as `take`/`skip`. The matrix has no case covering it, which is why the differential run stayed green.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
...
+    dialect.inList(
+      dialect.quoteIdent(join.parentColumn),
+      false,
+      values.length,
+      () => values,
+    ),
+  );
+
+  const { text, binders } = render(statement, dialect);
+  const db = app(DatabaseManager);
```
nstfkc: **Worth a note here for iteration 5.**

Reading the join table outside `$exec` is well argued and I would not change it — `_PostToTag` has no model and nothing an application could scope.

But this resolves `app(DatabaseManager).sql` directly. If iteration 5 resolves the ambient transaction inside `$exec` rather than inside `DatabaseManager`, this hop runs outside it, and a many-to-many read straddles the transaction boundary — the second hop enrolled, the first not. Cheaper to leave the constraint written down at the call site now than to rediscover it from a flaky test later.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
...
+function validateSubtree(
+  schema: ModelSchema,
+  node: unknown,
+  operation: string,
+  depth: number,
+): void {
+  if (typeof node !== "object" || node === null || Array.isArray(node)) return;
+
+  const nodes = relationNodes(schema, node, operation);
+  if (nodes.length === 0) return;
```
nstfkc: **Field-name validation does not get the whole-tree treatment that structure does.**

The walk exists because *"a guard that only ever sees one level catches an unbounded tree one query too late, every time"* — which is exactly right. But `relationNodes` skips non-relation keys, and `resolveSelection` only runs at each model's own compile, so a typo in a `select` at depth 3 surfaces as `UnknownFieldError` only after two queries have already run.

Minor, and consistent with where the error is raised. Flagging it because it is a hole in the "catch it at root compile" story this function otherwise commits to, and the fix is the same walk.

packages/gemi/orm/compile/plan-relations.ts
```diff
@@ -0,0 +1,890 @@
...
+
+/**
+ * `Map` keys compare by identity for objects, so two `Date`s holding the same
+ * instant are two different keys. Both sides of a stitch are decoded by the
+ * same dialect from the same declared type, so this only has to cover the
+ * decoded shapes that are objects.
+ */
+function keyOf(value: unknown): unknown {
+  return value instanceof Date ? value.getTime() : value;
+}
```
nstfkc, line 1003: **`Bytes` has the same problem `Date` has here.**

The comment scopes this to *"the decoded shapes that are objects"* and then handles one of them. A `Uint8Array` / `Buffer` join key still compares by reference, so both sides decode to distinct objects, every `byKey.get` misses, and the relation silently stitches to `[]` / `null` — the identical failure mode the `Date` branch was written to prevent, silent in the identical way.

Exotic (a `Bytes` `@id` referenced by a relation), but cheap to close: hex-encode alongside the `Date` case, or throw on an unrecognised object kind so the next one that shows up is loud rather than empty.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Fixed, and thank you for verifying it rather than describing it — I had measured under `bun test` and never ran the declared command.

Both fixes are in. `RelationExecutor` is now injected: `exec(model, op, args)` for a read through another model's `$exec`, `query(text, values)` for the one statement with no model behind it. `Model.$exec` supplies it, `compile/plan-relations.ts` imports nothing from `database/` or `foundation/`, and the whole planner — batching, stitching, key hiding, and now the m-n two-hop — is unit-testable with a fake executor and no database. `bunx vitest run orm/` is 9 files / 261 passed.

`"test": "bun --bun vitest"` too: 349 tests rather than 326, since two more files (`app/App.test.ts`, `services/router/assertNoReservedRoutePaths.test.ts`) fail to load under the node pool for the same reason, unrelated to the ORM.

One thing the injection bought beyond purity: it answers your other comment about iteration 5 for free — see below.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Right on both counts, and fixed.

`planKey` now takes the `SqlDialect` rather than its name and asks it `bindsListAsOneParameter` — so the branch stays out of `plan.ts` the same way it stays out of the compiler. Where the dialect binds the whole list to one parameter, `in` / `notIn` operands collapse to a length-independent token.

The collapse is exactly as wide as the SQL is identical:
- `in: []` keeps its own key — it compiles to a constant-false predicate, not to `= any($1)`.
- Arrays that are *structure* are untouched: `AND: [a, b]` vs `AND: [a]` and `orderBy: [a, b]` vs `orderBy: [a]` are still distinct on every dialect. Tested.
- SQLite still keys on length, because there the length really is the text.

`plan.ts`'s comments are corrected rather than left to mislead, and the four-length loop now asserts `compiles: 1, size: 1` on Postgres.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Half right, and the half that is wrong is worth writing down — I checked against Prisma before changing anything, and it disagrees with the premise.

```
include: { organization: { where: { name: "x" } } }      -> OK, returns rows
include: { organization: { orderBy: { id: "asc" } } }    -> THREW
include: { accounts:     { where: { ... } } }            -> OK
```

Prisma's own error for the `orderBy` case lists what it *does* take:

```
organization: {
  orderBy: {
  ~~~~~~~
    id: "asc"
  },
? where?: OrganizationWhereInput
}
Unknown argument `orderBy`. Available options are marked with ?.
```

So a to-one include takes `{ where?, select?, include? }`. `orderBy` is now gated on `relation.kind === "many"` with its own message — at most one row, nothing to order. `where` is not gated: doing so would have made gemi refuse a query Prisma runs, which is the divergence in the more expensive direction.

Your point about the matrix stands either way — it had no case for it. Three now: to-one `where` that matches, to-one `where` that matches nothing (relation comes back `null`, and Prisma agrees), and to-one `orderBy`, where the assertion is that *both* clients refuse it.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Better than a note: the executor injection from your first comment removes the constraint instead of documenting it.

The pair query now goes through `executor.query`, which `Model.$exec` builds from the `db` it already resolved — so both hops of a many-to-many read are on whatever connection that call resolved, including a future ambient transaction, without this file knowing anything about it.

The comment at the call site now says that, framed as the thing iteration 5 must not undo: if this hop went back to reaching for `DatabaseManager` directly, a m-n read would straddle the boundary exactly as you describe.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Fixed — you are right that it was a hole in the story this function tells about itself.

`validateSubtree` now runs `resolveSelection` at every level, so a misspelled column three levels down raises `UnknownFieldError` at root compile rather than after two queries. It validates the scalar keys and skips the relation ones, which are the walk's own business.

Same reasoning as the depth guard, so it lives in the same walk with the same comment explaining why.

packages/gemi/orm/compile/plan-relations.ts
nstfkc: Closed, and closed at both ends since the comment was writing a cheque the code was not cashing.

- `keyOf` normalises `Bytes` alongside `Date`, so a `Uint8Array` key stitches by value.
- Anything else that decodes to an object is refused **at plan time**, where the model and field are in scope for the message — `Json` is the only one Prisma could produce, and only from a hand-edited artifact. Better there than as a throw on the hot path.
- The fallthrough in `keyOf` is now an assertion rather than a reference comparison, in the style of `render()`'s placeholder-count check.

Two tests: a `Bytes` join key that stitches across distinct `Uint8Array`s holding the same bytes, and a `Json` key rejected by name. The first one needed the in-memory stand-in to compare bytes by value too — otherwise it would have passed for the wrong reason.

nstfkc:
Review addressed — all six, in `8cbaaa3`.

The two that changed the design:

**`compile/` is pure again.** You were right that `bun run test` was broken, and right that the executor was the fix rather than only the script. `RelationExecutor` is injected by `Model.$exec`; `plan-relations.ts` imports nothing from `database/` or `foundation/`. `bunx vitest run orm/` is 9 files / 261 passed, and the m-n two-hop is now unit-testable with a fake executor and no database at all. The script moves to `bun --bun vitest` as well (349 tests vs 326). It also resolves your iteration-5 comment outright: both hops of a many-to-many read now run on the connection `$exec` resolved.

**`planKey` asks the dialect.** One Postgres entry per `in` shape instead of one per list length — which from this iteration means one per distinct *parent row count*. The collapse stops exactly where the SQL stops being identical: empty lists keep their own key, and `AND` / `orderBy` arrays are untouched.

One correction, with evidence, in the thread: Prisma **accepts** `where` on a to-one include and rejects only `orderBy` (its error even lists `where?` as an available option). So `orderBy` is now kind-gated and `where` is not — gating both would have refused a query Prisma runs. Three differential cases pin it, including the mutual refusal.

Re-verified: SQLite 148 template + 261 ORM package tests; Postgres 138 (126 differential + 12 m-n, provider flipped, `TZ=UTC`); lint 0 errors; `tsc` clean in both packages.

PR #49, opened 2026-07-27
ORM iteration 4 — writes, on two dialects, verified against Prisma

nstfkc:
Iteration 4 of [the ORM plan](../blob/feat/orm/plans/orm/04-writes.md). Stacked on #47 — **review that first, and merge this into `feat/orm-03-motorcycle`**.

After this, the ORM covers everything the template application actually does today, which makes it the first point at which a real migration off the Prisma client is conceivable.

## What this adds

The full write surface — `create`, `createMany`, `update`, `updateMany`, `delete`, `deleteMany`, `upsert` — with client-side defaults, `@updatedAt`, `RETURNING`, and `select` / `include` on the returned record.

```
orm/compile/write.ts           the seven operations
orm/compile/nested-writes.ts   connect and shallow create, both directions
orm/compile/unique.ts          the unique-key matcher, lifted out of read.ts
orm/defaults.ts                cuid v1, uuid, now, @updatedAt
```

Everything runs through `Model.$exec` (invariant 1) and compiles through the same shape/value split as the reads (invariant 2): a `create` with the same field set is one plan whatever the values are, and no value is ever inlined into the SQL text — there is a test asserting no digit appears outside an identifier.

`compile/` stays a pure schema-in / SQL-out directory. Nested writes are planned there as `before` / `after` steps and *run* by the choke point, so the compiler still needs no database and no runtime import.

## Four decisions, each recorded where the code makes them

**Multi-statement writes are not atomic.** A nested `create` is more than one statement with no transaction around it: if a later step fails, the earlier rows stay written. This is a recorded decision rather than an oversight — iteration 5 introduces the ambient-transaction ALS at the choke point and makes every one of these atomic without changing a call site. Building an interim mechanism now would be work iteration 5 deletes. Noted at `NestedWriteStep`, in `$exec`, and in the test suite.

**Errors are gemi's own, not Prisma's codes.** `UniqueConstraintError` is typed, catchable, and names the fields — as *field* names, the way Prisma's `meta.target` does, not the database columns the driver reported. Mirroring `P2002` would ease migration for code that branches on it, but nothing in this repository does (checked), and carrying a `P` code implies the other fifty are implemented too.

**Row counts come from `RETURNING`, not driver metadata.** Measured rather than assumed: Bun's SQLite driver leaves `affectedRows` **null** on every statement and puts the number on an undocumented `count` property, neither of which is verifiable against Postgres from here. Counting the returned rows is one statement shape on every dialect and depends on nothing undocumented. It costs one key column per affected row on the wire — a real cost on a very large `deleteMany`, which iteration 7 can revisit with a benchmark rather than a guess.

**`createMany` fills defaults explicitly rather than grouping rows.** Prisma allows rows with different key sets in one call; the column list becomes the union, and a row missing one gets its client-side default or `NULL`. The single case that cannot be filled is a column with a *database* default that some rows supply and others do not — `NULL` would overwrite the default rather than request it, and the `DEFAULT` keyword is not available (SQLite: `near "DEFAULT": syntax error`). That case is refused by name rather than silently written wrong.

## Three bugs a real Postgres found, that SQLite could not

I ran the suite against Postgres 16 in Docker rather than trusting the loud skip. All three were live before that:

1. **Unique violations never registered on Postgres.** Bun puts the SQLSTATE on `errno`; `code` holds its own `ERR_POSTGRES_SERVER_ERROR`. Matching on `code` — the obvious reading, and what the Postgres manual would lead you to — matched nothing at all, so every violation escaped as a raw driver error.
2. **`on conflict do update set "n" = "n" + $1` is rejected as ambiguous.** Inside a conflict clause Postgres has both the existing row and `excluded` in scope, so the right-hand side has to name its table. SQLite accepts either form, and so does Postgres in a plain `update` — so the qualifier is applied where it is needed rather than everywhere.
3. **`upsert` diverged silently.** `on conflict` only fires if the insert actually collides, so `upsert({ where: { id: 1 }, create: { email } })` inserted where Prisma updated — the `create` supplies no `id`, so the update branch was unreachable. Now refused at compile time, plus a bind-time check that `where` and `create` agree on the key (compilation can prove the key is *present*, but not that it holds the same value — values do not exist at compile time).

And one more from the coercion fixture: **Postgres returns `BIGINT` as a string and `jsonb` as unparsed text**, where Prisma returns a `BigInt` and an object. The dialect had `needsDecode` false for everything, which was correct for every column the template's schema happens to have.

## Tests

- **Compiler units** — exact SQL text and parameters per dialect for all seven operations, including the multi-row column-order case, which is a data-corruption-shaped bug rather than an error: with the column list taken from row one and values taken per row by key order, a row written in a different key order binds into the wrong columns.
- **A write differential harness** comparing both the returned value *and* the resulting table contents against Prisma. The second half earns its keep: `updateMany` returns only a count, and a `create` that binds the right values into the wrong columns returns a perfectly plausible payload. Generated values are compared by *descriptor* — length and prefix — so a cuid shaped differently from Prisma's still fails.
- **Round-trip coercion** for every scalar type through a fixture schema, since the template has no Boolean, Json, BigInt or Bytes column. Decimal is absent on purpose: iteration 1 refuses it at generation time.
- **Plan-cache discrimination** over the write shapes, where a collision means wrong rows *written* rather than wrong rows read.

Verified on SQLite and on Postgres 16 through Docker:

| | |
|---|---|
| `packages/gemi` | 449 passed |
| `templates/saas-starter` | 219 passed (was 151) |
| Postgres, serial | 232 passed |
| `tsc` / `oxlint` | clean, 0 findings in ORM |

The 7 failures in `packages/gemi` are the pre-existing router/i18n ones, byte-identical to the baseline taken before this work and untouched by it.

## Notes for the reviewer

- **Postgres needs `--no-file-parallelism`.** Every suite points at the one database `TEST_POSTGRES_URL` names, and the write suites truncate it between cases; run in parallel, one file empties the tables another is reading. Documented in the harness. On SQLite each suite builds its own file, so the flag is unnecessary.
- **`prisma generate` runs the *built* generator** from `dist/bin/orm-generator.js`, not the source. `bun run build:bin` is required before regenerating after any `emit.ts` change — otherwise you get a confusing no-op diff.
- **`read.ts` shrank by ~116 lines.** `withKeyFields` moved to `select.ts` and the unique-key assertion to `unique.ts`, so the writes reuse them instead of growing parallel copies. No behaviour change; the read suite is unchanged and green.
- `Binder` gained a second parameter, a per-call `BindContext`. Existing binders ignore it and remain assignable. It carries the one instant a logical operation shares — Prisma returns `createdAt` and `updatedAt` from a single `create` as the *same* instant, verified against 6.19.2 — and whatever the `before` steps resolved.
- The generated artifacts are committed and regenerate byte-identically; flipping the datasource to postgresql and regenerating produces no diff, which is the dialect-agnosticism rule holding.

## Out of scope, deliberately

Deep nested writes (`connectOrCreate`, `set`, `disconnect`, nested `update` / `upsert` / `deleteMany`) — each throws `UnsupportedQueryError` naming itself. Cascade emulation, MySQL/MariaDB write paths, transactions (iteration 5), policies on writes (iteration 6). `skipDuplicates` on `createMany` throws.

### DIFF
--- a/packages/gemi/orm/compile/fragment.ts
+++ b/packages/gemi/orm/compile/fragment.ts
@@ -1,4 +1,5 @@
 import type { SqlDialect } from "../dialect";
+import { ParameterLimitError } from "../errors";
 
 /**
  * Pulls one value out of the argument tree at bind time. Compilation closes
@@ -106,6 +107,12 @@ export function joinFragments(parts: Fragment[], separator: string): Fragment {
 export function render(
   fragment: Fragment,
   dialect: SqlDialect,
+  /**
+   * Where the statement came from, for the parameter-ceiling error. Optional so
+   * a fragment can still be rendered in isolation by a test; every compiler
+   * passes it.
+   */
+  origin?: { model: string; operation: string },
 ): { text: string; binders: Binder[] } {
   const segments = fragment.text.split(PARAM_MARKER);
   const count = segments.length - 1;
@@ -116,6 +123,37 @@ export function render(
     throw new Error(
       `Compiled ${count} parameter placeholders but collected ` +
         `${fragment.binders.length} binders.`,
+    );
+  }
+
+  // The parameter ceiling, checked here because this is the one place that
+  // knows a *statement's* final count — and because a guard at any single
+  // compiler is a guard the next compiler forgets.
+  //
+  // It started life inside `compileCreateMany`, on the reasoning that
+  // `rows × columns` was the only count that scaled with the caller's data.
+  // That was wrong: on SQLite an `in` list binds one placeholder per element,
+  // so `findMany({ where: { id: { in: <40 000 ids> } } })` walks into the same
+  // driver error, and such a list is routinely request-derived. The relation
+  // loader has the same shape, batching an `in` over the parent keys — so a
+  // large `findMany` with an `include` reaches it without anyone writing a big
+  // list by hand. Postgres is unaffected either way: `= any($1)` is one
+  // parameter however long the array.
+  //
+  // Compile-time is exact rather than approximate, on both dialects: where the
+  // length changes the placeholder count it also changes the SQL text, so it is
+  // already part of the plan's identity.
+  if (origin && count > dialect.maxBoundParameters) {
+    throw new ParameterLimitError(
+      origin.model,
+      origin.operation,
+      count,
+      dialect.maxBoundParameters,
+      dialect.name,
+      dialect.bindsListAsOneParameter
+        ? `That is one parameter per value in the statement.`
+        : `That is one parameter per value, and on ${dialect.name} each ` +
+            `element of an 'in' list counts separately.`,
     );
   }
 
--- a/packages/gemi/orm/compile/fragment.ts
+++ b/packages/gemi/orm/compile/fragment.ts
@@ -5,8 +5,56 @@ import type { SqlDialect } from "../dialect";
  * over the *path* to a value, never the value itself — that is what makes two
  * calls with the same argument shape and different values produce byte
  * identical SQL, and it is what the plan cache depends on.
+ *
+ * Writes add a second source: values that do not exist in the argument tree at
+ * all. A `@default(cuid())` is minted per call, `@default(now())` has to be the
+ * *same* instant across every field and row of one logical operation, and a
+ * nested `create` produces a foreign key only once its own insert has run. All
+ * three arrive through the context rather than through `args`, so the binder
+ * stays a pure function of (arguments, call) and compilation stays pure.
  */
-export type Binder = (args: any) => unknown;
+export type Binder = (args: any, context: BindContext) => unknown;
+
+/**
+ * Per-call state a binder may read. Created once per `$exec`, never cached with
+ * the plan — a plan outlives the call, and `now` must not.
+ */
+export interface BindContext {
+  /**
+   * One instant for the whole operation. Prisma returns `createdAt` and
+   * `updatedAt` from a single `create` as the same instant (verified against
+   * 6.19.2), which reading the clock per field would not reproduce.
+   */
+  now: Date;
+  /**
+   * Values produced before the main statement runs, keyed by field name. Today
+   * that is exactly the foreign keys a nested `connect` or `create` resolved.
+   */
+  resolved: Record<string, unknown>;
+}
+
+export function createBindContext(): BindContext {
+  return { now: new Date(), resolved: {} };
+}
+
+/**
+ * The default context for a read. Reads never consult it, but `bind` takes one
+ * argument on every path so the choke point does not have to branch.
+ */
+const READ_CONTEXT: BindContext = { now: new Date(0), resolved: {} };
+
+/** Turns a compiled binder list into a plan's `bind`. */
+export function bindValues(
+  binders: Binder[],
+): (args: any, context?: BindContext) => unknown[] {
+  return (args: any, context: BindContext = READ_CONTEXT) => {
+    const values: unknown[] = [];
+    for (let i = 0; i < binders.length; i++) {
+      values.push(binders[i](args, context));
+    }
+    return values;
+  };
+}
 
 /**
  * Marks a parameter position inside a fragment's text. Placeholders are only
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -556,6 +556,94 @@ describe("upsert", () => {
       /2024-01-01.*2024-06-01/s,
     );
   });
+
+  /**
+   * The refusal message has to *distinguish* the two values, or it reads as
+   * "'x' is A ... but A" and looks like a framework bug rather than a wrong
+   * argument. Fixing the `Date` case is easy to half-do: `String()` collapses
+   * two different `Bytes` of the same length, and collapses `1n` against `1`.
+   */
+  test.each([
+    [
+      "Bytes of the same length",
+      new Uint8Array([1, 2]),
+      new Uint8Array([3, 4]),
+    ],
+    ["a bigint against a number", 1n, 1],
+  ])("a mismatched %s is described distinguishably", (_label, a, b) => {
+    const digest: any = {
+      name: "Digest",
+      table: "Digest",
+      fields: {
+        id: {
+          name: "id",
+          column: "id",
+          type: "Int",
+          nullable: false,
+          isId: true,
+          isUpdatedAt: false,
+          default: { kind: "autoincrement" },
+        },
+        key: {
+          name: "key",
+          column: "key",
+          type: typeof a === "bigint" ? "BigInt" : "Bytes",
+          nullable: false,
+          isId: false,
+          isUpdatedAt: false,
+        },
+        note: {
+          name: "note",
+          column: "note",
+          type: "String",
+          nullable: true,
+          isId: false,
+          isUpdatedAt: false,
+        },
+      },
+      primaryKey: ["id"],
+      uniques: [["key"]],
+      relations: {},
+    };
+
+    const args = {
+      where: { key: a },
+      create: { key: b },
+      update: { note: "n" },
+    };
+    const compiled = compileWrite(digest, "upsert", args, postgres);
+
+    let message = "";
+    try {
+      compiled.bind(args, createBindContext());
+    } catch (error: any) {
+      message = error.message;
+    }
+
+    expect(message).toMatch(/must agree on the key/);
+
+    // The two halves of "is X in the where clause but Y in 'create'" must not
+    // be the same string.
+    const [, left, right] =
+      /'key' is (.+?) in the where clause but (.+?) in 'create'/s.exec(
+        message,
+      ) ?? [];
+    expect(left).toBeDefined();
+    expect(left).not.toBe(right);
+  });
+
+  // Deliberate strictness, recorded in the plan's known-differences list:
+  // Prisma accepts a `where` naming two unique keys, but `on conflict` takes
+  // exactly one target and picking either would be silent narrowing.
+  test("a where naming two different unique keys is refused", () => {
+    expect(() =>
+      text("upsert", {
+        where: { id: 1, email: "a@b.c" },
+        create: { id: 1, email: "a@b.c" },
+        update: { name: "N" },
+      }),
+    ).toThrow(/Name exactly one unique key/);
+  });
 });
 
 describe("nested writes", () => {
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -1,14 +1,15 @@
-import { beforeEach, describe, expect, test } from "vitest";
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
 
 import { createBindContext } from "./fragment";
 import { PostgresDialect } from "../dialect/postgres";
 import { SqliteDialect } from "../dialect/sqlite";
 import {
   MissingRequiredValueError,
+  ParameterLimitError,
   UnknownFieldError,
   UnsupportedQueryError,
 } from "../errors";
-import { account, organization, user } from "../fixtures";
+import { account, bare, organization, reading, user } from "../fixtures";
 import * as registry from "../registry";
 import { compileWrite } from "./write";
 
@@ -183,6 +184,51 @@ describe("createMany", () => {
       text("createMany", { data: [{ organization: { connect: { id: 1 } } }] }),
     ).toThrow(UnsupportedQueryError);
   });
+
+  // `default values` inserts exactly one row, and there is no portable
+  // multi-row spelling of it — so this used to succeed and insert one row for
+  // three, reporting `{ count: 1 }`.
+  test("one empty row is still `default values`", () => {
+    const plan = compileWrite(bare, "createMany", { data: [{}] }, sqlite);
+    expect(plan.text).toBe(`insert into "Bare" default values returning "id"`);
+  });
+
+  test("several empty rows are refused rather than silently collapsed", () => {
+    expect(() =>
+      compileWrite(bare, "createMany", { data: [{}, {}, {}] }, sqlite),
+    ).toThrow(/All 3 rows are empty/);
+  });
+
+  // `rows × columns` is the one parameter count an application does not choose
+  // by writing the query, so it is the one that can walk into a driver limit.
+  test.each([
+    ["sqlite", sqlite, 32766],
+    ["postgres", postgres, 65535],
+  ])("the %s parameter ceiling is named, not hit", (_name, dialect, limit) => {
+    // Six columns per row: publicId, email, locale, globalRole, createdAt,
+    // updatedAt.
+    const rows = Math.ceil(limit / 6) + 1;
+    const data = Array.from({ length: rows }, (_, i) => ({
+      email: `a${i}@b.c`,
+    }));
+
+    expect(() => compileWrite(user, "createMany", { data }, dialect)).toThrow(
+      ParameterLimitError,
+    );
+    expect(() => compileWrite(user, "createMany", { data }, dialect)).toThrow(
+      new RegExp(`accepts at most ${limit}`),
+    );
+  });
+
+  test("a createMany just under the ceiling compiles", () => {
+    const data = Array.from({ length: 5461 }, (_, i) => ({
+      email: `a${i}@b.c`,
+    }));
+    // 5461 × 6 = 32766, exactly SQLite's limit.
+    expect(() =>
+      compileWrite(user, "createMany", { data }, sqlite),
+    ).not.toThrow();
+  });
 });
 
 describe("update", () => {
@@ -221,6 +267,43 @@ describe("update", () => {
     );
     const plan = compileWrite(user, "updateMany", args, sqlite);
     expect(plan.shape([{ id: 1 }, { id: 2 }, { id: 3 }])).toEqual({ count: 3 });
+  });
+
+  // There is no nested-write planner behind `updateMany` — Prisma types it as
+  // unchecked input and rejects a relation key. Without this the key was
+  // neither executed nor reported: `suppliedFields` skips relations silently,
+  // so the `connect` simply vanished from an otherwise successful statement.
+  test("a nested write in updateMany data is refused by name", () => {
+    expect(() =>
+      text("updateMany", {
+        where: { id: 1 },
+        data: { name: "x", organization: { connect: { id: 2 } } },
+      }),
+    ).toThrow(/updateMany cannot contain nested writes/);
+  });
+
+  // And the degenerate version, which used to report the wrong problem
+  // entirely: "At least one field must be updated".
+  test("updateMany data of only a relation names the relation", () => {
+    expect(() =>
+      text("updateMany", {
+        where: { id: 1 },
+        data: { organization: { connect: { id: 2 } } },
+      }),
+    ).toThrow(/updateMany cannot contain nested writes/);
+  });
+
+  test("update still accepts the same nested write", () => {
+    registry.clearRegistry();
+    registry.register("User", class { static $schema = user });
+    registry.register("Organization", class { static $schema = organization });
+    expect(() =>
+      text("update", {
+        where: { id: 1 },
+        data: { name: "x", organization: { connect: { id: 2 } } },
+      }),
+    ).not.toThrow();
+    registry.clearRegistry();
   });
 
   test.each([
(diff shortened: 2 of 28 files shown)


nstfkc (commented):
Reviewed against a clean checkout at `887e4e1`. Verified the claims rather than taking them: `vitest run` in `packages/gemi` gives **449 pass / 7 fail**, and all 7 are in `services/router` and i18n — untouched by this PR. `vitest run app/models/` in the template gives **216 pass**. `tsc --noEmit` and `oxlint orm/ bin/orm/` are both clean.

The architecture holds: `compile/` stays free of runtime imports, everything routes through `$exec`, no value reaches the SQL text, and the discrimination table covers the write shapes including `createMany` row counts and both nested-write directions. The three Postgres bugs are real ones SQLite genuinely cannot surface, and the `RETURNING`-for-counts and `createMany`-defaults decisions are sound and recorded where the code makes them. The differential harness comparing *table contents* — not just the returned payload — is the part that earns its keep, and the volatile-field descriptor comparison is better than skipping those fields.

Six findings inline. Three of them are silent divergences on writes, which is the specific failure mode the rest of this PR goes to unusual lengths to prevent:

1. `upsert` silently drops extra non-unique `where` filters (confirmed by compiling one).
2. `updateMany` silently drops nested relation writes in `data` (confirmed) — `createMany` refuses them by name.
3. `withConflictKeyCheck` compares encoded values by identity, so a valid upsert on a non-primitive key is rejected with a message printing two identical strings (reproduced with a `DateTime @unique` on Postgres).

The other three (`createMany` of all-empty rows, no parameter-count ceiling, `delete` + `include` after a cascade) are narrower and reasonable follow-ups.

One structural note beyond the diff: the Postgres path is guarded by a loud skip and there is no test workflow in `.github/workflows` at all, so the three Postgres-only fixes are protected only by someone remembering to run Docker. Not this PR's debt, but this is the first surface that materially depends on it.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+ * both see "absent" and both insert, and one gets a unique violation. Since
+ * there is no transaction until iteration 5, the single-statement form is not
+ * just faster here, it is the only correct one.
+ *
+ * The conflict target comes from the `where`'s unique key, so it is a schema
+ * identifier like every other. Nested writes are refused inside an upsert:
+ * they would have to run before an insert that may turn out to be an update,
+ * and there is nothing sensible to do with the row they created if it does.
+ */
+function compileUpsert(
```
nstfkc, line 453: **`upsert` silently drops extra non-unique `where` filters.**

`where` is used only for the conflict target and the bind-time key check — the rest is never compiled. Confirmed by compiling one:

```
where: { publicId: "p", deletedAt: null }
→ insert into "User" (...) values (...) on conflict ("publicId") do update set ...
```

`deletedAt` vanishes. `unique.ts:18` explicitly documents that Prisma 5 allows extra filters in a `WhereUniqueInput` and that they narrow further — and `update` / `delete` honour that, since they run the whole `where` through `compileWhere`. Only `upsert` drops them, and the outcomes genuinely differ: Prisma finding no match would attempt the create and hit a unique violation, where this updates the row.

Given this function already refuses the `create`-omits-the-key shape rather than diverging quietly, refusing a `where` with non-key members looks more consistent than silently narrowing it.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+  schema: ModelSchema,
+  op: Operation,
+  args: any,
+  dialect: SqlDialect,
+): QueryPlan {
+  const many = op === "updateMany";
+
+  if (!many) matchUniqueKey(schema, args?.where, op);
+
+  const nested = many
```
nstfkc, line 353: **`updateMany` silently drops nested relation writes.**

`planNestedWrites` is skipped when `many`, and `suppliedFields` skips relation keys without complaint, so the nested write is neither executed nor refused:

```
updateMany({ where: { id: 1 }, data: { name: "x", organization: { connect: { id: 2 } } } })
→ update "User" set "name" = ?, "updatedAt" = ? where "id" = ? returning "id"
```

The `connect` is gone, no error. Prisma rejects it (`Unknown argument 'organization'`). `createManyColumns` refuses nested writes by name at line 235 — `updateMany` deserves the same treatment. Note also that if `data` contains *only* a relation key, the caller gets the misleading `"At least one field must be updated"` instead.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+
+    const locate = whereValue(key, column.field.name);
+
+    return {
+      field: column.field,
+      value: (args, context) => {
+        const inserted = column.value(args, context);
+        const selected = dialect.encode(locate(args), column.field);
+
+        if (inserted !== selected) {
```
nstfkc: **Identity comparison rejects valid upserts on non-primitive keys.**

`inserted !== selected` is a reference comparison for any encoded value that is not a primitive: Postgres `DateTime` (pass-through `Date`) and `Bytes` on both dialects. Two `Date` objects for the same instant are never `===`, so a correct call is refused — with a message that prints two identical strings:

```
THREW: 'at' is "Tue Nov 14 2023 23:13:20 GMT+0100" in the where clause
       but "Tue Nov 14 2023 23:13:20 GMT+0100" in 'create'.
```

Reproduced by compiling an upsert on a fixture with `at DateTime @unique` against `PostgresDialect`. SQLite is unaffected because it encodes `DateTime` to a number, which is exactly the kind of asymmetry the rest of this PR hunts down.

Needs a value comparison that normalizes `Date` to `getTime()` and typed arrays byte-wise before falling back to `!==`.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+}
+
+/**
+ * `(cols) values (?, ?), (?, ?)`, or `default values` when the row supplies no
+ * column at all — legal on both dialects, and the only way to insert a row made
+ * entirely of database defaults.
+ */
+function valuesClause(grid: WriteColumn[][], dialect: SqlDialect): Fragment {
+  const columns = grid[0];
+  if (columns.length === 0) return sql(" default values");
```
nstfkc, line 763: **A `createMany` of all-empty rows inserts one row.**

The column list comes from `grid[0]`, and the zero-column short-circuit discards the rest of the grid entirely:

```
createMany({ data: [{}, {}, {}] })   // model where no field has a client-side default
→ insert into "Bare" default values returning "id"   → { count: 1 }
```

Prisma inserts 3. Narrow — it needs a model where the union of supplied keys and client-side-default fields is empty, so the template's schema cannot reach it — but the result is a wrong row count and a wrong `count`, rather than an error, which is the failure class this file is otherwise arranged to prevent.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+    valuesClause([columns], dialect),
+    returning.clause,
+  );
+
+  return plan(statement, dialect, op, returning, nested);
+}
+
+// --- createMany ------------------------------------------------------------
+
+function compileCreateMany(
```
nstfkc, line 156: **No parameter-count ceiling on `createMany`.**

A multi-row insert binds `rows × columns` parameters. SQLite's `SQLITE_MAX_VARIABLE_NUMBER` is 32766, and the Postgres wire protocol caps parameters at 65535 (the count is an int16) — so a `createMany` of a few thousand template `User` rows fails at the driver with a message naming neither the model nor the cause. Prisma chunks automatically.

The chunking itself is fair to leave to iteration 7, but this looks like a candidate for the same treatment `ReturningUnsupportedError` gets: a named gap rather than a driver error, so an import path fails legibly instead of mysteriously.

Secondary, for the same iteration: `canonicalShape` walks the whole `data` array on *every* call, hit or miss, so a large `createMany` pays an O(rows × fields) key build per call and stores a key of that size.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1094 @@
...
+  if (op === "delete") matchUniqueKey(schema, args?.where, op);
+
+  const where = compileWhere(
+    schema,
+    args?.where,
+    { dialect, operation: op },
+    (callArgs) => callArgs?.where,
+  );
+
+  const returning = returningClause(schema, args, dialect, op, undefined);
```
nstfkc, line 425: **`delete` with `include` loads relations after the row is gone.**

`returningClause` plans relations for `delete`, and `$exec` attaches them once the statement has run. On a schema with `onDelete: Cascade` the children are already deleted by then, so the `include` comes back empty where Prisma — which does the delete inside a transaction — returns them.

The template schema declares no cascades, so the differential harness cannot see this. Worth a recorded note here, or a `delete with include` case against a cascading fixture once iteration 5 provides the transaction that would actually fix it.

packages/gemi/orm/compile/write.test.ts
```diff
@@ -0,0 +1,569 @@
...
+        create: { organization: { connect: { id: 1 } } },
+        update: {},
+      }),
+    ).toThrow(/Nested writes inside an upsert/);
+  });
+});
+
+describe("nested writes", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
```
nstfkc, line 651: Nit: this clears the registry in `beforeEach` but never in `afterEach`, where `compile/relations.test.ts:129` does both.

Invisible under vitest, which isolates modules per file. Under a runner that shares process state it leaks: `bun test orm/` fails `read.test.ts:288` ("a relation in a select needs the registry") because `Account` is still registered from this file. One `afterEach(() => registry.clearRegistry())` restores the local convention and the isolation.

packages/gemi/orm/plan.writes.discrimination.test.ts
```diff
@@ -0,0 +1,255 @@
...
+  planKey,
+  type Operation,
+} from "./plan";
+
+const sqlite = new SqliteDialect();
+const postgres = new PostgresDialect();
+
+beforeEach(() => {
+  clearPlanCache();
+  registry.clearRegistry();
```
nstfkc, line 21: Same nit as `compile/write.test.ts`: registered in `beforeEach`, never cleared in `afterEach`, so the registry outlives the file under a shared-process runner.

The table itself is excellent — `createMany` row counts, the two `connect` forms compiling to different plans, and the same-count-different-union case are exactly the collisions that would be silent.

templates/saas-starter/app/models/differential.ts
```diff
@@ -180,6 +274,55 @@ export async function createDifferential(options: {
...
+      await this.reset();
+      clearPlanCache();
+      const gemiModel = options.models[model];
+      if (!gemiModel) throw new Error(`No gemi model registered for ${model}.`);
+      const fromGemi = await settle(() => (gemiModel as any)[operation](args));
+      const afterGemi = await readTables(prisma, tables);
+
+      if (fromPrisma.threw || fromGemi.threw) {
+        expect(
+          { threw: fromGemi.threw, at: label },
```
nstfkc: Minor: this compares only the *fact* of throwing, not the kind — a gemi error thrown for an unrelated reason (a compile-time refusal, say) passes as agreement with a Prisma unique violation.

The dedicated tests below the table do check the error type, so this is table-level looseness rather than a hole. Comparing a coarse class — `threw` plus whether it is a `UniqueConstraintError` — would close it without coupling to message text.

packages/gemi/orm/dialect/index.ts
```diff
@@ -43,6 +43,37 @@ export interface SqlDialect {
...
+   * parameter count as an int16 in the Bind message, and SQLite compiles
+   * `SQLITE_MAX_VARIABLE_NUMBER` in. Only `createMany` can approach either,
+   * since it is the one operation whose parameter count scales with the
+   * caller's data — `rows × columns` — rather than with the query's shape.
+   *
+   * It lives here rather than as a constant in the write compiler for the usual
+   * reason: the number differs per dialect, and the compiler is not allowed to
+   * know which dialect it is compiling for.
+   */
+  readonly maxBoundParameters: number;
```
nstfkc, line 86: The new capability is right, but the claim beside it isn't quite — `createMany` is **not** the only operation whose parameter count scales with the caller's data. On SQLite an `in` list binds one placeholder per element, and nothing checks it:

```
findMany({ where: { id: { in: <40 000 ids> } } })
→ compiles, plan.bind(...).length === 40000   (limit 32766)
```

Verified against `3b51cd4`. `assertParameterCount` is called only from `compileCreateMany`, so the read path walks into the same driver error this PR just taught `createMany` to avoid — and the list length is routinely request-derived (`?ids=…`), which `plan.ts` already flags as reachable from untrusted input. Postgres is unaffected: `= any($1)` is one parameter.

The relation loader has the same shape, batching an `in` over the parent keys, so a large `findMany` with an `include` can reach it without anyone writing a big list by hand.

Small fix given the machinery now exists — the guard wants to be in the SQLite `inList` path (or wherever the read compiler knows the length). Worth either doing here or narrowing the comment to say the ceiling is currently enforced only for writes.

nstfkc (commented):
Re-reviewed `3b51cd4` in an isolated worktree. All six findings and all three nits are addressed, and each fix landed with a test.

| Verified at `3b51cd4` | |
|---|---|
| `vitest run orm/` | **382 pass** (was 369 — +13 for the fixes) |
| `vitest run app/models/` (template) | **216 pass** |
| `tsc --noEmit` | clean |

**Resolved:**

1. **`upsert` extra `where` members** — refused by name, with a message that says why an `on conflict` target cannot hold a predicate. Verified: `{ publicId, deletedAt: null }` now throws.
2. **`updateMany` nested writes** — refused, and factored into `assertNoNestedWrites` shared with the `upsert` branches. The "data of only a relation" case now names the relation instead of reaching "At least one field must be updated."
3. **Identity comparison** — `sameEncoded` handles `Date` by `getTime()` and typed arrays byte-wise, and `describe` no longer runs a `Date` through `String()`. Both the equal and the genuinely-differing `DateTime` cases are tested against the new `reading` fixture.
4. **All-empty `createMany`** — refused above 1 row; the single-row `default values` case is preserved and pinned by a test.
5. **Parameter ceiling** — `maxBoundParameters` as a dialect capability plus `ParameterLimitError`, with a boundary test at exactly 32766 (5461 × 6). Cleanly placed: the compiler asks the dialect rather than knowing the number.
6. **`delete` + `include`** — recorded at `compileDelete` as a known divergence with the transaction-shaped fix named. That is what I was asking for.

Nits all addressed too: `afterEach` registry clears in both write test files, and `failureKind` gives the harness a coarse unique/notFound/other class instead of only the fact of throwing.

The two new fixtures (`reading` for `DateTime @unique` + `Bytes`, `bare` for no-client-side-default) are the right shape — they exist precisely because the template's schema cannot reach those paths.

Three small follow-ups inline, none blocking. Nothing here should hold up the merge.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1277 @@
...
+      if (left[i] !== right[i]) return false;
+    }
+    return true;
+  }
+
+  return false;
+}
+
+/** A value for an error message, without pretending a `Date` is a string. */
+function describe(value: unknown): string {
```
nstfkc, line 707: The `Date` branch fixes the case I reported, but the `Bytes` branch reintroduces it one type over: two different `Uint8Array`s of the same length both describe as `"2 bytes"`, so a genuinely-mismatched `Bytes` conflict key produces the same "is X … but X" message that made the `Date` case look like a framework bug.

A short hex prefix would keep it honest and stay bounded — `2 bytes (0x0102)`, truncated past a few bytes.

Same class, even narrower: `1n` vs `1` falls through `sameEncoded` to `false` and describes as `"1"` on both sides. Prisma coerces a number into a `BigInt` field, so it's reachable, if barely. Either normalize `bigint`/`number` in `sameEncoded` or include the `typeof` in `describe` — the latter fixes both cases at once.

packages/gemi/orm/compile/write.ts
```diff
@@ -0,0 +1,1277 @@
...
+  // honour that for free: their whole `where` goes through `compileWhere`.
+  //
+  // An upsert cannot. Its `where` becomes an `on conflict (...)` target, which
+  // is a *key*, not a predicate — there is nowhere to put `deletedAt: null`.
+  // Compiling only the key part would mean `where: { publicId, deletedAt: null }`
+  // updating a soft-deleted row that Prisma would have left alone (Prisma would
+  // find no match, attempt the create, and hit a unique violation). Refused
+  // rather than narrowed, for the same reason the create-omits-the-key shape a
+  // few lines down is.
+  const named = key.length > 1 ? key.join("_") : key[0];
```
nstfkc, line 485: Worth being explicit in the PR description, since it is a deliberate strictness rather than a bug: this refuses a `where` naming **two different unique keys**, which Prisma accepts.

```
upsert({ where: { id: 1, publicId: "p" }, ... })
→ "The where clause carries publicId beside the unique key 'id'."
```

I think refusing is the right call — `on conflict` takes one target, and picking `id` while ignoring `publicId` is exactly the silent narrowing this commit set out to remove. But it means a migrating application hits a compile error on a call Prisma ran happily, and the error text (correctly) explains the `on conflict` mechanics rather than saying "pick one key." Adding that to the out-of-scope/known-differences list would save the next reader the trip through this function.

One micro-nit on the mechanics: `named` is derived from whichever candidate `matchUniqueKey` returned first (primary key, then `uniques` in order), so with two keys present the one reported as "the unique key" is an artifact of declaration order rather than of anything the caller wrote.

nstfkc:
All three follow-ups addressed in `a2f5155`.

**The `in`-list gap was the real one, and the fix is broader than the report.** You were right that `createMany` is not the only shape that scales with the caller's data — but the conclusion I drew from that is that a guard living in *any single compiler* is a guard the next compiler forgets. So it moved into `render`, which is the one place that sees a statement's final parameter count:

```ts
if (origin && count > dialect.maxBoundParameters) throw new ParameterLimitError(…)
```

Every compiler now passes an origin, so the check is unconditional and covers `createMany`, both read paths, and the implicit m-n join query at once. The `createMany`-only helper is gone. Compile-time is exact rather than approximate on both dialects, since where the list length changes the placeholder count it also changes the SQL text — so it is already part of the plan's identity.

Five tests, including the two you'd want: the same 40 000-element list is one parameter on Postgres (`= any ($1)`), and two 20 000-element lists in one `where` trip the ceiling — which is the case a per-clause guard would have missed.

The claim in `dialect/index.ts` now says what is true: three shapes can approach the limit, and Postgres escapes two of them.

**`describe` — you caught the same bug one type over, twice.** Correct on both counts. Bounded hex prefix for `Bytes` (`2 bytes (0x0102)`, truncated past 8), and the `typeof` suffix on non-strings, which fixes `1n` vs `1` at the same time. The test asserts the general property rather than the two instances: parse the "is X … but Y" halves out of the message and assert they differ.

**The two-unique-keys strictness is now written down.** New "Known differences from Prisma, as shipped" section in `plans/orm/04-writes.md`, covering all six deliberate refusals with the reason each one cannot be expressed rather than only that it is refused. The error text also leads with the actionable half now — "Name exactly one unique key here" — before explaining the `on conflict` mechanics.

On your micro-nit: correct, and I left it as-is with a comment saying so. Which key gets *named* is an artifact of declaration order, there is no principled way to prefer one, and the message lists every other member by name regardless — so the caller has what they need either way.

| | |
|---|---|
| `vitest run orm/` | **390 pass** (was 382) |
| `vitest run app/models/` | **216 pass** |
| `tsc --noEmit` | clean |

All three fixes are compile-layer, and both dialects are covered by unit tests, so this needed no Docker round-trip.

#50 (iteration 5, ambient transactions) is stacked on this branch and has been merged forward — 400 orm tests, 234 in the template.

packages/gemi/orm/compile/fragment.ts
```diff
@@ -58,6 +107,12 @@ export function joinFragments(parts: Fragment[], separator: string): Fragment {
 export function render(
   fragment: Fragment,
   dialect: SqlDialect,
+  /**
+   * Where the statement came from, for the parameter-ceiling error. Optional so
+   * a fragment can still be rendered in isolation by a test; every compiler
+   * passes it.
+   */
+  origin?: { model: string; operation: string },
```
nstfkc, line 115: Making this optional reintroduces, in miniature, the hole the commit message says it closed: "a guard at any single compiler is a guard the next compiler forgets" — but a new `render` call site that omits `origin` skips the ceiling silently, which is the same failure one level down.

The stated reason is "so a fragment can still be rendered in isolation by a test." No test does: `grep -rn "render(" --include="*.test.ts"` over `packages/gemi` returns nothing, and `render` is exported from neither `orm/index.ts` nor `compile/index.ts`, so the only callers are the six inside `compile/`. All six pass it today (verified).

So making it required costs one line at each of the six and turns "every compiler passes it" from a convention into something `tsc` enforces — which is the standard the rest of this file holds itself to.

(If a future test really does want to render bare, passing `{ model: "test", operation: "test" }` is cheaper than the loophole.)

nstfkc (commented):
Reviewed `a2f5155`. All three follow-ups are resolved, and the parameter-ceiling fix is better than what I suggested — I proposed patching the SQLite `in`-list path, which would have left the next statement shape to be found the same way. Moving the check into `render` puts it at the one place that sees a statement's final count, so `createMany`, both read paths and the implicit m-n join query are covered by one guard.

| Verified at `a2f5155` | |
|---|---|
| `vitest run orm/` | **390 pass** (was 382) |
| `vitest run app/models/` | **216 pass** |
| `tsc --noEmit` | clean |

Re-ran the two probes from the last round against this commit:

```
sqlite:   ParameterLimitError: User.findMany would bind 40000 parameters,
          and the 'sqlite' driver accepts at most 32766 in one statement
postgres: compiled          <- correct, `= any($1)` is one parameter

date mismatch: 'at' is "2023-11-14T22:13:20.000Z" in the where clause
               but "2023-11-14T22:13:29.999Z" in 'create'
```

Both fixed. `describe` now carries a bounded hex prefix for `Bytes` and a `typeof` suffix for everything non-string, so the `1n` vs `1` case falls out too — and the `Array.from` note about `Uint8Array.prototype.map` coercing strings back to bytes is the kind of trap worth a comment.

"The ceiling counts the whole statement, not one clause" is the test that earns the refactor, and the `= any($1)` counterpart pins that Postgres genuinely escapes it rather than being untested.

The **Known differences from Prisma, as shipped** section in `plans/orm/04-writes.md` is more than I asked for — six refusals with the reason and the call-site workaround for each. That is the document a migrating application actually needs.

One small residual, inline. Nothing blocking.

nstfkc:
**Correction to my own verification claims on this PR and on #50.**

While starting iteration 6 I found that `packages/gemi/tsconfig.json` lists directories explicitly and **`orm` is not among them.** Only the seven ORM files reachable from an included directory were ever type-checked — `facades/DB.ts` imports `orm/context.ts`, and so on. `Model.ts`, the whole compiler, both dialects and every error type were not.

So every "`tsc --noEmit` clean" in this PR's description and in my review replies meant much less than I stated. Pushed as `9ea8bfc`.

Adding `orm` to `include` surfaced **three real errors, all mine from iteration 4** — the commit in this PR that widened `Binder` from `(args) => unknown` to `(args, context) => unknown`:

```
orm/dialect/postgres.ts:75        values(args)       — missing context
orm/dialect/sqlite.ts:109         values(args)       — missing context
orm/compile/plan-relations.ts:896 binder(undefined)  — missing context
```

Benign today only by luck: all three wrap binders that read neither argument, so the missing parameter is never dereferenced. A binder on any of those paths that read `context.now` — which is exactly what `defaultBinder` does — would have thrown. The two dialect ones are in `inList`, which every `in` list and every batched relation read goes through.

Fixed by forwarding the context rather than loosening the type. `tsc --noEmit` is clean over the ORM for the first time; 390 orm tests still pass.

Worth flagging for the re-review: **the type-level half of every earlier iteration's verification is only being established now.** Anything in `orm/` you previously took as type-checked was not, on iterations 1 through 5.

Merged forward into #50.

nstfkc:
Fixed in `bf1ad0e`. You're right, and the reasoning is the one I wrote in the commit message a round earlier and then didn't apply to my own escape hatch: optional `origin` is a per-compiler guard one level down.

I checked the justification I'd given and it was false — `render` is exported from neither `orm/index.ts` nor `compile/index.ts`, no test calls it, and all six callers already pass it. So the loophole bought nothing and cost the guarantee. Required now, which turns "every compiler passes it" into something `tsc` enforces.

390 orm tests, `tsc --noEmit` clean. Merged forward through #50.

PR #50, opened 2026-07-27
ORM iteration 5 — ambient transactions, on two dialects

nstfkc:
`Model.transaction(cb)` — every query inside the callback joins it, at any call depth, with no `tx` threaded through anything.

```ts
await Model.transaction(async () => {
  const user = await User.create({ data: { email } })
  await audit(user)          // its queries are in the transaction too
})
```

This is one of the two features that justify the project. Prisma's `$transaction` hands back a distinct client, so every participating function must accept it — the transaction ends up in the signature of everything between the boundary and the query.

Stacked on #49. Review that one first.

## The integration is one line

```ts
const conn = currentTransaction() ?? db.sql
```

That covers every operation, both directions of nested write, the relation reads an `include` fans out into, and the one query with no model behind it (the implicit m-n join table) — but only because invariant 1 held through iterations 1–4. An operation that had acquired a private path to the database would show up here as a statement committing while its neighbours roll back. This suite is the test that would fail first.

## The open decision, now settled

The plan flagged this one as needing sign-off, and it has it. The transaction lives in a **second, ORM-owned `AsyncLocalStorage`** (`orm/context.ts`) holding `{ tx, depth }` and nothing else. `kernel/context.ts` keeps carrying only the Application — and now carries the reasoning too, so the "exactly one ALS" claim there does not silently become false.

Both alternatives are rejected on the record:

- **Re-entering the kernel store with a `{ app, tx }` wrapper.** Keeps the single-store claim literally true, at the cost of every reader of that store — `app()` above all — handling two shapes forever. A miss there is not a transaction bug, it is a service resolving off the wrong Application.
- **Hanging the handle off the Application.** One Application serves every concurrent request, so two overlapping transactions overwrite each other's handle and statements land in the wrong one. Silent data corruption, not a style question. It fails acceptance criterion 4 by construction.

The honest cost is two stores on the hot path of every query. The ORM's is shallow and entered only inside a transaction, so a query outside one pays a single `getStore()` returning `undefined`.

## Three spike findings that shaped the code

Verified against both dialects before building on them, per the plan's instruction not to assume Bun's semantics:

1. **`tx.begin` inside a transaction is refused outright** — *"cannot call begin inside a transaction use savepoint() instead"*. So nesting is a savepoint, which is the better semantics anyway: an inner failure rolls back to it and leaves the outer transaction usable.
2. **Bun hands the savepoint callback the same handle object** (`sp === tx`), so depth is the only thing that actually changes on the way in. The callback's argument is still what gets stored, so this keeps working if that stops being true.
3. **The handle stays callable after `begin` resolves.** Queries on a leaked one run on the pooled connection, outside any transaction, and *succeed* — no error to notice. This is the one that shaped the API: it is why `Model.transaction`'s callback takes no argument at all, why `currentTransaction()` is documented as read-and-use rather than something to store, and why there is a test asserting the scope does not outlive the callback.

## `DB.transaction` interop

It keeps its existing signature and still hands its callback the raw handle, but routes through the same `withTransaction`. Without that the two would be transaction systems that silently ignore each other: a `User.create` inside a `DB.transaction` would commit separately on a pooled connection and survive the rollback, and a `DB.transaction` nested inside a `Model.transaction` would throw on Bun's `begin` guard. Both directions are tested.

## Writes are atomic exactly when the caller asks

`$exec` does not open a transaction per call. It cannot know whether it is one step of a larger unit, and doing so would put a `BEGIN` around every query in the framework and turn a caller's transaction into a nest of savepoints they never asked for. The not-atomic note iteration 4 left in `Model.$exec` is rewritten to say that rather than deleted — outside a transaction, a nested write's later step failing still leaves the earlier rows written.

## Tests

**`packages/gemi/orm/context.test.ts`** — the storage property, with fake handles and no database. The one that earns its keep: two concurrent transactions with awaits interleaved between every step, asserting each one's *statements* landed on its own handle, not just that each one read the right value. That is the test that fails if the handle ever moves somewhere shared, and a fake makes the interleaving deterministic in a way a real database is not.

**`templates/saas-starter/app/models/transactions.test.ts`** — the rest, against both dialects: commit and rollback, rollback of everything rather than the last statement, a query three frames deep through a service that never heard of the transaction, an `include`'s nested relation reads, a nested write's extra `before`/`after` statements, savepoint nesting three deep with depth asserted at each level, inner-rollback-outer-survives and inner-success-outer-rollback, `DB.transaction` interop both ways, no ambient handle outside a transaction, a rolled-back transaction not poisoning the next query, and no plan-cache pollution.

The two real-database concurrency cases are **Postgres-only and say why**: SQLite allows one writer at a time, so two concurrent write transactions there serialise or fail with `SQLITE_BUSY`. The interleaving cannot happen, and a test that "passes" because nothing overlapped proves nothing.

## Verification

| | |
| --- | --- |
| `packages/gemi` | 472 passed, 7 failed |
| baseline (stashed) | 462 passed, 7 failed |
| `templates/saas-starter` (SQLite) | 234 passed |
| Postgres 16 (Docker, `--no-file-parallelism`) | 267 passed |
| `tsc --noEmit` | clean |
| `oxlint` | 2 warnings, both pre-existing in `facades/Auth.ts` |
| `bun run build:core` + `build:types` | clean |

The 7 failures are the same pre-existing router/i18n ones, byte-identical to the baseline — confirmed by stashing this branch's changes and re-running. On the Postgres pass the only failing files are the two differential suites that cannot run with a Postgres-generated client, which is documented behaviour in `differential.ts`.

## For the reviewer

- **Postgres runs need `--no-file-parallelism`.** Every suite points at the one database `TEST_POSTGRES_URL` names, and this suite truncates it between cases. The new suite truncates *all* tables rather than the three it writes to — a row another suite left in `SocialAccount` otherwise makes its `DELETE FROM "User"` fail on a foreign key, and the failure lands here.
- `applyMigrations` is now exported from `differential.ts` so the transaction suite can build a SQLite database without pulling in a Prisma client.
- I confirmed again that the generated artifacts are dialect-agnostic: flipping the datasource to `postgresql` and regenerating produced no diff under `app/models/generated/`.

## Not in scope

Isolation-level configuration, retry-on-serialization-failure, distributed transactions, and a unit-of-work model, all per the plan. Also left alone: the optional development-mode warning for long-lived transactions — cheap, but the plan marks it not required, and it wants a duration threshold that should come from a measurement rather than a guess.

One thing iteration 5 now *makes* fixable, recorded in `compileDelete` during the #49 review: a `delete` with an `include` on a cascading relation reads the children after the database has removed them. The fix is to read relations before the delete inside one transaction. It is not done here because the template schema declares no cascades, so there is nothing to test it against yet.

### DIFF
--- a/packages/gemi/kernel/context.ts
+++ b/packages/gemi/kernel/context.ts
@@ -1,3 +1,35 @@
 import { AsyncLocalStorage } from "async_hooks";
 
+/**
+ * The kernel's AsyncLocalStorage. It holds the Application, and nothing else —
+ * `Kernel.run()` enters it once per request, websocket message or cron tick,
+ * and `foundation/app.ts` reads it to resolve services.
+ *
+ * There was exactly one AsyncLocalStorage in the framework until iteration 5 of
+ * the ORM. There are now two, and this note is here so that claim does not
+ * quietly become false where someone would otherwise still be relying on it.
+ *
+ * The second is `packages/gemi/orm/context.ts`, holding an open transaction
+ * handle and its nesting depth. It is deliberately separate rather than merged
+ * into this one:
+ *
+ * - A transaction scope nests *inside* an Application scope without replacing
+ *   it. Re-entering this store with a `{ app, tx }` wrapper would mean every
+ *   reader of it — `app()` above all — has to handle two shapes forever, and a
+ *   miss there is not a transaction bug, it is a service resolving off the
+ *   wrong Application.
+ * - The alternative of hanging the handle off the Application is worse than a
+ *   style question: one Application serves every concurrent request, so two
+ *   overlapping transactions would overwrite each other's handle and statements
+ *   would land in the wrong one. That is silent data corruption.
+ *
+ * The cost is honest: two stores on the hot path of every query instead of one.
+ * The ORM's is shallow — one small object, entered only inside
+ * `Model.transaction` or `DB.transaction`, never per request — so a query
+ * outside a transaction pays one `getStore()` returning undefined.
+ *
+ * If a third is ever proposed, this is the note to read first: the bar is that
+ * the state is genuinely per-async-scope *and* cannot live in an existing
+ * store without making that store's readers ambiguous.
+ */
 export const kernelContext = new AsyncLocalStorage();
--- a/packages/gemi/kernel/context.ts
+++ b/packages/gemi/kernel/context.ts
@@ -5,13 +5,22 @@ import { AsyncLocalStorage } from "async_hooks";
  * `Kernel.run()` enters it once per request, websocket message or cron tick,
  * and `foundation/app.ts` reads it to resolve services.
  *
- * There was exactly one AsyncLocalStorage in the framework until iteration 5 of
- * the ORM. There are now two, and this note is here so that claim does not
- * quietly become false where someone would otherwise still be relying on it.
+ * "The framework's only AsyncLocalStorage" is a claim that used to sit here. It
+ * was already not true — the *kernel's* is the only one, but `http/`, `pubsub/`
+ * and now the ORM each keep their own:
  *
- * The second is `packages/gemi/orm/context.ts`, holding an open transaction
- * handle and its nesting depth. It is deliberately separate rather than merged
- * into this one:
+ *   kernel     kernel/context.ts                     the Application
+ *   request    http/requestContext.ts                req, user, cookies, locale
+ *   broadcast  services/pubsub/BroadcastManager.ts   headers, cookies
+ *   orm        orm/context.ts                        an open transaction
+ *
+ * The pattern is one store per scope with one owner, entered by that owner and
+ * read through its own accessor — `app()`, `RequestContext.getStore()`,
+ * `currentTransaction()`. What matters is that this one holds the Application
+ * and nothing else, because `foundation/app.ts` reads it unconditionally.
+ *
+ * The ORM's, added in iteration 5, holds an open transaction handle and its
+ * nesting depth. It is deliberately separate rather than merged into this one:
  *
  * - A transaction scope nests *inside* an Application scope without replacing
  *   it. Re-entering this store with a `{ app, tx }` wrapper would mean every
@@ -23,13 +32,13 @@ import { AsyncLocalStorage } from "async_hooks";
  *   overlapping transactions would overwrite each other's handle and statements
  *   would land in the wrong one. That is silent data corruption.
  *
- * The cost is honest: two stores on the hot path of every query instead of one.
- * The ORM's is shallow — one small object, entered only inside
- * `Model.transaction` or `DB.transaction`, never per request — so a query
- * outside a transaction pays one `getStore()` returning undefined.
+ * The cost is honest: one more store on the hot path of every query. The ORM's
+ * is shallow — one small object, entered only inside `Model.transaction` or
+ * `DB.transaction`, never per request — so a query outside a transaction pays
+ * one `getStore()` returning undefined.
  *
- * If a third is ever proposed, this is the note to read first: the bar is that
- * the state is genuinely per-async-scope *and* cannot live in an existing
- * store without making that store's readers ambiguous.
+ * If another is ever proposed, this is the note to read first: the bar is that
+ * the state is genuinely per-async-scope *and* cannot live in an existing store
+ * without making that store's readers ambiguous.
  */
 export const kernelContext = new AsyncLocalStorage();
--- a/packages/gemi/kernel/context.ts
+++ b/packages/gemi/kernel/context.ts
@@ -7,7 +7,9 @@ import { AsyncLocalStorage } from "async_hooks";
  *
  * "The framework's only AsyncLocalStorage" is a claim that used to sit here. It
  * was already not true — the *kernel's* is the only one, but `http/`, `pubsub/`
- * and now the ORM each keep their own:
+ * and now the ORM each keep their own. Four in the shipped framework (a fifth
+ * sits in `rfc/workflow.ts`, an unreferenced design sketch that is outside
+ * `package.json`'s exports and that nothing imports):
  *
  *   kernel     kernel/context.ts                     the Application
  *   request    http/requestContext.ts                req, user, cookies, locale
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -1,6 +1,8 @@
+import type { SQL } from "bun";
 import { DatabaseManager } from "../database/DatabaseManager";
 import { app } from "../foundation/app";
 import { type BindContext, createBindContext } from "./compile/fragment";
+import { currentTransaction, withTransaction } from "./context";
 import {
   type NestedWriteStep,
   type RelationExecutor,
@@ -53,14 +55,45 @@ export abstract class Model {
     return schema;
   }
 
+  /**
+   * Runs `fn` inside a transaction. Every ORM query in it — at any call depth,
+   * through any number of services, including the nested reads an `include`
+   * fans out into — joins it automatically.
+   *
+   * The callback takes **no argument**, and that is the whole feature. Handing
+   * back a `tx` would put it in the signature of every function between here
+   * and the query, which is precisely the threading Prisma's `$transaction`
+   * forces and this exists to remove. If a raw query genuinely needs the
+   * handle, `currentTransaction()` hands it over without putting it in anyone's
+   * parameter list.
+   *
+   * Nesting is a savepoint: an inner failure rolls back to it and leaves the
+   * outer transaction usable, so a caller that catches keeps going.
+   *
+   *     await User.transaction(async () => {
+   *       const user = await User.create({ data: { email } })
+   *       await audit(user)          // its queries are in the transaction too
+   *     })
+   */
+  static transaction<T>(fn: () => Promise<T>): Promise<T> {
+    return withTransaction(app(DatabaseManager).sql, () => fn());
+  }
+
   static async $exec(op: Operation, args: any = {}): Promise<unknown> {
     const schema = this.$modelSchema();
 
     // Resolved per call, never captured at module scope: that is what keeps the
-    // connection swappable in tests, and it is the hook iteration 5 uses to
-    // pick up an ambient transaction instead of the pooled connection.
+    // connection swappable in tests.
     const db = app(DatabaseManager);
     const dialect = dialectFor(db.dialect);
+
+    // The ambient-transaction hook, and the entire integration. It is one line
+    // — and it covers every operation, every nested write and every relation
+    // read — only because invariant 1 held: `$exec` is the single door. An
+    // operation that acquired a private path to the database would show up
+    // here as a statement silently running outside the transaction, committed
+    // while its neighbours roll back.
+    const conn = currentTransaction() ?? db.sql;
 
     const plan = getOrCompile(schema, op, args, dialect);
 
@@ -69,18 +102,23 @@ export abstract class Model {
         registry
           .get<typeof Model>(model)
           .$exec(operation as Operation, relationArgs),
-      query: (text, values) => db.sql.unsafe(text, values),
+      // The one query with no model behind it — the implicit m-n join table —
+      // resolves its connection here rather than reaching for the pool, so it
+      // joins the transaction like everything else.
+      query: (text, values) => conn.unsafe(text, values),
     };
 
     // One context per call, holding the instant every `now()` and `@updatedAt`
     // on this operation shares, plus whatever the `before` steps resolve.
     const context = createBindContext();
 
-    // NOT ATOMIC until iteration 5. A write with nested `create`/`connect` runs
-    // more than one statement, and there is no transaction around them: if a
-    // later step fails, the earlier rows stay written. This is a recorded
-    // decision — iteration 5 introduces the ambient-transaction ALS here and
-    // makes all of it atomic without changing the call sites.
+    // A write with a nested `create` / `connect` runs more than one statement.
+    // Atomic exactly when the caller wrapped the call in `Model.transaction` —
+    // *not* implicitly. A write does not open its own transaction, because
+    // `$exec` cannot know whether it is one step of a larger unit; opening one
+    // per call would put a `BEGIN` around every query in the framework and turn
+    // a caller's transaction into a nest of savepoints it never asked for.
+    // Outside one, a failing later step still leaves the earlier rows written.
     await runSteps(plan.before, args, context, executor, []);
 
     // `unsafe` despite the name. Bun's tagged template cannot express a query
@@ -89,7 +127,7 @@ export abstract class Model {
     // identifiers only ever come from the generated schema, and every value is
     // a bound parameter. Do not "fix" this into a tagged template.
     const rows = await execute(
-      db,
+      conn,
       dialect,
       schema,
       op,
@@ -159,8 +197,8 @@ async function runSteps(
   rows: readonly Record<string, unknown>[],
 ): Promise<void> {
   if (steps === undefined) return;
-  // Sequentially, for the same reason relations load sequentially: iteration 5
-  // puts an ambient transaction on a single reserved connection, where
+  // Sequentially, for the same reason relations load sequentially: inside an
+  // ambient transaction every statement runs on one reserved connection, where
   // concurrent statements are not safe.
   for (const step of steps) {
     await step.run(args, context, executor, rows);
@@ -178,7 +216,7 @@ async function runSteps(
  * has never seen the database's names.
  */
 async function execute(
-  db: { sql: { unsafe(text: string, values: unknown[]): unknown } },
+  conn: Pick<SQL, "unsafe">,
   dialect: SqlDialect,
   schema: ModelSchema,
   op: Operation,
@@ -186,7 +224,7 @@ async function execute(
   values: unknown[],
 ): Promise<unknown> {
   try {
-    return await db.sql.unsafe(text, values);
+    return await conn.unsafe(text, values);
   } catch (error) {
     const violation = dialect.constraintViolation(error);
     if (!violation) throw error;
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -74,6 +74,41 @@ export abstract class Model {
    *       const user = await User.create({ data: { email } })
    *       await audit(user)          // its queries are in the transaction too
    *     })
+   *
+   * **KNOWN DIALECT ASYMMETRY — catching an error inside a transaction.**
+   * The sentence above holds for a *nested* `Model.transaction`. It does not
+   * hold for a bare statement: on Postgres any failed statement aborts the
+   * whole transaction block, so catching it and carrying on loses everything.
+   * SQLite does not care. Verified against Postgres 16 and SQLite:
+   *
+   *     await Model.transaction(async () => {
+   *       try { await User.create({ data: { email } }) }
+   *       catch (e) { if (!(e instanceof UniqueConstraintError)) throw e }
+   *       await Audit.create({ ... })   // fine on SQLite; on Postgres this
+   *     })                              // throws 'current transaction is
+   *                                     // aborted', and the whole transaction
+   *                                     // is lost
+   *
+   * That is a bug that passes in development on SQLite and takes out the
+   * transaction in production on Postgres, with an error naming neither the
+   * statement that failed nor the one that caused it.
+   *
+   * The fix is already here: wrap the fallible statement in a nested
+   * `Model.transaction`, which makes it a savepoint, and rolling back to that
+   * savepoint clears the aborted state. Works on both dialects.
+   *
+   *     await Model.transaction(async () => {
+   *       try {
+   *         await Model.transaction(() => User.create({ data: { email } }))
+   *       } catch (e) { if (!(e instanceof UniqueConstraintError)) throw e }
+   *       await Audit.create({ ... })   // fine on both
+   *     })
+   *
+   * **One statement at a time.** Every query in scope runs on one reserved
+   * connection, so `Promise.all` over several ORM calls inside the callback is
+   * not safe here — the ordinary, encouraged thing everywhere else in a Bun
+   * codebase. Await them in sequence. (`$exec` already runs its own nested
+   * writes and relation reads sequentially for this reason.)
    */
   static transaction<T>(fn: () => Promise<T>): Promise<T> {
     return withTransaction(app(DatabaseManager).sql, () => fn());
--- a/packages/gemi/orm/context.ts
+++ b/packages/gemi/orm/context.ts
@@ -5,13 +5,15 @@ import { AsyncLocalStorage } from "node:async_hooks";
  * The ambient transaction: the ORM's own `AsyncLocalStorage`, holding nothing
  * but the open handle and how deep the nesting is.
  *
- * This is the second `AsyncLocalStorage` in the framework, and the first one
- * that is not the kernel's. `packages/gemi/kernel/context.ts` carries the
- * Application and says so; the reasoning for adding a second rather than
- * widening that one is written up beside it, and the short version is that a
- * transaction scope has to *nest inside* an Application scope without replacing
- * it, and `foundation/app.ts` must keep reading exactly one shape out of the
- * kernel store or resolving a service becomes a transaction concern.
+ * One store per scope with one owner, which is the pattern the framework
+ * already follows — `kernel/context.ts` holds the Application,
+ * `http/requestContext.ts` the request and its user,
+ * `services/pubsub/BroadcastManager.ts` a socket's headers. The reasoning for
+ * not folding this into the kernel's is written up beside it, and the short
+ * version is that a transaction scope has to *nest inside* an Application scope
+ * without replacing it, and `foundation/app.ts` must keep reading exactly one
+ * shape out of the kernel store or resolving a service becomes a transaction
+ * concern.
  *
  * What makes an ALS the right home rather than a field somewhere: two concurrent
  * requests each in their own transaction must never see each other's handle.
--- a/packages/gemi/orm/context.ts
+++ b/packages/gemi/orm/context.ts
@@ -92,6 +92,18 @@ export function withTransaction<T>(
 ): Promise<T> {
   const current = ormContext.getStore();
 
+  // SINGLE-CONNECTION ASSUMPTION, pinned here so it is found rather than
+  // discovered. The savepoint branch ignores `pool` entirely, which is correct
+  // today: there is one `DatabaseManager` and one `SQL`, so the ambient handle
+  // and whatever pool the caller passed are necessarily the same database.
+  //
+  // It stops being correct the moment a second connection is configurable. A
+  // `DB.transaction` against connection B, called inside a `Model.transaction`
+  // on A, would open a savepoint on **A** and hand the callback A's handle —
+  // statements landing in the wrong database, with no error. That is the same
+  // failure shape as the "handle on the Application" alternative this file
+  // argues against. Multi-connection support must compare the two and either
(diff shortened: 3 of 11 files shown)


nstfkc (commented):
Reviewed `9ee27d4` in an isolated worktree. This is a small, well-argued PR and the integration really is one line — which is the strongest evidence yet that invariant 1 was worth holding for four iterations. Only one `.begin(` call site exists in the whole package (`orm/context.ts:102`), so there is no second path to verify.

| Verified at `9ee27d4` | |
|---|---|
| `vitest run` (packages/gemi) | **472 pass / 7 fail** — the 7 are the pre-existing router/i18n ones, matching the PR's table exactly |
| `vitest run app/models/` (template, SQLite) | **233 pass, 3 skipped**; the 1 failure is `generated.test.ts`, which needs `build:bin` my worktree doesn't have |
| `tsc --noEmit` | clean |
| `bun run lint` | 88 warnings on **both** this head and its base — no regression (the table's "2 warnings" is `oxlint orm/ facades/ kernel/`, not the package's own `lint` script) |

The ALS decision is right and, more importantly, the rejected alternatives are recorded where the next person will hit them — the note in `kernel/context.ts` keeping the "exactly one ALS" claim honest is the sort of thing that usually gets skipped. `context.test.ts` is the best test file in the ORM so far: fake handles make the interleaving deterministic, it asserts the *statements* landed on the right handle rather than just the reads, and the final assertion that the interleaving actually happened is what stops it from silently degrading into two sequential runs.

The SQLite-only-has-one-writer justification for making the two concurrency cases Postgres-only is correct, and saying so beats a test that passes because nothing overlapped.

**One finding worth acting on before merge**, and it is a documentation gap rather than a bug — details inline: catching an error inside a transaction and continuing works on SQLite and destroys the whole transaction on Postgres. I confirmed it against Postgres 16 in Docker through this PR's own `withTransaction`. Everything else is minor.

packages/gemi/orm/Model.ts
```diff
@@ -53,34 +55,70 @@ export abstract class Model {
...
+   * fans out into — joins it automatically.
+   *
+   * The callback takes **no argument**, and that is the whole feature. Handing
+   * back a `tx` would put it in the signature of every function between here
+   * and the query, which is precisely the threading Prisma's `$transaction`
+   * forces and this exists to remove. If a raw query genuinely needs the
+   * handle, `currentTransaction()` hands it over without putting it in anyone's
+   * parameter list.
+   *
+   * Nesting is a savepoint: an inner failure rolls back to it and leaves the
```
nstfkc, line 70: **The savepoint note is right, but the case beside it is the one that will bite — and it is an unnamed SQLite/Postgres asymmetry.**

"an inner failure rolls back to it and leaves the outer transaction usable, so a caller that catches keeps going" holds for a *nested* `Model.transaction`. It does not hold for a bare statement: on Postgres any failed statement aborts the whole transaction block, so catching and continuing loses everything.

Ran this through this PR's own `withTransaction` against Postgres 16 in Docker and against SQLite:

```
[postgres 16]
  bare statement caught + continue : THREW: current transaction is aborted,
                                    commands ignored until end of transaction block
    rows after: []                 <- the whole transaction lost
  wrapped in nested transaction    : continued fine
    rows after: [1,2]

[sqlite]
  bare statement caught + continue : continued fine
    rows after: [1,2]
  wrapped in nested transaction    : continued fine
    rows after: [1,2]
```

So the natural thing to write —

```ts
await Model.transaction(async () => {
  try { await User.create({ data: { email } }) }
  catch (e) { if (!(e instanceof UniqueConstraintError)) throw e }
  await Audit.create({ ... })            // silently fine on SQLite, fatal on Postgres
})
```

— works in development on SQLite and takes out the entire transaction in production on Postgres, with an error naming neither the statement that failed nor the one that caused it.

The good news is that this PR already ships the fix: wrapping the fallible statement in a nested `Model.transaction` makes it a savepoint and it works on both (verified above). Nothing tells the caller that, and this JSDoc is where they would look. Given `dialect/postgres.ts` names the `timestamp(3)` protocol divergence, `sqlite.ts` names the `like` case-sensitivity split and `write.ts` names the `on conflict` qualifier, this one deserves the same paragraph: *inside a transaction on Postgres, any error is terminal unless the failing statement was wrapped in a nested `Model.transaction`.*

packages/gemi/orm/context.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Anything reachable from the Application — which is one object shared by every
+ * in-flight request — fails that by construction, and fails it as silently
+ * written data, not as an error. "two concurrent transactions never see each
+ * other's handle" in `context.test.ts` is the test that would catch it.
+ */
+export interface TransactionScope {
+  /**
+   * Bun's transaction handle. Every ORM statement in scope runs on it, which
+   * means they all run on one reserved connection — so nothing here may issue
+   * two statements concurrently. `Model.$exec` already runs its nested writes
```
nstfkc, line 29: "nothing here may issue two statements concurrently" is the single most important constraint the feature places on a caller, and it is stated on an internal interface's field rather than on the public API.

An application author writing

```ts
await Model.transaction(async () => {
  await Promise.all([User.create(...), Account.create(...)])
})
```

is doing the ordinary, encouraged thing everywhere else in a Bun codebase, and here it puts two statements on one reserved connection. `Model.transaction`'s JSDoc says nothing about it — worth one line there, since that is the doc that appears on hover at every call site.

(No suggestion that it be *enforced*; a note is the proportionate fix.)

packages/gemi/orm/context.ts
```diff
@@ -0,0 +1,105 @@
...
+ * is Bun's behaviour and is left alone; it is only surprising if unmentioned.
+ */
+export function withTransaction<T>(
+  pool: SQL,
+  fn: (tx: TransactionSQL) => Promise<T>,
+): Promise<T> {
+  const current = ormContext.getStore();
+
+  if (current) {
+    return current.tx.savepoint((sp) => {
```
nstfkc, line 108: Minor / forward-looking: the savepoint branch ignores `pool` entirely, which is exactly right today — one `DatabaseManager`, one `SQL`, so the ambient handle and the pool a caller passed can only be the same database.

It stops being right the moment a second connection is configurable. A `DB.transaction` against connection B, called inside a `Model.transaction` on A, would silently open a savepoint on **A** and hand the callback A's handle — statements landing in the wrong database, no error. That is the same failure shape as the rejected "handle on the Application" alternative, which this file argues against so carefully.

One line noting the single-connection assumption would pin it, so whoever adds multi-connection support finds the constraint here rather than discovering it from a misplaced write.

nstfkc (commented):
Checked `5052ed8`. The delta since `9ee27d4` is only the merge of #49's `a2f5155` — no changes of this PR's own — so this is a verification pass rather than a re-review.

**The merge is clean and the merged head is green**, which was worth confirming: #49 changed `render`'s signature and touched `plan-relations.ts`, both of which this branch's transaction suite runs through.

| Verified at `5052ed8` | |
|---|---|
| `vitest run orm/` (packages/gemi) | **400 pass** |
| `vitest run app/models/` (template, SQLite) | **234 pass, 3 skipped**, all 7 files |

The three points from my review of `9ee27d4` are still open, since nothing in the merge touched them:

1. **The Postgres aborted-transaction asymmetry** — catching an error from a bare statement inside `Model.transaction` and continuing works on SQLite and destroys the whole transaction on Postgres. Confirmed against Postgres 16 in Docker through this PR's own `withTransaction`. The remedy already ships here (wrap the fallible call in a nested `Model.transaction`); it just isn't documented where a caller would look.
2. The "no concurrent statements in a transaction" constraint is documented on `TransactionScope.tx` rather than on `Model.transaction`'s JSDoc.
3. `withTransaction`'s savepoint branch ignores `pool` — correct with one `DatabaseManager`, silently wrong if multi-connection support ever lands.

Only the first is worth acting on before merge, and it is a paragraph rather than a code change.

packages/gemi/kernel/context.ts
```diff
@@ -1,3 +1,44 @@
...
+/**
+ * The kernel's AsyncLocalStorage. It holds the Application, and nothing else —
+ * `Kernel.run()` enters it once per request, websocket message or cron tick,
+ * and `foundation/app.ts` reads it to resolve services.
+ *
+ * "The framework's only AsyncLocalStorage" is a claim that used to sit here. It
+ * was already not true — the *kernel's* is the only one, but `http/`, `pubsub/`
+ * and now the ORM each keep their own:
+ *
+ *   kernel     kernel/context.ts                     the Application
```
nstfkc, line 14: The inventory is itself one short. `git grep -n "new AsyncLocalStorage" -- packages/gemi` at this commit returns **five**, not four:

```
http/requestContext.ts:36
kernel/context.ts:44
orm/context.ts:37
rfc/workflow.ts:22          <- not in the table
services/pubsub/BroadcastManager.ts:12
```

The fifth is defensible to leave out of the *pattern* discussion — `rfc/` is a directory of design sketches, nothing outside it imports anything in it (verified), it is absent from `package.json`'s exports, and `workflow.ts` is visibly a stub down to a bare `WorkflowContext;` expression statement. It is not part of the shipped framework.

But the sentence says "There are four in source," and there are five in source. Given this commit exists precisely because an inherited claim was stated more strongly than the code supported, the same standard applies here — the fix is a parenthetical:

> four in the shipped framework (a fifth sits in `rfc/workflow.ts`, an unreferenced sketch)

That also saves the next person the ten seconds of thinking they found a fifth store the note missed.

nstfkc (commented):
Checked `1cf1d05`. Comment-only — I diffed with `-U0` and filtered out comment lines, and nothing remains, so no re-run of the suites was needed.

Good catch to make, and the right way to make it: correcting an inherited claim rather than working around it, and reframing the ORM's store as a fourth instance of an existing pattern instead of an exception to a rule that never held. "A comment that overstates an invariant is worse than no comment" is the correct principle, and `Auth.user()` reading a store that was never counted is a concrete demonstration rather than a hypothetical.

I verified the inventory's descriptions against source and they are accurate — `http/requestContext.ts`'s `Store` does carry `req`, `user`, `cookies`, `headers` and `locale`; `BroadcastManager`'s carries headers and cookies.

One correction to the correction, inline. It is half a line.

nstfkc:
Two corrections, both pushed.

**1. `orm/` was never in the tsconfig `include` list**, so my "`tsc --noEmit` clean" claim on this PR covered only the seven ORM files reachable from an included directory. Fixed on #49 (`9ea8bfc`) and merged forward here; it surfaced three real `Binder` arity errors from iteration 4. Details on that PR.

**2. The ALS inventory in `kernel/context.ts` was wrong, and I wrote it.** The comment this PR added says there was exactly one AsyncLocalStorage until iteration 5 and there are now two. Both halves are false, and the claim it inherited was already false before this branch. There are four in source:

```
kernel     kernel/context.ts                     the Application
request    http/requestContext.ts                req, user, cookies, locale
broadcast  services/pubsub/BroadcastManager.ts   headers, cookies
orm        orm/context.ts                        an open transaction
```

Found while mapping how the authenticated user is reached for iteration 6 — `Auth.user()` reads the request store, which is one of the two that were never counted.

Rewritten in `1cf1d05` to say what is true, keeping the part that matters: the *kernel's* store holds the Application and nothing else, because `foundation/app.ts` reads it unconditionally. The framework's actual pattern is one store per scope with one owner, read through its own accessor — `app()`, `RequestContext.getStore()`, `currentTransaction()` — which makes the ORM's a fourth instance of an existing pattern rather than a deviation from a single-store rule that never held. The reasoning for not folding it into the kernel's is unchanged and stands on its own.

A comment that overstates an invariant is worse than none: someone relying on "there is only one ALS" would have been wrong about `Auth.user()` today, not at some future point.

nstfkc (commented):
Reviewed `8ac08bf` (the merge of #49 plus `9ea8bfc`). This is the most valuable commit in the stack, and I should say first that it invalidates a line I put in three separate reviews: I reported "`tsc --noEmit` clean" on #49 twice and on #50 once, and for `orm/` that proved nothing. Correcting it in the commit rather than letting it stand is the right call.

**The gap was worse than the commit message says.** It reports seven ORM files reachable from an included directory. Measured with `tsc --listFiles` against the pre-fix config on this branch, it is **one** — `orm/context.ts`, reached because `facades/DB.ts` imports `withTransaction` from it, and that file imports nothing but `bun` types and `node:async_hooks`, so the walk stops there. It is the only import of `orm/` from any included directory.

And that single import is **this branch's own addition**. At `a2f5155` — merged #49 — `facades/DB.ts` has no `orm` import at all, so the type-check reached **zero** ORM files. Iterations 1 through 4 were never type-checked, not partially type-checked.

The three errors are real and the fix is the right one: forwarding the context rather than widening the signature back. The note that all three sit on paths a `defaultBinder` would have thrown on is the part worth keeping — `inList` is on every `in` list and every batched relation read.

| Verified at `8ac08bf` | |
|---|---|
| `tsc --noEmit -p tsconfig.json` | clean — and now meaningful |
| `vitest run orm/` | 400 pass |

**Where this commit lives, and why that matters.** `9ea8bfc` is on `feat/orm-04-writes` — #50's *base* — not in #50's diff, which is still the same nine files. #49 is merged, so this commit is currently in no open PR at all: a reviewer reading either PR's diff would never see it. Worth either folding it into this branch or opening a small PR for it, because it is the one commit in the stack that changes what every previous verification claim meant.

**One finding: the same gap is still open one directory over, on code this stack changed.**

The fix is right, but `bin/orm/` is still outside the tsconfig's `include` — and that is the ORM's **generator**, which #49 modified (`bin/orm/emit.ts`, the write-operation emitters). Only `bin/migrate` is listed, so the same gap this commit closes for `orm/` remains open on the code that produces every model class.

Probed it by adding `bin/orm` to `include`:

```
bin/orm/emit.ts(87,19): error TS2339: Property 'name' does not exist on type
  'ReadonlyDeep<{ name: string; args: (string | number)[] }> | readonly ReadonlyDeep<FieldDefaultScalar>[]'
bin/orm/emit.ts(93,30): error TS2339: (same)
```

Calibrating honestly: these two are **not** the same severity as the three you just fixed. `defaultSpec` already guards with `typeof value === "object" && !Array.isArray(value)`, so the runtime logic is correct — TypeScript simply does not narrow a `readonly T[]` out of a union via `Array.isArray`. They need a cast or a `readonly`-aware guard, not a behaviour change. But they are exactly the kind of thing the include list is supposed to surface, and `emit.ts` is where a silent type error turns into a wrong *generated artifact*.

I also probed the other unlisted server directories — `auth`, `config`, `internal`, `runtime` — and all four are clean, so `bin/orm` is the only remaining gap. Adding it closes the class completely for server code, at the cost of those two annotations.

(`client` is covered separately by `tsconfig.browser.json`; `rfc`, `scripts`, `bun`, `vite`, `ide` are sketches and build glue, reasonable to leave out.)

nstfkc (commented):
Reviewed `c48c549`. Every point from the last two rounds is addressed, and the abort asymmetry got the treatment I hoped for — documented **and** pinned by a test, on both dialects rather than skipping the inconvenient one.

**I ran the Postgres half against a real server**, since that is the part CI cannot reach and the assertions are dialect-branched — so a green SQLite run proves only half of each test. Mirrored both new cases through this branch's own `withTransaction` against Postgres 16 in Docker and against SQLite:

```
[postgres]
  test 1  first = true    userCount = 0     suite asserts: first === true,  userCount === 0   ✓
  test 2  ok              userCount = 1, accounts = 1   suite asserts: 1, 1                    ✓

[sqlite]
  test 1  first = false   userCount = 1     suite asserts: first === false, userCount === 1   ✓
  test 2  ok              userCount = 1, accounts = 1   suite asserts: 1, 1                    ✓
```

Every assertion matches real-server behaviour, including the non-obvious one: `first === true` on Postgres means the transaction *resolves* rather than rejecting, because `COMMIT` on an aborted block returns `ROLLBACK` without raising. Asserting `true` rather than "not false" is what stops that from passing for the wrong reason, and it is correct.

The rest, all confirmed at `c48c549`:

- **`render`'s `origin` is now required** (`1101012`, on the base branch), with the reasoning written into the parameter's own doc. That is the loophole closed properly rather than by convention.
- **One-statement-at-a-time** moved onto `Model.transaction`'s JSDoc, where it appears on hover at every call site.
- **The single-connection assumption** pinned in `withTransaction`'s savepoint branch, naming the exact failure — savepoint on A, statements to B, no error.
- **The ALS inventory** now says four in the shipped framework with the `rfc/workflow.ts` sketch called out as the fifth.

| Verified at `c48c549` | |
|---|---|
| `vitest run orm/` | 400 pass |
| `vitest run app/models/` (SQLite) | 236 pass, 3 skipped, 7 files |
| `tsc --noEmit` | clean — and meaningful now that `orm` is in `include` |

One item from the last round is still open, and it is the only thing I would still do: **`bin/orm` is not in the tsconfig `include`** (checked at this commit — the list ends at `bin/migrate`). It hides two type-level errors in `emit.ts:87,93`, which are cosmetic rather than latent bugs, but `emit.ts` is where an unchecked type error becomes a wrong *generated artifact*. `auth`, `config`, `internal` and `runtime` are all clean, so that one line closes the class for server code.

Nothing blocking. This is ready as far as I can tell.

nstfkc:
All three addressed in `31c0dc9`, and the first one got a test rather than only a paragraph.

**The Postgres aborted-transaction asymmetry.** This is the best finding on the stack so far — it is a bug that *passes in development and fails in production*, which is the exact failure class this project keeps going out of its way to catch, and I shipped the misleading sentence myself. Documented on `Model.transaction`'s JSDoc, which is what appears on hover at the call site, with both the broken form and the nested-transaction remedy spelled out.

But a documented divergence with no test is how the three Postgres bugs in #49 got in, so it is now pinned by two cases in the transaction suite — both running on **both** dialects rather than skipping the inconvenient one. The assertion is that the two genuinely differ, so a test that only ran where the answer is comfortable would not be making it.

I re-derived your result independently before writing them, through this branch's own `withTransaction` against Postgres 16:

```
caught: duplicate key value violates unique constraint "p_e_key"
next statement: current transaction is aborted, commands ignored until end of transact
outcome: aborted
rows: []
```

So the Postgres assertion is `toBe(true)` plus `userCount() === 0`, not `not.toBe(false)` — otherwise it would pass if the transaction failed for any other reason.

**"One statement at a time" moved to the public JSDoc.** Correct that it was on the wrong doc. `Promise.all` over ORM calls inside the callback is the ordinary, encouraged thing everywhere else in a Bun codebase, so the note belongs where the call site sees it.

**The single-connection assumption** is now pinned in `withTransaction`. Your framing is what sold it: it is the same failure shape as the "handle on the Application" alternative this file argues against at length — statements landing somewhere they were not addressed to, with no error.

**The ALS inventory was one short.** Five in source; the fifth is `rfc/workflow.ts`. You're right that the standard applies to itself — this note exists *because* an inherited claim was stated more strongly than the code supported, so "there are four in source" was the same mistake in the same paragraph. Now reads "four in the shipped framework (a fifth sits in `rfc/workflow.ts`, an unreferenced design sketch)".

| | |
|---|---|
| `vitest run orm/` | **400 pass** |
| template, SQLite | **236 pass, 3 skipped** |
| template, Postgres 16 | **271 pass** — only the two documented one-dialect-at-a-time differential files fail |

On your `bun run lint` note: agreed, and thank you for checking it against the base. The 88 warnings are pre-existing and unrelated; my table's "2 warnings" was the narrower `oxlint orm/ facades/ kernel/`, which I should have labelled as such.

nstfkc:
Done in `41c7234`, and I went wider than the ask.

**First, your correction to my correction is right and I'll take it plainly: it was zero, not seven.** I verified it — at `a2f5155`, `facades/DB.ts` has no `orm/` import at all and `orm` is not in the include list, so iterations 1 through 4 were never type-checked. My "seven" came from measuring on a tree that already had *this* branch's `DB.ts` change merged in, which is measuring after the fact on the wrong tree. That's the same mistake as the original claim, one level up.

**`bin/orm` added, plus the two `emit.ts` errors fixed.** Your severity calibration is right — the runtime guard was already correct, TS just won't narrow a `readonly T[]` out of a union through `Array.isArray`. Replaced with a real predicate, which incidentally closes something the inline form didn't: `typeof null === "object"`, so a null default would have been dereferenced rather than falling to the literal branch. Not reachable from any schema I can construct, but the guard is now actually a guard.

Your framing is what made me go further — `emit.ts` is where an unchecked type error becomes a wrong *generated artifact*, which is worse than a wrong query because every model class in every application is downstream.

**So I audited the whole include list against the export map**, since "which directories are checked" was evidently something nobody had looked at:

```
auth       NOT listed  ->  added, clean
config     NOT listed  ->  added, clean   <- a public export: ./config
internal   NOT listed  ->  added, clean
runtime    NOT listed  ->  added, clean
```

Matching your finding exactly — all four clean. But it means `orm` was not a one-off omission: **six of the package's directories were outside the type-check, two of them public exports.** Adding all of them is one line each and closes the class for server code, rather than patching whichever directory happened to be under review.

Worth noting consumers were never affected — the export map points at source `.ts`, so an application type-checks these files itself. What the gap cost was the repo's own `tsc --noEmit` and the declarations `build:types` emits.

**On where `9ea8bfc` lives:** #49 merged at 20:56, and that commit was on its head branch, so it went in with the merge — it isn't stranded. But you're right that it's invisible in any open diff. Note the same now applies to `1101012` (the `render` origin fix), which I pushed to `feat/orm-04-writes` a few minutes *after* #49 merged. Both are in #50's base rather than its diff, so getting them past `feat/orm-04-writes` needs a follow-up PR at that level once #50 lands. Flagging it rather than restructuring the stack mid-review.

Thank you for running the Postgres half of the abort tests against a real server — that's the part CI can't reach, and the assertions being dialect-branched means a green SQLite run genuinely only proves half of each. Your note on *why* `first === true` (the transaction resolves because `COMMIT` on an aborted block returns `ROLLBACK` without raising) is a better explanation than the one I put in the test comment.

| Verified at `41c7234` | |
|---|---|
| `tsc --noEmit` | clean, with all six directories added |
| `build:types`, `build:bin` | both succeed |
| `vitest run` (packages/gemi) | 480 pass / same 7 pre-existing |
| template, SQLite | 236 pass, 3 skipped |

packages/gemi/tsconfig.json
```diff
@@ -22,7 +22,12 @@
     "support",
     "database",
     "orm",
-    "bin/migrate"
+    "bin/migrate",
+    "bin/orm",
```
nstfkc, line 48: The audit is one directory short of its own criterion. `bun/` is dismissed as "sketches and build glue" alongside `rfc`, `scripts`, `vite` and `ide` — but unlike those four it holds **two public exports**:

```
"./bun/preload": "./bun/preload.ts",
"./bun/plugin":  "./bun/plugin.ts",
```

That is the same argument that justified adding `config`. An application doing `import "gemi/bun/preload"` gets source that the repo's own `tsc` has never looked at.

Probed what it would cost — and unlike the other five, it is not a one-liner:

```
bun/customRequestParser.ts(170,26): error TS2307: Cannot find module
  'recast/parsers/typescript' or its corresponding type declarations.
bun/preload.ts(13,16): error TS1309: The current file is a CommonJS module
  and cannot use 'await' at the top level.
```

The first is a genuinely missing type declaration. The second is not a code bug at all — it says the inherited `module` setting treats these as CommonJS, which is wrong for files Bun loads as ESM, so including `bun/` needs a module override (or a nested tsconfig) rather than an edit to the file.

So I would not fold it into this commit. But "build glue" is doing work here that the export map does not support, and the honest version is a line saying `bun/` is knowingly excluded *despite* exporting publicly, with the module mismatch named — otherwise the next person to audit this list reaches the same wrong conclusion from the same comment.

nstfkc (commented):
Reviewed `41c7234`. This goes past what I reported — I flagged `bin/orm`; you audited the whole include list and found six directories outside the type-check. That is the right response to "the guard was in the wrong place" twice in a row.

I checked the claims that carry risk, since a tsconfig `include` change is exactly the sort of thing that breaks a build quietly:

| Verified at `41c7234` | |
|---|---|
| `tsc --noEmit` | clean with all six added |
| `bun run build:types` | succeeds; declarations now emitted for all six (`dist/orm` 23, `dist/auth` 12, `dist/internal` 3, `dist/config` 2, `dist/bin/orm` 1, `dist/runtime` 1) |
| `bun run build:bin` | succeeds, both entry points |
| `vitest run` (packages/gemi) | 480 pass / 7 fail — the same pre-existing router and i18n files |
| `vitest run app/models/` | 236 pass, 3 skipped |

"Two of them public exports" checks out against the export map: `./orm` and `./config` both point at source in directories that were unchecked.

`isFunctionDefault` is a better fix than the cast I expected. The `typeof null === "object"` hole it closes is real even if no schema reaches it, and writing it as a predicate rather than an inline guard is what makes the narrowing survive the next edit. One behavioural difference worth having noticed rather than discovered: an object default *without* a string `name` used to reach `switch (undefined)` and come out as `{ kind: "dbgenerated" }`, and now falls through to `{ kind: "value" }`. DMMF's `FieldDefault` always carries `name`, so it is unreachable — but it is a change, not a pure narrowing.

One last gap, by the commit's own standard. Details inline.

nstfkc:
Both in `db1b85f`.

**The behaviour change you spotted is reverted, not just documented.** You're right that it wasn't a pure narrowing, and on reflection the old branch is the one to keep: `dbgenerated` omits the column and lets the database supply it, while `value` would try to bind the object itself as the column's value. Unreachable either way given DMMF always carries `name` — but the safer of two unreachable branches is still the one to pick, and I'd rather that reasoning sit in the comment than in a diff nobody reads again. The predicate is now exactly the old condition plus the `null` check.

**The `bun/` gap is real and the dismissal was wrong.** `./bun/preload` and `./bun/plugin` are public exports, so "build glue" was doing work the export map doesn't support — the same argument that justified adding `config`. The conclusion happens to stand but the reason I gave for it didn't.

Agreed on not folding it in. The two errors are different in kind:

```
bun/customRequestParser.ts  TS2307  no types for 'recast/parsers/typescript'
bun/preload.ts              TS1309  top-level await under the inherited CommonJS `module`
```

The second isn't a code bug — those files are loaded by Bun as ESM, so the inherited `module` setting is simply wrong for them. That wants a nested tsconfig with its own `module`, not an edit to the source.

So it's recorded instead: a comment above `include` naming the six directories that were missing, why `client` lives in `tsconfig.browser.json`, and that `bun/` is knowingly excluded **despite exporting publicly** with both blockers named. The point being the one you made — otherwise the next audit reaches "build glue" and stops where I did.

`tsc --noEmit`, `build:types` and `build:bin` all clean; 400 orm tests.

Thank you for checking the declaration output per directory — that's the half of a tsconfig `include` change that fails quietly, and I hadn't enumerated it.

nstfkc (commented):
Reviewed `db1b85f`. Both points addressed, and the first one better than I framed it.

I flagged the `isFunctionDefault` behaviour change as a curiosity — "unreachable, but it is a change, not pure narrowing." You went further and asked which of the two unreachable branches is *safer*, which is the question I should have asked: `dbgenerated` omits the column and lets the database supply it, while `value` would try to bind the object itself as a column value. Reverting to the old condition plus the `null` check keeps the safe branch and still closes the `typeof null === "object"` hole. Right call.

On `bun/`: recording the exclusion with both errors named is exactly the outcome I was after — the point was never that it must be included, only that "build glue" was doing work the export map does not support. The comment block above `include` now says what was missing, which two were public exports, why `client` lives elsewhere, and why `bun/` needs a nested tsconfig rather than a source edit. The next audit of that list starts from the truth.

| Verified at `db1b85f` | |
|---|---|
| `tsc --noEmit` | clean |
| `bun run build:types` | clean |
| `bun run build:bin` | clean, both entry points |
| `vitest run orm/` | 400 pass |

One thing I checked because the comments make `tsconfig.json` JSONC rather than JSON: nothing in the repo parses it as strict JSON — no build script, no vite config, no turbo task reads it — and both `tsc` and Bun accept comments. So this is safe. (Worth knowing if a future script ever reaches for `JSON.parse` on it.)

PR #51, opened 2026-07-27
ORM iteration 6 — policies that apply to nested reads

nstfkc:
Per-model policies that can deny an operation, scope its `where`, or redact fields — applied to **every** query, including nested relation reads.

```ts
export class Account extends AccountModel {
  static $policy = {
    scope: (ctx) => ({ organizationId: ctx.user.organizationId }),
    onCreate: (ctx, data) => ({ ...data, organizationId: ctx.user.organizationId }),
  }
}

// Accounts from other organisations do not come back here, and nothing
// was written at the include site to make that true:
await User.findMany({ include: { accounts: true } })
```

The second of the two features that justify the project. Prisma's query extensions do not fire on nested reads, so a scope written for `Account` is bypassed by that exact call. Ours is not — and **not because it takes care to cover that case**: a relation read *is* `$exec` on the related model's own class, so it goes through the same hook as a root query and there is no second path to keep in step. The headline test is as much a proof that invariant 1 held for five iterations as it is a test of policies.

Stacked on #50. Review that first.

## The two gating decisions, settled before any code

The plan's deliverable 1 says to settle these first because the hook signature depends on them.

**The user is read synchronously off the request store.** `Auth.user()` is the fuller answer — it falls back to the cookie and resolves a session from the database — but it is `async`, and this hook runs once per query *per node of an include tree*. An awaited user there is a round trip per node on every request, for a value already sitting in the store on any route that authenticated. The consequence is stated in the code rather than left to be discovered: a route without the `auth` middleware sees no user even when a cookie is present. That is the correct reading — the route did not ask to authenticate — but it makes middleware config load-bearing for data access.

**Deny by default.** A policy that needs a user and does not have one raises. The alternative treats "no user" as "no policy", which makes the dangerous case the silent one: a request whose auth middleware was misconfigured would read every tenant's rows rather than failing.

`Model.asSystem(fn)` is the opt-out. `Model.asUser(user, fn)` is its narrow counterpart, and it is **beyond what the plan asked for** — added because `asSystem` alone is the wrong tool for the common case. A queue worker handling "send the invoice for organisation 7" acts *as somebody*; giving it only the blunt instrument leaves it suspending policies and hand-scoping every query, which is precisely the unscoped-by-accident outcome deny-by-default exists to prevent. The narrow tool should be the easy one. It also makes policies testable without reaching into `RequestContext`'s internals.

## Design, with the reasoning in the code

- **Args in, args out.** A policy never sees SQL. That is what makes a scope a two-line `AND` instead of string surgery, what keeps the relation strategies interchangeable — iteration 7's lateral joins put the scoped `where` inside the subquery with no policy change — and what keeps the plan cache sound.
- **Policies run before the plan key is computed.** They change the SQL, so a plan keyed on pre-policy args lets two differently-scoped users share one plan. That is a cross-tenant data leak wearing a caching bug's clothes: it returns wrong rows rather than failing. Both halves are tested — same shape means one plan and no leak, differently-*shaped* scopes mean different keys.
- **Base-first dispatch, narrowing only.** A subclass's scope is `AND`ed after its base's, so a base can only ever be narrowed further. A derived policy that could drop its base's would make a shared tenant guard unenforceable, which is the opposite of why a base would carry one.
- **`onCreate` is separate from `scope`**, because the read semantics do not generalise: an insert has no `where`, and "restrict which rows are affected" is meaningless for a row that does not exist yet.
- **`asSystem` shares the ORM's existing scope object** rather than minting a fifth `AsyncLocalStorage`. The inventory note in `kernel/context.ts` sets the bar; two booleans of ORM state do not clear it.
- **Redaction runs in shaping**, on the result — so the column is still selected and the plan key stays user-independent. Dropping it from the `select` instead would change the SQL per user.

## Two operations refused rather than run unscoped

Both found by writing the tests, not by inspection:

- **`upsert` under a scoped policy.** Its `where` becomes an `on conflict` target, which takes a key and not a predicate, so there is nowhere for `organizationId: 7` to go.
- **`create` under a policy with `scope` and no `onCreate`.** The natural tenant policy has both and is unaffected. `scope` alone is the dangerous shape: reads confined to the caller, inserts free to name any tenant — a policy that looks complete and is half of one.

## Soft deletes, as the proof case

`softDeletes()` is thirty lines against the same `ModelPolicy` an application writes, with no privileged access to anything. A soft-deleted row disappears from nested `include` reads too — the half ORM-level soft deletes usually miss.

The rewrite half (`softDelete`, `softDeleteMany`) is a call-site helper rather than part of the policy, because a policy rewrites *arguments* and this changes the *operation*, and `$exec` dispatches on the operation before consulting a policy. Widening `ModelPolicy` to let a policy swap the operation would make every `$exec` call's meaning depend on a hook — a far larger claim than scoping a `where`.

Writing it as an application would exposed two real bugs, which is the return on shipping it rather than documenting it:

**Deny-by-default was denying policies that never ask who you are.** `softDeletes()` scopes on `deletedAt: null` and never touches the user; it is a data-integrity rule, not an authorization one, and it has to work in a cron tick. The up-front "this model has a policy, so it needs a user" check made that impossible.

Deny-by-default now lives on the context's `user` accessor: reading it with no user raises, so a policy that never reads it is unaffected. Raising on *access* rather than returning null is the safe direction — a scope built from a null user is `{ organizationId: undefined }`, and an undefined value is an **absent filter**, so the scope would silently vanish and the read would be unscoped. The error is the loud version of the leak. `context.hasUser` is there for a policy that wants to serve anonymous access differently rather than refuse it.

**A scoped policy broke every `update` and `delete`.** `withScope` wrapped the caller's where as `{ AND: [caller, scope] }` — correct SQL, and it buries `{ id: 1 }` one level down where `matchUniqueKey` cannot see it. Every scoped `update`/`delete` failed with "needs a unique field".

The scope is now an extra member of the where's own `AND`, beside the caller's keys rather than around them: `{ ...caller, AND: [...theirs, scope] }`. That satisfies both requirements at once — the caller's top-level keys stay where `matchUniqueKey` reads them, and the scope still conjoins rather than replacing a filter on the same field.

**And an earlier test of mine was passing for the wrong reason because of it.** "a scoped delete of another tenant's row finds nothing" asserted a bare `rejects.toThrow()`, which the unique-field error satisfied just as well as the row genuinely not matching. It now asserts `RecordNotFoundError`, and two positive cases were added — a scoped delete and update of the caller's *own* row succeeding — because that is the half proving the refusal was about the scope and not about the key being unreachable.

## The differential harness

Every comparison now runs through `Model.asSystem`, so it asserts against the **pre-redaction** payload, as the plan requires. Teaching it which fields to ignore instead would also have masked a redaction that removed the *wrong* field, which is the bug worth catching. The template declares no policies today, but the moment an application adds one every comparison would otherwise start failing for a reason unrelated to the compiler.

## Tests

`packages/gemi/orm/policy.test.ts` — 48 cases with no database and no request, which is the design working: dispatch, ordering and rewriting are all pure functions of (policies, context, args).

`templates/saas-starter/app/models/policies.test.ts` — the database-dependent half on both dialects: the nested-include proof both ways round, a deny on the related model denying the whole include, base-class narrowing, both plan-cache cases, write scoping, soft deletes including through an include, and the interaction with iteration 5's transactions.

One test bug worth recording, because it is the kind that survives review: the seed hard-coded ids 1 and 2, and SQLite's AUTOINCREMENT high-water mark survives a `DELETE` — so the headline test passed in isolation and returned zero accounts in a full run, with the accounts pointing at a `userId` no longer present. It would have passed on Postgres either way, where the harness restarts identity. The ids are captured from the seed now.

## Verification

| | |
| --- | --- |
| `vitest run` (packages/gemi) | **531 pass**, 7 fail |
| baseline | same 7 — pre-existing router/i18n |
| template, SQLite | **267 pass**, 4 skipped |
| template, Postgres 16 | **333 pass** |
| `tsc --noEmit` | clean |

On Postgres the only failing files are the two differential suites that cannot run with a Postgres-generated client — documented behaviour in `differential.ts`.

## Acceptance criteria

All nine. 1–3 deny/scope/redact and the nested-include proof and base-class ordering; 4 both plan-cache cases; 5 write coverage with per-operation semantics documented in `policy.ts`; 6 redaction in shaping plus the pre-redaction differential; 7 `asSystem`/`asUser` as explicit call-site opt-ins; 8 soft deletes shipped with tests; 9 lint and tests pass.

## Not in scope

Field-level *write* permissions, database-delegated row-level security, policy memoisation (iteration 7 measures before optimising), and a policy DSL — policies stay plain functions, per the plan.

## For the reviewer

- Postgres runs need `--no-file-parallelism`; the new suites truncate the shared database.
- The one thing I would look at hardest is `withScope`. Both shapes are valid SQL and only one keeps `matchUniqueKey` working, so it is the kind of place where a future refactor could reintroduce the bug without any test outside this PR noticing.

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -208,21 +208,40 @@ export abstract class Model {
     let effective = args;
 
     if (policies.length > 0) {
-      // The registry must resolve this model's name to *this* class, or nested
-      // relation reads — which go through the registry — would run on a
-      // different class and skip these policies entirely. Scoped at the root,
-      // unscoped inside an include; see `UnregisteredPolicyClassError`. Checked
-      // here rather than at registration because a class can acquire a policy
-      // at any point after it is defined.
+      // A nested relation read resolves its target through the registry, so it
+      // runs whatever is registered under this model's name — not necessarily
+      // the class the caller queried. If the two disagree about *policies*, the
+      // same query is scoped at the root and unscoped inside an include. See
+      // `UnregisteredPolicyClassError`.
+      //
+      // The comparison is on the resolved policy **chain**, not on class
+      // identity, and the difference matters: a plain subclass of a registered,
+      // policied class carries the same policy objects in the same order by
+      // inheritance, so a nested read resolving to the parent applies exactly
+      // what the root query applied. There is no divergence and nothing to
+      // refuse. Checking identity instead rejected `class AdminUser extends User
+      // {}` — a typed view that adds nothing — with an error about policies it
+      // did not write.
+      //
+      // Checked here rather than at registration because a class can acquire a
+      // policy at any point after it is defined.
       const registered = registry.has(schema.name)
         ? registry.get<unknown>(schema.name)
         : undefined;
+
       if (registered !== undefined && registered !== this) {
-        throw new UnregisteredPolicyClassError(
-          schema.name,
-          (registered as { name?: string }).name ?? String(registered),
-          this.name,
-        );
+        const theirs = policiesFor(registered);
+        const diverges =
+          policies.length !== theirs.length ||
+          policies.some((policy, index) => policy !== theirs[index]);
+
+        if (diverges) {
+          throw new UnregisteredPolicyClassError(
+            schema.name,
+            (registered as { name?: string }).name ?? String(registered),
+            this.name,
+          );
+        }
       }
 
       // Deny-by-default lives on the context's `user` accessor, not here: a
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -105,6 +105,41 @@ export abstract class Model {
    *       const user = await User.create({ data: { email } })
    *       await audit(user)          // its queries are in the transaction too
    *     })
+   *
+   * **KNOWN DIALECT ASYMMETRY — catching an error inside a transaction.**
+   * The sentence above holds for a *nested* `Model.transaction`. It does not
+   * hold for a bare statement: on Postgres any failed statement aborts the
+   * whole transaction block, so catching it and carrying on loses everything.
+   * SQLite does not care. Verified against Postgres 16 and SQLite:
+   *
+   *     await Model.transaction(async () => {
+   *       try { await User.create({ data: { email } }) }
+   *       catch (e) { if (!(e instanceof UniqueConstraintError)) throw e }
+   *       await Audit.create({ ... })   // fine on SQLite; on Postgres this
+   *     })                              // throws 'current transaction is
+   *                                     // aborted', and the whole transaction
+   *                                     // is lost
+   *
+   * That is a bug that passes in development on SQLite and takes out the
+   * transaction in production on Postgres, with an error naming neither the
+   * statement that failed nor the one that caused it.
+   *
+   * The fix is already here: wrap the fallible statement in a nested
+   * `Model.transaction`, which makes it a savepoint, and rolling back to that
+   * savepoint clears the aborted state. Works on both dialects.
+   *
+   *     await Model.transaction(async () => {
+   *       try {
+   *         await Model.transaction(() => User.create({ data: { email } }))
+   *       } catch (e) { if (!(e instanceof UniqueConstraintError)) throw e }
+   *       await Audit.create({ ... })   // fine on both
+   *     })
+   *
+   * **One statement at a time.** Every query in scope runs on one reserved
+   * connection, so `Promise.all` over several ORM calls inside the callback is
+   * not safe here — the ordinary, encouraged thing everywhere else in a Bun
+   * codebase. Await them in sequence. (`$exec` already runs its own nested
+   * writes and relation reads sequentially for this reason.)
    */
   static transaction<T>(fn: () => Promise<T>): Promise<T> {
     return withTransaction(app(DatabaseManager).sql, () => fn());
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -19,6 +19,7 @@ import {
   MissingModelSchemaError,
   RecordNotFoundError,
   UniqueConstraintError,
+  UnregisteredPolicyClassError,
 } from "./errors";
 import { getOrCompile, type Operation, type QueryPlan } from "./plan";
 import {
@@ -207,6 +208,23 @@ export abstract class Model {
     let effective = args;
 
     if (policies.length > 0) {
+      // The registry must resolve this model's name to *this* class, or nested
+      // relation reads — which go through the registry — would run on a
+      // different class and skip these policies entirely. Scoped at the root,
+      // unscoped inside an include; see `UnregisteredPolicyClassError`. Checked
+      // here rather than at registration because a class can acquire a policy
+      // at any point after it is defined.
+      const registered = registry.has(schema.name)
+        ? registry.get<unknown>(schema.name)
+        : undefined;
+      if (registered !== undefined && registered !== this) {
+        throw new UnregisteredPolicyClassError(
+          schema.name,
+          (registered as { name?: string }).name ?? String(registered),
+          this.name,
+        );
+      }
+
       // Deny-by-default lives on the context's `user` accessor, not here: a
       // policy that never consults the user — a soft-delete scope, say — has
       // nothing to deny and must keep working with no request in sight. See
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -2,7 +2,7 @@ import type { SQL } from "bun";
 import { DatabaseManager } from "../database/DatabaseManager";
 import { app } from "../foundation/app";
 import { type BindContext, createBindContext } from "./compile/fragment";
-import { currentTransaction, withTransaction } from "./context";
+import { currentTransaction, isSystemScope, runAsSystem, withTransaction } from "./context";
 import {
   type NestedWriteStep,
   type RelationExecutor,
@@ -11,10 +11,19 @@ import {
 import { dialectFor, type SqlDialect } from "./dialect";
 import {
   MissingModelSchemaError,
+  PolicyDeniedError,
   RecordNotFoundError,
   UniqueConstraintError,
 } from "./errors";
 import { getOrCompile, type Operation, type QueryPlan } from "./plan";
+import {
+  applyPolicies,
+  applyRedaction,
+  currentUser,
+  policiesFor,
+  type ModelPolicy,
+  type PolicyContext,
+} from "./policy";
 import * as registry from "./registry";
 import type { ModelSchema } from "./schema";
 
@@ -48,6 +57,28 @@ const ORTHROW = new Set([
 export abstract class Model {
   /** Assigned by the generated subclass from `app/models/generated/schema.ts`. */
   static $schema: ModelSchema;
+
+  /**
+   * The model's own policy. Optional, inherited through the prototype chain so
+   * a shared base can contribute one to every subclass — see `policiesFor` for
+   * the order, which is base first and narrowing-only.
+   */
+  static $policy?: ModelPolicy;
+
+  /**
+   * Runs `fn` with policies suspended, for code that has no user and knows it:
+   * a cron tick, a queue worker, a seed script, a migration.
+   *
+   * Deliberately a wrapper rather than a flag or a fallback. Under
+   * deny-by-default the alternative to writing this is an error, which is the
+   * point — unscoped access is a sentence someone typed, never what happens
+   * when a user fails to turn up.
+   *
+   *     await Model.asSystem(() => User.findMany({}))
+   */
+  static asSystem<T>(fn: () => Promise<T>): Promise<T> {
+    return runAsSystem(fn);
+  }
 
   static $modelSchema(): ModelSchema {
     const schema = this.$schema;
@@ -95,7 +126,39 @@ export abstract class Model {
     // while its neighbours roll back.
     const conn = currentTransaction() ?? db.sql;
 
-    const plan = getOrCompile(schema, op, args, dialect);
+    // POLICIES, AND THE ORDER MATTERS MORE HERE THAN ANYWHERE ELSE IN THE ORM.
+    //
+    // Policies rewrite the argument tree, so they change the SQL. They must
+    // therefore run *before* `getOrCompile`, which keys the plan cache on those
+    // arguments. Reordered — compile first, scope after — two users with the
+    // same query shape and different scopes would share one plan, and one of
+    // them would run the other's SQL. That is a cross-tenant data leak produced
+    // by a caching bug, and it would not look like one.
+    //
+    // A scope that injects a *value* keeps the same shape for every user, so
+    // the plan is still shared and only the bound value differs. That is the
+    // desired outcome, not a compromise, and `policy.plan-cache.test.ts` pins
+    // both halves: same shape means one plan and different parameters, and a
+    // scope whose shape varies by user means a different plan key.
+    // `asSystem` suspends the whole hook, not just the no-user check: a script
+    // that has said it is a script should not then be scoped to a user that
+    // happens to be in the request store.
+    const policies = isSystemScope() ? [] : policiesFor(this);
+    let policy: PolicyContext | undefined;
+    let effective = args;
+
+    if (policies.length > 0) {
+      policy = { user: currentUser(), operation: op, model: schema.name };
+
+      // Deny by default. "No user" is not "no policy" — see PolicyDeniedError.
+      if (policy.user === null) {
+        throw new PolicyDeniedError(schema.name, op, "no-user");
+      }
+
+      effective = applyPolicies(policies, policy, args);
+    }
+
+    const plan = getOrCompile(schema, op, effective, dialect);
 
     const executor: RelationExecutor = {
       exec: (model, operation, relationArgs) =>
@@ -119,7 +182,7 @@ export abstract class Model {
     // per call would put a `BEGIN` around every query in the framework and turn
     // a caller's transaction into a nest of savepoints it never asked for.
     // Outside one, a failing later step still leaves the earlier rows written.
-    await runSteps(plan.before, args, context, executor, []);
+    await runSteps(plan.before, effective, context, executor, []);
 
     // `unsafe` despite the name. Bun's tagged template cannot express a query
     // whose *shape* is dynamic, which every ORM query is. Safety here does not
@@ -132,7 +195,7 @@ export abstract class Model {
       schema,
       op,
       plan.text,
-      plan.bind(args, context),
+      plan.bind(effective, context),
     );
 
     const result = this.$shape(plan, rows as unknown[]);
@@ -147,7 +210,7 @@ export abstract class Model {
     // Rows that could not exist until this one did: the far side of a relation
     // whose foreign key lives on the child. Run before relations are attached,
     // so an `include` on the same call sees what was just written.
-    await runSteps(plan.after, args, context, executor, rowsOf(result));
+    await runSteps(plan.after, effective, context, executor, rowsOf(result));
 
     // Relations are loaded after the root rows are shaped, one query per node
     // in the include tree. Each of those queries is `$exec` on the *related
@@ -163,7 +226,7 @@ export abstract class Model {
         plan.relations,
         plan.hidden,
         result,
-        args,
+        effective,
         executor,
       );
     } else if (plan.hidden !== undefined && plan.hidden.length > 0) {
@@ -174,6 +237,11 @@ export abstract class Model {
         for (const key of plan.hidden) delete row[key];
       }
     }
+
+    // Redaction last, on the shaped result. After relations, not before: a
+    // related row was shaped by its own model's `$exec` and has already been
+    // through its own policy's `redact` — this one only owns its own rows.
+    if (policy) applyRedaction(policies, policy, result);
 
     return result;
   }
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -207,24 +207,38 @@ export abstract class Model {
     let policy: PolicyContext | undefined;
     let effective = args;
 
-    if (policies.length > 0) {
-      // A nested relation read resolves its target through the registry, so it
-      // runs whatever is registered under this model's name — not necessarily
-      // the class the caller queried. If the two disagree about *policies*, the
-      // same query is scoped at the root and unscoped inside an include. See
-      // `UnregisteredPolicyClassError`.
-      //
-      // The comparison is on the resolved policy **chain**, not on class
-      // identity, and the difference matters: a plain subclass of a registered,
-      // policied class carries the same policy objects in the same order by
-      // inheritance, so a nested read resolving to the parent applies exactly
-      // what the root query applied. There is no divergence and nothing to
-      // refuse. Checking identity instead rejected `class AdminUser extends User
-      // {}` — a typed view that adds nothing — with an error about policies it
-      // did not write.
-      //
-      // Checked here rather than at registration because a class can acquire a
-      // policy at any point after it is defined.
+    // POLICY-DIVERGENCE GUARD.
+    //
+    // A nested relation read resolves its target through the registry, so it
+    // runs whatever is registered under this model's name — not necessarily the
+    // class the caller queried. When those two disagree about *policies*, the
+    // same policy applies to some queries and not others, silently. Both
+    // directions are reachable and both are the same bug:
+    //
+    //   policied class not registered  -> root scoped, nested reads unscoped
+    //   policied class registered, but -> nested reads scoped, a direct query
+    //   the caller queries the base       through the base unscoped
+    //
+    // The second is the one this check sat inside `policies.length > 0` and
+    // therefore missed: querying `AccountModel` when `Account` owns the name
+    // finds no policies on `AccountModel`, so the old placement never looked.
+    // `AccountModel` and `Account` come out of the same barrel with near
+    // identical names, so it needs no include and no mistake worth calling one.
+    //
+    // Keyed on the *divergence*, not on either class having policies. The
+    // comparison is of the resolved policy chain rather than class identity: a
+    // plain subclass of a registered, policied class inherits the same policy
+    // objects in the same order, so nothing diverges and there is nothing to
+    // refuse — checking identity rejected `class AdminUser extends User {}`, a
+    // typed view, for policies it had not written.
+    //
+    // `!system` is load-bearing. Under `asSystem` policies are suspended by
+    // design, so a seed script querying the generated base directly is correct
+    // and must not raise.
+    //
+    // Cost on the common path: one `Map.get` and one reference compare, for a
+    // model where nothing is registered under a different class.
+    if (!system) {
       const registered = registry.has(schema.name)
         ? registry.get<unknown>(schema.name)
         : undefined;
@@ -233,17 +247,20 @@ export abstract class Model {
         const theirs = policiesFor(registered);
         const diverges =
           policies.length !== theirs.length ||
-          policies.some((policy, index) => policy !== theirs[index]);
+          policies.some((entry, index) => entry !== theirs[index]);
 
         if (diverges) {
           throw new UnregisteredPolicyClassError(
             schema.name,
             (registered as { name?: string }).name ?? String(registered),
             this.name,
+            policies.length > 0 ? "queried" : "registered",
           );
         }
       }
+    }
 
+    if (policies.length > 0) {
       // Deny-by-default lives on the context's `user` accessor, not here: a
       // policy that never consults the user — a soft-delete scope, say — has
       // nothing to deny and must keep working with no request in sight. See
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -2,7 +2,13 @@ import type { SQL } from "bun";
 import { DatabaseManager } from "../database/DatabaseManager";
 import { app } from "../foundation/app";
 import { type BindContext, createBindContext } from "./compile/fragment";
-import { currentTransaction, isSystemScope, runAsSystem, withTransaction } from "./context";
+import {
+  currentTransaction,
+  isSystemScope,
+  runAsSystem,
+  runAsUser,
+  withTransaction,
+} from "./context";
 import {
   type NestedWriteStep,
   type RelationExecutor,
@@ -78,6 +84,23 @@ export abstract class Model {
    */
   static asSystem<T>(fn: () => Promise<T>): Promise<T> {
     return runAsSystem(fn);
+  }
+
+  /**
+   * Runs `fn` with policies scoped to `user`, for code that acts on somebody's
+   * behalf without a request to read them from — a queue job, a scheduled
+   * report, a test.
+   *
+   * The narrow counterpart to `asSystem`, and deliberately the easier of the
+   * two to reach for. A worker handling "send the invoice for organisation 7"
+   * acts *as* a user; giving it only `asSystem` would suspend policies outright
+   * and leave it hand-scoping every query, which is the unscoped-by-accident
+   * outcome deny-by-default exists to prevent.
+   *
+   *     await Model.asUser(owner, () => Invoice.findMany({}))
+   */
+  static asUser<T>(user: unknown, fn: () => Promise<T>): Promise<T> {
+    return runAsUser(user, fn);
   }
 
   static $modelSchema(): ModelSchema {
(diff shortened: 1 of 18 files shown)


packages/gemi/orm/context.ts
```diff
@@ -53,7 +78,61 @@ export function currentTransaction(): TransactionSQL | undefined {
...
+ * It is also what makes policies testable without reaching into
+ * `RequestContext`'s internals.
+ *
+ * Takes precedence over the request store, so a job that says who it is acting
+ * as is not quietly overridden by an ambient request that happens to enclose it.
+ */
+export function runAsUser<T>(user: unknown, fn: () => Promise<T>): Promise<T> {
+  const current = ormContext.getStore();
+  return ormContext.run(
+    { ...current, depth: current?.depth ?? 0, actor: { user } },
```
nstfkc: **`asUser(undefined)` walks straight through deny-by-default and reads unscoped.**

`user` is `unknown` and stored verbatim, so `undefined` reaches `policyContext`, where `hasUser: user !== null` is **true** and the accessor's `user === null` check does not fire. Ran it against this branch:

```
user=null      : DENIED (PolicyDeniedError)          ← correct
user=undefined : hasUser=true  where={"active":true,"AND":[{}]}
```

That `AND: [{}]` is the leak this file's own doc describes: `{ organizationId: ctx.user?.organizationId }` collapses to an empty object, an absent filter, and the query returns **every tenant's rows** — while `hasUser` tells a branching policy there is a user.

The request path cannot produce it (`requestStore()?.user ?? null` normalises), so this is specific to `asUser` — and `asUser` is the one the PR argues should be the easy tool a queue worker reaches for. `asUser(usersById.get(job.userId), …)` and `asUser(list.find(…), …)` both hand it `undefined` on a miss, which is precisely the "user failed to turn up" case deny-by-default exists for.

One line, either end:

```ts
actor: { user: user ?? null }        // here, or
if (user == null && !system)         // loose equality in the accessor
```

The accessor version also fixes `hasUser`, which should be `user != null` for the same reason.

packages/gemi/orm/context.ts
```diff
@@ -53,7 +78,61 @@ export function currentTransaction(): TransactionSQL | undefined {
...
+ * unscoped-by-accident failure the deny-by-default rule exists to prevent. The
+ * narrow tool should be the easy one.
+ *
+ * It is also what makes policies testable without reaching into
+ * `RequestContext`'s internals.
+ *
+ * Takes precedence over the request store, so a job that says who it is acting
+ * as is not quietly overridden by an ambient request that happens to enclose it.
+ */
+export function runAsUser<T>(user: unknown, fn: () => Promise<T>): Promise<T> {
```
nstfkc, line 139: **`asUser` nested inside `asSystem` is silently ignored — the query stays unscoped.**

`runAsUser` spreads `...current`, which carries `system: true` forward, and `$exec` short-circuits on `isSystemScope()` before it ever looks at the actor:

```
inside asSystem             : isSystemScope=true
inside asUser(org 7) within : isSystemScope=true  currentUser={"organizationId":7}
  -> $exec uses policies = []   (UNSCOPED)
```

So the actor is set, `currentUser()` returns it, and nothing scopes on it. The reverse nesting behaves correctly (`asSystem` inside `asUser` suspends, as it should).

This is a realistic shape rather than a contrived one: a worker enters `asSystem` to read its job queue, then narrows to the job's owner for the actual work. The PR's own argument for shipping `asUser` is that the narrow tool must be the easy one so nobody reaches for the blunt one — and here reaching for the narrow one inside the blunt one silently does nothing.

Acting as a user is strictly narrower than acting as the system, so I would have `runAsUser` clear it:

```ts
{ ...current, depth: current?.depth ?? 0, system: false, actor: { user } }
```

If the intended semantics really are "system wins", that deserves a sentence in `asSystem`'s JSDoc and a test pinning it — right now it is neither documented nor covered, and the failure is silent.

packages/gemi/orm/policy.ts
```diff
@@ -0,0 +1,497 @@
...
+      : Array.isArray(existing)
+        ? existing
+        : [existing];
+
+  return { ...args, where: { ...where, AND: [...members, scope] } };
+}
+
+/** Runs an `onCreate` over whichever shape the operation's payload takes. */
+function withCreated(args: any, context: PolicyContext, policy: ModelPolicy): any {
+  const run = (data: any) => policy.onCreate!(context, data ?? {});
```
nstfkc: Minor, and only a doc/robustness point: `applyPolicies` promises "The caller's `args` is never mutated" and gives the right reason — a mutation would apply twice if the args object were reused. The top-level spreads honour that, but this hands the caller's *actual* `data` object to `onCreate`, so a policy written as

```ts
onCreate: (ctx, data) => { data.organizationId = ctx.user.organizationId; return data }
```

mutates the caller's object and reintroduces exactly the double-apply hazard the comment cites. `redact` has the same property by design, but there the mutation *is* the interface.

Policies are application-authored, so this is worth either a shallow copy here (`run({ ...data })`) or a line in `onCreate`'s doc saying it must not mutate — the current comment reads as a guarantee the mechanism provides rather than one the policy author has to uphold.

nstfkc (commented):
Reviewed `e80a0c0` in an isolated worktree. The headline claim holds and is the strongest result in the stack: a relation read *is* `$exec` on the related model, so nested scoping falls out of invariant 1 rather than being maintained separately. There is no second path to keep in step, which is exactly why Prisma's extensions cannot do this.

| Verified at `e80a0c0` | |
|---|---|
| `vitest run` (packages/gemi) | **531 pass / 7 fail** — the same pre-existing router/i18n files |
| `vitest run app/models/` (SQLite) | **267 pass, 4 skipped**, 8 files |
| `tsc --noEmit` | clean |

Every number matches the PR's table. The ordering argument is right and the code implements it: policies run before `getOrCompile`, so the plan key is computed on post-policy args — a scope that injects a value shares a plan and differs only in the bound parameter, and a scope whose *shape* varies gets its own key. Redaction correctly runs last, after relations, with each model redacting only its own rows.

Moving deny-by-default onto the `user` accessor rather than an up-front model check is the right correction, and the reasoning — raising on *access* because `{ organizationId: undefined }` is an absent filter, so the leak is the quiet version of the error — is exactly the right instinct.

**Which is why two findings below matter: both are that same leak, through doors the accessor does not cover.** Both are one-line fixes, and I confirmed both by running this branch's own code rather than by reading it.

The `withScope` fix (`{ ...caller, AND: [...theirs, scope] }` rather than wrapping) is subtle and correct — keeping the caller's keys at the top level where `matchUniqueKey` reads them, while still conjoining. Finding that a previous test was passing on the wrong error, and replacing it with `RecordNotFoundError` plus two positive cases, is the right repair.

`softDeletes()` earns its place: thirty lines against the same public `ModelPolicy`, and keeping the operation-rewriting half at the call site — because `$exec` dispatches on the operation before consulting a policy — is the correct boundary.

nstfkc (commented):
Checked `1f115a8`. The delta since `e80a0c0` is only the merge of #50's `db1b85f` — `policy.ts`, `context.ts` and `Model.ts` are byte-identical, so this is a verification pass, not a re-review.

Merged head is green:

| Verified at `1f115a8` | |
|---|---|
| `vitest run orm/` | 451 pass, 13 files |
| `vitest run app/models/` (SQLite) | 267 pass, 4 skipped |
| `tsc --noEmit` | clean |

**The two policy-bypass findings from my review of `e80a0c0` are still open**, since nothing in the merge touched them:

1. `asUser(undefined)` — `hasUser` is `true`, the accessor's `user === null` check does not fire, and the scope collapses to `AND: [{}]`, reading every tenant's rows. Fix: `user ?? null` in `runAsUser`, or `== null` in the accessor (which also fixes `hasUser`).
2. `asUser` nested inside `asSystem` — `runAsUser` spreads `system: true` forward and `$exec` short-circuits to `policies = []`, so an explicitly-named user is silently unscoped. Fix: `system: false` in `runAsUser`, or document and test the "system wins" semantics.

Plus the minor `onCreate` mutation note. Nothing new from this commit.

nstfkc:
All three fixed in `297bda6`. Both bypasses were real, and your framing of them is the part I want to acknowledge: they are the *same* leak the accessor exists to prevent, reached through doors the accessor doesn't cover. That's a sharper diagnosis than "two bugs".

**`asUser(undefined)`.** Confirmed. What makes it nasty is that *defensive* code is what makes it silent — `ctx.user?.organizationId` collapses to `{}` and Prisma reads that as no filter, where a non-optional `ctx.user.organizationId` would at least have thrown. So the more careful policy is the one that leaks.

I took your accessor suggestion rather than only normalising at the boundary, because it covers any future caller of `policyContext` instead of just this one — `== null` in the accessor and `!= null` for `hasUser`. I also added `?? null` in `runAsUser` so the store's contents are honest, but the accessor is the load-bearing half.

**`asUser` inside `asSystem`.** Also confirmed, and I agree with your reading over the alternative: acting as a user is strictly narrower than acting as the system, so naming one has to narrow back. `runAsUser` clears the flag. The shape you describe is the one that decided it — a worker entering `asSystem` for its queue and narrowing to the job's owner — and it makes the failure the exact opposite of what this PR argues for: reaching for the narrow tool inside the blunt one did nothing at all. "System wins" would have needed a defence I can't make.

**The `onCreate` mutation.** Right, and the distinction you drew is the one I'd missed: my comment reads as a guarantee the mechanism provides, while it was really a request to the policy author. Shallow copy, so it's now the former. `redact` stays as it is — there the mutation *is* the interface, as you say.

Six tests, each failing without its fix: null and undefined both denying, `hasUser` false for both, a mutating `onCreate` unable to reach the caller's object on `create` and on every row of a `createMany`, and both nesting directions end-to-end.

| Verified at `297bda6` | |
|---|---|
| `vitest run orm/` | **457 pass** (was 451) |
| template, SQLite | **271 pass**, 4 skipped |
| template, Postgres 16 | **341 pass** |
| `tsc --noEmit`, `oxlint orm/` | clean |

Ran the Postgres pass too, since these are authorization fixes and the scope lands in a `where` — the only failing files there are the two documented one-dialect-at-a-time differential suites.

nstfkc (commented):
Reviewed `297bda6`. All three fixes from the last round are correct and each landed with a test that fails without it — `== null` in the accessor covers any future caller rather than just `asUser`, `system: false` makes the narrower tool actually narrow, and the shallow copy turns the no-mutation promise into one the mechanism keeps. 457 orm tests, 271 template + 4 skipped, tsc and oxlint clean.

**But while checking that the fixes held, I found something bigger, and I think it blocks the PR.**

**Under the pattern this PR documents, the headline feature does not fire — a nested `include` reads across tenants.** I ran the PR description's own example verbatim against the template's schema and a real SQLite database:

```ts
class Account extends AccountModel { static $policy = tenantScoped }   // as documented
class User    extends UserModel {}

Account.findMany({})                                 -> 1 account,  orgs=[1]      ✅ scoped
User.findMany({ include: { accounts: true } })       -> 2 accounts, orgs=[1,2]    ❌ LEAK
```

Bob's account, from the other organisation, comes back inside Alice's include. That is precisely the Prisma behaviour this iteration exists to fix — "a scope written for `Account` is bypassed by that exact call. Ours is not." Under the documented pattern, ours is.

**Root cause, and it is one line away from the claim being true.** `generated/index.ts` registers the *generated* class:

```ts
register("Account", AccountModel);
```

A nested read resolves through that registry — `Model.ts:222` — so `$exec` runs on `AccountModel`, and `policiesFor(this)` at `Model.ts:205` walks **`AccountModel`'s** prototype chain. The policy lives on `Account extends AccountModel`, which is not on that chain. Nothing is registered under the app subclass, and nothing in the template re-registers one (I checked: the only `register(` calls outside `generated/index.ts` are in two test files).

So the split is exact: **root queries scope correctly** (you call the app subclass, so `this` carries the policy) and **every nested read is unscoped** — the one case that motivated the feature. `redact` and a `before` deny are bypassed on nested reads the same way, for the same reason.

Why the suite is green on this: details inline on the headline test.

I have not proposed a fix because the right one is a design call between at least three — registering the app subclass instead of the generated one, having the generator emit the app-layer registration, or keying policies by schema name in their own registry rather than by class identity. All three are small; picking between them is yours.

Everything else in this PR is in good shape, and the mechanism itself is right — it is the wiring between the registry and the documented authoring pattern that does not connect.

templates/saas-starter/app/models/policies.test.ts
```diff
@@ -0,0 +1,738 @@
...
+     * `Account` is scoped to the caller's organisation. Reading users with
+     * `include: { accounts: true }` must not return accounts from another one.
+     * Prisma's extensions do not fire on that nested read at all; ours does,
+     * because the nested read is `$exec` on `AccountModel` and therefore goes
+     * through the same hook as a root query.
+     *
+     * Note `User` is deliberately *unpolicied* here, so the only thing that can
+     * scope the accounts is `Account`'s own policy firing on a nested read.
+     */
+    test("a scope on the related model applies to a nested include", async () => {
```
nstfkc, line 246: **This test passes for a reason that does not hold in an application.**

```ts
(AccountModel as any).$policy = tenantScoped;
```

It attaches the policy to the **generated** class — which is the one `generated/index.ts` registers, so the nested read finds it. But that is not how the PR tells an author to write a policy. The description's example, and `app/models/User.ts`'s own comment ("This class is the place your own code goes: scopes, policies, observers"), both put it on the subclass:

```ts
export class Account extends AccountModel {
  static $policy = { scope: …, onCreate: … }
}
```

and that class is never registered, so `registry.get("Account")` returns the unpolicied `AccountModel` and the nested read runs unscoped. I confirmed the leak end-to-end against a real database — 2 accounts across 2 organisations returned to a caller scoped to one.

So the single most important test in the iteration is green while the documented usage leaks. Test-shaped access to the mechanism is a legitimate thing to want, but here it routes around the exact wiring the headline claim depends on.

This PR already fixed one test that was "passing for the wrong reason" (the bare `rejects.toThrow()` that the unique-field error satisfied). This is the same category, one level up: it is not the assertion that is wrong, it is the setup.

Whatever the fix to the registration is, this test should be rewritten to declare its policy the way an application does — on a subclass — so that it fails if the wiring regresses. It would fail today.

nstfkc:
**You're right, and this was the most important finding on the stack.** Fixed in `0b62355`.

I reproduced it against a real database before designing anything, and got exactly your numbers — 1 account scoped at the root, 2 accounts across 2 organisations inside the include. The headline feature did not work under the pattern this PR documents, and the test asserting it was green.

Your diagnosis of *why the suite missed it* is the part I want to sit with: the assertion was right and the **setup** routed around the wiring the claim depends on. This PR had already fixed one test passing on the wrong error; this is the same category one level up. A test that constructs its subject differently from the documentation is testing something the documentation does not describe.

**The fix, in two halves.**

*Registration.* The app subclass owns the model's name. The template now shows it — `register("User", User)` beside the class, with the reasoning there rather than in a commit message.

*A guard, because a required line is a forgettable one.* `$exec` raises `UnregisteredPolicyClassError` when a class carries policies and is not the registered one, with the exact registration to add in the message. I chose this over your other two options: registering the app subclass alone leaves the same forget-and-leak hole, and a name-keyed policy registry only moves the problem — `before` and `redact` dispatch by class too, and so will observers.

It earned its place immediately: **turning it on failed 12 existing tests**, each a place where a policy-carrying class wasn't registered. Twelve latent leaks in the suite that introduced the feature.

It also refuses two policy-carrying classes for one model name, which surfaced something in my own base-class test: it queried both levels of its chain, and only one can own the name because a nested read resolves to exactly one. It now queries the leaf and inherits the base's scope through the prototype chain — which is what the test was actually about.

Every case in the policy and soft-delete suites now declares its policy on a registered subclass, plus the new case you asked for: the unregistered arrangement **raises** rather than reading across tenants, and it fails if the wiring regresses.

**One residual gap I want on the record rather than discovered later.** The guard fires on a query *through* the policy-carrying class. A model only ever reached by a nested include, never queried at its root, would not trip it — the registry would quietly hold the unpolicied base. Closing that properly needs the generator to emit the app-layer registration, which couples it to app file layout. Worth doing; not this fix.

| Verified at `0b62355` | |
|---|---|
| `vitest run orm/` | **457 pass** |
| template, SQLite | **272 pass**, 4 skipped |
| template, Postgres 16 | **343 pass** |

Thank you for running the documented example verbatim instead of reading the test. That is the check I should have made when I wrote the claim.

packages/gemi/orm/Model.ts
```diff
@@ -130,7 +185,55 @@ export abstract class Model {
...
+      // The registry must resolve this model's name to *this* class, or nested
+      // relation reads — which go through the registry — would run on a
+      // different class and skip these policies entirely. Scoped at the root,
+      // unscoped inside an include; see `UnregisteredPolicyClassError`. Checked
+      // here rather than at registration because a class can acquire a policy
+      // at any point after it is defined.
+      const registered = registry.has(schema.name)
+        ? registry.get<unknown>(schema.name)
+        : undefined;
+      if (registered !== undefined && registered !== this) {
```
nstfkc, line 246: **The identity check is stricter than the leak it prevents, and it refuses a legitimate arrangement.**

A plain subclass of the *registered, policied* class — carrying no policy of its own — is rejected:

```ts
class Acct extends AccountModel { static $policy = tenantScoped }
register("Account", Acct)

class AdminAcct extends Acct {}          // a typed view, no policy of its own
AdminAcct.findMany({})
// -> UnregisteredPolicyClassError: AdminAcct carries policies but the registry
//    resolves 'Account' to Acct
```

(Run against this commit.) `policiesFor` walks the prototype chain, so `AdminAcct` "carries policies" by inheritance, `policies.length > 0`, and the identity check fires. But there is no leak in this arrangement: `policiesFor(AdminAcct)` and `policiesFor(Acct)` are the *same policy objects in the same order*, so a nested read resolving to `Acct` applies exactly what the root query applied.

The commit message says the guard "correctly refuses two policy-carrying classes for one model name" — which it does, and should. This case is different: one policy, two classes, no divergence.

The condition that matches the actual risk is a difference in the *policy chain*, not in class identity:

```ts
const mine = policies;                       // already computed above
const theirs = registered === undefined ? mine : policiesFor(registered);
const diverges =
  mine.length !== theirs.length || mine.some((p, i) => p !== theirs[i]);
if (registered !== undefined && diverges) throw new UnregisteredPolicyClassError(…);
```

That still catches the original leak (`policiesFor(Account)` = `[tenantScoped]` vs `policiesFor(AccountModel)` = `[]`), still catches two rival policy-carrying classes (different objects), and stops refusing subclasses that add nothing.

If the intent really is "exactly one class per model name, no exceptions", that is defensible too — but then it is a broader rule than the error message describes, and it deserves saying so, because `class AdminUser extends User {}` as a typed alias is a natural thing to write and currently fails with an error about policies it did not write.

nstfkc (commented):
Reviewed `0b62355`. **The leak is fixed.** I re-ran my original repro against this commit and it now raises instead of returning another tenant's rows:

```
root query through unregistered policy class -> UnregisteredPolicyClassError
```

The two-part response is the right shape: registration is the actual fix, and the guard exists because a required line is a forgettable one. Converting a silent cross-tenant read into a loud failure at the first query is the correct trade, and the error message naming the exact `register(...)` line to add is what makes it actionable at 3am. That turning it on failed 12 existing tests — twelve places where a policy-carrying class was not registered, in the suite that introduced the feature — is the strongest possible argument for the guard existing.

I also appreciate the framing in the commit message: a test that constructs its subject differently from the documentation is testing something the documentation does not describe. That is the general lesson, and it is worth more than the specific fix.

| Verified at `0b62355` | |
|---|---|
| `vitest run orm/` | 457 pass, 13 files |
| `vitest run app/models/` (SQLite) | 272 pass, 4 skipped |
| `tsc --noEmit` | clean |

**I reproduced the residual gap you named**, so it is on the record with a number rather than as a caveat:

```
class Account extends AccountModel { static $policy = tenantScoped }   // never queried at its root
User.findMany({ include: { accounts: true } })  ->  2 accounts, orgs=[1,2]   still leaks
```

Your description of it is accurate. Worth stating the shape of the limit precisely: the guard fires on a query *through* the policy-carrying class, so its coverage depends on that class being queried at its root at least once in the process — which is not a property an application controls, asserts, or notices the absence of. It is a good net, not a proof, and an include-only model (a membership or pivot-ish model — exactly the kind that carries a tenant scope) is the shape that slips through. Agreed it wants the generator, and agreed it is not this fix.

One new finding on the guard itself, inline — it is stricter than the risk it protects against, and rejects a legitimate arrangement.

nstfkc:
Fixed in `bfa8e57`. You're right — the guard was stricter than the risk, and your condition is the correct one.

The distinction I'd conflated: my previous commit said the guard "correctly refuses two policy-carrying classes for one model name", which it does and should. `class AdminAcct extends Acct {}` is a different case wearing the same shape — **one policy, two classes, no divergence.** `policiesFor` returns the same objects in the same order, so a nested read resolving to the parent applies exactly what the root query applied. Refusing it was refusing a typed view for policies it hadn't written.

The comparison is on the resolved policy chain now. It still catches the original leak (`[tenantScoped]` vs `[]`) and still catches two rival policy-carrying classes, since those are different objects.

Two tests, one per side — a subclass adding nothing is allowed *and still scoped*, and a subclass adding its own policy must own the name. The second is what keeps this from being a loosening.

**On the residual gap: thank you for reproducing it with a number.** Your framing sharpens what I wrote — the guard's coverage depends on the policy-carrying class being queried at its root at least once in the process, and that is "not a property an application controls, asserts, or notices the absence of". That is a better statement of the limit than mine, and the observation that an include-only model is *exactly* the shape that carries a tenant scope is the part that makes it worth prioritising rather than filing.

I've left it out of this PR deliberately — the fix is generator-side and couples the generator to app file layout, which is a design decision that deserves its own review rather than being bolted onto a fix commit. It should be the first thing in whatever follows, not a "notes and risks" bullet.

| Verified at `bfa8e57` | |
|---|---|
| `vitest run orm/` | 457 pass |
| template, SQLite | 274 pass, 4 skipped |

nstfkc:
Correction to my previous comment: the commit is **`b0e3f34`**, not `bfa8e57`. I wrote the hash before pushing rather than reading it back. Everything else stands, and the verification numbers were taken from the working tree that became `b0e3f34`.

packages/gemi/orm/Model.ts
```diff
@@ -130,7 +185,74 @@ export abstract class Model {
...
+    // scope whose shape varies by user means a different plan key.
+    // `asSystem` suspends the whole hook, not just the no-user check: a script
+    // that has said it is a script should not then be scoped to a user that
+    // happens to be in the request store.
+    const system = isSystemScope();
+    const policies = system ? [] : policiesFor(this);
+    let policy: PolicyContext | undefined;
+    let effective = args;
+
+    if (policies.length > 0) {
```
nstfkc: **The guard lives inside `if (policies.length > 0)`, so a query through a class with *no* policies never reaches it — including the generated base whose name a policied subclass owns.**

Run against this commit:

```ts
class Acct3 extends AccountModel { static $policy = tenantScoped }
register("Account", Acct3)

AccountModel.findMany({})   // the generated base, still exported from ./generated
// -> 2 rows, orgs=[1,2]   UNSCOPED, no error
```

This is the previous bug reflected: there, the policied class was unregistered and *nested* reads were unscoped; here the policied class **is** registered — so nested reads are correctly scoped — and a *direct* query through the base skips the policy instead. Both are "the scope silently does not apply", and this one needs no include to reach.

It is not exotic. `AccountModel` and `Acct`/`Account` are exported from the same barrel with near-identical names, and querying the generated base is exactly what every test in this suite does — `UserModel.findMany(...)`, `AccountModel.create(...)`. The template gets away with it only because it declares no policies today, which is the same reason the original leak survived review.

The condition wants to key off the *divergence*, not off the queried class having policies. Something like:

```ts
const registered = registry.has(schema.name)
  ? registry.get<unknown>(schema.name)
  : undefined;

// Only when they actually differ — the common path is one Map.get and a
// reference compare, so an unpolicied model still pays almost nothing.
if (!system && registered !== undefined && registered !== this) {
  const theirs = policiesFor(registered);
  const diverges =
    policies.length !== theirs.length ||
    policies.some((policy, index) => policy !== theirs[index]);
  if (diverges) throw new UnregisteredPolicyClassError(…);
}
```

Note the `!system` guard is load-bearing if you move it out: `policies` is already `[]` under `asSystem`, so without it a seed script legitimately querying the base would start raising.

The error message would need a second voice for this direction — "the registry resolves 'Account' to `Acct3`, which carries policies you are skipping by querying `AccountModel` directly" — since the current wording assumes the queried class is the policied one.

(If querying the generated base is *intended* as an escape hatch, that is a defensible design — but then it is a second, undocumented one alongside `asSystem`, and it is the one nobody had to type `asSystem` to get.)

nstfkc (commented):
Reviewed `b0e3f34`. The chain comparison is right, and I verified both sides of the distinction against a real database:

```
AdminAcct.findMany()           -> 1 row, orgs=[1]  (allowed, and still scoped)
Rival (own policy).findMany()  -> UnregisteredPolicyClassError
```

So the typed-view case works *and* is scoped — checking the scoping half matters, since "allowed" would be no good if it were allowed and unpolicied. Two rival policy-carrying classes are still refused.

| Verified at `b0e3f34` | |
|---|---|
| `vitest run orm/` | 457 pass |
| `vitest run app/models/` (SQLite) | 274 pass, 4 skipped |
| `tsc --noEmit` | clean |

**One case the guard still does not reach, and it is the mirror image of the original bug.** Details inline — it is the same "policy silently does not apply" family, reachable from a plain import.

nstfkc:
Fixed in `ab6dcb0`. Correct on all counts — and thank you for checking the *scoping* half of the typed-view case, not just that it was allowed. "Allowed and unpolicied" would have been a worse outcome than refused.

**The mirror image is real and I'd missed it structurally**, not incidentally: the check sat inside `if (policies.length > 0)`, so a query through a class with no policies never looked. Your repro is the shape that matters — `AccountModel` and `Account` come out of the same barrel with near-identical names, and querying the generated base is what every test in this file did before the previous commit. The template only escapes it by declaring no policies, which is the same reason the original leak survived review.

The condition keys off the **divergence** now rather than off the queried class having policies, so it fires both ways. The error message has a second voice for the new direction, since the old wording assumed the caller was the policied party — it now points at the registered class and at `asSystem`, because if skipping is intended that is the way to say so visibly.

`!system` is load-bearing exactly as you flagged, and there is a test for the exemption specifically because it is the kind of condition someone later "simplifies" away. Without it every seeded fixture in the repo would raise.

**On your parenthetical** — no, querying the generated base is not an intended escape hatch. You put it precisely: it would be a second, undocumented one, and the one nobody had to type `asSystem` to get. `asSystem` should stay the only way to skip policies, because it is the only one that is visible at the call site.

One thing I want to name rather than let pass: this guard has now been wrong twice in opposite directions — too strict, then too narrow. Both times I wrote the condition from the example in front of me instead of from the property. The property is that *any* divergence between the queried class's policy chain and the registered one's is a silent partial application. That is what it says now, and it is what I should have written the first time.

| Verified at `ab6dcb0` | |
|---|---|
| `vitest run orm/` | 457 pass |
| template, SQLite | 276 pass, 4 skipped |
| template, Postgres 16 | 351 pass |

nstfkc (commented):
Reviewed `ab6dcb0`. Fixed, including the `!system` exemption and the second error voice. I ran all four arrangements against a real database:

```
base query (was the leak)   -> UnregisteredPolicyClassError [carries: registered]
base query under asSystem   -> 2 rows, orgs=[1,2]   allowed, as intended
typed view AdminAcct        -> 1 row,  orgs=[1]     allowed and scoped
registered policied class   -> 1 row,  orgs=[1]     scoped
```

Every cell is what it should be, and the two directions of the guard now report through different messages — the second one pointing at `asSystem` as the visible way to say "skip policies on purpose" is the right closing move, because it turns the one remaining bypass into something a reader can see at the call site.

| Verified at `ab6dcb0` | |
|---|---|
| `vitest run orm/` | 457 pass |
| `vitest run app/models/` (SQLite) | 276 pass, 4 skipped |
| `tsc --noEmit` | clean |

I also checked the guard cannot fire on an application that has no policies at all: both chains are empty, nothing diverges, nothing raises. So this is not a change existing code can trip over.

On the note at the end of your commit message — that the guard was wrong twice in opposite directions, both times from writing the condition against the example rather than the property — that is the right read, and worth keeping. The property (any divergence between the queried chain and the registered one is a silent partial application) is what the code now states, and it is why this version covers cases neither of us enumerated: two sibling subclasses sharing one policy object are correctly allowed, because the chains match.

**Nothing outstanding from me except the residual you already documented**: a policied model that is only ever reached through an include, and never queried at its root, still never trips the guard. That one needs the generator to emit the app-layer registration, and I agree it is not this PR's to fix. Everything I raised on this branch is now closed.

PR #52, opened 2026-07-27
ORM iteration 7 — performance: measured, and one optimisation justified

nstfkc:
Measure first, then close the gap.

> **Revised after review.** The first version of this PR concluded that deliverable 2 (`LATERAL` + `json_agg`) should not be built. That conclusion rested on a scenario that was not measuring what its name said — see "Deliverable 2" below. With the measurement corrected, **deliverable 2 is justified on Postgres and is the largest unclaimed win in the ORM.** Figures below are deliberately not restated from any single run; `plans/orm/benchmarks.md` is the artifact with provenance.

Stacked on #51. Review that first.

## What the numbers say

`plans/orm/benchmarks.md` is committed with machine and version provenance, and is the thing to read — every gemi row decomposes into compile / lookup / execute / shape, which is possible only because compile is pure and split from bind. Prisma's stages cannot be separated from outside it, so its column is a total.

Deliberately not restating figures here: an earlier version of this description quoted one run while the committed table held another, and the table is the artifact with provenance.

The shape of the results: gemi is a few percent to ~2× hand-written SQL depending on scenario, and **2–15× faster than Prisma** on identical queries. Transaction overhead is one round-trip pair, with the ALS ~25ns of it. Policy dispatch is ~40ns. The plan cache holds 4 entries across the whole run with zero evictions.

## Deliverable 2 (`LATERAL` + `json_agg`): justified on Postgres, not built here

The first version of this PR declined it. That was wrong, and the way it was wrong is the most useful thing in this PR.

Scenario 4 was added specifically to show the batched planner falling behind at depth 3, and it appeared not to — depth-3 read as *faster* than depth-2. Two defects, found because the reviewer asked me to check this claim hardest:

1. Scenario 4's baseline ran on a different connection from every other scenario's, so its `×raw` column was not computed on the same footing as scenario 3's.
2. With that equalised, the baseline still went the wrong way for strictly more work — because **the seed gave accounts no `organizationId`**, so every third-level foreign key was null, the batched loader correctly skipped that query, and the scenario measured depth-2 plus a filter pass.

Corrected, the Postgres cost is linear in include depth — roughly one whole round trip per level, against a point read of comparable size. gemi sits at ~1.0× hand-written SQL at both depths, so there is nothing to win at this *shape*; the win is in changing the shape, which is precisely the deliverable. The ceiling is about 2× on depth-3 and grows with depth.

It stays unjustified on **SQLite**, matching the plan's own note: in-process round trips are nearly free, and the SQLite depth-2/depth-3 pair is not consistently ordered anyway, since scenario 4 runs second and benefits from a warmer cache.

**Not built in this PR.** It is a new relation strategy plus the type-flattening and empty-to-many handling the plan describes, and it wants its own review rather than being appended to the PR that justified it. Acceptance criteria 3, 4 and 8's lateral clauses are therefore still open, and named as open rather than N/A — which is a change from the first version of this description.

## Deliverable 4: a generated shaper, 3.4×

Three variants measured before choosing, because a compiled body costs debuggability and is unavailable wherever `eval` is refused:

```
interpreted (closure, dynamic keys)   190µs    baseline
flat (parallel arrays)                198µs    -4%
compiled (new Function, literal)       55µs    +71%
```

**The flat variant measuring nothing is the informative result.** It rules out property-load indirection, leaving the object's *shape*: `shaped[key] = value` transitions the hidden class once per property, thirteen times per row, while a literal with known keys gets one hidden class and one allocation. That also means this is the end of the stage rather than a first step.

Both implementations ship — availability is probed once at module load, and the interpreted shaper is the fallback. It is not dead code: it runs there, and it is the reference the compiled one is asserted equal to. That equivalence suite is the reason two paths are acceptable at all, since a divergence would appear only on whichever runtime the author was not using. It also asserts compilation is genuinely available in CI, without which every "compiled" case could be silently comparing the interpreted shaper to itself.

`compile` roughly doubled, 3µs → 6µs, because compiling a plan now generates the shaper source. The right trade, stated rather than hidden: compile happens once per argument shape and is cached; shaping happens per row on every call; `lookup` — what a warm query pays — is unchanged.

**Positional row mode is verified to exist and not justified.** Bun's query object does expose `.values()`. It measured 14% *slower* on SQLite and 17% faster on Postgres, and the sign flipped between runs at 30 samples. With p95 near double p50 on Postgres, this workload cannot resolve it.

## Deliverables 3, 5, 6

**Strategy observability.** `RelationPlan.strategy` existed since iteration 3 and nothing read it, which made it a comment rather than a guarantee. `QueryPlan.strategies` now reports it. The boundary is asserted, not implied: a nested level compiles its own plan and picks its own strategy, so the root can only answer for its own nodes — which is also what makes a mixed-strategy tree expressible.

**Every deferred cost answered with a number.** ALS ~25ns, policy dispatch ~40ns, plan cache as above, and stitching on 100 parents at +247µs (SQLite) / +328µs (Postgres) — labelled an *upper bound*, because most of that is the second query rather than the stitch.

**Column selection**, swept across all thirteen operations on both dialects rather than the single `findMany` assertion that existed. No `select *`, no `returning *`, no qualified `u.*`; the default set is every scalar in schema order for reads *and* a write's `RETURNING`; `select` narrows the emitted SQL, with the include-adds-a-key-column exception pinned at exactly one extra column.

## Things I got wrong, and what they cost

Four in the harness, because a performance suite that is wrong is worse than none:

1. **`execute` came out larger than `total`** — impossible for a subset. It was timed on a second `SQL` instance rather than the one `$exec` resolves, so the two had different prepared-statement state. A decomposition that does not sum is worse than no decomposition.
2. **The per-call overhead table reported 42ns for four different things** — not four measurements agreeing, one `performance.now()` tick divided by one. Sub-microsecond work is now timed in batches of 2000.
3. **The positional measurement produced a confident +15%, then −9%.** A benchmark readable either way is worse than none, so that table reports p50 *and* p95 at 7× the samples, and the conclusion is "unresolvable" rather than a number.
4. **The bench's baseline row called `buildRowShaper`**, which returns the compiled shaper once it became the default — so it compared the winner against itself and reported +1%. It names `buildInterpretedShaper` explicitly now, and additionally asserts the default *is* the compiled path.

Two more in the test I wrote for deliverable 6: it first asserted no `*` anywhere and failed on `count(*)` — an assertion wider than the guarantee, with a comment confidently claiming the compiler emits `count(<pk>)`, which I had not checked. And the include-exception test registered two models then compiled a case with no relation, going through the motions of the exception without exercising it.

## Where these numbers should not be trusted

- **Ratios below 1.00× are noise, not a win.** gemi cannot beat hand-written SQL doing the same work — it *is* that driver call plus overhead. The Postgres rows land at 0.98–1.06× because a ~150µs loopback round trip varies by more than the difference being measured.
- **Postgres was measured over loopback**, which understates round-trip cost — and round-trip count is the entire case for deliverable 2. The plan requires a real socket before that win can be claimed, and this is not one. It is a floor on the gap, not a measurement of it.

## Verification

| | |
| --- | --- |
| `vitest run orm/` | **513 pass** |
| `vitest run` (packages/gemi) | 7 fail — the same pre-existing router/i18n files |
| template, SQLite | 276 pass, 4 skipped |
| template, Postgres 16 | 351 pass |

## Not in scope

MySQL tuning, query-plan analysis, connection-pool tuning, and result caching — per the plan. Also not done: a CI regression guard for the ratios. The plan floats it and warns against a flaky perf test becoming a reason to ignore CI; with p95 near double p50 on the Postgres scenarios, a threshold tight enough to catch a regression would flake, so it wants either a much longer run or a dedicated machine. Worth its own decision.

## For the reviewer

- `bun run app/models/bench/run.ts` regenerates the report; `bench/shaper.ts` is the variant comparison, so the 3.4× is re-runnable rather than remembered.
- Postgres numbers need two runs, one per generated Prisma client, since a client speaks one provider. The `.json` sidecar carries the other dialect's Prisma column forward and marks it.
- The claim I would check hardest is deliverable 2's negative. It rests on depth-3 not being slower than depth-2 on loopback Postgres, which is exactly where the measurement is weakest.


templates/saas-starter/app/models/bench/run.ts
```diff
@@ -0,0 +1,892 @@
...
+    "- **Ratios below 1.00× are noise, not a win.** gemi cannot be faster than",
+    "  hand-written SQL doing the same work; it *is* the same driver call plus",
+    "  overhead. The Postgres point read and depth-2 include come out at 0.82×",
+    "  and 0.94× because a ~150µs loopback round trip varies by more than the",
+    "  difference being measured, and because the `raw` baseline is a second",
+    "  `SQL` instance with its own prepared-statement state. Read them as \"at",
+    "  the floor\", not as better than it.",
+    "- Postgres was measured over loopback, so every Postgres round trip here is",
+    "  optimistic — see the note below.",
+    "- Scenario 4 (depth-3 include) is **not implemented yet**, and it is the one",
```
nstfkc: **The report contradicts its own headline finding, and regenerating reproduces it.**

This line is hardcoded into the generator, so it is in the committed artifact at `plans/orm/benchmarks.md:156`:

> - Scenario 4 (depth-3 include) is **not implemented yet**, and it is the one the lateral strategy would be judged on.

But scenario 4 *is* implemented — right below at line 511 — its numbers are in the table (`benchmarks.md:25` and `:34`), and conclusion 3 at `benchmarks.md:119` draws the whole deliverable-2 decision from them.

So the document simultaneously reports scenario 4's measurements, declines an optimisation on their strength, and tells the reader under "Where these numbers should not be trusted" that the deciding scenario does not exist. Whoever comes back to audit "should we build `LATERAL`?" — which is the express purpose of committing this file — hits that caveat and stops.

One line to delete. Flagging it as more than a typo only because this artifact *is* the deliverable, and the contradiction lands precisely on the claim it exists to support.

templates/saas-starter/app/models/bench/run.ts
```diff
@@ -0,0 +1,892 @@
...
+    results.push(
+      await scenario({
+        name: `4. depth-3 include, ${PARENTS} parents`,
+        dialect,
+        rows: PARENTS,
+        runs: 30,
+        raw: async () => {
+          // Three round trips by hand, matching what the batched planner does,
+          // so the floor is the same shape rather than a straw man.
+          const parents: any = await database.sql.unsafe(
```
nstfkc: **Scenario 4's baseline runs on a different connection from every other scenario's — which is the same defect class as harness bug #1, and it lands on the comparison the lateral decision is made from.**

`raw` is a second client (`const raw = new SQL(gemiUrl)`, line 389). Scenarios 1, 2 and 3 baseline through it. Scenario 4 alone uses `database.sql.unsafe(...)` — the connection `$exec` itself resolves — here and at the two calls below.

That is the inverse of the bug you already fixed: *"it was timed on a second `SQL` instance rather than the one `$exec` resolves, so the two had different prepared-statement state."* The consequence is the same, one level up: the `×raw` column for scenario 4 is not computed on the same footing as scenario 3's, and **comparing scenarios 3 and 4 is exactly what declines deliverable 2.**

There is an unexplained anomaly in the committed numbers that this would account for. Depth 3 does strictly more work than depth 2 — three queries against two — yet on SQLite both columns go the wrong way:

```
sqlite  3. depth-2 include   raw 170.2µs   gemi 411.0µs   2.41×
sqlite  4. depth-3 include   raw 148.5µs   gemi 286.2µs   1.93×
```

More work, less time, on both sides. On Postgres the raw column behaves (573.3 → 588.7), which is what makes the SQLite pair stand out rather than read as ordinary noise.

I cannot prove causation without re-running, and I am not claiming the conclusion is wrong — the Postgres depth-2/depth-3 pair (0.95× and 0.96×) does support it, and the structural argument (one query per include *node*, not per row) is sound on its own and verifiable from the planner. What I am claiming is narrower: the one measurement singled out as decisive is the one measured differently from its comparator, and the plausible direction of that bias — a warm shared connection making scenario 4 look faster — is the direction that favours the conclusion.

Switching these three calls to `raw.unsafe` and re-running should settle it in a minute, and would let the decision rest on numbers gathered the same way.

packages/gemi/rfc/req.ts
```diff
@@ -0,0 +1,7 @@
+class XRequest extends Request {}
```
nstfkc: This file has nothing to do with iteration 7 — a seven-line sketch of a request/router idea, referencing `ApiRouter` and `XController`, neither of which is imported or defined. It looks like it was picked up by an unstaged `git add`.

Worth noting *why* nothing caught it: `rfc/` is outside the tsconfig `include`, which #50 established and deliberately kept that way for exactly this kind of sketch. That exclusion is defensible, and this is its cost — a file referencing undefined identifiers lands in `packages/gemi` and neither `tsc` nor `oxlint` has an opinion.

If it is a real note-to-self it belongs in a commit of its own; otherwise drop it from this PR.

nstfkc (commented):
Reviewed `60dccdc`. Declining an optimisation on the strength of a measurement is the right kind of result, and the four self-reported harness bugs are exactly the ones that matter — a decomposition that does not sum, and a baseline that compared the winner against itself, are both the sort of thing that makes a perf suite worse than none.

| Verified at `60dccdc` | |
|---|---|
| `vitest run orm/` | **513 pass** |
| `vitest run app/models/` (SQLite) | **276 pass, 4 skipped** |
| `tsc --noEmit` | clean |

**The generated shaper holds up under the attacks I could think of.** `@map` column names are arbitrary strings, so I fed the codegen a column containing `"`, `\`, a newline, a backtick and `${x}`, plus a relation key colliding with a scalar field name. Compiled and interpreted agree on all of them, and `JSON.stringify` closes the quoting question properly. The equivalence suite is well chosen too — `@map`, both dialects, Bytes/DateTime/Json/BigInt, missing columns, empty sets, and key order pinned via `JSON.stringify` rather than deep-equal. Asserting that compilation is *actually available* in CI is the detail that keeps the whole suite from silently comparing the interpreted shaper to itself.

You asked for deliverable 2's negative to be checked hardest. I did, and **two things there need fixing before the decision rests on it** — details inline. Neither says the conclusion is wrong; both say the evidence does not yet carry it.

Also inline: a stray file unrelated to this iteration.

Minor, no comment needed — the PR body's summary numbers do not match the committed table. The body says depth-2 and depth-3 "sit at ~1.1× hand-written SQL"; `benchmarks.md` says 0.95× and 0.96×. The body says the 1000-row SQLite read went to 1.20×; the table says 1.16×. The table is the artifact with provenance, so it should probably be the one quoted — or the body should say which run it is quoting.

One footnote I cannot resolve: on my first run of `vitest run app/models/` I got `1 failed | 275 passed | 4 skipped`, and I did not capture which test. Three subsequent runs were clean at 276/4. Recording it because an intermittent failure in a suite that just grew timing-adjacent code is worth knowing about, not because I can point at it.

nstfkc:
**You were right to push on this, and correcting the evidence reverses the conclusion.** `9999b71`.

## The connection defect, and what it exposed

Fixed — scenario 4's baseline now runs on `raw` like every other. But with the connections equal, SQLite's raw column *still* went the wrong way: 213µs at depth 2 against 194µs at depth 3, for strictly more work.

Your "more work, less time" observation was the tell, and the cause was a level deeper than the connection: **scenario 4 was not measuring a depth-3 include at all.** The seed created accounts with no `organizationId`, so every foreign key at the third level was null, the batched loader correctly skipped that query entirely, and the scenario measured depth-2 plus a filter pass.

Seed now creates 20 organisations and spreads accounts across them, so the third level has a realistic `in` list.

## Which reverses the recommendation

On Postgres, with the seed fixed, the cost is linear in include depth:

```
100 parents, no include    286µs     1 round trip
depth-2 include            690µs     2 round trips
depth-3 include            911µs     3 round trips
```

~220µs per level against a 180µs point read — each level is very nearly one whole round trip, which is exactly what a lateral join collapses to one. gemi is at 0.96× and 1.00× hand-written SQL, so there is nothing to win at this *shape*; the win is in changing the shape.

Ceiling is roughly 350–450µs against 911µs — about 2×, growing with depth. **That makes deliverable 2 the largest unclaimed win in the ORM, not a decline.** The report now says so, and says why the earlier version said the opposite.

It stays unjustified on SQLite, which matches the plan's own note — and I'll add that the SQLite depth-2/depth-3 pair is not even consistently ordered, since scenario 4 runs after scenario 3 and benefits from a warmer cache. That difference is inside the noise either way.

## The other two

**The self-contradicting caveat** is deleted. You are right that it is more than a typo: anyone auditing "should we build LATERAL?" — the express purpose of committing that file — would hit it and stop.

**`rfc/req.ts` is untracked again and left on disk.** Not mine; a `git add packages/gemi` in the shaper commit swept it up. Your note on *why* nothing caught it is the useful part — `rfc/` is outside the tsconfig by the decision made in #50, so a file referencing undefined identifiers had nothing checking it. That exclusion is still right; this was its cost.

**The body/table mismatch** was the body quoting an earlier run. The table is the artifact with provenance, so I've stopped putting run-specific figures in prose — I'll update the PR description to point at the file rather than restate it.

## On the intermittent failure you saw

Recorded, and I have not reproduced it in several full runs either. The suites most likely to do that are the Postgres-shared-database ones under `--no-file-parallelism`, and the new timing-adjacent code in the `@updatedAt` test uses a 5ms sleep to force a distinguishable timestamp — that is a real candidate for flaking on a loaded machine, and it is in iteration 8's branch rather than this one. I will make it deterministic there rather than leave a sleep in a test.

## The lesson I'm taking

This is twice now that a benchmark of mine was not measuring what its name said, and both times it was visible in the committed table as a number going the wrong direction. I published it anyway, twice, without asking why. A benchmark that is wrong is worse than none precisely because it produces a confident recommendation — and this one produced a decision to *not build* the largest available win.

| Verified at `9999b71` | |
|---|---|
| `vitest run orm/` | 513 pass |
| template, SQLite | 276 pass, 4 skipped |

templates/saas-starter/app/models/bench/run.ts
```diff
@@ -0,0 +1,939 @@
...
+    "   is worth stating plainly: scenario 4 was not measuring a depth-3 include.",
+    "   The seed created accounts with no `organizationId`, so every foreign key",
+    "   at the third level was null, the batched loader correctly skipped that",
+    "   query entirely, and the scenario measured depth-2 plus a filter pass. It",
+    "   read as *faster than* depth-2 — more nodes, less time — which was the",
+    "   tell, and it was the number the recommendation rested on.",
+    "",
+    "   With the seed fixed, the Postgres cost is linear in include depth:",
+    "",
+    "       100 parents, no include    286µs     1 round trip",
```
nstfkc: **These three figures — and the two ratios derived from them — appear nowhere in the data the report renders above them.** They are hardcoded prose, so this is the same root cause as the "not implemented yet" caveat you just deleted: narrative in the generator that does not track its own results.

What the conclusion says, against what the committed tables say for the same quantities on Postgres:

| | conclusion | report's own tables |
|---|--:|--:|
| 100 parents, no include | 286µs | **231.1µs** (stitching table) |
| depth-2 include | 690µs | **660.0µs** scenario 3 — or 595.4µs in the stitching table |
| depth-3 include | 911µs | **1057.0µs** scenario 4 |
| point read | 180µs | **123.3µs** scenario 1 |

And the derived claims ride on them:

- *"~220µs per level"* — from the tables it is 231 → 660 → 1057, so **~410µs per level**, nearly double.
- *"against a 180µs point read — so each level is very nearly one whole round trip"* — against the real 123µs point read, each level is more than **three**.
- *"roughly 350–450µs against 911µs — call it 2×"* — the denominator is 1057µs, and the ceiling estimate is built on the 286µs no-include figure that is really 231µs.

**The recommendation survives all of this, and is strengthened by it**: the true per-level cost is about double what the report claims, so the case for collapsing levels into one lateral statement is better than argued. That is the only reason I am not calling this blocking. But the artifact exists so the decision can be audited, and right now an auditor who checks the arithmetic against the table two screens up finds none of it reconciles — on the one conclusion the whole iteration turns on.

The fix that stops this recurring is to derive these from `results` the way the tables are, rather than writing them out — a lookup by scenario name and dialect, interpolated. Three instances of stale hardcoded narrative in two commits is the argument for it.

nstfkc (commented):
Reviewed `9999b71`. Following the thread past the connection defect to *"the third level had no rows to load"* is the right instinct — a null foreign key meaning the batched loader correctly skips the query, so the scenario measured depth-2 plus a filter pass, is a much better explanation than mine, and it is the one that actually explains "more work, less time." Reversing a published recommendation on the strength of it, and saying in the report why the earlier version said the opposite, is the correct handling.

**The anomaly is genuinely resolved.** Depth-3's raw baseline now exceeds depth-2's on *both* dialects, which is the ordering the work implies:

```
sqlite    d2 raw 146.4µs   d3 raw 234.3µs
postgres  d2 raw 685.3µs   d3 raw 1043.0µs
```

Both fixes verified in the tree: scenario 4's three baseline calls now use `raw` like every other scenario, the seed creates 20 organisations and spreads accounts across them, and the "not implemented yet" caveat is gone.

| Verified at `9999b71` | |
|---|---|
| `vitest run orm/` | 513 pass |
| `vitest run app/models/` (SQLite) | 276 pass, 4 skipped |

**But the conclusion that now carries the reversal is argued from numbers that are in none of the report's tables** — inline. The recommendation survives it (the real figures make the case *stronger*), so this is about the evidence, not the call.

Also: `packages/gemi/rfc/req.ts` is still tracked. `git ls-tree 9999b71 packages/gemi/rfc/` lists it, identical to `60dccdc`, and it is still in `gh pr diff 52`. The commit message says it was "removed from the branch and left on disk" — the working-copy half happened, the `git rm --cached` half did not.

On the flake I mentioned last time: it recurred once on my first run of this commit, then passed 6/6 when I ran the suite in isolation. Two failures in ~14 runs, never reproducible on demand, identity still not captured. I would not chase it on my evidence, but I would not call it noise either — recording it in case it turns up in CI.

nstfkc:
Both fixed in `5ec0f6a`, and you were right that the arithmetic did not reconcile.

**The hardcoded prose was the third instance of one root cause**, so I stopped rewriting numbers and removed the ability to write them. Every figure in conclusion 3 is now read out of `results` through a `measured()` lookup — the same data the tables render — and it *throws* rather than emitting a placeholder if a cited scenario is missing, because a conclusion quoting an unmeasured figure is precisely the failure being prevented.

**But the estimate was worse than stale — it was invented.** Checking your table against mine, I found "roughly 2×" came from `depth3 / depth2`, which is the ratio to *two* round trips, not to a lateral's one. That denominator does not correspond to anything the deliverable would produce.

A lateral replaces the whole tree with a single trip, so the win is bounded rather than estimable:

```
pessimistic   depth3 / depth2      if it saved only one of the two extra trips
optimistic    depth3 / noInclude   if carrying the extra rows were free
```

The report quotes that interval now, and says why a single figure is not available: the true cost includes `json_agg`'s server-side work and a wider result to parse, neither of which this suite can measure without building the thing. The interval is wider and less flattering than the number I invented, which is rather the point.

Your read stands — the recommendation survives and is stronger than the first version argued, since each added level costs on the order of a whole round trip.

**`rfc/req.ts` is untracked now, for real.** Thank you for checking `git ls-tree` rather than taking the commit message. The working-copy half happened; the `git rm --cached` half was undone by my own `git add` in the same commit, so the claim was false when I made it. `git ls-tree HEAD packages/gemi/rfc/` no longer lists it.

**On the flake:** two failures in ~14 runs with no captured identity is more than I can dismiss. The most likely candidate was a 5ms sleep in iteration 8's `@updatedAt` test — already replaced there with an assertion that the stored value equals the returned one, no timing dependency. That is on `feat/orm-08-eloquent-doorway`, so it would not have affected your runs of *this* commit. Since it recurred here, that rules it out and I have no better candidate; recording it as open rather than closed.

If it surfaces again, the thing worth capturing is `--reporter=verbose` output, since the shared-Postgres suites and the plan-cache assertions in `policies.test.ts` are the two places where cross-test state could bite.

templates/saas-starter/app/models/bench/run.ts
```diff
@@ -0,0 +1,1061 @@
...
+    `   is worth stating plainly: scenario 4 was not measuring a depth-3 include.`,
+    `   The seed created accounts with no \`organizationId\`, so every foreign key`,
+    `   at the third level was null, the batched loader correctly skipped that`,
+    `   query entirely, and the scenario measured depth-2 plus a filter pass. It`,
+    `   read as *faster than* depth-2 — more nodes, less time — which was the`,
+    `   tell, and it was the number the recommendation rested on.`,
+    ``,
+    `   With the seed fixed, the Postgres cost is linear in include depth:`,
+    ``,
+    `       100 parents, no include    ${us(noInclude)}\t1 round trip`,
```
nstfkc: **Now that these figures track the data, compare the last two runs of identical code.** The only change to `run.ts` between `9999b71` and `5ec0f6a` is this reporting rewrite — scenario definitions, `runs: 30`, seed and baselines are untouched (I diffed for it). So this is pure run-to-run variance on Postgres:

| | `9999b71` | `5ec0f6a` | |
|---|--:|--:|--:|
| depth-2 raw | 685.3 | 600.5 | −12% |
| depth-2 gemi | 660.0 | 735.9 | +11% |
| depth-3 raw | 1043.0 | 753.9 | **−28%** |
| depth-3 gemi | 1057.0 | 753.0 | **−29%** |

And the decision variable — what the third level actually costs:

```
9999b71   660.0 -> 1057.0   = +397µs
5ec0f6a   735.9 ->  753.0   = + 17µs
```

A 23× swing on the quantity the recommendation turns on. In *this* run the third include level costs 17µs against a 146µs point read, so the conclusion's own sentence — "each level costs on the order of a round trip" — is not supported by the depth-2→depth-3 step it is rendered beside. The linear reading comes entirely from the first step (236 → 736), with the second step contributing almost nothing.

It also propagates into the interval: the pessimistic bound is `depth3 / depth2` = 753/736 = **1.02×**, which is why the range opens at "1.0×" — a floor of *no win at all*. The report is being honest, and what it is honest about is that this run cannot distinguish "large win" from "none".

I want to be clear this is not a criticism of the change — it is the change working. The hardcoded prose was concealing this; deriving it surfaced it in one commit.

Two things would settle it, and the second is the one I would actually do:

1. Raise `runs` for scenarios 3 and 4 and report the depth-2→depth-3 **delta** with a spread rather than two independent p50s. The report already notes p95 ≈ 2× p50 on Postgres, which is the same instability seen from the other side.
2. Stop inferring round trips from wall clock. The count is **deterministic** — one query per include node, which the planner guarantees and a statement counter can assert — and the per-round-trip cost is what scenario 1 measures directly. "3 RTTs → 1 RTT at ~146µs each" is a stable argument built from a deterministic count and the single most reliable number in the suite, and it does not move 29% between runs. The end-to-end delta is the noisiest possible way to measure the one thing that is not actually uncertain.

nstfkc (commented):
Reviewed `5ec0f6a`. Both items are properly closed, and you found a third that I had missed: the "roughly 2×" ceiling came from `depth3 / depth2`, which is the ratio to *two* round trips rather than to a lateral's one. I checked the arithmetic and not the denominator's meaning — quoting a bounded interval instead, and saying why a single figure is not available without building the thing, is more honest than what I asked for.

Deriving the prose through `measured()` — throwing rather than emitting a placeholder when a cited scenario is missing — is the right shape. `rfc/req.ts` is untracked at this commit; I verified with `git ls-tree`.

**First, a retraction.** I twice reported an unidentified flaky failure in the template suite and said I could not call it noise. I caught it this time: `generated.test.ts > running prisma generate again produces a zero-line diff`, and it is **my environment, not this PR**. My review worktrees symlink `node_modules/.bin` at the main checkout, so `prisma generate` runs a generator binary built from work *newer* than this branch — the regenerated file adds `type ExecOptions` to an import that does not exist here. First run rewrites the files and fails; every later run sees them already rewritten and passes, which is exactly the pattern I saw. Nothing for you to chase, and I should not have raised it twice without pinning it down.

**The real finding is that `measured()` immediately earned its keep — by exposing that the data underneath the conclusion is not stable.** Details inline. The recommendation may well still be right; it is the reproducibility of the evidence that is now the open question.

| Verified at `5ec0f6a` | |
|---|---|
| `vitest run orm/` | 513 pass |
| `vitest run app/models/` | 276 pass, 4 skipped (after the environment artifact above) |
| `rfc/req.ts` | untracked ✓ |

nstfkc:
`6731fc8`. This is the best finding on the PR, and your suggestion #2 is the one I took.

**You're right that the deciding quantity is not reproducible**, and right that `measured()` is what surfaced it — the hardcoded prose had been concealing a 23× swing on the number the recommendation turned on. In the second run the third include level cost 17µs, which contradicts "each level costs on the order of a round trip" in the sentence rendered directly beside it.

**So the argument no longer rests on wall clock.** Round trips are counted now, through the same container-wrapping seam `differential.ts` uses: 1, 2 and 3 statements for no-include / depth-2 / depth-3, identical on both dialects and immune to load. Combined with scenario 1's point read — the suite's steadiest number — the case becomes "3 statements collapse to 1, at the measured cost of one", where every input is either a count that cannot drift or the most stable timing available.

Your framing is what made it obvious: *the end-to-end delta is the noisiest possible way to measure the one thing that is not actually uncertain.* I had been treating a deterministic property as an empirical one.

The report also now states what the wall clock does *not* support, and renders the two `×raw` ratios rather than characterising them — an earlier draft called them "~1.0× at both depths" when they were 1.25× and 1.02×.

**And I wrote run-dependent prose again while fixing run-dependent prose.** A sentence observing that depth-2 had come out slower than depth-3 was true of the run in front of me and false of the next. It renders from the data now, direction included. Four instances of one mistake in this file — the lesson isn't "be careful writing numbers", it's "don't write them".

**On the flake: thank you for pinning it, and I reproduced it on myself in this commit.** `dist/bin/orm-generator.js` is gitignored and mine was built from iteration 8's branch, so `prisma generate` during benchmark prep rewrote `models.ts` with iteration 8's output — `type ExecOptions`, `wrap`, the declaration-merging interface — and `generated.test.ts` correctly failed on the diff. Same mechanism you described, arrived at from the other side. That is a real hazard of a gitignored build artifact shared across worktrees, and worth knowing about independently of either of our runs; a `build:bin` before `prisma generate` is the discipline.

| Verified at `6731fc8` | |
|---|---|
| `vitest run orm/` | 513 pass |
| template, SQLite | 276 pass, 4 skipped |
| statement counts | 1 / 2 / 3, both dialects |
| `tsc --noEmit` | clean |

nstfkc (commented):
Reviewed `6731fc8`. This is the right resolution, and I checked the part that matters by running it rather than reading it.

**The counts reproduce exactly.** I ran `bun run app/models/bench/run.ts` on SQLite in a clean worktree and got the committed table back, cell for cell:

```
| sqlite | no include      | 1 |
| sqlite | depth-2 include | 2 |
| sqlite | depth-3 include | 3 |
```

Which is the whole argument for the change: the timed quantity moved 23× between two runs, and this one came back identical from a different machine state. The Proxy seam is the right way to get it, too — counting through the container the way `differential.ts` does means it cannot drift from what the runtime actually issues, and hitting `pool.begin is not a function` with a hand-written stub is a good reason to have reached for a Proxy rather than a stub.

**Every figure in conclusion 3 now reconciles with the tables.** I checked each one: point read 155.0 → "155µs", depth-2 591.0 → "591µs" at 1.05×, depth-3 802.6 → "803µs" at 1.00×, removable cost 2 × 155 = 310µs against 803µs. The ratios are rendered rather than characterised, and the sentence about what the wall clock does *not* support is rendered too — including its direction, which is the specific thing that went wrong last time.

Using scenario 1 as the price of a round trip is exactly right for this: a single-row point read is nearly pure latency, so it is the cleanest available proxy, and it is the most stable number in the suite.

**On the flake — confirmed from both ends.** You reproduced it by rebuilding `dist/bin/orm-generator.js` from iteration 8; I confirmed the fix by building the generator *inside* the worktree before running, and the template suite came back 276 pass / 4 skipped on the first run, where every previous first run in a fresh worktree had failed. A gitignored build artifact shared across worktrees is a real hazard and neither of us guessed it up front.

| Verified at `6731fc8` | |
|---|---|
| `vitest run orm/` | 513 pass |
| `vitest run app/models/` | 276 pass, 4 skipped, first run |
| round-trip counts | reproduced independently, exact match |

One observation, not a problem. The counting Proxy is installed on the container for the whole run, so every gemi timing now goes through a `get` trap and a `bind` per property access while the `raw` baselines use the bare client. It is on the order of 100ns against a 155µs round trip, so it cannot matter — and the bias runs *against* gemi, which is the safe direction for a ratio the report uses to argue there is nothing left to win at this shape. Worth a line in the provenance section only because this document is otherwise scrupulous about stating what conditions the numbers were taken under.

nstfkc:
`df9c5f2`. Fixed by removing the asymmetry rather than documenting it.

You're right that it cannot move a number, and right that the bias runs the safe direction — but it was an asymmetry in the one document that is otherwise careful about stating its conditions, and scoping it turned out cheaper than the caveat would have been. The wrapper is now installed only for the duration of each counting call and removed in a `finally`, so no ratio in the report divides a proxied measurement by an unproxied baseline. Counts unchanged and still exact.

The provenance section states it anyway, alongside the two-runs-per-Prisma-client condition that was also unstated.

**Thank you for reproducing the counts independently.** That was the claim worth checking, and "identical from a different machine state" is precisely the property the timed version did not have. Confirming the generator fix from the other end — building inside the worktree and getting a clean first run where every previous first run had failed — closes that one properly too.

For what it's worth, this PR's review found six things I had wrong, and the two that mattered most were both cases of me publishing a confident number without asking why it pointed the wrong way. The scenario-4 seed and the 23× swing were each visible in the committed table before you flagged them.

PR #53, opened 2026-07-27
ORM iteration 8 — the Eloquent doorway, and an ORM-backed auth adapter

nstfkc:
The first step past the POJO baseline, without breaking it. Three deliberate levels, each opt-in, and **no return type or signature changes**.

Plus the ORM-backed auth adapter, which is the piece `plans/orm/README.md` schedules for "once iteration 4 lands writes" and the thing that lets `auth/adapters/prisma.ts` be deprecated.

Stacked on #52 (merged).

## The three levels

| Level | What you get | What it costs |
| --- | --- | --- |
| POJOs *(default)* | Plain objects, full `select` narrowing | Nothing |
| `track` + `save` | Mutate and persist, only changed columns | ~+100% on a 1000-row SQLite read; provenance tied to object identity |
| `wrap` | Methods on a record | Requires a complete row |

`docs/orm-rows-and-entities.md` frames them as deliberate levels rather than one unfinished abstraction, with the out-of-scope position stated plainly: no identity map, no unit of work, and *half* an identity map is worse than none.

## `track` and `save`

```ts
const user = await User.findUnique({ where: { id } }, { track: true })
user.name = "New name"
await User.save(user)      // update "User" set "name" = ?, "updatedAt" = ? where "id" = ?
```

Diffs against the values that came back and writes only what changed. An unchanged row is a no-op — no statement, no `@updatedAt` stamp.

**`track` is a second parameter, not a key in `args`.** Intersecting Prisma's arg types with a gemi-specific key would mean `Prisma.UserFindManyArgs` no longer describes what the operation accepts. It also keeps the flag away from the compiler, so it cannot reach the plan key — which it must not, since it does not change the SQL. A test asserts the args type stays unintersected.

There is a test asserting a tracked result is `toEqual` an untracked one with the same keys, because *that* — not `save` itself — is invariant 5's actual claim.

**The cost, since criterion 1 asks for the number:** +~100% on a 1000-row SQLite read, +~40% on Postgres. That more than doubles a SQLite read, which is the argument for off-by-default and makes `track: true` on a large read a real decision. The default path is unchanged.

**Provenance is tied to object identity.** A row that is spread, cloned or JSON round-tripped is a different object with none, and `save` raises rather than guessing — the guess would be "write every column", and writing a column never fetched is how a partial `select` silently reverts data.

Everything goes through `$exec("update", …)` rather than around it, so policies scope it, `@updatedAt` stamps it, an open transaction contains it, and two saves touching the same columns share one cached plan.

## `wrap`, and why completeness is the design

```ts
export class User extends UserModel {
  get displayName() { return this.name ?? this.email ?? "anonymous" }
}

const user = User.wrap(await User.findUniqueOrThrow({ where: { id } }))
await user.save()
```

`wrap` requires a complete row, and a `Pick` is a **compile** error — criterion 6 asks for a type test, so it is `wrap.test-d.ts` under `vitest --typecheck`, where the assertions are the type checker's.

That constraint is what lets instances and `select` coexist: `displayName` reads `this.email`, so it is safe exactly as long as `wrap` cannot be handed a row that did not fetch `email`. It is also why `wrap` needs no `track` — a complete row is one that can be snapshotted in full, so the type constraint and the save capability are the same constraint rather than two.

### Two gaps the type test caught, both of which would have shipped

1. **`this.name` did not exist inside a subclass method.** The first version intersected the columns into `wrap`'s *return* type only, which typed the call site correctly and left the class body blind — a wrapper whose methods cannot read the row is not a wrapper. Fixed with declaration merging against Prisma's own payload type, so the two cannot drift.
2. **The return type dropped `include`d relations**, because it intersected with the fixed scalar payload rather than the argument.

Then oxlint flagged the declaration merging as unsafe — correctly, since TypeScript does not check merged properties are initialised, so `new UserModel()` would type as carrying every column while holding none. **I closed the hazard rather than suppressing it:** `Model`'s constructor is `protected`, so `wrap` is the only way to get an instance. That forced `wrap`'s `this` constraint from `new (...args: never[]) => any` onto `{ prototype: unknown }`, since a protected constructor cannot satisfy a public construct signature.

## `ActiveRecordModel` confirms invariant 3 paid

Every query returns instances, achieved **purely** by overriding `$shape` — no changes to any of the twelve operations, no per-operation branching, and `count` / `updateMany` fall through untouched because they are not rows. Criterion 7 said a failure here would be a finding to report rather than a workaround to write; it was not one.

It ships as a **documented example, not a supported tier**, and its cost is asserted in a test rather than only warned about: hydration is unconditional there, so a partial row produces an instance whose methods silently read unfetched fields. `wrap` refuses that at compile time; this trades the guarantee for ergonomics. The return *types* also do not change, so the methods are invisible at the call site — fixing that means conditional return types across twelve operations, which is the complexity the plain-object default exists to avoid.

## The ORM auth adapter

Twenty-two methods, and **all of them translated with no changes to the ORM.** That is the first evidence the query surface is *sufficient* rather than merely tested, because these were written against a real application's needs rather than against the compiler's known capabilities. What they happen to exercise: `findUnique` with an extra non-unique filter, two different compound `@@unique` keys on one model chosen per call, a nested `connect`, a three-level `select` (session → user → accounts → organization), and `deleteMany`-as-idempotent-delete.

Two behaviours a switching application must know, both asserted:

- **Policies are suspended.** Authentication runs *before* there is an authenticated user, so a policy scoping by `ctx.user` cannot be satisfied — and under deny-by-default it would raise rather than return null, turning "wrong password" into a 500. One `asSystem` wrapper, applied uniformly, so no method can forget.
- **`password` is returned** where Prisma's adapter used `omit`, which the ORM does not implement. Emulating it with a `select` naming every other column would drift as the schema grows. The *session* queries name their columns and exclude it — there is a test asserting `"password" in session.user` is false, since that is the case that reaches a client.

**Not done here:** pointing the template's `app/config/auth.ts` at it. That is a behaviour change to a working application and wants its own decision, not to be smuggled in beside the adapter that enables it.

## Verification

| | |
| --- | --- |
| `vitest run orm/ bin/` | **543 pass** |
| template, SQLite | **316 pass**, 6 skipped |
| template, Postgres 16 | **431 pass** |
| `wrap.test-d.ts` | 6 type assertions, no type errors |

## For the reviewer

- `bun --bun vitest run --typecheck --typecheck.ignoreSourceErrors app/models/wrap.test-d.ts` runs the type test. `ignoreSourceErrors` is needed because the template's own views carry pre-existing errors unrelated to this work.
- **Run `bun run build:bin` in `packages/gemi` before `prisma generate`.** `dist/bin/orm-generator.js` is gitignored, so a stale one built from a different branch will rewrite `models.ts` and fail `generated.test.ts`. That bit both of us on #52.
- The claim I would check hardest is the `protected` constructor. It closes a real hazard, but it constrains how these classes can ever be constructed, and I would rather hear now than after something needs to subclass differently.


packages/gemi/auth/adapters/orm.ts
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  async createUser(args: CreateUserArgs): Promise<User> {
+    // Prisma's adapter uses `omit: { password: true }`, which the ORM does not
+    // implement — `omit` arrived in Prisma 5.16 and iteration 2 built the read
+    // surface without it. Rather than emulate it with a `select` listing every
+    // other column (which would silently drift as the schema grows), the field
+    // is returned and the difference is documented on the class. A caller that
+    // serialises this to a client must strip it.
+    return await this.run(() => this.models.User.create({ data: args }));
```
nstfkc: **Switching to this adapter makes `POST /sign-up` respond with the user's password hash.**

The class comment says callers "that hand a user object straight to a client must not assume the field is absent" — but the framework *is* that caller, in this package:

- `auth/routes.ts:32` — `"/sign-up": this.post(AuthController, "signUp")`, so the handler's return value is the response body.
- `auth/AuthController.ts:362` — `signUp` ends `return newUser;`.
- `newUser` is whatever `createUser` returned. Prisma's adapter returns it with `omit: { password: true }` (`adapters/prisma.ts:68`); this one returns the full row.

So the difference is not a caveat for application authors to absorb — it is a change in what an existing, framework-owned, unauthenticated endpoint sends back. The same object also reaches `config.onSignUp(newUser, …)`, so an application hook that logs its argument now logs a credential hash.

Blast radius today is limited because the template's `app/config/auth.ts` is not pointed at this adapter yet — which you flagged as deliberately out of scope, and I agree with that call. But the adapter is shipped as the thing that lets the Prisma one be deprecated, so this fires on whoever flips that config first, and it will not be visible to them at the call site.

On the reasoning for not narrowing: *"emulating `omit` with a `select` listing every other column would silently drift as the schema grows"* is a fair objection, and it is the same objection that `SESSION_USER` overrides fourteen lines above — naming columns explicitly, for exactly this reason ("what keeps the *session* queries from selecting `password`"). So the principle is already accepted where a client-facing path is at stake; this is the other such path.

The fix that avoids the drift objection entirely is not a `select` at all — the adapter can drop the key from the object it returns. One line, no column list to maintain, no divergence as the schema grows:

```ts
async createUser(args: CreateUserArgs): Promise<User> {
  const user = await this.run(() => this.models.User.create({ data: args }))
  delete (user as { password?: unknown }).password
  return user
}
```

`updateUserPassword` returns the full row too, though I did not find a path that serialises it.

(If the preferred fix is instead to stop `signUp` returning the raw user, that is also fine — but it is a change to `AuthController`, and this PR is where the behaviour changes.)

packages/gemi/orm/provenance.ts
```diff
@@ -0,0 +1,176 @@
...
+ * it, every save of a row carrying a timestamp would rewrite that timestamp.
+ */
+export function changedFields(row: object): Record<string, unknown> {
+  const record = provenanceOf(row);
+  if (!record) return {};
+
+  const source = row as Record<string, unknown>;
+  const changed: Record<string, unknown> = {};
+
+  for (const name of Object.keys(record.snapshot)) {
```
nstfkc, line 119: **Assigning to a field the query did not fetch is silently dropped.** Run against this branch:

```ts
const row = await UserModel.findFirst({ select: { id: true, name: true } }, { track: true })
row.email = "changed@x.test"    // never fetched
row.name  = "Renamed"
await UserModel.save(row)
// name written; email SILENTLY DROPPED — no statement, no error
```

`changedFields` iterates the *snapshot's* keys, so a field outside it cannot be seen as changed. The doc frames the partial case as "can still be saved, it just cannot write what it never read", and that is right for a column the caller never touched — but this is the caller explicitly assigning a value and the write disappearing.

It is the same shape as the divergences this codebase has been refusing all the way up the stack: `save` looks like it worked and did less than it said. And the detection is cheap and precise, because everything needed is already here — a key on the row that is in `schema.fields`, is not a relation, and is *absent from the snapshot* can only mean an assignment to something unfetched:

```ts
for (const name of Object.keys(source)) {
  if (skip.has(name) || !(name in schema.fields)) continue
  if (!(name in record.snapshot)) throw /* names the field and the select that omitted it */
}
```

Refusing it also composes with the existing message style: it can say "`email` was not fetched by the query this row came from, so `save` cannot write it — select it, or update it explicitly."

packages/gemi/orm/Model.ts
```diff
@@ -104,6 +133,106 @@ export abstract class Model {
...
+        `This row came from ${record.model}, not ${schema.name}. Save it ` +
+          `through the model it was read from.`,
+      );
+    }
+
+    const changed = changedFields(row);
+    if (Object.keys(changed).length === 0) return null;
+
+    const updated = await this.$exec("update", {
+      where: record.key,
```
nstfkc, line 200: Two smaller things, both about this being where the row's identity and freshness are decided.

**1. A tracked row selected without its primary key fails, but blames the wrong thing.**

```ts
const row = await UserModel.findFirst({ select: { name: true } }, { track: true })
row.name = "Changed"
await UserModel.save(row)
// UnsupportedQueryError: gemi ORM does not support 'where' yet (User.update).
// update needs a unique field. User declares: id, publicId, …
```

`track` captured `key = { id: undefined }`, `matchUniqueKey` drops undefined members and reports the generic failure. The caller's actual mistake — selecting without the primary key — is never named, and the message points at a `where` they did not write. `untracked()` two files over is a model of how to do this well: both causes named, with the fix. A check here (`record.key` has an undefined member → the select omitted the key) would match that standard, and it can only fire for this one reason.

**2. The in-memory row is stale after `save`.**

```
in-memory row.updatedAt = 2026-07-27T23:23:41.847Z
database    updatedAt   = 2026-07-27T23:23:41.856Z
```

`resnapshot(row, changed)` writes back the values *sent*, so `@updatedAt` — stamped by the compiler, never in `changed` — is refreshed in neither the row nor the snapshot. No incorrect write follows (the row and snapshot still agree, so it stays clean), but a caller reading `user.updatedAt` after a successful save silently gets the old instant.

The fix is free: `updated` is already in hand and is what the database holds. Copying its scalar keys onto the row before resnapshotting makes the row current and keeps the snapshot honest, which is also what makes a subsequent `save` correct if the database rewrote anything else.

nstfkc (commented):
Reviewed `876c3c9`. The three-level framing is the right shape, and `ActiveRecordModel` being *purely* an `$shape` override is a real result — invariant 3 paid, and saying so with a working example rather than an assertion is better evidence than a test would have been.

| Verified at `876c3c9` | |
|---|---|
| `vitest run orm/ bin/` | **543 pass** |
| `vitest run app/models/` (SQLite) | **316 pass, 6 skipped** — first run, no artifact |
| `tsc --noEmit` | clean |
| `oxlint` | `orm/` 0 warnings; `auth/` 2; whole package 88, unchanged from the baseline measured on #50 |

Building the generator inside the worktree first, per your note, gave a clean first run — that hazard is now fully closed on both sides.

**You asked me to check the `protected` constructor hardest. I think it is right, and the reason is asymmetry rather than taste.** Relaxing a `protected` constructor later is a non-breaking change; tightening a public one later is not. So if the choice is uncertain, this is the direction that keeps the option open. I checked it does not box in the tier that needs instances — `ActiveRecordModel` constructs within the hierarchy, so it is unaffected — and the one thing it genuinely costs is an application building a model instance directly for a factory or fixture, which has to go through `wrap` and therefore needs a complete row. Given `wrap`'s completeness constraint is the whole safety argument, making that the only door seems consistent rather than incidental.

**One finding I would treat as blocking**, on the auth adapter rather than the ORM — details inline. Three smaller ones on `save`.

nstfkc:
All four in `7663c96`. The first one is the important one and you're right that it's blocking.

**The credential leak.** I had documented this as a difference for application authors to absorb, and that was the wrong frame: **the framework itself is the caller that serialises it.** `auth/routes.ts` → `AuthController.signUp` → `return newUser` means whatever `createUser` returns is the response body of an unauthenticated endpoint, and `config.onSignUp` receives the same object.

Your fix is the right one and it dissolves the objection I'd made. I argued against a `select` because the column list would drift as the schema grows — deleting the key has no list to maintain. And you're right that the principle was already conceded fourteen lines up: `SESSION_USER` names its columns *specifically* to keep `password` out of a client-facing path. I applied it to one such path and not the other.

`updateUserPassword` gets the same treatment. You noted no path serialises it; I'd rather not depend on that staying true for a method whose entire purpose is setting a password.

Tests assert the field is absent from the return *and* still stored, so the strip is provably outbound only.

**The silently-dropped assignment.** Also right, and it's the class this stack has been refusing the whole way up — `save` reporting success having done less than it was told. Your detection is exact and I used it as written: a key the schema declares as a field, not a relation, absent from the snapshot, can only be an assignment to something unfetched. The message names the field and both ways forward.

I replaced the existing test rather than adjusting it — it had been asserting the silent-drop as correct. It now covers the case that genuinely is legal: touching only fetched columns.

**The missing-primary-key error** now names the actual mistake instead of pointing at a `where` the caller never wrote. Your framing of `untracked()` as the standard to match was the useful part.

**The stale row** is fixed by copying the returned row back, as you suggested — and taking the snapshot from the same source, so a second `save` is still a no-op.

**On the `protected` constructor:** thank you for the asymmetry argument. "Relaxing later is non-breaking, tightening later is not" is a better reason than the one I had, and it's the one I'd give if asked again.

One process note: `git add packages/gemi` has now swept `rfc/req.ts` into my index three separate times. I've caught it each time, but I'm switching to explicit paths rather than relying on catching it.

| Verified at `7663c96` | |
|---|---|
| `vitest run orm/ bin/` | 543 pass |
| template, SQLite | 320 pass, 6 skipped |
| template, Postgres 16 | 439 pass |
| `tsc --noEmit` | clean |

nstfkc (commented):
Reviewed `7663c96`. All four fixed, and I verified each by running it rather than reading the diff.

**The credential leak is closed, and — the part that could have gone wrong quietly — the strip is scoped correctly:**

```
createUser            -> returned has password: false   stored: "HASH123"
findUserByEmailAddress -> password present (correct)
```

That second line is the one I most wanted to see. A `withoutPassword` applied one method too widely would have broken sign-in by removing the hash the verifier compares against, and it would have failed as "wrong password" rather than as an error — so the fix landing on exactly `createUser` and `updateUserPassword`, with the lookups untouched, is the whole of it. Extending it to `updateUserPassword` on "no path today is not a property that stays true" is the right instinct; that is the one I would have left and been wrong to.

Deleting the key rather than naming columns also dissolves the drift objection instead of overriding it, which is a better resolution than the one I proposed.

**The three `save` fixes, each confirmed:**

```
assigning an unfetched column  -> UnsupportedQueryError, naming the field
row fetched without its PK     -> raised from save, not from `where`
updatedAt after save           -> current
```

The unfetched-column message meets the bar `untracked()` set — it names the field, says why the diff cannot place it, and gives both ways forward. Replacing the test that asserted the old silent-drop behaviour rather than adjusting it is the right call: that test was encoding the bug.

| Verified at `7663c96` | |
|---|---|
| `vitest run orm/ bin/` | 543 pass |
| `vitest run app/models/` (SQLite) | 320 pass, 6 skipped |
| `tsc --noEmit` | clean |

One small observation inline, and it is a pattern this stack has already settled once.

packages/gemi/orm/provenance.ts
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * Compared with `!==`, which is right for the scalars an encoder produces and
+ * wrong for a `Date` — two Dates for the same instant are different objects. So
+ * `Date` is compared by instant and typed arrays byte-wise, the same three cases
+ * `sameEncoded` in the write compiler handles, and for the same reason: without
+ * it, every save of a row carrying a timestamp would rewrite that timestamp.
+ */
+export function changedFields(
+  row: object,
+  schema?: ModelSchema,
```
nstfkc: `schema` is optional, and the new guard only runs `if (schema)` — so the unfetched-column check is off for every caller that omits it. This is the same shape as `render`'s `origin?` on #49, where the conclusion was that optional "reintroduces in miniature the hole this check exists to close", and it was made required.

The argument is a little stronger here than it was there, because `changedFields` is **public API** — `orm/index.ts:87` exports it — so an application computing its own diff gets the permissive behaviour, which is precisely the silent drop this commit just refused. Inside the framework there is one caller and it passes the schema, so nothing is broken today; it is the exported surface that carries the old semantics.

Worth noting the 13 call sites in `provenance.test.ts` all omit it, so the guard is exercised only through `Model.save`. Making the parameter required would pull those onto the checked path too, which is where the interesting cases are.

If it stays optional, the thing I would want is a line saying which behaviour the no-schema form has and why a caller might want it — right now the signature reads as "schema is an extra", when what it actually selects is whether `save`'s central safety check runs.

nstfkc:
`1fae711`. Required now — and you're right that this is the third time the shape has come up. `render`'s `origin?` on #49 settled it once; I reintroduced it here from the same instinct that optional is politer.

Your point about it being **public API** is what makes this stricter than #49's case rather than equal to it. An application computing its own diff got the permissive behaviour from a signature reading as though the schema were an extra, when what it selected was whether `save`'s central check ran at all.

The thirteen test call sites are on the checked path now, and **one of them was encoding the bug** — "a field the row never fetched is never written" asserted the silent drop as correct at the unit level, exactly as the integration test I replaced last round did. Same treatment: it asserts the refusal, plus a second test for the legal half where a partial row touches only what it read.

**On the scoping check you ran:** that is the one I had not thought to make, and it is the failure that would have been worst. `withoutPassword` applied one method too widely would have stripped the hash the verifier compares against, and sign-in would have failed as *"wrong password"* rather than as an error — a silent break introduced by a fix for a silent leak. Confirming `findUserByEmailAddress` still returns it is the assertion that matters, and I should have written it myself.

| Verified at `1fae711` | |
|---|---|
| `vitest run orm/ bin/` | 544 pass |
| template, SQLite | 320 pass, 6 skipped |

nstfkc (commented):
Reviewed `1fae711`. Done, and it paid for itself immediately — which is the argument for making these required rather than carefully remembering to pass them.

Pulling the thirteen unit call sites onto the checked path surfaced a **second** test encoding the bug:

```
- test("a field the row never fetched is never written")
+ test("assigning to a field the row never fetched is refused")
```

That is the same defect as the integration test replaced one commit earlier, at the unit level, and it would not have been found any other way — an optional parameter means the tests exercising the interesting cases are exactly the ones opting out of the check. Adding the legal-half test beside it (a partial row touching only what it read) is what keeps the refusal from being over-broad.

| Verified at `1fae711` | |
|---|---|
| `vitest run orm/ bin/` | 544 pass (was 543) |
| `vitest run app/models/` (SQLite) | 320 pass, 6 skipped |
| `tsc --noEmit` | clean |

Three instances of this shape across the stack now — `render`'s `origin`, and this — and both times the optional form was hiding a real case rather than merely being untidy. That is worth more than the individual fixes.

**Nothing outstanding from me on #53.** Everything I raised on this branch is closed: the sign-up credential leak, the three `save` defects, and this. The `protected` constructor I checked separately and still think is the right call.

PR #54, opened 2026-07-28
ORM iteration 9 — the lateral strategy, and two findings that had to land first

nstfkc:
`LATERAL` + `json_agg` on Postgres: one round trip for a relation instead of one per node. **Now the default there.**

This is iteration 7's deliverable 2, which that iteration *declined* on a measurement that turned out to be wrong. It gets its own iteration because the decision needs its own record, and because building it required widening a seam — which iteration 7 said would be a finding rather than a workaround.

Stacked on #53.

## Two findings before a line of strategy was written

**1. The seam could not express it.** Invariant 4 lists three strategies, two of them single-round-trip, and iteration 7 says the lateral form "has to be a sibling `RelationStrategy`". It cannot be: a strategy's entire output is a `load()` that runs *after* the root query plus the field names the root must select for stitching. There is no way to contribute a column expression, a join clause, or a table alias — all three of the things a lateral join needs.

So the seam is genuinely swappable for *"fetch children separately, by some means"* and closed to *"fold children into the root statement"*. Widened additively with `RelationPlan.root`; the batched strategy is untouched and 543 existing tests passed unchanged, which is the evidence that "additive" is a fact rather than an intention.

**2. Policies did not survive the fold, and that was blocking.** The README asserts scoping applies under every strategy because policies rewrite the arg tree before planning. **It did not.** Policies are applied in `$exec` to the model being queried; a nested read acquired its own *only* because the batched strategy recurses through the child's `$exec`. A lateral join has no such call, so the child's policies would never be consulted and the subquery would be unscoped — a cross-tenant read arriving through a new door.

The obvious fix was wrong: letting the strategy resolve policies while building the subquery would give `compile/` knowledge of the registry and the ambient user, at which point it is no longer a pure function of the argument shape and the plan cache stops being sound. Instead nested policies are applied **to the argument tree, in `$exec`, before the plan key** — which makes the README's claim true rather than papering over it.

That landed *before* the strategy, deliberately. The first version of lateral must not be one that silently bypasses policies.

## Three corrections to my own plan document

Written up in `plans/orm/09-lateral-strategy.md`, struck through rather than rewritten, for the same reason the invariant-4 correction sits beneath the original claim.

The doc predicted root-table aliasing would be "the largest part of the work" and "the first iteration that changes emitted SQL for existing queries". Neither is true: **no alias is needed** (Postgres lets a lateral subquery reference the outer table by name), **qualification is conditional** so existing queries emit byte-identical SQL, and **writes need nothing** since `insert … returning` cannot carry a lateral join.

## The strategy

It **declines rather than emitting SQL it cannot get right**, per *node* rather than per query — so a mixed tree reports `["batched", "lateral"]`. Declines: not Postgres; a node with its own `include` or a relation in its `select`; an implicit many-to-many.

**The correctness lives in the decoder.** JSON aggregation flattens types: `json_build_object` renders a `timestamp(3)` as ISO text where the driver gives a `Date`, `bytea` as hex, `bigint` as text. The batched path never sees any of that. A decoder that merely parsed would return strings where the shipped path returns dates — so it is schema-driven, per field, and asserted against a real server.

`coalesce(json_agg(…), '[]'::json)` because `json_agg` over zero rows returns `NULL` and an empty to-many must shape to `[]`. `orderBy` goes *inside* the aggregate, because beside it it would order the single row the aggregate produces.

Values stay parameters: the subquery is built from `Fragment`s and its `where` goes through `compileWhere`. This is the first strategy to emit SQL, and invariant 2 applies to it.

## Selection, and the plan key it requires

**The strategy is part of the plan key.** Two strategies emit different SQL for the same arguments, so without it a query naming one could get a plan compiled for the other — and a batched plan run where lateral was asked for never loads the children at all, because `attachRelations` skips a plan carrying `root`. Wrong results from a caching bug.

`ExecOptions.strategy` is the per-query override, and it is the first option there that *does* reach the compiler. `track` deliberately does not, since it changes no SQL. An option that changes the statement must be in the key; one that does not must not be.

## What justifies making it the default

Thirty relation cases from the differential matrix run against Prisma under lateral and pass. They are **derived from** the batched table rather than listed separately, so the two cannot drift — the whole risk of two strategies is a shape one handles and the other does not. Cases the strategy declines still run, because a fallback returning wrong rows would be the worst outcome available. An assertion guards that the derived list is non-empty, since a filter matching nothing would make thirty cases vacuous.

The mechanism shipped one commit before the default flipped, deliberately: same discipline as measuring before optimising.

## Things I got wrong, caught by tests

- **The strategy did not propagate to nested `$exec`.** A query explicitly asking for `batched` got the default for every level below the root. Results were identical either way, so only a statement count could find it — and the query-count test did, within a minute of the default changing.
- **A decline for `take`/`skip` was dead code**: the *planner* refuses pagination on a relation node before a strategy is consulted, so the tests asserting that fallback asserted an unreachable state.
- **My database test drove both strategies by hand**, and the *batched* side returned `accounts: []` every time — because batched loads children in `attachRelations`, which `$exec` calls and a manual compile does not. It looked like lateral inventing rows when it was the only one loading them.
- **I predicted a declined node costs three statements**, assuming a decline reverts its descendants to batching. It does not — four becomes two. A decline costs a statement, not a subtree.
- **Three read tests asserted qualification while passing `include: undefined`** — no relation means nothing folds, means no qualifier. They asserted a consequence of a premise they had not set up.

## Two tests were asserting the default rather than a strategy

Pinning them is the fix, not widening them.

`"each relation query is $exec on the related model's class"` is iteration 6's **mechanism**, not its guarantee. Under lateral the child's `$exec` is never entered, and the guarantee survives only because nested policies now reach the arg tree. There is a test asserting that absence directly, beside the batched one, so nobody reads the first and concludes the recursion *is* the guarantee.

## Verification

| | |
| --- | --- |
| `vitest run orm/ bin/` | **591 pass** |
| template, SQLite | **324 pass**, 19 skipped |
| template, Postgres 16 | **489 pass**, lateral as default |
| differential matrix under lateral | 30 cases vs Prisma |

## Not in scope

SQLite's `json_group_array` — its round trips are in-process and cost tens of microseconds; nothing measured asks for it. Folding a *whole tree* into one statement (a node with its own include still declines, which costs one statement rather than the subtree). MySQL.

## For the reviewer

- The claim I would check hardest is the decoder. It is the only place where lateral and batched can disagree about a *value* rather than about SQL, and the template's schema exercises `DateTime` but not `Bytes` or `BigInt` through a relation.
- `bun run build:bin` in `packages/gemi` before `prisma generate`, per #52.


packages/gemi/orm/compile/lateral.ts
```diff
@@ -0,0 +1,322 @@
...
+      // ISO text out of `json_build_object`, where the driver would have given a
+      // `Date`. Prisma returns a `Date`, so this is not optional.
+      return (value) =>
+        value === null || value === undefined
+          ? null
+          : value instanceof Date
+            ? value
+            : new Date(String(value));
+
+    case "BigInt":
```
nstfkc, line 295: **`BigInt` loses precision through the lateral path, silently, above 2^53 — and no decoder here can fix it.**

Against live Postgres 16, one row, one `bigint` column on an included relation:

```
counter   batched = bigint:9007199254740993      lateral = bigint:9007199254740992
```

`9007199254740993` is 2^53 + 1. `json_build_object` renders a Postgres `bigint` as a **JSON number**, so by the time this converter runs the value has already been through a double — `BigInt(String(value))` is faithfully converting an already-wrong number. The batched path is exact because the driver types the column itself.

So the fix has to be server-side, in the projection rather than the decoder: emit the column as text inside the JSON, e.g. `'counter', "Widget"."counter"::text`, at which point this converter's `BigInt(String(…))` is already correct. Only `BigInt` needs it — `Int` is inside the safe range by construction.

Worth noting the comment above says "`bigint` becomes a number or a string depending on magnitude". That is the behaviour of some *drivers*; `json_build_object` is unconditional, so the string case never arrives and the number case is always lossy past 2^53.

Severity, stated plainly: this is a wrong value, silently, on the path that just became the default for Postgres, for a type Prisma returns exactly. It is unreachable from the template's schema, which is why the thirty derived cases pass.

packages/gemi/orm/compile/lateral.ts
```diff
@@ -0,0 +1,322 @@
...
+      return (value) =>
+        value === null || value === undefined ? null : BigInt(String(value));
+
+    case "Bytes":
```
nstfkc, line 317: **`Bytes` comes back as a different type under the two strategies.** Same live-server run:

```
digest    batched = Buffer:1,2,255        lateral = Uint8Array:1,2,255
```

The bytes match, so this is milder than the `BigInt` one — but the *type* is observable and the difference bites the ordinary things a caller does with a `bytea`:

- `Buffer.isBuffer(row.digest)` — true under batched, false under lateral.
- `row.digest.toString("hex")` — `"0102ff"` under batched; under lateral `Uint8Array.prototype.toString` ignores the argument and returns `"1,2,255"`. Wrong answer, no error.
- `.equals()`, `.subarray().toString()` and the rest of the Buffer surface are simply absent.

The hex parsing itself is right; it is the container that differs. `Buffer.from(bytes)` — or `Buffer.from(text, "hex")` directly, skipping the manual loop — would make the two paths return the same type as well as the same content.

Whether `Buffer` or `Uint8Array` is the *desired* answer is a separate question worth settling once: the driver gives `Buffer`, and this codebase's rule is to match what the shipped path already returns, so `Buffer` is the one that keeps `lateral` invisible to callers.

nstfkc (commented):
Reviewed `c5b738d`. I stood up Postgres 16 with the template's schema (`prisma db push` against a copy with the provider flipped) so the lateral path could be exercised for real rather than reasoned about.

**The policy finding was the right thing to land first, and it holds.** This was my main worry — a fold that skips the child's `$exec` skipping the child's policies is a cross-tenant read through a new door. Verified against the live server, with a policied `Account` subclass registered the way an application would:

```
batched  -> 1 accounts, orgs=[1]  (scoped)
lateral  -> 1 accounts, orgs=[1]  (scoped)
```

Applying nested policies to the argument tree in `$exec` — rather than letting the strategy resolve them while building SQL — is also the right call for the reason given: the other way puts the registry and the ambient user inside `compile/`, and the plan cache stops being sound. Making the README's claim true instead of narrowing it is the better repair.

The `PRE_SCOPED` module-private `Symbol` is a good choice too. A `policiesApplied?: boolean` on the public `ExecOptions` would have been precisely the skip-policies door iteration 6 spent three rounds closing.

| Verified at `c5b738d` | |
|---|---|
| `vitest run orm/ bin/` | 591 pass |
| `vitest run app/models/` (SQLite) | 324 pass, 19 skipped |
| `lateral.test.ts` vs real Postgres 16 | 13 pass |
| `tsc --noEmit` | clean |

**You said the decoder is what you would check hardest, and that the template's schema exercises `DateTime` but not `Bytes` or `BigInt` through a relation. That instinct was right — I built a fixture relation carrying those types and found two divergences.** Details inline; one of them is silent data corruption on the new default path.

Separately, on the structural point: the thirty differential cases are *derived from* the batched table, which is the right way to stop the two drifting — but it also means they can only ever cover the types the template's schema has. No number of derived cases reaches `Bytes`/`BigInt` through a relation. That argues for a fixture model with the exotic columns (the way `reading` and `bare` were added in #49 for exactly this reason) rather than more cases.

nstfkc:
Follow-up in `09e09e2`, on the thing I flagged as worth checking hardest.

I said the decoder is the only place lateral and batched can disagree about a *value*, and that the template's schema reaches a relation carrying a `DateTime` but not one carrying `Bytes`, `BigInt`, `Json`, `Boolean` or `Float` — so the thirty-case matrix cannot see those conversions. Building the fixture found **two real bugs**:

**`BigInt` lost precision.** `9007199254740993n` came back as `…992n`. JSON has no integer type, so `json_build_object` renders a `bigint` as a JSON *number* and `JSON.parse` makes it a float64 before any decoder runs — the low bit gone, in a value whose whole reason for being a `BigInt` is that it does not fit a double. Fixed with a `::text` cast in the emitted SQL.

**`Json` came back as a string.** The batched path parses it in `PostgresDialect.decode`; the folded path returned whatever the outer parse gave. Fixed with the same `typeof` check, so the two agree regardless of whether the column is `jsonb` or text-typed.

Both are the class this project keeps refusing — a value quietly wrong rather than a query that fails — and neither was reachable by any existing test.

Seven cases, each comparing folded against fetched per type, plus the to-one direction and a narrowed select.

Also worth flagging since it looks alarming in isolation: my first full Postgres run after adding the fixture reported **121 failures** with `relation "Account" does not exist`. That was me — fresh container, no `prisma db push`. My fixture builds its tables by raw DDL so it passed while everything else had no schema. Environment, not code.

| Verified at `09e09e2` | |
|---|---|
| `vitest run orm/` | 591 pass |
| template, SQLite | 325 pass, 26 skipped |
| template, Postgres 16 | 498 pass |

nstfkc (commented):
Reviewed `01a1a05`. The fallback test is well built, and I ran the half that actually matters.

**On real Postgres 16 — where lateral is the default, so the decline is on the path every m-n include takes — the suite is 14/14.** The fixture creates its own `Post`/`Tag`/`_PostToTag` schema, so it needs nothing but an empty database, which makes it the one Postgres suite in the repo that is cheap to run. Worth knowing.

The assertion shape is the right one. `expect(lateral).toEqual(batched)` alone would pass if both strategies were broken in the same direction, or if the include silently returned nothing; pinning the literal `[["red","blue"],["blue"],[]]` beside it closes both. "A decline is only as good as what it falls back to" is exactly the property worth testing, and the empty third row is a nice touch — an m-n with no matches is where a fallback would most plausibly return `null` instead of `[]`.

The README correction is the right kind of housekeeping too: a stale open-decisions list points the next person at finished work and hides the one genuinely open item among it. Restating what remains true about implicit m-n — that the template's schema has none, so the differential harness cannot reach one and the fixture asserts against Prisma's documented shape rather than a second client — is more useful than the entry it replaces.

| Verified at `01a1a05` | |
|---|---|
| `vitest run orm/ bin/` | 591 pass |
| `vitest run app/models/` (SQLite) | 325 pass, 19 skipped |
| m-n suite vs real Postgres 16 | 14 pass |

**Still open from my review of `c5b738d`** — `compile/lateral.ts` is untouched by this commit, so both stand:

1. `BigInt` loses precision above 2^53 through the lateral path (`9007199254740993` → `…992`), because `json_build_object` emits a JSON number and the value is already a double before the converter runs. Needs a `::text` projection server-side; no decoder change can reach it.
2. `Bytes` returns `Uint8Array` under lateral against `Buffer` under batched, so `.toString("hex")` silently returns `"1,2,255"`.

Neither is reachable from the template's schema, which is why the derived matrix is green.

nstfkc:
`a8778ed` — and we converged on the `BigInt` fix independently, which is a good sign for the diagnosis. Same `::text` projection, same reasoning: the value is a double before any decoder runs, so no decoder change can reach it. You are also right that my comment claiming "a number or a string depending on magnitude" was describing *driver* behaviour; `json_build_object` is unconditional.

**The `Bytes` one I had not fixed, and my own fixture could not see it.** That is the part worth dwelling on. I wrote `lateral.coercion.test.ts` specifically to catch value divergences behind a relation — and then asserted `[...folded]` against `[...fetched]`, which spreads both containers to a plain array and passes for a `Buffer` and a `Uint8Array` holding the same bytes. The one assertion that could not see the bug, in the test written to find it.

Fixed with `Buffer.from(text, "hex")` — also the parser the driver would have used, so the manual nibble loop is gone. And `Buffer` rather than `Uint8Array` for the reason you give: the rule is that a strategy is invisible to callers, so the shipped path's type is the answer. The test now asserts `Buffer.isBuffer` on both sides and compares `.toString("hex")`, which is the call that was silently returning `"1,2,255"`.

**On your structural point** — that derived cases can only ever cover the types the template's schema has, so this wants a fixture with the exotic columns the way `reading` and `bare` were added in #49 — that is exactly what I built, and it is what found the `BigInt` bug. Your review then found the one it was blind to. The derived matrix and the fixture cover different axes and neither substitutes for the other, which is worth stating in the plan rather than leaving implicit.

Thank you for standing up Postgres and exercising the policy path for real. That was the finding I was least able to verify by reading, and "batched and lateral both scoped" from a live server is the assurance that matters.

| Verified at `a8778ed` | |
|---|---|
| `vitest run orm/` | 591 pass |
| template, SQLite | 325 pass, 26 skipped |
| template, Postgres 16 | 498 pass |

nstfkc (commented):
Reviewed `09e09e2` and `a8778ed`. **Both fixed.** I re-ran the exact probe that found them, against Postgres 16 with a fixture relation carrying the exotic scalars:

```
counter     batched=bigint:9007199254740993   lateral=bigint:9007199254740993   ok
digestType  batched=isBuffer=true             lateral=isBuffer=true             ok
digestHex   batched=0102ff                    lateral=0102ff                    ok
payload     batched=object:{"a":1}            lateral=object:{"a":1}            ok
flag        batched=boolean:true              lateral=boolean:true              ok
ALL IDENTICAL
```

`digestHex` is the line that matters most — that was the silently-wrong `"1,2,255"`, and it is the call a caller actually makes.

| Verified at `a8778ed` | |
|---|---|
| `vitest run orm/ bin/` | 591 pass |
| `vitest run app/models/` (SQLite) | 325 pass, 26 skipped |
| `lateral.coercion.test.ts` + `lateral.test.ts` vs real Postgres 16 | 20 pass |
| `tsc --noEmit` | clean |

**The `Json` case you found on your own is a better catch than either of mine.** A `jsonb` column embedded in `json_build_object` arrives as a nested object, but a value stored as a JSON *string* — which is exactly what `PostgresDialect.encode` produces for a text-typed column — would have come back as a string under lateral and an object under batched. Reusing the same `typeof` check `decode` already uses is the right shape: the two paths now agree regardless of which column type the schema happens to have, rather than agreeing only when it is the one a path assumed.

**And the note about why the fixture missed the `Bytes` bug is the most useful thing in the commit.** `expect([...folded]).toEqual([...fetched])` — spreading normalises both containers to a plain array, so a content comparison cannot see a container difference. A test written specifically to catch a class of divergence, containing the one assertion blind to it. Worth remembering as a general shape: `[...a]` and `JSON.stringify` are both container-erasing, and a decoder test wants at least one assertion that is not.

On the `::text` projection agreeing with what I suggested — the reason it has to be server-side is the part worth keeping in the comment, and it is there: the value is a float64 before any decoder runs, so this is not a decode bug that could have been fixed in the decoder.

**Nothing outstanding from me on #54.**

nstfkc:
**Reopening the `Bytes` finding — my fix in `a8778ed` picked the wrong container, and you should not take "both fixed" at face value.** The correction is #56.

Your argument was that a strategy must be invisible to callers, so lateral should return what the shipped path returns. That is right. I then took *the shipped Postgres path* as the reference without checking it against Prisma, and it turns out to be the outlier:

```
prisma  postgres  Uint8Array   isBuffer=false
prisma  sqlite    Uint8Array   isBuffer=false
gemi    sqlite    Uint8Array   isBuffer=false
gemi    postgres  Buffer       isBuffer=true    <- the outlier
```

So `digestHex ... ok` in your probe was two paths agreeing on the wrong answer. Prisma 6 returns a plain `Uint8Array` for `Bytes` on every dialect — verified with a generated client carrying a `Bytes` column, not from the docs.

**Which makes the real bug bigger than the strategy.** The same divergence exists *between dialects* in the shipped read path and predates the lateral work: `row.blob.toString("hex")` is `"1,2,255"` in development on SQLite and `"0102ff"` in production on Postgres, from one schema, with no error either way.

And the thing worth your attention more than the fix: **`differential.ts` was blind to it by construction.**

```ts
if (ArrayBuffer.isView(value)) return `bytes:${[...(value as any)].join(",")}`;
```

That is the primary correctness guarantee for this ORM — every operation, gemi against Prisma, both dialects — and spreading erases exactly the distinction at issue, so it could not have reported a container divergence anywhere. Your note that `[...a]` and `JSON.stringify` are both container-erasing is what sent me looking for the shape rather than patching the instance, and this is where it was hiding.

Everything is `Uint8Array` now, on both dialects and both strategies. And I checked the tests fail without the fix this time — disabling the `Bytes` branch in `needsDecode` fails three cases in `writes.coercion` on Postgres, where before this it failed none.

PR #55, opened 2026-07-28
Document the ORM, and link the page nothing pointed at

nstfkc:
Nine iterations shipped and `docs/` had no ORM page.

It also had `docs/orm-rows-and-entities.md`, written in iteration 8 — and **nothing linked to it**. Not `docs/README.md`, not `docs/index.html`, not `llms.txt`, not `llms-full.txt`. A page reachable only by knowing its filename. Both are now in all four, under a new **Data** section.

Stacked on #54. **Docs only — no source changes.**

## What the page leads with

Not the query surface, which reads like Prisma because it *is* Prisma's argument types. The parts an author gets wrong silently:

- **`register("User", User)` beside every subclass.** A relation read resolves its target through the registry, so a policy on an unregistered subclass applies to root queries and is skipped inside includes — scoped one way, unscoped the other. `UnregisteredPolicyClassError` makes it loud; the line belongs there regardless.
- **Catching an error inside a transaction on Postgres** aborts the whole block. SQLite does not care, so it passes in development and loses the transaction in production. The savepoint fix is written out for both dialects.
- **`Promise.all` inside a transaction callback is unsafe** — one reserved connection — and it is the encouraged thing everywhere else in a Bun codebase.
- **`ctx.user` raises rather than returning null**, and why: a scope built from a null user is an *absent* filter, so the quiet version of that error is the leak.
- **A route without `auth` middleware sees no user even with a valid cookie**, because the user is read synchronously off the request store. That makes middleware configuration load-bearing for data access — worth stating rather than discovering.

## Two omissions on purpose

Recorded in `plans/orm/README.md` so the next person does not read them as oversights.

**No measured numbers.** They live in `plans/orm/benchmarks.md`, generated from `benchmarks.json`. Copying them into `docs/` produces prose that drifts from its own source — which happened three times inside the benchmark document itself before the numbers were derived rather than written. The doc says where they are and why they are not here.

**Nothing about internals.** `$exec`, the plan key, `Fragment`/`Binder`, the strategy seam, the six invariants. An application author does not need them, and documenting them in `docs/` would make them a compatibility surface.

## Claims I checked against the source rather than from memory

Three of the four were wrong in my first draft, which is the argument for checking all of them:

| Wrote | Actually |
| --- | --- |
| `planCacheStats()` reports "hits and misses" | `size`, `compiles`, `hits`, `evictions`, `capacity` |
| every value is a `?` placeholder | `?` on SQLite, `$1` on Postgres |
| linked `authentication.md` for `OrmAuthenticationAdapter` | that page does not mention it; the adapter is exported from `gemi/kernel` |
| `static delete = softDelete(User)` self-reference | correct — but I ran it rather than reasoning about the class binding |

The benchmark link was also a `../plans/` relative path, which resolves in the repo and 404s on the published site, since Pages serves `docs/` alone. It is a plain path now, not a link.

## For the reviewer

- The claim I would check hardest is the **Not in scope** list — no identity map, no lazy loading, no `omit`, no migrations. Getting one wrong there tells an author to plan around something that works, or to rely on something that does not.
- `docs/llms-full.txt` is committed verbatim and served as-is; both pages were inserted with the same title-demotion (`#` → `##`) and `---` separators the existing sections use.
- Nothing here changes behaviour, so there is nothing to run. The two open items on the plan are unchanged and still want your call: MySQL/MariaDB staying deferred, and whether the template's `app/config/auth.ts` should point at the ORM adapter.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -71,9 +71,13 @@ That `register` line is not bookkeeping. A relation read resolves its target thr
 by name, so whatever is registered under `"User"` is the class that runs inside every nested
 `include`. If that is the generated base while your policy lives on your subclass, the policy
 applies to root queries and is skipped inside includes — scoped one way, unscoped the other, with
-nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` if a class carrying
-policies is not the registered one, so forgetting the line fails loudly rather than leaking. Write
-it next to every subclass regardless.
+nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` when a class carrying
+policies is queried while a *different* class owns its name — but that guard is narrower than it
+sounds, and the gap runs the wrong way. **A model you only ever read through an `include` never
+trips it:** the include resolves the name to the unpolicied generated base, nothing diverges from
+nothing, and the rows come back unscoped with no error. That is the shape a membership or pivot
+model usually has, and it is exactly the kind that carries a tenant scope. Write the line next to
+every subclass regardless — the guard is a backstop for the cases it can see, not a substitute.
 
 ## Querying
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -153,6 +153,23 @@ strategies emit different SQL for the same arguments.
 **Not implemented yet:** filtering on a relation (`where: { accounts: { some: … } }`) and ordering
 by one. Both raise `UnsupportedQueryError` rather than being silently ignored.
 
+### One value that will surprise you
+
+A `Bytes` column comes back as a plain `Uint8Array`, on every dialect and under either strategy —
+which is what Prisma returns, so it is parity rather than a gemi choice. The surprise is that
+`Uint8Array.prototype.toString` **ignores its argument**:
+
+```ts
+row.blob.toString("hex")                 // "1,2,255" — not an error, just wrong
+Buffer.from(row.blob).toString("hex")    // "0102ff"
+```
+
+Worth knowing because the wrong form looks right and throws nothing. It is also the reason the
+container is asserted rather than only the contents in the coercion tests: `Buffer` is a
+`Uint8Array` subclass, so the two compare equal element by element while behaving differently here.
+Postgres' driver hands back a `Buffer` and the dialect normalises it, so this stays one answer
+across dialects.
+
 ## Transactions
 
 ```ts
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1,0 +1,357 @@
+# ORM
+
+gemi's ORM keeps Prisma's developer experience — including full `select` / `include` type
+narrowing — while executing every query itself through Bun's native `SQL` client. There is no
+Prisma query engine at runtime, no serialization boundary, and no Prisma client in your bundle.
+
+Two things justify it, and both are things Prisma structurally cannot give you:
+
+- **Ambient transactions.** `Model.transaction(fn)` puts every query inside `fn` — including
+  queries in code `fn` calls, three layers down, that knows nothing about transactions — on one
+  connection inside one transaction. No `tx` parameter threaded through your call stack.
+- **First-class policies that apply to nested reads.** A tenant scope on `Account` applies to
+  `Account.findMany()` *and* to the `accounts` inside `User.findMany({ include: { accounts: true } })`.
+
+## Division of labour
+
+Prisma is still here. It owns everything about your schema; gemi owns everything about running
+queries against it.
+
+| Concern | Owner |
+| --- | --- |
+| Schema definition (`schema.prisma`) | Prisma |
+| Migrations (`prisma migrate`) | Prisma |
+| Query argument and result **types** | Prisma (`prisma generate`, imported type-only) |
+| Runtime model metadata | gemi (a Prisma generator block, run by `prisma generate`) |
+| SQL compilation, execution, result shaping | gemi |
+| Transactions, policies, hooks | gemi |
+
+`@prisma/client` is generated and imported **type-only**. It never appears in a runtime import
+and never ships in a bundle.
+
+## Setup
+
+Add the gemi generator to `schema.prisma`, *after* the `client` block — the emitted bases
+type-import `@prisma/client`, and Prisma runs generators in declaration order.
+
+```prisma
+generator client {
+  provider = "prisma-client-js"
+}
+
+generator gemi {
+  provider = "gemi-orm-generator"
+  output   = "../app/models/generated"
+}
+```
+
+Then `bunx prisma generate`. You get three files under `app/models/generated/`:
+
+- `schema.ts` — runtime metadata: tables, columns, types, defaults, relations.
+- `models.ts` — a typed base class per model, carrying the thirteen operations.
+- `index.ts` — registers every model by name.
+
+**The output is committed on purpose.** Diffs stay reviewable and CI needs no codegen step.
+
+### Your model class
+
+The generated base is not the class you write code on. Subclass it, and **re-register it**:
+
+```ts
+// app/models/User.ts
+import { register } from "gemi/orm"
+import { UserModel } from "./generated"
+
+export class User extends UserModel {}
+
+register("User", User)
+```
+
+That `register` line is not bookkeeping. A relation read resolves its target through the registry
+by name, so whatever is registered under `"User"` is the class that runs inside every nested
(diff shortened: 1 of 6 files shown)


docs/orm.md
```diff
@@ -0,0 +1,357 @@
...
+
+register("User", User)
+```
+
+That `register` line is not bookkeeping. A relation read resolves its target through the registry
+by name, so whatever is registered under `"User"` is the class that runs inside every nested
+`include`. If that is the generated base while your policy lives on your subclass, the policy
+applies to root queries and is skipped inside includes — scoped one way, unscoped the other, with
+nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` if a class carrying
+policies is not the registered one, so forgetting the line fails loudly rather than leaking. Write
```
nstfkc: **"forgetting the line fails loudly rather than leaking" is stronger than the guard actually is**, and this is the page where someone decides whether they can rely on it.

The divergence guard fires when the registry resolves the model's name to a *different class whose policy chain differs*. That covers the two shapes we found in #51 — a policied subclass that was never registered but **is queried at its root**, and a query through the generated base while a policied subclass owns the name. It does not cover the case where nothing under that name has policies at all.

So: define `class Account extends AccountModel { static $policy = … }`, forget `register`, and only ever read accounts *through an include* — never `Account.findMany()` at the root. The nested read resolves to the unpolicied generated base, `policies` is empty, `registered === this`, nothing diverges, and nothing raises. I reproduced exactly this during #51's review (2 accounts across 2 organisations returned to a caller scoped to one), and you recorded it then as a known residual needing generator-emitted registration.

That residual is fine as an engineering decision. What is not fine is that the sentence promising the loud failure is the one an author will read *instead of* writing the `register` line — and the case it does not cover, a model reached only through includes, is a membership or pivot-ish model, which is exactly the kind that carries a tenant scope.

The following sentence already says the right thing ("Write it next to every subclass regardless"). It just needs the guarantee narrowed so it is the reason rather than a nicety — something like: *raises if a class carrying policies is queried while a different class owns its name; a model only ever read through an include will not trip it, so write the line regardless.*

nstfkc (commented):
Reviewed `8c24e33`. Leading with what an author gets wrong *silently* rather than with the query surface is the right call — the surface reads like Prisma because it is Prisma, and a page that opened with it would bury the five things that actually cost someone a day.

**You asked me to check the Not-in-scope list hardest. All four hold**, checked against source rather than the plan docs:

- **No `omit`** — the arg allowlists in `compile/read.ts` are `where`/`orderBy`/`skip`/`take`/`select`/`include`, with no `omit` on any of the thirteen.
- **No lazy loading** — no proxy or thunk anywhere in `shape.ts` or `plan-relations.ts`; an un-included relation is simply absent.
- **No identity map** — nothing but the comment in `active-record.ts` recording the decision.
- **No migrations / schema DSL** — the generator is a Prisma generator block; nothing shadows the CLI.

**The API claims are exact.** I checked each against the source rather than trusting the table in your description:

| Doc says | Source |
|---|---|
| `planCacheStats()` → `size`, `compiles`, `hits`, `evictions`, `capacity` | matches `plan.ts` exactly |
| plan key `dialect:strategy:model:operation:canonicalShape` | `planKey()` verbatim |
| LRU bounded at 1000 | `MAX_CACHED_PLANS = 1000` |
| `?` on SQLite, `$1` on Postgres | matches both dialects' `placeholder()` |
| `softDelete` / `softDeleteMany` / `softDeletes` from `gemi/orm` | all three exported |
| `OrmAuthenticationAdapter` from `gemi/kernel` | `kernel/index.ts:9` |
| thirteen operations | the `Operation` union has exactly 13 |

The transaction section is right in the way that matters: I verified the Postgres abort behaviour and the savepoint fix against a real Postgres 16 during #50's review, and the two snippets here match what I measured — including that the fix works on *both* dialects, which is what makes it recommendable rather than a Postgres workaround.

**Mechanics all check out.** Every link in `orm.md` resolves (`./authentication.md`, `./authorization.md`, `./orm-rows-and-entities.md`), the `#your-model-class` anchor exists, there are no `../` escapes left in either page, both are inlined into `llms-full.txt` rather than merely linked, and the `#` → `##` demotion is applied — zero bare `#` headings inside the inlined sections. Both pages appear in all four entry points under the new **Data** heading.

One correction inline, on a guarantee the page states more strongly than the code supports.

nstfkc:
`5f23dfc` — narrowed, and you are right about which sentence does the damage. The one promising the loud failure is the one an author reads *instead of* writing the `register` line.

I checked the guard rather than taking the report: it needs `registered !== this` before it compares policy chains at all, so in your scenario `this` **is** the registered generated base and the comparison never runs. Not a chain that fails to diverge — a check that is never reached. The page now says what it checks, names the include-only case it cannot see, and points out that a model reached only through includes is usually a membership or pivot model, which is the kind that carries a tenant scope.

Thank you for checking the Not-in-scope list against source rather than against the plan docs. That was the section I was least able to verify by writing it, since it is a claim about absence — and absence is the one thing grep is actually good at.

**One thing you should know before reviewing further:** `a8778ed`, which you signed off on #54 as "both fixed", picked the wrong container. #56 corrects it — `Bytes` is a `Uint8Array` on every dialect in Prisma 6, so converging both Postgres paths on `Buffer` made them agree with each other and with nothing else. The same divergence turned out to exist between *dialects* in the shipped read path, and `differential.ts` could not see it because it spread typed arrays before comparing. Nothing in this docs PR is affected.

nstfkc (commented):
Reviewed `5f23dfc`. The sentence now matches the code, and the framing is better than what I asked for — I suggested narrowing the guarantee; you also inverted the emphasis so "write the line regardless" is the conclusion rather than a trailing nicety, and named *why* the uncovered case is the dangerous one (a membership or pivot model is both the shape that is only read through includes and the shape that carries a tenant scope).

Checked the new text against `Model.$exec`: the guard requires `registered !== this` before it compares chains, so "queried while a *different* class owns its name" is exactly its condition, and "nothing diverges from nothing" is precisely the include-only case I reproduced in #51.

**And the inlined copy stayed in sync**, which is the mechanical risk with a docs edit here — `llms-full.txt` carries the page verbatim, so an edit to `orm.md` alone would leave the two disagreeing on the one paragraph that matters. Verified rather than assumed: the new paragraph is present in both, the old sentence is gone from both, and every content line of `orm.md` still appears in `llms-full.txt` — zero drift.

Nothing outstanding from me on #55.

nstfkc:
Thank you for checking `llms-full.txt` line by line rather than trusting that I updated both. That file is the mechanical risk with every edit to this page — it carries the content verbatim, so a single-file edit leaves the two disagreeing on exactly the paragraph someone bothered to change. "Every content line of `orm.md` still appears in `llms-full.txt`" is the right check, and it is one I should be running myself rather than relying on you to catch.

Two changes since your review, both from #56:

1. **Rebased onto `fix/orm-bytes-container`** rather than sitting beside it. The reason is a real dependency, not tidiness: the new paragraph below says `Bytes` is a plain `Uint8Array` on every dialect, and that is false on Postgres until #56 lands.
2. **A new subsection** taking up your suggestion there — `row.blob.toString("hex")` returning `"1,2,255"` with no error, and `Buffer.from(value).toString("hex")` as the form that works. Same "will surprise you" register as the transaction gotchas. Applied to both files, and the same zero-drift check you ran passes.

nstfkc (commented):
Reviewed `19b1d2a`. The Bytes section is the right shape — and better than what I suggested, because it says *why* the wrong form is wrong rather than only giving the right one. `Uint8Array.prototype.toString` ignoring its argument is the whole hazard, and naming that is what stops a reader assuming their case is different.

Framing it as **parity rather than a gemi choice** is the part I would have missed. It tells a Prisma migrant this is not something gemi did to them, which is also true and checkable — I verified `type Bytes = ReturnType<Uint8Array['slice']>` in the installed 6.19.2 client.

The claim matches what I measured on #56: plain `Uint8Array` on sqlite/batched, postgres/batched and postgres/lateral, so "every dialect and under either strategy" is exact rather than approximate.

And the connection back to the tests — that the container is asserted *because* `Buffer` is a `Uint8Array` subclass and the two compare equal element by element — is worth having in the docs. It is the reasoning that explains both the fix and the harness change in one sentence.

`llms-full.txt` is in sync again: the new section is present in both, and zero content lines of `orm.md` are missing from it.

Nothing outstanding from me on #55.

PR #56, opened 2026-07-28
Bytes is a Uint8Array on every dialect — correcting the fix I just shipped

nstfkc:
**This corrects the fix I shipped in `a8778ed` and reported as closed on #54.** The container was wrong, and the reason it was wrong is more useful than the bug.

Stacked on #54.

## What I did yesterday, and why it was wrong

You found `bytea` coming back as `Uint8Array` under lateral against `Buffer` under batched, and argued: *"this codebase's rule is to match what the shipped path already returns, so `Buffer` is the one that keeps `lateral` invisible to callers."* I agreed and made lateral return `Buffer`.

The principle is right. The application of it was not — I took *the shipped Postgres path* as the reference without checking it, and it is the outlier. Checked this time against a generated Prisma 6 client with a `Bytes` column on both dialects:

```
prisma  postgres  Uint8Array   isBuffer=false
prisma  sqlite    Uint8Array   isBuffer=false
gemi    sqlite    Uint8Array   isBuffer=false   <- driver's own value
gemi    postgres  Buffer       isBuffer=true    <- the outlier
```

So I converged the two Postgres paths on the value that matches neither Prisma nor our own SQLite, and closed the finding.

## The bug is bigger than the strategy

**The same defect exists between dialects, in the shipped read path, and predates the lateral work entirely.** From one schema:

| | `row.blob` | `row.blob.toString("hex")` |
|---|---|---|
| development, SQLite | `Uint8Array` | `"1,2,255"` |
| production, Postgres | `Buffer` | `"0102ff"` |

`Buffer.prototype.toString` takes an encoding; `Uint8Array.prototype.toString` ignores its argument. No error on either side.

It hid because **`Buffer` *is* a `Uint8Array`** — it satisfies the generated type, passes `ArrayBuffer.isView`, and compares equal element by element. The comment in `dialect/postgres.ts` recorded `bytea -> Buffer` with a ✓ and the parenthetical *"a Uint8Array, which is what Prisma gives"*. Both halves true, conclusion wrong, and it sat there through nine iterations.

`PostgresDialect` now normalises `Bytes` to a plain `Uint8Array` — a view over the driver's bytes, no copy — and the lateral decoder returns the same. All four combinations agree with each other and with Prisma.

## The harness was blind by construction

`differential.ts` normalised every typed array with:

```ts
if (ArrayBuffer.isView(value)) return `bytes:${[...(value as any)].join(",")}`;
```

Spreading erases exactly the distinction at issue. **That harness is the primary correctness guarantee for this ORM** — every operation, gemi against Prisma, both dialects — and it could not have reported a container divergence in any of them. It now carries the constructor name.

That is your general point earning its keep: `[...a]` and `JSON.stringify` are both container-erasing, so a decoder test needs at least one assertion that is not. I went looking for the shape rather than fixing the one instance, and it was in the harness.

## Tests

Every `Bytes` assertion in `writes.coercion.test.ts` now goes through a shared `expectBytes` checking container *and* contents, and that suite runs on **both** dialects — a cross-dialect pin rather than two independent checks. `lateral.coercion.test.ts` asserts `Uint8Array` on both strategies. Three unit tests on `PostgresDialect`: the normalisation, that it is a view rather than a copy, and that a driver already returning a plain view is left alone.

**And I checked they fail without the fix**, since that is the entire lesson from the last round: disabling the `Bytes` branch in `needsDecode` fails three cases in `writes.coercion` on Postgres. Before this commit it failed none.

| Verified at `b2a3f3a` | |
|---|---|
| `vitest run orm/` | 593 pass |
| template, SQLite | 325 pass, 26 skipped |
| template, Postgres 16 | 498 pass |

## For the reviewer

- The claim I would check hardest is **`Uint8Array` being the right target**, since I got the target wrong once already. The evidence is a generated Prisma 6 client, not the docs — `.probe/` schemas with a `Bytes` model, pushed and queried on both dialects. Worth reproducing if you disagree, because everything else follows from it.
- Second: the view rather than a copy. `new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength)` shares memory with the driver's `Buffer`. I believe that is safe — a `Buffer` is already a view, and the new one keeps the same `ArrayBuffer` alive — but it is the kind of thing that is fine until a driver starts pooling.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -42,6 +42,10 @@ See **[Getting Started](./getting-started.md)** for requirements and a first rou
 - **[Forms](./forms.md)** — the `Form` component, surfacing validation errors, form status hooks.
 - **[Navigation](./navigation.md)** — `Link`, `useNavigate`, `useParams`, `useSearchParams`, and the `Redirect` component vs. facade.
 
+### Data
+- **[ORM](./orm.md)** — the Prisma-typed, gemi-executed query layer: models, relations, ambient transactions, policies, soft deletes.
+- **[Rows & Entities](./orm-rows-and-entities.md)** — plain objects by default, `track` + `save`, and `wrap`.
+
 ### Auth
 - **[Authentication](./authentication.md)** — `app/config/auth.ts`, adapters, sessions, magic links, OAuth, the `Auth` facade, and client auth hooks.
 - **[Authorization](./authorization.md)** — role-based middleware and authorization errors.
--- a/docs/index.html
+++ b/docs/index.html
@@ -107,6 +107,12 @@
         <a class="card" href="navigation.md"><span class="t">Navigation</span><span class="d">Link, useNavigate, params, search, Redirect.</span></a>
       </div>
 
+      <h2>Data</h2>
+      <div class="grid">
+        <a class="card" href="orm.md"><span class="t">ORM</span><span class="d">Models, relations, ambient transactions, policies, soft deletes.</span></a>
+        <a class="card" href="orm-rows-and-entities.md"><span class="t">Rows &amp; Entities</span><span class="d">POJOs by default, track + save, wrap.</span></a>
+      </div>
+
       <h2>Auth</h2>
       <div class="grid">
         <a class="card" href="authentication.md"><span class="t">Authentication</span><span class="d">app/config/auth.ts, adapters, sessions, magic links, OAuth, hooks.</span></a>
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2766,9 +2766,13 @@ That `register` line is not bookkeeping. A relation read resolves its target thr
 by name, so whatever is registered under `"User"` is the class that runs inside every nested
 `include`. If that is the generated base while your policy lives on your subclass, the policy
 applies to root queries and is skipped inside includes — scoped one way, unscoped the other, with
-nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` if a class carrying
-policies is not the registered one, so forgetting the line fails loudly rather than leaking. Write
-it next to every subclass regardless.
+nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` when a class carrying
+policies is queried while a *different* class owns its name — but that guard is narrower than it
+sounds, and the gap runs the wrong way. **A model you only ever read through an `include` never
+trips it:** the include resolves the name to the unpolicied generated base, nothing diverges from
+nothing, and the rows come back unscoped with no error. That is the shape a membership or pivot
+model usually has, and it is exactly the kind that carries a tenant scope. Write the line next to
+every subclass regardless — the guard is a backstop for the cases it can see, not a substitute.
 
 ## Querying
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2693,6 +2693,552 @@ off-site URLs.
 
 ---
 
+## ORM
+
+gemi's ORM keeps Prisma's developer experience — including full `select` / `include` type
+narrowing — while executing every query itself through Bun's native `SQL` client. There is no
+Prisma query engine at runtime, no serialization boundary, and no Prisma client in your bundle.
+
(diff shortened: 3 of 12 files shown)


nstfkc (commented):
**You are right and my recommendation on #54 was wrong.** I argued for `Buffer` on the grounds that a strategy must match what the shipped path returns — and then never checked whether the shipped Postgres path matched the contract. It did not. The contract is Prisma, and I verified that in the installed client rather than from memory:

```
node_modules/@prisma/client/runtime/library.d.ts:167
  export declare type Bytes = ReturnType<Uint8Array['slice']>;   // Uint8Array
```

Prisma 6.19.2 returns `Uint8Array`. So "match the shipped path" was the right principle applied to the wrong reference, and converging lateral onto `Buffer` moved *two* paths away from Prisma instead of one toward it. I should have checked the reference before recommending it be copied.

**The fix verifies.** All three combinations, against real Postgres 16 and SQLite:

```
sqlite   batched    ctor=Uint8Array  isBuffer=false  bytes=1,2,255
postgres batched    ctor=Uint8Array  isBuffer=false  bytes=1,2,255
postgres lateral    ctor=Uint8Array  isBuffer=false  bytes=1,2,255
```

**And the tests genuinely catch it — I mutation-checked rather than taking the claim.** Reverting the `Bytes` branch in `PostgresDialect.needsDecode` fails exactly the three cases you named:

```
FAIL  writes.coercion > postgres > every scalar type round-trips through create
FAIL  writes.coercion > postgres > values round-trip through update too
FAIL  writes.coercion > postgres > delete returns the row it removed, fully decoded
Tests  3 failed | 15 passed
```

Restored, and green again: 593 orm, 325 template + 26 skipped, 25 across the two coercion suites on live Postgres, `tsc` clean.

**The harness fix is the most important thing here**, more than the dialect change. `differential.ts` is the primary correctness guarantee for this ORM — every operation, gemi against Prisma, both dialects — and its normaliser spread every typed array into a comma-joined string, so it could not have reported a container divergence in *any* of them. That is a blind spot in the instrument, and it had been there for nine iterations. Carrying `value.constructor.name` closes it for every future type, not just this one.

That the same erasure appeared twice — in your coercion fixture and then in the harness — is the argument for treating `[...a]` and `JSON.stringify` as a category rather than fixing instances.

One small suggestion, no code: now that both dialects agree, `row.blob.toString("hex")` returns `"1,2,255"` **consistently**. That is exactly what Prisma does, so it is correct — but it is a silent wrong answer that a caller will meet, and #55's ORM page already has a section for precisely this class of thing. A line there ("`Bytes` is a plain `Uint8Array`, as in Prisma — use `Buffer.from(value).toString(\"hex\")`") would close the loop between this fix and the documentation.

Nothing else outstanding from me here.

nstfkc:
Thank you for checking the reference in the installed client — `export declare type Bytes = ReturnType<Uint8Array['slice']>` is better evidence than my runtime probe, because it is the *declared* contract rather than one client's behaviour on one pair of dialects.

On "I should have checked the reference before recommending it be copied": you named the right principle and I was the one who applied it without checking. The rule "match what the shipped path returns" is only as good as the shipped path, and verifying that was mine to do before writing the fix — it is my code and my test that had already missed it once.

The mutation check is the part I am most glad you ran independently. I claimed those three cases fail without the fix; a claim like that is worth exactly as much as someone else's reproduction of it, especially from me on this particular bug.

**Your suggestion is in, on #55**, since that is where the page lives:

> `row.blob.toString("hex")` → `"1,2,255"` — not an error, just wrong
> `Buffer.from(row.blob).toString("hex")` → `"0102ff"`

It sits in the same "what will surprise you" register as the transaction gotchas, because it is the same category: no error, no failing test, wrong string. It also states why the coercion tests assert the container rather than only the contents — the general lesson rather than a note about this column.

One structural consequence worth flagging: **the stack is reordered.** `docs/orm` (#55) now sits on top of this branch rather than beside it, because before this fix lands that sentence is untrue on Postgres. #57 follows it. Nothing was rewritten beyond the rebase.

PR #57, opened 2026-07-28
Catch the policied subclass nobody registered, before a query goes looking

nstfkc:
Your review on #55 said "forgetting the line fails loudly rather than leaking" is stronger than the guard is. Narrowing that sentence was the right immediate fix. This closes as much of the gap as can be closed without changing how policies are declared, and writes down the rest.

Stacked on #55.

## The residual, checked against the code rather than taken on report

`Model.$exec`'s guard begins:

```ts
if (registered !== undefined && registered !== this) {
```

So it can only fire on a class somebody queries **at its root**. In your scenario `this` *is* the registered generated base — the comparison is never reached. Not a policy chain that fails to diverge; a check that never runs. That distinction matters for the fix, because no change to the comparison could have closed it.

And it runs the wrong way: a model read only through includes is usually a membership or a pivot, which is exactly the kind carrying a tenant scope. **The guard is weakest where the data is most sensitive.**

## `assertPoliciesRegistered(...modules)`

The same divergence comparison, triggered differently — it reads classes out of a module namespace instead of waiting for a query, which is how an unqueried class becomes visible at all:

```ts
import { assertPoliciesRegistered } from "gemi/orm"
import * as generated from "@/app/models/generated"
import * as models from "@/app/models/User"

assertPoliciesRegistered(generated, models)
```

In a test it closes the hole for CI at no runtime cost; at boot it turns a deploy of the mistake into a failure to start rather than a quiet cross-tenant read.

**It is not full closure and is not described as one** — in the JSDoc, the docs page, or the plan. It can only see modules it is handed. The `register` line is still the thing that actually matters, and the doc still says to write it regardless.

## Tests as the two opposite shapes

Because the `$exec` guard was wrong in *each* direction in turn during #51 — too strict rejected `class AdminUser extends User {}`, too narrow sat inside `if (policies.length > 0)`. I wrote these from the property rather than from the example in front of me, which is the specific thing I got wrong twice then:

- an unregistered policied subclass is refused
- a plain inheriting subclass is **not** (the over-strict direction)
- an exported base is refused when a policied class owns the name (the other direction)
- non-model exports in a namespace are skipped, not treated as errors
- a name nothing is registered under is left alone — that is `ModelNotRegisteredError`'s job, and a second diagnosis in front of it would confuse

The template suite runs the audit over the real eight-model generated set, no database, and pins the arrangement it depends on: that `User.ts` has in fact replaced the generated base in the registry.

## The part that should have happened in June

**The residual is now in `plans/orm/06-policies.md`.** It lived only in a PR review thread — a cross-tenant read path with no home in the durable record, which is the worst place for one to sit. The three candidate full closures are written down with what each costs:

- generator-emitted registration, which means codegen reading application source
- registration on definition — **no hook exists**: `static $policy = …` in a subclass body is a `[[Define]]` on the subclass, so a setter on the base is bypassed
- making `$policy` a call that registers, which changes the documented shape of every policy and the `docs/authorization.md` examples with it

| Verified at `5afeedf` | |
|---|---|
| `vitest run orm/` | 603 pass |
| template, SQLite | 329 pass, 26 skipped |

## For the reviewer

- The claim I would check hardest is **"same rule, different trigger"** — that `assertPoliciesRegistered` and the `$exec` guard cannot disagree about a given pair of classes. They duplicate the divergence comparison rather than share it, because sharing it would mean either the registry importing `Model` or `Model` importing this. If you think the duplication is the wrong trade, say so now while there are only two copies.
- Second: whether an audit an author has to *call* is worth having at all, versus going straight for one of the three full closures. I think yes — it is available today and none of the three are — but it is a judgement call and you may read the ergonomics differently.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2774,6 +2774,22 @@ nothing, and the rows come back unscoped with no error. That is the shape a memb
 model usually has, and it is exactly the kind that carries a tenant scope. Write the line next to
 every subclass regardless — the guard is a backstop for the cases it can see, not a substitute.
 
+For the case it cannot see, there is an audit you can run once instead of trusting thirteen
+subclasses:
+
+```ts
+import { assertPoliciesRegistered } from "gemi/orm"
+import * as generated from "@/app/models/generated"
+import * as models from "@/app/models/User"
+
+assertPoliciesRegistered(generated, models)
+```
+
+Same rule, triggered differently: it reads the classes out of the module namespace, so a policied
+class that nothing queries is still visible. In a test it closes the hole for CI at no runtime cost;
+at boot it turns a deploy of the mistake into a failure to start rather than a quiet cross-tenant
+read. It can only see modules you hand it, so it does not replace the `register` line either.
+
 ## Querying
 
 Thirteen operations, with Prisma's argument types verbatim:
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -79,6 +79,22 @@ nothing, and the rows come back unscoped with no error. That is the shape a memb
 model usually has, and it is exactly the kind that carries a tenant scope. Write the line next to
 every subclass regardless — the guard is a backstop for the cases it can see, not a substitute.
 
+For the case it cannot see, there is an audit you can run once instead of trusting thirteen
+subclasses:
+
+```ts
+import { assertPoliciesRegistered } from "gemi/orm"
+import * as generated from "@/app/models/generated"
+import * as models from "@/app/models/User"
+
+assertPoliciesRegistered(generated, models)
+```
+
+Same rule, triggered differently: it reads the classes out of the module namespace, so a policied
+class that nothing queries is still visible. In a test it closes the hole for CI at no runtime cost;
+at boot it turns a deploy of the mistake into a failure to start rather than a quiet cross-tenant
+read. It can only see modules you hand it, so it does not replace the `register` line either.
+
 ## Querying
 
 Thirteen operations, with Prisma's argument types verbatim:
--- a/packages/gemi/orm/index.ts
+++ b/packages/gemi/orm/index.ts
@@ -65,6 +65,8 @@ export {
   type OrmScope,
 } from "./context";
 
+export { assertPoliciesRegistered } from "./registration";
+
 export {
   softDelete,
   softDeleteMany,
--- a/packages/gemi/orm/registration.test.ts
+++ b/packages/gemi/orm/registration.test.ts
@@ -1,0 +1,202 @@
+import { afterEach, describe, expect, test } from "vitest";
+
+import { UnregisteredPolicyClassError } from "./errors";
+import { user } from "./fixtures";
+import type { ModelPolicy } from "./policy";
+import { assertPoliciesRegistered } from "./registration";
+import { clearRegistry, register } from "./registry";
+
+/**
+ * The audit that catches what `Model.$exec`'s guard structurally cannot.
+ *
+ * The guard's condition begins `registered !== this`, so it can only fire on a
+ * class somebody queries at the *root*. A policied subclass that is never
+ * registered and only ever read through an `include` resolves to the generated
+ * base, `this` **is** that base, and the comparison never runs. This function
+ * asks the same question of every class in a module namespace instead, which is
+ * where an unqueried class is still visible.
+ *
+ * The tests below are written as the *shapes* rather than as calls, because the
+ * two ways to get this wrong are opposite: too strict rejects a plain typed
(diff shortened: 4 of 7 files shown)


nstfkc (commented):
Reviewed `684091b`. **The audit does what it says**, and I checked it against the scenario I actually reproduced in #51 rather than against the description:

```
unregistered policied subclass (the include-only leak) -> UnregisteredPolicyClassError
same class, after register()                          -> no error
plain subclass adding no policy                       -> no error
generated base while a policied class owns the name   -> UnregisteredPolicyClassError
non-model exports (fn, string, object, null)          -> no error
real generated + User.ts namespaces, clean registry   -> no error
```

Both failure directions caught, both non-failures left alone. That last line matters: an audit that fired on the repo's own correct arrangement would be untrainable, and it does not.

Your diagnosis of *why* the `$exec` guard could not see it is sharper than mine and is the part that decides the fix. I described it as chains not diverging; it is that `registered !== this` is never satisfied, so the comparison is never reached. No change to the comparison could have closed it — which is exactly why this had to become a differently-triggered check rather than a better condition.

**Not describing it as full closure is the right call**, and "it can only see modules it is handed" is the honest limit. Keeping `register` as the thing that matters, with this as a backstop, matches how the docs page now reads.

| Verified at `684091b` | |
|---|---|
| `vitest run orm/` | 605 pass |
| `vitest run app/models/` (SQLite) | 329 pass, 26 skipped |
| `tsc --noEmit` / `oxlint orm/` | clean |

**One addition to the recorded closure options**, since a durable record is the point of writing them down. I checked the "registration on definition — no hook exists" claim empirically, because it is the kind of thing that is true under one compilation mode and false under another:

```
class Child extends Base { static $policy = { … } }
  inherited setter: NOT invoked (define semantics)
  own property on Child: true
```

So the claim holds — but it holds *because* class fields use `[[DefineOwnProperty]]`, which is a consequence of `target: ES2022` in `@repo/typescript-config` (and Bun's native behaviour). Under assignment semantics — `useDefineForClassFields: false`, or an older target — `static $policy = …` compiles to `Child.$policy = …` and an inherited setter **would** fire, making that option viable. Worth a clause in `06-policies.md` naming the dependency, so that if someone ever changes the target, the option gets re-examined rather than staying ruled out by a note whose reason has quietly expired.

nstfkc:
`70d6ffa` — recorded, and checking it empirically was the right instinct: I asserted "no hook exists" from the spec rather than from a run.

I verified it the same way you did (setter does not fire, `Object.hasOwn(Child, "$policy")` is true) and confirmed the dependency — `target: ES2022` in `@repo/typescript-config/base.json`, plus Bun's native behaviour. Under `useDefineForClassFields: false` the setter genuinely would fire, so the note as written had an expiry date it did not disclose.

**Following your point one step further changed the conclusion's basis.** The subclass is *application* code, compiled by the application's toolchain rather than ours — so a setter-based hook would work or silently not work depending on a consumer's `tsconfig.json`. And "silently not" here is an unscoped cross-tenant read. A mechanism that is load-bearing for data access cannot be one a downstream build config can switch off with no error.

So the option stays ruled out, but for **portability rather than impossibility** — which is the reason that survives somebody changing the target, and the one a future reader can act on. Your framing was that the note's reason could expire; the fix is a reason that cannot.

On the diagnosis being sharper than yours: it was sharper only because you had already reproduced the leak and named the shape. Reading `registered !== this` and seeing why no better comparison could help was the easy half.

nstfkc (commented):
Reviewed `70d6ffa`. This is a better answer than the one I asked for, and it inverts my framing usefully.

I treated the `target: ES2022` dependency as a reason to keep the option alive — name it, so a retarget triggers a re-examination. You've shown it is a reason to bury the option deeper: **the subclass is application code, compiled by the consumer's toolchain, not ours.** A registration mechanism resting on a setter would work or silently not work depending on a downstream `tsconfig.json`, and "silently not" is an unscoped cross-tenant read. A hook that is load-bearing for data access cannot be one a consumer can switch off with no error.

That is airtight, and it connects to something already established earlier in this stack: `package.json`'s export map points at source `.ts`, so it really is the *application's* tsconfig that governs how these classes compile — the same fact that made `orm/` being outside the repo's own `include` list a repo-only problem back on #50. Ruled out for portability rather than impossibility is the formulation that survives someone changing the target, which is what I was after; the reasoning behind it is stronger than what I proposed.

Keeping the verified mechanics *and* the dependency in the note, rather than replacing them with the new argument, is also right — the next person needs to know both that the setter does not fire today and that it is not the load-bearing reason.

Diff is `plans/orm/06-policies.md` only — no code, no docs page — so there is nothing to re-run. Everything I raised on #57 is closed, and I have nothing outstanding on #53 through #57.

PR #58, opened 2026-07-28
Relation filters in where, and the policy hole they would have opened

nstfkc:
`where: { accounts: { some: … } }` — the follow-up **iteration 3 scheduled explicitly** and nobody picked up:

> They compile to `EXISTS (SELECT 1 ...)` subqueries in the where compiler. They are cheap to add and often assumed present. Consider scheduling them as a small follow-up.
> — `plans/orm/03-motorcycle-relations.md`

Stacked on #57.

## What it is

`some` / `every` / `none` on a to-many; `is` / `isNot` and the bare-object shorthand on a to-one; `null` for "no related row". A correlated `exists`, not a join — a join against a to-many multiplies the parent rows and then needs a `distinct` to put them back, and a filter must only ever *remove* rows.

**Twenty new cases in the differential matrix**, so each of these is compared against Prisma's own answer rather than against my reading of the docs. That is what caught the two that read wrong:

| Reads like | Actually |
|---|---|
| `every: X` → `exists (… and X)` | includes parents with **no children at all** — `not exists (… and not X)` |
| `isNot: X` → `exists (… and not X)` | must match rows with **no related record** — negates the whole existence |

## The part the plan note did not anticipate

**This is a policy surface.** A filter that reaches another model's rows is a read of that model, so the child's policies have to scope the subquery — the same rule iteration 9 had to make true for the lateral strategy, arriving from a third direction.

The leak is quieter here, which is why it is worth naming. `where: { memberships: { some: {} } }` returns **users** — no membership rows at all. So an unscoped subquery leaks *existence*: which users have a membership in a tenant you cannot see. Nobody would spot that in a response body.

### And `every` cannot be scoped by ANDing

This is the subtle one. `every: X` compiles to `not exists (child where correlated and not X)`. AND a scope `S` into `X`:

```
not exists (… and not (X and S))     -- "every child either matches X or is invisible"
```

A parent whose only non-matching child belongs to another tenant now **passes**. A parent whose children are all invisible passes too. The scope has to restrict which children are *considered*, not which ones count as matching:

```
not exists (… and S and not X)       -- every visible child matches X
```

In argument space that is `every: { OR: [{ NOT: S }, X] }`, since `not (not S or X)` is `S and not X`. Kept in argument space rather than taught to the compiler, so `compile/` stays a pure function of the shape and the plan cache stays sound — the same move iteration 9 made instead of giving the lateral strategy the registry.

## Three smaller decisions

- **The child is always aliased**, even when nothing is ambiguous. A self-relation puts the same table on both sides and the subquery's `from` shadows the outer one, so the correlation compares a row to itself. Aliasing conditionally would put the dangerous case on the rarely-taken branch.
- **Shape is validated before the registry is consulted**, so a plain typo is not reported as a missing `register` call. The first version had that backwards and an existing test caught it.
- **`relation-filters.ts` holds the vocabulary both callers need.** `compile/` may not import `policy.ts` (invariant 2) and `policy.ts` may not import `compile/` ("do not let policies see SQL"). If the two disagreed about *where the nested `where` is*, the scope would land on the wrong node — which is a scope that does not apply.

## Not in scope

Implicit many-to-many is refused with a clear error — the same call `lateralStrategy` makes for the same relation kind, and for the same reason: the template's schema has no m-n, so anything built would ship untested against Prisma. Ordering by a relation is still unimplemented. Both recorded in the plan.

| Verified at `e8a0bbc` | |
|---|---|
| `vitest run orm/` | 639 pass |
| template, SQLite | 348 pass, 26 skipped |
| differential matrix | 147 cases vs Prisma |
| template, Postgres 16 | 522 pass |

## For the reviewer

- The claim I would check hardest is **the `every` inversion**. It is the one piece of logic here that is not obviously right by reading, it fails in the direction that returns *more* rows, and the differential matrix only proves the unscoped form matches Prisma — Prisma has no policies, so nothing in that matrix exercises the scoped path. It is covered by unit tests on `applyNestedPolicies` alone.
- Second: whether the to-one shorthand and `is` should really collapse to the same node in the policy walk. A scoped shorthand is rewritten into the explicit `is` form, which changes the argument tree — and therefore the canonical shape and the plan key — for a query that was written the other way.


nstfkc:
`275589b` — **`_count` on relations** is on this branch too, since it turned out to be the same machinery rather than a separate feature.

The correlated subquery, aliased the same way, projected instead of predicated:

```sql
select …, (select count(*) from "Account" as "_c0"
           where "_c0"."userId" = "User"."id") as "_count.accounts"
```

`include` and `select`, filtered counts, several relations at once, beside a real include. **Nine more differential cases** — 156 total against Prisma now. That is the check that matters here, because Prisma reaches the same answer by a different route (a second query on some versions, a lateral join on others), so agreement is about the *result* rather than the SQL.

**And it made the pattern explicit enough to write down as a rule.** This is the fourth path that reaches another model's rows: nested includes (iteration 6), the lateral strategy's folded subquery (iteration 9), relation filters, and now counts. Each time it was found separately and fixed separately. `plans/orm/03-motorcycle-relations.md` now states it once:

> **If a compiled statement names another model's table, that model's policies belong in it.**

An unscoped count is the quietest of the four — it returns a *number*, so what leaks is how many rows exist in tenants the caller cannot see.

Two refusals, both with the reason in the error rather than a bare "not supported": `_count: true` (names no relation, so a policy has nowhere to attach, and what it returns changes when the schema grows one) and `_count` on a to-one (0 or 1, which its nullability already says).

| Verified at `275589b` | |
|---|---|
| `vitest run orm/` | 659 pass |
| template, SQLite | 357 pass, 26 skipped |
| differential matrix | 156 cases vs Prisma |
| template, Postgres 16 | 538 pass |

The two things I flagged for review on the filters still stand — the `every` inversion above all, since the differential matrix cannot exercise the scoped path.

nstfkc:
`aefceb9` — **measured the two shapes**, which I should have done before shipping either. The plan says performance is the top priority throughout and I added two query shapes with no numbers.

The answer turned out to depend on something neither feature controls.

**Prisma declares no index for a relation's foreign key**, on either dialect. A correlated subquery runs once per parent row, so without one every run scans the child table. `Account.userId` in the template has no index — so the unindexed column is what an author gets by default:

| Dialect | plain | `_count` | `_count` +index | include+`.length` | `exists` | `exists` +index |
| --- | --: | --: | --: | --: | --: | --: |
| sqlite | 74.0 | 615.5 | **97.3** | 252.6 | 346.0 | **86.9** |
| postgres | 247.5 | 1093.7 | 1118.0 | 1651.9 | 299.7 | 291.1 |

One `@@index([userId])` is worth **6.3× on `_count`** and **4.0× on an `exists` filter** on SQLite — and it reverses the advice. Unindexed, `_count` is 2.4× *slower* than loading the children and counting them in JavaScript. Indexed, 2.6× faster.

**On Postgres the index changes nothing measurable, and I do not think that means what it looks like.** The child table is 200 rows over loopback, so the round trip dominates and the scan is free at that size. The report says that rather than letting the row read as "Postgres does not need one" — same caveat class as the loopback note the lateral work carries.

Two method notes:

- **`_count` is compared against loading the children and counting them in JS**, not against a per-parent count query. Nobody writes the latter; measuring against it would flatter the feature by comparing it to a strawman.
- **The conclusion sentences are derived, including their direction** — `faster`/`slower` is chosen by the numbers, not typed. This report has drifted from its own tables three times and the fix each time was to stop writing prose that could.

`docs/orm.md` carries the index advice with the schema line. A feature whose cost swings 6× on a line the author has to write themselves is not a footnote.

The two things I flagged earlier are still open for you — the `every` inversion most of all.

nstfkc:
`e57b638` — **ordering by a relation**, the last thing this compiler still said was "not implemented yet".

```ts
orderBy: { organization: { name: "asc" } }   // to-one: a field
orderBy: { accounts: { _count: "desc" } }    // to-many: the count
```

Third use of the same correlated subquery — `exists` for a filter, `count(*)` for `_count`, a scalar in `order by` for this. A subquery rather than the `left join` Prisma emits, because a join changes what the `from` clause contains and, on a to-many, the row count. An ordering must not be able to do either.

**Eight differential cases, and they matter more here than elsewhere.** The two clients emit structurally different SQL, so the only thing worth checking is that the *orderings* agree — reading either statement would not tell you. They include the negative-`take` path, where every term flips and the page reverses back.

## The one place Prisma's grammar has no slot for a scope

Worth your attention, because it is a new shape rather than a fourth instance of the old one.

An `include` node takes a `where`. A relation filter's operand *is* one. A `_count` entry takes one. **`orderBy: { rel: { field: "asc" } }` takes nothing** — Prisma has no such key.

It still needs scoping. Ordering by a column of rows the caller cannot read leaks their contents by comparison: sort ascending, sort descending, and an invisible row's position tells you where its value falls. Ordering by `_count` leaks the number outright.

So the policy walk **writes** a `where` into that node and the compiler reads it. The key is not in Prisma's `orderBy` type, so the generated bases — which type `args` as Prisma's own — cannot accept it from application code. The policy pass is its only producer, which is what stops it becoming a way to hand-write a scope. If you think that is the wrong trade, this is the moment: the alternative I considered was refusing relation orderings on policied models outright, which is safe and blocks a legitimate query.

Two things that would have been silent, both caught by writing the test from the property rather than the example:

- **The binder path must be indexed for the array form and not for the object form.** Wrong either way, a scope's parameters bind `undefined` and the ordering still produces valid SQL.
- **Two relation terms need distinct aliases** (`_o0`, `_o1`), or the second subquery's correlation resolves against the first.

| Verified at `e57b638` | |
|---|---|
| `vitest run orm/` | 680 pass |
| template, SQLite | 365 pass, 26 skipped |
| differential matrix | 164 cases vs Prisma |
| template, Postgres 16 | 547 pass |

That closes every "not implemented yet" in the read surface except implicit many-to-many, which is refused in three places now and recorded in the plan as one item.

nstfkc:
`bd88732` — **implicit many-to-many**, which closes the last "not implemented yet" on the read surface. The interesting part is that the fix was not about m-n.

Filters, `_count` and `orderBy` each refused it separately, with three different messages. Same reason every time: the join table needs a **second table inside the subquery**, and none of the three had anywhere to put one — each carried its own inline copy of "resolve the link, quote the two key columns, write the correlation".

So it became one function. `correlate()` returns the three pieces every correlated subquery over a relation needs, and the join table is a difference in exactly one of them:

```
from "Account" as "_r0"                        where "_r0"."userId" = "User"."id"

from "Tag" as "_r0"
  join "_PostToTag" as "_r0j" on "_r0j"."B" = "_r0"."id"
                                               where "_r0j"."A" = "Post"."id"
```

**All three call sites got shorter**, three duplicated key-column lookups went with them, and m-n became a property of one function rather than a capability each caller had to grow. Worth naming because the alternative — a join-table branch in each of the three — was available and would have worked, and would have left the next relation shape needing three more.

Exercised against a real database in the Post/Tag fixture rather than only in the compiler: a filter through the join table in all three operator forms, `every`'s vacuous case, `_count` plain and filtered, ordering by the count both ways. The emitted SQL is asserted in the compiler's tests; what those cannot say is whether it returns the right rows, and the template's schema has no m-n for the differential harness to reach.

Still refused, and it is the artifact's limit rather than the compiler's: a **self-referential** implicit m-n. Same model on both ends, so the generated record cannot say which join column is which — Prisma disambiguates by field order, which the artifact does not carry. That refusal already lived in `resolveLink`; it is now *reached* rather than duplicated, so there is one message for it instead of four.

| Verified at `bd88732` | |
|---|---|
| `vitest run orm/` | 682 pass |
| template, SQLite | 371 pass, 26 skipped |
| template, Postgres 16 | 559 pass |

**The read surface has no remaining "not implemented yet" except the self-referential m-n above.** What is left on the plan is the two entries that are decisions rather than work — MySQL/MariaDB staying deferred, and whether the template's `app/config/auth.ts` should point at the ORM adapter.

This PR has grown to four commits (relation filters, `_count`, benchmarks, ordering, m-n). Say the word if you would rather review them as separate PRs and I will split it.

nstfkc:
`7c48cd8` — **self-referential implicit many-to-many**, the last refusal on the read surface. It was the generated artifact's limit rather than the compiler's, so the fix is in the generator.

**And the note we had been carrying was wrong in the dangerous way.** `plan-relations.ts` said Prisma "disambiguates by field order". Half right: it is field **name** order, not declaration order.

Those two rules agree on every schema whose relation fields happen to be declared alphabetically and disagree on the rest — so guessing would have passed its own tests and reversed the relation on somebody else's schema. Plausible rows, wrong direction, nothing failing.

Established by experiment against a generated Prisma 6 client, connecting through each field in turn and reading the join table:

```
model Thing { zeta Thing[] @relation("Link")    // declared first, sorts second
              alpha Thing[] @relation("Link") }

one.zeta.connect(two)      ->  _Link (A=2, B=1)   owner of `zeta`  in B
three.alpha.connect(four)  ->  _Link (A=3, B=4)   owner of `alpha` in A
```

The alphabetically-first field's owner is in `A`. The generator emits it per field as `joinTable.ownerColumn`.

**Optional rather than required, deliberately.** An artifact generated before this raises — naming `prisma generate` — rather than guessing. Adding the key changes nothing for any relation that already worked, so no artifact version bump and nobody regenerates for a case that previously threw.

The fixture declares `zeta` before `alpha` on purpose: under declaration order the columns come out swapped, so a wrong rule shows up as the two directions **exchanged** rather than as empty results. That is the failure worth shaping a test around — a reversed relation is the one that looks like it works.

| Verified at `7c48cd8` | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 373 pass, 26 skipped |
| template, Postgres 16 | 563 pass |

**That is every "not implemented yet" on the read surface closed**, except `_count` on a to-one, which is a refusal on purpose rather than a gap. What remains on `plans/orm` is the two entries that are decisions rather than work:

1. **MySQL / MariaDB staying deferred** — the dialect seam is there and `DatabaseManager` infers all four, but only SQLite and Postgres are built and tested. The plan asks you to confirm that is acceptable.
2. **The template's `app/config/auth.ts`** still points at the Prisma adapter. `OrmAuthenticationAdapter` has been ready since #53; switching is a behaviour change to a working application and wants your call, not mine.

Both have been waiting a while, and neither is something I should decide.

nstfkc:
`1fcd9c4` — **`delete` with an `include` on a cascading relation** returns the children again. Iteration 4 shipped this as a known divergence and named its own unblocking condition:

> Not fixable at this layer. The fix is to read the relations before the delete inside one transaction, which is iteration 5's to provide — and once it does, **this is the case to write the test for**.

Iteration 5 provided it five iterations ago and nobody went back. `Model.$exec` now reads the caller's projection first and deletes second, inside one transaction — a savepoint when the caller already has one — and returns what it read.

**This was a wrong answer with no error.** A cascading `include` came back `[]`, which reads as "this row had no children" rather than as a failure.

Three decisions the original note did not anticipate:

- **The miss is reported as a `delete`.** The pre-read is a `findFirst` underneath, and an error naming an operation the caller never issued is worse than no error.
- **A `delete` with nothing to read is untouched** — one statement, no transaction. Opening one unconditionally would put a `BEGIN` around every delete in the framework, which is the argument `$exec` already makes for not opening one per write.
- **The pre-read is scoped as the delete was**, nested policies included, rather than re-scoped as a read. These are the rows the delete is about to remove.

## The test was wrong before it was right, in the way that matters

The template declares no cascades — which is why the differential harness never saw this — so it needed a fixture. And then: **`PRAGMA foreign_keys = ON` is per connection.** Setting it only on the test's raw handle left the ORM's own connection with cascades off, and the suite passed. For the wrong reason: with no cascade the children survive the delete, so reading them afterwards finds them and **the bug does not reproduce at all**. A green suite that never exercised the thing it was written for.

Verified the tests fail without the fix — three of six, including the rollback case.

| Verified at `1fcd9c4` | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 379 pass, 27 skipped |
| template, Postgres 16 | 575 pass |

Two of iteration 4's known differences remain, both also marked "not before iteration 5" and neither returning a wrong answer today — `createMany` chunking (raises `ParameterLimitError`) and `upsert` refusing a `create` that omits the conflict key. Those are next unless you would rather I stopped adding to this branch, which is a fair thing to want at seven commits.

nstfkc:
`fa7c475` — **`createMany` splits itself** past the driver's parameter ceiling. Second of iteration 4's three "not before iteration 5" items.

> **No automatic chunking.** Prisma splits a large `createMany`; gemi raises `ParameterLimitError`… Chunking means several statements, which cannot be made atomic before iteration 5.

The chunks run inside one transaction, and the caller gets the `{ count }` a single statement would have returned.

**The assertion iteration 4 was waiting for is not "all the rows arrive."** It is what happens when a later chunk fails — several statements that are not atomic leave the first chunk written, which is a *worse* answer than the refusal they replace: a partial insert behind a thrown error. The test puts the duplicate in the second chunk, so the first is already written when the failure lands, and asserts the table is empty afterwards. A second test does it inside a caller's transaction, where the chunks are a savepoint.

Two details easy to get wrong in the direction that still fails:

- **The split is a `catch`, not a size check.** The ceiling lives in `render`, so the ordinary path pays nothing — a `createMany` small enough to compile never enters the branch. A pre-flight size check would cost every write to serve the one that is too big.
- **The chunk size comes from binding one row**, not from dividing `required`. Division looks equivalent and is not: `required` counts everything the statement binds, so on any shape with fixed overhead it understates the per-row cost and yields a chunk that is *still* over the limit.

`ParameterLimitError` stays, and I fixed its message — it had claimed since iteration 4 that "automatic chunking is not implemented", which is now false. On a `createMany` it means splitting cannot help (one row alone is over the ceiling); on a read it means narrowing the query.

Verified the tests fail without it: four of five, including both atomicity cases.

| Verified at `fa7c475` | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 384 pass, 27 skipped |
| template, Postgres 16 | 580 pass |

One of iteration 4's three remains: **`upsert` refusing a `create` that omits the conflict key**, where the note says "Prisma means find-then-write there; expressing that takes a read and a write inside one transaction, which is iteration 5's." Same shape as the two I have just done, so it is next.

Still open and still yours: MySQL/MariaDB staying deferred, and whether the template's `app/config/auth.ts` should point at the ORM adapter.

nstfkc:
`eaeeb99` — **`upsert` whose `create` omits the conflict key** now runs as a read and a write inside one transaction. That is the last of iteration 4's three "not before iteration 5" items.

**Prisma's semantics here are surprising, so I checked them against a generated client rather than reasoning about them:**

```
upsert({ where: { publicId: "does-not-exist" }, create: { email } })
  -> inserts a row whose publicId is GENERATED, not "does-not-exist"
```

The `where` selects; it contributes nothing to the insert. Two differential cases pin both branches, so the claim rests on Prisma's answer rather than mine.

**`on conflict` is kept wherever it works.** It is one atomic statement and read-then-write is not — two callers can both miss and both insert. Prisma's upsert has that race and this inherits it, but only on calls that previously *raised*, so nothing that worked becomes racy. That is the whole reason the fallback is narrow rather than a replacement.

Compiler and runtime share one predicate (`upsertAbsentConflictKey`) instead of two copies of the rule. If they disagreed, either a divertable call would raise or a raising call would be compiled into SQL that can never conflict.

## It broke an invariant the differential harness had written down about itself

Worth flagging on its own. The harness handed `$exec` a `{ unsafe }` stub, reasoning that `$exec` reads exactly `sql.unsafe` and `dialect`, and that **"a test seam the runtime does not know about cannot drift."**

It drifted the moment `$exec` needed `begin` — `pool.begin is not a function`. It is a Proxy now, delegating everything and intercepting one method, which is the conclusion `bench/run.ts` had already reached independently after the identical error. A stub that lists what the runtime uses today is a promise about what it will use tomorrow.

The other half of that bullet — a `create` whose key *disagrees* with the `where` — stays refused, and the plan now records it as a divergence we are **choosing** rather than one waiting on anything. Prisma ignores the `where` and inserts the `create`'s value; someone who wrote two different keys in one call almost certainly meant one of them.

| Verified at `eaeeb99` | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 387 pass, 27 skipped |
| template, Postgres 16 | 583 pass |

**That is every deferred item in `plans/orm` closed** — the read surface, the write surface, and all three of iteration 4's known differences that were waiting on transactions. What is left is the two entries that are decisions rather than work, and they have been waiting several iterations now:

1. **MySQL / MariaDB staying deferred.**
2. **Whether the template's `app/config/auth.ts` should point at `OrmAuthenticationAdapter`.**

Without an answer to either, the next useful thing I can do is review rather than build — so unless you say otherwise I will spend the next pass auditing what is here instead of adding to it.

nstfkc:
`5aa0233` — the audit pass turned up a real one, and it is a **cross-tenant write**.

With every planned item closed I went looking for the failure this project keeps rediscovering — a path that reaches another model's rows without that model's policies. Five have been closed (nested includes, the lateral fold, relation filters, `_count`, relation orderings). The sixth was open, and it is the first on the **write** side:

```ts
// the child's onCreate never ran; the row landed with orgId unset
Folder.create({ data: { code: "ours", notes: { create: { label: "n" } } } })

// the lookup saw every tenant's rows; this attached org 99's folder
Note.create({ data: { label: "n", folder: { connect: { code: "theirs" } } } })
```

Both silent. The first writes a row belonging to nobody. The second attaches another tenant's row and returns successfully.

## The cause was one boolean with no call site

`RelationExecutor.exec` routes every nested operation through the target model's own `$exec` — invariant 1 holding exactly as designed — but always with `markPreScoped`, which says *"this model's policies are already applied to these args."*

For a relation read that is true: `applyNestedPolicies` walks the include tree before the plan key. For a nested write it is false, because **nothing walks `data.<relation>.create`**. So the marker meant "skip policies", and they were skipped.

`preScoped` is now a required parameter on `exec`. Defaulted either way, one of the two callers gets the quiet wrong answer — and they need opposite answers. Fourth parameter in this codebase made required for that reason.

**The observation is broader than the fix, and I would rather it were on the record than in a commit message.** `markPreScoped` is a *capability*, and it was being handed out by position rather than by decision. It was carefully designed to be unforgeable from application code — a module-private Symbol, deliberately not exported, which took three review rounds on #51 — and then applied unconditionally inside the framework. Unforgeable from outside and automatic on the inside is the same mistake one layer in. Anything that can suppress a policy needs its call sites countable.

Four tests, written before the fix and failing on both counts. `orgId` is nullable in the fixture on purpose: `NOT NULL` would turn the first case into a database error, which passes for a reason unrelated to policies and would keep passing if the scope were removed again.

| Verified at `5aa0233` | |
|---|---|
| `vitest run orm/` | 684 pass |
| template, SQLite | 391 pass, 27 skipped |
| template, Postgres 16 | 587 pass |

Recorded in `plans/orm/06-policies.md` under "Found by audit", beside the residual from #57 — that section is now where the policy-surface story lives.

The two decisions are still open and still yours: **MySQL/MariaDB staying deferred**, and **whether the template's `app/config/auth.ts` should point at the ORM adapter**.

nstfkc:
`d92f837` — second finding from the audit, and it is the same shape as the one you caught on #57: **a rule kept after its stated reason stopped being true.**

`assertScopable` refuses to put a scope on an `upsert`, because "its where clause compiles to an `on conflict` target, which is a key rather than a filter". Exactly true of the `on conflict` path — and **false** of the read-then-write path I added one commit earlier, which is three ordinary statements.

So a tenant-scoped model could not upsert at all, including in the shape that is perfectly scopable. The reason had expired and the refusal outlived it.

**Fixed by moving the fallback decision before policies are applied, not by relaxing the check.** The three statements then run as `findFirst`, `create` and `update`, each scoped normally by its own `$exec` — so that branch is not marked pre-scoped, and nothing had to be special-cased in the policy layer. The refusal that remains is the one whose reason still holds, and its message now names the escape rather than only telling an author to give up on upsert.

One consequence worth stating rather than leaving to be found: a policy branching on `context.operation` sees the three operations that actually run rather than the one the caller named. That is the honest reading — the operation really *is* three statements here.

The test covers the half that matters: the read is scoped, so a user in org 7 upserting `{ where: { id: 1 } }` against org 99's row does not find it and **creates** rather than updating somebody else's row. Getting that wrong would have been a cross-tenant *update* dressed as an upsert.

**And the first version of that test was wrong in the usual way** — it passed the conflict key in `create`, so it never took the path it was written for and simply asserted the `on conflict` refusal it was meant to be avoiding. Caught because it failed; it would have been invisible had the refusal not still been there.

| Verified at `d92f837` | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 393 pass, 27 skipped |
| template, Postgres 16 | 589 pass |

Two audit passes, two findings, both in the policy layer and both of the "the mechanism was right, the wiring was not" kind. I will keep going in the same direction unless you would rather I stopped — and the two decisions are still open: **MySQL/MariaDB staying deferred**, and **the template's `app/config/auth.ts`**.

nstfkc:
`9de30cf` — third audit finding, and the biggest of the three. **`redact` was skipped for every nested relation, under both strategies.**

A `redact` policy protected a root query and was silently skipped inside every `include`. Scoped one way, unscoped the other — the failure iteration 6 exists to prevent, arriving through the mechanism built to make it impossible.

## Same root cause as the first finding

`$exec` built the policy *context* inside `if (policies.length > 0 && !preScoped)`, and `applyRedaction` is keyed on that context. Every nested relation read is pre-scoped, so `policy` was undefined for all of them.

`preScoped` means "the scope is already in these args" — not "this model has no policies". Only `applyPolicies` is idempotency-sensitive: re-running it would `AND` the same predicate twice, while re-running `redact` on an already-redacted row is a no-op. The context is built either way now; only the rewrite is skipped.

That is **two of three audit findings caused by the same flag suppressing more than it names.** I said last time that a capability being handed out by position was the general lesson; this is the same lesson a second time, in a place I did not think to look.

## The lateral half needed a different fix

Worth separating, because iteration 9's argument does not carry over. `scope` survives the fold because policies rewrite the *argument tree* before planning, so the scoped `where` lands inside the subquery. `redact` has no argument to rewrite — it is a row transform in the shaping stage — and a folded child never enters the child's `$exec`. So the parent runs the child's `redact` on its behalf, which is why `RelationPlan` now carries the related model's name.

The comment in `$exec` said *"a related row was shaped by its own model's `$exec` and has already been through its own policy's `redact`"*. True of batched, false of lateral, and unchanged since before lateral existed.

The test is Postgres-only and compares the strategies, because lateral is the default there: **this divergence was on in production and off in a SQLite development environment.**

## And the first version of the test was wrong

It drove both strategies inside `Model.asSystem`, which suspends policies for the whole subtree — so it reported a redaction hole in the *root* query too, one that was not there. A test for a policy cannot run in the scope that turns policies off. I nearly reported the false half.

| Verified at `9de30cf` | |
|---|---|
| `vitest run orm/` | 684 pass |
| template, SQLite | 393 pass, 30 skipped |
| template, Postgres 16 | 592 pass |

I also want to flag a process error on my side: the first Postgres run I did for this commit reported 121 failures, and it was my environment — I ran the suite before pushing the schema and regenerating the client. Same mistake I made earlier in this stack. The numbers above are from a clean run.

nstfkc:
`d2cb33e` — fourth audit pass. No new defect this time; two gaps in how the previous three were checked.

**A folded to-one had no test.** `redactFolded` hands the folded value to `applyRedaction`, which normalises one row and many rows the same way — but the only case exercising it was a to-many. "It happens to also accept an object" is true until it is not, and that is twice in this stack that an untested adjacent shape turned out to be the interesting one. Verified it fails without the fix.

Two adjacent assumptions I checked and which **do** hold, recorded rather than changed:

- `rowsOf` returns `[row]` for a single object, so the to-one path was correct.
- `attachRelations` still deletes hidden key columns for folded plans — the `continue` skips only `load`, and the deletion is after the loop. A `select` that omits a join key does not leak it under lateral.

## And a process fix I owe you

**I have misread my own Postgres environment three times in this stack**, including twice in the last two comments. Four steps have to happen in order — a server, `provider = "postgresql"`, `db push`, `prisma generate` — and getting it wrong does not error. The suite runs, the Postgres describes skip or fail, and the number at the bottom looks like a result: 121 failures that are nothing but a client generated for the other dialect.

`app/models/postgres.sh` does the sequence. It polls `pg_isready` instead of sleeping a guessed number of seconds, and restores `provider = "sqlite"` in a trap — on failure too, because a left-over `postgresql` provider then makes the *SQLite* suites fail for an unrelated reason, which is the same trap one step later. `plans/orm/README.md` says why it exists.

Cheaper than remembering, and it means the numbers in these comments stop depending on me getting a four-step ritual right by hand.

| | |
|---|---|
| `vitest run orm/ bin/` | 684 pass |
| template, SQLite | 393 pass, 31 skipped |
| template, Postgres 16 | 593 pass |
| `oxlint` | clean |

Four audit passes: three real findings (nested writes unscoped, a refusal whose reason had expired, `redact` skipped for every nested relation) and one pass that found only test gaps. That is the shape I would expect near the end of a productive audit, so unless you want me to keep going I think the useful work here is done — and the **two decisions are still open**: MySQL/MariaDB staying deferred, and the template's `app/config/auth.ts`.

nstfkc:
`7954fe2` — fifth audit pass, on a different axis: **the plan cache.** `06-policies.md` calls a collision here "a cross-tenant data leak produced by a caching bug", and the three surfaces I added after iteration 9 each gave arguments a new way to change SQL — none with a discrimination test.

Reading `canonicalShape` says all three are keyed correctly, and the tests agree. That is the outcome I expected, and the entries are still worth having: this file exists so a future change to `canonicalShape` cannot quietly merge two of them, and the newest surfaces are the likeliest to be refactored.

Twenty-two new shapes, which get both the key-distinctness and plan-distinctness checks for free. The pairs that matter most are `some` / `none` / `every` against each other — `every` is `not exists` over the *negated* condition, so sharing a plan with `some` would answer close to the opposite question.

## A genuine gap, in the harness rather than the cache

The plan-identity string was text plus binders. A batched relation contributes **no root column** — it is a separate query — so `include: { accounts: true, _count: … }` and `include: { _count: … }` compile to the same root statement, and the test reported a collision that was not one. The keys differ and so do the plans; the identity was too narrow.

It now includes the relation list, which is the same reasoning the comment above it already made about binders ("a plan is the text *and* its binders, not the text alone"). Found by adding cases, which is the argument for adding them.

## And the direction that actually leaks

A second block pins **policy-written arguments key on shape, not on value.** Iteration 6 established that for a root `where`; each of these three surfaces is one more place a policy's fragment lands. Two tenants with same-shaped scopes must share one plan — asserted as one compile and one hit — and differently-shaped scopes must not. Getting the first wrong is a cache that never hits; getting the second wrong is one tenant running another's SQL.

| Verified at `7954fe2` | |
|---|---|
| `vitest run orm/ bin/` | 691 pass |
| template, SQLite | 393 pass, 31 skipped |
| template, Postgres 16 | 593 pass (via `postgres.sh`) |

Five audit passes now: three real defects, one that found only test gaps, and this one that found a harness gap and left the cache alone. The returns have clearly flattened.

**The two decisions remain the only open items on the plan** — MySQL/MariaDB staying deferred, and the template's `app/config/auth.ts`. Neither is mine to make.

nstfkc:
`9d987bc` — docs, and this is a gap my own fixes created rather than one that was always there.

When I wrote the ORM page three passes ago it described a policy system that **was not true in four places**. Nested writes were unscoped, `redact` was skipped for every nested relation, and a scoped model could not upsert at all. Those are fixed now, so the page can make the promise — and it should make it as a *rule*, because the list is longer than anyone would guess and was found one case at a time.

Six paths reach another model's rows, and every one carries that model's policies:

| | |
| --- | --- |
| `include` / nested `select` | the child's scope narrows the child query |
| a folded relation (`lateral`) | the scope lands *inside* the subquery |
| `where: { rel: { some: … } }` | the `exists` subquery is scoped |
| `_count` | you can only count rows you could read |
| `orderBy: { rel: … }` | you can only sort by rows you could read |
| `data: { rel: { create / connect } }` | the child's `onCreate` runs; a `connect` cannot reach a row you cannot read |

Four of those six were broken when this session started.

**`redact` on nested rows is newly true and was undocumented.** It is the one policy member that cannot be expressed as an argument — a row transform rather than a predicate — which is why it applies after shaping, and why the folded case needed its own fix.

Two limits are stated *with* the guarantee rather than away from it: `asSystem` suspends all of it for the whole subtree, and a policy only applies if its class is the registered one. The second is the gap the guarantee genuinely has, and pointing at it beside the table is more honest than letting the table read as unconditional.

There is also a new **upsert** section, since that is the one operation where a scoped model meets a refusal. It gives both paths and records the surprising half of Prisma's semantics that makes the second work — the `where` selects and contributes nothing to the insert, so the created row carries the schema default rather than the key you searched for.

`llms-full.txt` carries both edits; zero content lines of `orm.md` are missing from it. Docs only, nothing to re-run.

That is the last thing I can see to do that does not need an answer from you. **The two open decisions are the only remaining items on `plans/orm`** — MySQL/MariaDB staying deferred, and whether the template's `app/config/auth.ts` should point at `OrmAuthenticationAdapter`.

nstfkc:
`27776df` — sixth audit pass, on the injection invariant. Iteration 2's acceptance criterion 7 ("not one value is inlined into SQL text anywhere") has a test, and it is a **fixed list of shapes that predates** relation filters, `_count` and relation orderings. All three emit SQL, so the invariant applies to them and nothing was checking it.

**The existing digit proxy cannot be extended, and that is the interesting part.** It asserts no digit survives outside an identifier — exactly right for what iteration 2 emits, and too strict for a correlated subquery: a relation filter emits `exists (select 1 from …)`, and that `1` is a structural constant with no relation to any argument.

Both ways of keeping the proxy are worse than a second check:

- emitting `select null` changes the SQL to fit its own measurement;
- `.replace("select 1", "")` puts a hole in a security check, and the next exception goes through the same hole.

So these assert the property instead of a proxy for it: **no supplied value appears anywhere in the text, and every one comes back from `bind`.** The second half is load-bearing — "the value is not in the SQL" is also satisfied by *dropping* the value, which would be a filter that silently matches everything, and a test that only looked for absence would call that a pass.

Fourteen cases across both dialects. The values are deliberately unmistakable (`314159`, `27-hyphen-8`) rather than small integers that could collide with something structural, and one case pins binder **order** across a relation filter, a scalar and a `take` together — the arrangement where a mis-threaded path binds the right values in the wrong places.

Recorded under criterion 7 in `plans/orm/02-scooter-query-surface.md`, which is where someone will look to find out whether that criterion still holds.

| Verified at `27776df` | |
|---|---|
| `vitest run orm/ bin/` | 705 pass |

No template run: this is compiler-only and touches no runtime path.

Six passes: three real defects, one harness gap, two that confirmed invariants and left the code alone. **The two decisions are still the only open items** — MySQL/MariaDB, and the template's `app/config/auth.ts`.

nstfkc:
`c8151dd` — seventh pass, and the first that changes the shipped starter rather than the framework.

The benchmark established that one `@@index` line is worth several times the query on SQLite for `_count` and relation filters — enough to **reverse** the advice about whether to use them. The template's own schema had **no index on any foreign key**. Prisma declares none, so a schema gets one only by asking, and the starter was not asking.

Seven indexes, one per `@relation(fields:)`, with the migration generated by `prisma migrate diff` rather than hand-written and verified by re-running the diff until it reported an empty migration. Generated artifacts are byte-identical — the gemi generator emits `uniques`, not indexes.

## And it silently broke the benchmark that justified it

The suite measured "unindexed" against whatever the schema declared. With the index in place, **both columns would have been indexed**, the derived sentence would have read "the index is worth 1.0x", and the section would have answered a different question than its own heading — while looking like a clean result.

It is an explicit with/without now: drop `Account_userId_idx`, measure, recreate, measure. And it **refuses to run** if any index on `userId` survives the drop, because that failure is one typo in an index name away and would otherwise be invisible.

## Two prose corrections, both mine

- The report claimed the unindexed column "is what an author gets by default". It is the counterfactual now, and says so.
- It reported Postgres ratios near 1.0x with no caveat, which reads as *"indexes do not help on Postgres"*. The honest statement is that this fixture cannot measure it — 200 child rows over loopback, so the round trip dominates and the ratio moves either side of 1.0 on noise. The report now says the table is decisive on SQLite and says **nothing** about a real child table over a real socket.

And `docs/orm.md` no longer quotes multipliers at all. It cited 6.3× and 4.0×; this re-run made them 6.1× and 3.9×. That is exactly the drift the benchmark document learned to avoid three times over by deriving its own sentences, and I had reintroduced it by hand in the docs one pass later. The page now makes the qualitative claim and points at the generated report.

| Verified at `c8151dd` | |
|---|---|
| template, SQLite | 393 pass, 31 skipped |
| template, Postgres 16 | 593 pass |
| `prisma migrate diff` | empty — schema and migrations in sync |

Seven passes: three real defects, one harness gap, two invariant confirmations, and this one — a real improvement that came with two self-inflicted staleness bugs attached, which is a fair indication of where the returns now are.

**The two decisions remain the only open items on `plans/orm`**: MySQL/MariaDB staying deferred, and whether the template's `app/config/auth.ts` should point at `OrmAuthenticationAdapter`.

nstfkc:
`a24ad32` — no new reviews since 01:00, all plan work closed, so this pass is maintenance and a consolidated view rather than another audit.

**Checked first: the stack still applies.** Nine levels deep is the kind of thing that quietly stops being mergeable, so I verified rather than assumed — all six are `MERGEABLE`/`CLEAN` against their bases, no rebase outstanding.

**Then fixed a staleness bug in the plan's own README.** Its stack diagram described three levels, which was accurate the day it was written and would now send someone branching *under* work that has landed on top. Same class of thing I have been correcting in that document's other notes all session — a record kept after the thing it records moved.

The original diagram stays, because it is right about where the work began and which base to branch from. Beneath it, the six PRs in merge order:

| PR | Branch | What it adds |
| --- | --- | --- |
| #53 | `feat/orm-08-eloquent-doorway` | opt-in provenance, `save(row)`, `wrap`, the ORM auth adapter |
| #54 | `feat/orm-09-lateral-strategy` | `LATERAL` + `json_agg`, nested policies on the arg tree |
| #56 | `fix/orm-bytes-container` | `Bytes` is a `Uint8Array` on every dialect; the harness stops erasing containers |
| #55 | `docs/orm` | `docs/orm.md`, both ORM pages linked from all four indexes |
| #57 | `feat/orm-registration-audit` | `assertPoliciesRegistered` |
| #58 | this | relation filters, `_count`, relation orderings, implicit m-n, six audit findings |

## Where this leaves things

Everything `plans/orm` schedules is built and verified. Seven passes past that produced three real defects, one harness gap, two invariant confirmations, and one starter improvement that arrived with two self-inflicted staleness bugs — which is a clear signal the returns are gone.

**I am going to stop adding to this stack.** Continuing to look for work in a finished plan is how the last pass ended up fixing its own mistakes, and six unreviewed PRs is already more than is comfortable to review at once.

The two open items are decisions, not work, and both have been waiting since #53:

1. **MySQL / MariaDB staying deferred** — the dialect seam exists and `DatabaseManager` infers all four, but only SQLite and Postgres are built and tested. The plan asks for confirmation that this is acceptable.
2. **The template's `app/config/auth.ts`** still points at `PrismaAuthenticationAdapter`. `OrmAuthenticationAdapter` has been ready since #53 and is tested against both dialects; switching is a behaviour change to a working application, which is why I have not done it.

Either answer unblocks me. Merging the stack, or telling me to keep going in some direction, works too — but I would rather ask than invent a task.

nstfkc:
No new reviews. Holding on new work as I said last pass, so this is pre-merge diligence on my own stack rather than another audit.

**File footprint.** 76 files across the six PRs, and every one is in `packages/gemi/{orm,bin/orm,auth,kernel}`, `templates/saas-starter/app/models`, `templates/saas-starter/prisma`, `docs/` or `plans/orm/`. Nothing outside those areas.

**No stray artifacts.** Checked specifically for the things a long session leaves behind — scratch probes, `.probe/` directories, `.orig`/`.rej`, and committed databases. None. `prisma/dev.db` is untouched across the whole stack: I restored it after the index work rather than commit an unreviewable binary whose diff would also have carried leftover test rows.

**`packages/gemi/rfc/req.ts` is still untracked**, which is deliberate — it is yours, not mine, and it got swept into my index three times early in this stack before I switched to explicit paths. Worth confirming it stayed out.

**Green at the tip:**

| | |
|---|---|
| `vitest run orm/ bin/` | 705 pass |
| template, SQLite | 393 pass, 31 skipped |
| template, Postgres 16 | 593 pass |
| working tree | clean |

The two SQLite differential describes that report FAIL in the Postgres run are the expected skip-guard — they refuse to run against a client generated for the other dialect, which is the behaviour `postgres.sh` exists to make legible.

Nothing further from me until one of these gets an answer:

1. **MySQL / MariaDB staying deferred** — confirm, or schedule it.
2. **The template's `app/config/auth.ts`** — point it at `OrmAuthenticationAdapter`, or leave it on Prisma.

Merging the stack works too, and #58 has not been reviewed at all yet.

packages/gemi/orm/policy.ts
```diff
@@ -642,5 +680,366 @@ export function applyNestedPolicies(
...
+  // before.
+  const before = andMembers(inner);
+  const after = andMembers(scoped);
+  const added = after.slice(before.length);
+  if (added.length === 0) return inner;
+
+  const restriction =
+    added.length === 1 ? added[0] : { AND: added };
+
+  return { OR: [{ NOT: restriction }, inner] };
```
nstfkc, line 1049: Confirming this line specifically, since it is the one a future reader will stop at — `{ OR: [{ NOT: S }, X] }` does not *look* like "restrict which children are considered", and the temptation to "simplify" it to `AND` will recur.

Working it through in the same direction the compiler does: `every: Y` becomes `not exists (child correlated and not Y)`. With `Y = (not S) or X`:

```
not (not S or X)  ==  S and not X
→ not exists (correlated and S and not X)
→ "no visible child fails X"  ==  "every visible child matches X"
```

which is the intended meaning. Whereas `Y = X and S` gives `not exists (correlated and not (X and S))`, where an invisible child makes `X and S` false and so *counts as a failure* — the scope ends up disqualifying parents on the strength of rows the caller cannot see.

My database check agrees with the algebra: as Alice, `every: { organizationRole: 0 }` returns her even though her only non-matching account is in the tenant she cannot see, and returns Bob vacuously since none of his are visible.

The reason I would want this comment to stay long: the failure is not a crash or a wrong row count in an obvious direction — it is a filter that silently consults invisible data, and the unit tests on `applyNestedPolicies` are the only thing standing over it, since the differential matrix cannot reach a scoped path at all. Worth a line in the comment saying *that* — that this logic is deliberately outside the differential's reach and therefore depends on its unit tests — so nobody later assumes the 147 cases cover it.

nstfkc (commented):
Reviewed `a24ad32`. Recognising that a relation filter is a *read of another model* — and therefore a policy surface — before writing the filter is the whole review in one decision. The framing that this is the same rule arriving from a third direction (batched recursion, then lateral's fold, now a correlated subquery) is the right way to hold it.

**You flagged the `every` inversion as the thing to check hardest, and noted the differential matrix structurally cannot reach it — Prisma has no policies, so nothing in those 147 cases exercises the scoped path. So I tested it against a database**, with the arrangement that discriminates between the two candidate rewrites:

> Alice (org 1) has two accounts — org1/role0 **visible**, org2/role5 **invisible**. Bob has one account, org2/role0, invisible to Alice.

```
as Alice:
  some:  {}                     -> alice          existence of Bob's account not leaked
  some:  { organizationRole: 5 } -> (none)         her own invisible account not matchable
  every: { organizationRole: 0 } -> alice, bob     every *visible* child matches; Bob vacuous
  none:  { organizationRole: 0 } -> bob
```

All four are the "every visible child" semantics, which is the correct reading. The `every` row is the load-bearing one: Alice's only non-matching child is the one in the tenant she cannot see, and she is correctly returned.

**And the scope is doing the work rather than being a no-op** — the same query under `asSystem` gives a different answer, which is what proves the rewrite is live:

```
every { role: 0 } as alice   -> alice, bob
every { role: 0 } as system  -> bob
```

The difference is exactly the invisible role-5 account: it disqualifies Alice when policies are suspended and is not considered when they are not. That is the inversion working.

**`some: {}` returning only `alice` is the finding you named being closed.** It is worth restating why it deserved its own paragraph: the leak is of *existence*, in a response body containing no rows of the leaking model at all. Nobody reviewing a payload would see it.

### Your second question — the to-one shorthand collapsing into `is`

I think this one is safe, and for a stronger reason than "harmless". Rewriting the shorthand into the explicit form is a **normalisation**, and a normalisation can only ever *merge* cache keys, never split them. Merging is dangerous only if the two merged forms mean different things — and here they cannot, because the shorthand is defined as `is`. So the outcome is one plan where there were two, for queries that were always the same query.

The hazard worth ruling out is the opposite direction — a rewrite that made one written form produce *two* keys depending on context, which would be churn rather than a collision. That is not what this does.

Verified both forms return the same rows under a policy, and the cache stayed healthy across the run (7 entries, 7 compiles, 4 hits, 0 evictions).

| Verified at `a24ad32` | |
|---|---|
| `vitest run orm/` | **705 pass** |
| `vitest run app/models/` (SQLite) | **393 pass**, 31 skipped |

(Both higher than the table in the description, which was taken at `e8a0bbc`.)

One note inline on the rewrite itself. Nothing blocking.

nstfkc:
**The five merges did not reach `feat/orm`, and nine PRs now read as shipped when the branch #45 proposes carries iteration 1 alone.** Flagging before anything else gets built on that assumption.

## What I checked

`feat/orm` is at `a06534e` — the merge of #44 — and is **91 commits behind** the stack tip. Every iteration after the first is absent from it:

```
relations   (it3)  MISSING      provenance (it8)  MISSING
writes      (it4)  MISSING      lateral    (it9)  MISSING
transactions(it5)  MISSING      registration audit MISSING
policies    (it6)  MISSING      _count             MISSING
```

## Why

The topology, not the merges. Every PR in this stack targets its **parent**, so merging a child lands its content on the immediate parent and nowhere below. Nothing ever re-merges downward, so content pools one level up from where it started.

Today's five merges landed within 100 seconds, bottom-up, which compounded it — each base was merged before its own child had received the next one:

| branch | bytes fix | docs | audit | `_count` |
|---|---|---|---|---|
| `feat/orm-07-performance` | no | no | no | no |
| `feat/orm-08-eloquent-doorway` | no | no | no | no |
| `feat/orm-09-lateral-strategy` | **yes** | no | no | no |
| `fix/orm-bytes-container` | yes | **yes** | no | no |
| `docs/orm` | yes | yes | **yes** | no |
| `feat/orm-relation-filters` | yes | yes | yes | **yes** |

So the only branch holding the complete ORM is this one, and it is still open.

## Options, none of which I have taken

I have not merged or retargeted anything — pushing 91 commits into `feat/orm` is exactly the kind of thing I should ask about first, and you asked me to leave #45 alone.

1. **Merge `feat/orm-relation-filters` into `feat/orm` directly.** One merge, everything lands, #45 carries the whole ORM. The intermediate PRs stay marked merged and nothing is lost administratively. Cheapest correct fix.
2. **Retarget #58 to `feat/orm` and merge it.** Same result and closes #58 — but its diff becomes 91 commits, which buries the relation-filter work you have not reviewed yet. I would not choose this.
3. **Cascade downward, thirteen merges**, each branch into its base from the top. Preserves the per-iteration history on each branch. Most faithful, most work, and the intermediate branches stop mattering once #45 merges.

I would take (1) after #58 is reviewed — and (1) works whether or not #58 lands first, since it can be repeated.

Worth saying plainly: **this is not a new defect, it is nine PRs old.** It has been true since #44 merged and I did not check it until the merge activity made the state visible. A stack where every PR targets its parent will do this every time, so the shape is worth knowing independently of this one repair.

nstfkc:
`fbd1aca` — thank you for standing up a database and testing the `every` inversion. That was the claim I could least support from reading, and I said so; you closing it independently is worth more than my saying it again.

**Your fixture is the contribution, so I have made it a test rather than only writing the note.** The arrangement is the part that matters:

| | account | visible to alice? |
| --- | --- | --- |
| alice (org 1) | org 1, role 0 | yes |
| alice | org 2, role 5 | **no** |
| bob (org 2) | org 2, role 0 | no |

Alice's only non-matching child is the one she cannot see. A fixture where every child is visible cannot distinguish the two candidate rewrites — and I verified this one does rather than asserting it: **mutating `invertForEvery` to return the ANDed scope fails exactly one case, the `every` one.** The other five pass under both, which is precisely why the discriminating row had to be constructed deliberately.

Six cases, including the two halves you separated: `every` considers only visible children, and the same query under `asSystem` gives a *different* answer — which is what proves the rewrite is live rather than a no-op that happens to agree.

**The comment now says what you asked for**, and I took the wording from your reasoning: this logic is deliberately outside the differential's reach, so nobody reads the matrix being green as coverage of it. It also names the failure mode, since the tempting `AND` is how it comes back — not a crash or an obviously wrong count, but a filter that silently consults rows the caller cannot see.

**On the to-one shorthand, your argument is better than mine.** I offered "harmless"; you gave a reason: a normalisation can only *merge* cache keys, never split them, and merging is dangerous only if the merged forms mean different things — which they cannot, because the shorthand is defined as `is`. That is a property, not a hope, and it also names the hazard worth ruling out (a rewrite producing two keys for one written form, which would be churn). Recorded in the test.

| Verified at `fbd1aca` | |
|---|---|
| `vitest run orm/ bin/` | 705 pass |
| template, SQLite | 399 pass, 31 skipped (was 393) |
| template, Postgres 16 | 599 pass |

Still outstanding and not mine to act on — **`feat/orm` is 91 commits behind and carries iteration 1 alone**, per my previous comment. Merging this branch into `feat/orm` is the one-step fix whenever you want it.

plans/orm/README.md
```diff
@@ -18,9 +18,63 @@ This work is a **third level in an open PR stack**. Branch from
...
+| --- | --- | --- | --- |
+| `feat/orm-07-performance` | no | no | no |
+| `feat/orm-09-lateral-strategy` | yes | no | no |
+| `docs/orm` | yes | yes | yes |
+
+**The check is one command**, and it is worth running before treating a stack as
+shipped, because every PR showing `MERGED` looks exactly like success:
+
+```
+git rev-list --count origin/<trunkward-branch>..origin/<stack-tip>
```
nstfkc, line 85: Confirmed, and three refinements from running it across every level of the stack.

**1. It is worse than "behind" — `feat/orm` has iteration 1 and nothing else.** Its tip is still `Merge pull request #44`, and `git log origin/feat/orm | grep -c 'iteration [4-9]'` is `0`. The tip is **93 commits** ahead of it.

**2. The stall starts at iteration 3, not at the bottom.** Worth recording because it narrows where to look:

```
feat/database-layer               0 commits ahead of feat/orm     <- landed
feat/orm-03-motorcycle           11 ahead of feat/database-layer  <- stalls here
feat/orm-04-writes               11 ahead of feat/orm-03-motorcycle
feat/orm-05-ambient-transactions 12 ahead of feat/orm-04-writes
feat/orm-06-policies             13 ahead of feat/orm-05-…
feat/orm-07-performance          14 ahead of feat/orm-06-policies
feat/orm-08-eloquent-doorway     15 ahead of feat/orm-07-performance
```

Every level holds its child's work and passes none of it down, exactly as the section describes.

**3. The repair really is one merge, and there is evidence for that rather than hope.** Your note says "the repair itself is a merge into `feat/orm`", singular, and I wondered whether the intermediate branches held anything unique — most of them are *not* ancestors of this branch's tip, which looks alarming. They are not ancestors only because each carries the merge commit its own child PR created:

```
feat/orm-03-motorcycle  ← Merge pull request #49 …
feat/orm-06-policies    ← Merge pull request #52 …
feat/orm-09-lateral…    ← Merge pull request #56 …, Merge pull request #55 …
```

That is the complete list of what they hold and the tip does not — merge commits, no unique work. So one merge of the stack tip into `feat/orm` recovers all of it, and the intermediates can be deleted rather than repaired one at a time.

That last point seems worth adding to the section: "nine PRs merged and nothing landed" reads like nine repairs, and the check you give returns a number that reinforces it. Knowing the fix is one merge — and how to confirm that (`git log <tip>..origin/<base>` showing only merge commits) — is the other half of the runbook.

Agreed the repair is not yours to push, since `feat/orm` is #45's head.

nstfkc (commented):
Reviewed `fbd1aca` and `5747941`.

**The `every` guard is now real rather than decorative — I mutation-checked it.** Changing `{ OR: [{ NOT: restriction }, inner] }` to `{ AND: [restriction, inner] }` fails exactly one test, and it is the right one:

```
FAIL  relation-filter-policies > a scoped relation filter
      > every considers only the children the caller can see
      Tests  1 failed | 5 passed
```

So the tempting simplification is now caught by name. 705 orm tests with it restored. The comment saying the differential matrix cannot reach this path is the part I most wanted — "do not read the matrix being green as coverage of this rewrite" is precisely the misreading that was available.

**And the stacked-merge hazard is real. I confirmed it independently with your own command**, and can add three details that sharpen it — inline.

nstfkc:
`5e6c414` — you are right that the section was missing the half that matters operationally, and I would not have found it: the detection command returns a number that reads like *one job per level*, and my note reinforced that by describing the repair in a single clause at the end.

**Verified your claim rather than accepting it**, across all thirteen branches:

```
git rev-list --no-merges --count origin/<tip>..origin/<branch>
→ 0 on every one
```

Zero non-merge commits anywhere. What the intermediates hold that the tip does not is only the merge commits their own child PRs created — exactly as you found. So the repair is one merge and the intermediates can be deleted rather than fixed one at a time.

Both additions are in, as a **runbook** with the two commands separated by what they answer:

- `rev-list --count <trunkward>..<tip>` — *did the stack land?*
- `rev-list --no-merges --count <tip>..<branch>` — *how big is the repair?*

Plus your per-level table, since "the stall starts at iteration 3, not at the bottom" narrows where to look on a stack of any depth, and `feat/database-layer` being level with `feat/orm` is the fact that locates it.

**The observation I would not have made is the ancestry one.** Most intermediates are not ancestors of the tip, which looks alarming and is not — a merge commit is enough to break ancestry while carrying no work. That is the reading available *without* the second command, and it is the expensive one. It is recorded as the reason the second command exists.

Thank you also for mutation-checking the `every` guard independently. Two of us flipping it to `AND` and getting the same single named failure is worth more than my saying the fixture discriminates.

Nothing outstanding from me on this branch. **`feat/orm` is 93 commits behind**; the one merge is yours to make whenever you want it.

nstfkc (commented):
Reviewed `5e6c414`. **Verified the claim rather than the reasoning** — ran the new command across every branch in the stack:

```
git rev-list --no-merges --count 5e6c414..origin/<branch>

  feat/orm                           0
  feat/database-layer                0
  feat/orm-03-motorcycle             0
  feat/orm-04-writes                 0
  feat/orm-05-ambient-transactions   0
  feat/orm-06-policies               0
  feat/orm-07-performance            0
  feat/orm-08-eloquent-doorway       0
  feat/orm-09-lateral-strategy       0
  docs/orm                           0
  fix/orm-bytes-container            0
  feat/orm-registration-audit        0
```

Zero on all twelve, so "the repair is a single merge" holds with evidence behind it.

`--no-merges` is a better instrument than the one I described. I said "inspect the log and check you see only merge commits", which is a judgement call a tired person makes wrong at the end of a long merge session; a command that must print `0` is not. Same answer, no reading required.

The section now carries both halves — *that* it happened, and *how big* the repair is — which is what makes it a runbook rather than a warning. And recording that the alarming reading ("most intermediates are not ancestors of the tip") is the one available without the second command is the right thing to keep: that is exactly where I stopped before checking, and a merge commit breaking ancestry while carrying no work is not obvious until you look.

Diff is `plans/orm/README.md` only, so nothing to re-run.

Nothing outstanding from me on #58.

PR #59, opened 2026-07-28
Policies become a list, and the write halves stop being optional

nstfkc:
Stacked on `feat/orm` (#45), like #58 was.

## `$policy` → `$policies`

A model now carries a list. The single slot left two ways to compose and both were traps: spreading two objects that each carry a `scope` silently keeps only the last, and the intermediate-base-class route needs one throwaway class *per model*, because every generated base is a different class.

```ts
static $policies: UserPolicy[] = [softDeletes<User>(), new TenantPolicy()]
```

Each level of the prototype chain contributes its own array and they concatenate **base first**, so a shared base can still impose a policy a subclass can only narrow.

Entries may be objects or classes — a class is instantiated once, with stable identity, because `$exec`'s divergence guard and `assertPoliciesRegistered` compare chains element by element.

## Typed policies

`ModelPolicy<TWhere, TCreate, TRow>` is generic, and the generator emits `<Model>Policy` and `<Model>ScopedPolicy` per model off Prisma's own types. The annotation is load-bearing rather than cosmetic: TypeScript does not contextually type an initializer from an inherited static declaration, so a bare `static $policies = [{ scope: (ctx) => … }]` leaves `ctx` an implicit `any`. Extending `<Model>ScopedPolicy` makes `scope`/`onCreate`/`onUpdate` abstract members, so a half-written policy is `TS2515` at the declaration.

## Two write guards, asked per policy

**The create guard was defeatable.** `assertCreateCovered` was a `some()` over the whole chain, so one policy could satisfy it on another's behalf — and `softDeletes()` ships a pass-through `onCreate` precisely so its own scope does not trip it. Adding soft deletes to a model therefore disarmed the guard for that model's tenant policy and the insert ran with the scoped column unset: the exact outcome the error message describes, produced by the guard meant to prevent it.

| policies on `create` | before | after |
| --- | --- | --- |
| `[tenantScope]` | throws | throws |
| `[softDeletes(), tenantScope]` | **passes, column unset** | throws |

**`onUpdate` closes the same hole on the other write.** A `scope` narrows *which rows* an update may touch and says nothing about the values written, so a tenant-scoped update could hand one of its own rows to another tenant — after which the caller can neither read it back nor put it right. Writing a column the policy's own `scope` selects on now raises `ScopeEscapeError` unless the policy has an `onUpdate`. `onUpdate: (_ctx, data) => data` is a complete answer; the point is that the policy says which.

Scope-owned columns are read off the top level of the returned fragment, so `{ organizationId: 7 }` is understood and `{ OR: [...] }` is not attributed to any column and is not checked. That limit is documented and pinned by a test so changing it is a decision.

`plans/orm/06-policies.md` §5 asked for exactly this — *"define what `scope` means per operation rather than assuming the read semantics generalise"* — and it was answered for `create` and `delete` and left open for `update`.

## Multi-statement writes are atomic on their own

A nested write step runs the *child's* `$exec`, so a child policy is consulted mid-sequence, after the parent row is already written. A denial used to leave the parent behind — "your policy stopped the write" and "your policy stopped half the write" are not the same promise.

`$exec` now opens a transaction for exactly the calls whose compiled plan has `before`/`after` steps. A plain `create` still compiles to one statement and opens nothing (asserted), and inside a caller's own transaction it takes a savepoint. The rollback test fails without the change — verified by disabling the flag.

## Also

`softDeletes<M>()` is generic, so `field` is constrained to the model's own keys: pointing it at a column that does not exist is a compile error rather than `no such column` on the first read.

## Verification

- sqlite: **401 passed**, 31 skipped
- Postgres via `app/models/postgres.sh`: **601 passed** — the 2 failures are the dialect guard refusing to run the sqlite differential against a Postgres-generated client, which is by design
- `tsc --noEmit`: clean on the package; template app has the same 45 pre-existing errors before and after
- `oxlint`: 90 warnings before and after
- The 7 failing router/translation tests in the package suite are pre-existing on `feat/orm` (confirmed against a stashed tree)

## Breaking

`static $policy` is gone rather than kept as sugar — #45 is unmerged, so it costs nothing now and could not be done later. All call sites in the package, the template suite and the benchmark harness are migrated.

### DIFF
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -471,18 +471,22 @@ export interface ${model}Model extends Prisma.${model}GetPayload<{}> {}
  * declaration, but the author gets no help writing the body. Naming the type is
  * what turns the hook into a typed one.
  *
- * `${model}Policy` is a class rather than a `type` so that it serves both uses:
- * a class is also a type, its members are all optional through `Policy`'s merged
- * interface, and it carries no private members — so an object literal satisfies
- * it structurally and `extends` works for anyone who prefers methods.
+ * `${model}Policy` is a `type` and `${model}ScopedPolicy` is a class, which is
+ * the split their uses actually want. The scoped one *has* to be a class —
+ * abstract members are the entire mechanism. The plain one is only ever an
+ * annotation (`static $policies: UserPolicy[] = [...]`), and emitting it as a
+ * class would put two runtime declarations plus a value import of `Policy` into
+ * every generated `models.ts`, once per model, to serve a use nobody has: an
+ * author who wants methods without the mandatory write halves extends
+ * `Policy<…>` from `gemi/orm` directly.
  */
 function policyTypes(model: string): string {
   return `
-export abstract class ${model}Policy extends Policy<
+export type ${model}Policy = ModelPolicy<
   Prisma.${model}WhereInput,
   Prisma.${model}CreateInput,
   Prisma.${model}GetPayload<{}>
-> {}
+>;
 
 // The same shape with \`scope\`, \`onCreate\` and \`onUpdate\` abstract, so a policy
 // that scopes cannot omit the write halves. The runtime refuses those
@@ -600,9 +604,9 @@ export function emitModelsFile(schemas: ModelSchema[]): string {
     `\nimport type { Prisma } from "@prisma/client";\n`,
     `import {
   Model,
-  Policy,
   ScopedPolicy,
   type ExecOptions,
+  type ModelPolicy,
   type PolicyEntry,
 } from "gemi/orm";\n`,
     `\nimport * as schema from "./schema";\n`,
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -454,6 +454,48 @@ export interface ${model}Model extends Prisma.${model}GetPayload<{}> {}
 `;
 }
 
+/**
+ * The model's policy types, emitted so an application never writes
+ * `ModelPolicy<Prisma.UserWhereInput, Prisma.UserCreateInput, …>` by hand.
+ *
+ * Two names because the two authoring forms need different things, and one name
+ * cannot be both a class to extend and a plain type alias:
+ *
+ *   static $policies: UserPolicy[] = [{ scope: (ctx) => … }]   // object form
+ *   class Tenant extends UserScopedPolicy { … }                // class form
+ *
+ * The annotation on the object form is load-bearing rather than decorative.
+ * TypeScript does not contextually type an initializer from an inherited static
+ * declaration, so a bare `static $policies = [{ scope: (ctx) => … }]` leaves
+ * `ctx` an implicit `any` — the return type is still checked against the base's
+ * declaration, but the author gets no help writing the body. Naming the type is
+ * what turns the hook into a typed one.
+ *
+ * `${model}Policy` is a class rather than a `type` so that it serves both uses:
+ * a class is also a type, its members are all optional through `Policy`'s merged
+ * interface, and it carries no private members — so an object literal satisfies
+ * it structurally and `extends` works for anyone who prefers methods.
+ */
+function policyTypes(model: string): string {
+  return `
+export abstract class ${model}Policy extends Policy<
+  Prisma.${model}WhereInput,
+  Prisma.${model}CreateInput,
+  Prisma.${model}GetPayload<{}>
+> {}
+
+// The same shape with \`scope\`, \`onCreate\` and \`onUpdate\` abstract, so a policy
+// that scopes cannot omit the write halves. The runtime refuses those
+// combinations too — this makes it \`TS2515\` at the class declaration instead of
+// an error on the first write that reaches production.
+export abstract class ${model}ScopedPolicy extends ScopedPolicy<
+  Prisma.${model}WhereInput,
+  Prisma.${model}CreateInput,
+  Prisma.${model}GetPayload<{}>
+> {}
+`;
+}
+
 function operation(model: string, op: ReadOperation): string {
   const argsType = `Prisma.${model}${op.args}`;
   const returns = op.returns(model);
@@ -556,7 +598,13 @@ export function emitModelsFile(schemas: ModelSchema[]): string {
   const parts = [
     HEADER,
     `\nimport type { Prisma } from "@prisma/client";\n`,
-    `import { Model, type ExecOptions } from "gemi/orm";\n`,
+    `import {
+  Model,
+  Policy,
+  ScopedPolicy,
+  type ExecOptions,
+  type PolicyEntry,
+} from "gemi/orm";\n`,
     `\nimport * as schema from "./schema";\n`,
     `
 // Copied from Prisma's own generated client. It is what makes \`select\` and
@@ -571,9 +619,19 @@ type Subset<T, U> = {
 
   for (const schema of schemas) {
     parts.push(instanceShape(schema.name));
+    parts.push(policyTypes(schema.name));
     parts.push(`
 export class ${schema.name}Model extends Model {
   static $schema = schema.${schema.name};
+
+  // Narrowed from \`Model\`'s \`PolicyEntry[]\` to this model's own, so a policy
+  // written for another model is a type error here rather than a scope compiled
+  // against columns that do not exist.
+  static $policies?: readonly PolicyEntry<
+    Prisma.${schema.name}WhereInput,
+    Prisma.${schema.name}CreateInput,
+    Prisma.${schema.name}GetPayload<{}>
+  >[];
 ${READ_OPERATIONS.map((op) => operation(schema.name, op)).join("")}${countOperation(
       schema.name,
     )}${WRITE_OPERATIONS.map((op) => operation(schema.name, op)).join(
--- a/packages/gemi/orm/policy.ts
+++ b/packages/gemi/orm/policy.ts
@@ -327,6 +327,15 @@ const INSERTING = new Set<string>(["create", "createMany"]);
  * call would make every comparison fail — every query on a class-form policy
  * would raise `UnregisteredPolicyClassError` against itself. Keyed on the
  * constructor, which is as stable as the class object.
+ *
+ * Keyed on the constructor and nothing else, so a policy class listed in two
+ * models' `$policies` is **one instance shared across both** — once per process,
+ * not once per model. Correct for every policy that exists today, all of which
+ * are stateless, and the reason the class form is documented as a place for
+ * behaviour rather than for state: a policy that memoised anything per model
+ * would silently serve one model's cache to the other. A policy that genuinely
+ * needs per-model state should be a factory returning a fresh object, which is
+ * what `PolicyEntry`'s other half is for.
  */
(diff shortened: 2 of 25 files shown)


packages/gemi/orm/policy.ts
```diff
@@ -196,34 +319,89 @@ const CREATING = new Set<string>(["create", "createMany", "upsert"]);
...
+ * **Deliberately uncached.** This runs per query per node of an include tree, so
+ * a cache is tempting — but `$policies` is an ordinary static that tests and
+ * feature flags reassign at runtime, and a cache keyed on the class would serve
+ * the stale chain with no way to notice. The single-level fast path below is
+ * what keeps the common case free: one `Object.hasOwn` per level and no
+ * allocation at all when nothing needs instantiating.
  */
-export function policiesFor(model: unknown): ModelPolicy[] {
-  const found: ModelPolicy[] = [];
+export function policiesFor(model: unknown): readonly ModelPolicy[] {
```
nstfkc, line 375: **A leftover `static $policy` is now silently unpolicied.** Confirmed:

```ts
class Legacy { static $policy = { scope: () => ({ orgId: 7 }) } }
expect(policiesFor(Legacy)).toEqual([])   // passes
```

No type error either — `Model` no longer declares `$policy`, so a subclass writing it is just adding a static nobody reads.

This PR migrated every call site, so the tree is correct. But the failure mode of missing one is "the tenant scope stops applying, quietly", and the next rebase across this commit is where that happens. Suggest the walk throws when it sees the old name:

```ts
if (Object.hasOwn(current, "$policy")) {
  throw new Error(
    `${(current as any).name ?? "model"} declares \`$policy\`, which is now \`$policies\` ` +
      `and takes a list. The old name is not read, so this model is currently unpolicied.`,
  );
}
```

Deletable once `feat/orm` merges and the window closes. It is the same argument `UnregisteredPolicyClassError` already makes one function over: a policy that is present but not in effect has to be loud.

packages/gemi/orm/policy.ts
```diff
@@ -275,6 +462,12 @@ export function applyPolicies(
     }
   }
 
+  if (MUTATING.has(context.operation)) {
```
nstfkc, line 500: **Ordering gap: the guard inspects the caller's `data`, this loop then rewrites it.** `assertNoScopeEscape` runs in the scope loop above, before any `onUpdate` has been applied — so a policy's `onUpdate` can write a column a *different* policy's scope owns, and nothing checks it. Verified:

```ts
const injector: ModelPolicy = { onUpdate: (_c, d) => ({ ...d, orgId: 999 }) }
const tenant: ModelPolicy = { scope: () => ({ orgId: 7 }) }   // no onUpdate

applyPolicies([injector, tenant], ctx("update"), { where: { id: 1 }, data: { name: "x" } })
// => { data: { name: "x", orgId: 999 } }   — no ScopeEscapeError
```

This is the same shape as the `assertCreateCovered` bug this PR fixes — one policy acting on another's behalf — just arriving through the rewrite rather than through a `some()`. It is narrower in practice (it needs a policy that writes a column it does not itself scope, which is unusual), so I would not hold the PR for it, but the current comment on `assertNoScopeEscape` says the check is "whether the columns this scope constrains are being written" and that is only true of the *caller's* write.

If it is not worth closing, it is worth a sentence in `ScopeEscapeError`'s "what it cannot see" section, which already does this honestly for combinator scopes.

packages/gemi/orm/policy.ts
```diff
@@ -196,34 +319,89 @@ const CREATING = new Set<string>(["create", "createMany", "upsert"]);
...
+/**
+ * Instances made from constructor entries, so identity is stable across calls.
+ *
+ * Not an optimisation. `$exec`'s divergence guard and `assertPoliciesRegistered`
+ * both compare two policy chains element by element, and a fresh instance per
+ * call would make every comparison fail — every query on a class-form policy
+ * would raise `UnregisteredPolicyClassError` against itself. Keyed on the
+ * constructor, which is as stable as the class object.
+ */
+const instantiated = new WeakMap<object, ModelPolicy>();
```
nstfkc, line 340: Keyed on the constructor and global, so one policy class listed in two models' `$policies` yields **one shared instance across both models**. Right for the stable-identity requirement and right for stateless policies, which is all of them today — but the class form is newly promoted by this PR, and "instantiated once" in the docs reads as once per model rather than once per process. Worth one clause saying the instance is shared, since a policy that memoises anything per-model would be quietly wrong.

packages/gemi/bin/orm/emit.ts
```diff
@@ -454,6 +454,48 @@ export interface ${model}Model extends Prisma.${model}GetPayload<{}> {}
...
+ * `ctx` an implicit `any` — the return type is still checked against the base's
+ * declaration, but the author gets no help writing the body. Naming the type is
+ * what turns the hook into a typed one.
+ *
+ * `${model}Policy` is a class rather than a `type` so that it serves both uses:
+ * a class is also a type, its members are all optional through `Policy`'s merged
+ * interface, and it carries no private members — so an object literal satisfies
+ * it structurally and `extends` works for anyone who prefers methods.
+ */
+function policyTypes(model: string): string {
```
nstfkc, line 483: Two runtime `abstract class` declarations per model, plus a value import of `Policy`/`ScopedPolicy` into every generated `models.ts`. `<Model>ScopedPolicy` has to be a class — the abstract members are the whole point. `<Model>Policy` does not: every documented use of it is as a type annotation (`static $policies: AccountPolicy[] = [...]`), and the docs never show anyone extending it.

The comment justifies the class form with "`extends` works for anyone who prefers methods" — but that is what `<Model>ScopedPolicy` is for, and someone who wants methods without the mandatory write halves can extend `Policy<…>` directly. `export type AccountPolicy = ModelPolicy<Prisma.AccountWhereInput, Prisma.AccountCreateInput, Prisma.AccountGetPayload<{}>>` would give the identical authoring experience for the object form and cost nothing at runtime, which matters at N models.

Not a defect — just checking the class was chosen rather than defaulted to, since the file emits both names anyway.

templates/saas-starter/app/models/nested-write-policies.test.ts
```diff
@@ -301,3 +301,123 @@ describe("policies on nested writes", () => {
...
+              return sql.unsafe(text, values);
+            };
+          },
+        });
+      },
+    });
+
+    const application = new Application();
+    application.instance(DatabaseManager, counted as never);
+    Application.setInstance(application);
```
nstfkc: The global `Application` is swapped in mid-test and restored inline rather than in a `finally` or `afterEach`. If the `$exec` below throws — which is exactly what a regression in the atomicity change would do — the proxied `DatabaseManager` stays installed for every test after this one in the file, and the failure surfaces somewhere unrelated.

Also: the restore builds a *fresh* `Application` rather than putting back the one `beforeAll` created, so anything else registered on it is dropped. `afterAll` then restores `previous` over the top, which works today only because nothing else is registered.

Suggest `try { … } finally { Application.setInstance(application) }` against the instance `beforeAll` already holds.

nstfkc (commented):
## Review

The three changes here are independently well-argued, and two of them fix real holes rather than cosmetic ones. The `assertCreateCovered` bug in particular is the good kind of find — a guard that was disarmed by the one policy the codebase ships, and the table in the PR body makes the before/after unambiguous. The transaction change is correctly scoped: keying on `plan.before`/`plan.after` rather than on "is this a write" is what keeps the `BEGIN`-around-everything objection answered, and the single-statement assertion pins it.

**Verified locally on `80940e1`:**

- `tsc --noEmit` on `packages/gemi` — clean
- `oxlint orm/policy.ts` — 0 warnings, so the two suppressions on the `Policy` merge are doing their job
- `orm/policy.test.ts`, `orm/registration.test.ts`, `bin/orm/emit.test.ts` — 125 passed
- template `nested-write-policies.test.ts`, `policies.test.ts` — 50 passed, 1 skipped
- Both documented authoring forms (`static $policies: UserPolicy[] = [softDeletes<User>(), new TenantPolicy()]` and the object literal) compile against the template's generated types
- The claim in `emit.ts` that a cross-model policy is a type error holds — an `AccountPolicy` on a `UserModel` subclass is `TS2417` on the static side

One thing I'd want addressed before merge, plus a doc inconsistency and some smaller notes inline.

### The one blocking-ish thing

Removing `$policy` outright is the right call while #45 is unmerged, but the removal is **silent in both directions**. A class still carrying `static $policy = {…}` compiles clean (the base no longer declares the name, so it is just a new static) and `policiesFor` returns `[]` — the model is unpolicied with nothing to notice it. I confirmed this:

```ts
class Legacy { static $policy = { scope: () => ({ orgId: 7 }) } }
policiesFor(Legacy)  // []  — no error, no warning
```

This PR migrated every call site in the package, the template suite and the bench harness, so the tree is correct today. The risk is the next branch that rebases across this, or a stacked PR on `feat/orm` that still has the old name. Given the failure mode is "tenant scoping silently stops applying", the cheap insurance is worth it — a throw in `policiesFor` (or in `assertPoliciesRegistered`, which already exists to catch exactly this class of "policy present but not in effect" mistake) when `Object.hasOwn(current, "$policy")`. That is the same argument `UnregisteredPolicyClassError` already makes for the sibling case, and it can be deleted once `feat/orm` merges.

### Docs

`plans/orm/06-policies.md` gets the new §5 note but the rest of the file is still on the old API — line 57 and line 198 show `static $policy = {…}`, and the whole passage at 227–247 reasons about `Object.hasOwn(Child, "$policy")` and "make `$policy` a method call". Since that file is the design record this PR cites as having asked for `onUpdate`, leaving the examples on the removed name reads as though the plan predates the change rather than being updated by it. `docs/orm.md` and `docs/llms-full.txt` are fully migrated.

### Test coverage

Everything about `onUpdate` and `ScopeEscapeError` is unit-level, calling `applyPolicies` directly with a hand-built context. The create guard and soft deletes both have template-level tests against a real database; the update guard has none. The interaction I would most want covered end-to-end is the one that mirrors the bug this PR fixes: `[softDeletes(), tenantScope]` on a model, then `softDelete(Model)` issuing `update({ data: { deletedAt } })` and **not** tripping `ScopeEscapeError`, while `update({ data: { orgId } })` on the same model does. That is asserted in `policy.test.ts`, but only against `applyPolicies` — nothing proves it survives the path through `$exec`, `save()` and the plan cache. Same for the Postgres differential, which the PR body notes was run but which cannot see a guard that throws before any SQL is emitted.

plans/orm/06-policies.md
```diff
@@ -116,6 +116,19 @@ affected; scoping a `create` means validating or defaulting the tenant column.
 Define what `scope` means per operation rather than assuming the read semantics
 generalise.
 
+**Answered, and one part of it was missed the first time.** `create` got `onCreate`
```
nstfkc, line 119: This note is added, but the rest of the file still documents the removed API — `static $policy = {…}` at lines 57 and 198, and the passage at 227–247 reasoning about `Object.hasOwn(Child, "$policy")` and "make `$policy` a method call". That last section is specifically about the mechanics of the static this PR renamed, so it is the one most likely to be read as current.

Since this file is cited in the PR body and in `ModelPolicy.onUpdate`'s doc comment as the record that asked for this, having it half-migrated undercuts it. `docs/orm.md` and `docs/llms-full.txt` are fully done.

PR #60, opened 2026-07-28
Warn in development when a transaction holds its connection too long

nstfkc:
Iteration 5 left this as an open note: *"Long-lived transactions hold a pooled connection. Nothing here prevents a callback from doing network I/O while holding one. A development-mode warning above a duration threshold is cheap and worth considering, though not required."*

This is that warning.

## Why it earns its place

An ambient transaction reserves one pooled connection for as long as its callback runs. The API gives no hint that awaiting a `fetch` inside it differs from awaiting a query — which is the point of ambient transactions, and also the trap:

```ts
await User.transaction(async () => {
  const user = await User.create({ data })
  await billing.createCustomer(user)   // holds a connection for the whole round-trip
})
```

Under load the cost lands somewhere else entirely: unrelated queries block on connection acquisition, and the resulting error names neither the transaction nor the I/O that held it. That's the failure this makes traceable.

```
[gemi] A database transaction has not settled after 2000ms. It reserves a pooled connection
until it does — check for network or filesystem I/O inside the callback, and move it outside
the transaction if you find any.
    at Object.<anonymous> (app/services/signup.ts:34:9)
    …
(development only; set GEMI_SLOW_TRANSACTION_MS to change the threshold)
```

## Decisions worth reviewing

**A timer, not a duration measured on the way out.** The case that matters most is the transaction that never settles at all — a hung request, a lock wait, a deadlock. An after-the-fact measurement is silent for exactly that one. The timer fires while the transaction is still in flight.

**The clock spans the pool wait and the commit round-trip, not just the callback.** Deliberate: a transaction stuck waiting on an exhausted pool is the symptom a *different* long transaction creates, so scoping the timer to the callback would go quiet precisely when the warning is most useful. It's also why the message reads "has not settled after" rather than "has been open for" — during the pool wait no connection is held yet, and the wording must not claim one was.

**Outermost scope only.** A savepoint reserves no connection of its own; its lifetime is bounded by the transaction it sits in, which is already being timed. Warning per depth would report one slow block several times over and point at the innermost frame instead of the one actually holding the connection.

**Development only, read per call.** `scripts/build.ts` rewrites `process.env.NODE_ENV` to `Bun.env.NODE_ENV` specifically so mode stays a runtime question in the built framework — caching it at module scope would undo that.

**An unparseable threshold falls back to the default rather than disabling.** A typo in an env var silently switching a diagnostic off is worse than ignoring the typo.

**The timer is `unref`'d**, so a pending warning can't keep a short-lived script or CLI command alive. The `Error` carrying the call site is allocated up front but only formatted if the timer fires — V8 builds `.stack` lazily, so a fast transaction pays for one allocation and nothing else.

`DB.transaction` is covered too, since it already routes through `withTransaction`.

It reports; it does not cancel. Nothing here shortens or aborts a long transaction.

## Tests

Seven, in `orm/context.test.ts`. They drive a real timer against a 10ms threshold rather than vitest's fake clock, because the `unref` behaviour under test lives on the very object a fake clock replaces.

Both negative assertions are load-bearing, confirmed by mutation:

| Mutation | Test that fails |
| --- | --- |
| Remove the `NODE_ENV` gate | "production is silent, however long the transaction runs" |
| Neuter the `.finally(stopWatching)` | "a throwing transaction disarms the warning" |

The second is the one worth keeping: a rollback releases the connection exactly as a commit does, so a throwing callback has to disarm the timer or every failed transaction leaves a warning armed against a connection already back in the pool.

## Verification

- ORM suite: 730 pass
- Template `transactions.test.ts`: 20 pass, 3 skipped
- `tsc` clean; `oxlint` clean (90 warnings, all pre-existing, none in these files)
- **Not exercised against Postgres** — the template's Postgres suite skips without `TEST_POSTGRES_URL`. The change is a JS timer with no SQL in it, so the dialect cannot reach it.

## Note on the base

This is stacked on `feat/orm-policy-list` (`80940e1` + `9b12b17`), so those two commits show here until that branch lands in `feat/orm`. **Only `6b30bb9` is this PR's.**

One crossover to know about: the `Model.transaction` doc comment pointing at this warning is in `9b12b17` rather than here — the two branches were being written at the same time and that hunk landed a commit early.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3044,14 +3044,23 @@ opened it:
 until it does — check for network or filesystem I/O inside the callback…
 ```
 
-Set `GEMI_SLOW_TRANSACTION_MS` to change the threshold. There is no separate off-switch: a seed
-script or a dev-time migration with legitimately long transactions silences the warning by setting
-a threshold longer than they take. A value that is not a positive finite number — `0`, `-1`,
-`Infinity`, a typo — falls back to the 2-second default rather than disabling, so it cannot be
-switched off by accident.
+The threshold is `slowTransactionThreshold` in `app/config/database.ts`:
 
-The warning is development-only and never fires in production. It is a diagnostic, not a limit:
-nothing cancels a long transaction.
+```typescript
+export default defineDatabaseConfig({
+  url: process.env.DATABASE_URL,
+
+  slowTransactionThreshold: 5_000,   // milliseconds, or `false` to switch it off
+});
+```
+
+`false` is the only way to disable it. Any other unusable value — `0`, a negative, `Infinity`,
+`NaN` — falls back to the 2-second default rather than turning the warning off, so it cannot be
+lost to a mistake. Say `false` when you mean off; raise the number for a seed script or a data
+migration whose transactions are legitimately long.
+
+The warning is development-only and never fires in production, whatever the config says. It is a
+diagnostic, not a limit: nothing cancels a long transaction.
 
 ## Policies
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3029,6 +3029,30 @@ still compiles to one statement and opens nothing, and inside a transaction you 
 one becomes a savepoint. You do not need `Model.transaction` to make a single nested write whole;
 you need it to make *several separate calls* whole.
 
+### Keep slow work out of the callback
+
+The reserved connection is held for as long as the callback runs, whether or not it is running
+queries. A `fetch`, an upload or a queue push inside a transaction holds a connection for the
+length of that I/O, and under load the symptom is unrelated queries elsewhere in the process
+blocking on connection acquisition — an error that names neither the transaction nor the I/O.
+
+In development, a transaction still open after 2 seconds warns with the stack of the call that
+opened it:
+
+```
+[gemi] A database transaction has not settled after 2000ms. It reserves a pooled connection
+until it does — check for network or filesystem I/O inside the callback…
+```
+
+Set `GEMI_SLOW_TRANSACTION_MS` to change the threshold. There is no separate off-switch: a seed
+script or a dev-time migration with legitimately long transactions silences the warning by setting
+a threshold longer than they take. A value that is not a positive finite number — `0`, `-1`,
+`Infinity`, a typo — falls back to the 2-second default rather than disabling, so it cannot be
+switched off by accident.
+
+The warning is development-only and never fires in production. It is a diagnostic, not a limit:
+nothing cancels a long transaction.
+
 ## Policies
 
 A model carries a **list** of policies. Every member of a policy is optional; a model with none
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -327,6 +327,13 @@ await Model.transaction(async () => {
 `Promise.all` over several ORM calls inside the callback is not safe — the ordinary, encouraged
 thing everywhere else in a Bun codebase. Await them in sequence.
 
+**A multi-statement write is atomic on its own.** A write with a nested `create` or `connect` runs
+more than one statement, and `$exec` opens a transaction for exactly those calls — so a nested
+step that fails, or a child policy that denies, rolls back the parent row too. A plain `create`
+still compiles to one statement and opens nothing, and inside a transaction you opened the nested
+one becomes a savepoint. You do not need `Model.transaction` to make a single nested write whole;
+you need it to make *several separate calls* whole.
+
 ### Keep slow work out of the callback
 
 The reserved connection is held for as long as the callback runs, whether or not it is running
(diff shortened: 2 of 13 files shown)


nstfkc:
## Review

Scoped to `6b30bb9` — `80940e1`/`9b12b17` reviewed as base, as the description says.

The core design holds up, and the PR description defends the parts that actually needed defending. Three choices I'd have questioned are already argued correctly in the code:

- **Timer at `begin`, not elapsed-time on the way out.** The never-settles case is the one that matters and it's the one an after-the-fact measurement can't see. Right call.
- **Clock spans the pool wait.** This is the subtle one and the reasoning is sound: a transaction blocked on an exhausted pool is the *symptom* of the problem this warns about, so scoping to the callback would go quiet exactly when it's most useful. The "has not settled after" wording follows from that, and the comment says why.
- **`.finally` rather than `then`/`catch`.** Rollback releases the connection the same as commit. The test covering it is load-bearing (verified below).

`slowTransactionThreshold()` reading `NODE_ENV` per call rather than at module scope is correct and non-obvious — `scripts/build.ts` sets `define: { "process.env.NODE_ENV": "Bun.env.NODE_ENV" }` precisely so mode stays runtime-resolved in the published bundle, and a module-scope cache would defeat it.

---

### 1. `docs/llms-full.txt` was not updated (should block)

`docs/orm.md` gains the "Keep slow work out of the callback" section; `docs/llms-full.txt` mirrors that page and did not. The insertion point exists verbatim there — `llms-full.txt:3021-3024` has the `**One statement at a time.**` paragraph followed immediately by `**A multi-statement write is atomic on its own.**`, with nothing between.

This is the repo's own stated convention (`plans/orm/README.md:495` lists the four surfaces docs must be reachable from), and the base commit `9b12b17` updated `orm.md` and `llms-full.txt` in lockstep. Only this commit broke step.

### 2. The unparseable-threshold test doesn't test what it's named

`orm/context.test.ts` — *"an unusable threshold falls back to the default rather than disabling"*. The comment is candid that the assertion is indirect:

```ts
// The default is 2s, so nothing fires inside this test — the assertion is
// that it fell back to *some* real threshold instead of returning null.
expect(warnings).toEqual([]);
```

It can't carry that. `toEqual([])` is exactly what you observe when the fallback is *absent*. I mutated `context.ts:167` to the failure it claims to guard:

```ts
if (!Number.isFinite(parsed) || parsed <= 0) return null;   // disable instead of fall back
```

All 16 tests still pass. The other two negative assertions in the file do hold under their stated mutations — this is the one that doesn't.

Cheapest fix: export `slowTransactionThreshold` and assert the number directly (`expect(slowTransactionThreshold()).toBe(2_000)`), which also covers `""`, `"0"`, `"-1"` and `"Infinity"` without burning wall-clock. The behavioural alternative — sleeping past 2s — isn't worth 2 seconds of suite time.

### 3. The new heading orphans a paragraph from the base branch

`### Keep slow work out of the callback` was inserted between `**One statement at a time.**` and `**A multi-statement write is atomic on its own.**`. That last paragraph is about nested-write atomicity and now renders *inside* a section about keeping I/O out of callbacks.

It arrived in `80940e1`, so this is a stacking artefact rather than a mistake in the prose — but the insertion point is this PR's. Moving the new heading below that paragraph fixes both files at once (see #1).

---

### Minor

- **A synchronous throw from `pool.begin` leaks an armed timer.** `watchForSlowTransaction()` runs before the `begin` call, and `stopWatching` is only reachable via `.finally` on the returned promise. If `begin` throws synchronously rather than rejecting — pool closed, say — the timer survives and prints a warning about a transaction that never opened. `unref` keeps it from holding the process, so this is cosmetic, but it's the same class of gap the `.finally` comment reasons about carefully. A `try`/`catch` around the `begin` call closes it.

- **`unref` is the justification for real timers but nothing asserts it.** The `describe` block explains that vitest's fake clock is avoided because "the `unref` behaviour lives on the very object a fake clock replaces" — yet no test observes `unref`. Deleting `timer.unref?.()` passes the suite. Either assert it (spy on the returned timer) or drop the claim from the comment, since it's currently the stated reason for the slower approach.

- **Flake margin on the 10ms/40ms test.** *"warns while it is still open"* needs the timer to fire within the 40ms sleep — 4× headroom. Fine locally; on a loaded CI box with the suite running in parallel, 4× is thinner than it looks. A longer sleep costs nothing that matters here.

- **No documented off-switch.** The docs say to set `GEMI_SLOW_TRANSACTION_MS` to change the threshold, but a developer running a seed script or a dev-time migration with legitimately long transactions has no stated way to silence it. A very large value works; it's just not written down. Worth one sentence given `Infinity` falls back to 2s rather than disabling.

### Verification

- `bun --bun vitest run orm/context.test.ts` — 16 pass, ~500ms
- Mutation of `context.ts:167` to `return null` — 16 pass (finding #2)
- Deleting `timer.unref?.()` — 16 pass (minor #2)

Nothing here is a correctness problem in the warning itself. #1 and #2 are the two I'd want addressed before merge; #3 is a one-line move that goes with #1.

nstfkc:
Pushed `baa3557`. Both blocking findings were right; all four minors addressed too.

**#1 `llms-full.txt` not updated** — mirrored. `diff` of the two sections is now identical, and it's the same insertion point you named.

**#2 the fallback test asserted nothing** — confirmed your mutation locally: `return null` at `context.ts:167` passed all 16. You're right that no behavioural test of that fallback can work, because "no warning fired" is what you observe in both the working and the broken case. Took your suggestion: `slowTransactionThreshold` is exported and asserted as a value, which also covers `""`, `"   "`, `"0"`, `"-1"`, `"NaN"` and `"Infinity"` for free. `Infinity` is the one worth having — it's why the guard is `Number.isFinite` and not `!Number.isNaN`, since it would otherwise schedule the timer for never.

**#3 orphaned paragraph** — heading moved below `**A multi-statement write is atomic on its own.**` in both files.

**Minors** — all four:
- Synchronous `begin` throw: `try`/`catch`, with a test that fails without it. You were right that it's cosmetic given `unref`, but it's the same gap the `.finally` reasons about.
- `unref` unasserted: now spied and asserted, so the describe block's justification for avoiding the fake clock is backed.
- Flake margin: 4× → 20×, as a named `PAST_THRESHOLD` rather than five scattered literals.
- Off-switch: documented — set a threshold longer than the work takes, and non-positive/non-finite values fall back rather than disabling, so it can't be silenced by accident.

Mutation battery, all five now fail where they should:

| Mutation | Tests failing |
| --- | --- |
| fallback → `null` | 7 |
| delete `timer.unref?.()` | 1 |
| remove the `try`/`catch` | 1 |
| remove the `NODE_ENV` gate | 2 |
| neuter the `.finally` | 2 |

ORM suite 741 pass, template transactions 20 pass / 3 skipped, `tsc` clean, oxlint at the 90-warning baseline. Postgres still unexercised — no `TEST_POSTGRES_URL`, and the change remains a JS timer with no SQL in it.

Issue #61, opened 2026-07-28
UnsupportedQueryError: separate 'not yet' from 'out of scope', and always say what to use instead


nstfkc:
`UnsupportedQueryError` carries the right structure, and it is right about *what* was refused. Two things it gets wrong are worth fixing before the ORM ships, because both of them cost an application author real time.

Messages below are captured by executing each case against a real database, not read off the source.

## What already works

- Structured, inspectable fields: `argument`, `model`, `operation`, plus `name`.
- The argument path survives nesting: `data.organization.connectOrCreate`, `accounts.take`, `where.name.path`.
- Where a `detail` is supplied, it is genuinely good — this one is the standard the rest should meet:

  > gemi ORM does not support `'accounts.take'` yet (User.findMany). A 'take' inside a to-many relation is per parent, which the batched relation planner cannot express. It arrives with the lateral join strategy; until then, load 'accounts' as its own query, or narrow it with 'where'.

  What was refused, why it cannot work today, when it arrives, and two things to do instead.
- `UnknownFieldError` and `UnknownRelationError` both enumerate the valid names.

## 1. "yet" promises a roadmap that does not exist

`detail` is optional, and the refusals that omit it are the highest-traffic ones:

```
gemi ORM does not support 'distinct' yet (User.findMany).
gemi ORM does not support 'cursor' yet (User.findMany).
gemi ORM does not support 'omit' yet (User.findMany).
gemi ORM does not support 'skipDuplicates' yet (User.createMany).
gemi ORM does not support 'where.name.path' yet (User.findMany).
```

Several of these are not "yet" at all. `docs/orm.md` has them under a heading called **Not in scope**:

> - **No `omit`.** Use `select`, or a policy's `redact`.

and `plan.ts:8-12` says `aggregate`, `groupBy` and the raw operations "are deliberately not among them".

So for `omit` the documentation states the exclusion is permanent *and* names two alternatives, while the runtime says it is coming and names none. An author who trusts the error waits for a feature that is never arriving; one who finds the docs learns they wrote the wrong call. The error is the one they will hit first.

**Fix:** distinguish *not yet implemented* from *out of scope*, and make the guidance mandatory rather than optional for the second kind. Either a second error class, or a required discriminator on this one — the point is that "we will get to it" and "use `select` instead" cannot share a sentence template.

Minimum per argument:

| argument | status | say instead |
|---|---|---|
| `omit` | out of scope | use `select`, or a policy's `redact` |
[...]

PR #72, opened 2026-07-28
Per-parent `take` / `skip` inside a to-many, which needed the fold to recurse

### LINKED ISSUE #62: ORM: per-parent `take` / `skip` inside a to-many relation
Branch: `feat/orm`. This is the highest-priority gap in the ORM's query surface — it is the only one on the list with no workaround that is both correct and affordable.

## What is refused

`take` / `skip` inside a to-many `include`:

```ts
Store.findMany({
  include: {
    products: {
      orderBy: { position: "asc" },
      take: 10,                     // ← ten products per store
      include: { media: { take: 1, orderBy: { position: "asc" } } },
    },
  },
})
```

```
UnsupportedQueryError: gemi ORM does not support 'products.take' yet (Store.findMany).
A 'take' inside a to-many relation is per parent, which the batched relation
planner cannot express. It arrives with the lateral join strategy; until then,
load 'products' as its own query, or narrow it with 'where'.
```

Raised at `packages/gemi/orm/compile/plan-relations.ts:519`.

## Why it has no workaround

- Fetching every child and slicing in JS reads the whole table — the point of `take: 1` on media is to *not* read 40k rows to render 10.
- Pushing the limit into the batched child query gives the first ten children **in total**, not per parent. Plausible, and wrong.
- One query per parent is the N+1 the `include` exists to avoid.

Measured against a real Prisma application (~1,400 call sites, 79 models, Postgres, in production), this shape sits in the storefront's page-load query, twice, at two different depths. It is the single call site that blocks that application from compiling against the ORM at all.

## Where the fix goes

**The refusal lives in the planner, not in a strategy** — worth knowing before starting. `LATERAL` + `json_agg` (`packages/gemi/orm/compile/lateral.ts`, shipped in iteration 9) is the shape that *can* express this: its subquery already runs once per parent row, so a `LIMIT` inside it means exactly what Prisma means. It refuses anyway, because `planRelations` validates the node before a strategy is chosen.

So the work is:

1. Let the planner carry `take` / `skip` on a to-many node instead of rejecting it.
2. Emit `LIMIT` / `OFFSET` inside the lateral subquery, under the node's own `orderBy`.
3. Keep the batched strategy refusing it — with a message that says *this query needs the lateral strategy*, not *this arrives later*. Postgres-only support is a complete answer for the applications that need it.

## Acceptance

- The nested query above returns ten products per store and one media row per product, matching Prisma, under the differential harness on Postgres.
- Both depths work — a per-parent `take` under another per-parent `take`.
- `skip` alongside `take`, and a negative `take`, agree with Prisma.
- The batched strategy still refuses, and its message names the strategy that would work.
- SQLite (no lateral) refuses with the same message rather than computing a wrong page.

## Related

- Error-message quality is #61; the message this issue changes is the one #61 holds up as the standard, so keep its four parts (what, why, when, what to do instead).

nstfkc:
Closes #62.

`include: { products: { take: 10 } }` means ten products **per store**. The batched planner serves every parent from one query, so the only `limit` it could emit pages the parents' children as a single set — plausible, and wrong — and it refused. The refusal lived in `assertNodeArgs`, *before* a strategy was chosen, so the one strategy that can express it never got the chance.

## What moved

Whether a per-parent page is expressible is a property of the **strategy**, not of the argument tree. So the refusal moves to where that is known — `assertPaginable`, called by the batched plan and by lateral when it declines. `assertNodeArgs` now only validates the argument (an integer; a `skip` cannot be negative) and still refuses `take` / `skip` on a to-one, as Prisma does.

The new message keeps #61's four parts and names the strategy that would work rather than a future release, plus *why this node reached one that cannot page it*:

```
gemi ORM does not support 'tags.take' yet (Post.findMany). A 'take' inside a
to-many relation is per parent, and the batched relation planner serves every
parent from one query — so the only limit it could emit would page every
parent's children as one set. Only the lateral join strategy can express it, by
paginating inside a subquery that already runs once per parent row, and it needs
Postgres: this node declined to fold (implicit many-to-many). Load 'tags' as its
own query, or narrow it with 'where'.
```

The tail varies: `not postgres`, `implicit many-to-many`, `this query planned with the batched strategy`.

## The page goes before the aggregate

`json_agg` is an aggregate, so a `limit` beside it caps the aggregate's **single row** and silently returns every child. That is the trap, and it is the shape that survives a careless test. A paginated node therefore gets a subquery of its own:

```sql
select coalesce(json_agg("__page_accounts"."data"
                order by "__page_accounts"."__ord_0" asc), '[]'::json) as "data"
from (select json_build_object(…) as "data", "Account"."createdAt" as "__ord_0"
      from "Account"
      where "Account"."userId" = "User"."id"
      order by "__ord_0" asc limit $1 offset $2) as "__page_accounts"
```

The ordering terms are carried out as `__ord_N` columns so the aggregate can re-apply them — aggregating over an ordered subquery and hoping the order survives is not something Postgres promises. A negative `take` pages in flipped order and the aggregate puts the page back into the caller's, so there is no array reversal to do. Paginating without an `orderBy` orders by the child's primary key, exactly as the root does.

An unpaginated node emits byte-identical SQL to before.

## The fold had to recurse

Acceptance criterion 2 is a page under a page, and it could not be met without this. A folded node's children never enter the related model's `$exec`, so a nested `include` under one has nowhere to be loaded from — the strategy declined the node instead. It now folds the **whole subtree**, and declines a node whose subtree contains anything it cannot express, because half a fold is not a thing. A decline still costs one statement rather than the subtree: the node runs as its own query with the strategy applied again one level down.

Side effect worth calling out for review: **a depth-3 tree on Postgres is now one statement rather than three.** That is the headline win iteration 9 deferred, and it changes what every existing nested include does there. The full nested differential matrix runs against it.

Declines are now: not Postgres, implicit many-to-many, self-relation, a `_count` on a node, or any of those below it.

### Two things had to follow the recursion

- **`redact` walks the subtree.** It is the one policy member that cannot ride along in the argument tree — `scope` becomes a `where` inside the subquery, `redact` is a row transform with nothing to rewrite — so the parent runs it on the child's behalf. At depth 1 only, a folded **grandchild came back unredacted**. `RootContribution` now carries the shape of what it folded and `redactFolded` recurses. Verified by deleting the recursive call and watching the new test fail.
- **A self-relation declines.** The correlation reads `"T"."fk" = "T"."pk"`, which inside the subquery compares its own row against itself rather than the outer one. Latent since iteration 9 and reachable at any depth once the fold recurses; batching handles it correctly, so declining is the whole fix. Expressing it properly needs an alias in the `from` clause — deliberately not in this PR.

### And one plan-cache fix

A relation node's `skip` sits inside `include`, which is a *literal* subtree in the plan key, so its value would have been recorded verbatim and minted one cache entry per page. The root's `skip` never had the problem, which is exactly why it went unnoticed until a node could carry one. A `take`'s sign stays structural, since it changes the SQL.

## Verification

Against **Prisma on real Postgres**, not asserted SQL:

- Twelve per-parent shapes: `take`, `skip`, both together, `take: 0`, a take larger than any parent's child count, a skip past the end (an empty array, not `null`), both negative-take orderings, no-`orderBy` fallback, a `where` narrowing before the page, a `select` alongside it, and a page on `findFirst`.
- Two nested shapes — `take` under `take`, and `take: -1, skip: 1` over `take: 1, skip: 1`.
- The fixture discriminates: one user has two accounts and another has one, so a limit applied to the combined set returns the right answer for the first parent and the wrong one for the second.
- The batched refusal, and the SQLite refusal, pinned next to the cases that now work.
- The full existing nested relation matrix, unchanged, under the recursive fold.

Plus 761 unit tests in `packages/gemi` (up 18) and the SQLite differential suite.

Pre-existing failures, unchanged by this branch and confirmed on the base commit: seven router/i18n tests in `packages/gemi`, and `generated.test.ts`'s generator-linkage probe in the template.

### Closes a `_count` divergence reported on #45 as a side effect

The new `_count on a folded node` decline also fixes the nested-`_count` bug reported against #45: an `include` carrying a `_count` used to fold *without* it and return rows missing the key the caller asked for. Confirmed by the same probe that reproduced it — it returned `[{"id":1}]` on the base branch and returns `_count` correctly here.

### Declines a node ordered by a relation

Added in review. A node's `orderBy` naming a relation compiles to a correlated subquery that `foldNode` has no `OrderContext` to build, so such a node cannot fold. The depth-1 case was already broken on `feat/orm`; recursion widened its reach to any node with a nested `include`, and one decline covers both.

## Not in this PR

`plans/orm/` is a historical record of each iteration rather than a living document, and iteration 9's plan does not state the nested-include decline as a permanent decision, so it is left alone. `docs/orm.md` — the user-facing page — gains a **Paging a relation** section and an updated decline list.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -148,14 +148,18 @@ them, and they return identical rows:
 
 - **`batched`** — one query per include node, with the children stitched onto the parents in
   process. Works on every dialect.
-- **`lateral`** — the children folded into the root statement with a `LATERAL` join and
-  `json_agg`. **Postgres only, and the default there.** One round trip instead of one per node.
+- **`lateral`** — the whole include tree folded into the root statement with `LATERAL` joins and
+  `json_agg`. **Postgres only, and the default there.** One round trip for the tree instead of one
+  per node.
 
 The lateral strategy **declines per node** rather than emitting SQL it cannot get right, falling
-back to batching for that node alone. It declines a node that has its own `include`, a relation
-inside a `select`, and an implicit many-to-many. A mixed include tree therefore uses both, which
-is fine — the results are the same either way, and there are tests asserting exactly that against
-Prisma across thirty relation shapes.
+back to batching for that node alone. It declines an implicit many-to-many, a self-relation (a
+relation onto the same table), a node carrying a `_count`, and any node with one of those below it
+— half a fold is not a thing, so a descendant that cannot be expressed declines its whole node. A
+declined node still runs under this strategy one level down, so a decline costs one statement, not
+the subtree. A mixed include tree therefore uses both, which is fine — the results are the same
+either way, and there are tests asserting exactly that against Prisma across thirty relation
+shapes.
 
 Override per call when you need to:
 
@@ -165,6 +169,45 @@ await User.findMany({ include: { accounts: true } }, { strategy: "batched" })
 
 Unlike `track`, `strategy` **does** reach the compiler and **is** part of the plan cache key: two
 strategies emit different SQL for the same arguments.
+
+### Paging a relation
+
+```ts
+await Store.findMany({
+  include: {
+    products: {
+      orderBy: { position: "asc" },
+      take: 10,                                       // ten products *per store*
+      include: { media: { take: 1, orderBy: { position: "asc" } } },
+    },
+  },
+})
+```
+
+`take` and `skip` inside a to-many are **per parent**, exactly as they are in Prisma — ten products
+for each store, not ten in total. A negative `take` means "the last N" and comes back in the order
+your `orderBy` describes, the same as at the root. Paginating without an `orderBy` orders by the
+child's primary key, because "the first ten" is otherwise only meaningful if the scan happens to be
+stable.
+
+**This needs the `lateral` strategy, so it needs Postgres.** The batched strategy serves every
+parent from one query, so the only `limit` it could emit would page every parent's children as a
+single set — plausible, and wrong. It refuses instead:
+
+```
+UnsupportedQueryError: gemi ORM does not support 'products.take' yet (Store.findMany).
+A 'take' inside a to-many relation is per parent, and the batched relation planner serves
+every parent from one query — so the only limit it could emit would page every parent's
+children as one set. Only the lateral join strategy can express it, by paginating inside a
+subquery that already runs once per parent row, and it needs Postgres: this node declined to
+fold (not postgres). Load 'products' as its own query, or narrow it with 'where'.
+```
+
+The tail of that message is the part to read: it says *why* this node reached a strategy that
+cannot page it — the dialect, an explicit `strategy: "batched"`, or a decline. There is no
+workaround that is both correct and affordable, which is why the refusal is loud: fetching every
+child and slicing in JavaScript reads the whole table, and one query per parent is the N+1 the
+`include` exists to avoid.
 
 ### Filtering on a relation
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -153,10 +153,17 @@ them, and they return identical rows:
   per node.
 
 The lateral strategy **declines per node** rather than emitting SQL it cannot get right, falling
-back to batching for that node alone. It declines an implicit many-to-many, a self-relation (a
-relation onto the same table), a node carrying a `_count`, and any node with one of those below it
-— half a fold is not a thing, so a descendant that cannot be expressed declines its whole node. A
-declined node still runs under this strategy one level down, so a decline costs one statement, not
+back to batching for that node alone. It declines:
+
+- an implicit many-to-many,
+- a self-relation (a relation onto the same table),
+- a node carrying a `_count`,
+- a node ordered by a *relation* — `orderBy: { user: { email: "asc" } }` or
+  `orderBy: { accounts: { _count: "desc" } }` — which compiles to a correlated subquery,
+- and any node with one of those below it: half a fold is not a thing, so a descendant that cannot
+  be expressed declines its whole node.
+
+A declined node still runs under this strategy one level down, so a decline costs one statement, not
 the subtree. A mixed include tree therefore uses both, which is fine — the results are the same
 either way, and there are tests asserting exactly that against Prisma across thirty relation
 shapes.
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -10,6 +10,7 @@ import {
   withTransaction,
 } from "./context";
 import {
+  type FoldedRelation,
   type NestedWriteStep,
   type RelationExecutor,
   type RelationStrategy,
@@ -817,7 +818,16 @@ export abstract class Model {
       if (plan.relations !== undefined && !system) {
         for (const relation of plan.relations) {
           if (relation.root === undefined) continue;
-          redactFolded(relation, result, op, system);
+          redactFolded(
+            {
+              as: relation.as,
+              model: relation.model,
+              folded: relation.root.folded,
+            },
+            result,
+            op,
+            system,
+          );
         }
       }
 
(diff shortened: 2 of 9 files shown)


nstfkc:
## Review — per-parent `take` / `skip` (closes #62)

Read the full diff, ran the unit suite (761 pass) and the template suite against real Postgres 16 (629 pass, 226 skipped; the two "failures" are the by-design "only one dialect's suite can run at a time" guard, not regressions).

**The headline works.** Verified against Postgres with a discriminating fixture — one user with three accounts, another with one — so a limit applied to the combined set would give the right answer for the first parent and the wrong one for the second:

```ts
User.findMany({
  orderBy: { id: "asc" },
  include: { accounts: { orderBy: { id: "asc" }, take: 2 } },
})
// accounts per parent: [2, 1]     ✓ per-parent, not [2, 0]
```

**It also fixes the nested-`_count` divergence I reported on #45**, via the new `_count on a folded node` decline. Same probe that returned `[{"id":1}]` on the base branch now returns `_count` correctly. Worth noting in the PR body — it closes that finding as a side effect.

The three moves the description leads with are all the right ones, and each is the kind that is only obvious in retrospect: moving the refusal from `assertNodeArgs` to `assertPaginable` because paginability is a property of the *strategy*; putting the page in a subquery beneath the aggregate rather than beside it; and carrying the ordering out as `__ord_N` columns instead of trusting an ordered subquery to stay ordered through `json_agg`. `redactFolded` recursing with `RootContribution.folded` is the one I'd have most expected a PR this size to miss.

---

### Blocking: the recursive fold regresses relation orderings inside an include

`decline()` gained a case for `_count` but not for an `orderBy` that names a relation. `foldNode` calls `parseOrderBy(child, node?.orderBy, operation)` (`lateral.ts:332`) with **no** `OrderContext`, and `parseOrderBy` throws when a key is in `schema.relations` and no context was supplied (`orderBy.ts:91`). So a folded node cannot express a relation ordering — but nothing stops it from folding one.

That gap is pre-existing at depth 1. What this PR changes is *reach*: a node with a nested `include` used to decline ("nested include") and go batched, where the ordering worked. Now it folds, and throws. Same probe, same fixture, same Postgres, base branch vs this one:

| shape (`strategy: "lateral"`) | `feat/orm` | this branch |
| --- | --- | --- |
| `accounts: { orderBy: { user: { email: "asc" } } }` | THREW | THREW |
| `accounts: { orderBy: { user: { email: "asc" } }, include: { organization: true } }` | **OK (1)** | **THREW** |

```
gemi ORM does not support 'orderBy.user' yet (Account.findMany).
Ordering by a relation is not implemented here.
```

Both shapes work under `strategy: "batched"` on both branches, so this is the dialect split again: works on SQLite in development, throws on Postgres in production, where lateral is the default. That is the same failure class as the `_count` one this PR fixes — and the PR's own reasoning for that fix applies here verbatim: a node that folds *without* something the caller asked for is a node that should have declined.

The fix looks like the same one line, next to the `_count` check in `decline()`:

```ts
// A relation ordering compiles to a correlated subquery that `foldNode` has no
// OrderContext to build, so a node carrying one cannot fold.
const ordering = args.orderBy;
if (namesARelation(child, ordering)) return "relation ordering on a folded node";
```

Supporting it properly instead — threading an `OrderContext` through `foldNode` so `order-relation.ts` can build the correlated subquery with the right qualifier — is a bigger change and probably its own PR. Either way `docs/orm.md`'s decline list ("an implicit many-to-many, a self-relation, a node carrying a `_count`") needs the fourth entry, and the pre-existing depth-1 case is worth a test whichever route it takes.

### Smaller

- **Unmeasured, and I am raising it as a question rather than a finding.** In the paginated shape the nested laterals sit inside the paged subquery (`scan` carries `joins`, and `joins` goes into the page's `from`), so a grandchild's subquery is evaluated for child rows the `LIMIT` then discards — and with an explicit `order by` the sort has to consume every input row before the limit applies. For the motivating shape (`take: 10` products, `take: 1` media each) over a large child set that could be a lot of `json_agg` thrown away. Whether Postgres defers the projection past the sort is plan-dependent and I have not benchmarked it. `plans/orm/benchmarks.md` already has the harness; a row for "paged node with a subtree" would answer it.
- The plan-key fix for a node's `skip` (`plan.ts:254`) is right and the reasoning is worth keeping — worth noting it also silently fixes anything else that grows a `skip` inside a literal subtree later.
- `suffixed()` keeping the bare `__lat_<relation>` form at depth 1 and only indexing deeper is a nice call for query-log readability, and the `accounts.user.accounts` shadowing case it cites is real.

Everything else in the diff reads as correct to me. Once the decline is in, this is a clear improvement on a gap that #62 documents as having no affordable workaround.

nstfkc:
Fixed and pushed as 6e3dcc9. Thanks — the finding was right, and reproducing it first turned up one thing worth adding.

**The decline is in.** `decline()` gains `relation ordering on a folded node`, checking both of Prisma's relation-ordering forms (`{ user: { email: "asc" } }` and `{ accounts: { _count: "desc" } }`) in both containers the grammar accepts (one object, or an array of them).

**It also fixes the pre-existing depth-1 case.** I expected to leave that one and add a test for it, as you suggested — but the decline is per node and does not care why the node was going to fold, so the flat shape now lands on batched and works too. Reproduced before and after, on the base branch and this one:

| shape (`strategy: "lateral"`) | `feat/orm` | before fix | after fix |
| --- | --- | --- | --- |
| `accounts: { orderBy: { user: { email: "asc" } } }` | THREW | THREW | **OK (batched)** |
| `accounts: { orderBy: { user: { email: "asc" } }, include: { organization: true } }` | OK | THREW | **OK (batched)** |
| `accounts: { orderBy: { user: { _count: "desc" } } }` | THREW | THREW | **OK (batched)** |
| `accounts: { orderBy: { organizationRole: "desc" } }` | OK (lateral) | OK (lateral) | **OK (lateral)** |

That last row is a test too — without it the decline could be matching every `orderBy` and nothing would notice.

Also pinned: a per-parent `take` beneath a relation ordering now refuses rather than being silently batched, and the message names which of the two boundaries it hit (`declined to fold (relation ordering on a folded node)`).

**Three shapes added to the differential matrix**, so they run against Prisma on both dialects and under both strategies rather than only as a compile assertion.

One thing worth recording, because my first draft of those cases failed and the failure was mine: each now carries an `id` tiebreaker. gemi orders by a correlated subquery and Prisma by a join, so the two are free to break a tie differently — and the fixture ties on purpose, since both of Ada's accounts point at the same user and therefore the same email. Without the tiebreaker the harness reported a divergence that was not one, which is the failure mode a differential harness must not have. The array-form case passed throughout, because `organizationRole` differs and there was no tie to break.

`docs/orm.md`'s decline list is now a real list with the fourth entry in it.

---

**On the `_count` side effect** — noted, and thank you for checking it against #45. I have added it to the PR description.

**On the unmeasured question** (nested laterals evaluated for rows the page then discards): agreed that it is real and that I have not measured it either. I have left it alone rather than guess — it is a benchmark row and an optimisation, not a correctness fix, and folding it into this PR would mean shipping a plan change with no number behind it. Happy to take it as a follow-up against `plans/orm/benchmarks.md` if you want it before this lands rather than after.

Full suite green on both dialects after the change: 767 unit tests, and the template suite against Postgres 16 (634 pass; the two differential "failures" are the by-design one-dialect-at-a-time guard, and `generated.test.ts`'s generator-linkage probe is the pre-existing environment one).

nstfkc:
### Re-reviewed `6e3dcc9` — the decline fixes it, including the pre-existing half

Verified against Postgres 16, same probe as before, both strategies:

| shape (`strategy: "lateral"`) | `402f830` | `6e3dcc9` |
| --- | --- | --- |
| `accounts: { orderBy: { user: { email: "asc" } } }` | THREW | **OK** |
| …same **+ nested include** | THREW | **OK** |
| `users: { orderBy: { accounts: { _count: "desc" } } }` | — | **OK** |
| root `orderBy: { organization: { name: "asc" } }` + `include` | OK | OK |
| `accounts: { orderBy: { id: "asc" }, take: 1 }` | OK | OK |

Every one matches `strategy: "batched"`, and the headline `take` still refuses correctly under batched. So this closes the regression *and* the depth-1 case that predated the PR — which was more than I asked for and the right call, since the two were the same missing decline.

Two things I checked rather than assumed:

- `ordersByRelation` walks both containers Prisma accepts (one object, or an array of them) and keys on `key in schema.relations`, which is the same test `parseOrderBy` makes — so the decline and the throw it is preventing cannot disagree about what counts as a relation ordering. It is asked against `child`, matching `foldNode`'s own `parseOrderBy(child, node?.orderBy, …)`.
- It sits above the `relationNodes` recursion, so a relation ordering at depth 2+ declines the whole subtree rather than only its own node. Confirmed by the nested-include row above.

Full template suite on Postgres: 635 passed, 229 skipped (up 6, the new differential cases), with only the by-design "one dialect's suite at a time" guard failing. Unit suite 741 pass.

The doc list now has the fourth entry with both spellings. Nothing further from me on this one — the `take`/`skip` work itself I verified as correct in the previous round, and my remaining note there was the unmeasured question about nested laterals inside the paged subquery, which is a benchmark rather than a blocker.

Issue #68, opened 2026-07-28
ORM: decide `omit`, `distinct` and `cursor` — implement, or make the refusal permanent


nstfkc:
Branch: `feat/orm`. Three read arguments in Prisma's `findMany` type are accepted by the signature and refused by the compiler. `packages/gemi/orm/compile/read.ts:27-45` allows `where`, `orderBy`, `skip`, `take`, `select`, `include` and nothing else, so `cursor`, `distinct` and `omit` all fail as:

```
gemi ORM does not support 'cursor' yet (User.findMany).
```

Grouped in one issue because the decision on each is the same *kind* of decision — implement, or make the refusal permanent and say what to use instead — and because two of the three land on the same dependency.

Counts are from a large production Prisma application read end to end.

---

## `omit` — 7 call sites

The motivating case is "every column except the password hash". `docs/orm.md` already lists it under **Not in scope** with two alternatives (`select`, or a policy's `redact`), so the documentation and the runtime currently disagree — the error says "yet".

The argument for implementing it anyway: with `select`, the exclusion list is rewritten every time the model gains a column, and the day someone forgets is the day the new column silently stops being returned. That is the safer of the two failure directions, but it is a maintenance tax per column, and `redact` is a policy — a per-model decision, not a per-query one.

Compilation is trivial: `omit` is the default column list minus the named fields, resolved in `packages/gemi/orm/compile/select.ts` where the default list is already built. The type side is the real work — `omit` narrows the payload the same way `select` does, and the generated bases have to thread `Prisma.UserGetPayload` for it.

**Decision needed:** implement, or make the refusal permanent and align the error with the docs.

## `distinct` — 2 call sites

Worth noting that the application is already wary of Prisma's version: it documents that Prisma applies `distinct` **in memory**, so `take` / `skip` on a distinct `findMany` neither reduces the rows pulled nor paginates by group, and it rewrites the query as `DISTINCT ON` in raw SQL instead.

So refusing may be the *better* answer than Prisma's — but only if the rewrite has somewhere to go, which is composable SQL fragments (#63). Implementing it as real `DISTINCT ON` (Postgres) would diverge from Prisma's semantics and the differential harness would correctly call that a divergence; that divergence may still be the right call, but it has to be a deliberate one.

**Decision needed:** implement as `DISTINCT ON` and accept a documented divergence from Prisma, or refuse permanently and point at fragments.

## `cursor` — 1 call site

[...]

Issue #64, opened 2026-07-28
ORM: `aggregate` and `groupBy`


nstfkc:
Branch: `feat/orm`. `packages/gemi/orm/plan.ts:14-27` lists thirteen operations; `aggregate` and `groupBy` are deliberately not among them, and the generated bases never emit them. `count` is the only aggregate the ORM has.

## Why revisit the exclusion

Read against a large production Prisma application, this is five call sites — small — but two of them are not analytics:

- `_max: { sortIndex }` to compute the next sort index when reordering a list
- `_max: { position }` to compute the next position for a newly uploaded media row

That is ordinary CRUD, and it means the floor is higher than a call-site count implies: any list-reordering feature reaches for `_max` on the way in. The other three are `by: ["operation"]` + `_sum` for a usage tile, a `by: "conversation_id"` grouping (note the **string** form, which Prisma also accepts), and a price-offer grouping.

Unlike every other gap on the list, there is no partial workaround inside the ORM. Each of these has to drop to raw SQL, which is currently `DB.sql` and therefore also blocked on composable fragments (#63).

## Scope

Two separable pieces, and the smaller one carries most of the value:

**`aggregate`** — `_count`, `_sum`, `_avg`, `_min`, `_max` over a filtered set, returning one object. This covers the two next-position sites and is a small compiler addition: the `where` / `orderBy` / `skip` / `take` machinery already exists, and the result is a single row with a fixed shape.

**`groupBy`** — `by`, `_count`/`_sum`/`_avg`/`_min`/`_max`, and `having`. Larger: `having` is a second predicate compiler over aggregate expressions rather than columns, and Prisma requires every `by` field to appear in `orderBy` when one is given.

Suggest shipping `aggregate` first and deciding on `groupBy` after, rather than treating them as one unit.

## Notes

- Both need to be emitted onto the generated bases (`packages/gemi/bin/orm/emit.ts`) with Prisma's own arg and payload types, so narrowing stays exact — the same concrete-base approach as the other thirteen.
- Prisma's `count({ select: { … } })` per-field-count form is refused today at `packages/gemi/orm/compile/read.ts:40-44`, deliberately, as aggregate territory. It belongs with `aggregate`, and the two decisions should be made together.
- Numeric aggregate results come back as strings from Postgres for `bigint`/`numeric` sums. Coercion has to be aware of that, the same way the lateral strategy's JSON path already is.

## Acceptance

[...]

PR #73, opened 2026-07-28
Composable SQL fragments — `sql` / `join` / `empty`, `unsafeSql`, and rowcount from raw writes

### LINKED ISSUE #63: ORM: composable SQL fragments — `sql` / `join` / `empty`, an inline escape hatch, and rowcount from writes
Branch: `feat/orm`. The ORM deliberately does not implement every SQL shape — that is the right call. What it does not have is the place those shapes go instead.

`DB.sql` covers a single self-contained statement. It does not cover **composition**: fragments built as values, passed around, and interpolated into each other.

## The evidence

Read off a large production Prisma application (~1,400 call sites, 79 models, Postgres): 105 raw call sites — `$queryRaw` ×71, `$executeRaw` ×34 — of which `Prisma.sql` ×29, `Prisma.join` ×6, `Prisma.empty` ×7, `Prisma.raw` ×2 are composition rather than one-shot statements.

The hardest of them assembles `Prisma.Sql` values conditionally, reuses one *inside* `array_agg(… ORDER BY …)`, and injects them into `HAVING` and `ORDER BY`. Its own comment records that the JavaScript equivalent loaded ~11 MB per poll.

Every unsupported ORM shape below funnels here too — `distinct` (#68), cursor keyset pagination (#68), `groupBy` (#64) all have "drop to raw SQL" as their stated workaround, and that workaround needs somewhere to land.

## Three capabilities, hardest-to-live-without first

**1. Fragments as first-class values** — `sql`, `join`, `empty`. A conditional `WHERE` built from a list is the most common raw shape by a distance.

```ts
const filters = []
if (q) filters.push(sql`name ilike ${`%${q}%`}`)
if (orgId) filters.push(sql`"organizationId" = ${orgId}`)
const where = filters.length ? sql`where ${join(filters, " and ")}` : empty
await DB.sql`select * from "Product" ${where}`
```

The internal `Fragment` / `Binder` type in `packages/gemi/orm/compile/fragment.ts` is already exactly this shape — text plus positional binders, composable, never string-concatenating a value. The question is what of it becomes public API, not whether the machinery exists.

**2. A bind-vs-inline escape hatch.** Plan quality can depend on *not* parameterising. One site inlines a constant with `Prisma.raw` and explains why in nine lines: the partial index is `WHERE name IS DISTINCT FROM '<constant>'`, a bound parameter is opaque at plan time, so the planner cannot prove the query matches the index and falls back to a sequential scan over ~1M rows. A builder with no escape hatch cannot express this at any price.

This one needs a deliberate design call, because it is the only door in the ORM through which a value reaches the SQL text — the exact thing invariant 2 exists to prevent. It should look dangerous at the call site and be impossible to reach by accident.

**3. Rowcount back from a write.** A credit-ledger service uses `$executeRaw`'s return value as a compare-and-swap: `UPDATE … WHERE status = 'reserved'` returning 1 means this caller won the race, 0 means it lost. The rowcount *is* the concurrency primitive, so a raw-write API that discards it is not usable for that class of code.

## Constraints

- Whatever ships must run on the ambient transaction (`packages/gemi/orm/context.ts`) like every other query — a raw statement that escapes `Model.transaction` is worse than no raw statement.
- Fragments carry binders, not interpolated values, everywhere except the explicit escape hatch in (2).
- Both dialects: SQLite's `?` and Postgres's `$n` differ, so a fragment's placeholders have to be assigned at assembly time, not at authoring time.

## Acceptance

- `sql` / `join` / `empty` exported, composable to arbitrary depth, with values always bound.
- Both dialects renumber placeholders correctly when fragments are nested or reused.
- Raw statements participate in ambient transactions.
- A write returns its affected-row count.
- The escape hatch exists, is separately named, and is documented with the index-plan case as the reason it is there.

nstfkc:
Closes #63.

The ORM declines plenty of SQL shapes on purpose, and every one of them has *"drop to raw SQL"* as its stated workaround — `distinct` and cursor pagination (#68), `groupBy` (#64), window functions, recursive CTEs. That workaround had nowhere to land. `DB.sql` runs one self-contained statement; it cannot express a predicate built conditionally, held as a value, and interpolated into a larger query — which is what raw SQL actually looks like in an application.

```ts
import { DB } from "gemi/facades"
import { sql, join, empty } from "gemi/orm"

const filters = []
if (q) filters.push(sql`"name" ilike ${`%${q}%`}`)
if (organizationId) filters.push(sql`"organizationId" = ${organizationId}`)

const where = filters.length ? sql`where ${join(filters, " and ")}` : empty
const rows = await DB.query(sql`select * from "Product" ${where} limit ${20}`)
```

## It is the compiler's own `Fragment`, made public

Not a parallel mechanism. `compile/fragment.ts` has held exactly this shape since iteration 1 — text plus positional binders, composable, never concatenating a value into the text — because that is what invariant 2 needs internally. Reusing it means a public fragment binds by the same rule the compiler's own SQL does, and the parameter ceiling, the placeholder numbering and the dialect split are handled rather than reimplemented one door down.

Placeholders are assigned **at assembly, not at authoring**, which is what lets a fragment nest, be reused twice in one statement, or be built before the dialect is known. There is a test that renders one fragment for SQLite, then Postgres, then SQLite again — a fragment that memoised its rendered text would fail it in the direction that is *not* a syntax error.

### Rowcount from a write

`compile/write.ts` already recorded that driver metadata was not portable, and that the ORM's own `{ count }` therefore comes from `RETURNING <pk>`. I measured it again, on **both dialects**, across `update`, `delete`, `insert`, a statement matching nothing, and one carrying `returning`:

| | `count` | `affectedRows` |
| --- | --- | --- |
| SQLite | correct, every case | `null` |
| Postgres | correct, every case | `null` |

So `DB.execute` can report it. That matters because the count *is* the concurrency primitive, not a diagnostic:

```ts
const won = await DB.execute(
  sql`update "Reservation" set "status" = 'claimed'
      where "id" = ${id} and "status" = 'reserved'`,
)
if (won === 0) throw new AlreadyClaimed()
```

The ORM's own choice is unchanged — `RETURNING` is still exact and still asks the driver for nothing — but its note claimed `count` was "not verifiable against Postgres from here", which is no longer true, so that half is corrected and pointed at the new measurement.

### `unsafeSql`

The one door through which text reaches the SQL. It exists for something a builder cannot otherwise express: a partial index declared `where "name" is distinct from 'reaper'` is only usable if the planner can prove the query's predicate matches it, and a bound parameter is opaque at plan time — so `$1` there means a sequential scan of the whole table instead of an index hit. The constant has to be *in the text*.

It refuses a non-string, and it refuses **tagged-template use** — as a tag it would interpolate values straight into the SQL, which is the one thing it must never offer, and both spellings read identically at the call site. `sql("…")` called as a plain function is also refused, with a message pointing at `unsafeSql` since that is what the caller was reaching for.

## Two things found while building it

- **A `Date` was not portable through a fragment.** Bun's SQLite driver rejects one outright (`Binding expected string, TypedArray, boolean, number, bigint or null`); Postgres takes it. `SqlDialect` gains `encodeUntyped`, for a value with no declared column type: a `Date` becomes milliseconds and a boolean `0`/`1` on SQLite, which is *exactly* what `encode` writes for those columns — so a raw statement compares against ORM-written rows rather than against nothing. A plain object is deliberately **not** JSON-encoded: the compiler only knows to do that because a field says `Json`, and guessing from the value would turn a mistyped parameter into a successfully-written string.
- **`DB.query` threw synchronously** for a rejected fragment while returning a promise otherwise, so `DB.query(...).catch(…)` would miss precisely the errors it refuses to run. Both entry points are `async`.

## The ambient transaction

`DB.query` / `DB.execute` resolve the handle through the same `currentTransaction()` that `Model.$exec` uses. A raw statement that quietly ran on a pooled connection would commit while its neighbours rolled back — the worst outcome available here, and invisible in any test that does not deliberately roll back. Covered on both dialects: a raw write rolls back with the transaction, a raw read sees that transaction's *uncommitted* writes (which is what proves it is the same connection rather than a second one that also committed), and a savepoint rolls back to itself leaving the outer transaction usable.

`DB.sql` is untouched and still Bun's own tagged template, with a regression test saying so. It still does not join the ambient transaction; the docs now say which to reach for.

## Also documented

**Policies do not apply to raw SQL, and cannot** — a policy scopes a model's queries by rewriting the argument tree, and a raw statement has neither. Worth stating explicitly next to a page that otherwise says "every path that reaches another model carries that model's policies", because that sentence stops at this door.

### Only fragments this module made are spliced

Added in review, and it was a live SQL injection. `isFragment` asked whether a value had a string `text` and an array `binders` — so `{"text": "…", "binders": []}` arriving in a parsed request body was spliced into the statement as text, through the intended call site with the intended safe value. One of the reproduced payloads returned the password column out of a statement selecting only `email`.

Membership in a `WeakSet` this module owns, granted only by `sql`, `join`, `unsafeSql` and `empty`, registering what each constructor *returns* (since `concat`/`joinFragments` return fresh objects). Six tests cover it, all of which fail against the structural check — including a structural *clone* of a real fragment, which shape cannot tell apart.

The internal `Fragment` never had this problem, because nothing outside `compile/` could construct one. Making the type public is what turned "text plus binders" into a shape an attacker can name.

## Verification

- 46 unit tests in `packages/gemi/orm/sql.test.ts`, with no database anywhere near them. Every case asserts the emitted text **and** the bound values — a test that checked only the text would pass for an implementation that interpolated, which is the one failure this exists to prevent.
- 14 integration tests in `templates/saas-starter/app/models/raw-sql.test.ts`, run against **both dialects** (28 total with `TEST_POSTGRES_URL` set, all green) — rowcounts, the compare-and-swap, transaction and savepoint rollback, a value that looks like SQL staying data, and the composition example end to end.
- Full template suite green on SQLite and on Postgres.

Pre-existing failures, unchanged and confirmed on the base commit: seven router/i18n tests in `packages/gemi`, and `generated.test.ts`'s generator-linkage probe in the template.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -593,6 +593,101 @@ Measured numbers, the methodology, and the reasoning behind the lateral default 
 `plans/orm/benchmarks.md` in the repository — deliberately not here, because they are tied to a
 machine and a dataset and would go stale as documentation.
 
+## Raw SQL
+
+The ORM does not implement every SQL shape, and that is deliberate. What it owes you is somewhere
+for the rest to go — not just a way to run one hand-written statement, but a way to *compose* SQL,
+because a predicate built conditionally and passed around is the shape raw SQL actually takes in an
+application.
+
+```ts
+import { DB } from "gemi/facades"
+import { sql, join, empty } from "gemi/orm"
+
+const filters = []
+if (q) filters.push(sql`"name" ilike ${`%${q}%`}`)
+if (organizationId) filters.push(sql`"organizationId" = ${organizationId}`)
+
+const where = filters.length ? sql`where ${join(filters, " and ")}` : empty
+
+const rows = await DB.query<Product>(
+  sql`select * from "Product" ${where} order by "position" limit ${20}`,
+)
+```
+
+| | |
+| --- | --- |
+| `` sql`…` `` | A fragment. Every `${value}` is **bound**; a `${fragment}` is spliced in, carrying its own parameters. Nests to any depth. |
+| `join(items, sep)` | Fragments *or* values, joined. `join(ids)` is `$1, $2, $3`. An empty list is `empty`, not a dangling separator. |
+| `empty` | The fragment that contributes nothing, so a conditional predicate stays a value instead of becoming a branch. |
+| `DB.query(fragment)` | The rows. |
+| `DB.execute(fragment)` | How many rows the statement **touched**. |
+
+Three properties worth knowing, because they are the reason this is not just `DB.sql`:
+
+- **Placeholders are assigned at assembly, not at authoring.** SQLite writes `?` and Postgres
+  writes `$1`; a fragment does not know where it will land, so the numbering happens once the whole
+  statement exists. The same fragment can be nested, reused twice in one statement, or built before
+  the dialect is known.
+- **It runs on the ambient transaction.** A `DB.query` inside `Model.transaction` joins it, exactly
+  as a `User.create` does. A raw statement that quietly committed while its neighbours rolled back
+  would be worse than no raw statement at all.
+- **A plain string is refused.** Accepting one would make an interpolated template literal the path
+  of least resistance, which is the injection everything else here exists to prevent.
+
+**Policies do not apply here, and cannot.** A policy scopes a *model's* queries by rewriting the
+argument tree; a raw statement has no argument tree and no model behind it, so nothing scopes it and
+nothing redacts its rows. That is not a gap to be closed later — it is what "raw" means. If the
+statement reads a policied model, the tenant predicate is yours to write, and
+[`Every path that reaches another model carries that model's policies`](#every-path-that-reaches-another-model-carries-that-models-policies)
+stops at this door.
+
+Values are handed to the driver as they arrive, with two exceptions that exist so a fragment means
+the same thing on both dialects: a `Date` binds as milliseconds on SQLite — which is what the ORM
+stores — and a boolean as `0`/`1`. A plain object is **not** JSON-encoded for you; the compiler only
+knows to do that because a field says `Json`, and guessing from the value would turn a mistyped
+parameter into a successfully-written string.
+
+### The rowcount is a primitive, not a diagnostic
+
+```ts
+const won = await DB.execute(
+  sql`update "Reservation" set "status" = 'claimed'
+      where "id" = ${id} and "status" = 'reserved'`,
+)
+if (won === 0) throw new AlreadyClaimed()
+```
+
+`1` means this caller won the compare-and-swap and `0` means it lost. That is the whole concurrency
+control for a large class of code, so an API that discarded the number could not express it.
+
+### `unsafeSql` — the one door into the SQL text
+
+```ts
+import { unsafeSql } from "gemi/orm"
+
+// A literal in this file. Never a request, a form, a header or a database row.
+const REAPER = unsafeSql(`'reaper'`)
+
+await DB.query(sql`
+  select … from "Job" where "name" is distinct from ${REAPER} …
+`)
+```
+
+**Never give it a value that came from outside the program.** There is no escaping and there will
+not be — escaping that is "usually right" is worse than none, because it survives review.
+
+It exists because plan quality can depend on *not* parameterising. A partial index declared
+`where "name" is distinct from 'reaper'` is only usable if the planner can prove the query's
+predicate matches the index's, and a bound parameter is opaque at plan time — so `$1` there means a
+sequential scan of the whole table instead of an index hit. The constant has to be in the text. The
+same applies to identifiers and sort directions, which cannot be parameters in SQL at all; the rule
+is the same for those, and it is about where the value came from, not what it is.
+
+`DB.sql` is unchanged and still Bun's own tagged template, for a single self-contained statement
+that needs none of this. It does **not** join the ambient transaction — use `DB.query` when that
+matters.
+
 ## Dialects
 
 **SQLite and Postgres** are built and tested, on every operation, against a differential harness
@@ -612,6 +707,8 @@ Stated so you can plan around them rather than discover them:
   touched.
 - **No `omit`.** Use `select`, or a policy's `redact`.
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
+- **No `groupBy`, `aggregate`, `distinct` or cursor pagination.** These land in
+  [Raw SQL](#raw-sql), which exists so that "not implemented" has an answer rather than a shrug.
 
 ## See also
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -644,9 +644,18 @@ stops at this door.
 
 Values are handed to the driver as they arrive, with two exceptions that exist so a fragment means
 the same thing on both dialects: a `Date` binds as milliseconds on SQLite — which is what the ORM
-stores — and a boolean as `0`/`1`. A plain object is **not** JSON-encoded for you; the compiler only
-knows to do that because a field says `Json`, and guessing from the value would turn a mistyped
-parameter into a successfully-written string.
+stores — and a boolean as `0`/`1`.
+
+A plain object is **not** JSON-encoded for you. The compiler only knows to do that because a field
+says `Json`, and guessing from the value would turn a mistyped parameter into a successfully-written
+string. What the driver then does with it differs: Bun binds an object into a Postgres `jsonb`
+column correctly, and on SQLite binds it to something no row matches. Stringify it yourself if the
+column is JSON and you want the same answer on both.
+
+**Only a fragment built by `sql`, `join`, `unsafeSql` or `empty` is spliced.** Anything else in an
+interpolation slot is bound — including an object that merely *looks* like a fragment, which is what
+a parsed request body can be made to contain. That check is membership, not shape, precisely so that
+`{"text": "…", "binders": []}` arriving from outside the program is a parameter and not a statement.
 
 ### The rowcount is a primitive, not a diagnostic
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -618,7 +618,7 @@ const rows = await DB.query<Product>(
 | | |
 | --- | --- |
 | `` sql`…` `` | A fragment. Every `${value}` is **bound**; a `${fragment}` is spliced in, carrying its own parameters. Nests to any depth. |
-| `join(items, sep)` | Fragments *or* values, joined. `join(ids)` is `$1, $2, $3`. An empty list is `empty`, not a dangling separator. |
+| `join(items, sep)` | Fragments *or* values, joined. `join(ids)` is `$1, $2, $3`. An empty list is `empty`, not a dangling separator. The separator is **glue only** — see below. |
 | `empty` | The fragment that contributes nothing, so a conditional predicate stays a value instead of becoming a branch. |
 | `DB.query(fragment)` | The rows. |
 | `DB.execute(fragment)` | How many rows the statement **touched**. |
@@ -651,6 +651,20 @@ says `Json`, and guessing from the value would turn a mistyped parameter into a 
 string. What the driver then does with it differs: Bun binds an object into a Postgres `jsonb`
 column correctly, and on SQLite binds it to something no row matches. Stringify it yourself if the
 column is JSON and you want the same answer on both.
+
+**The separator `join` takes is written into the SQL**, so it is not a free string: whitespace,
+commas, or `and` / `or`, and nothing else. Anything more expressive is legitimate but has to say so
+with `unsafeSql`, which keeps that the only door:
+
+```ts
+join(filters, " and ")               // fine
+join(parts, unsafeSql(") and ("))    // fine, and visibly a decision
+join(ids, req.query.sep)             // refused
+```
+
+That is an allowlist rather than a blocklist on purpose. The obvious blocklist — reject quotes,
+semicolons and comment markers — lets `") or 1=1 union select password from Users where 1=(1"`
+through, which has none of them and is still an injection.
 
 **Only a fragment built by `sql`, `join`, `unsafeSql` or `empty` is spliced.** Anything else in an
 interpolation slot is bound — including an object that merely *looks* like a fragment, which is what
--- a/packages/gemi/facades/DB.ts
+++ b/packages/gemi/facades/DB.ts
@@ -1,7 +1,9 @@
 import type { SQL } from "bun";
 import { DatabaseManager } from "../database/DatabaseManager";
 import type { Dialect } from "../database/dialect";
-import { withTransaction } from "../orm/context";
+import { currentTransaction, withTransaction } from "../orm/context";
+import { dialectFor } from "../orm/dialect";
+import { renderFragment, type SqlFragment } from "../orm/sql";
 import { Facade } from "./Facade";
 
 // Access to the app's database connection. Bun's `SQL` client is a tagged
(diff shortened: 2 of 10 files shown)


nstfkc:
## Review — composable SQL fragments (closes #63)

Read the full diff, ran the unit suite (787 pass) and the new `raw-sql.test.ts` against SQLite (14 pass, 1 skipped).

The design is the right one, and the argument for it in `sql.ts` is the strongest part of the PR: reusing the compiler's own `Fragment` rather than building a parallel mechanism means a public fragment binds by the same rule the compiler's SQL does, and the parameter ceiling, placeholder numbering and dialect split come along for free. Assigning placeholders at assembly rather than authoring — with the render-twice-on-two-dialects test — is the detail that makes nesting and reuse actually work. Routing `DB.query` / `DB.execute` through `currentTransaction()` is right and the reasoning for it is the same one `Model.$exec` makes. Making both entry points `async` so a rejected fragment rejects rather than throwing synchronously is a good catch. `encodeUntyped` is well-scoped, and the measurement table for `count` vs `affectedRows` is worth more than the paragraph it replaces.

**One blocking issue.**

### `isFragment` is structural, so a JSON request body can splice text into the SQL

```ts
function isFragment(value: unknown): value is Fragment {
  return (
    typeof value === "object" &&
    value !== null &&
    typeof (value as Fragment).text === "string" &&
    Array.isArray((value as Fragment).binders)
  );
}
```

`interpolate` splices anything `isFragment` accepts and binds everything else. That check is duck-typing, and `{"text": "...", "binders": []}` is two lines of JSON — so any attacker-controlled object reaching an interpolation slot is spliced into the statement as **text**, which is exactly the thing the whole mechanism exists to make impossible.

The call site does not have to be unusual. This is the intended usage, with the intended safe value:

```ts
const body = await request.json()
await DB.query(sql`select "email" from "User" where "email" = ${body.email}`)
```

Against a real SQLite database through the public facade, with two seeded users:

```
body = {"email":"alice@x.test"}
  -> select "email" from "User" where "email" = ?      values: ["alice@x.test"]
  -> [{"email":"alice@x.test"}]                                            ✓

body = {"email":{"text":"'' or 1=1","binders":[]}}
  -> select "email" from "User" where "email" = '' or 1=1   values: []
  -> [{"email":"alice@x.test"},{"email":"bob@x.test"}]                     ✗

body = {"email":{"text":"'' union select \"password\" from \"User\"","binders":[]}}
  -> [{"email":"hunter2"},{"email":"s3cret"}]                              ✗
```

The last one returns the password column out of a statement that selects only `email`. `join()` is the same door and slightly wider, since `join(body.filters)` maps `interpolate` over an array the caller never inspected.

Worth being precise about the severity: the internal `Fragment` never had this problem, because nothing outside `compile/` could construct one. Making the type public is what turns "text plus binders" from an implementation detail into a shape an attacker can name. `unsafeSql` is carefully built so that reaching for the escape hatch is a decision — this is a second, unmarked door next to it.

**The fix is small and I verified it.** Membership in a `WeakSet` the module controls, instead of shape:

```ts
const FRAGMENTS = new WeakSet<object>()
const brand = (f: Fragment): SqlFragment => (FRAGMENTS.add(f), f)

// sql(), join(), unsafeSql() return brand(...); `empty` is branded once.

function isFragment(value: unknown): value is Fragment {
  return typeof value === "object" && value !== null && FRAGMENTS.has(value)
}
```

Applied to this branch: the forged bodies above bind as parameters instead of splicing (`= ?`, no rows returned, nothing leaked), `raw-sql.test.ts` still passes 14/1 skipped, and all 787 unit tests still pass — including the nesting, reuse and cross-dialect-render cases, since every publicly-built fragment is branded at its constructor. A `Symbol` property would also work but has to survive `concat`/`joinFragments`, which return fresh objects; registering the returned object is the version that cannot drift.

### Smaller

- With the brand in place, a forged object reaches the driver as a bound value. `dialect/sqlite.ts`'s new comment says a plain object is left alone "because the driver rejects it" — on this path it did not: Bun bound it without complaint and the query simply matched nothing. Silent rather than the loud failure the comment promises. If the intent is loud, `encodeUntyped` refusing a non-`Date` plain object by name would deliver it; if the intent is "the driver's problem", the comment overstates what the driver does.
- `DB.execute` reads `Number(result?.count ?? 0)`. The PR says `count` was measured on a statement carrying `returning`, where Bun's result is an array — worth a line in the doc comment saying that an array result still carries `count`, since `result.count` on an array reads like a bug to anyone who has not seen the measurement.

Everything else reads as correct. With the branding change this is a good addition, and it is the piece that makes #64 and #68's "drop to raw SQL" an honest answer rather than a deferral.

nstfkc:
Fixed and pushed as b10efea. You were right, and it was worse than a theoretical shape collision — I reproduced both of your payloads before touching anything, and the second one did return the password column out of a statement selecting only `email`.

**The brand is in**, as you specified: a `WeakSet` this module owns, membership granted only by `sql`, `join`, `unsafeSql` and `empty`, registering the object each constructor *returns* rather than stamping a property — for exactly the reason you gave, that `concat`/`joinFragments` return fresh objects.

One implementation note: `empty` is initialised at module scope, so the `WeakSet` and `brand` had to move above it or the first `brand()` call hits the temporal dead zone. They now sit at the top of the file, which also puts the argument for identity-over-shape where someone reading the module meets it first.

**Six tests, and all six fail against the structural check** — I verified that by reverting `isFragment` and re-running rather than assuming:

| test | structural | branded |
| --- | --- | --- |
| forged body → `'' or 1=1` | ✗ splices | ✓ binds |
| forged body → `union select "password"` | ✗ splices | ✓ binds |
| forged entry inside `join` | ✗ splices | ✓ binds |
| forged object handed to `renderFragment` | ✗ runs | ✓ refused |
| **structural clone of a real fragment** | ✗ splices | ✓ binds |
| forged body through `DB.query`, real database | ✗ leaks | ✓ 0 rows |

The clone is the one I added beyond your list: same `text`, same `binders`, copied from a real fragment rather than hand-written. Shape cannot tell it apart; membership can. Two of the six run through the public facade against a real database on both dialects, since what matters is not what the compiler emits but what comes back. Composition and two-level nesting are pinned next to them, or the fix could have traded an injection for a mechanism that silently stopped splicing.

---

**On the SQLite comment** — you were right that it overstates, and measuring it changed what I did about it. Bun binds a plain object *silently on both dialects*, and on Postgres it lands correctly in a `jsonb` column:

```
object -> jsonb column, bound as $1  ->  [{"payload":{"a":1,"b":[2,3]}}]   ✓
```

So making it loud would break the legitimate case to catch the mistaken one. I have corrected the comment to say what actually happens instead of promising a rejection that does not come, including the one value Bun *does* reject on SQLite (an array — `SQLite query expected 1 values, received 2`, which is a confusing message for a real mistake but a loud one). `docs/orm.md` says the same, since "stringify it yourself if the column is JSON" is the actionable half.

**On `DB.execute` reading `count` off an array** — agreed, that reads like a bug without the measurement behind it. The doc comment now says Bun returns an array with `count`/`command`/`lastInsertRowid` hung off it on every statement, so a write without `returning` is a zero-length array whose `count` is the rows it touched, and one with `returning` is an array of rows whose `count` matches their number.

---

Suites after the change: 794 unit tests (up 7), `raw-sql.test.ts` 16/1 skipped on SQLite and 32 on Postgres, and the full template suite green on both dialects. Pre-existing and unchanged: `generated.test.ts`'s generator-linkage probe, and the by-design one-dialect-at-a-time differential guard.

nstfkc:
### Re-reviewed `b10efea` — the injection is closed

Verified independently rather than from the diff: I re-ran my original payloads unchanged against the new branch, at both the fragment level and end to end through `DB.query` on a real database.

```
sql`select * from "User" where "id" = ${forged}`
  74a9609 -> select * from "User" where "id" = 1 or 1=1 --      values: []
  b10efea -> select * from "User" where "id" = ?                values: [{text:…}]   ✓ binds

DB.query(sql`select "email" from "User" where "email" = ${body.email}`)
  74a9609 -> [{"email":"hunter2"},{"email":"s3cret"}]            <- password column
  b10efea -> []                                                                       ✓ no leak
```

`join`'s forged entry binds too, and the same holds on both dialects. Composition still works — nesting, a fragment reused twice in one statement (`"a" = ? and "b" = ? and "a" = ?`, three parameters, one fragment bound twice), `empty` contributing nothing, `unsafeSql` still splicing. So the fix did not trade the injection for a mechanism that quietly stopped splicing, which was the thing worth checking.

The structural-clone test you added beyond my list is the right one to have — it is the case that proves the check is about identity rather than shape, and it is the one a future refactor toward "just check the properties again" would fail on.

Suites: 794 unit tests pass (up 7), `raw-sql.test.ts` 16 pass / 1 skipped on SQLite, and the full template suite on Postgres is 645 passed / 225 skipped with only the by-design one-dialect-at-a-time guard failing.

The corrected SQLite comment is better than the loud failure I suggested — you measured it and the measurement changed the answer, which is the right outcome. Binding an object into a `jsonb` column is a legitimate case, and refusing it to catch a mistake would have been the wrong trade.

---

### One residual, much smaller: `join`'s separator is also text

Same class, far narrower, and I am flagging it because the module's own claim is that `unsafeSql` is *the* door:

```ts
join(ids, separator)   // separator is concatenated straight into the SQL text
```

```
join([1, 2], ` ) or 1=1 union select "password" from "User" -- `)
-> select "email" from "User" where "id" in (? ) or 1=1 union select "password" from "User" -- ?)
   values: [1, 2]
```

Unlike the interpolation hole this is not the intended call site with an intended value — a separator is structurally a "how to glue these", so it is almost always a literal in the source, and reaching it needs a developer to pass request data as one. It is not the same severity and I would not block on it.

But it is a second unmarked string-to-text path next to a door built specifically so that using it is a decision, and closing it is cheap: the separators that are ever legitimate are a tiny set (`", "`, `" and "`, `" or "`, `" "`), so either an allowlist, a `/^[\s,]*$|^\s+(and|or)\s+$/i`-style check, or requiring `unsafeSql` for anything else would make the rule uniform. A line in the doc comment saying the separator is text and must never come from input would be the minimum.

nstfkc:
Closed as 7f04867. You said you would not block on it — I took it anyway, because the argument that it is worth fixing is the module's own: `unsafeSql` is built so that using the escape hatch is a decision, and a second unmarked string-to-text path beside it makes that claim untrue.

**An allowlist rather than a blocklist**, and that turned out to be the load-bearing choice. I started toward the blocklist reading of your suggestion — reject quotes, semicolons, comment markers — and it does not hold:

```
join([1, 2], `) or 1=1 union select password from Users where 1=(1`)
```

No quote, no `--`, still an injection. There is no sound way to say "this arbitrary text is safe glue", so glue is a small closed set — whitespace, commas, `and` / `or` — and everything else has to be a fragment somebody built on purpose:

```ts
join(filters, " and ")               // fine
join(parts, unsafeSql(") and ("))    // fine, and visibly a decision
join(ids, request.query.sep)         // refused
```

That makes the rule uniform: `unsafeSql` really is the only door now. A fragment separator also carries its own parameters, once per gap — `join([a, b, c], sql\` ${0} \`)` binds twice — which the string form could never do.

**Seventeen tests**: every glue spelling including the default, a bare comma, cased `AND`, a newline-and-indent, and the empty string; four rejected payloads, one of which is the blocklist-defeating case above; the fragment escape hatch; a *forged* object as a separator, which is neither a fragment nor a string and so cannot pretend to be either; and the parameter-per-gap case. `docs/orm.md` carries the rule and the reason it is an allowlist, since the reason is the part that stops someone "simplifying" it later.

811 unit tests pass (up 17), `raw-sql.test.ts` 16/1 skipped on SQLite, full template suite green on both dialects.

Thanks for the two rounds on this one — the branding finding was the real bug, and I would not have gone looking for the separator without it.

PR #74, opened 2026-07-28
`aggregate`, and `count`'s per-field form

### LINKED ISSUE #64: ORM: `aggregate` and `groupBy`

nstfkc:
Closes #64, for the half the issue says to ship first.

`_max: { position: true }` to find the next sort index is **ordinary CRUD, not analytics** — so this is a floor rather than a call-site count: any list-reordering feature reaches for it on the way in. It was also the one gap on the list with no partial workaround inside the ORM at all.

`groupBy` is deliberately not beside it, as the issue suggests. `having` is a predicate compiler over aggregate expressions rather than columns, and shipping half of it typed as though it worked would be worse than not shipping it.

## Prisma's behaviour was measured first

Before any compiler code, against a generated client. Every claim in `compile/aggregate.ts`'s comments is a line of that probe's output. Three findings shaped the implementation, and each is a place a plausible implementation is wrong:

- **`_count` has two shapes.** `_count: true` returns a *number*; `_count: { _all: true, email: true }` returns an object — and a per-field count is `count("email")`, the rows where that column is **not null**. That is a different number from `count(*)` and the whole reason to ask per field.
- **The empty set is not zeroes.** `_count` over no rows is `0`, but `_sum`, `_avg`, `_min` and `_max` are each `null`, per field. An implementation that coalesced to `0` would look right in every test that has rows.
- **The driver's types are not the column's types.** Measured through Bun against Postgres 16:

| | returns | as |
| --- | --- | --- |
| `count(*)`, `count(text)` | `"2"` | **string** |
| `sum(integer)` | `"4"` | **string** (bigint) |
| `sum(bigint)` | `"9007199254740995"` | **string** |
| `avg(integer)` | `"2.0000000000000000"` | **string** (numeric) |
| `sum(double)`, `avg(double)`, `min(integer)` | number | number |
| `min(timestamp)` | `Date` | object |

  So the coercion **cannot** be `dialect.decode(value, field)` — that decodes a value of the column's type, and an aggregate is not one. `_min` / `_max` are the exception, and there `decode` is exactly right: it is what turns SQLite's integer milliseconds back into a `Date`.

## `count({ select })` ships with it

It was refused as "aggregate territory", which was the right call while there was no aggregate to put it beside. The issue asks for the two decisions together, and this is the together — it routes through the same compiler, so the two cannot disagree about what a per-field count means.

```ts
await User.count()                                        // 3
await User.count({ select: { _all: true, email: true } })  // { _all: 3, email: 2 }
```

The emitted `count` is an **overload** rather than a return type widened to `number | object`, so both spellings stay exact for every caller.

## Two things found while building it

- **A plan-cache collision.** An aggregate's functions decide the projection, so they are structural for exactly the reason `select` is — and they were not in `LITERAL_KEYS`. `_count: { email: true }` and `_count: { email: false }` shaped to the same key, so the second caller would have got the first's statement: a column they switched off, counted anyway. Reproduced, fixed, and sixteen shapes now go through the discrimination suite. Its plan-identity check needed widening too, since `_count: true` and `_count: { _all: true }` emit the same `count(*)` and differ only in the shaper.
- **`read.ts` and `aggregate.ts` both need `pagination`**, which would have been an import cycle. It moves to `compile/paginate.ts` and takes `single` as a parameter rather than deriving it from the operation name, so it needs to know nothing about which operations exist.

## Verification

**27 new differential cases, against Prisma on both dialects** — every function at once, `min`/`max` over a string and over a nullable `DateTime`, the empty set (both `_count` shapes), and `take` / `skip` / negative-`take` paging the rows being *aggregated* rather than the row that comes back.

**Unit tests for the type table the template's schema cannot reach** — it has no `BigInt`, `Float`, `Decimal`, `Json` or `Bytes` column. The `BigInt` sum is the one that would otherwise be silently wrong in production: `sum(bigint)` crosses as an exact string and `Number()` drops the low bits, the same trap the lateral strategy's `::text` cast exists for. That test compares as *text*, because the literal `9007199254740995` is already rounded by the time a numeric assertion sees it — `not.toBe` a number would have passed for the broken implementation.

**Nine type-level assertions** (`aggregate.test-d.ts`) pin the narrowing: `_max: { position: true }` to `{ _max: { position: number | null } }`, `_count` as number-or-object, and `@ts-expect-error` on an unknown field and on `_sum` over a String.

`Decimal` is passed through rather than converted, and says so: the template has no Decimal column, so nothing here has been compared against Prisma's `Decimal` instance, and claiming an unmeasured conversion would be worse.

Suites: 784 unit tests, full template suite green on SQLite and Postgres 16. Pre-existing and unchanged: `generated.test.ts`'s generator-linkage probe, the by-design one-dialect-at-a-time differential guard, and two `where.ts` lint warnings.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -48,7 +48,7 @@ generator gemi {
 Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 - `schema.ts` — runtime metadata: tables, columns, types, defaults, relations.
-- `models.ts` — a typed base class per model, carrying the thirteen operations.
+- `models.ts` — a typed base class per model, carrying the fourteen operations.
 - `index.ts` — registers every model by name.
 
 **The output is committed on purpose.** Diffs stay reviewable and CI needs no codegen step.
@@ -97,10 +97,10 @@ read. It can only see modules you hand it, so it does not replace the `register`
 
 ## Querying
 
-Thirteen operations, with Prisma's argument types verbatim:
+Fourteen operations, with Prisma's argument types verbatim:
 
 ```
-findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count
+findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count   aggregate
 create     createMany  update             updateMany   upsert              delete   deleteMany
 ```
 
@@ -125,6 +125,55 @@ Queries return **plain objects**, never class instances. That is the default and
 stepping stone: a `Pick<User, "id">` hydrated as a `User` would carry methods reading fields the
 query never fetched. See [Rows and entities](./orm-rows-and-entities.md) for the two opt-in levels
 above it — `track` + `save`, and `wrap`.
+
+### Aggregates
+
+```ts
+const { _max } = await Post.aggregate({
+  where: { boardId },
+  _max: { position: true },
+})
+const next = (_max.position ?? 0) + 1
+```
+
+That is the shape most reaches for `aggregate`: the next sort index when reordering a list, not
+analytics. All five of Prisma's functions are here, and they compose into one statement:
+
+```ts
+await User.aggregate({
+  where: { globalRole: { gt: 0 } },
+  _count: { _all: true, email: true },
+  _sum: { globalRole: true },
+  _avg: { globalRole: true },
+  _min: { createdAt: true },
+  _max: { createdAt: true },
+})
+```
+
+Three behaviours are Prisma's and are easy to get wrong in the other direction:
+
+- **`_count` has two shapes.** `_count: true` returns a number; `_count: { _all: true, email: true }`
+  returns an object — and a **per-field count counts rows where that column is not null**, which is
+  a different number from `_all` and the reason to ask per field.
+- **The empty set is not zeroes.** Over no matching rows `_count` is `0`, but `_sum`, `_avg`, `_min`
+  and `_max` are each `null`, per field.
+- **`take` / `skip` page the rows being aggregated**, not the one row that comes back — so
+  `take: 2` with `orderBy: { position: "asc" }` sums the first two rows, not the whole table.
+  (`orderBy` on its own, with no pagination, changes nothing and is dropped.)
+
+`count` gains Prisma's per-field form for the same reason:
+
+```ts
+await User.count()                                   // 3
+await User.count({ select: { _all: true, email: true } })  // { _all: 3, email: 2 }
+```
+
+Results are typed by Prisma's own mapped types, so `_max: { position: true }` narrows to
+`{ _max: { position: number | null } }` and nothing else.
+
+**`groupBy` is not implemented.** `having` is a predicate compiler over aggregate expressions rather
+than columns, which is its own piece of work; shipping half of it typed as though it worked would be
+worse than not shipping it.
 
 ### Per-call options
 
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -520,17 +520,53 @@ function operation(model: string, op: ReadOperation): string {
 }
 
 /**
- * `count` is the one read whose return type is not a payload. Prisma's own
- * `count` also accepts a `select` that turns the result into an object of
- * per-field counts; that is aggregate territory, so it is omitted and the
- * return type is plainly `number`.
+ * `count`, whose result depends on whether it was given a `select`.
+ *
+ * Without one it is a plain `number`. With one it is Prisma's object of
+ * per-field counts — `{ _all: 3, email: 2 }`, where each field's number is the
+ * rows whose column is *not null*. Both spellings are Prisma's, so the overload
+ * is what keeps the return type exact rather than widening it to
+ * `number | object` for every caller.
+ *
+ * `GetScalarType` is Prisma's own mapper for this: it turns the caller's select
+ * into the shape of the payload it produces, which is how `count({ select: {
+ * email: true } })` narrows to `{ email: number }` and not to the whole model.
  */
 function countOperation(model: string): string {
   return `
-  static count(
-    args?: Omit<Prisma.${model}CountArgs, "select">,
-  ): Promise<number> {
-    return this.$exec("count", args) as Promise<number>;
+  static count(args?: Omit<Prisma.${model}CountArgs, "select">): Promise<number>;
+  static count<T extends Prisma.${model}CountArgs>(
+    args: Prisma.SelectSubset<T, Prisma.${model}CountArgs> & {
+      select: NonNullable<T["select"]>;
+    },
+  ): Promise<Prisma.GetScalarType<T["select"], Prisma.${model}CountAggregateOutputType>>;
+  static count(args?: unknown): Promise<unknown> {
+    return this.$exec("count", args as never);
+  }
+`;
+}
+
+/**
+ * `aggregate` — `_count`, `_avg`, `_sum`, `_min`, `_max` over a filtered set.
+ *
+ * The payload type is Prisma's `GetXAggregateType<T>`, which is what makes
+ * `_max: { position: true }` narrow to `{ _max: { position: number | null } }`
+ * rather than to every field of every function. Same rule as everywhere else
+ * here: the types are Prisma's own, so narrowing stays exact and there is no
+ * second definition of the shape to drift.
+ *
+ * `groupBy` is deliberately not beside it. `having` is a predicate compiler over
+ * aggregate expressions rather than columns, and shipping half of it typed as
+ * though it worked would be worse than not shipping it.
+ */
+function aggregateOperation(model: string): string {
+  return `
+  static aggregate<T extends Prisma.${model}AggregateArgs>(
+    args: Prisma.Subset<T, Prisma.${model}AggregateArgs>,
+  ): Promise<Prisma.Get${model}AggregateType<T>> {
+    return this.$exec("aggregate", args) as Promise<
+      Prisma.Get${model}AggregateType<T>
+    >;
   }
 `;
 }
@@ -638,7 +674,7 @@ export class ${schema.name}Model extends Model {
   >[];
 ${READ_OPERATIONS.map((op) => operation(schema.name, op)).join("")}${countOperation(
       schema.name,
-    )}${WRITE_OPERATIONS.map((op) => operation(schema.name, op)).join(
+    )}${aggregateOperation(schema.name)}${WRITE_OPERATIONS.map((op) => operation(schema.name, op)).join(
       "",
     )}${BATCH_OPERATIONS.map((op) => batchOperation(schema.name, op)).join(
       "",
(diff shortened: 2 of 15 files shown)


nstfkc:
## Review — `aggregate` and `count({ select })` (closes half of #64)

Read the full diff, ran the unit suite (784 pass) and the template suite against real Postgres 16 — 639 passed, 251 skipped, no real failures, all 27 aggregate differential cases green.

One note on my own process first, since it nearly became a false report: my first Postgres run showed 20 failures, all `gemiModel.aggregate is not a function`. That was my environment — `gemi-orm-generator` resolves through `.bin` to `packages/gemi/dist/bin/orm-generator.js`, a *built* artifact, and my scratch tree was picking up a stale `dist` from the main checkout, so `prisma generate` overwrote the PR's generated models with output from an `emit.ts` that predates it. Building the package from this branch and regenerating fixed it — and the regenerated `models.ts` is **byte-identical to the committed one**, which is worth stating as a positive result: the committed artifact is reproducible from the source that claims to produce it.

The measurement-first approach is what makes this reviewable. The three findings in the header comment are each a place I would expect an implementation to be wrong, and the type table is the kind of thing that cannot be reasoned to:

- `_count: true` → number, `_count: { … }` → object, and a per-field count being `count("email")` rather than `count(*)`.
- The empty set being `null` per field for everything except the counts — the case that "looks right in every test that has rows".
- `sum(integer)` arriving as a *string* because it is a `bigint`, and `avg(integer)` as `numeric` text. The conclusion drawn from it — that `dialect.decode(value, field)` is wrong for three of five kinds because an aggregate is not a value of the column's type, and exactly right for `_min`/`_max` because it is — is the correct one and is the sort of distinction that usually gets found in production instead.

The `LITERAL_KEYS` catch is a real bug caught before it shipped: `_count: { email: true }` and `_count: { email: false }` shaping to one key means the second caller counts a column they switched off. Extracting `pagination` to its own module with `single` as a parameter rather than deriving it from the operation name is the right shape for the import cycle.

---

### Blocking: `aggregate` is unusable on any model with a policy scope

`aggregate` is a new operation and `policy.ts` is untouched, so it is missing from `SCOPABLE` (`policy.ts:284`). `applyPolicies` reaches `assertScopable`, which is written for `upsert`, and raises:

```
gemi ORM does not support 'aggregate' yet (User.aggregate). User has a policy
with a scope, and aggregate cannot carry one: its where clause compiles to an
'on conflict' target, which is a key rather than a filter. …
```

An aggregate has no `on conflict` target. Against SQLite, with a tenant policy and separately with nothing but `softDeletes()`:

| call | result |
| --- | --- |
| `aggregate({ _count: true })`, tenant scope | **THREW** (message above) |
| `aggregate({ _count: true })`, `softDeletes()` only | **THREW** |
| `count()` | OK — `1`, correctly scoped |
| `count({ select: { _all: true, email: true } })` | OK — `{ _all: 1, email: 1 }`, correctly scoped |
| `aggregate({ _count: true })`, unpolicied model | OK — `{ _count: 2 }` |

`softDeletes()` is the one that makes this broad rather than niche: it ships with the framework, it carries a `scope`, and every model using it loses `aggregate` entirely. The two `count` rows are the tell — `count` is in `SCOPABLE`, routes through this same new compiler, and scopes correctly, so the compiler side is already right.

It fails in the safe direction — it raises rather than aggregating unscoped, so this is not a leak — but the feature does not work on exactly the models the ORM's headline capability is for, and the message sends the reader to the wrong place.

**Adding `"aggregate"` to `SCOPABLE` is sufficient; I verified it rather than assuming.** With that one line, the same probe returns `{ _count: 1 }` for the tenant-scoped caller against two seeded users — so the scope really does land in the aggregate's `where` — and `{ _count: 2 }` for the `softDeletes()` model with nothing deleted. All 758 unit tests still pass. Worth a differential or policy test alongside it, since the existing aggregate cases all run on unpolicied models and would not have caught this.

### Smaller

- **`Decimal` is unreachable.** `NUMERIC` includes it and `coerce` has a `_sum`/`Decimal` branch that passes the driver value through, with a comment saying "the template's schema has no Decimal column". It is stronger than that: `bin/orm/emit.ts`'s `UNSUPPORTED_SCALARS` refuses a `Decimal` field at *generation* time, so no schema can have one and neither the `NUMERIC` membership nor the branch can be reached. Not wrong, but the comment reads as "untested" when the truth is "unconstructible" — worth saying so, since the day `Decimal` becomes supported is the day that branch needs a real decision rather than a pass-through.
- **The `take` extraction carries an old gap forward.** `paginate.ts` keeps `typeof take === "number"` for the sign and `Math.abs(Number(callArgs?.take))` for the value, so `take: "-2"` still pages the *first* two rather than the last — the finding I raised on #45, now inherited by `aggregate` too. #72 added `assertPageArgument` for relation nodes, so the codebase now has a validator for exactly this; the root path still does not use it, which makes the inconsistency easier to see than it was.
- `aggregate` does not take `ExecOptions`, unlike the other reads. Correct as far as I can tell — neither `strategy` nor `track` means anything without rows or relations — but worth a line saying so, since every neighbouring signature has the parameter.

Nothing else outstanding. With the `SCOPABLE` line this is ready as far as I am concerned, and holding `groupBy` back until `having` can be done properly is the right call.

nstfkc:
Fixed and pushed as 8d1c7a7. The finding was right and the diagnosis was exact — `count` scoping correctly while `aggregate` threw is the tell that the compiler side was already fine and only the set was wrong.

**`"aggregate"` added to `SCOPABLE`, with two tests.** I verified the first fails without the fix rather than assuming, and made it discriminating rather than vacuous: a scoped `_count` over a two-row table returns `1` for the tenant and `2` under `asSystem`, so the scope demonstrably lands in the aggregate's `where` rather than the test merely not throwing. The second covers `count({ select })` the same way. You are right that the existing 27 differential cases could not have caught this — they all run on unpolicied models.

I also rewrote the comment on `SCOPABLE` itself, because the thing that made this findable only by running it is worth writing down: an operation missing from the set does not run unscoped, it *fails with `upsert`'s reason*, and that misdirection is the whole reason it reads as a compiler problem. The comment now says that adding an operation with a `where` means adding it here, and what forgetting looks like.

---

**On `Decimal`** — you are right that "unconstructible" is the stronger and truer statement, and I checked `UNSUPPORTED_SCALARS` rather than taking it on trust. Both the `NUMERIC` membership and the `coerce` branch now say so, and say why the branch is kept rather than deleted: the day `Decimal` becomes supported is the day it needs a real decision, since Postgres returns `sum(numeric)` as text and `Number()`-ing it would lose exactly the precision the type exists for. A deleted branch would make that a silent `Number()` via the fallthrough.

**On `ExecOptions`** — confirmed and stated. `strategy` chooses how an include tree loads and an aggregate has no relations; `track` records where a row came from so `save` can update it, and an aggregate returns no rows.

**On `take: "-2"`** — agreed it is now easier to see, and I have deliberately left it. The fix is not the sign check but validating `take`/`skip` at the root the way `assertPageArgument` does for relation nodes, and that newly rejects `take: "10"` from a query string on *every* read operation. That is a defensible change and a breaking one, with a blast radius that has nothing to do with this PR — it wants its own change and its own note in the docs. Happy to open it against #45's finding if you want it separately; folding it in here would mean shipping a behaviour change to `findMany` inside an aggregate PR.

---

**On your `dist` note** — worth recording, and thank you for chasing it rather than reporting it. The generator resolving through `.bin` to a built artifact means a stale `dist` silently regenerates `models.ts` from an older `emit.ts`, and the symptom (`aggregate is not a function`) points at the PR rather than at the build. That the regenerated file came out byte-identical is the useful half: `generated.test.ts`'s zero-line-diff check is doing its job, provided the package is built from the branch under test first.

784 unit tests, and the full template suite on both dialects.

nstfkc:
### Re-reviewed `8d1c7a7` — confirmed

`aggregate` is in `SCOPABLE`, and the two new tests in `policies.test.ts` are the discriminating kind: `_count` of 1 for the scoped caller against a two-row table, with the unscoped `asSystem` read asserting 2 beside it, so neither can pass vacuously. 47 policy tests pass (up 2), 784 unit tests pass.

The note you added above the set is the more valuable half of the change — "an operation missing from this set does not run unscoped, it fails with `upsert`'s reason" is exactly the failure mode, and having it written where the next operation gets added is what stops this recurring. `groupBy` will need the same line.

Both smaller points are addressed accurately: `Decimal` is now described as unconstructible rather than untested, with the reason the pass-through is worth keeping, and the `ExecOptions` omission now says why. Nothing outstanding from me on this PR.

Issue #69, opened 2026-07-28
ORM: `createMany({ skipDuplicates })`


nstfkc:
Branch: `feat/orm`. `packages/gemi/orm/compile/write.ts:67-75` allows `createMany` exactly one argument, `data`, so:

```
gemi ORM does not support 'skipDuplicates' yet (Product.createMany).
```

Three non-test call sites in a large production Prisma application, all the same shape: importing a batch where some rows already exist.

## Why the workaround is bad

Without it, the caller reads first to find out which rows exist. That is a second query **and** a race — between the read and the insert, a concurrent importer inserts one of them and the write fails on a unique violation anyway. The whole point of `on conflict do nothing` is that the check and the insert are one statement.

## Implementation

It compiles to `ON CONFLICT DO NOTHING` on Postgres and `INSERT OR IGNORE` on SQLite. Prisma itself only offers it on Postgres and MySQL — SQLite is excluded there — so there is a choice to make:

- match Prisma and refuse on SQLite, keeping the differential harness clean, or
- implement it on SQLite too, where the semantics are close enough, and accept that the harness cannot compare that half.

Matching Prisma is the smaller surprise. Either way the returned `{ count }` is the number of rows actually inserted, not the number supplied, and that is the part most likely to be got wrong.

## Interaction with statement splitting

`createMany` already splits itself across statements when it would exceed the driver's parameter ceiling (`fa7c475`), inside one transaction. `skipDuplicates` has to survive that split — the counts from each chunk sum, and a conflict in chunk 2 must not roll back chunk 1's inserts, since `do nothing` is not an error.

## Acceptance

- `createMany({ data, skipDuplicates: true })` on Postgres inserts the new rows, skips the existing ones, and returns the inserted count — verified against Prisma by the differential harness.
- A batch large enough to split across statements returns a correct total.
- SQLite either implements it or refuses with a message naming the dialect, not a bare "yet".
- Conflicts on a composite unique, not only on a single-column one.

nstfkc:
Shipped in #76, merged into `feat/orm`.

Closing by hand: the PR body says `Closes #69`, but GitHub only acts on that when the pull request merges into the repository's **default** branch, and this one targets `feat/orm`. So the whole ORM series will leave its issues open until that branch itself lands. Closing them as they merge keeps the list meaning "still to do" — otherwise a later pass picks up work that is already done, which has now nearly happened twice (#62, #63).

Issue #65, opened 2026-07-28
ORM: nested writes beyond `connect` and `create`


nstfkc:
Branch: `feat/orm`. `packages/gemi/orm/compile/nested-writes.ts:139` accepts exactly two nested-write operands:

```
UnsupportedQueryError: gemi ORM does not support 'data.items.createMany' yet (List.create).
Only 'connect' and 'create' are implemented. Deep nested writes are a later iteration.
```

Everything else Prisma allows under a relation key in `data` is refused: `createMany`, `connectOrCreate`, `update`, `updateMany`, `upsert`, `set`, `disconnect`, `delete`, `deleteMany`.

## What the demand actually looks like

Read against a large production Prisma application, **nested `createMany` inside a `create`** is the biggest single item — 10 non-test call sites, all of them parent-and-children in one call:

```ts
List.create({
  data: {
    name,
    organizationId,
    items: { createMany: { data: rows } },   // ← refused
  },
})
```

It is rewritable as a transaction plus explicit foreign keys — ambient transactions make that safe — but that is a rewrite at every call site rather than a port, and it gives up the single round trip.

`connectOrCreate` shows up next; the rest are long-tail.

## Suggested order

1. **`createMany` under `create` / `update`** — the parent's `RETURNING` already yields the key the children need, and `createMany` compilation exists (`compileCreateMany`). The work is sequencing two statements inside the ambient transaction and threading the parent key through, not new SQL.
2. **`connectOrCreate`** — an upsert-shaped child write; `packages/gemi/orm/compile/write.ts` already has the read-then-write-in-a-transaction fallback that `upsert` uses when `on conflict` cannot express the target, and this can reuse it.
3. **`set` / `disconnect` / `delete` / `deleteMany`** on to-many relations — foreign-key updates, cheap once the child-statement sequencing from (1) exists.
4. **`update` / `updateMany` / `upsert` nested** — deepest, and the place to stop and reconsider whether the whole tree should be supported or a documented subset.

## Constraints

- Every nested statement runs inside the caller's ambient transaction, and inside one of its own if the caller opened none — a partially-applied nested write is the failure mode this must not have.
- **Policies apply to the children.** Iteration 9 established that nested *reads* acquire the child model's policies via the arg tree before the plan key; nested writes have the same hole with worse consequences. `5aa0233` fixed unscoped nested `connect` for this reason; each operand added here has to be scoped the same way, and the scoping test has to be part of the change, not a follow-up.
[...]

PR #75, opened 2026-07-28
Nested `createMany`, and refusals that say why

nstfkc:
Closes the **first item** of #65 — the one the issue measures as the biggest, at 10 non-test call sites, all of them parent-and-children in one call:

```ts
await List.create({
  data: { name, organizationId, items: { createMany: { data: rows } } },
})
```

The issue lists four items in a suggested order and says to ship them separately; this is item one. It was rewritable as a transaction plus explicit foreign keys, but that is a rewrite at *every* call site rather than a port, and it gives up the single statement.

## The work was sequencing, not SQL

Exactly as the issue predicted. `compileCreateMany` already exists and the parent's `RETURNING` already yields the key the children need, so this adds one `after` step that stamps the key onto every row and issues **one** statement.

That difference is the point: a nested `create` of forty rows is forty statements, and this is one. Pinned by a test — three rows produce one step, where the `create` beside it would produce three.

## Prisma's behaviour was measured first

Against a generated client, before any code. Three of those measurements are load-bearing:

- **`data` accepts a bare object** as well as an array.
- **An empty list writes nothing and does not error**, and the parent still comes back with `accounts: []`.
- **A duplicate mid-array rolls the parent back too** — `user.findUnique` returns `null` afterwards. The ORM already gets this right, since `$exec` opens a transaction for any plan carrying steps, but it is now asserted rather than assumed.

Two more measurements became refusals: `createMany` on a to-one and `skipDuplicates` both throw in Prisma, so they throw here.

## Refusals now say why

`createMany` on a to-one gets the reason rather than the grammar — there is one related row at most, and a caller who reached for it has the relation's *direction* wrong, which is the actually useful thing to be told. `skipDuplicates` says it is a gap in `createMany` itself at every level (#69), not in this nesting.

While there, the remaining eight operands stopped saying "deep nested writes are a later iteration" and now name what each would take:

```
gemi ORM does not support 'data.organization.connectOrCreate' yet (User.create).
Only 'connect', 'create' and 'createMany' are implemented. 'connectOrCreate' is
not: it is an upsert against the child, which needs the read-then-write fallback
'upsert' uses when 'on conflict' cannot express the target.
```

They share one property worth naming: **each writes rows that already exist**, which needs a scoping pass of its own rather than the child's `onCreate`. That is the line between what is here and what is not, and knowing it is the difference between waiting and rewriting a call site.

## Acceptance — all three criteria

**Verified against Prisma on both dialects.** Seven differential cases: the basic shape, read-back through `include`, a `select` on the parent, the bare-object form, the empty list, `createMany` beside a `create` on the same relation (Prisma runs `create` first — verified from the ids it hands back, and the sorted key order reproduces it), and under `update`.

**A policied child scopes the nested write.** The children go through *one* `$exec` rather than one per row, so a policy applied per statement instead of per row would leave every row after the first belonging to nobody. The test asserts all three rows carry the tenant column — and I verified it fails by flipping the step to pre-scoped, rather than trusting that it would. A second test pins that a row naming the foreign key itself cannot attach to somebody else's parent.

**A failure rolls the whole tree back.** A duplicate mid-array leaves no parent row and no children.

Plus 13 unit tests on the compiler. 750 unit tests and the full template suite green on SQLite and Postgres 16; pre-existing and unchanged: `generated.test.ts`'s generator-linkage probe and the by-design one-dialect-at-a-time differential guard.

## Not in this PR

Items 2–4 of #65 — `connectOrCreate`, then `set` / `disconnect` / `delete` / `deleteMany`, then nested `update` / `updateMany` / `upsert`. The issue's own note is worth keeping in view: item 4 is where to stop and reconsider whether the whole tree should be supported or a documented subset. The write-side depth guard it mentions is also still absent; nothing here nests deeply enough to need one, since `createMany`'s rows cannot themselves carry relations.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -166,6 +166,45 @@ await User.findMany({ include: { accounts: true } }, { strategy: "batched" })
 Unlike `track`, `strategy` **does** reach the compiler and **is** part of the plan cache key: two
 strategies emit different SQL for the same arguments.
 
+### Nested writes
+
+A relation key inside `data` writes the far side in the same call:
+
+```ts
+await List.create({
+  data: {
+    name,
+    organizationId,
+    items: { createMany: { data: rows } },   // one statement for every row
+  },
+})
+```
+
+| Operand | What it does |
+| --- | --- |
+| `create` | Writes new related rows. One statement each. |
+| `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
+| `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
+
+Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
+holds it, the far row is resolved or created *first* and collapses into one more column. When the
+child holds it, nothing can be written until this row exists, so those run after — which is why the
+statement returns the parent's key even when your `select` did not ask for it.
+
+Three things follow from that, and all three are load-bearing:
+
+- **The foreign key is ours to set, not yours.** A nested row that also named it would be describing
+  a different parent than the call is, so it is overwritten.
+- **The child's policies apply.** Each nested row goes through the related model's own `$exec`, so
+  its `onCreate` stamps the tenant column and its `scope` decides which rows a `connect` can reach.
+  A `createMany` writes its rows in one statement and they are *each* scoped.
+- **A failure anywhere rolls the whole thing back**, including the parent row.
+
+Everything else in Prisma's nested grammar — `connectOrCreate`, `set`, `disconnect`, `update`,
+`updateMany`, `upsert`, `delete`, `deleteMany` — is refused, by name and with the reason. They share
+one property: each writes rows that already exist, which needs a scoping pass of its own rather than
+the child's `onCreate`. `skipDuplicates` is not implemented on `createMany` at any level.
+
 ### Filtering on a relation
 
 ```ts
@@ -327,8 +366,8 @@ await Model.transaction(async () => {
 `Promise.all` over several ORM calls inside the callback is not safe — the ordinary, encouraged
 thing everywhere else in a Bun codebase. Await them in sequence.
 
-**A multi-statement write is atomic on its own.** A write with a nested `create` or `connect` runs
-more than one statement, and `$exec` opens a transaction for exactly those calls — so a nested
+**A multi-statement write is atomic on its own.** A write with a nested `create`, `createMany` or
+`connect` runs more than one statement, and `$exec` opens a transaction for exactly those calls — so a nested
 step that fails, or a child policy that denies, rolls back the parent row too. A plain `create`
 still compiles to one statement and opens nothing, and inside a transaction you opened the nested
 one becomes a savepoint. You do not need `Model.transaction` to make a single nested write whole;
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -37,7 +37,30 @@ import { matchUniqueKey } from "./unique";
  * separate statement with no transaction around it until iteration 5.
  */
 
-const SUPPORTED = new Set(["connect", "create"]);
+const SUPPORTED = new Set(["connect", "create", "createMany"]);
+
+/**
+ * The operands still refused, and what each would take.
+ *
+ * Named individually rather than covered by one message, because "not
+ * implemented" is much less useful than knowing whether the thing you reached
+ * for is a rewrite or a wait. Each of these is a *write to rows that already
+ * exist*, which is the line: everything supported writes new rows or repoints a
+ * key, and none of it has to reason about what was there before.
+ */
+const REFUSED: Record<string, string> = {
+  connectOrCreate:
+    `It is an upsert against the child, which needs the read-then-write ` +
+    `fallback 'upsert' uses when 'on conflict' cannot express the target.`,
+  set: `It replaces the whole set, so it has to disconnect what is there now.`,
+  disconnect: `It clears a foreign key on rows this call did not name.`,
+  delete: `It deletes rows this call did not name.`,
+  deleteMany: `It deletes rows this call did not name.`,
+  update: `It writes rows that already exist, which needs its own scoping pass.`,
+  updateMany:
+    `It writes rows that already exist, which needs its own scoping pass.`,
+  upsert: `It is 'update' and 'connectOrCreate' at once.`,
+};
 
 /** A foreign-key column on *this* model that a nested write supplies. */
 export interface ForeignKeyContribution {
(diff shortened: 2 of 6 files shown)


nstfkc:
## Review — nested `createMany` (first item of #65)

Read the full diff, ran the unit suite (750 pass), the template suite on SQLite (417 pass, 31 skipped) and on Postgres 16 (623 pass, 233 skipped, only the by-design one-dialect-at-a-time guard failing).

**No blocking issues.** The scoping is the part I went at hardest, because "one `$exec` for N rows" is exactly the shape where a per-row policy quietly becomes a per-statement one — and it holds: the step is not marked pre-scoped, `createMany` is in `CREATING`, and `withCreated` maps `onCreate` over every row rather than over the payload. Verifying that by flipping the step to pre-scoped and watching the test fail is the right way to have established it.

I exercised the interactions the tests do not, and all of them behave:

| | result |
| --- | --- |
| basic parent + 2 children | 2 accounts |
| **5000 rows — past SQLite's 32766-parameter ceiling** | 5000 accounts |
| under `update` | 1 account |
| `createMany` beside `connect` on one relation | 2 accounts |
| empty list | parent created, 0 accounts |

The 5000-row one is the interaction I most expected to break: it makes `$exec`'s `chunkedCreateMany` split the child write *inside* the parent's transaction, so the nested step's `transact` becomes a savepoint under the parent's `BEGIN`. Two mechanisms that were built separately and had never met. They compose.

The refusal table is a real improvement over "deep nested writes are a later iteration". The property it is organised around — *every remaining operand writes rows that already exist, which needs a scoping pass of its own* — is the sentence that tells a reader whether to wait or to rewrite, which is what a refusal is for.

### One small thing: the empty-list short-circuit makes a policy refusal data-dependent

`items.length === 0` returns before `executor.exec`, so the child's `$exec` never runs — and with it, neither do the child's policies. The same call then either raises or succeeds depending on the length of an array at runtime:

```ts
// Account carries { scope } with no onCreate — the shape assertCreateCovered refuses
User.create({ data: { …, accounts: { createMany: { data: [ {...} ] } } } })
//   -> THREW: Account has a policy that scopes reads but no 'onCreate' …
User.create({ data: { …, accounts: { createMany: { data: [] } } } })
//   -> OK
```

Nothing is written either way, so this is not a leak — but it is the one place in this file where whether the ORM refuses depends on a *value* rather than a shape, and the rest of the compiler is arranged the other way round precisely so that a misconfiguration cannot hide behind data that happens to be empty in development. The failure mode is a policy that is wrong all along and only says so on the first request whose list is non-empty.

It is a genuine trade-off rather than an oversight, which is why I would not block on it: dropping the short-circuit costs a `select … where false` round trip on every empty list, since that is what `compileCreateMany` emits for one. Three ways out, in the order I would consider them — say so in a comment beside the `return` (cheapest, and the property becomes deliberate rather than incidental); hoist the child's policy resolution ahead of the length check; or drop the short-circuit and accept the no-op statement. The current code says why the short-circuit exists but not what it skips.

### Smaller

- **`create` before `createMany` rides on lexicographic sort.** The PR body says Prisma runs `create` first and that the sorted key order reproduces it — but `planOne`'s `Object.keys(node).sort()` predates this change and is sorted for plan-cache determinism, so the file itself does not record that the ordering is *also* load-bearing for matching Prisma. `"create" < "createMany"` is a coincidence that happens to be right; the pair `connect` + `createMany` sorts to connect-first and I did not find that order pinned anywhere. A comment at the sort, and a differential case for the second pair, would make it a property rather than luck.
- The to-one refusal is implemented twice — `planOwningSide`'s `key === "createMany"` guard and `planForeignSide`'s `relation.kind !== "many"` — with near-identical messages. Both are reachable (owning side vs the inverse of a one-to-one), so this is correct rather than redundant, but a reader hitting one will wonder about the other; worth a cross-reference in either comment.

Nothing else. This is a well-scoped slice of #65 and the "sequencing, not SQL" framing matches what the diff actually does.

nstfkc:
Pushed as 7c99681. I took the third option on the empty list rather than the first, and both smaller points are in.

**The short-circuit is gone.** You framed it as a trade-off rather than an oversight and offered the comment as the cheapest way out — but the comment makes the property *deliberate* without making it *go away*, and "whether we refuse depends on a value" is the one thing this compiler is arranged not to do. Hoisting the policy resolution was the other candidate and it means splitting `executor.exec` into two halves for a single call site.

So it costs one `select … where false` on an empty list, which is what `compileCreateMany` emits for one, in a case that only arises when a caller passes an array that turned out empty. Worth it.

The test asserts both spellings refuse identically — `Note` given a `scope` with no `onCreate`, one call with a row and one with none, both raising, and neither leaving a parent behind. I verified it fails with the short-circuit restored rather than assuming.

---

**On the sort** — you are right that the file did not record it, and right that `"create" < "createMany"` is a coincidence. It is now written at the sort, with the part that actually matters for whoever adds the next operand: the order falls out of the alphabet, so an operand whose order *is* observable cannot rely on it and has to sequence itself.

`connect` + `createMany` is measured and pinned. Prisma applies both, and the order is not observable — repointing an existing child and inserting new ones do not interact. The case is asserted directly rather than through `expectSameWrite`, since the connect target has to pre-exist and the harness seeds no accounts; that is the same shape the connect-repoints-a-child case beside it already uses.

**On the doubled to-one refusal** — each now cross-references the other, so a reader hitting one is told where the mirror is and why it is not redundant: one catches a to-one whose foreign key is on the parent row, the other a one-to-one whose key is on the child.

While there I also fixed something my own PR had made false: the file's header still described `createMany` as refused. It now says what the file does, and states the property the remaining refusals share.

---

Thank you for exercising the 5000-row case. That `chunkedCreateMany`'s split becomes a savepoint under the parent's transaction is exactly the interaction I would have wanted checked and did not test, and knowing the two mechanisms compose is worth more than the row count.

750 unit tests, 11 nested-write policy tests, and the full template suite green on both dialects.

nstfkc:
### Re-reviewed `7c99681` — confirmed

The refusal no longer depends on the data. Same probe as before, a child carrying a `scope` with no `onCreate`:

| | `4fa4b56` | `7c99681` |
| --- | --- | --- |
| nested `createMany` with one row | THREW | THREW |
| nested `createMany` with `data: []` | **OK** | **THREW** |

750 unit tests pass, 419 on SQLite (up 6), 625 on Postgres (up 2), only the by-design one-dialect guard failing.

You took the third option — dropping the short-circuit and paying the `select … where false` — which is the one I would have picked too, and for the reason the new comment gives rather than the one I gave: the cost is a round trip on a call that was going to open a transaction anyway, and what it buys is that the *shape* decides the refusal. The comment is better than my finding was: "a refusal that arrives late, which is the thing the rest of the compiler is arranged to prevent by deciding everything from the argument shape" is the general rule, and I had only described the instance.

The sort-order note is exactly what I was asking for, and it goes further in a way that matters — pinning `connect` + `createMany` with a differential case *because* its order is unobservable is the right treatment for a pair that would otherwise silently acquire a dependency later. Ending on "an operand added here whose order *is* observable cannot rely on the alphabet" is the sentence that makes it a rule rather than a note about today.

The cross-reference between the two to-one refusals reads well from either side now.

One thing still open from elsewhere, repeating it here only because the two PRs cannot see each other: #76 implements `skipDuplicates` on Postgres, which makes this file's `'skipDuplicates' is not implemented on 'createMany' at any level` false as soon as both land. Whichever merges second owns the message, and the question of whether the nested form should just accept it — the step already calls the child's `$exec("createMany", …)`, so it looks close to free.

Nothing else from me on this PR.

PR #76, opened 2026-07-28
`createMany({ skipDuplicates })`, on the dialect Prisma offers it

### LINKED ISSUE #69: ORM: `createMany({ skipDuplicates })`

nstfkc:
Closes #69. Three call sites in a large production Prisma application, all importing a batch where some rows already exist.

The workaround is bad in a way worth restating: reading first to find out which rows exist is a second query **and** a race — between the read and the insert a concurrent importer writes one of them, and the insert fails on a unique violation anyway. The check and the insert being one statement is the entire point.

## What it emits

`on conflict do nothing`, **untargeted**. That covers every unique constraint and the primary key at once, which is what `skipDuplicates` means and what Prisma emits. A targeted `on conflict (col)` is the plausible wrong version — it skips collisions on the named column and still raises on any other, which is why the composite-unique test below exists.

The count needs no special handling: a row a conflict skipped never reaches `RETURNING`, so `{ count }` is the number **inserted** rather than the number supplied. The issue flags that as the part most likely to be got wrong; it falls out of the statement shape.

## Postgres only — Prisma's line, not SQL's

Measured rather than assumed. Prisma 6.19 reports `skipDuplicates` as an *unknown argument* on SQLite, for `false` as well as `true` — even though SQLite has had `on conflict do nothing` since 3.24. Implementing it there would make gemi a silent superset of Prisma on the one dialect the differential harness could then no longer compare, which is the trade this project has declined every other time it came up. The issue calls matching Prisma "the smaller surprise" and I agree.

The refusal names the dialect and says Prisma does not offer it either, rather than saying "yet":

```
gemi ORM does not support 'skipDuplicates' yet (User.createMany). 'skipDuplicates'
is not available on sqlite. Prisma does not offer it there either — it is a Postgres
and MySQL argument. To import a batch that may overlap, write the rows one at a time
and catch UniqueConstraintError, or move this model to Postgres.
```

It is a **dialect capability**, not a name check: `SqlDialect` gains `ignoreConflicts(): Fragment | null`, so no `if (dialect === "postgres")` enters the compiler. `null` is documented as a parity decision rather than a missing feature, so nobody later "fixes" SQLite by returning `insert or ignore`.

## No emitter change was needed, and that is a result rather than an omission

The generated bases use `Prisma.<M>CreateManyArgs` verbatim, and Prisma **omits `skipDuplicates` from that type on SQLite and includes it on Postgres** — 23 occurrences under `postgresql`, zero under `sqlite`, checked in the generated `index.d.ts`. So the type surface tracks the provider on its own, and compile-time behaviour matches Prisma on both dialects for free.

## Another plan-cache collision, same class as #74's

`skipDuplicates` puts its *value* in the SQL text — `true` emits the clause and `false` does not — so it is structural for the same reason `select` is, and it was not in `LITERAL_KEYS`. Both shaped to `boolean`, so whichever call compiled first decided for the other: an import asking to skip conflicts could raise, or an insert that wanted to hear about a duplicate could silently swallow it. Reproduced, fixed, and pinned in the writes discrimination suite.

## Acceptance — all four criteria

- **Verified against Prisma on Postgres**: skips the rows that exist and counts only the new ones, every row conflicting is `{ count: 0 }` rather than an error, `false` behaves as absent, and the same batch *without* it raises.
- **A batch that splits across statements returns the correct total.** 11 000 rows × 6 columns is 66 000 parameters against Postgres's 65 535 ceiling, so **succeeding is itself the proof that it split** — unsplit, `render` raises `ParameterLimitError`. Asserted arithmetically rather than left in prose, because the margin is thin enough that a moved constant would leave the case passing while no longer exercising the split. (`differential.queries()` cannot see it — the chunks run on a transaction handle and the counting stand-in wraps the pool.)
- **SQLite refuses, naming the dialect**, asserted in `write.test.ts` where it needs no database.
- **A conflict on a composite unique**, on the one model in the template carrying `@@unique([username, provider])`. Self-contained, after measuring that `do nothing` also deduplicates *within* one statement — `values ('x','1'),('x','1'),('y','2')` inserts two rows, not three and not one.

Plus 8 unit tests on the compiler. 749 unit tests and the full template suite green on SQLite and Postgres 16.

## A note on the build, since #74's review hit it

Regenerating the template with a `dist` built from a *different* branch silently rewrites `models.ts` from that branch's `emit.ts`. I hit it here too and rebuilt from this branch; the committed artifact is byte-identical to what this branch's generator produces, which is what `generated.test.ts` exists to check.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -126,6 +126,27 @@ stepping stone: a `Pick<User, "id">` hydrated as a `User` would carry methods re
 query never fetched. See [Rows and entities](./orm-rows-and-entities.md) for the two opt-in levels
 above it — `track` + `save`, and `wrap`.
 
+### Importing a batch that may overlap
+
+```ts
+await Product.createMany({ data: rows, skipDuplicates: true })   // Postgres only
+```
+
+`ON CONFLICT DO NOTHING`, untargeted — so it covers every unique constraint and the primary key at
+once, which is what `skipDuplicates` means. The returned `{ count }` is the number of rows
+**inserted**, not the number supplied, so it answers "how many were new".
+
+The check and the insert being one statement is the whole point. Reading first to find out which
+rows exist is a second query *and* a race: between the read and the insert a concurrent importer
+writes one of them, and the insert fails on a unique violation anyway.
+
+**Postgres only, and that is Prisma's line rather than SQL's.** SQLite can express the same clause;
+Prisma rejects the argument there — for `false` as well as `true` — so gemi does too, rather than
+becoming a silent superset on the one dialect the differential harness could then no longer compare.
+On SQLite the error names the dialect and says so. If a batch is too large for one statement it is
+split inside a transaction, and `skipDuplicates` survives the split: the counts sum, and a conflict
+in a later chunk does not roll back an earlier one, because `do nothing` is not an error.
+
 ### Per-call options
 
 A second parameter, not a key inside `args` — intersecting Prisma's own arg types with a
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -185,6 +185,95 @@ describe("createMany", () => {
     ).toThrow(UnsupportedQueryError);
   });
 
+  /**
+   * `skipDuplicates` — the check and the insert in one statement, which is the
+   * whole point: reading first to find out which rows exist is a second query
+   * *and* a race, since a concurrent importer can write one of them in between
+   * and the insert fails on a unique violation anyway.
+   */
+  describe("skipDuplicates", () => {
+    const rows = { data: [{ email: "a@b.c" }, { email: "d@e.f" }] };
+
+    test("true emits an untargeted conflict clause, before returning", () => {
+      const emitted = text("createMany", { ...rows, skipDuplicates: true }, postgres);
+
+      // Untargeted, so it covers every unique constraint and the primary key at
+      // once — which is what `skipDuplicates` means, and what Prisma emits. A
+      // targeted `on conflict (col)` would still fail on any other constraint.
+      expect(emitted).toContain(` on conflict do nothing returning "id"`);
+    });
+
+    test("false is the same statement as not asking", () => {
+      expect(text("createMany", { ...rows, skipDuplicates: false }, postgres)).toBe(
+        text("createMany", rows, postgres),
+      );
+    });
+
+    /**
+     * The count has to be the number *inserted*, not the number supplied —
+     * the part the issue flags as most likely to be got wrong. It falls out:
+     * a row a conflict skipped never reaches `RETURNING`.
+     */
+    test("the count is what was inserted, not what was offered", () => {
+      const plan = compileWrite(
+        user,
+        "createMany",
+        { ...rows, skipDuplicates: true },
+        postgres,
+      );
+
+      expect(plan.shape([{ id: 1 }])).toEqual({ count: 1 });
+      expect(plan.shape([])).toEqual({ count: 0 });
+    });
+
+    /**
+     * SQLite *can* express it, and Prisma still rejects the argument there —
+     * for `false` as well as `true`, verified against a generated 6.19 client.
+     * Accepting `false` because it happens to be a no-op would make the two
+     * dialects disagree about which calls are legal.
+     */
+    test.each([true, false])("%s is refused on sqlite, naming the dialect", (value) => {
+      expect(() =>
+        text("createMany", { ...rows, skipDuplicates: value }),
+      ).toThrow(/not available on sqlite/);
+      expect(() =>
+        text("createMany", { ...rows, skipDuplicates: value }),
+      ).toThrow(/Prisma does not offer it there either/);
+    });
+
+    /**
+     * Validation must not depend on how much data happened to be supplied. The
+     * empty-list shortcut returns before the statement is built, so the check
+     * has to come first or `createMany({ data: [], skipDuplicates: true })`
+     * succeeds on SQLite and the same call with one row throws.
+     */
+    test("an empty list is validated too", () => {
+      expect(() =>
+        text("createMany", { data: [], skipDuplicates: true }),
+      ).toThrow(/not available on sqlite/);
+      expect(() =>
+        compileWrite(user, "createMany", { data: [], skipDuplicates: true }, postgres),
+      ).not.toThrow();
+    });
+
+    test("a non-boolean is refused", () => {
+      expect(() =>
+        compileWrite(user, "createMany", { ...rows, skipDuplicates: "yes" }, postgres),
+      ).toThrow(/Expected true or false/);
+    });
+
+    test("it is only an argument of createMany", () => {
+      expect(() =>
+        compileWrite(
+          user,
+          "create",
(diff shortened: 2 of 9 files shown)


nstfkc:
## Review — `createMany({ skipDuplicates })` (closes #69)

Read the full diff against the merge base (`31fbf29` — this branched before #72/#73 landed, so the diff against `feat/orm`'s tip is misleading; the real change is the 9 files the description lists). Unit suite 749 pass, template suite 413 pass on SQLite and 619 pass on Postgres 16, with only the by-design one-dialect-at-a-time guard failing.

**No issues found in the change itself.** Untargeted `on conflict do nothing` is the right emission and the reasoning for it is the part worth having in writing — a targeted `on conflict (col)` is the version that looks equivalent, passes a single-unique test, and then raises on the second constraint. Putting it behind `SqlDialect.ignoreConflicts(): Fragment | null` rather than a dialect-name check keeps the rule the compiler has everywhere else, and documenting `null` as a *parity* decision rather than a gap is what stops the next person returning `insert or ignore` for SQLite.

I exercised the edges beyond the committed cases, on Postgres:

| | result |
| --- | --- |
| 2 rows, 1 already committed | `{ count: 1 }` |
| **two identical rows inside one statement** | `{ count: 1 }` |
| `skipDuplicates: false` | `{ count: 1 }`, clause absent |
| **11 000 rows past the 65 535 ceiling, duplicate seeded at index 10 500** | `{ count: 10999 }` |
| row total afterwards | 11 004 — arithmetic checks out |

The last two are the ones I most wanted to see. Chunking and `skipDuplicates` were built independently: the argument has to ride onto every chunk, and a duplicate landing in a *later* chunk has to be skipped rather than raising — because by then the conflicting row was written by an earlier chunk of the same call. Both hold. The in-batch duplicate is worth having as a committed case too, since it is a conflict against a row the same statement is inserting rather than one already committed, and it is the shape a caller hits when their input list is not deduplicated.

`{ count }` being the number *inserted* falling out of `RETURNING` rather than needing a second count is the neatest part of the change.

### Worth pre-empting: this contradicts #75's refusal message

#75 (nested `createMany`, currently open) refuses the argument at the nested level with:

```
'skipDuplicates' is not implemented on 'createMany' at any level.
```

That sentence is true today and false the moment this merges. Whichever of the two lands second inherits the fix, and there are two decisions in it rather than one:

- The message has to stop saying "at any level", since `Account.createMany({ data, skipDuplicates: true })` will work on Postgres while `User.create({ data: { accounts: { createMany: { data, skipDuplicates: true } } } })` does not.
- Whether the nested form should simply *accept* it on Postgres. The nested step already calls the child's own `$exec("createMany", { data: items })`, so passing the flag through looks close to free — and #69's motivating shape (importing a batch that overlaps) is at least as natural one level down, under the parent it belongs to.

Not a defect in this PR; flagging it because the two are open simultaneously and the interaction is invisible from inside either.

### One thing this PR gets right that #75 does not

`skipDuplicatesClause` runs **before** the empty-list shortcut, with the reason stated: *"validation that depends on how much data happened to be supplied is the kind that passes in a test and fails in production."* That is exactly the property I raised against #75's nested step, where the empty-list short-circuit returns before the child's `$exec` and so skips the child's policy checks — the same call refusing or succeeding on array length. The rule is already written down here; #75 could adopt it verbatim.

Also worth noting for the merge order: this branch predates #72 and #73, so it will need a rebase, and `plan.ts`'s `LITERAL_KEYS` is now being edited by three separate branches (#74's aggregate kinds, this one's `skipDuplicates`). All three are additive to the same `Set`, so the conflicts should be trivial, but the third merge is the one to read carefully.

nstfkc:
Pushed as d9b8785 — the in-batch duplicate is a committed case now.

You are right that it is a different thing for `on conflict do nothing` to resolve: a conflict against a row the *same statement* is inserting, not one already committed. And it is the shape a caller actually hits, since an import list is rarely deduplicated before it is handed over. It goes through `expectSameWrite`, so Prisma's answer is the assertion rather than mine.

Thank you for exercising the two I most wanted covered and could not justify committing at that size — the 11 000-row split with a duplicate at index 10 500 is the one where chunking and `skipDuplicates` had to compose without ever having been designed together.

---

**On the contradiction with #75** — you called it before either of us hit it, and I have taken the half that is safe to take now.

`nested-writes.ts` said:

```
'skipDuplicates' is not implemented on 'createMany' at any level.
```

That sentence is false the moment this merges. It now names the level and points at the spelling that works:

```
'skipDuplicates' is not implemented on a *nested* 'createMany'. Write the
children with their own 'Account.createMany' call, which takes it.
```

True whichever of the two lands second, which is the property I wanted — a refusal that is wrong about the rest of the API is worse than a generic one, because it sends the reader looking for a gap that is not there. Pushed on #75 as a696106.

**The second decision I have left open deliberately.** Forwarding the flag through the nested step is close to free — it already calls the child's `$exec("createMany", …)` — but it needs this PR's support to exist first, and until then it fails with an error about an unknown argument on a model the caller never named. So it is a follow-up against whichever merges second, and I have recorded it in the code beside the refusal rather than in a comment thread that will scroll away.

**On `LITERAL_KEYS`** — noted, and agreed the third merge is the one to read. Three branches add to that `Set` (#74's five aggregate kinds, this one's `skipDuplicates`) and all three are additive, but they are also all *fixes for the same class of bug* — a value that reaches the SQL text while shaping to a bare type. Two of the three were found by probing for it rather than by a failing test, which suggests the check is worth making structural rather than remembering: something that walks the compiled text for a value that is not a placeholder. Not proposing it in this PR.

68 write-differential cases on SQLite and Postgres.

Issue #70, opened 2026-07-28
ORM: no `Json` column is exercised against a real database, and JSON path filters are unsupported


nstfkc:
Branch: `feat/orm`. Two related things, one a coverage hole and one a real gap.

## 1. No `Json` column is exercised against a real database

`Json` is implemented throughout — `packages/gemi/orm/schema.ts:36`, both dialects' encode and decode paths (`postgres.ts:151,183,198`, `sqlite.ts:190,205,235`), the lateral strategy's own decoder (`lateral.ts:299`), and the write path (`write.ts:1097`).

**`templates/saas-starter/prisma/schema.prisma` has no `Json` column** — nor `Bytes`, nor `Decimal`. So the differential harness, which is the only thing that compares gemi's decoding against Prisma's against a live database, never sees one. The coverage that exists is unit-level, against fixtures the same code wrote its expectations for.

This is precisely the shape of the bug that already shipped twice: `a8778ed` and `b2a3f3a` are two consecutive commits arriving at *opposite* conclusions about whether `Bytes` decodes to a `Buffer` or a `Uint8Array`, with the note that "my own fixture could not see that". A fixture asserting its own assumptions agrees with them.

Under the lateral strategy the risk is higher again, not lower: `json_agg` flattens types on the way out, so a `Json` column nested inside a lateral include goes through `JSON.parse` twice and has to come back identical to the batched path and to Prisma. Nothing checks that today.

**Fix:** add `Json`, `Bytes` and `Decimal` columns to the template schema. The existing coercion suites and the differential matrix then cover them with no new test code — including null, empty object, empty array, nested, and a JSON string that looks like a number.

For a production comparison point: the reference application carries 23 `Json` columns across 79 models. This is not an edge type.

## 2. `where` filters on JSON paths are refused

```
gemi ORM does not support 'where.metadata.path' yet (Product.findMany).
```

`packages/gemi/orm/compile/where.ts:466-479` lists the twelve scalar operators; Prisma's JSON filters — `path`, `string_contains`, `string_starts_with`, `string_ends_with`, `array_contains`, and `equals` against a path — are none of them.

Postgres expresses all of these (`#>`, `@>`, `jsonb_path_query`); SQLite has `json_extract` and covers most. The reference application filters on JSON paths and also mutates JSON in place with `jsonb_set`, which is a *write*-side gap — `data: { metadata: { set: … } }` replaces the whole document today, and there is no path-scoped update.

[...]

PR #77, opened 2026-07-28
`Json` against a real database, and the two bugs that were hiding each other

nstfkc:
Closes the **first half** of #70 — the coverage hole. The second half, JSON path filters in `where`, is a feature with its own dialect surface and is not here.

The issue's claim was that `Json` is implemented everywhere and compared nowhere: the template's schema has no `Json` column, so the differential harness — the only thing that checks gemi's decoding against Prisma's against a live database — had never seen one. Every test that "covered" it was written against the same assumption as the code.

Adding `Json` and `Bytes` columns to `User` and to `Account` — root, and one relation away — turned **20 differential cases red on Postgres immediately**.

### Read

`PostgresDialect.decode` re-parsed anything that was a string, on the reasoning that jsonb "arrives as text" and that a `typeof` check would cope either way. Measured through Bun 1.3.14 against Postgres 16 — Bun already parses `json` and `jsonb`:

| stored | Bun returns |
| --- | --- |
| `'{"a":1}'` | object |
| `'"42"'` | string `"42"` |
| `'42'` | number `42` |
| `'null'` | `null` |

And the check *cannot* cope, which is the part worth keeping: **a JS string is ambiguous** between "raw JSON text nobody parsed" and "a JSON string value somebody did". So a column holding the JSON string `"42"` came back as the number 42, `"true"` as a boolean, and `'{"a":1}'` as an object. `"text"` survived only because `JSON.parse` threw and the catch handed it back.

`compile/lateral.ts` carried the same guard, and there it is worse: a folded include puts the value through JSON **twice**. The same query returned `"42"` under the batched strategy and `42` under lateral — two answers, decided by which strategy the dialect happened to pick. That one was caught by seeding a JSON string *one relation down*, which nothing had reached before.

### Write

`PostgresDialect.encode` pre-serialized, and Bun then stored that string **as a JSON string**: `{a:1}` was written as `"{\"a\":1}"`. Removing the read bug is what exposed it — the two had been cancelling, and the round trip looked right as long as nothing else read the column.

SQLite had the same ambiguity on the write side, from the same `typeof` guard: `"42"` was stored as the bare text `42` and read back as the number 42.

The fixes are symmetric and **remove the guessing rather than adjusting it** — Postgres hands `Json` to the driver raw in both directions, SQLite serializes unconditionally.

## One boundary, pinned rather than hidden

A bare JSON *number* or *boolean* still fails on Postgres: Bun binds it as its own type and the server answers `is of type jsonb but expression is of type integer`. It **raises** rather than storing something wrong, and it is narrower than Prisma only on that dialect.

Fixing it means always serialising *and* emitting `::jsonb` on the parameter — which has to reach the insert, the update's set clause, and any `where` comparing a Json column. That is a compiler change with its own differential pass, not a dialect tweak, so I deliberately did not fold it into the change that found it. It is pinned by a test asserting the raise on Postgres and the round trip on SQLite, so a future fix has to update it on purpose.

## On the issue's third column

**`Decimal` cannot be added.** `bin/orm/emit.ts`'s `UNSUPPORTED_SCALARS` refuses a Decimal field at *generation* time, with the precision reasoning recorded there — a Decimal column would break `prisma generate` rather than extend coverage. So the issue's "add `Json`, `Bytes` and `Decimal`" is only two-thirds achievable today; supporting Decimal is its own decision and its own change. (This is the same conclusion #74's review reached from the other direction, about `aggregate`'s unreachable `Decimal` branch.)

## Coverage

The seeded values discriminate rather than merely exist: a nested object, an empty object, an empty array, a JSON null, a JSON string that looks like a number, and bytes carrying both a zero and a high byte. The whole existing matrix now carries them — root and nested, batched and lateral, both dialects — **with no new differential cases**, which is what the issue predicted would happen once the columns existed. Seven round-trip cases cover write → read → write.

839 unit tests, and the full template suite green on SQLite and Postgres 16. Pre-existing and unchanged: `generated.test.ts`'s generator-linkage probe.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -761,6 +761,28 @@ is the same for those, and it is about where the value came from, not what it is
 that needs none of this. It does **not** join the ambient transaction — use `DB.query` when that
 matters.
 
+## `Json` columns
+
+A `Json` column round-trips whatever you give it: an object, an array, a string, a number, `null`.
+The value is stored as JSON and comes back as the same JavaScript value, on both dialects and
+whether it is read at the root or nested inside an `include`.
+
+Two things are worth knowing:
+
+- **A bare JSON number or boolean is refused on Postgres.** `metadata: 42` raises rather than being
+  stored, because the driver binds it as an integer and the column is `jsonb`. Wrap it — `{ value:
+  42 }` — or store it as a string. SQLite has no such limit. This is the one shape where the two
+  dialects disagree, and it fails loudly rather than storing the wrong thing.
+- **A string is a string.** `metadata: "42"` stores the JSON string `"42"`, not the number, and
+  `metadata: '{"a":1}'` stores that text as a string rather than as an object. If you want an
+  object, pass an object.
+
+> **Upgrading:** `Json` values written by a *pre-release* build of this ORM on Postgres were stored
+> as JSON **strings** rather than as objects — `{ a: 1 }` landed as `"{\"a\":1}"`. Reads undid it,
+> so nothing looked wrong from inside the ORM, but the column was wrong for anything else that read
+> it. Those rows now read back as strings. Re-seed development databases; there is no released
+> version affected.
+
 ## Dialects
 
 **SQLite and Postgres** are built and tested, on every operation, against a differential harness
--- a/packages/gemi/orm/compile/lateral.ts
+++ b/packages/gemi/orm/compile/lateral.ts
@@ -662,22 +662,24 @@ function jsonConverter(
         value === null || value === undefined ? null : BigInt(String(value));
 
     case "Json":
-      // A `jsonb` column embedded in `json_build_object` should arrive as a
-      // nested object, and does — unless the value was stored as a JSON *string*,
-      // which is what `PostgresDialect.encode` produces and what the column then
-      // holds if it is text-typed rather than `jsonb`. The batched path handles
-      // exactly this with the same `typeof` check in `PostgresDialect.decode`, so
-      // the two paths agree either way rather than only when the column type is
-      // what one of them assumed.
-      return (value) => {
-        if (value === null || value === undefined) return null;
-        if (typeof value !== "string") return value;
-        try {
-          return JSON.parse(value);
-        } catch {
-          return value;
-        }
-      };
+      // Nothing to do, for the same reason `PostgresDialect.decode` does
+      // nothing: the value arrives already parsed. A `jsonb` column embedded in
+      // `json_build_object` is a nested JSON value, and Bun parses the whole
+      // `data` payload — so by the time it reaches here it is an object, an
+      // array, a string, a number or null, exactly as stored.
+      //
+      // This used to re-parse anything that was a string, on the reasoning that
+      // it might be raw JSON text. **A JS string is ambiguous** between "text
+      // nobody parsed" and "a JSON string value somebody did", and here the
+      // ambiguity is worse than on the batched path, because the value has been
+      // through JSON *twice* — once for the column, once for `json_agg`. A
+      // column holding the JSON string `"42"` came back as the number 42 from a
+      // folded include and as `"42"` from a batched one: the same query, two
+      // answers, decided by which strategy the dialect happened to pick.
+      //
+      // Caught by seeding a JSON string one relation down. Nothing reached it
+      // before, because the template's schema had no `Json` column at all.
+      return undefined;
 
     case "Bytes":
       // Postgres renders `bytea` into JSON as `\x` followed by hex.
--- a/packages/gemi/orm/dialect/postgres.test.ts
+++ b/packages/gemi/orm/dialect/postgres.test.ts
@@ -199,13 +199,19 @@ describe("decoding the types the template schema cannot reach", () => {
   });
 
   /**
-   * `needsDecode` stays true even though `decode` is now a pass-through for
-   * `Json`: it is asked per *field type*, and the shaper skips the call
-   * entirely when it is false. Flipping it would be a micro-optimisation that
-   * silently stops `null` becoming `null`.
+   * `needsDecode` is "false when the driver already returns exactly what Prisma
+   * would", and that is now true of `Json` here — so the shaper skips the call
+   * rather than paying for one per value to be handed back unchanged.
+   *
+   * Safe because a NULL column arrives as `null` from the driver, which is
+   * already what Prisma returns; the same reason `Int` and `DateTime` are not
+   * in this set either.
    */
-  test("Json still reports that it needs decoding", () => {
-    expect(postgres.needsDecode(field("Json") as any)).toBe(true);
+  test("Json needs no decoding, since the driver already parsed it", () => {
+    expect(postgres.needsDecode(field("Json") as any)).toBe(false);
+    // The two that genuinely do: `bigint` crosses as text, `bytea` as a Buffer.
+    expect(postgres.needsDecode(field("BigInt") as any)).toBe(true);
+    expect(postgres.needsDecode(field("Bytes") as any)).toBe(true);
   });
 
   test.each(["Int", "String", "Boolean", "Float", "DateTime"])(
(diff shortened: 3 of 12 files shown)


nstfkc:
## Review — `Json` / `Bytes` coverage, and the two bugs it found (first half of #70)

Read the full diff against the merge base (`9b20919`). Unit suite 839 pass, template suite 435 pass on SQLite and 683 on Postgres 16, only the by-design one-dialect-at-a-time guard failing.

**Both bugs are real and the fix is correct — I verified by reverting rather than by reading.** Keeping this branch's new `Json` columns and restoring only the three dialect files to their base-branch versions reproduces exactly what the description claims:

```
                            base dialects        this branch
string "42"          ->     42      ✗            "42"     ✓
string "true"        ->     true    ✗            "true"   ✓
string '{"a":1}'     ->     {a:1}   ✗            '{"a":1}' ✓
object, array, {}, null ->  correct              correct
through a relation, batched -> 42   ✗            "42"     ✓
through a relation, lateral -> 42   ✗            "42"     ✓
```

Both dialects, and the relation rows are the ones worth dwelling on: under the old code the folded and batched paths agreed *because both were wrong the same way*, which is why "the same query returns two answers" was only half the story. The diagnosis — that a JS string is irreducibly ambiguous between "raw JSON text" and "a JSON string value", so the `typeof` guard cannot be adjusted, only removed — is right, and it is the reason the fix is symmetric on both sides rather than a patch on one.

Worth saying plainly since the description is modest about it: this is a data-corruption bug in a shipped-looking feature, found only because the coverage hole was closed. That is the whole argument of #70 discharged in one PR.

### The bare-scalar boundary is a regression, not just a gap

The description frames it as a boundary that "raises rather than storing something wrong", and pins it. Both true. But it is narrower than that: on the base branch, `metadata: 42` and `metadata: true` **round-tripped correctly on Postgres** — the old encoder stringified every non-string, so `42` reached the driver as text, Postgres cast it into `jsonb`, and the old decoder handed back the number. Same probe, same rows:

```
postgres  bare number   base: OK 42     this branch: THREW  column "metadata" is of type jsonb …
postgres  bare boolean  base: OK true   this branch: THREW
```

So the trade is not "gains correctness, keeps a known gap" — it is "gains correctness for JSON strings, loses bare JSON scalars on Postgres". That is still the right trade, and the reasoning for deferring `::jsonb` is sound (it has to reach the insert, the update's set clause, and any `where` on a Json column, and it wants its own differential pass). I would just say so in the description and in the test's comment, because "this used to work" is what a bisect will conclude and it should find an answer already written down.

If it turns out to be small, doing it here rather than next has an argument: the pinned test currently encodes a *regression* as expected behaviour, and those are the tests that get read as intent later.

### Data written by the old encoder does not read back

Follows from the encode fix, and I confirmed the storage shape rather than inferring it. Writing `{ a: 1 }` through the base-branch encoder against Postgres produces:

```sql
select jsonb_typeof("metadata"), "metadata"::text from "User"
-- string    "{\"a\":1}"
```

A jsonb *string* containing JSON text, not an object. The old decoder re-parsed it, so the round trip looked right; the new one returns it verbatim, so that row now reads as the JS string `{"a":1}`. Every `Json` value in any Postgres database written by this branch's earlier code is in that state.

On an unreleased branch the answer is "re-seed your development database", which is fine — but it is silent (wrong type, no error), so it belongs in the PR description and probably in `docs/orm.md` beside the `Json` section. SQLite is unaffected for objects; only values that were legitimately JSON strings differ there, and those were already being stored wrong.

### Smaller

- `PostgresDialect.needsDecode` still returns `true` for `Json`, but `decode`'s `Json` case is now `return value` — so every Json value on the hot path takes a function call to be handed back unchanged. `needsDecode` documents itself as "false when the driver already returns exactly what Prisma would", which is now exactly true of `Json` here. Dropping it from that predicate looks like a one-word change with a measurable-by-construction win on wide reads, and it keeps the predicate honest.
- `encode(value, _field)` no longer reads its second parameter on this dialect. Fine as it is — the interface requires it — but the comment says "which is why this takes no `field` at all any more", which reads as though the signature changed when only the usage did.
- The note that `Decimal` cannot be added — because `emit.ts` refuses it at generation time — is the same conclusion #74's review reached from the other direction. Two independent paths arriving at "Decimal is unconstructible, and supporting it is its own change" is worth one line in #70 so the issue stops asking for three columns.

Nothing blocking. The reverting exercise is the part I would keep: it is the only thing that distinguishes "these two bugs cancelled" from "there was never a bug", and the description's claim survives it.

nstfkc:
Pushed as e85b3de. All four points are in, and the first one turned out differently once I measured it.

## The bare-scalar boundary is not the regression it looks like

You are right that `metadata: 42` appeared to work before and raises now. But it was not *stored* correctly — I checked with `jsonb_typeof` against the old encoder rather than trusting the round trip:

```
in 42       -> stored kind=string   raw="42"
in true     -> stored kind=string   raw="true"
in {a:1}    -> stored kind=string   raw="{\"a\":1}"
```

A bare number was stored as the jsonb **string** `"42"` — the same mis-store as the object case, and for the same reason. It only looked correct because the old decoder re-parsed on the way out. The column was wrong in the database the whole time, and anything reading it other than this ORM — Prisma, `psql`, a report — saw a string.

So the trade is not "gains JSON strings, loses bare scalars". It is **a silent mis-store replaced by a loud failure**, which is the same trade as everywhere else in this PR and needs no separate justification. That also removes your reason for doing the `::jsonb` work here: the pinned test is not encoding a regression as intent, it is pinning a shape that never worked and now says so.

Recorded in both places a bisect will land — the test comment and the dialect comment — because "this used to work" is what the commit range says and it should find the answer already written down.

## The other three

- **`needsDecode`** returns `false` for `Json` on Postgres now. You are right that the predicate documents itself as "false when the driver already returns exactly what Prisma would", and that is exactly true once `decode` is a pass-through. Safe for the same reason `Int` and `DateTime` are already absent: a NULL column arrives as `null`, which is what Prisma returns.
- **The `encode` comment** said the signature had changed when only the usage had. Reworded.
- **The stale-data hazard** is in `docs/orm.md` now, not just the description — it is silent, so a commit message is the wrong place for it. `docs/orm.md` also gains the `Json` contract it did not have: what round-trips, that a string stays a string, and the Postgres bare-scalar refusal.

## On `Decimal`

Agreed, and worth doing: two independent paths — #74's review from the `aggregate` side, this one from the schema side — arrived at "unconstructible, and supporting it is its own change". I have not edited #70 from here since it is your issue, but it should stop asking for three columns.

---

The reverting exercise is the part I would keep too. I had verified the fix forwards; you verified it backwards, and those are different claims — "these two bugs cancelled" only survives the second one.

839 unit tests, full template suite green on both dialects.

nstfkc:
### Re-reviewed `e85b3de` — and my "regression" framing was wrong

You are right and I was not. I measured only the object case and generalised from it; you measured the shape I had asserted about. Running it per value against the pre-fix encoder:

```
number 42     -> jsonb_typeof=string   text="42"
boolean true  -> jsonb_typeof=string   text="true"
object        -> jsonb_typeof=string   text="{\"a\":1}"
array         -> jsonb_typeof=string   text="[1,2]"
```

Every shape, including the bare scalars. So `metadata: 42` never round-tripped correctly on Postgres — it round-tripped *through this ORM*, because the decoder undid the encoder, and the column held the string `"42"` for anything else that read it. "Loses bare JSON scalars" was the wrong description: what is lost is a silent mis-store, replaced by a loud failure. The new comment says that better than my finding did, and pinning it with `jsonb_typeof` rather than through the ORM's own read path is the right way to have settled it — a round-trip assertion could not have distinguished the two.

The upgrading note in `docs/orm.md` is the right home for the second point, and it is more precise than I was: naming what the value looked like *from outside the ORM* is the part that tells a reader whether their data is affected.

`needsDecode` dropping `Json` is correct — `decode`'s `Json` case is now `return value`, so the predicate's own definition applies. 839 unit tests pass.

PR #78, opened 2026-07-28
`omit`, and the end of "yet" for `distinct` and `cursor`

### LINKED ISSUE #68: ORM: decide `omit`, `distinct` and `cursor` — implement, or make the refusal permanent

nstfkc:
Closes #68, which asks for **three decisions** rather than three features. Each was made against a measurement of what Prisma actually does, not against its documentation.

## `omit` — implemented

7 call sites, all the same shape: every column except the password hash.

```ts
const user = await User.findUnique({ where: { id }, omit: { password: true } })
```

It is a **real projection** in Prisma — read off its query log, the column is simply absent from the `SELECT` list rather than filtered afterwards — and it is the complement of `select` for a reason worth writing down: with `select` the exclusion list is rewritten every time the model gains a column, and the day somebody forgets is the day the new column silently stops being returned. `omit` says the thing that is actually stable about the query.

Implemented in `resolveSelection`, which reads, writes and relation nodes all share — so a write's `RETURNING` list narrows exactly as a read's column list does, with no second implementation to drift.

Behaviours reproduced from the probe rather than assumed: `omit: { x: false }` keeps the column; `select` + `omit` together is refused (Prisma raises); omitting *every* field is refused.

## `distinct` — refused permanently

The issue suspected Prisma's version was in-memory. It is — the query log for `findMany({ distinct: ["name"], take: 1 })` is:

```sql
SELECT `User`.`id`, `User`.`name` FROM `User` WHERE 1=1 ORDER BY `User`.`id` ASC LIMIT ? OFFSET ?
```

No `DISTINCT` anywhere, and `take` did not reduce the rows pulled. Reproducing that faithfully would mean reading the whole result set and deduplicating in JavaScript behind an argument that reads like a database operation. Emitting a real `DISTINCT ON` instead would silently diverge from Prisma, and the differential harness would be right to call it a divergence.

Neither is worth doing now that composed SQL has somewhere to go (#73, merged).

## `cursor` — refused permanently

Only correct under a *total* ordering, which Prisma does not enforce — under a non-unique `orderBy` it skips or repeats rows at the page boundary. The real demand behind it is keyset pagination, which `sql` + `DB.query` express directly and honestly.

## The criterion that made this more than three edits

> The refusal and `docs/orm.md` agree — no argument says "yet" while the docs say "not in scope".

They *could not* agree, because every refusal used one error whose message says "yet". So the docs listed `omit` under **Not in scope** while the runtime promised it was coming — a contradiction a reader has no way to resolve.

`UnsupportedByDesignError` says **"and this is a decision rather than a gap"**, and each of the two carries what to reach for instead (#61's rule). It **subclasses** `UnsupportedQueryError`, so `catch (e instanceof UnsupportedQueryError)` keeps working for callers that do not care which kind it is. An argument that genuinely is unimplemented still says "yet", and there is a test holding the two apart — otherwise the next refusal added would quietly inherit the wrong word.

## Verification

13 differential cases for `omit` against Prisma on both dialects: one column, several, `false`, mixed true/false, an empty object, a nullable column, the primary key, beside a `where`, beside `orderBy` + `take`, on `findFirst` and `findUnique`, beside an `include`, and — the one I most wanted — **omitting the very key the include stitches on**, which the planner has to fetch anyway and then hide again.

841 unit tests, and the full template suite green on SQLite and Postgres 16. Pre-existing and unchanged: `generated.test.ts`'s generator-linkage probe.

Branched after #72 and #73 merged, so it needs no rebase.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -141,7 +141,9 @@ later is included by default.
 It is a real projection: the omitted column never enters the `SELECT` list, so it is not read, not
 shaped and not decoded. `select` and `omit` together are refused, because one names what to keep and
 the other what to drop; Prisma rejects that pair too. It works on every operation that returns a
-payload, reads and writes alike.
+payload, reads and writes alike, and **inside an `include`** — which is where it usually matters,
+since a column you never want to leave the process is as likely to be on a related row as on the
+root one. Both relation strategies honour it.
 
 **It is not a substitute for a policy's `redact`.** `omit` is the caller choosing; `redact` is the
 model refusing, and a caller cannot opt out of it. Use `redact` for "nobody may read this", and
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -126,6 +126,27 @@ stepping stone: a `Pick<User, "id">` hydrated as a `User` would carry methods re
 query never fetched. See [Rows and entities](./orm-rows-and-entities.md) for the two opt-in levels
 above it — `track` + `save`, and `wrap`.
 
+### Returning everything except one column
+
+```ts
+const user = await User.findUnique({ where: { id }, omit: { password: true } })
+```
+
+`omit` is the complement of `select`, and the reason to have both is what happens when the model
+gains a column. With `select` the exclusion list has to be rewritten every time, and the day
+somebody forgets is the day the new column silently stops being returned. `omit` says the thing that
+is actually stable about the query — *this column must never leave the process* — so a column added
+later is included by default.
+
+It is a real projection: the omitted column never enters the `SELECT` list, so it is not read, not
+shaped and not decoded. `select` and `omit` together are refused, because one names what to keep and
+the other what to drop; Prisma rejects that pair too. It works on every operation that returns a
+payload, reads and writes alike.
+
+**It is not a substitute for a policy's `redact`.** `omit` is the caller choosing; `redact` is the
+model refusing, and a caller cannot opt out of it. Use `redact` for "nobody may read this", and
+`omit` for "I do not need this here".
+
 ### Per-call options
 
 A second parameter, not a key inside `args` — intersecting Prisma's own arg types with a
@@ -778,10 +799,21 @@ Stated so you can plan around them rather than discover them:
   an identity map is worse than none.
 - **No lazy loading.** A relation you did not `include` is absent, not a proxy that queries when
   touched.
-- **No `omit`.** Use `select`, or a policy's `redact`.
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
-- **No `groupBy`, `aggregate`, `distinct` or cursor pagination.** These land in
-  [Raw SQL](#raw-sql), which exists so that "not implemented" has an answer rather than a shrug.
+- **No `groupBy` or `aggregate`.** These land in [Raw SQL](#raw-sql), which exists so that "not
+  implemented" has an answer rather than a shrug.
+- **No `distinct`, and this one is deliberate rather than pending.** Prisma applies it **in
+  memory** — its query log shows no `DISTINCT` at all, so `take` neither reduces the rows pulled
+  nor paginates by group. Reproducing that faithfully would mean reading the whole result set and
+  deduplicating in JavaScript behind an argument that reads like a database operation; emitting a
+  real `DISTINCT ON` instead would silently diverge from Prisma. Write it as SQL.
+- **No `cursor`, also deliberate.** It is only correct under a *total* ordering, which Prisma does
+  not enforce — under a non-unique `orderBy` it skips or repeats rows at the page boundary. Use
+  `take` with a `where` on the last row's sort key, or compose the keyset comparison with `sql`.
+
+Both of the last two throw `UnsupportedByDesignError`, which says *"and this is a decision rather
+than a gap"* rather than *"yet"* — so a refusal you can plan around reads differently from one that
+might lift next release.
 
 ## See also
 
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -48,11 +48,12 @@ const RELATION_ARGS = new Set([
   "where",
   "orderBy",
   "select",
+  "omit",
   "include",
   "take",
   "skip",
 ]);
-const TO_ONE_ARGS = new Set(["where", "select", "include"]);
+const TO_ONE_ARGS = new Set(["where", "select", "omit", "include"]);
 
 /**
  * How deep an include tree may nest. Not a modelling limit — a legal tree can
@@ -1162,6 +1163,25 @@ function childQuery(
 
   if (relationArgs.orderBy !== undefined) args.orderBy = relationArgs.orderBy;
   if (relationArgs.include !== undefined) args.include = relationArgs.include;
+
+  // `omit` gets the same treatment as `select`, and it has to: the lateral
+  // strategy resolves a node's selection directly, so a node's `omit` already
+  // narrowed its columns there. Not forwarding it here would drop the omission
+  // on the batched path only — the same query returning the column under one
+  // strategy and not the other, decided by the dialect.
+  const omit = relationArgs.omit as Record<string, unknown> | undefined;
+  if (omit !== undefined) {
+    if (omit[childField] !== true) {
+      args.omit = omit;
+    } else {
+      // The stitch key cannot be omitted from the *query* — nothing could
+      // match the children to their parents without it. So it is fetched and
+      // then deleted, exactly as a `select` that leaves it out is.
+      const { [childField]: _dropped, ...rest } = omit;
+      args.omit = rest;
+      return { args, hidden: true };
+    }
+  }
 
   const select = relationArgs.select as Record<string, unknown> | undefined;
   if (select === undefined) return { args, hidden: false };
--- a/packages/gemi/orm/compile/read.test.ts
+++ b/packages/gemi/orm/compile/read.test.ts
@@ -5,6 +5,7 @@ import { SqliteDialect } from "../dialect/sqlite";
 import {
   ParameterLimitError,
   UnknownFieldError,
+  UnsupportedByDesignError,
   UnsupportedQueryError,
 } from "../errors";
 import { USER_COLUMNS, account, mapped, user } from "../fixtures";
@@ -481,13 +482,43 @@ describe("postgres", () => {
   });
 });
 
-describe("unimplemented arguments still throw", () => {
+/**
+ * `cursor` and `distinct` are **decisions, not gaps**, and the error says so.
+ *
+ * The distinction is one a caller can act on: "yet" means wait for a release,
+ * "will not" means change the code. They shared one error until #68, which is
+ * how `docs/orm.md` came to list an argument under *Not in scope* while the
+ * runtime said "yet".
+ */
+describe("arguments refused by design", () => {
   test.each([
     ["cursor", { cursor: { id: 1 } }],
     ["distinct", { distinct: ["email"] }],
-  ])("%s", (argument, args) => {
+  ])("%s says it is a decision, not a gap", (argument, args) => {
+    expect(() => text(args)).toThrow(UnsupportedByDesignError);
     expect(() => text(args)).toThrow(
-      `gemi ORM does not support '${argument}' yet (User.findMany).`,
+      `gemi ORM does not implement '${argument}' (User.findMany), and this is ` +
+        `a decision rather than a gap.`,
+    );
+    // Still an `UnsupportedQueryError`, so a caller that does not care which
+    // kind it is keeps working.
+    expect(() => text(args)).toThrow(UnsupportedQueryError);
+  });
+
+  /** Each names what to reach for instead — #61's rule. */
+  test("distinct explains why Prisma's own version is the problem", () => {
+    expect(() => text({ distinct: ["email"] })).toThrow(/in memory/);
+    expect(() => text({ distinct: ["email"] })).toThrow(/DB\.query/);
+  });
+
+  test("cursor names the failure it would have", () => {
+    expect(() => text({ cursor: { id: 1 } })).toThrow(/total.* ordering/);
+    expect(() => text({ cursor: { id: 1 } })).toThrow(/skips or repeats rows/);
+  });
+
+  test("an argument that really is unimplemented still says 'yet'", () => {
+    expect(() => text({ nonsense: 1 } as never)).toThrow(
+      `gemi ORM does not support 'nonsense' yet (User.findMany).`,
     );
   });
 
--- a/packages/gemi/orm/compile/read.ts
+++ b/packages/gemi/orm/compile/read.ts
@@ -24,7 +24,7 @@ import { assertUniqueWhere } from "./unique";
 import { compileWhere } from "./where";
 
 /** Every read operation, and what each accepts. */
-const READ_ARGS: Record<string, Set<string>> = {
+export const READ_ARGS: Record<string, Set<string>> = {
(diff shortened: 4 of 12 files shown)


nstfkc:
## Review — `omit`, and the `distinct` / `cursor` decisions (closes #68)

Read the full diff against the merge base (`9b20919`). 841 unit tests pass.

The three-decisions framing is right, and the two refusals are argued from measurements rather than from intuition — the `distinct` query log showing no `DISTINCT` and an unreduced `take` is the kind of evidence that settles a question permanently. `UnsupportedByDesignError` subclassing `UnsupportedQueryError` so existing `catch` sites keep working, with a test holding "by design" and "yet" apart, is the right shape: the failure it prevents is the next refusal quietly inheriting the wrong word. Implementing `omit` inside `resolveSelection` so reads, writes and the `RETURNING` list narrow through one path is the reason this is +325 rather than +900.

**One blocking issue, and it is the third instance of a class this codebase has now hit three times.**

### `omit` is missing from `LITERAL_KEYS`, and it fails toward exposure

`omit` decides the projection — it is the complement of `select`, which is in that set for exactly this reason. It is not, and `plan.ts` is not in the diff:

```
key(omit: { password: true  }) = sqlite:batched:User:findMany:{omit:{password:boolean}}
key(omit: { password: false }) = sqlite:batched:User:findMany:{omit:{password:boolean}}
same key: true
```

Both shape to `boolean`, so whichever compiles first decides for the other. Against a real database, two requests in one process, in the order an application would actually make them:

```ts
// an internal screen that explicitly keeps the column
await User.findMany({ omit: { password: false } })   // -> password = "hunter2"
// the public profile endpoint, asking for it to be hidden
await User.findMany({ omit: { password: true } })    // -> password = "hunter2"   ✗
```

The second call gets the first's statement and the password comes back. The reverse order fails harmlessly (the column is missing where the caller wanted it), so the asymmetry is the wrong way round: the dangerous direction is the one where the *permissive* call warms the cache, which is also the more likely one in a process that serves admin and public traffic from the same models.

This is the same bug as #74's aggregate kinds and #76's `skipDuplicates`, and it lands on the feature whose motivating example in this very PR is "every column except the password hash".

**Fix and verification:** adding `"omit"` to `LITERAL_KEYS` beside `"select"` is sufficient — with it the second call returns `password: undefined`, and all 841 unit tests still pass. Worth a case in `plan.discrimination.test.ts` alongside the ones #74 and #76 added; three occurrences suggest the discrimination suite wants a rule rather than a row — something that enumerates every key a compiler branches on and asserts each is either in `LITERAL_KEYS` or provably not projection-affecting, so the fourth one fails at test time instead of at review time.

### `omit` is not accepted on a relation node

Prisma allows it there, and this refuses it:

```
include: { accounts: { omit: { userId: true } } }
  -> gemi ORM does not support 'accounts.omit' yet (User.findMany).
```

`RELATION_ARGS` gained `take` and `skip` in #72 but not `omit`. Loud rather than silent, so it is a scope question rather than a defect — but it is worth deciding explicitly in this PR rather than by omission, because the nested case is where the motivating shape actually bites: a password column on an *included* relation is exactly the read that `redact` had to exist for, and `omit` would be the cheaper answer. The message also says "yet", which is now the word this PR is careful about elsewhere.

If it is deferred, the docs section deserves a line saying `omit` is root-only today, since a reader who has just been told it is the complement of `select` will reasonably expect it wherever `select` goes.

### Smaller

- `omit` interacting with a hidden stitch key is the case I would have gone looking for, and it is already in the differential list — omitting the very column the include joins on, which the planner must fetch and then hide. Good.
- The docs' **Not in scope** list and the runtime messages now agree, which was the criterion #68 set and the one that is easy to declare done without checking. I did check: `distinct` and `cursor` raise `UnsupportedByDesignError` and neither says "yet".

Nothing else. With the `LITERAL_KEYS` line this is ready.

nstfkc:
Pushed as 2d041ab. The finding was right, the direction is the bad one, and the meta-point about three occurrences turned into the more valuable half of the change.

## The leak

Reproduced through the public API before touching anything, in the order an application actually makes the calls:

```
User.findMany({ omit: { password: false } })   -> password selected
User.findMany({ omit: { password: true } })    -> password selected   ✗
```

`"omit"` now sits in `LITERAL_KEYS` beside `"select"`, whose complement it is.

## The rule, and a way it can be written wrong

`plan.coverage.test.ts` walks **every argument every operation accepts**, compiles two calls differing only in that argument's literal content, and asserts the plan key separates them whenever the statements differ. A coverage check fails for any argument added to `READ_ARGS` or `WRITE_ARGS` without a probe — so the fourth instance is a failing test *when it is written*.

Worth reporting: my first version of it **passed against a knowingly-broken `LITERAL_KEYS`**. Compiled through the shared cache, the test agrees with the bug — when the two keys collide the second `getOrCompile` is a cache *hit*, so both plans are the same object, the texts match, and the `if (textA === textB) continue` skips the case it exists to catch. Clearing the cache between the two compiles is what makes "the statements differ" a fact about the compiler rather than about what the cache returned.

I only found that because I checked the test failed with `omit` removed rather than trusting it passed with `omit` present. With the fix it fails eight operations, each naming `LITERAL_KEYS` in the message.

There is a converse case too, so the rule cannot be satisfied by making everything structural: `where`, `take` and `skip` must still *share* a plan.

## Nested `omit` — implemented, and the audit found something worse

You were right that it is where the shape bites. It also turned out that refusing it was hiding a divergence rather than a gap: `resolveSelection` already handled `omit`, so allowing it in `RELATION_ARGS` made the **lateral** strategy honour it immediately — while the **batched** path would have dropped it silently, because `childQuery` forwards only `where`, `orderBy`, `include` and `select`.

That is the same query returning the column under one strategy and not the other, decided by the dialect. So the choice was not "implement or defer" but "implement or ship a strategy-dependent leak the moment somebody adds it". `omit` now forwards to the child query, and omitting the *stitch key* fetches and hides it exactly as a `select` that leaves it out does.

Four differential cases on both dialects and both strategies, including `omit` at both levels of one query. The docs line you asked for is there too — it says `omit` works inside an `include`, rather than that it does not.

856 unit tests, full template suite green on both dialects.

nstfkc:
### Re-reviewed `2d041ab` — fixed, and the guard works

The leak is closed. My original two-request probe, unchanged:

```
omit: { password: false }  first  ->  password = "hunter2"
omit: { password: true }   after  ->  password = undefined     ✓
```

**`plan.coverage.test.ts` is the right answer to "three times is a rule", and I checked that it actually fires rather than merely existing.** Removing `"omit"` from `LITERAL_KEYS` again fails it in 8 of 15 cases, with the message naming the argument and where to fix it:

```
delete.omit compiles to two different statements but shares one plan key
  — add it to LITERAL_KEYS in plan.ts
```

Walking `READ_ARGS` / `WRITE_ARGS` so an argument added without a probe fails the suite is what makes it a mechanism rather than three more rows — the fourth instance is now a failing test at the time it is written. That is worth more than this PR's own feature.

**`omit` on a relation node**, which you took further than I flagged: I asked whether it should be accepted, and you also caught that the lateral strategy resolves a node's selection directly, so forwarding it to the batched child query was needed or the two paths would disagree by dialect. Verified against Postgres, both strategies, including the case that has to be fetched and then deleted:

```
omit an ordinary column  [batched] userId present, organizationRole absent
                         [lateral] userId present, organizationRole absent
omit the stitch key      [batched] userId absent, children still attached
                         [lateral] userId absent, children still attached
```

Identical key sets under both, and the children still stitch — which is the whole difficulty of omitting the join key.

856 unit tests pass. Nothing outstanding from me on this PR.

nstfkc:
### Post-merge check on `5d893df` — the merge is clean, but the guard now has a blind spot

`feat/orm` gained #74 (aggregate), and three branches merged it in. I checked `plan.ts` on all three, since `LITERAL_KEYS` is the one `Set` several open branches edit and a conflict resolution that dropped a side would be silent:

```
feat/orm-nested-create-many  orderBy select include distinct mode _count _avg _sum _min _max
feat/orm-skip-duplicates     … + skipDuplicates
feat/orm-omit                … + omit
```

All three keep #74's five aggregate keys alongside their own, and all three retain #72's `skip` handling and #74's `"aggregate"` operation. Nothing lost. 874 tests pass here.

**But `plan.coverage.test.ts` does not cover `aggregate`, and that is exactly the operation it should.**

The guard walks `READ_ARGS` and `WRITE_ARGS`. `aggregate` has neither — it carries its own `AGGREGATE_ARGS` in `compile/aggregate.ts` — so the operation and its five function arguments are outside the enumeration entirely. Which means the guard does not catch the bug it was written for, on the operation that supplied one of its three motivating instances:

```
# delete "_sum" from LITERAL_KEYS — precisely the bug #74 fixed
bun test packages/gemi/orm/plan.coverage.test.ts
  ->  15 pass, 0 fail
```

This was not visible when the guard was written: `aggregate` did not exist on this branch until the merge. It is visible now, which makes this the moment to close it.

**Verified fix, two lines.** Export `AGGREGATE_ARGS` and add it to the walked list:

```ts
const OPERATIONS = [
  ["aggregate", AGGREGATE_ARGS] as [string, Set<string>],
  ...Object.entries(READ_ARGS),
  ...Object.entries(WRITE_ARGS),
]
```

The guard's own coverage assertion fires immediately, naming every argument it now wants a probe for:

```
error: add a two-value probe to PROBES for: aggregate._count, aggregate._avg,
       aggregate._sum, aggregate._min, aggregate._max
```

and with those probes in place, deleting `_sum` from `LITERAL_KEYS` fails the suite instead of passing it. That is the mechanism working as designed — it simply was not pointed at this operation.

Worth deciding how the list stays honest as a fifth argument set appears. Right now `OPERATIONS` is a hand-written union of two (soon three) exported constants, which is the same shape as #82's three copies of the dialect list — and #82's answer was to derive the enumeration from something `tsc` checks. A `Record<Operation, Set<string>>` over the `Operation` union would make a new operation without an argument set a compile error, rather than a silently unwalked one. Not necessary for this PR; worth a line in the file's header so the next operation added does not repeat this.

nstfkc:
### Post-merge status — the branch survived, the guard's blind spot did not close

`e7781bd` merges `feat/orm` (eight PRs) cleanly and this branch keeps its own contribution: `LITERAL_KEYS` now reads

```
orderBy select omit include distinct mode _count _avg _sum _min _max skipDuplicates
```

— `omit` alongside #76's `skipDuplicates` and #74's five aggregate keys, with `plan.coverage.test.ts` still present. Nothing was lost in the resolution, which is the thing worth checking when this many land at once.

**The `aggregate` gap is still open, and the merge made it more consequential rather than less.** The guard still walks only `READ_ARGS` and `WRITE_ARGS`, and `aggregate` carries its own `AGGREGATE_ARGS` — so deleting `_sum` from `LITERAL_KEYS`, the exact bug #74 fixed, still passes:

```
bun --bun vitest run orm/plan.coverage
  ->  15 passed
```

`aggregate` is on `feat/orm` now rather than queued behind this branch, so the operation the guard cannot see is one every model already exposes. And #92 adds `groupBy` with its own argument set and its own projection key (`by`), which would be a second blind spot on arrival — the same two lines close both if `OPERATIONS` is built from something that has to name every operation, which is the `Record<Operation, …>` shape #82 landed on for exactly this reason.

nstfkc:
### This merged with the `aggregate` blind spot open — it is live on `feat/orm` now

Flagging rather than letting it sit, since the PR is closed and the guard is the thing that was supposed to stop this class recurring. On `origin/feat/orm` (`1ec25d3`):

```
bun --bun vitest run orm                 953 passed        the guard is in
delete "_sum" from LITERAL_KEYS
bun --bun vitest run orm/plan.coverage   15 passed         and does not notice
```

That deletion reintroduces the exact bug #74 fixed. `plan.coverage.test.ts` walks `READ_ARGS` and `WRITE_ARGS`; `aggregate` carries its own `AGGREGATE_ARGS`, so the operation is outside the enumeration entirely — and `aggregate` is on `feat/orm`, so this is an operation every model already exposes.

The merge itself is clean: `LITERAL_KEYS` reads `orderBy select omit include distinct mode _count _avg _sum _min _max skipDuplicates`, so `omit` landed alongside #76's and #74's entries with nothing lost.

Two lines close it — export `AGGREGATE_ARGS` and add `["aggregate", AGGREGATE_ARGS]` to the walked list — after which the guard's own coverage assertion names every probe it wants. And #92 will bring `groupBy` with a third argument set and a third projection key (`by`), so the same two lines are worth spending as a `Record<Operation, …>` that cannot omit an operation, which is the shape #82 landed on for exactly this reason.

Worth saying plainly: this is the second finding this session to merge unaddressed and end up with nothing tracking it — the other is #91's un-awaited `ready`, which turns a bad `DATABASE_URL` into an asynchronous process exit and is also live on `feat/orm`. Both are one-line fixes on a branch that is otherwise in good shape; neither has a PR or an issue holding it.

Issue #67, opened 2026-07-28
ORM: multi-field (composite) relations in `include`


nstfkc:
Branch: `feat/orm`. A relation that joins on more than one field is refused at `packages/gemi/orm/compile/plan-relations.ts:667`:

```
UnsupportedQueryError: gemi ORM does not support 'items' yet (Order.include).
items joins on 2 fields (from: tenantId, orderId). Multi-field relations are not
implemented yet.
```

Prisma allows composite `@relation(fields: [a, b], references: [x, y])`, and the shape shows up mostly in one place: **tenant-scoped composite keys**, where every table carries `tenantId` and every relation joins on `(tenantId, parentId)`. That is a deliberate schema style rather than an exotic one, and an application using it cannot use `include` at all.

## What has to change

The join-key resolution assumes a single field on each side and returns `fields[0]`. Every consumer of that assumption widens to a list:

- **Batched strategy** — the child query filters `parentKey in (…)`, which becomes a tuple `in` (`(a, b) in ((…),(…))` on Postgres; SQLite needs the `or`-of-`and`s form or a values join). The parameter budget doubles, which matters because `ParameterLimitError` is already reachable on SQLite for to-many includes.
- **Lateral strategy** — the correlation predicate in `packages/gemi/orm/compile/correlate.ts` becomes a conjunction. Cheapest of the three.
- **Stitching** — the shaper keys parents by one value today; it needs a composite key that cannot collide (not string concatenation of user data).
- **Relation filters, `_count`, relation ordering** — all four surfaces added after iteration 9 correlate through the same helper and inherit the fix, but each needs its own plan-discrimination test, per the invariant `27776df` established.

## Not in scope here

Composite **unique** keys in `where` already work (`packages/gemi/orm/compile/where.ts:373-394`) and composite primary keys are handled. This issue is only about relation *joins*.

## Acceptance

- `include` across a two-field relation, both directions, both strategies, both dialects, verified by the differential harness.
- A relation filter, an `_count`, and an `orderBy` through the same relation.
- Plan-cache discrimination: two composite relations differing only in field order produce different plans.
- The parameter-limit check accounts for the multiplied placeholder count on SQLite.

nstfkc:
Resolved by a PR now merged into `feat/orm`. Closing by hand — GitHub only auto-closes on merges to the default branch.

PR #79, opened 2026-07-28
A multi-field relation refuses on every surface, and says which one

nstfkc:
**Groundwork for #67, not the feature** — and I want to be plain about that split rather than dress it up.

The feature is a composite key comparison on both sides: a tuple `in` for the batched strategy, a conjunction for the lateral one, a composite stitch key that cannot collide, and a template schema that actually has a two-field relation so the differential harness can compare it. That touches four compilers and wants the coverage the issue asks for (both directions, both strategies, both dialects, plus relation filters, `_count`, `orderBy`, and plan discrimination). It deserves its own PR.

What this does is establish what is true today — and the audit turned up one thing that was wrong.

## The safety property holds, and is now pinned

Seven surfaces correlate over a relation, each with its own compiler:

| surface | before | after |
| --- | --- | --- |
| `include`, to-many | refused | refused |
| `include`, to-one | refused | refused |
| `include` under `lateral` | refused | refused |
| a relation filter | refused | refused |
| `_count` | refused | refused |
| `orderBy` through a relation | refused | refused |
| a nested write | refused | refused |

All seven resolve their link through one function, so none can reach a one-field assumption and join on `fields[0]` — which would return **plausible wrong rows from a query that succeeds**, the worst failure available here.

That was already true. It was not tested, and "all seven happen to go through one function" is exactly the kind of property that stops being true quietly when the eighth surface arrives.

## The message was wrong about where it came from

`single()` hardcoded `"include"` as the operation. So a composite relation named in a `where` reported `(Ledger.include)`, and a nested `connect` on a `create` reported `(LedgerEntry.include)` — sending the reader to a query they had not written. Measured before and after:

```
relation filter (some)      (Ledger.include)   ->  (Ledger.findMany)
_count                      (Ledger.include)   ->  (Ledger.findMany)
orderBy a relation field    (Entry.include)    ->  (Entry.findMany)
nested write: connect       (Entry.include)    ->  (Entry.create)
nested write: child create  (Ledger.include)   ->  (Ledger.create)
```

The fix is one argument threaded from five call sites. The message now also names what the shape needs and offers a way through, instead of only naming what it will not do:

```
gemi ORM does not support 'entries' yet (Ledger.findMany). entries joins on 2
fields (to: tenantId, code). A multi-field relation needs a composite key
comparison on both sides — a tuple 'in' for the batched strategy, a conjunction
for the lateral one — which is not implemented. Query the two models separately
and join them in process, or write the join with 'sql' and 'DB.query'.
```

## The fixture

A `ledger` / `ledgerEntry` pair carrying `@relation(fields: [tenantId, ledgerCode], references: [tenantId, code])` — the tenant-scoped shape the issue names, where every table has `tenantId` and every relation joins on `(tenantId, parentId)`. The template's schema has no example of one.

17 tests walk the seven surfaces, both directions of the relation (they resolve differently — the owning side reads `from`/`to` off its own relation, the other has to find its counterpart through `relationName`), and **one single-field relation on the same models**, so the suite cannot pass for a compiler that refused every relation.

`docs/orm.md` gains the entry, marked *pending* rather than declined — unlike `distinct` and `cursor` in #78, this one is a gap and should say so.

When the feature lands, this file inverts: the refusals become assertions about emitted SQL, and the list of surfaces is already written down.

855 unit tests, full template suite green on SQLite and Postgres 16.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -779,6 +779,12 @@ Stated so you can plan around them rather than discover them:
 - **No lazy loading.** A relation you did not `include` is absent, not a proxy that queries when
   touched.
 - **No `omit`.** Use `select`, or a policy's `redact`.
+- **No multi-field relations.** `@relation(fields: [a, b], references: [c, d])` is refused wherever
+  a relation is correlated — `include` under either strategy, a relation filter, `_count`, an
+  `orderBy` through a relation, and nested writes — rather than joining on the first field and
+  returning plausible wrong rows. The error names the fields on both sides and the operation it came
+  from. This one *is* pending rather than declined; it needs a composite key comparison on both
+  sides (a tuple `in` for the batched strategy, a conjunction for the lateral one).
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
 - **No `groupBy`, `aggregate`, `distinct` or cursor pagination.** These land in
   [Raw SQL](#raw-sql), which exists so that "not implemented" has an answer rather than a shrug.
--- a/packages/gemi/orm/compile/composite-relations.test.ts
+++ b/packages/gemi/orm/compile/composite-relations.test.ts
@@ -1,0 +1,185 @@
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
+
+import { PostgresDialect } from "../dialect/postgres";
+import { SqliteDialect } from "../dialect/sqlite";
+import { UnsupportedQueryError } from "../errors";
+import { ledger, ledgerEntry } from "../fixtures";
+import * as registry from "../registry";
+import { lateralStrategy } from "./lateral";
+import { compileRead } from "./read";
+import { compileWrite } from "./write";
+
+/**
+ * Multi-field relations — `@relation(fields: [a, b], references: [c, d])`.
+ *
+ * Not implemented (#67), and the point of this file is the *shape* of the
+ * refusal rather than the feature:
+ *
+ * 1. **Every surface refuses.** A one-field assumption reached by any path
+ *    would join on the first field and silently return the wrong rows — the
+ *    worst failure available here, because the query succeeds. Seven surfaces
+ *    correlate over a relation, and each has its own compiler; they are safe
+ *    only because all seven resolve their link through one function. That is a
+ *    property to hold, not a coincidence to rediscover.
+ *
+ * 2. **Each names the operation it came from.** They all used to report
+ *    `(Model.include)` — a composite relation named in a `where`, in an
+ *    `orderBy`, or in a nested `connect` on a `create` sent the reader to an
+ *    `include` they had not written.
+ *
+ * When the feature lands this file inverts: the refusals become assertions
+ * about emitted SQL, and the list of surfaces is already here.
+ */
+
+const sqlite = new SqliteDialect();
+const postgres = new PostgresDialect();
+
+beforeEach(() => {
+  registry.clearRegistry();
+  registry.register("Ledger", class { static $schema = ledger });
+  registry.register("LedgerEntry", class { static $schema = ledgerEntry });
+});
+
+afterEach(() => registry.clearRegistry());
+
+/** `[surface, run, model, operation]` — the seven paths that correlate. */
+const SURFACES: [string, () => unknown, string, string][] = [
+  [
+    "include, to-many",
+    () => compileRead(ledger, "findMany", { include: { entries: true } }, sqlite),
+    "Ledger",
+    "findMany",
+  ],
+  [
+    "include, to-one",
+    () => compileRead(ledgerEntry, "findMany", { include: { ledger: true } }, sqlite),
+    "LedgerEntry",
+    "findMany",
+  ],
+  [
+    "include under the lateral strategy",
+    () =>
+      compileRead(
+        ledger,
+        "findMany",
+        { include: { entries: true } },
+        postgres,
+        lateralStrategy,
+      ),
+    "Ledger",
+    "findMany",
+  ],
+  [
+    "a relation filter",
+    () =>
+      compileRead(
+        ledger,
+        "findMany",
+        { where: { entries: { some: { amount: 1 } } } },
+        sqlite,
+      ),
+    "Ledger",
+    "findMany",
+  ],
+  [
+    "_count",
+    () =>
+      compileRead(
+        ledger,
+        "findMany",
+        { include: { _count: { select: { entries: true } } } },
+        sqlite,
+      ),
+    "Ledger",
+    "findMany",
+  ],
+  [
+    "an orderBy through a relation",
+    () =>
+      compileRead(
+        ledgerEntry,
+        "findMany",
+        { orderBy: { ledger: { title: "asc" } } },
+        sqlite,
+      ),
+    "LedgerEntry",
+    "findMany",
+  ],
+  [
+    "a nested write",
+    () =>
+      compileWrite(
+        ledger,
+        "create",
+        {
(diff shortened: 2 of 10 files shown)


nstfkc:
## Review — composite-relation groundwork (#67)

Read the full diff against the merge base (`9b20919`). 855 unit tests pass.

Shipping the audit separately from the feature is the right call and the description is honest about which it is. The valuable part is the seven-surface table: "all seven happen to go through one function" is exactly the kind of property that holds until it doesn't, and the failure it guards against — joining on `fields[0]` and returning plausible wrong rows from a query that *succeeds* — is the worst thing available in this area. Pinning it with a test that also carries a single-field relation on the same models, so the suite cannot pass for a compiler that refuses everything, is what makes it worth having.

The message fix is real: reporting `(Ledger.include)` for a relation filter or `(Entry.include)` for a nested `connect` sends the reader to a query they did not write, and the before/after table measures it rather than asserting it. The new text naming what the shape needs and offering two ways through is the #61 rule applied properly.

### The one thing I would change: make `operation` required

```ts
export function resolveLink(
  parent: ModelSchema,
  child: ModelSchema,
  relation: RelationSchema,
  operation = "include",   // <-
): Link
```

The bug this PR fixes *is* a wrongly-defaulted operation. Leaving the default in place means the eighth surface — the one the seven-surface test exists to anticipate — reintroduces it by omission, silently, and reports `include` for a query that is not one. That is the same failure, one call site later, and it will not be caught by the new test because the new test only knows about the seven that exist.

This codebase has already decided this question, three times, and in each case wrote down why:

- `render`'s `origin` (`compile/fragment.ts`): *"Required, and deliberately so. Optional would reintroduce in miniature the hole this check exists to close — a new call site that omits it skips the ceiling silently… Making it required is the difference between a convention and something `tsc` enforces."*
- `changedFields`' `schema` (`provenance.ts`): *"Required, not optional… an optional schema would mean every caller that omitted it silently got the permissive behaviour."*
- `RelationExecutor.exec`'s `preScoped` (`plan-relations.ts`), which even counts itself: *"It is the fourth parameter in this codebase to be made required for that reason."*

`operation` is the same shape. The justification given — *"Defaulted for the callers that predate it; every call site in the compiler passes a real one"* — is the description of a convention, and the second half of it says the default is already dead code.

**It costs nothing.** I removed the default from `resolveLink` and `correlate`, and `tsc --noEmit` on `packages/gemi` passes with no errors: all eight call sites already pass an operation, and the test files that touch this area do not call either function directly. 829 orm tests still pass. Whether to make it the fifth is a one-line change to two signatures.

### A note on vocabulary, since two things here are called "composite"

This PR is about composite *relations* — `@relation(fields: [a, b], references: [c, d])`. It does not touch composite **primary keys**, which are a different path and currently broken: `matchUniqueKey` accepts a compound `@@id` under Prisma's `a_b` name, but `compileWhere`'s `compileCompoundKey` searches `schema.uniques` alone, and the generator never puts an `@@id` group there — so `findUnique` / `update` / `delete` on a model with `@@id([userId, orgId])` fail both spellings, each error pointing at the other. That is finding 4 on #45 and it is unrelated to this change; flagging it only because "composite support" is about to become an ambiguous phrase in this repo, and the `ledger` / `ledgerEntry` fixture added here is close to what a test for the other one would need.

Marking the docs entry *pending* rather than declined, in contrast to #78's `distinct` and `cursor`, is the right distinction and worth keeping consistent as more of these land.

nstfkc:
Pushed as 0442139. You are right, and the argument you made against the default is the argument I should have made against it myself — the bug this parameter exists to fix *is* a wrongly-defaulted operation, so leaving a default is the same bug one call site later, in the place the seven-surface test cannot see.

The justification I wrote was the tell: *"every call site in the compiler passes a real one"* is a statement that the default is already dead code. Now required, `tsc` clean with no call-site changes — which is the evidence.

I recorded it beside the parameter as the fifth in this codebase made required for this reason, with the three precedents named. Having them enumerated in one place seemed more useful than a fourth independent restatement of the same rule, and `preScoped` was already counting.

---

**On the vocabulary point — I verified the other one, and it is exactly as you describe.** Using this PR's own fixture, which has `@@id([tenantId, code])` and no `@@unique`:

```
where: { tenantId_code: { tenantId: 1, code: "a" } }
  -> 'tenantId_code' is not a field on model Ledger. Known fields: tenantId, code, title.

where: { tenantId: 1, code: "a" }
  -> findUnique needs a unique field. Ledger declares ...
```

Both spellings fail and each error points at the other, so `findUnique` / `update` / `delete` are unusable on any model with a compound `@@id`. That is a sharper failure than the one this PR is about: composite *relations* refuse loudly and consistently, composite *primary keys* refuse in a loop.

Agreed it is out of scope here — it is finding 4 on #45 and a different code path (`compileCompoundKey` searching `schema.uniques`, which the generator never populates from an `@@id` group). But you are right that the fixture is most of what a test for it needs, and it is the natural next piece: it is a smaller, sharper bug than the composite-relation feature, and it lands in `where.ts` and the generator rather than in the four relation compilers.

I will pick it up next unless you would rather have the composite-relation feature itself.

855 unit tests, full template suite green on both dialects.

nstfkc:
### Re-reviewed `0442139` — done, and a correction to my own evidence

Both signatures now require `operation`, no default remains anywhere in `orm/`, and all eight call sites pass a real one — enumerated on this head rather than assumed:

```
correlate.ts:67       resolveLink  4 args  operation
lateral.ts:100        resolveLink  4 args  request.operation
lateral.ts:310        resolveLink  4 args  operation
nested-writes.ts:152  resolveLink  4 args  operation
plan-relations.ts:957 resolveLink  4 args  request.operation
order-relation.ts:88  correlate    6 args  operation
relation-count.ts:103 correlate    6 args  operation
where.ts:213          correlate    6 args  context.operation
```

855 unit tests pass. Writing the rule into the parameter's own doc comment — and incrementing the count to five — is what makes it findable by the person who adds the eighth surface, which is the whole point.

**A correction to my previous comment.** I claimed "`tsc --noEmit` on `packages/gemi` passes with no errors" as evidence that making the parameter required was free. That verification was not sound: `npx tsc` in this environment resolves to a decoy package that prints a banner and exits 0 without type-checking anything, and my grep for `error TS` found nothing because nothing ran. TypeScript is not installed in the checkout I was working from, so I could not have run it.

The substantive claim survives — every call site did already pass an operation, which I had established by enumerating them, and the change now compiles and passes the suite. But the tsc line was evidence I did not have, and I should not have offered it. Flagging it rather than leaving it in the thread, since a later reader would reasonably take it at face value.

nstfkc:
Thanks for the retraction — and it prompted me to check my own toolchain rather than assume it differed, since I have been citing `tsc` as evidence all session.

Mine is real, and I verified it the only way that settles it — by injecting an error rather than trusting a clean run:

```
$ bunx tsc --version
Version 5.9.3

$ printf 'const __probe: number = "not a number";\n' >> orm/plan.ts
orm/plan.ts(373,7): error TS2322: Type 'string' is not assignable to type 'number'.
```

So the typecheck claims in my PRs stand. Worth recording here rather than only knowing it privately, because a reader of this thread now has one retracted `tsc` claim and several unretracted ones, and no way to tell which environment produced which.

The general lesson is the one this session keeps producing from different directions: **a passing check is not evidence until you have seen it fail.** It is the same thing that caught my `plan.coverage.test.ts` on #78 — that test passed against a knowingly-broken `LITERAL_KEYS` until I cleared the cache between compiles, and I only found out because I removed `"omit"` to watch it break. A green run proves nothing about a check that cannot go red.

Nothing outstanding on this PR from either of us.

PR #81, opened 2026-07-28
A compound `@@id` is a usable key

### LINKED ISSUE #80: ORM: `findUnique` / `update` / `delete` are unusable on a model with a compound `@@id`
Branch: `feat/orm`. Found while reviewing #79 (composite *relations*); this is the other thing "composite" means in this repo and it is a separate code path.

A model whose primary key is a compound `@@id([a, b])` — and which declares no `@@unique` — cannot be read or written by key at all. **Both spellings fail, and each error points at the other.**

Reproduced against the `ledger` fixture added in #79, which has `@@id([tenantId, code])` and no `@@unique`:

```ts
Ledger.findUnique({ where: { tenantId_code: { tenantId: 1, code: "a" } } })
// -> 'tenantId_code' is not a field on model Ledger.
//    Known fields: tenantId, code, title.

Ledger.findUnique({ where: { tenantId: 1, code: "a" } })
// -> gemi ORM does not support 'where' yet (Ledger.findUnique).
//    findUnique needs a unique field. Ledger declares ...
```

So the compound form is rejected as an unknown field, and the field-by-field form is rejected for not naming a unique key. A caller following either message is sent to the other.

## Cause

Two places disagree about what counts as a unique key.

- `packages/gemi/orm/compile/unique.ts` — `uniqueKeys()` returns `[schema.primaryKey, ...schema.uniques]`, so `matchUniqueKey` **accepts** `tenantId_code`: the operation is allowed to compile.
- `packages/gemi/orm/compile/where.ts` — `compileCompoundKey` searches `schema.uniques` **alone**:

  ```ts
  const members = schema.uniques.find(
    (group) => group.length > 1 && group.join("_") === key,
  );
  if (!members) return null;   // caller then reports an unknown field
  ```

The generator never copies an `@@id` group into `uniques` — it is already in `primaryKey` — so a compound primary key is unique to one of the two and invisible to the other.

A single-field `@id` is unaffected: it is matched as a plain field, not through the compound path. Composite `@@unique` is unaffected too, which is why the template's `SocialAccount` covers `findUnique` on a compound key and this went unseen — **it has `@@unique([username, provider])`, not `@@id`.**

## Scope

Every operation Prisma types with a `WhereUniqueInput`: `findUnique`, `findUniqueOrThrow`, `update`, `delete`, `upsert`. A join-table model or any tenant-scoped `@@id([tenantId, id])` schema hits it immediately.

## Acceptance

- `findUnique` / `update` / `delete` / `upsert` accept Prisma's compound form for a compound `@@id`, on both dialects.
- The template's schema carries a model with a compound `@@id`, so the differential harness compares it — the gap that hid this is that no such model existed.
- A partial compound key is still refused (it is only unique as a whole).
- `upsert`'s `on conflict` target names the primary key correctly for such a model.

nstfkc:
Closes #80 — which I filed this iteration, after it turned up while reviewing #79. It is the *other* thing "composite" means in this repo, and a separate code path from composite relations.

## The bug

A model whose primary key is a compound `@@id([a, b])`, declaring no `@@unique`, could not be read or written by key **at all**. Both spellings failed, and each error pointed at the other:

```
where: { organizationId_userId: { … } }  ->  'organizationId_userId' is not a field on Membership
where: { organizationId: 1, userId: 2 }  ->  findUnique needs a unique field
```

That affects every operation Prisma types with a `WhereUniqueInput`: `findUnique`, `findUniqueOrThrow`, `update`, `delete`, `upsert`. A join-table model, or any tenant-scoped `@@id([tenantId, id])` schema, hits it immediately.

## Cause

Two places disagreed about what counts as a unique key.

- `unique.ts` — `uniqueKeys()` returns `[primaryKey, ...uniques]`, so `matchUniqueKey` **accepted** `organizationId_userId` and the operation compiled.
- `where.ts` — `compileCompoundKey` searched `schema.uniques` **alone**, found nothing, returned `null`, and the caller fell through to reporting an unknown field.

The generator never copies an `@@id` group into `uniques` — it is already in `primaryKey` — so a compound primary key was unique to one of the two and invisible to the other. The fix is that one search: `uniqueKeys(schema)` instead of `schema.uniques`.

## Why it hid, and what now covers it

The template's only compound key is `SocialAccount`'s `@@unique([username, provider])` — a different path, and one that **was** covered by a differential case. So a compound *primary* key had no model anywhere.

The schema now carries `Membership` with `@@id([organizationId, userId])`, and the harness compares `findUnique` (hit and miss), `update`, `delete`, and both halves of `upsert` against Prisma. `upsert` is the one most worth having: it compiles the key into an `on conflict (…)` target, so naming the wrong columns there would resolve conflicts against the wrong constraint rather than failing.

## A harness assumption the new model exposed

`readTables` — which reads every table back after a write comparison — ordered by `id`:

```ts
prismaDelegate(prisma, model).findMany({ orderBy: { id: "asc" } })
```

Every model in this schema happened to have one. Prisma answers `Unknown argument 'id'` for one that does not, so the comparison never ran and the write cases failed for a reason unrelated to the bug under test. It orders by the model's **declared primary key** now. A stable order still matters — the two clients write from the same seeded state, so an unordered read could differ by storage order alone and report a divergence that is not one.

## Guards against the obvious over-fix

The fix must not become "accept any group of fields joined by an underscore", so three cases are pinned:

- a **partial** compound key is still refused by name (it is only unique as a whole);
- a compound `@@unique` still works;
- a made-up `organizationId_role` is still an unknown field.

846 unit tests (8 new), full template suite green on SQLite and Postgres 16.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -126,6 +126,21 @@ stepping stone: a `Pick<User, "id">` hydrated as a `User` would carry methods re
 query never fetched. See [Rows and entities](./orm-rows-and-entities.md) for the two opt-in levels
 above it — `track` + `save`, and `wrap`.
 
+### Looking a row up by a composite key
+
+`findUnique`, `update`, `delete` and `upsert` need a key that is unique. A composite one — whether
+it is a compound `@@id([a, b])` or an `@@unique([a, b])` — is named in Prisma's compound form:
+
+```ts
+await Membership.findUnique({
+  where: { organizationId_userId: { organizationId, userId } },
+})
+```
+
+Every member is required. A composite key is only unique as a whole, so a partial one would quietly
+become a non-unique lookup — which is the failure `findUnique` exists to prevent, and it is refused
+by name.
+
 ### Per-call options
 
 A second parameter, not a key inside `args` — intersecting Prisma's own arg types with a
--- a/packages/gemi/orm/compile/compound-id.test.ts
+++ b/packages/gemi/orm/compile/compound-id.test.ts
@@ -1,0 +1,125 @@
+import { describe, expect, test } from "vitest";
+
+import { SqliteDialect } from "../dialect/sqlite";
+import { UnsupportedQueryError } from "../errors";
+import { account, membership } from "../fixtures";
+import { compileRead } from "./read";
+import { compileWrite } from "./write";
+
+/**
+ * A compound `@@id`, which was unreachable by key at all (#80).
+ *
+ * Two places disagreed about what counts as a unique key: `matchUniqueKey` asks
+ * `uniqueKeys()`, which includes the primary key, so the operation *compiled* —
+ * and then `compileCompoundKey` searched `schema.uniques` alone, found nothing,
+ * and the caller reported an unknown field. Both spellings failed and each
+ * error pointed at the other:
+ *
+ *     where: { organizationId_userId: { … } }  ->  not a field on Membership
+ *     where: { organizationId: 1, userId: 2 }  ->  needs a unique field
+ *
+ * It hid because the template's only compound key is `SocialAccount`'s
+ * `@@unique([username, provider])` — a different path, and one that was
+ * covered.
+ */
+
+const sqlite = new SqliteDialect();
+const key = { organizationId_userId: { organizationId: 1, userId: 2 } };
+
+describe("a compound @@id is a usable key", () => {
+  test("findUnique compiles both members into the predicate", () => {
+    const plan = compileRead(membership, "findUnique", { where: key }, sqlite);
+
+    expect(plan.text).toContain(`"organizationId" = ?`);
+    expect(plan.text).toContain(`"userId" = ?`);
+    // Both members bound, in the key's declared order. The trailing 1 is
+    // `findUnique`'s own `limit 1`, which every single-row read carries.
+    expect(plan.bind({ where: key })).toEqual([1, 2, 1]);
+  });
+
+  test.each([
+    ["update", { where: key, data: { role: 1 } }],
+    ["delete", { where: key }],
+    [
+      "upsert",
+      {
+        where: key,
+        create: { organizationId: 1, userId: 2 },
+        update: { role: 1 },
+      },
+    ],
+  ])("%s takes it too", (op, args) => {
+    expect(() =>
+      compileWrite(membership, op as never, args, sqlite),
+    ).not.toThrow();
+  });
+
+  /**
+   * `upsert` is the one that needs the key's *identity* rather than a yes: it
+   * compiles into an `on conflict (…)` target, so naming the wrong columns
+   * there would resolve conflicts against the wrong constraint.
+   */
+  test("upsert targets the compound key in its on-conflict clause", () => {
+    const plan = compileWrite(
+      membership,
+      "upsert",
+      {
+        where: key,
+        create: { organizationId: 1, userId: 2 },
+        update: { role: 1 },
+      },
+      sqlite,
+    );
+
+    expect(plan.text).toContain(`on conflict ("organizationId", "userId")`);
+  });
+
+  /**
+   * Still refused, because a composite key is only unique as a whole — a
+   * partial one would quietly become a non-unique lookup, which is the failure
+   * `findUnique` exists to prevent.
+   */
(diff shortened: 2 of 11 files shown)


nstfkc:
## Review — compound `@@id` (closes #80)

Read the full diff against the merge base (`9b20919`). 846 unit tests pass; template suite 432 pass on SQLite and 672 on Postgres 16, only the by-design one-dialect-at-a-time guard failing.

This closes a finding I raised, so I verified it with my **original** reproduction rather than against the cases the PR added — compiled against the artifact this branch actually generates:

```
Membership  primaryKey ["organizationId","userId"]  uniques []

findUnique  { organizationId_userId: {…} }  -> select … where ("organizationId" = ? and "userId" = ?)
update      { organizationId_userId: {…} }  -> update … where ("organizationId" = ? and "userId" = ?)
delete      { organizationId_userId: {…} }  -> delete from … where ("organizationId" = ? and "userId" = ?)
upsert      { organizationId_userId: {…} }  -> on conflict ("organizationId", "userId")
```

The fixture reproduces the exact shape — a compound `@@id` with an empty `uniques` — which is what made this unreachable before. `upsert` is the one I would have insisted on: the key becomes an `on conflict` target, so naming the wrong columns there resolves conflicts against the wrong constraint instead of failing, and the emitted target is right.

The three guards against the over-fix all hold, checked independently:

```
partial compound  { organizationId_userId: { organizationId } }  -> refused by name
made-up compound  { organizationId_role: … }                     -> unknown field
flat form         { organizationId, userId }                     -> still refused (Prisma refuses it too)
composite @@unique  SocialAccount username_provider              -> still compiles
```

The flat form staying refused is the right call and worth keeping explicit: it is what Prisma does, and the temptation when fixing "both spellings fail" is to make both work.

`uniqueKeys(schema)` rather than `schema.uniques` is the one-line fix, and the comment records why the two halves disagreed — which is the part that keeps someone from "simplifying" it back. Naming the reason it hid (the template's only compound key was an `@@unique`, a different path that *was* covered) is the more useful half: the gap was in the fixtures, not in anyone's attention.

### The harness fix is the more interesting half

`readTables` ordering by a hardcoded `id` was an assumption no model had violated, and Prisma answers `Unknown argument 'id'` rather than ignoring it — so the write comparisons would have failed for a reason unrelated to the bug under test, which is the kind of failure that gets misread as "the fix broke something". Ordering by the model's declared primary key is right, and keeping the ordering at all — rather than dropping it — is the detail I would have expected to be lost: the two clients write from the same seeded state, so an unordered read can differ by storage order alone and report a divergence that is not one.

That assumption was load-bearing for every future model without an `id`, so finding it now is worth more than the case that exposed it.

### One process note, about my own run

My first SQLite run showed 65 failures. That was my environment, not this branch: I had just rebuilt `packages/gemi/dist` to get the generator from this branch, and the suite raced the build. A clean re-run passes 432. Recording it because the number looks alarming in isolation and I would rather it be in the thread with its explanation than not.

Nothing outstanding from me. That clears the last of my open findings on #45 — 1 and 3 there remain (the `create` + `include` policy refusal, and `take`/`skip` not being validated as numbers), both untracked and neither touched by this.

nstfkc:
Thanks — and noted on the 65-failure run. That is the third time in this session a stale or mid-rebuild `packages/gemi/dist` has produced an alarming number that meant nothing about the branch, so recording it in the thread with its explanation is the right instinct. It has bitten me on #74, #77 and #79.

It is worth saying plainly, since it keeps recurring: `prisma generate` runs the generator from `packages/gemi/dist/bin/`, a **built artifact**, so the template's committed models are only reproducible if the package is built from the branch under test *first*. `generated.test.ts` checks the artifact matches — but it cannot tell you the generator that produced it was the right one.

On the two remaining #45 findings: both are still untracked, and I would rather they were tracked than remembered.

- **`take` / `skip` not validated as numbers** — I declined this on #74 deliberately: the fix is validating them at the root the way `assertPageArgument` does for relation nodes, which newly rejects `take: "10"` from a query string on *every* read operation. Defensible and breaking, so it wants its own change and its own note.
- **`create` + `include` policy refusal** — I have not looked at it.

I will file both as issues next time round unless you would rather they stayed on #45, so they stop depending on either of us remembering.

PR #82, opened 2026-07-28
Refuse MySQL and MariaDB explicitly, and say what still works

### LINKED ISSUE #71: ORM: MySQL / MariaDB — support it, or refuse it explicitly
Branch: `feat/orm`. `plans/orm/README.md` carries this as one of the two remaining open decisions:

> **MySQL / MariaDB.** Deferred. `DatabaseManager` already infers all four dialects, and the strategy seam keeps the door open, but only SQLite and Postgres are built and tested. Confirm this is acceptable.

Filing it so the decision is visible outside the plan directory rather than to pre-empt it.

## What "deferred" means concretely today

`DatabaseManager` infers a MySQL dialect from a connection URL, and `packages/gemi/orm/dialect/` has two implementations. An application pointed at MySQL therefore gets as far as resolving a dialect and then fails on whatever the first query needs — which is not necessarily an error that says "MySQL is not supported".

One gap is already named properly, and is a good model for the rest: `ReturningUnsupportedError` (`packages/gemi/orm/errors.ts`) exists specifically so the MySQL write path is a named gap rather than a wrong answer. MySQL has no `RETURNING`, so every write that reports its result — `create`, `update`, `delete`, `upsert` — needs `lastInsertRowid` plus a re-select, which is a different statement shape and is not implemented.

## The work, if it is taken

- A third `SqlDialect`: identifier quoting (backticks), `LIMIT`/`OFFSET` form, `ON DUPLICATE KEY UPDATE` instead of `ON CONFLICT`, no `NULLS FIRST/LAST`, no `mode: "insensitive"` (collation-dependent), a different parameter limit.
- The `RETURNING` fallback above, and the decision about what `create` returns when the re-select races.
- Relation strategy: MySQL 8 has `LATERAL` and `JSON_ARRAYAGG`; MariaDB does not have lateral. So the two cannot share a strategy default, which is an argument for treating them as two dialects rather than one.
- A third differential-harness target in CI, which is where most of the cost is — the Postgres half already needed `templates/saas-starter/app/models/postgres.sh` and a documented run order to avoid producing 121 meaningless failures.

## The decision

Three ways to close this, and any of them is better than leaving it ambiguous:

1. **Support it** — the work above.
2. **Refuse it explicitly** — a clear error at dialect resolution naming SQLite and Postgres as the supported set, plus a line in `docs/orm.md`. Cheap, and it converts a confusing runtime failure into a supported-matrix statement.
3. **Support it partially and say so** — reads only, or MySQL 8 only.

(2) is the minimum that should ship with the ORM, whichever of the three is chosen long-term.

nstfkc:
Closes #71 at its **option (2)** — the one the issue names as *"the minimum that should ship with the ORM, whichever of the three is chosen long-term."* Building a third dialect, or shipping a documented partial, remain open decisions and are not pre-empted here.

## Measured first

A MySQL URL today:

```
inferDialect        -> mysql
new DatabaseManager -> ok, dialect = mysql
dialectFor          -> THREW
```

`dialectFor` is called by `Model.$exec` **per query**. So an application pointed at MySQL starts, passes a health check, serves everything that does not touch a model, and raises on the first `findMany`. The refusal existed; it arrived at the latest and most expensive moment.

## The message separates the compiler from the connection

It said *"does not support the 'mysql' dialect **yet**"*, which invites the reading that the database is unusable. It is not — and this is the part worth being precise about:

**Bun's client speaks all four dialects.** Pointed at MySQL, the connection works, `DB.query` / `DB.sql` run, and transactions work. What is missing is a `SqlDialect` for the query *compiler*. A reader who concludes the connection is dead migrates a database they did not need to.

The message now says what is unavailable, what still works, and what to do:

```
The gemi ORM does not support the 'mysql' dialect. Only sqlite and postgres are
implemented — see the supported matrix in docs/orm.md.

The *connection* is fine: Bun's client speaks mysql, so raw SQL through
DB.query / DB.sql and transactions all work. It is the query compiler that has
no mysql dialect, so every model operation raises this.

Point DATABASE_URL at Postgres or SQLite to use the ORM, or keep using Prisma's
own client for this database.
```

It also **stops saying "yet"**. That word became load-bearing in #78/#68: `distinct` and `cursor` say *"a decision rather than a gap"*, genuinely unimplemented arguments say *"yet"*, and this is neither — it is a supported-matrix statement, and "yet" promised a release that is not planned.

## `ormSupports(dialect)`, and a boot-time warning

Exported, so an application can assert it at startup rather than discover it on the first query:

```ts
import { ormSupports } from "gemi/orm"
if (!ormSupports(DB.dialect)) throw new Error("this deployment needs Postgres")
```

`DatabaseServiceProvider` warns in development when the configured dialect has no compiler.

**A warning rather than a throw, deliberately.** Refusing to construct the manager would break raw SQL that works correctly today, in order to pre-empt an error the ORM already raises clearly. The supported set is a property of the compiler, not of the connection — and the change should not contradict the message it just wrote.

A test pins that `ormSupports` and `dialectFor` **cannot disagree**: a boot check that differed from the query-time failure about which databases work would be worse than no check at all.

## Verification

8 new tests; 859 orm/database tests pass. Full template suite green on SQLite. Pre-existing and unchanged: the six router/i18n files in `packages/gemi` (7 tests), confirmed against the base commit, and `generated.test.ts`'s generator-linkage probe.

Not run against a live MySQL server — nothing here connects to one, and the behaviour under test is dialect *resolution*, which is a pure function of the URL.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -767,8 +767,24 @@ matters.
 that runs each query through both gemi and Prisma and compares the rows. `DatabaseManager` infers
 the dialect from `DATABASE_URL`.
 
-MySQL and MariaDB are **not** implemented. The dialect seam is there and `DatabaseManager` already
-recognises them, but nothing is built or tested behind it.
+MySQL and MariaDB are **not** implemented, and that is a statement about the supported matrix rather
+than a release that is coming. The dialect seam is there and `DatabaseManager` already recognises
+them, but nothing is built or tested behind it.
+
+**The connection still works, which is the part worth being precise about.** Bun's client speaks all
+four dialects, so pointed at MySQL an application connects, `DB.query` / `DB.sql` run, and
+transactions work. What is missing is a `SqlDialect` for the query *compiler*, so every model
+operation raises `UnsupportedDialectError` — with a message that says exactly that, rather than
+leaving you to conclude the database is unusable.
+
+In development the mismatch is reported **at boot** instead of on the first model query, since
+otherwise an application pointed at MySQL starts, passes a health check, and fails on the first
+`findMany`. `ormSupports(dialect)` is exported if you would rather assert it yourself:
+
+```ts
+import { ormSupports } from "gemi/orm"
+if (!ormSupports(DB.dialect)) throw new Error("this deployment needs Postgres")
+```
 
 ## Not in scope
 
--- a/packages/gemi/database/DatabaseServiceProvider.ts
+++ b/packages/gemi/database/DatabaseServiceProvider.ts
@@ -1,3 +1,4 @@
+import { ormSupports } from "../orm/dialect";
 import { ServiceProvider } from "../support/ServiceProvider";
 import { withDefaults } from "../support/withDefaults";
 import { databaseConfigDefaults, type DatabaseConfig } from "./config";
@@ -7,13 +8,49 @@ export class DatabaseServiceProvider extends ServiceProvider {
   register() {
     this.app.singleton(
       DatabaseManager,
-      () =>
-        new DatabaseManager(
+      () => {
+        const database = new DatabaseManager(
           withDefaults(
             databaseConfigDefaults(),
             this.app.config.get<DatabaseConfig>("database", {}),
           ),
-        ),
+        );
+
+        warnIfOrmUnavailable(database);
+        return database;
+      },
     );
   }
+}
+
+/**
+ * Say it at boot, not on the first model query.
+ *
+ * `DatabaseManager` connects to MySQL and MariaDB without complaint — Bun's
+ * client speaks them — and the ORM's compiler has no dialect for either. So a
+ * deploy pointed at one starts, passes a health check, serves whatever does not
+ * touch a model, and raises `UnsupportedDialectError` on the first `findMany`.
+ * That is the latest and most expensive moment to learn it.
+ *
+ * A **warning rather than a throw**, deliberately: raw SQL through `DB.query`
+ * and `DB.sql` genuinely works on those dialects, and so do transactions.
+ * Refusing to construct the manager would break code that is correct today to
+ * pre-empt an error the ORM already raises clearly. The supported set is a
+ * property of the query compiler, not of the connection.
+ *
+ * Development only, for the same reason the slow-transaction warning is: it is
+ * a diagnostic aimed at whoever is choosing the database, and it should not
+ * add noise to a production log every time the container starts.
+ */
+function warnIfOrmUnavailable(database: DatabaseManager): void {
+  if (process.env.NODE_ENV !== "development") return;
+  if (ormSupports(database.dialect)) return;
+
+  console.warn(
+    `[gemi] DATABASE_URL points at ${database.dialect}, which the gemi ORM ` +
+      `does not implement — only sqlite and postgres are. The connection ` +
+      `works, so raw SQL through DB.query / DB.sql and transactions are fine, ` +
+      `but every model operation will raise UnsupportedDialectError.\n` +
+      `(development only)`,
+  );
 }
--- a/packages/gemi/orm/dialect/index.ts
+++ b/packages/gemi/orm/dialect/index.ts
@@ -183,14 +183,47 @@ export interface ConstraintViolation {
   constraint?: string;
 }
 
+/**
+ * The ORM was asked to compile against a dialect it does not implement.
+ *
+ * **The split this message has to convey**, because it is the surprising part:
+ * `DatabaseManager` connects to MySQL and MariaDB perfectly well — Bun's client
+ * speaks all four — so raw SQL through `DB.query` / `DB.sql` works, and
+ * transactions work. What does not exist is a `SqlDialect` for them, so
+ * everything that *compiles* a statement stops here.
+ *
+ * A caller who reads only "not supported" reasonably concludes the connection
+ * is unusable, which is wrong and is the more expensive misreading: it is the
+ * ORM that is unavailable, not the database.
+ */
 export class UnsupportedDialectError extends Error {
   constructor(dialect: Dialect) {
     super(
-      `The gemi ORM does not support the '${dialect}' dialect yet. ` +
-        `Only sqlite and postgres are implemented.`,
+      `The gemi ORM does not support the '${dialect}' dialect. Only sqlite ` +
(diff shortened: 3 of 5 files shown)


nstfkc:
## Review — refuse MySQL explicitly (closes #71 at option 2)

Read the full diff against the merge base (`9b20919`). 841 orm/database tests pass, and the typecheck is clean — verified real this time by injecting an error first (1 error), restoring (0), after your note on #79. Thank you for checking your own toolchain rather than assuming mine was the odd one; that saved me from repeating the mistake.

The framing is right and the decisions are the ones I would defend:

- **Warn rather than throw at boot.** Refusing to construct the manager would break raw SQL that genuinely works on MySQL today, in order to pre-empt an error the ORM already raises clearly. The reasoning that "the supported set is a property of the query compiler, not of the connection" is also what makes the *message* correct, and the change not contradicting the message it just wrote is the kind of consistency that is easy to lose.
- **Separating connection from compiler in the error text.** "Not supported" reasonably reads as "this database is unusable", and the expensive misreading is migrating a database you did not need to. Naming what still works — `DB.query`, `DB.sql`, transactions — is the part that prevents it.
- **Dropping "yet."** That word has become load-bearing across #78 and #68, and this is neither a by-design refusal nor an unimplemented argument; it is a supported-matrix statement. Getting a third category right rather than forcing it into one of the two is the correct call.

Pinning that `ormSupports` and `dialectFor` cannot disagree is exactly the right instinct — a boot check that differed from the query-time failure would be worse than no check.

### The guard does not survive the change it is guarding against

The supported set is written three times: `ormSupports`'s `||` chain, `dialectFor`'s `if` chain, and the agreement test's hand-written `["sqlite", "postgres", "mysql", "mariadb"]`. The test enumerates a copy of the `Dialect` union rather than the union itself, so it only covers the members someone remembered to list.

Adding a fifth member and changing nothing else:

```
# packages/gemi/database/dialect.ts
export type Dialect = "sqlite" | "postgres" | "mysql" | "mariadb" | "duckdb"

bunx tsc --noEmit -p tsconfig.json   ->  0 errors
bun test packages/gemi/orm/dialect   ->  91 pass, 0 fail
```

Nothing notices. And "someone adds a dialect to the union" is precisely the moment the guard exists for — it is the first step of implementing MySQL, and the step where a half-added dialect would report as supported by one function and unsupported by the other.

**A `satisfies` map in source fixes it, and I verified both halves.** With the map in `dialect/index.ts`:

```ts
const SUPPORTED = {
  sqlite: true, postgres: true, mysql: false, mariadb: false,
} satisfies Record<Dialect, boolean>

export function ormSupports(dialect: Dialect): boolean {
  return SUPPORTED[dialect]
}
```

adding `duckdb` to the union becomes a compile error in a file the build actually checks:

```
orm/dialect/index.ts(230,3): error TS1360: Type '{ sqlite: true; postgres: true;
mysql: false; mariadb: false; }' does not satisfy the expected type 'Record<Dialect, boolean>'.
```

and with the union restored it is 0 errors, 91 tests passing. `dialectFor` can read the same map, which also collapses the two hand-maintained lists into one.

**Worth knowing before reaching for the obvious placement:** putting the exhaustiveness in the *test* file does not work. `packages/gemi/tsconfig.json` excludes `**/*.test.ts`, and vitest transpiles without type-checking, so a `Record<Dialect, true>` there is a type error nothing ever evaluates — I confirmed it only errors when the file is compiled directly. The check has to live in source to be enforced.

This is the same rule the codebase has now invoked five times, most recently on #79: where omitting or forgetting something silently produces the wrong behaviour rather than an error, `tsc` should be the one enforcing it rather than a convention. The agreement *test* is still worth keeping — it checks the two functions agree in behaviour, which the type cannot — but its input should come from the type.

Nothing blocking. Not running against a live MySQL is the right scope call, and saying so explicitly is better than a test that pretends to.

nstfkc:
Pushed as 20c6a9b. The finding is right and the fix is the one you specified — including the part that would have caught me out.

**One `COMPILERS` map, `satisfies Record<Dialect, SqlDialect | null>`**, which collapses both hand-maintained lists into one. Verified by adding a fifth member rather than by reading:

```
export type Dialect = … | "duckdb"

orm/dialect/index.ts(264,3): error TS1360: … Property 'duckdb' is missing in type
'{ sqlite; postgres; mysql; mariadb; }' but required in type 'Record<Dialect, SqlDialect>'
```

0 errors with the union restored. Same discipline as the `LITERAL_KEYS` guard on #78 — a check that has not been seen to go red is not a check.

**Your note about placement saved me the wrong version of this.** The test file is exactly where I would have put it, and it would have been a type error nothing ever evaluates: the tsconfig excludes test files and vitest transpiles without type-checking. That is a trap worth having in the code rather than only in a review thread, so the comment says it.

The agreement test stays, for the reason you gave — it checks the two functions agree in *behaviour*, which the type cannot — but it walks `everyDialect()` now instead of a copy of the union.

That makes six things in this codebase moved from convention to `tsc`: `render`'s `origin`, `changedFields`' `schema`, `exec`'s `preScoped`, `resolveLink`'s `operation`, `ScopedPolicy`'s abstract members, and this. The rule they share is the one worth stating: **where forgetting something silently produces the wrong behaviour rather than an error, the type system should be the one that notices.**

---

On the toolchain: glad it helped. Checking mine was prompted entirely by your retraction — the useful part was not that mine turned out fine, it was that neither of us had reason to believe either way until one of us injected an error.

867 orm/database tests, full template suite green on SQLite.

nstfkc:
### Re-reviewed `20c6a9b` — confirmed, and better than what I proposed

One `COMPILERS` map that both `ormSupports` and `dialectFor` read, plus `everyDialect()` so the test walks the union instead of a copy — that collapses all three hand-maintained lists into one, where I had only suggested fixing two of them.

Verified the guard now fires. Adding a fifth member to `Dialect` and changing nothing else:

```
orm/dialect/index.ts(264,3): error TS1360: Type '{ sqlite: SqliteDialect;
postgres: PostgresDialect; mysql: null; mariadb: null; }' does not satisfy
the expected type 'Record<Dialect, SqlDialect>'.
```

Before, the same edit was 0 errors and 91 passing tests. With the union restored it is clean again and 841 tests pass.

Keeping the behavioural agreement test *and* driving it from `everyDialect()` is the right combination: the type cannot check that the two functions agree at runtime, and the test could not know about a dialect nobody added to its list.

PR #83, opened 2026-07-28
Writes through an implicit many-to-many join table

### LINKED ISSUE #66: ORM: writes through an implicit many-to-many join table
Branch: `feat/orm`. Reading an implicit many-to-many works — iteration 3 built the two-hop load through the generated join table, in both directions, on both dialects, and iteration 9 taught the lateral strategy to decline the shape and fall back to batching with identical rows.

**Writing one does not.** `packages/gemi/orm/compile/nested-writes.ts:123`:

```
UnsupportedQueryError: gemi ORM does not support 'data.tags' yet (Post.create).
tags is an implicit many-to-many. Writing through its join table is not implemented yet.
```

## Why it is separate from nested writes generally

Prisma's implicit m-n writes rows into a join table that **has no model**. There is no registered class, no generated base, no `$schema` to compile against — so the two operands that already work for ordinary relations (`connect`, `create`) have nothing to compile *to*. The compiler has to emit statements against a table it knows only from relation metadata:

- name `_PostToTag`, columns `A` and `B`
- assignment by **field name**, not declaration order (`7c48cd8` — the self-referential case is where the difference becomes visible and wrong)
- the pair is the primary key, so a repeated `connect` is a unique violation rather than a no-op

The read side already resolves all of this in `packages/gemi/orm/compile/plan-relations.ts`; the write side needs the same resolution pointed the other way.

## Scope

`connect`, `disconnect`, `set`, `create`, `connectOrCreate` under an implicit m-n relation, in `create` and `update`. `set` is the one worth calling out — Prisma implements it as delete-all-then-insert for that parent, which is two statements and therefore needs the transaction.

## Testing

The template's own schema has **no many-to-many relation**, so the differential harness cannot reach one. The existing coverage is `templates/saas-starter/app/models/relations.many-to-many.test.ts`, which builds a `Post`/`Tag` pair with DDL taken verbatim from `prisma migrate diff` and asserts against Prisma's *documented* shape rather than a second generated client.

That is adequate for reads and thin for writes, where the failure mode is a row written into the wrong column of the join table — which a fixture asserting its own expectations will happily agree with. Adding an m-n pair to the template schema, so the differential harness compares against a real Prisma client, is the cheaper fix and would also close the same gap for the read side.

## Acceptance

- `connect` / `disconnect` / `set` / `create` on an implicit m-n, compared against a real Prisma client rather than a hand-written expectation.
- The self-referential case (`followers` / `following` on one model) assigns `A` / `B` by field name, in both directions.
- A repeated `connect` behaves as Prisma does rather than raising a raw driver error.
- Every statement inside one transaction.

nstfkc:
Closes #66 — `connect`, `disconnect`, `set` and `create` on an implicit many-to-many, in both `create` and `update`.

## Why it needed its own path

The join table has **no model**: no registered class, no generated base, no `$schema`. So the two operands that already work for an ordinary relation have nothing to compile *to*. These statements are emitted directly against `_PostToTag`, the way the read side's `readPairs` already does — with placeholders from the dialect rather than hardcoded, since SQLite writes `?` and Postgres `$1`.

Which column is which end comes from `resolveLink`, never from declaration order. That distinction is invisible on `Post` / `Tag` and **wrong** on a self-relation, where both ends are the same model and only the field name separates them.

Three properties worth calling out:

- **Policies still apply to the rows.** Only the join table — which has nothing an application could scope — is written directly. The rows the pairs point at go through the related model's own `$exec`, so a `connect` cannot reach a row that model's policies hide, and a `create` gets its `onCreate`.
- **A repeated `connect` is a no-op**, via `on conflict do nothing`. The pair is the table's primary key, so without it the second connect of the same pair is a raw driver unique violation — neither Prisma's behaviour nor a useful one. Both dialects spell it identically, so it needs no dialect branch.
- **`set` is two statements** — delete then insert — and `$exec` already opens a transaction for any plan carrying steps, so it is atomic.

## The coverage was the point, and it found a second bug

#66 notes that the template has no m-n relation, so the differential harness could not reach one **on the read side either** — the existing coverage builds its own tables and asserts Prisma's *documented* shape, which is a fixture agreeing with itself, and the issue calls that "thin for writes, where the failure mode is a row written into the wrong column of the join table".

So a `Post` / `Tag` pair is in the template schema now, and every case here is compared against a real Prisma client with the table contents read back.

Doing that immediately failed five of the eight, for a reason that has nothing to do with join tables:

```
Post.update({ where, data: { tags: { connect: [...] } } })
  -> gemi ORM does not support 'data' yet (Post.update).
     At least one field must be updated.
```

An update whose `data` names only relations has **no column to set**. Prisma accepts it, and there is no `UPDATE … SET` that expresses it. Such an update now compiles to a **select** of the same returning list: the nested steps need the parent's key and the caller's `select` / `include` still has to be honoured, and both come from a list a select produces just as well as `update … returning` does. Same plan shape, same steps, one statement that happens to read.

With neither an assignment nor a nested write it is still an error — that case genuinely has nothing to do.

## Verification

Eight differential cases against a real Prisma client, table contents compared: `connect`, `create`, both together, `disconnect`, `set`, `set: []`, the repeated `connect`, and the relation written **from the other side** (`Tag.update({ data: { posts: { connect } } })`).

838 unit tests, full template suite green on SQLite and Postgres 16.

## Not in this PR

`connectOrCreate` on an m-n. It is refused generally for every relation (#75's list, with the reason), and adding it only for join tables would make the operand set inconsistent across relation kinds for no good reason.

Note this touches `nested-writes.ts`, which #75 also changes — the conflict is in different functions (`planOne`'s routing versus `planForeignSide`) and should be mechanical, but it is the file to read carefully on whichever merges second.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -317,7 +317,7 @@ scope. The rows the pairs point at still go through the related model's own oper
 `connect` cannot reach a row that model's policies hide, and a `create` gets its `onCreate`.
 
 That extends to `set`, which has to delete before it inserts: it clears **the links you can see**,
-not every link. A pair pointing at a row the related model's policies hide survives — otherwise
+not every link — in one `delete … in (…)`, not one statement per link. A pair pointing at a row the related model's policies hide survives — otherwise
 `set` would quietly do what `disconnect` refuses. With no policy on that model every link is
 visible and `set` clears all of them, exactly as Prisma does.
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -316,6 +316,11 @@ Only the join table itself is written directly; it has no model and nothing an a
 scope. The rows the pairs point at still go through the related model's own operations, so a
 `connect` cannot reach a row that model's policies hide, and a `create` gets its `onCreate`.
 
+That extends to `set`, which has to delete before it inserts: it clears **the links you can see**,
+not every link. A pair pointing at a row the related model's policies hide survives — otherwise
+`set` would quietly do what `disconnect` refuses. With no policy on that model every link is
+visible and `set` clears all of them, exactly as Prisma does.
+
 Self-referential ones work too — `Thing.related Thing[]` — as long as the generated files are
 current. Prisma assigns that join table's two columns by *field name*, which the generator now
 records; an artifact from before it does raises rather than guessing, and `prisma generate` fixes
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -306,6 +306,16 @@ Filtering, counting and ordering all work across an **implicit many-to-many** to
 table is a second table inside the subquery and changes nothing else about how you write the
 query.
 
+**Writing one works too**, through the same relation key inside `data`: `connect`, `disconnect`,
+`set` and `create`. `set` replaces the whole set — a delete and then inserts, inside the same
+transaction — and connecting a pair that is already there is a no-op rather than an error, as it is
+in Prisma. An `update` whose `data` names only relations is fine: there is no column to set, so the
+row is read rather than written and the relation work still happens.
+
+Only the join table itself is written directly; it has no model and nothing an application could
+scope. The rows the pairs point at still go through the related model's own operations, so a
+`connect` cannot reach a row that model's policies hide, and a `create` gets its `onCreate`.
+
 Self-referential ones work too — `Thing.related Thing[]` — as long as the generated files are
 current. Prisma assigns that join table's two columns by *field name*, which the generator now
 records; an artifact from before it does raises rather than guessing, and `prisma generate` fixes
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -179,11 +179,27 @@ function planOne(
 
     if (owning) {
       planOwningSide(
-        schema, relation, child, link, key, operand, operation, at, out,
+        schema,
+        relation,
+        child,
+        link,
+        key,
+        operand,
+        operation,
+        at,
+        out,
       );
     } else {
       planForeignSide(
-        schema, relation, child, link, key, operand, operation, at, out,
+        schema,
+        relation,
+        child,
+        link,
+        key,
+        operand,
+        operation,
+        at,
+        out,
       );
     }
   }
@@ -242,7 +258,11 @@ function planOwningSide(
   // Whether that is the case is a property of the argument *shape*, not of its
   // values, so the choice is made here at compile time and the two forms are two
   // plans. That is what keeps a `connect` from silently costing a round trip.
-  if (typeof operand !== "object" || operand === null || Array.isArray(operand)) {
+  if (
+    typeof operand !== "object" ||
+    operand === null ||
+    Array.isArray(operand)
+  ) {
     throw new UnsupportedQueryError(
       `data.${relation.name}.connect`,
       schema.name,
@@ -456,9 +476,6 @@ function planJoinTable(
       const parentKey = parent[parentField];
       if (parentKey === null || parentKey === undefined) return;
 
-      // `set` replaces the whole set, so what is there now goes first. Prisma
-      // does the same, and it is why this operand needs the transaction the
-      // step already runs in.
       // `set` replaces the whole set, so what is there now goes first — but
       // **only the part of it this caller can see**.
       //
@@ -502,13 +519,38 @@ function planJoinTable(
             false,
           )) as Record<string, unknown>[];
 
-          for (const row of visible) {
+          // One statement, not one per link. The scoped form is a `delete`
+          // over an `in` list rather than a loop, because this repo counts
+          // *statements*: `plans/orm/benchmarks.md` measures include trees that
+          // way, and the lateral strategy exists to remove `N - 1` of them. A
+          // loop here would have made `set` the one write that reintroduces
+          // them — invisible on SQLite, where round trips are in-process, and
+          // paid in full on Postgres, where they are not.
+          //
+          // Placeholders are built per length rather than bound as one array
+          // parameter. `dialect.inList` does the latter on Postgres, but it
+          // returns a `Fragment` and this is the one path that emits SQL
+          // without that pipeline — there is no model to route the join table
+          // through. The cost is one server-side plan per distinct link count;
+          // the ORM's own plan cache is unaffected, since it keys on the
+          // argument shape and this text never reaches it.
+          const targets = Array.from(visible, (row) => row[childField]);
+
+          // Every existing link points at a row this caller cannot see, so
+          // there is nothing to clear and `in ()` is a syntax error on both
+          // dialects. The loop this replaced degenerated to zero iterations
+          // here; the single statement has to say so explicitly.
+          if (targets.length > 0) {
             await runPairStatement(
               executor,
               `delete from ${quoted(join.table)} where ` +
                 `${quoted(join.parentColumn)} = ${dialect.placeholder(0)} and ` +
-                `${quoted(join.childColumn)} = ${dialect.placeholder(1)}`,
-              [parentKey, row[childField]],
+                `${quoted(join.childColumn)} in (` +
+                targets
+                  .map((_, index) => dialect.placeholder(index + 1))
+                  .join(", ") +
+                `)`,
+              [parentKey, ...targets],
             );
           }
         }
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -1,7 +1,9 @@
 import { UnsupportedQueryError } from "../errors";
+import type { SqlDialect } from "../dialect";
 import type { ModelSchema, RelationSchema } from "../schema";
 import type { Binder } from "./fragment";
 import {
+  type Link,
   type NestedWriteStep,
   relatedSchema,
   resolveLink,
@@ -71,6 +73,12 @@ export function planNestedWrites(
   operation: string,
   /** Re-locates `data` inside the call's argument tree at bind time. */
   locateData: (args: any) => any,
+  /**
+   * Needed only by the implicit many-to-many path, which emits statements
+   * against a table with no model and therefore has no `Fragment` pipeline to
+   * number its placeholders — SQLite writes `?` and Postgres `$1`.
+   */
+  dialect: SqlDialect,
 ): NestedWritePlanning {
   if (typeof data !== "object" || data === null || Array.isArray(data)) {
     return EMPTY;
@@ -95,7 +103,7 @@ export function planNestedWrites(
     const node = (data as Record<string, unknown>)[key];
     const locate = (args: any) => locateData(args)?.[key];
 
-    planOne(schema, relation, node, operation, locate, planning);
+    planOne(schema, relation, node, operation, locate, planning, dialect);
   }
 
   return planning;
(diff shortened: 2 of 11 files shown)


nstfkc:
## Review — implicit many-to-many writes (closes #66)

Read the full diff against the merge base (`9b20919`). 812 unit tests pass.

Emitting against the join table directly is the only thing available — it has no model to route through — and taking the column identity from `resolveLink` rather than declaration order is the detail that matters, for the reason given: invisible on `Post`/`Tag`, wrong on a self-relation. `on conflict do nothing` for a repeated `connect` is right and the note that both dialects spell it identically is worth having. Finding the "update whose `data` names only relations" bug by adding real coverage, and answering it with a select of the same returning list rather than a special case, is a good outcome — the shape genuinely has no `UPDATE … SET`.

Policy routing is right for three of the four operands: `connect`, `disconnect` and `create` all go through the child's own `$exec`, so a caller cannot name a row the child's policies hide.

### `set` is the fourth, and it deletes links to rows the caller cannot see

The `set` branch runs an unscoped delete before it inserts:

```sql
delete from "_PostToTag" where "A" = ?
```

That removes *every* pair for the parent, including pairs pointing at children the child's policies would hide — and unlike `disconnect`, nothing consults the child model on the way. Against SQLite, with `Tag` scoped to labels starting `mine`:

```
seeded pairs:      [{A:1,B:1}, {A:1,B:2}]      B=1 "mine-a"  B=2 "theirs-b"
caller sees tags:  ["mine-a"]                   the hidden tag is invisible to this caller

set: [{ id: 1 }]        -> pairs now [{A:1,B:1}]     the hidden link is gone, silently
disconnect: [{ id: 2 }] -> THREW  No Tag found (Tag.findUniqueOrThrow)
```

So the *same effect* — removing the link to a row the caller cannot see — is refused when named explicitly and permitted when reached through `set`. That is an authorization boundary that disagrees with itself inside one PR, and `disconnect`'s refusal is the evidence that the intended answer is the strict one.

It is not a read leak: nothing comes back, and the caller cannot tell from the result that a second link existed. It is an unscoped *write* — state the caller could not otherwise reach, mutated without an error.

**#75 refuses `set` on an ordinary relation for exactly this reason**, and names it in its `REFUSED` table: *"It replaces the whole set, so it has to disconnect what is there now."* The reasoning transfers verbatim; only the storage differs. Two open PRs currently disagree about whether that operand is safe.

Three ways out, in the order I would consider them:

1. **Scope the delete.** Resolve the child keys the caller can actually see — the child's `$exec` is already being called for the named items — and delete `"A" = ? and "B" in (…)`. `set` then means "replace the set I can see", which is what every other policy in this ORM does: narrow, never widen. Unpolicied models are unaffected, so Prisma parity is unchanged where there is no policy to apply.
2. **Refuse `set` when the child model has policies**, consistent with #75, and say why.
3. Document it as a known divergence — weakest, since `disconnect` right beside it does the opposite.

Whichever, it wants the test the other way round from the one I wrote: a policied child, a hidden link, and an assertion that it survives.

### Smaller

- **`quote()` is hand-rolled while `dialect` is in scope.** `planJoinTable` takes a `SqlDialect` and uses it for `placeholder()`, then quotes identifiers with a local `name.replace(/"/g, '""')` helper. `dialect.quoteIdent` does the same thing *and* rejects a NUL byte, which exists because NUL is the parameter sentinel in `compile/fragment.ts`. The join table's identifiers are Prisma-generated so nothing can reach it today, but the whole discipline here is that identifiers go through the dialect — and this is the one file that now emits SQL without a `Fragment` pipeline, so it is the one most worth holding to it.
- **A second merge-order conflict, beyond the one you flagged.** This branch calls `resolveLink(schema, child, relation)` with three arguments. #79 makes the fourth (`operation`) **required** — deliberately, so a new call site cannot inherit `"include"` silently. Whichever of the two merges second gets a compile error here rather than a mechanical conflict, which is the good outcome, but it is worth knowing it is coming: the new call site wants `operation` threaded like the others.

Everything else reads correctly, and the differential coverage — eight cases with table contents compared, including the relation written from the other side — is the part that makes this reviewable at all.

nstfkc:
Pushed as 88c7e89. The `set` finding is right, and the argument that settled it for me is the one you made: **`disconnect` three lines below refuses the same effect.** A boundary that permits through one operand what it refuses through another is not a boundary, and the disagreement was inside one function.

I took option 1 — scope the delete — over refusing, because refusing would make `set` unavailable on any policied model while the narrow reading is available and is what every other policy here does. The existing links are read from the join table, filtered through the child's own `findMany`, and only those are deleted. `set` means "replace the set I can see".

**Two tests, and the second is the one I would have missed if you had only asked for the first.** You said the test wants to be the other way round — a hidden link that survives — and that is there. But a fix of "delete nothing" would also pass it. So there is a second: with no policy on the child, every link is visible and `set` still clears all of them, unchanged from Prisma. Verified by restoring the unscoped delete and watching the hidden-link test fail (1 failed / 46 passed), and by checking the unpolicied case passes both ways.

**On `quote()`** — you are right, and the reasoning is the part worth keeping: `dialect.quoteIdent` additionally rejects a NUL byte because NUL is the parameter sentinel in `compile/fragment.ts`. Nothing can reach that today since Prisma generates the identifiers, but this is the one file emitting SQL *without* the `Fragment` pipeline, which makes it the file to hold to the rule rather than the one to excuse from it. Replaced, with that noted.

**On the second merge-order conflict** — thank you for spotting it. Agreed a compile error is the good outcome, and it is the one #79 was designed to produce: the whole point of making `operation` required was that a new call site cannot inherit `"include"` silently, and this branch adds exactly such a call site. Whichever merges second, the fix is to thread `operation` like the other eight.

I have deliberately not pre-empted it by passing a fourth argument here, since on this branch that is a type error.

838 unit tests, 47 policy tests, full template suite green on SQLite and Postgres 16.

nstfkc:
### Re-reviewed `88c7e89` — the boundary agrees with itself now

My original probe, unchanged. A `Tag` scoped to labels starting `mine`, a post linked to one visible and one hidden tag:

```
                     6fc86ee                    88c7e89
set: [{ id: mine }]  [{A:1,B:1}]                [{A:1,B:1},{A:1,B:2}]   hidden link survives
disconnect hidden    THREW No Tag found         THREW No Tag found      unchanged
```

And `set` still means `set` — verified separately on an unpolicied model, where every link is visible and the behaviour should be exactly what it was:

```
set a -> b   [{A:1,B:2}]      the old link goes, the new one lands
set []       []               clears
```

Option 1 was the right one to take, and the comment gives the reason I would want a future reader to find: *"replace the set I can see", which is what every other policy in this ORM does: narrow, never widen* — with the parity note that an unpolicied child sees every link and gets byte-for-byte the old behaviour. `quoteIdent` replacing the local helper is the other half, and the reasoning for it — that this is the one file emitting SQL without the `Fragment` pipeline, so it is the one most worth holding to the rule rather than excusing from it — is better than the way I put it.

812 unit tests pass.

### One follow-up, not a blocker: `set` is now N statements where one would do

The scoped delete issues one `DELETE` per visible link:

```ts
for (const row of visible) {
  await runPairStatement(executor, `delete from … where "A" = ? and "B" = ?`, …)
}
```

Replacing a wide set is therefore `1 select + 1 findMany + N deletes`, where it used to be one statement. Measured on SQLite it is invisible — 200 links replaced in 3ms — but SQLite round trips are in-process, and this repo counts statements precisely because Postgres round trips are not: `plans/orm/benchmarks.md` measures include trees by statement count, and the whole lateral strategy exists to remove `N - 1` of them.

A single `delete … where "A" = ? and "B" in (…)` expresses the same scoped set in one statement, and the `in` list is a shape the batched relation loader already builds. Worth noting the SQLite ceiling applies to both forms — the `findMany` above already binds one placeholder per existing link there, so a parent with more links than `maxBoundParameters` raises `ParameterLimitError` either way, which is the loud failure rather than the quiet one.

I would not hold the PR for it; flagging it because the correctness fix is what introduced the shape, and "why is `set` slow on Postgres" is a question better answered by a comment now than by a profile later.

nstfkc:
Took the follow-up rather than commenting it — pushed as 580b325. Your framing decided it: the correctness fix is what introduced the shape, and `1 select + 1 findMany + N deletes` makes `set` the one write that reintroduces the round trips the lateral strategy exists to remove. One `delete … where "A" = ? and "B" in (…)`.

On the binding: I built placeholders per length rather than using `dialect.inList`, and the comment says why — `inList` returns a `Fragment`, and this is the one path that emits SQL without that pipeline, because a join table has no model to route through. The cost is a server-side plan per distinct link count; the ORM's own plan cache is untouched since it keys on argument shape and this text never reaches it. Your note about the `maxBoundParameters` ceiling applying to both forms is right and unchanged — the `findMany` above already binds one per existing link.

### The single statement needed a guard the loop did not, and it is Postgres-only

When every existing link points at a row the caller cannot see, the target list is empty:

```
sqlite    delete from t where a in ()   accepted, deletes nothing
postgres  delete from t where a in ()   ERROR:  syntax error at or near ")"
```

A loop over an empty list ran zero times and said nothing; one statement has to check first. I found this because the new test for it **passed on SQLite with the guard removed** — so I checked both engines directly rather than trusting the green, and it is a real dialect split. Without the guard, on Postgres:

```
× set clears nothing when every existing link is hidden
  AssertionError: promise rejected "PostgresError: syntax error at or near ")""
```

That is the whole reason the suite runs twice, and it would have shipped green on the default dialect.

### One thing I introduced in the first commit and only found now

`Post`, `Tag` and `_PostToTag` were never added to the policy suite's truncation list, so the many-to-many tests leave rows behind — and `Tag.label` is unique. On SQLite every run gets a fresh temp database, so it is invisible. On Postgres the *second* run against the same database fails in a seed with `Tag_pkey`, three tests at once, pointing nowhere near the cause. It cost me two confusing runs before I recognised it as residue rather than a regression.

Fixed by truncating them, and verified the way the bug demands: the full Postgres suite run **twice against a single database**, 680 passed both times, identically.

838 unit tests, 48 policy tests, full template suite green on SQLite and Postgres 16 apart from the pre-existing `generated.test.ts` linkage probe.

nstfkc:
### Re-reviewed `580b325` — one statement, and the edge it introduced is handled

`set` is a single scoped delete now. Verified all three properties on SQLite, including the case the rewrite created:

```
mixed visible/hidden   [{A:1,B:1},{A:1,B:2}]        hidden link survives, visible one replaced
all existing hidden    [{A:2,B:3},{A:2,B:4},{A:2,B:5}]   no throw, nothing wrongly cleared
unpolicied set a->b    [{A:3,B:7}]                  old link goes, new one lands
unpolicied set []      []                           clears
```

The middle row is the one I would have gone looking for: with every existing link pointing at a row the caller cannot see, `targets` is empty and `in ()` is a syntax error on both dialects. The loop this replaced degenerated to zero iterations for free; the single statement has to say so, and it does. Guarding it *and* writing down why the previous shape did not need the guard is the difference between a fix and a fix that survives the next edit.

812 unit tests pass.

The note about building placeholders per length rather than reaching for `dialect.inList` is the right call to make explicitly. Worth recording what it costs, since you have already named it: on Postgres `= any($1)` would collapse every length to one server-side plan, which is exactly what `PostgresDialect.inList` exists for — but it returns a `Fragment`, and this is the one path with no pipeline to render one. Paying one plan per distinct link count to keep the join-table path free of that plumbing is a reasonable trade, and it is now a documented one rather than an accident.

Issue #84, opened 2026-07-28
ORM: `take` / `skip` are not validated, and a string `take` silently returns the wrong rows


nstfkc:
Branch: `feat/orm`. Finding 3 on #45, untracked until now. Raised again on #74, where I deferred it deliberately — the fix is a behaviour change to every read operation and wanted its own issue rather than being folded into an `aggregate` PR.

## The failure

`take` and `skip` reach `pagination()` in `packages/gemi/orm/compile/paginate.ts` unvalidated. The *sign* decides the SQL — a negative `take` means "the last N", which flips every ordering term — and the sign is read with `typeof take === "number" && take < 0`. A string never satisfies it:

```
{ take: "-2" }   ->  COMPILED  order=asc (first N)   bind=[2]     ✗ wrong rows, no error
{ take: "2"  }   ->  COMPILED  order=asc             bind=[2]
{ skip: "x"  }   ->  COMPILED                        bind=[-1, null]
{ take: 1.5  }   ->  COMPILED                        bind=[1.5]
```

The first line is the one that matters: `take: "-2"` asks for the **last** two rows and gets the **first** two, in the opposite order, with no error. `Math.abs(Number(...))` then binds `2`, so the statement is valid and the result is plausible.

A string `take` is not an exotic input — it is what a query string gives you (`?take=-2`), and the ORM's own argument types say `number`, so nothing between the request and the compiler has a reason to coerce it.

`skip: "x"` binds `null` as the offset and `take: 1.5` binds a fraction; both are quieter, but neither is a number the caller meant.

## Why it is not just a coercion

Coercing would fix the value and not the sign — `Number("-2")` is negative, so coercing early would make it work. That *is* a candidate fix. The other is to refuse: Prisma types these as `number` and rejects a string with `Argument take: Got invalid value`, so refusing matches Prisma and coercing is a silent superset.

The relation-node path already chose refusing: #72 added `assertPageArgument` for `take` / `skip` inside an `include`, which requires an integer and rejects a negative `skip`. So the codebase already has the validator and the decision — the root path simply does not use it.

## The cost of refusing

It is a breaking change for anyone passing `take: "10"` from a query string today, which works by coercion. That is the reason to do it as its own change with a line in `docs/orm.md`, rather than inside an unrelated PR.

## Acceptance

- A non-integer `take` or `skip` is refused at compile time on every read operation, with the same message shape `assertPageArgument` uses for relation nodes.
- A negative `skip` is refused, as it is on a relation node.
[...]

PR #86, opened 2026-07-28
A nested `include` is scoped as a read, whatever encloses it

### LINKED ISSUE #85: ORM: a nested `include` under `create` is scoped as an insert, dropping the child's read scope
Branch: `feat/orm`. Finding 1 on #45, untracked until now.

`Model.$exec` calls `applyNestedPolicies(schema, effective, op, …)` for every operation, and the walk passes the **root** operation down to each nested node. A nested `include` / `select` node is always a *read* — but under a root `create` it is scoped as if it were an insert, and `applyPolicies` takes its `INSERTING` branch.

Two outcomes, both wrong, and the quiet one is worse.

## (a) Child scopes with no `onCreate` → a hard failure naming an operation nobody asked for

```ts
// Account carries { scope: ctx => ({ organizationId: ctx.user.organizationId }) }
await Model.asUser(alice, () =>
  User.create({ data: { email }, include: { accounts: true } })
)
```

```
UnsupportedQueryError: gemi ORM does not support 'create' yet (Account.create).
Account has a policy that scopes reads but no 'onCreate', and create has no
where clause for a scope to narrow.
```

Nothing is being created on `Account` — it is being *read back* on the new row.

## (b) Child scopes **and** has `onCreate` → the read scope is silently dropped

The `INSERTING` branch `continue`s past `withScope`, then the `CREATING` pass runs `onCreate` over the include node, so the node comes out carrying a `data` key where its `where` should be:

```
findMany     include: {"accounts":{"where":{"organizationId":7}}}
create       include: {"accounts":{"data":{}}}            <- scope gone
softDeletes  include: {"accounts":{"data":{}}}            <- deletedAt: null gone
```

No error. A `softDeletes` scope disappearing from a nested read is precisely the class of thing this layer exists to make impossible.

The blast radius is small *today* because the included children belong to the row just created — but it is the same code path that would carry a real scope if `include` on writes ever widens, and (a) is a hard failure on an ordinary call today.

## Cause

The operation is threaded whole. A nested node needs to be scoped as the *read* it is, not as the root operation — the walk should pass a read operation down to `include` / `select` nodes regardless of what the root is doing. Nested **write** nodes under `data` are a separate path (`planNestedWrites`) and already scope themselves correctly, so this is only about the read tree.

## Acceptance

- `create` / `createMany` / `update` / `upsert` with an `include` of a policied relation applies the child's **read** scope, not its `onCreate`.
- The case in (a) stops raising.
- A `softDeletes` child is still filtered when included from a write.
- A test that fails if the scope is dropped — the (b) case is silent, so a test asserting only "it did not throw" would pass against the bug.

nstfkc:
Closes #85 — finding 1 on #45, which was untracked until I filed it this iteration along with #84.

## The bug

`applyNestedPolicies` passed the **root** operation down to every nested node. An `include` / `select` node is always a *read*, so under a root `create` the child was scoped as an insert and `applyPolicies` took its `INSERTING` branch. Two failures, and the second is the one that matters:

**(a) A child that scopes reads and declares no `onCreate` raised.**

```
UnsupportedQueryError: gemi ORM does not support 'create' yet (Account.create).
Account has a policy that scopes reads but no 'onCreate' …
```

That is a legal policy — a model can scope reads and never create rows — and nothing was being created on `Account`. It was being read back on the new row. The error named an operation the caller never asked for.

**(b) A child with both silently lost its read scope.**

The inserting branch skips `withScope`, then the creating pass runs `onCreate` over the include node, so it came out carrying `data: {}` where its `where` should have been:

```
findMany     include: {"accounts":{"where":{"organizationId":7}}}
create       include: {"accounts":{"data":{}}}            <- scope gone
softDeletes  include: {"accounts":{"data":{}}}            <- deletedAt: null gone
```

No error. A `softDeletes` scope vanishing from a nested read is precisely the class of thing this layer exists to make impossible.

## The fix

The recursion and the policy context both use `findMany` rather than the enclosing operation. Nested *write* nodes under `data` are a separate path (`planNestedWrites`) and already scope themselves as writes — this walk only ever sees the read tree, so there is no case where the root operation was the right answer.

`findMany` is also in `SCOPABLE`, which an inserting operation is not; that mismatch is what produced the bug, and the constant carries the reason.

## Testing the silent half was the hard part

Under a root `create` the included children belong to a row that has **just been made**, so they come back empty whether the scope was applied or dropped. A test there passes against the bug — which is why #85's own acceptance criterion says so.

`upsert` is in the same inserting branch and can land on a row that **already has children**. So the discriminating test seeds two accounts on one user, in two different organisations, and upserts with an include: without the fix both come back, with it one.

All three tests were verified by reverting the two lines and watching them fail — 3 failed / 45 passed against the bug, 48 passed with it. The same discipline as the `LITERAL_KEYS` guard on #78 and the `satisfies` map on #82: a check that has not been seen to go red is not a check.

## Also filed this iteration

**#84** — `take` / `skip` are not validated, and `take: "-2"` from a query string silently returns the *first* two rows ascending instead of the last two. That is finding 3 on #45, which I deferred on #74 because the fix is a behaviour change to every read operation; it now has an issue rather than depending on either of us remembering it.

838 unit tests, 48 policy tests, full template suite green on SQLite and Postgres 16.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -468,12 +468,21 @@ a shared model base can impose a policy a subclass can only narrow, never drop.
 | `before(ctx)` | First | Return `false` (or throw) to deny outright. |
 | `scope(ctx)` | Reads and row-matching writes | Returns a `where` fragment `AND`ed into `args.where`. A policy can only ever **narrow**. |
 | `onCreate(ctx, data)` | `create` / `createMany` / `upsert` | Defaults or validates the payload. An insert has no `where` for a scope to narrow. |
-
-A relation reached through `include` or `select` is always scoped as a **read**, whatever the
-statement around it is doing — so `User.create({ data, include: { accounts: true } })` applies
-`Account`'s `scope`, not its `onCreate`. The children are being read back, not written.
 | `onUpdate(ctx, data)` | `update` / `updateMany` / `upsert` | Defaults or validates the payload of a write to existing rows. |
 | `redact(ctx, row)` | Shaping | Removes fields from a fetched row. |
+
+**Anything that names another model is scoped as a read of that model**, whatever the statement
+around it is doing. That covers all four ways a query reaches a model it did not name — an
+`include` / `select` node, a `_count` entry, a relation filter in a `where`, and a relation
+ordering in an `orderBy` — and it means `context.operation` is `findMany` inside each of them:
+
+```ts
+User.create({ data, include: { accounts: true } })      // Account's scope, not its onCreate
+User.updateMany({ where: { accounts: { some: {} } } })  // Account's read scope, not its write one
+```
+
+The children are being *read back*; only the row named by the statement itself is being written.
+Nested writes under `data` are a different tree and scope themselves as writes.
 
 **A scope alone does not protect writes, and the ORM says so rather than assuming.** A `scope`
 narrows *which rows* a statement may touch and says nothing about the values written, so a
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -468,6 +468,10 @@ a shared model base can impose a policy a subclass can only narrow, never drop.
 | `before(ctx)` | First | Return `false` (or throw) to deny outright. |
 | `scope(ctx)` | Reads and row-matching writes | Returns a `where` fragment `AND`ed into `args.where`. A policy can only ever **narrow**. |
 | `onCreate(ctx, data)` | `create` / `createMany` / `upsert` | Defaults or validates the payload. An insert has no `where` for a scope to narrow. |
+
+A relation reached through `include` or `select` is always scoped as a **read**, whatever the
+statement around it is doing — so `User.create({ data, include: { accounts: true } })` applies
+`Account`'s `scope`, not its `onCreate`. The children are being read back, not written.
 | `onUpdate(ctx, data)` | `update` / `updateMany` / `upsert` | Defaults or validates the payload of a write to existing rows. |
 | `redact(ctx, row)` | Shaping | Removes fields from a fetched row. |
 
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -506,7 +506,6 @@ export abstract class Model {
       effective = applyNestedPolicies(
         schema,
         effective,
-        op,
         currentUser(),
         system,
         (model) => {
--- a/packages/gemi/orm/compile/order-relation.test.ts
+++ b/packages/gemi/orm/compile/order-relation.test.ts
@@ -200,7 +200,6 @@ describe("policies", () => {
     const out = applyNestedPolicies(
       root,
       { orderBy: { accounts: { _count: "desc" } } },
-      "findMany" as any,
       { organizationId: 7 },
       false,
       lookup([tenant]),
@@ -216,7 +215,6 @@ describe("policies", () => {
     const out = applyNestedPolicies(
       root,
       { orderBy: [{ email: "asc" }, { accounts: { _count: "desc" } }] },
-      "findMany" as any,
       { organizationId: 7 },
       false,
       lookup([tenant]),
@@ -232,7 +230,6 @@ describe("policies", () => {
       applyNestedPolicies(
         root,
         args,
-        "findMany" as any,
         { organizationId: 7 },
         true,
         lookup([tenant]),
@@ -243,7 +240,7 @@ describe("policies", () => {
   test("an unpolicied ordering is returned unchanged, identically", () => {
     const args = { orderBy: { accounts: { _count: "desc" } } };
     expect(
-      applyNestedPolicies(root, args, "findMany" as any, {}, false, lookup([])),
+      applyNestedPolicies(root, args, {}, false, lookup([])),
     ).toBe(args);
   });
 
--- a/packages/gemi/orm/compile/relation-count.test.ts
+++ b/packages/gemi/orm/compile/relation-count.test.ts
@@ -229,7 +229,6 @@ describe("policies", () => {
     const out = applyNestedPolicies(
       root,
       { include: { _count: { select: { accounts: true } } } },
-      "findMany" as any,
       { organizationId: 7 },
       false,
       lookup([tenant]),
@@ -248,7 +247,6 @@ describe("policies", () => {
           _count: { select: { accounts: { where: { organizationRole: 1 } } } },
         },
       },
-      "findMany" as any,
       { organizationId: 7 },
       false,
       lookup([tenant]),
@@ -265,7 +263,6 @@ describe("policies", () => {
     const out = applyNestedPolicies(
       root,
       args,
-      "findMany" as any,
       { organizationId: 7 },
       true,
       lookup([tenant]),
@@ -280,7 +277,6 @@ describe("policies", () => {
     const out = applyNestedPolicies(
       root,
       args,
-      "findMany" as any,
       {},
       false,
       lookup([]),
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -658,7 +658,6 @@ describe("applyNestedPolicies", () => {
     const out = applyNestedPolicies(
       relations({ accounts: "Account" }),
       { include: { accounts: true } },
-      "findMany" as any,
       USER,
       false,
       lookup({ Account: [tenant] }),
@@ -672,7 +671,6 @@ describe("applyNestedPolicies", () => {
     const out = applyNestedPolicies(
       relations({ accounts: "Account" }),
       { include: { accounts: { where: { organizationRole: 0 } } } },
-      "findMany" as any,
       USER,
       false,
       lookup({ Account: [tenant] }),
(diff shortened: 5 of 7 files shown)


nstfkc:
## Review — nested `include` scoped as a read (closes #85)

Read the full diff against the merge base (`9b20919`). 838 unit tests pass.

The fix is the right one, the constant carries its reasoning, and the argument that "an `include` / `select` node is a read whatever the statement around it is doing" is the correct general statement — `NESTED_READ` being in `SCOPABLE` where an inserting operation is not is the mismatch that produced the bug, and naming that in the constant is what stops it being re-derived.

**The `upsert` test is the part worth calling out.** I raised this finding without a discriminating test for the silent half, precisely because under a root `create` the included children belong to a row that has just been made, so they come back empty whether the scope applied or not. Reaching for `upsert` — same inserting branch, but it can land on a row that already has children — is the move that makes the quiet half observable, and seeding two accounts in two organisations is what makes it discriminate. Verifying all three by reverting the two lines and watching them go red (3 failed / 45 passed against the bug) is the standard I would want applied to a fix for a silent bug.

### The same bug is still live one branch over, in `_count`

`applyNestedPolicies` handles three kinds of node in that loop, and only the `include` / `select` recursion was changed. The `_count` branch still passes the root operation:

```ts
const scopedCounts = scopeCounts(
  schema,
  (tree as Record<string, unknown>)[key],
  operation,        // <- still the root's
  …
);
```

`scopeCounts` then calls `applyPolicies` with a `policyContext` built from it, so under a root `create` a `_count` node takes the same `INSERTING` branch. Both halves reproduce — same tree, same policies, only the node type differs:

```
include, scope-only     [findMany]  {"accounts":{"where":{"organizationId":7}}}
include, scope-only     [create]    {"accounts":{"where":{"organizationId":7}}}   ✓ fixed
include, scope+onCreate [create]    {"accounts":{"where":{"organizationId":7}}}   ✓ fixed

_count,  scope-only     [findMany]  {"_count":{"select":{"accounts":{"where":{…}}}}}
_count,  scope-only     [create]    THREW  Account has a policy that scopes reads but no 'onCreate'
_count,  scope+onCreate [create]    {"_count":{"select":{"accounts":{"data":{}}}}}   <- scope gone
```

`User.create({ data, include: { _count: { select: { accounts: true } } } })` against a read-scoped `Account` raises an error about creating an `Account`, which is the loud half verbatim; with an `onCreate` present the scope is replaced by `data: {}`, which is the quiet half verbatim.

The justification transfers without modification — `policy.ts`'s own comment on `scopeCounts` already says it: *"`_count` counts another model's rows, so it is a read of that model and carries its policies… An unscoped count is the quietest of the three: it returns a number, so what leaks is how many rows exist in tenants the caller cannot see."* A `_count` node is as much a read as an `include` node is.

**Same one-word fix, verified:** passing `NESTED_READ` to `scopeCounts` as well makes all three `_count` rows match their `include` counterparts, and 812 tests still pass.

Two notes on scoping the change:

- `scopeRelationOrderings` and `scopeRelationFilters` are called with the root operation too, but neither is reachable from an inserting operation — `WRITE_ARGS.create` accepts no `where` or `orderBy`, and `createMany` accepts no `include` at all. So they are not part of this bug. Worth a sentence saying so rather than leaving the asymmetry to be rediscovered, since the next reader will notice three call sites and one change.
- The practical exposure of the `_count` case is bounded the same way the `include` case was: under `create` the count is of a new row's children. The loud half is the one that bites immediately, and it bites on a legal policy.

Given the fix already exists in this PR, folding `_count` into it seems better than a follow-up — and the `upsert` technique that made the `include` half testable does not transfer, since `upsert` is not in `INSERTING` and a `_count` there hits `assertScopable`'s refusal instead. A unit-level assertion on `applyNestedPolicies`, of the shape above, is probably the honest test for it.

Everything else reads correctly. Filing #84 rather than folding it in was the right call for the same reason it was deferred on #74.

nstfkc:
Pushed as 8eee549. The `_count` finding is right and reproduces exactly as your table shows. I did not take the one-word fix, and the reason is the rest of your review.

**Your two notes on scoping the change turned out to be the finding.** You asked for a sentence saying why `scopeRelationOrderings` and `scopeRelationFilters` are not part of this bug. I went to write it and could only make it true of *inserting* operations. They are unreachable from an insert for the reason you gave — `create` accepts only `data` / `select` / `include`, `createMany` only `data`, so there is no `where` or `orderBy` for a relation to be named in — but they are reachable from `update` and `delete`, and there they were wrong:

```
root      include                     _count                     where filter             orderBy
update    seenOperation: "findMany"   seenOperation: "update"    seenOperation: "update"  seenOperation: "update"
delete    seenOperation: "findMany"   seenOperation: "delete"    seenOperation: "delete"  seenOperation: "delete"
```

(A `scope` that returns `{ seenOperation: context.operation }`, so the walk reports what it was told.) A policy is entitled to answer `scope` differently per operation — that is what the parameter is for — so one that narrows reads and not writes leaves that filter unscoped under a root `update`. Quieter than the `_count` case, and the same bug.

**So the fix is that the walk no longer takes an operation.** Every node it touches is a read; the parameter was a slot with one legal value and four opportunities to fill it wrongly, and two were taken. Passing `NESTED_READ` to `scopeCounts` fixes the third and leaves the fourth, plus every future one. This is #79's argument applied to the same shape of parameter, which is why I would rather not repeat the pattern of fixing the call site: the bug this parameter exists to prevent *is* a wrong operation, so not accepting one is the version that cannot regress.

Verified against the pre-fix code: three of the new unit test's four assertions fail, and the fourth — `include` — passes, which is the shape of what was already fixed.

### One correction to your suggestion, and it is the interesting part

> the `upsert` technique that made the `include` half testable does not transfer, since `upsert` is not in `INSERTING` and a `_count` there hits `assertScopable`'s refusal instead

It hits the refusal, yes — but the technique fails for a different reason, and it took building the test to find it. **`_count` in a write's `include` is never compiled at all.** `planRelationCounts` is called from `compile/read.ts` and nowhere else, so the statement runs, the row comes back, and the key is simply absent:

```
include: { nosuchrelation: true }   ->  UnknownRelationError: 'nosuchrelation' is not a relation
                                        on model User. Known relations: organization, accounts
include: { _count: { … } }          ->  accepted, dropped, no error
```

The write path validates relation names in an `include` and then discards the one key that is not one. Filed as #87 with the two options.

That changes the honest account of the `_count` half of your finding, so it is worth being precise: the **loud** half is real and reachable today — a `create` with a read-scoped child raises about an operation nobody asked for. The **quiet** half is real as the walk's *output* and has no consumer, because nothing downstream reads that node for a write. It becomes a live leak the day #87 is implemented. I would rather fix it while it is latent than discover it is the reason a count is too high.

The test therefore asserts the loud half and says in its comment why it cannot assert the number, rather than asserting something weaker and looking complete.

### One thing your review did not catch, in my own commit

The doc paragraph I added landed **between two rows of the policy-members table**, splitting it — `onUpdate` and `redact` were orphaned below a paragraph. Moved below the table and widened to state the rule over all four node types.

839 unit tests, 50 policy tests, full template suite green on SQLite and Postgres 16, with only the pre-existing `generated.test.ts` generator-linkage probe failing (fails on `HEAD` too).

nstfkc:
### Re-reviewed `8eee549` — a better fix than the one I suggested, and a correction to my analysis

I proposed threading `NESTED_READ` to the `_count` call site. Removing the parameter from the walk entirely is the stronger answer, and the reasoning in the constant is the right generalisation:

> Threading the right value through four call sites leaves the wrong value expressible at all four. Not accepting one at all is what makes it unrepresentable.

That is the same move as #79's required `operation`, arrived at from the opposite direction — there the fix was to *demand* the argument, here to *refuse* it, and both because the wrong value was silently acceptable. All four node kinds now scope as reads:

```
include              {"accounts":{"where":{"organizationId":7}}}
include + onCreate   {"accounts":{"where":{"organizationId":7}}}
_count               {"_count":{"select":{"accounts":{"where":{"organizationId":7}}}}}
_count  + onCreate   {"_count":{"select":{"accounts":{"where":{"organizationId":7}}}}}
relation filter      {"accounts":{"some":{"AND":[{"organizationId":7}]}}}
relation ordering    {"accounts":{"_count":"desc","where":{"organizationId":7}}}
```

839 unit tests pass.

**And you were right to go further than I did.** I wrote that relation filters and orderings "are not part of this bug" because they are unreachable from an insert. That was the wrong test — unreachable from the *failing* case is not the same as correct. Under a root `update` a policy reading `context.operation` saw `"update"` for what is a read of another model, and answered a question it was not asked. Recording the reachability asymmetry as the reason it was invisible, rather than as a reason to leave it, is the version I should have written.

**A correction to my own framing, prompted by #87.** I said both halves of the bug reproduced "verbatim" for `_count`. The loud half did — the spurious `Account has a policy that scopes reads but no 'onCreate'` under a root `create` is real and user-facing. The quiet half is narrower than I implied, because `planRelationCounts` is only called from `read.ts`, so a `_count` in a write's `include` never compiles at all:

```
create + _count include ->  _count = undefined
read   + _count include ->  _count = {"accounts":0}
create + bad relation   ->  THREW 'nosuchrelation' is not a relation on model User
```

So the scope the walk computed for that node was being discarded downstream regardless. The fix is still necessary — the throw was the reachable failure — but "the child's read scope silently disappeared" overstated the consequence for `_count` specifically, where the node itself disappears. Filing that as #87 rather than folding it in is the right split, and the inconsistency it names is the sharper complaint: the write path validates relation names in an `include` and then drops the one key that is not one.

PR #88, opened 2026-07-28
`take` and `skip` are validated, at the one place that reads them

### LINKED ISSUE #84: ORM: `take` / `skip` are not validated, and a string `take` silently returns the wrong rows

nstfkc:
Closes #84.

## The failure

`take` and `skip` reached `pagination()` unvalidated, and they are the only two arguments whose **sign decides the SQL** rather than a bound value — a negative `take` means "the last N", which flips every ordering term. The sign is read as `typeof take === "number" && take < 0`, so a string does not merely arrive in the wrong type. It takes the other branch:

```
{ take: -2 }    order flipped, bind 2      the last two rows
{ take: "-2" }  order as written, bind 2   the *first* two rows
```

`Math.abs(Number(...))` binds `2` either way, so the statement is valid, the page is the right size, and the rows are wrong. Nothing raises. `?take=-2` from a query string is exactly where a string `take` comes from, and the generated types say `number`, so nothing in between has a reason to coerce it.

## The fix

Refused, not coerced. Coercing would fix that one case — `Number("-2")` is negative — and go on accepting `take: "abc"`, which binds `NaN`. #72 already chose refusal for the same two arguments on a *relation* node; this is that rule reaching the root, where it should have been first.

Two things about where it went:

- **`assertPageArgument` moved to `paginate.ts` and `plan-relations.ts` now calls it.** One function, not two with the same body. #84 *was* the two paths disagreeing; fixing it with a second copy is how they drift again. The only difference is the name reported — `take` at the root, `accounts.take` on a node — and that is a parameter.
- **The check runs inside `pagination()`,** not in each caller's argument allowlist. Three operations page — `findMany` and friends, `count`, `aggregate` — across two files, so putting it at the one place that *reads* these arguments means a fourth pager cannot be added without it. Same reasoning as invariant 1.

## A fraction is a deliberate divergence, and measuring it decided the direction

#84 assumed Prisma rejects a fractional `take`. **It does not** — it truncates toward zero. And binding one, which is what this did, has no single meaning to match:

| | `take: 1.5` |
| --- | --- |
| Prisma 6 | 1 row — truncates |
| SQLite, `limit ? = 1.5` | `SQLiteError: datatype mismatch` (`SQLITE_MISMATCH`) |
| Postgres, `limit $1 = 1.5` | 2 rows — rounds to nearest |

Read off a live Prisma client and both engines through Bun, not reasoned about. So the status quo is an opaque driver error on one dialect and a silent disagreement with Prisma *and* with the other dialect on the second — one argument, three behaviours, no line naming `take`.

Matching Prisma would mean truncating here **and** relaxing #72's relation-node rule, which refuses fractions today. Refusing keeps one rule at both levels and fails in the loud direction. `docs/orm.md` records it as a divergence rather than leaving it to be discovered.

## Tests

Seven, and every one verified **red against the unvalidated code** before being kept:

```
× a string take is refused, rather than silently paging the wrong end
× a fractional take is refused, which Prisma would have truncated
× a non-numeric skip is refused rather than bound as null
× a negative skip is refused, as it already was on a relation node
× the refusal names the operation that was asked for
× a single-row operation refuses it too, though it ignores the value
× a take that is not an integer is refused here too        (aggregate)
```

The string test also pins the **shape** of the bug — that a real `-2` flips the order where `"-2"` did not — so a future coercion cannot reintroduce it quietly. `findFirst` gets its own test because it pins `take` to 1 and never reads the caller's value: it is the one operation where the argument could have been ignored instead of refused, and refusing what was *written* rather than what survived is what keeps the rule one sentence long.

## Breaking

For anyone passing `take: "10"` from a query string today, which works by coercion. That is why #84 asked for it as its own change with a documented line rather than folded into #74.

## Verification

889 unit tests. Full template suite green on SQLite and Postgres 16, with only the pre-existing `generated.test.ts` generator-linkage probe failing (fails on the base branch too). `tsc` clean; oxlint clean apart from two pre-existing warnings in `where.ts`.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -117,6 +117,25 @@ const user = await User.findUniqueOrThrow({
 })
 ```
 
+**`take` and `skip` must be integers, and are refused rather than coerced.** They are the only
+arguments whose *sign* decides the SQL — a negative `take` means "the last N", which flips every
+ordering term — so a `take` arriving as a string does not merely have the wrong type, it takes the
+wrong branch: `take: "-2"` used to return the **first** two rows, in the opposite order, with no
+error. A query string is exactly where a string `take` comes from, so parse it before you pass it:
+
+```
+UnsupportedQueryError: gemi ORM does not support 'take' yet (User.findMany).
+Expected an integer, got "-2".
+```
+
+A negative `skip` is refused too — it counts rows to pass over — and both rules are the ones that
+already applied inside an `include`.
+
+> **Divergence from Prisma, on purpose.** Prisma accepts a *fractional* `take` and truncates toward
+> zero; this refuses it. Binding one instead is a `datatype mismatch` error from SQLite and a
+> silently different row count on Postgres, which rounds — so the fraction has no single meaning to
+> match. One rule at both levels, failing loudly, beats three behaviours.
+
 `select` and `include` narrow the **return type**, exactly as they do in Prisma — `user.email`
 type-checks, `user.name` does not. A key outside the operation's argument type collapses to
 `never` rather than being quietly accepted.
--- a/packages/gemi/orm/compile/aggregate.test.ts
+++ b/packages/gemi/orm/compile/aggregate.test.ts
@@ -155,6 +155,22 @@ describe("pagination", () => {
     expect(compiled.text).toContain(`order by "globalRole" desc`);
     expect(compiled.bind(args)).toEqual([2]);
   });
+
+  /**
+   * #84 reaches here because `aggregate` pages, and it validates its arguments
+   * in a different file from `read.ts`. The check lives in `paginate.ts` — the
+   * one place that *reads* these two — rather than in each caller's allowlist,
+   * so this passes without `aggregate.ts` knowing the rule exists.
+   */
+  test("a take that is not an integer is refused here too", () => {
+    expect(() => plan({ take: "-2", _count: true })).toThrow(
+      /Expected an integer, got "-2"/,
+    );
+    expect(() => plan({ skip: -1, _count: true })).toThrow(
+      /'skip' counts rows to pass over/,
+    );
+    expect(() => plan({ take: "-2", _count: true })).toThrow(/User\.aggregate/);
+  });
 });
 
 /**
--- a/packages/gemi/orm/compile/aggregate.ts
+++ b/packages/gemi/orm/compile/aggregate.ts
@@ -2,7 +2,14 @@ import type { SqlDialect } from "../dialect";
 import { UnknownFieldError, UnsupportedQueryError } from "../errors";
 import type { QueryPlan } from "../plan";
 import type { FieldSchema, ModelSchema, ScalarType } from "../schema";
-import { type Fragment, bindValues, concat, joinFragments, render, sql } from "./fragment";
+import {
+  type Fragment,
+  bindValues,
+  concat,
+  joinFragments,
+  render,
+  sql,
+} from "./fragment";
 import { compileOrderBy, parseOrderBy } from "./orderBy";
 import { pagination } from "./paginate";
 import { compileWhere } from "./where";
@@ -132,11 +139,16 @@ export function compileAggregate(
       flat
         ? `A 'select' on a count must name at least one field, or '_all'.`
         : `An aggregate needs at least one of ${KINDS.join(", ")}. Prisma ` +
-          `rejects an empty one too.`,
+            `rejects an empty one too.`,
     );
   }
 
-  const where = compileWhere(schema, args?.where, { dialect, operation: op }, (a) => a?.where);
+  const where = compileWhere(
+    schema,
+    args?.where,
+    { dialect, operation: op },
+    (a) => a?.where,
+  );
   const whereClause = where ? concat(sql(" where "), where) : sql("");
   const from = sql(` from ${dialect.quoteIdent(schema.table)}`);
 
@@ -164,7 +176,13 @@ export function compileAggregate(
   if (!paginating) {
     statement = concat(sql("select "), projection, from, whereClause);
   } else {
-    const { clause, terms: order } = pagination(schema, args, dialect, parsed);
+    const { clause, terms: order } = pagination(
+      schema,
+      op,
+      args,
+      dialect,
+      parsed,
+    );
     const ordering = compileOrderBy(order, dialect);
 
     // The subquery selects every column an aggregate names — an `order by` may
@@ -213,7 +231,8 @@ export function compileAggregate(
 
       if (flat) {
         const out: Record<string, unknown> = {};
-        for (const term of terms) out[term.as] = coerce(term, row[term.alias], dialect);
+        for (const term of terms)
+          out[term.as] = coerce(term, row[term.alias], dialect);
         return out;
       }
 
@@ -236,11 +255,7 @@ export function compileAggregate(
 /** The `as` a bare `_count: true` carries, which no field name can be. */
 const BARE = "";
 
-function aggregateTerms(
-  schema: ModelSchema,
-  op: string,
-  args: any,
-): Term[] {
+function aggregateTerms(schema: ModelSchema, op: string, args: any): Term[] {
   const terms: Term[] = [];
 
   for (const kind of KINDS) {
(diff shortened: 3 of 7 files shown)


nstfkc:
## Review — `take` / `skip` validation (closes #84)

Read the full diff against the merge base (`8825c87`). 889 unit tests pass.

This closes a finding I raised on #45 and then repeated on #74, so I checked it with my own reproduction rather than the PR's cases:

```
findMany take:-2      OK       order="id" desc      the legitimate "last N" still flips
findMany take:"-2"    REFUSED  Expected an integer, got "-2"
findMany take:"abc"   REFUSED
findMany take:1.5     REFUSED
findMany skip:-1      REFUSED  A 'skip' counts rows to pass over…
findMany skip:"2"     REFUSED
findMany take:0       OK
findFirst take:"x"    REFUSED  (User.findFirst)
count     take:"x"    REFUSED  (User.count)
aggregate take:"x"    REFUSED  (User.aggregate)
include accounts take:"x"  REFUSED  'accounts.take'
```

All four pagers, the relation node reporting its own name through the shared function, and the negative-`take` behaviour the fix could plausibly have broken still intact.

Three decisions are the right ones and worth naming:

- **Refuse rather than coerce.** Coercing fixes `"-2"` and keeps accepting `"abc"`, which binds `NaN`. The generated types already say `number`; the failure was that nothing checked, not that the contract was unclear.
- **One `assertPageArgument`, moved to `paginate.ts`.** #84 *was* two paths disagreeing about the same two arguments; fixing it with a second copy is how they drift again. Making the reported name a parameter is what let the two levels share it.
- **The check lives where the arguments are read, not in each allowlist.** Three operations page across two files, and a fourth pager now cannot be added without it. That is the same argument as invariant 1 and it is the reason this is a fix rather than three fixes.

**And you corrected an assumption I had put in the issue.** #84 says Prisma rejects a fractional `take`; it truncates. Measuring it across Prisma, SQLite and Postgres — `1 row` / `datatype mismatch` / `2 rows` — is what turns "match Prisma" from an obvious answer into a real choice, and refusing to keep one rule at both levels is the right call given #72 already refuses fractions on a relation node. Recording it in `docs/orm.md` as a divergence rather than leaving it to be found is the other half.

### One thing: the message says "yet" for a refusal this PR documents as permanent

`UnsupportedQueryError` hardcodes it:

```
gemi ORM does not support 'take' yet (User.findMany). Expected an integer, got "-2".
```

For a string `take` that is merely odd — there is no future in which `take: "-2"` is supported, so "yet" promises something nobody intends. For the *fraction* it is a contradiction with this PR's own documentation, which calls it:

> **Divergence from Prisma, on purpose.**

That word became load-bearing across #78 and #82: `distinct` and `cursor` say "a decision rather than a gap" through `UnsupportedByDesignError`, genuinely unimplemented arguments say "yet", and #82 took "yet" out of the dialect message for exactly this reason — it was a supported-matrix statement rather than a promise. A validation refusal is a fourth category again: not unimplemented, not a design refusal of a *feature*, but invalid input.

`UnsupportedByDesignError` is not on this branch — it comes from #78, still open — so this is a merge-order note rather than a change to make here. Whichever lands second, the fraction refusal at minimum belongs in the by-design category it is already documented as, and the type-mismatch messages would read better without a word that implies a plan.

Worth noting for the same merge: #78 also moves `paginate.ts`'s neighbours around, and this PR changes `pagination()`'s signature by inserting `operation` as the second parameter. Both touch `read.ts`'s call site.

Nothing blocking.

nstfkc:
Pushed as 2a987d2 — a comment, not a fix, and the reasoning for stopping there is worth stating.

**You are right about "yet", and it is worse than odd on the fraction.** That refusal is documented in this very PR as *"Divergence from Prisma, on purpose"*, and the message it raises says the opposite. Given how load-bearing the word has become — #78's `UnsupportedByDesignError` for "a decision rather than a gap", #82 taking it out of the dialect message — a validation refusal reading "yet" is not a wording slip, it is the wrong category.

I did not fix it here, and your own framing is why: validation is a **fourth** category, so the fix is a category rather than a reworded string. Inventing one on this branch while #78 is in flight would be the second copy `assertPageArgument` exists to avoid — the same mistake, one level up. So what I pushed is a comment beside the refusal saying exactly what changes when #78 merges: the fraction refusal moves to by-design, and the type-mismatch messages lose the word. The second merge should not have to rediscover that.

**On the merge-order warning — I checked it rather than noting it, and it does not bite.** The two branches touch four files in common:

```
docs/orm.md
packages/gemi/orm/compile/plan-relations.ts
packages/gemi/orm/compile/read.test.ts
packages/gemi/orm/compile/read.ts
```

and `git merge-tree` reports **zero conflict hunks** in any of them. #78 changes `READ_ARGS` and the `omit` handling in `read.ts`; it does not touch `pagination()`'s call site, which is where my signature change lands. Either order merges clean — so the only thing to carry across is the category, not the code.

Thank you for reproducing it independently rather than reading my cases. The `include accounts take:"x"` row is the one I would have wanted checked most: the whole argument for moving `assertPageArgument` into `paginate.ts` was that the relation node keeps reporting its own name through the shared function, and that only holds if the parameterised name actually survives the move.

889 unit tests, full template suite green on SQLite and Postgres 16 apart from the pre-existing `generated.test.ts` linkage probe.

nstfkc:
### Re-reviewed `2a987d2` — the right call, and the deferral checks out

Comment-only (0 non-comment lines added), 889 tests pass. Recording the decision next to the refusal rather than inventing a category while #78 is in flight is the right handling — my finding was itself a merge-order note, and "the fix is a category, not a reworded string" is the accurate reading of it.

I verified the load-bearing claim rather than taking it, since the deferral rests on it:

```
git merge-tree --write-tree feat/orm-omit(5d893df) feat/orm-page-validation(2a987d2)
  ->  MERGE CLEAN
```

Worth knowing it is head-specific: at #78's *previous* head (`2d041ab`, before it merged `feat/orm`) the same test conflicts in `read.ts` and `docs/orm.md`. So the claim is true now and became true when #78 rebased — if either branch moves again it is worth re-running, which is one command.

Issue #89, opened 2026-07-28
SQLite runs with `foreign_keys = OFF`, so no referential action ever fires


nstfkc:
Found while adding differential coverage for #87 — the first case in the harness that compares a *child* table after a parent write.

## The failure

Bun's SQLite driver leaves the pragma at its SQLite default, and nothing in gemi turns it on:

```
bun sqlite default foreign_keys -> [{"foreign_keys":0}]
```

Prisma enables it for every SQLite connection. So on SQLite, gemi:

- never fires `ON DELETE SET NULL` or `ON DELETE CASCADE`;
- never raises on a violated foreign key;
- happily writes a row referencing a parent that does not exist.

The migrations declare all of it — `Account_userId_fkey ... ON DELETE SET NULL`, `SocialAccount_userId_fkey ... ON DELETE RESTRICT` — and none of it is enforced.

Measured through the harness. Deleting a user with two accounts:

```
                       Account.userId after
prisma (fk on)         null, null
gemi   (fk off)        1, 1            <- rows now point at a user that is gone
```

Both clients, one database, one statement each.

## Why it went unnoticed

The differential harness compares the *written* model's table by default (`tables ?? [model]`). Every existing delete case compares `User` alone, so the orphaned children were never read back. It is not a gap in the comparison logic — it is that no case had ever asked.

## Why it matters beyond tidiness

- **`onDelete: Cascade` silently does nothing.** An application that relies on it to remove dependent rows — sessions, tokens, memberships — accumulates them forever in development and cleans them up correctly in production, which is the wrong way round for a bug to be discovered.
- **A restrict is not a restrict.** `SocialAccount_userId_fkey` is `ON DELETE RESTRICT`; on SQLite the delete succeeds anyway.
- **Development and production disagree**, and the template ships SQLite as the development dialect and Postgres as the production one. That is the exact split the differential harness exists to close.

It is also a *dialect* divergence inside gemi: Postgres enforces these and SQLite does not, so the same code has two behaviours.

## The fix

`pragma foreign_keys = ON` on every SQLite connection, at the point the driver hands one over — it is per-connection, not per-database, so it has to be set on each. Postgres needs nothing.

## Worth deciding alongside

Enabling it will make currently-passing writes start failing where they violate a constraint the schema declares — which is correct, and is a behaviour change. Seeds and test fixtures that write children before parents are the usual casualties.

## Acceptance

- A SQLite connection reports `foreign_keys = 1`.
[...]

PR #90, opened 2026-07-28
`_count` in a write's `include` is projected, not dropped

### LINKED ISSUE #87: `_count` in a write's `include` is silently dropped
`planRelationCounts` is called from `compile/read.ts` and nowhere else, so a `_count` entry in the `include` of a `create` / `update` / `delete` / `upsert` is never compiled. The statement runs, the row comes back, and the `_count` key the caller asked for is simply absent:

```ts
const created = await User.create({
  data: { email: "a@x.test" },
  include: { _count: { select: { accounts: true } } },
});

created._count        // undefined — no error, no warning
```

Prisma returns `{ accounts: 0 }` here.

**The reason it is worth a ticket rather than a shrug is the inconsistency.** The same `include`, one key over, is checked:

```
include: { nosuchrelation: true }   ->  UnknownRelationError: 'nosuchrelation' is not a
                                        relation on model User. Known relations: organization,
                                        accounts
include: { _count: { … } }          ->  accepted, dropped, no error
```

So the write path validates relation names in an `include` and then discards the one key that is not a relation name. A caller who reads the result gets `undefined` and, in TypeScript, whatever the generated type says — which is the shape Prisma would have returned.

Found while fixing #85's `_count` counterpart (PR #86). It matters there for a second reason: the policy walk *does* scope `_count` nodes under a write, so today the scope it computes has no consumer. The moment this is implemented, that scope becomes load-bearing — and the version of it that existed before #86 wrote `data: {}` over the child's `where`. The policy half is fixed; this is the half that would have made it visible.

Two things to decide:

1. **Support it** — call `planRelationCounts` from the write path's returning projection, the same way `compile/read.ts` does. This is the Prisma-compatible answer and the correlated subqueries are already built.
2. **Refuse it** — `UnsupportedQueryError` naming `_count`, consistent with how every other unimplemented argument is handled in this ORM. Cheap, honest, and leaves 1 open.

Either is better than the current silence. Refusing first and supporting later is a legal sequence; the reverse is not.

nstfkc:
Closes #87.

## The failure

`planRelationCounts` was called from `compile/read.ts` and nowhere else, so a `_count` in the `include` of a `create` / `update` / `delete` / `upsert` was never compiled. The statement ran, the row came back, and the key the caller asked for was absent — with no error:

```
include: { nosuchrelation: true }   ->  UnknownRelationError: 'nosuchrelation' is not a relation
                                        on model User. Known relations: organization, accounts
include: { _count: { … } }          ->  accepted, dropped, no error
```

The write path validated relation names in an `include` and then discarded the one key that is not one.

## Projected, not refused

Both options were on the ticket. Projecting, because both dialects evaluate a correlated subquery inside `RETURNING` — measured through Bun against SQLite and Postgres 16 — so the same plan `compileRead` builds goes straight into the returning list and costs **no second round trip**. The shaper assembles `_count` from the raw row exactly as the read path does, deliberately spelled the same way rather than abstracted.

## `delete` is where the dialects disagree

This is the part that needed more than plumbing. Under `on delete cascade`:

| | `delete … returning (select count(*) from ch …)` |
| --- | --- |
| Postgres | **3** — pre-statement snapshot, which is what Prisma returns |
| SQLite | **0** — evaluated after the cascade |

So on SQLite the count would come back as a *number*, just the wrong one, with nothing missing and nothing raised.

`$exec` already reads a `delete`'s relations **before** the row goes, inside a transaction — that is the documented fix for the same hazard on `include`. A `_count` now sets a flag that puts the delete on that same path, so the answer is right on both dialects rather than one. It has to be a flag rather than a wider check on `relations`: a count is not a relation plan, so `relations` stays empty for an `include: { _count: … }` on its own.

## Tests

**Seven differential cases** against Prisma, covering `create`, `update` (with `include`, beside a relation, in a `select`, and with a filtered count), `upsert` and `delete`. The `update` / `upsert` / `delete` ones return a **non-zero** count, which is what separates "projected correctly" from "projected as zero" — the seed grows two accounts on Ada so there is a number to be wrong about. Safe for the existing delete cases: `Account_userId_fkey` is `ON DELETE SET NULL`, so removing a user detaches its accounts on both sides rather than failing a constraint.

**Nine unit tests**, including the one that keeps this from being cosmetic: the `where` inside a `_count` is bound, never inlined.

## Finding out why the `delete` case only discriminates on Postgres turned up something worse

Bun's SQLite driver leaves the pragma at SQLite's default and nothing in gemi turns it on:

```
bun sqlite default foreign_keys -> [{"foreign_keys":0}]
```

So on SQLite **no referential action fires at all**: a deleted user's accounts keep pointing at it, `ON DELETE RESTRICT` does not restrict, and `ON DELETE CASCADE` silently does nothing. Prisma enables the pragma. Measured through the harness — one statement each, one database:

```
                    Account.userId after deleting the user
prisma (fk on)      null, null
gemi   (fk off)     1, 1
```

Filed as **#89**. It is why the `delete with _count` case compares `User` alone rather than `["User", "Account"]` — it could not distinguish a `_count` bug from a foreign-key one, and asserting the wrong thing loudly is worse than asserting less.

It went unnoticed because the harness compares the *written* model's table by default, and no case had ever read a child table back after a parent write.

## Verification

887 unit tests. Full template suite green on SQLite (461 passed) and Postgres 16 (703 passed), with only the pre-existing `generated.test.ts` generator-linkage probe failing — it fails on the base branch too. `tsc` clean; oxlint clean apart from two pre-existing warnings in `where.ts`.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -117,6 +117,12 @@ const user = await User.findUniqueOrThrow({
 })
 ```
 
+A write's `include` takes `_count` as a read's does — `User.create({ data, include: { _count: {
+select: { accounts: true } } } })` comes back with `_count.accounts`. It compiles to a correlated
+subquery inside the `RETURNING`, so it costs no extra statement; a `delete` carrying one reads the
+count before the row goes, for the same reason a `delete` carrying an `include` reads its children
+first.
+
 `select` and `include` narrow the **return type**, exactly as they do in Prisma — `user.email`
 type-checks, `user.name` does not. A key outside the operation's argument type collapses to
 `never` rather than being quietly accepted.
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -119,9 +119,10 @@ const user = await User.findUniqueOrThrow({
 
 A write's `include` takes `_count` as a read's does — `User.create({ data, include: { _count: {
 select: { accounts: true } } } })` comes back with `_count.accounts`. It compiles to a correlated
-subquery inside the `RETURNING`, so it costs no extra statement; a `delete` carrying one reads the
-count before the row goes, for the same reason a `delete` carrying an `include` reads its children
-first.
+subquery inside the `RETURNING`, so it costs no extra statement. Two orderings are handled so the
+number agrees with the row beside it: a `delete` carrying one reads the count *before* the row
+goes, and a write whose nested `create` writes children *after* the statement recomputes it once
+they have — otherwise `include` and `_count` would describe the same relation and disagree.
 
 `select` and `include` narrow the **return type**, exactly as they do in Prisma — `user.email`
 type-checks, `user.name` does not. A key outside the operation's argument type collapses to
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -795,6 +795,39 @@ export abstract class Model {
       // so an `include` on the same call sees what was just written.
       await runSteps(plan.after, effective, context, executor, rowsOf(result));
 
+      // ...and a `_count` on the same call has to be recomputed once they have,
+      // for the same reason `include` is attached after them rather than before.
+      //
+      // A count is a correlated subquery inside the write's own `RETURNING`, so
+      // it is evaluated at the instant the parent row is inserted — before any
+      // `after` step has written a child. That produced a single response
+      // contradicting itself:
+      //
+      //   User.create({ data: { …, accounts: { create: [a, b] } },
+      //                 include: { accounts: true, _count: { … } } })
+      //
+      //   accounts.length  2      attached after the steps
+      //   _count           0      projected before them
+      //
+      // Both keys describe the same relation on the same row. The row itself was
+      // never wrong — reading it back gives 2 — so this is the projection being
+      // frozen, which is the mirror of the `delete` case above: there the count
+      // is taken too late, here too early.
+      //
+      // **Only when there is something to be wrong about.** A write with no
+      // `after` steps keeps the projected value and costs nothing extra; one
+      // with both pays a single read. `plan.counts` is what makes that condition
+      // cheap enough to ask on every write.
+      if (plan.counts && plan.after !== undefined) {
+        await recountAfterSteps(
+          schema,
+          this as unknown as typeof Model,
+          effective,
+          rowsOf(result),
+          options,
+        );
+      }
+
       // Relations are loaded after the root rows are shaped, one query per node
       // in the include tree. Each of those queries is `$exec` on the *related
       // model's own class*, recursively — not a private helper — so a nested read
@@ -953,6 +986,48 @@ function fieldsForColumns(
     }
     return column;
   });
+}
+
+/**
+ * Re-reads the `_count` for rows whose children were written by an `after` step.
+ *
+ * One read per row rather than one for all of them, and that is not a
+ * concession: every write that can carry both a `_count` and an `after` step is
+ * a single-row operation — `create`, `update`, `upsert`. `createMany` takes no
+ * `include` at all. So the loop runs once, and writing it as a loop keeps a
+ * compound primary key expressible without an `in` list over tuples.
+ *
+ * **Pre-scoped, deliberately.** `effective` has already been through this
+ * model's policies and `applyNestedPolicies` has already scoped the `_count`
+ * node, so re-applying them would `AND` the same predicate twice — the same
+ * rows, a different plan key. This is the same reasoning `delete`'s read-first
+ * path uses, a few hundred lines up.
+ */
+async function recountAfterSteps(
+  schema: ModelSchema,
+  model: typeof Model,
+  effective: any,
+  rows: Record<string, unknown>[],
+  options: unknown,
+): Promise<void> {
+  const node = effective?.include?._count ?? effective?.select?._count;
+  if (node === undefined || rows.length === 0) return;
+
+  for (const row of rows) {
+    const where: Record<string, unknown> = {};
+    for (const field of schema.primaryKey) where[field] = row[field];
+
+    const fresh = (await (model as any).$exec(
+      "findFirst",
+      { where, select: { _count: node } },
+      markPreScoped({ strategy: (options as any)?.strategy }),
+    )) as Record<string, unknown> | null;
+
+    // `null` only if the row vanished between the write and this read, which
+    // inside the write's own transaction it cannot. Leaving the projected value
+    // alone is still the right fallback: it is stale, not invented.
+    if (fresh && fresh._count !== undefined) row._count = fresh._count;
+  }
 }
 
 function rowsOf(result: unknown): Record<string, unknown>[] {
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -818,6 +818,13 @@ export abstract class Model {
       // `after` steps keeps the projected value and costs nothing extra; one
       // with both pays a single read. `plan.counts` is what makes that condition
       // cheap enough to ask on every write.
+      //
+      // **Between `runSteps` and the `hidden` deletion, and that is
+      // load-bearing.** `select: { email, _count: … }` never asks for the
+      // primary key, so the key the recount reads the row by is on it only
+      // because the plan added it for the nested write and has not stripped it
+      // yet. Move this after the strip and the recount loses its identifier;
+      // move it before `runSteps` and there is nothing new to count.
       if (plan.counts && plan.after !== undefined) {
         await recountAfterSteps(
           schema,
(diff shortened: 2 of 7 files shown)


nstfkc:
## Review — `_count` in a write's `include` (closes #87)

Read the full diff against the merge base (`8825c87`). 861 unit tests pass.

Projecting rather than refusing is the right call given the measurement — both dialects evaluate a correlated subquery inside `RETURNING`, so the plan `compileRead` already builds goes straight into the returning list at no extra round trip. Verified on both dialects:

```
             create   update   upsert   read (oracle)   delete
sqlite       0        2        2        2               2
postgres     0        2        2        2               2
```

The `delete` column is the one that matters: `Account_userId_fkey` is `ON DELETE SET NULL`, so a count evaluated *after* the statement would be 0 on Postgres. Getting 2 there is the read-first path doing its job. Using a separate `plan.counts` flag rather than widening the `relations !== undefined` check is right for the reason given — a count is a column, not a second query, so `relations` stays empty for an `include: { _count: … }` on its own.

### The same hazard in the other direction: a `_count` cannot see an `after` step

The delete case is "the count is computed too late". The mirror is a write whose children are created *after* the statement, and it produces a single response that contradicts itself:

```ts
await User.create({
  data: { email, accounts: { create: [{…}, {…}] } },
  include: { accounts: true, _count: { select: { accounts: true } } },
})

// accounts.length = 2
// _count          = { accounts: 0 }
```

Both keys describe the same relation on the same row, in the same object. Reading the row back immediately afterwards gives `{ accounts: 2 }`, so the row is fine — it is the projection that is frozen.

The cause is the sequence, and `Model.$exec` already documents the half that works: nested `after` steps run first, *"so an `include` on the same call sees what was just written"*, and then `attachRelations` runs. A `_count` is not attached — it is a subquery inside the `RETURNING` of the insert, evaluated before any `after` step exists. So `include` sees the children and `_count` cannot, by construction.

This is the same class as the `delete` case this PR fixes, and it deserves the same shape of answer: **only when there is something to be wrong about.** A write with no `after` steps keeps the projection and costs nothing; a write that has both `after` steps and a `_count` recomputes the counts once the steps are done, the way `delete` reads once before the row goes. `plan.counts` is already the flag that makes that condition cheap to ask.

Worth deciding rather than leaving: `0` is not obviously wrong at a glance, which is what makes it the quiet kind — the same reason the SQLite `delete` count got its own paragraph here.

### #89 is real, and worse than the ticket states

Verified independently, since it undercuts the instrument this whole project is checked with. It is not only that referential actions do not fire — **nothing is enforced at all**:

```
pragma foreign_keys                                   ->  0
insert into child (pid) values (999)   -- no such parent  ->  ACCEPTED
delete from parent where id = 1        -- on delete cascade
select * from child                                   ->  [{id:10,pid:1},{id:11,pid:999}]
```

So on SQLite an orphan row inserts happily, a dangling foreign key survives a parent delete, and `ON DELETE RESTRICT` does not restrict. Prisma enables the pragma, which means the differential harness is comparing two databases with different integrity semantics — and a whole class of divergence is invisible to it rather than merely untested. Filing it separately was right, and the decision here to compare `User` alone rather than `["User", "Account"]` — because the case could not distinguish a `_count` bug from a foreign-key one — is the honest handling of a test you cannot yet trust.

I would treat #89 as the higher priority of the two.

nstfkc:
Pushed as 77830d6. The finding is right, it reproduces exactly as you wrote it, and it is the mirror of the case this PR already fixed — `delete` takes the count too late, this takes it too early.

Reproduced before touching anything:

```
accounts.length = 2
_count          = {"accounts":0}
reread _count   = {"accounts":2}
```

The third line is the one that settles what kind of bug it is: the row was never wrong. `include` is attached *after* the `after` steps — `$exec` says so in as many words, "so an `include` on the same call sees what was just written" — while `_count` is a subquery inside the insert's own `RETURNING`, evaluated when the parent row exists and no child does. So the two keys could not have agreed.

Fixed with the same shape of answer, including the condition you specified: **only when there is something to be wrong about.** A write with no `after` steps keeps the projected value and costs nothing; one carrying both pays a single read. That is the second thing `plan.counts` has now earned, which makes it look less like a flag added for one branch.

One read per row rather than one for all, and deliberately: every write that can carry both a `_count` and an `after` step is single-row — `create`, `update`, `upsert` — and `createMany` takes no `include`. So the loop runs once, and being a loop keeps a compound primary key expressible without an `in` over tuples.

**Two differential cases, both verified red first** (`0` where Prisma says `2`, and `0` where it says `1`). The first asserts `accounts` **and** `_count` in one comparison, which is the part worth naming: a test checking only `_count` would pass against a version that had also lost the children, and one checking only `accounts` never saw the bug. The second is the `select` form, where `_count` is the only key asked for and the narrowed key set has to survive — it does.

### On #89

Agreed, and your reproduction is worse than my ticket in the way that matters: I described referential *actions* not firing, and you showed **nothing is enforced at all** — an orphan insert accepted, a dangling key surviving a parent delete. I have added that to the issue. The consequence you drew is the one I had not stated plainly: the differential harness is comparing two databases with different integrity semantics, so a whole class of divergence is invisible to it rather than merely untested. That makes it a defect in the instrument, not just in the SQLite path.

Taking it as the next issue rather than leaving it queued.

887 unit tests, full template suite green on SQLite (463) and Postgres 16 (705), with only the pre-existing `generated.test.ts` linkage probe failing.

nstfkc:
### Re-reviewed `77830d6` — the contradiction is gone

My original probe, unchanged, plus the two cases the fix could plausibly have got wrong:

```
include: accounts.length=2   _count={"accounts":2}     agree now
select : keys=_count,email   _count={"accounts":3}     ground truth 3
plain  : _count={"accounts":0}                          no after step, projection kept
```

The middle one is the one I would have gone looking for: `select: { email, _count }` does not ask for the primary key, so the recount depends on the key still being on the row when it runs — and on it being gone from what the caller gets. Both hold, because the recount sits between `runSteps` and the `hidden` deletion. That ordering is load-bearing and it is not obvious from reading either site alone; it is worth the sentence it does not currently have.

Gating on `plan.counts && plan.after !== undefined` keeps the ordinary write free, and the reasoning for one read per row rather than one for all of them — that every write which can carry both is single-row, so the loop runs once — is the right justification to write down rather than the apologetic one.

861 unit tests pass.

### One latent edge, not reachable in this schema

`recountAfterSteps` builds its `where` from `schema.primaryKey`:

```ts
for (const field of schema.primaryKey) where[field] = row[field]
```

What the plan actually guarantees is on the row is the **link** field — `planForeignSide` pushes `link.parentField` into `keyFields`, and that is the column the relation `references`, which is the primary key only by convention. For a relation declared `references: [publicId]`, on a call whose `select` omits the primary key, `row[pk]` is `undefined`; `compileWhere` drops undefined keys, so the `findFirst` runs with no predicate and recounts against an arbitrary row. A wrong number rather than an error.

Not reachable today — every relation in the template schema references `id`, and I did not construct one that does not — so this is a note rather than a finding. The narrow version of the fix is to read the key the plan promised rather than the one convention suggests; the cheap version is to leave the projected value alone when any key component is `undefined`, which matches what the function already does for a `null` row: *"it is stale, not invented"*.

Nothing blocking. #89 remains the one I would prioritise.

nstfkc:
Pushed as 730171d. Both points taken, and the latent one turned out to be worth more than "not reachable today" suggests.

### The incomplete key

You are right that `schema.primaryKey` is not what the plan guarantees — `planForeignSide` pushes `link.parentField`, which is the column the relation `references`, and that is the primary key only by convention. I checked the failure mechanism rather than reasoning about it, and it is the bad kind:

```
findFirst({ where: { id: undefined } })   ->   select … from "User" limit ?
```

No predicate at all. `compileWhere` dropping `undefined` keys is correct — it is how an optional filter is spelled — but it means a `where` assembled entirely from missing components is not a *narrower* query, it is an unfiltered one. The recount would have read an arbitrary row and reported its count: a plausible number, no error.

I took your cheap version rather than the narrow one, and the reason is that it is not actually the weaker fix. Reading the key the plan promised would still need this guard for the case where neither key is present, and "leave the projected value alone" is the answer the function already gives for a missing row — *stale, not invented*. There is nothing there that guessing improves.

**On testing it: I did not add a model.** The shape that reaches the guard needs a relation referencing a non-primary-key column, and every relation in the template schema references `id` — so a differential case would mean a new model and a hand-written migration on both dialects to exercise six lines that are currently unreachable. What I pinned instead is the *hazard*, in `read.test.ts` where the behaviour lives: a `where` whose keys are all `undefined` compiles to no predicate, and one present key still compiles to one. That is the fact any future caller assembling a `where` from row fields needs, and it was written down nowhere. Say the word if you would rather have the model.

### The ordering

Agreed it was load-bearing and unstated, and your `select` probe is the one that shows why — `select: { email, _count }` never asks for the primary key, so the key the recount reads the row by is on it *only* because the plan added it for the nested write and has not stripped it yet. It now says so at the call site: after the strip there is no identifier, before `runSteps` there is nothing new to count. Two constraints, one line, and neither visible from `runSteps` or from `recountAfterSteps` alone.

### #89

Agreed, and it is now open as **#91**. It also makes this PR better tested: with `ON DELETE SET NULL` actually firing on SQLite, the read-before-delete path becomes load-bearing on both dialects instead of only Postgres, and the `delete with _count` case can widen from `["User"]` back to `["User", "Account"]` on whichever of the two merges second.

889 unit tests, full template suite green on SQLite (463) and Postgres 16 (705), with only the pre-existing `generated.test.ts` linkage probe failing.

nstfkc:
### Re-reviewed `730171d` — confirmed, and it took the better half of the suggestion

All three cases still behave, with the correct runner this time (889 tests, `bun --bun vitest run orm`):

```
include: accounts.length=2   _count={"accounts":2}
select : keys=_count,email   _count={"accounts":3}   ground truth 3
plain  : _count={"accounts":0}
```

I offered two fixes and you took the second, which is the right one: reading the key the plan promised would still have been guessing at *which* key, whereas refusing to identify the row at all has a defensible answer already in the file — the projected value stands, "stale, not invented", the same thing a missing row gets.

The reasoning added is stronger than my note was. I said an incomplete key would recount against an arbitrary row; the comment says *why* that is the bad kind of failure — `compileWhere` drops `undefined`, so the `where` compiles to no predicate at all and the result is a plausible number rather than an error. That is the distinction that makes the guard worth having rather than defensive.

Documenting the ordering — between `runSteps` and the `hidden` strip, with what breaks on each side — closes the other half. That was the sentence I said was missing and could not be reconstructed from either site alone.

PR #91, opened 2026-07-28
SQLite enforces foreign keys, like every other dialect

### LINKED ISSUE #89: SQLite runs with `foreign_keys = OFF`, so no referential action ever fires

nstfkc:
Closes #89.

## The failure

SQLite leaves `pragma foreign_keys` off by default and so does Bun's driver. Nothing in gemi turned it on, so on SQLite **nothing was enforced at all** — not "cascades did not fire":

```
pragma foreign_keys                                     ->  0
insert into child (pid) values (999)   -- no such parent   ->  ACCEPTED
delete from parent where id = 1        -- on delete cascade
select * from child                                     ->  [{pid:1}, {pid:999}]
```

An orphan inserted happily, a dangling reference survived the parent's delete, and `ON DELETE RESTRICT` did not restrict — while the migrations declare all three and Postgres enforces them always.

## Why it is worse than a dialect gap

The differential harness compares gemi against Prisma on one database per dialect, and **Prisma enables the pragma**. So the two clients were being asked to agree while running under different integrity rules: any case where gemi would violate a declared constraint passed on SQLite by not checking. A whole class of divergence was invisible to the instrument rather than merely untested — which is why this was worth doing before the `_count` work it was found by.

## The fix, and the two things it rests on

`pragma foreign_keys = ON` from the constructor. Two decisions worth naming:

- **From the constructor, not lazily.** SQLite documents the pragma as a no-op *inside* a transaction, so a "set it on first use" version would have been correct exactly until the first use happened to be a transaction.
- **Not awaited, and that needed proving.** The constructor is synchronous; the pragma is a statement. What makes it safe is that SQLite queues statements on one connection in the order they were issued, so a caller that queries immediately is already behind it. Both halves are asserted rather than assumed — the value, and that 32 concurrent queries agree on it — because they are exactly what breaks if Bun ever serves SQLite from a pool. `ready` is exposed for a caller that would rather have the guarantee than the argument.

## Tests

**Seven unit tests**, every one verified red against the unset pragma: the value, the ordering guarantee, a refused orphan insert, `ON DELETE CASCADE`, `ON DELETE SET NULL`, and a violated key inside a transaction.

**Five differential cases**, in the **shared** matrix rather than a Postgres-only block — because the dialect that used to be wrong is the default one:

```
× create naming a parent that does not exist
× createMany naming a parent that does not exist
× update repointing at a parent that does not exist
× upsert inserting against a parent that does not exist
× nested create naming a parent that does not exist
```

(shown failing against the old behaviour). Each succeeded on SQLite before this and refused on Postgres. The last writes its foreign key from an `after` step rather than from the statement itself, which is a different code path to the same constraint.

## The behaviour change

The full template suite passes with enforcement on — no fixture writes a child before its parent — so this lands without a test migration. Applications developing against SQLite may see writes start to raise that quietly succeeded. **Those were already raising in production**, which is the argument for the change rather than against it, and `docs/orm.md` says so.

## Notes for merge order

- The referential-action half is covered here at the SQL level on both actions. Its *differential* form needs children in the seed, which #90 adds — once that merges, its `delete with _count` case can widen from `["User"]` back to `["User", "Account"]`, which is the line #89's acceptance list asks for. #90 also becomes better tested by this change: with `SET NULL` actually firing, its read-before-delete path becomes load-bearing on SQLite too, not only on Postgres.
- A foreign-key violation currently lands in the harness's `other` failure kind on both sides, so the cases compare "both threw" rather than "both threw the same typed error". A `ForeignKeyConstraintError` mirroring `UniqueConstraintError` would sharpen that; out of scope here.

## Verification

910 unit tests. Full template suite green on SQLite (459) and Postgres 16 (701), with only the pre-existing `generated.test.ts` generator-linkage probe failing — it fails on the base branch too. `tsc` clean; oxlint clean on every changed file.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -121,6 +121,13 @@ const user = await User.findUniqueOrThrow({
 type-checks, `user.name` does not. A key outside the operation's argument type collapses to
 `never` rather than being quietly accepted.
 
+**Foreign keys are enforced on both dialects.** SQLite leaves `pragma foreign_keys` off by
+default and so does Bun's driver, so gemi turns it on for every SQLite connection — otherwise a
+`references` in your migration would mean nothing there while meaning everything on Postgres. An
+insert naming a parent that does not exist is refused, and `onDelete: Cascade` / `SetNull` actually
+run. If you have been developing against SQLite, writes that quietly succeeded may now raise; they
+were already raising in production.
+
 Queries return **plain objects**, never class instances. That is the default and it is not a
 stepping stone: a `Pick<User, "id">` hydrated as a `User` would carry methods reading fields the
 query never fetched. See [Rows and entities](./orm-rows-and-entities.md) for the two opt-in levels
--- a/packages/gemi/database/DatabaseManager.test.ts
+++ b/packages/gemi/database/DatabaseManager.test.ts
@@ -1,0 +1,151 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, describe, expect, test } from "vitest";
+
+import { DatabaseManager } from "./DatabaseManager";
+
+/**
+ * `pragma foreign_keys` on a SQLite connection (#89).
+ *
+ * SQLite defaults it **off** and so does Bun's driver, which reports `0` on a
+ * fresh connection. With it off SQLite enforces nothing: an insert naming a
+ * parent that does not exist is accepted, a dangling reference survives the
+ * parent's delete, and `ON DELETE RESTRICT` does not restrict — while the
+ * migrations declare all three and Postgres enforces them always.
+ *
+ * These run against a real file rather than `:memory:`, because a foreign key
+ * needs two tables that outlive one statement, and because the file case is the
+ * one an application actually has.
+ */
+describe("sqlite foreign keys", () => {
+  const workspaces: string[] = [];
+
+  afterEach(() => {
+    for (const path of workspaces.splice(0)) {
+      rmSync(path, { recursive: true, force: true });
+    }
+  });
+
+  /** The pragma's value, read off whichever connection served the statement. */
+  async function enabled(db: DatabaseManager) {
+    const rows = (await db.sql.unsafe(`pragma foreign_keys`)) as any;
+    return rows[0].foreign_keys;
+  }
+
+  function manager() {
+    const workspace = mkdtempSync(join(tmpdir(), "gemi-fk-"));
+    workspaces.push(workspace);
+    return new DatabaseManager({ url: `sqlite://${join(workspace, "t.db")}` });
+  }
+
+  async function schema(db: DatabaseManager) {
+    await db.sql.unsafe(`create table parent (id integer primary key)`);
+    await db.sql.unsafe(
+      `create table child (
+         id integer primary key,
+         pid integer references parent(id) on delete cascade
+       )`,
+    );
+    await db.sql.unsafe(`create table detachable (
+       id integer primary key,
+       pid integer references parent(id) on delete set null
+     )`);
+    await db.sql.unsafe(`insert into parent (id) values (1)`);
+  }
+
+  test("it is on", async () => {
+    const db = manager();
+    expect(await enabled(db)).toBe(1);
+    await db.close();
+  });
+
+  /**
+   * The constructor is synchronous and the pragma is a statement, so it cannot
+   * be awaited before returning. What makes that safe is that SQLite queues
+   * statements on one connection in the order they were issued — so a caller
+   * that queries immediately is already behind it.
+   *
+   * Asserted rather than assumed: this is the half that breaks first if Bun
+   * ever serves SQLite from a pool.
+   */
+  test("a query issued immediately after construction is already behind it", async () => {
+    const db = manager();
+    const [rows] = await Promise.all([
+      db.sql.unsafe(`pragma foreign_keys`) as Promise<any>,
+      db.ready,
+    ]);
+    expect(rows[0].foreign_keys).toBe(1);
+    await db.close();
+  });
+
+  /** The other half: one connection, so every statement sees the same answer. */
+  test("concurrent queries all see it", async () => {
+    const db = manager();
+    const results = await Promise.all(
+      Array.from({ length: 32 }, () => db.sql.unsafe(`pragma foreign_keys`)),
+    );
+    expect(new Set(results.map((rows: any) => rows[0].foreign_keys))).toEqual(
+      new Set([1]),
+    );
+    await db.close();
+  });
+
+  test("an insert naming a parent that does not exist is refused", async () => {
+    const db = manager();
+    await schema(db);
+
+    await expect(
+      db.sql.unsafe(`insert into child (pid) values (999)`),
+    ).rejects.toThrow(/FOREIGN KEY constraint failed/);
+
+    await db.close();
+  });
+
+  test("on delete cascade removes the children", async () => {
+    const db = manager();
+    await schema(db);
+    await db.sql.unsafe(`insert into child (pid) values (1), (1)`);
+
+    await db.sql.unsafe(`delete from parent where id = 1`);
+
+    const remaining = (await db.sql.unsafe(
+      `select count(*) as c from child`,
+    )) as any;
+    expect(remaining[0].c).toBe(0);
+    await db.close();
+  });
+
+  test("on delete set null detaches them instead", async () => {
+    const db = manager();
+    await schema(db);
+    await db.sql.unsafe(`insert into detachable (pid) values (1)`);
+
+    await db.sql.unsafe(`delete from parent where id = 1`);
+
+    const detached = (await db.sql.unsafe(`select pid from detachable`)) as any;
+    expect(detached[0].pid).toBeNull();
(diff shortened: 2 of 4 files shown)


nstfkc:
## Review — `pragma foreign_keys = ON` (closes #89)

Read the full diff against the merge base (`8825c87`). 910 tests pass (`bun --bun vitest run database orm bin/orm`), and the SQLite template suite is 460 passed / 32 skipped with enforcement on — so the claim that no fixture writes a child before its parent holds.

This was the right thing to prioritise, and the framing is right: it was not a dialect gap but an instrument failure. The differential harness compares gemi against Prisma with the pragma on for one side only, so any case where gemi would violate a declared constraint passed on SQLite by not checking. Putting the five new cases in the **shared** matrix rather than a Postgres-only block is the correct consequence — the dialect that used to be wrong is the default one.

Both design notes are load-bearing and correctly argued: from the constructor rather than lazily, because SQLite documents the pragma as a no-op inside a transaction; and the ordering guarantee asserted rather than assumed, because it is what breaks first if Bun ever pools SQLite connections.

### Blocking: an un-awaited `ready` turns a bad `DATABASE_URL` into process death

```ts
this.ready = this.configure()
```

Nothing attaches a rejection handler, and the doc comment is explicit that nothing has to — *"Nothing has to: SQLite queues statements on one connection…"*. That is true of the **ordering**, but it also means the default path leaves `ready` unhandled, and an unhandled rejection is fatal in Bun.

Before this change, constructing a manager issued no statement, so an unopenable database surfaced at the first query — inside a request, catchable, with a stack pointing at the query. Now the pragma is that first statement and it fires from the constructor. Same script, same bad path, on both branches:

```
=== base 8825c87 ===
constructing...
constructed; not awaiting ready
REACHED: process survived 400ms

=== this branch ===
constructing...
constructed; not awaiting ready
SQLiteError: unable to open database file   errno: 14
<process dies — "REACHED" never printed>
```

So a misconfigured `DATABASE_URL` goes from a catchable error at first use to an asynchronous process exit from a constructor, at a tick nobody controls, naming `configure` rather than anything the application did. In a server that is a crash loop on a config typo; the failure it replaces was a 500 with a usable stack.

It is one line to keep both properties — `ready` still rejects for a caller who wants the guarantee, and the rejection is no longer unhandled:

```ts
this.ready = this.configure()
this.ready.catch(() => {})
```

Storing the error and rethrowing it at first use would be closer to the old behaviour still, if the intent is that a broken connection should fail where it is used rather than where it is built. Either way the current shape makes `ready` a trap for exactly the callers the comment tells not to bother with it.

Worth a test in the same file, since the others in it were each verified red: construct against an unopenable path, do not await, assert the process is still alive.

### A note on my own runs, because it nearly became a false report

I have been running `bun test packages/gemi/...` all session. This package's runner is `bun --bun vitest`, and every test file in it — including this PR's new one — imports from `vitest`. Most files happen to work under either; `DatabaseManager.test.ts` does not, and hangs. I was one step from reporting that as a defect in this PR. It is my invocation: under the correct runner the file passes and the total is exactly the 910 the description claims.

Mentioning it because the same trap is one command away for anyone else checking this branch, and because my earlier per-PR test counts in this repo came from the wrong runner — they executed and passed, but `bun --bun vitest` is the number to trust.

### Smaller

The merge-order notes are accurate and worth keeping. The observation that #90 becomes *better* tested by this change — with `SET NULL` actually firing, its read-before-delete path becomes load-bearing on SQLite rather than Postgres only — is the kind of thing that is easy to notice and easy not to write down.

nstfkc:
### This merged with the unhandled-rejection finding unaddressed — it is live on `feat/orm` now

Flagging rather than letting it sit, since the PR is closed and the branch it landed on is the integration one. `origin/feat/orm` (`d38b367`):

```ts
this.ready = this.configure();   // no rejection handler
```

Same script as before, run against the merged branch rather than the PR branch:

```
constructing...
constructed; not awaiting ready (the documented common case)
SQLiteError: unable to open database file
<process dies — "REACHED: process survived 400ms" never printed>
```

So on `feat/orm` today, a `DATABASE_URL` pointing at an unopenable SQLite path kills the process asynchronously from the constructor. Before this change the same configuration was inert until the first query, where it surfaced as a catchable error with a useful stack. For a server that is the difference between a 500 on a misconfigured deploy and a crash loop.

Everything else about the change is good and I would not want it reverted — the enforcement itself is right, and 963 tests pass on the merged branch. It is one line:

```ts
this.ready = this.configure();
this.ready.catch(() => {});   // `ready` still rejects for anyone awaiting it
```

Happy to be told it is intended — a manager that cannot configure its connection is arguably fatal — but then it should be fatal *deliberately*, at construction, rather than at whatever tick the promise settles on, and the doc comment saying nothing has to await `ready` should say the opposite.

---

While verifying the merge I also checked the thing I have been tracking across these branches: `plan.ts`'s `LITERAL_KEYS` survived eight PRs landing together, carrying `skipDuplicates` from #76 alongside #74's five aggregate keys and #72's `skip` handling. `omit` and `plan.coverage.test.ts` are absent, which is correct — #78 is still open.

nstfkc:
## Follow-up — the un-awaited `ready` still kills the process, re-measured 25 PRs later

I raised this when reviewing this PR and it merged unaddressed, so I re-checked it against the current top of the ORM stack (`fe9679b`, #134). **Unchanged, and still fatal.**

`DatabaseManager.ts:47`:

```ts
this.ready = this.configure();
```

Constructed against a path that cannot be opened, with nobody awaiting `ready`:

```
constructed, ready is a promise: true      <- the constructor returns normally
SQLiteError: unable to open database file  (SQLITE_CANTOPEN)
bun exit = 1                               <- dead before a 300ms timer fires
```

The two lines together are the whole problem. The constructor **succeeds**, so a caller wrapping it in `try/catch` is not protected and has no way to be; then the rejection goes unhandled and takes the process down. Before this PR the same construction was inert, because there was no promise.

The trigger is ordinary rather than exotic — a `DATABASE_URL` pointing at a directory that does not exist yet is the first-run case, and the failure arrives as a bare stack trace from Bun's SQLite adapter rather than as anything the framework says.

One line, and I checked it in both directions rather than only the one I wanted:

```ts
this.ready = this.configure();
// A caller that never awaits `ready` must not take the process down: an
// unhandled rejection is fatal and the constructor has already returned.
// The original promise still rejects for anyone who does await it.
void this.ready.catch(() => {});
```

```
process survives          exit 0, "STILL ALIVE after 300ms"
await db.ready            still rejects — SQLiteError
database + orm suites     1214 passed
```

The second line is the one worth confirming: `.catch` returns a *new* promise, so marking the rejection handled does not swallow it for a caller who awaits `ready` deliberately. The failure stays reportable; it just stops being fatal to a caller who never asked.

Raising it here rather than opening an issue, since this is where the line landed and the reasoning for it lives.

PR #92, opened 2026-07-28
`groupBy`, and the four rules that are not what the arguments suggest

### LINKED ISSUE #64: ORM: `aggregate` and `groupBy`

nstfkc:
Closes #64, whose `aggregate` half shipped in #74.

## Why it is its own compiler

Not a flag on `compileAggregate`, for three reasons that all show up in the file:

- the result is many rows with a **mixed** shape — grouped columns flat, aggregates nested under their kind;
- `having` is a second predicate compiler whose operands are *function calls* rather than columns (`count("x") > $1` where `compileWhere` would emit `"x" > $1`);
- `orderBy` carries a restriction no other operation has, and may itself name an aggregate.

The term machinery is **shared, not copied** — `aggregateTerms`, `coerce`, `FUNCTION` and the kind list are now exported from `aggregate.ts`. Two copies of the `_sum`-over-`BigInt` coercion is exactly the drift #84 turned out to be.

## Every rule was measured, and four are not what the shape suggests

Three of these would have shipped wrong from reading the types or the docs:

- **`orderBy` is optional.** #64 predicted Prisma would require one whenever `by` is given. It does not — so requiring it would have refused legal queries.
- **`by` takes a bare string** as well as an array. The generated types read as array-only.
- **`having` on an ungrouped column depends on the form.** Prisma's refusal — *"must either be an aggregation filter or be included in the selection of the query"* — is an **or**. Reading it as "the field must be in `by`" refuses `having: { email: { _count: { gt: 1 } } }`, which Prisma answers, because `count(email)` has one value per group either way. I very nearly shipped that reading.
- **The `orderBy` restriction is worth reproducing rather than passing through.** SQLite will happily order by an ungrouped column, picking an arbitrary row's value. Silently arbitrary beats refused for nobody.

## One refusal with no Prisma behaviour behind it

```ts
having: { role: { gt: 0, _count: { gt: 1 } } }
```

mixes a column comparison and an aggregate filter under one key. Prisma's query engine **panics**:

```
thread 'tokio-runtime-worker' panicked at
  query-engine/core/src/query_graph_builder/read/aggregations/group_by.rs:136
internal error: entered unreachable code
```

There is a sensible statement for that shape, and inventing it would put gemi ahead of Prisma somewhere a differential test has no oracle. Refused by name, with the `AND` spelling that works.

## Plan cache and policies

**`by` joins `LITERAL_KEYS`.** The grouped columns reach the SQL as *identifiers*, and `["role"]` and `["locale"]` have the same shape — `[string]` — so without it two callers share one plan and the second is grouped by the first one's column. Well-formed, wrong, silent.

**`groupBy` joins `SCOPABLE`**, and it is load-bearing. Verified by removing it:

```
UnsupportedQueryError: gemi ORM does not support 'groupBy' yet (User.groupBy).
User has a policy with a scope, and groupBy cannot carry one: its where clause
compiles to an 'on conflict' target…
```

— the operation disappears on every policied model, pointing at `upsert`. `softDeletes()` carries a `scope`, so that is most models. This is how `aggregate` was found missing on #74.

## Tests

- **37 unit tests** over the SQL and the refusals — the `having` compiler, the `orderBy` restriction, argument validation, and that no `having` value reaches the SQL text.
- **16 differential cases** against Prisma on both dialects, including the null group, the empty result, per-field vs `_all` counts, ordering by an aggregate, and the ungrouped-aggregate `having` that the misreading would have refused.
- **A policy test asserting the *groups* change**, not only the counts: a scope applied after grouping would leave the other tenant's group in the result with a smaller number, and a test on counts alone would pass against that.

## Notes for merge order

`groupBy` is the **fourth pager**, which is the case #88 was written for: it moved `take` / `skip` validation into `pagination` itself rather than each caller's allowlist, arguing a fourth pager should not be able to arrive without it. On this base #88 has not merged, so `groupBy` inherits the #84 behaviour today and gains the check the moment it does, with no change here. The call site is the only line that differs between the two bases (`pagination` gained an `operation` parameter), and it is noted in the file.

## Verification

920 unit tests. Full template suite green on SQLite (471) and Postgres 16 (714), with only the pre-existing `generated.test.ts` generator-linkage probe failing — it fails on the base branch too. `tsc` clean on the package and on the template; oxlint clean.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -100,7 +100,7 @@ read. It can only see modules you hand it, so it does not replace the `register`
 Fourteen operations, with Prisma's argument types verbatim:
 
 ```
-findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count   aggregate
+findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count   aggregate   groupBy
 create     createMany  update             updateMany   upsert              delete   deleteMany
 ```
 
@@ -171,9 +171,37 @@ await User.count({ select: { _all: true, email: true } })  // { _all: 3, email: 
 Results are typed by Prisma's own mapped types, so `_max: { position: true }` narrows to
 `{ _max: { position: number | null } }` and nothing else.
 
-**`groupBy` is not implemented.** `having` is a predicate compiler over aggregate expressions rather
-than columns, which is its own piece of work; shipping half of it typed as though it worked would be
-worse than not shipping it.
+### `groupBy`
+
+```ts
+const perOperation = await Usage.groupBy({
+  by: ["operation"],                       // or "operation" — Prisma takes both
+  where: { occurredAt: { gte: since } },
+  _sum: { credits: true },
+  _count: true,
+  having: { credits: { _sum: { gt: 0 } } },
+  orderBy: { _count: { operation: "desc" } },
+})
+// [{ operation: "render", _count: 42, _sum: { credits: 1200 } }, …]
+```
+
+The grouped columns come back flat and the aggregates nested under their kind, which is Prisma's
+shape. Four rules are worth knowing because they are not what the arguments suggest:
+
+- **`orderBy` may only name grouped columns — or an aggregate.** A group has one value for a column
+  in `by` and many for everything else, so ordering by anything else is a question with no answer.
+  SQLite answers it anyway, by picking an arbitrary row's value, so this is refused rather than
+  passed through. `orderBy: { _count: { field: "desc" } }` is the top-N-by-count query and is fine.
+- **`orderBy` is optional.** `groupBy` with none is a legal query.
+- **`having` filters groups, `where` filters rows.** `having: { role: { gt: 0 } }` needs `role` to be
+  in `by`; `having: { email: { _count: { gt: 1 } } }` does not, because a count has one value per
+  group either way. That split is Prisma's, and it is easy to read as stricter than it is.
+- **`take` / `skip` page the groups**, not the rows — unlike `aggregate`, where they page the rows
+  being aggregated.
+
+One divergence, and it is a refusal rather than a difference: `having: { role: { gt: 0, _count: { gt: 1 } } }`
+mixes a column comparison and an aggregate filter under one key. Prisma's query engine panics on that
+shape rather than answering it, so there is no behaviour to match — spell it as an `AND` of the two.
 
 ### Per-call options
 
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -555,9 +555,20 @@ function countOperation(model: string): string {
  * here: the types are Prisma's own, so narrowing stays exact and there is no
  * second definition of the shape to drift.
  *
- * `groupBy` is deliberately not beside it. `having` is a predicate compiler over
- * aggregate expressions rather than columns, and shipping half of it typed as
- * though it worked would be worse than not shipping it.
+ * `groupBy` is beside it now, and its typing is the awkward one.
+ *
+ * Prisma's `groupBy` signature is not a plain `Subset`: it carries a conditional
+ * that reports a *missing* `by` field in `orderBy` as a type error naming the
+ * field. Reproducing that machinery would mean copying a large conditional out
+ * of the generated client, where it would then be a second definition free to
+ * drift from the first — the thing every other operation here avoids.
+ *
+ * So the args type is Prisma's `${model}GroupByArgs` directly and the return is
+ * its `Get${model}GroupByPayload<T>`, which still narrows the payload exactly.
+ * The `orderBy`-must-be-in-`by` rule is enforced at runtime instead, with a
+ * message naming the field and saying why — see `compile/group-by.ts`. A
+ * compile-time error would be better; a *wrong* compile-time error, or a silent
+ * one, would be worse than a precise runtime refusal.
  */
 function aggregateOperation(model: string): string {
   return `
@@ -566,6 +577,14 @@ function aggregateOperation(model: string): string {
   ): Promise<Prisma.Get${model}AggregateType<T>> {
     return this.$exec("aggregate", args) as Promise<
       Prisma.Get${model}AggregateType<T>
+    >;
+  }
+
+  static groupBy<T extends Prisma.${model}GroupByArgs>(
+    args: Prisma.Subset<T, Prisma.${model}GroupByArgs>,
+  ): Promise<Prisma.Get${model}GroupByPayload<T>> {
+    return this.$exec("groupBy", args) as Promise<
+      Prisma.Get${model}GroupByPayload<T>
     >;
   }
 `;
--- a/packages/gemi/orm/compile/aggregate.ts
+++ b/packages/gemi/orm/compile/aggregate.ts
@@ -46,8 +46,8 @@ import { compileWhere } from "./where";
  */
 
 /** The aggregate functions Prisma exposes, in the order it returns them. */
-const KINDS = ["_count", "_avg", "_sum", "_min", "_max"] as const;
-type Kind = (typeof KINDS)[number];
+export const KINDS = ["_count", "_avg", "_sum", "_min", "_max"] as const;
+export type Kind = (typeof KINDS)[number];
 
 const AGGREGATE_ARGS = new Set<string>([
   "where",
@@ -73,7 +73,7 @@ const NUMERIC = new Set<ScalarType>(["Int", "Float", "BigInt", "Decimal"]);
 const UNORDERABLE = new Set<ScalarType>(["Json", "Bytes"]);
 
 /** The SQL function each kind compiles to. */
-const FUNCTION: Record<Kind, string> = {
+export const FUNCTION: Record<Kind, string> = {
   _count: "count",
   _avg: "avg",
   _sum: "sum",
@@ -81,7 +81,7 @@ const FUNCTION: Record<Kind, string> = {
   _max: "max",
 };
 
-interface Term {
+export interface Term {
   kind: Kind;
   /** Absent for `_count`'s `_all`, which is `count(*)`. */
   field?: FieldSchema;
@@ -234,9 +234,9 @@ export function compileAggregate(
 }
 
 /** The `as` a bare `_count: true` carries, which no field name can be. */
-const BARE = "";
+export const BARE = "";
 
-function aggregateTerms(
+export function aggregateTerms(
   schema: ModelSchema,
   op: string,
   args: any,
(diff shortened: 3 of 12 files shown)


nstfkc:
Merged `feat/orm` — pushed as 3cd4cfa. **The compile error this PR predicted is the one it got**, which is the outcome #88 was arranged to produce:

```
orm/compile/group-by.ts(137,29): error TS2554: Expected 5-6 arguments, but got 4.
```

One argument at one call site. The interesting half is what it bought: #88 moved `take` / `skip` validation into `pagination` itself rather than into each caller's argument allowlist, on the argument that a **fourth** pager should not be able to arrive without it. `groupBy` is that fourth pager — written against a base where #88 had not merged, validated the moment it did, with nothing in this file doing anything about it.

That is now asserted rather than described:

```
take and skip are validated, without this file knowing the rule
  take: "-2"  ->  Expected an integer, got "-2"
  skip: -1    ->  'skip' counts rows to pass over
  take: 1.5   ->  ...and the message names User.groupBy
```

Two conflicts, both additive: `LITERAL_KEYS` gained `skipDuplicates` from #76 beside `by`, and `aggregateTerms` kept the exported signature this branch needs while taking the base's body.

Also dropped a stale sentence: `plan.ts`'s operation list still said `groupBy` was deliberately excluded, one paragraph above the note explaining that it is not.

973 unit tests, template suite green on SQLite (508) and Postgres 16 (764), with only the pre-existing `generated.test.ts` linkage probe failing.

nstfkc:
## Review — `groupBy` (closes #64)

Read the full diff against the merge base (`8825c87`). 1002 tests pass (`bun --bun vitest run orm bin/orm database`).

Two things I would have raised are already here, with the reasoning and the history attached: `by` in `LITERAL_KEYS` — grouped columns reach the SQL as identifiers and `["role"]` / `["locale"]` share a shape — and `groupBy` in `SCOPABLE`, verified by removing it and watching the operation disappear behind `upsert`'s message. That is the third time that second one has mattered, and pinning it with a policy test asserting the *groups* change rather than only the counts is the version that actually discriminates.

Sharing `aggregateTerms` / `coerce` / `FUNCTION` rather than copying them is right, and the justification — that two copies of the `_sum`-over-`BigInt` coercion is what #84 turned out to be — is the correct read of that history. Refusing the mixed column/aggregate `having` because Prisma's engine *panics* is the right call: there is no oracle for a shape the reference implementation cannot answer.

### `having`'s values reach the plan cache key

`having` is not in `VALUE_KEYS`, and the aggregate kinds are in `LITERAL_KEYS` — put there by #74 so a *projection* like `_count: { email: true }` keeps its booleans. Inside a `having` the same keys re-raise `literal`, so the comparison operands are recorded verbatim:

```
{ having: { role: { _count: { gt: 5    } } } }
  -> …:groupBy:{_count:true,by:["role"],having:{role:{_count:{gt:5}}}}
{ having: { role: { _count: { gt: 9999 } } } }
  -> …:groupBy:{_count:true,by:["role"],having:{role:{_count:{gt:9999}}}}
same key: false
```

`where` with the same two values is one key, which is the behaviour this should match. Measured over 500 requests differing only in a threshold:

```
500 distinct having thresholds -> size=500  compiles=500
500 distinct where  thresholds -> size=1    compiles=1
same SQL text for gt:1 and gt:999 : true
```

500 entries holding byte-identical statements. That is precisely the pathology `plan.ts` already describes for `in`-list lengths — *"one entry per distinct list length, each holding SQL identical to its neighbours'"* — and the cap is 1000, so a threshold off a query string (`?minCount=`) fills the cache and starts evicting every other query's plan. `VALUE_KEYS`' own comment names the second half: recording values verbatim *"would put user data into a long-lived global map"*.

**The obvious fix does not work, and I checked before suggesting it.** Adding `"having"` to `VALUE_KEYS` changes nothing: it resets `literal` at that level, and `_count` re-raises it one level down. And a blanket "nothing inside a value subtree is literal" would be wrong — `mode` lives inside `where` and *must* stay literal, since `insensitive` and `default` compile to `ilike` and `like`:

```
{where:{name:{contains:string,mode:"insensitive"}}}
{where:{name:{contains:string,mode:"default"}}}      same key: false   — correctly
```

So the real distinction is that an aggregate kind means two different things by position: in a projection its values are structural, in a `having` they are operands. Same key, two meanings, and the canonical-shape walk currently cannot tell them apart. Whatever form it takes, the property to land on is the one the `where` row above already has.

### Worth adding while it is fresh

#78's `plan.coverage.test.ts` asserts one direction — arguments that compile to *different* statements must not share a key. This is the converse: arguments that compile to the *same* statement must not mint separate keys. Nothing tests that today, which is why three instances of the first were caught by it and this one was not caught by anything. A companion assertion in the same file — compile twice with two values for the same argument, assert one entry when the SQL text matches — would cover the class rather than the instance, and `planCacheStats().size` makes it a two-line check.

### Smaller

The four measured rules are the part of this PR I would keep even if the feature changed — particularly `having` on an ungrouped column being an **or**. Reading Prisma's message as "the field must be in `by`" is the natural reading and refuses a query Prisma answers; saying you nearly shipped it is more useful to the next reader than a clean claim would have been.

`groupBy` being the fourth pager is exactly the case #88 was built for, and it arriving through `pagination()` without needing its own validation is the argument for that placement landing.

nstfkc:
### Still reproducing on `3cd4cfa`

That head move was a `feat/orm` merge, so the `having` finding is untouched — re-checked rather than assumed:

```
200 distinct having thresholds -> cache size = 200
```

Unchanged from the 500/500 measured before. Flagging only so it does not look addressed by the merge; nothing to re-review otherwise.

nstfkc:
### Re-reviewed `b994ca4` — fixed, and it satisfies all four constraints at once

500 distinct thresholds now share one entry, down from 500. The reason this was worth a flag rather than a one-liner is that three other properties had to survive it, and all four hold:

```
1 having gt:5 vs gt:9999 share a key : true    the fix
2 having gt   vs lt      separate    : true    the operator is still structural
3 projection _count true vs false    : true    #74's reason for LITERAL_KEYS intact
4 mode insensitive vs default        : true    ilike vs like still separate

500 thresholds -> cache size 1   (was 500)
```

1009 tests pass.

`inHaving` as a separate flag rather than clearing `literal` is the right shape, and the comment states the rule precisely: *"nothing below a value key is structural" is not the rule; "an aggregate kind below a `having` is not" is*. That is the distinction I could show by counter-example but had not managed to phrase — the `mode` case proves the general form wrong without telling you what the specific form should be.

Naming the five keys as the only ones with two positions, and saying so as the justification for a set rather than a rule, is what makes this readable later: the next person adding an aggregate-shaped key knows to ask which positions it has.

nstfkc:
Fixed and pushed as b994ca4 — and you were right that the merge had not touched it.

```
200 distinct having thresholds -> cache size = 1
```

**Your diagnosis was the whole fix.** The keys mean two different things by position: as a *projection* `_count: { email: true }` decides the select list, so #74 put it in `LITERAL_KEYS`; inside a `having` the same key introduces a comparison whose operand is a bound parameter. The walk could not tell them apart. So an aggregate kind no longer raises `literal` when it is inside a `having`, and `having` itself resets it.

Both of the traps you flagged are checked rather than assumed:

```
having gt:5 vs gt:9999        same key   ← fixed
where  gt:5 vs gt:9999        same key   ← unchanged
_count:{email:true} vs false  differ     ← #74's property, intact
mode insensitive vs default   differ     ← still literal, as you warned
```

Ruling out `VALUE_KEYS` before suggesting it saved me from shipping it — I would have tried that first, and it does nothing because `_count` re-raises one level down.

### The converse test

Added, and it is the more valuable half of the review. Every discrimination test in that file checks one direction; nothing checked the other, which is exactly why three instances of the first were caught by `plan.coverage`'s rule and this one was caught by nothing.

`SAME_STATEMENT` walks seven pairs that must each compile to one statement and one entry — a `where` threshold, `take` / `skip` magnitudes, a value inside a relation filter, a value inside a nested include's `where`, and both `having` forms. It asserts the *premise* first (`texts.size === 1`) so a badly-chosen pair fails loudly rather than passing vacuously, which is the failure mode of a test like this.

Verified red against the unfixed key: `size: 2` where it wants 1.

981 unit tests, template suite green on SQLite (508).

PR #94, opened 2026-07-28
Nested `connectOrCreate`, on both sides of the key

nstfkc:
The second item on #65's own suggested order, after `createMany`. **Stacked on #75**, which is where `createMany` lives — hence the base, so the diff here is only the new operand.

## The thing the name gets wrong

**A hit ignores `create` entirely.** It is connect-*or*-create, not upsert. Measured against Prisma rather than assumed: an existing organisation kept its own name where the `create` payload named a different one, and an existing account kept its own role.

The differential case is written so an implementation that upserted would fail it — the create payload is named `SHOULD-NOT-APPEAR`. A test that only checked "one organisation exists afterwards" would pass against the wrong implementation.

## Both directions, because they are different work

- **This row holds the foreign key** → a `before` step that resolves to one more bound column.
- **The child holds it** → a hit is an `update` of the child's key (Prisma *repoints* the existing row rather than duplicating it — verified), a miss is a `create` with the key stamped on.

`findUnique`, not `findUniqueOrThrow`: a miss is the *other branch* here, where for a plain `connect` it is an error.

## The scoped-away case is the interesting one, and it must not raise

Both branches go through the child's own `$exec`, so both are scoped — and they have to be, in **opposite directions**:

- an unscoped *update* would re-parent another tenant's row;
- an unscoped *read* would let a hit reveal that a row with that key exists.

Scoped, a hidden row reads as absent and the call creates its own — the same answer the caller would get if it genuinely did not exist. That symmetry is the point: `connect` raising while `connectOrCreate` succeeded would together confirm the row is there, which is a probe.

Verified by pre-scoping the lookup and watching the test fail — with the child's policy skipped it connects to the other tenant's folder, leaving one row where there should be two:

```
× connectOrCreate cannot see another tenant's row, and creates instead
  expected [ { code: 'theirs', orgId: 99 }, …(4) ] to have a length of 2 but got 1
```

## Validated at plan time

`where` and `create` both required, no other keys, `where` unique, and a list refused on a to-one — Prisma refuses all four the same way. Checked when the query compiles rather than after the parent row is written and the transaction has to unwind it, which is the same reason `createMany`'s operand is checked there.

## Tests

- **5 differential cases** — hit and miss on each side, plus a list that does both in one call. The foreign-side ones compare the `Account` table, not only the returned payload, because a hit is a write to a row the call did not create.
- **9 unit tests** over the plan shape (which side gets a `before` vs an `after` step) and the four refusals.
- **2 policy tests**, one of them verified red against a pre-scoped lookup.

## Verification

899 unit tests. Full template suite green on SQLite (473) and Postgres 16 (715), with only the pre-existing `generated.test.ts` generator-linkage probe failing — it fails on the base branch too. `tsc` clean; oxlint clean.

## What is left on #65

`set` / `disconnect` / `delete` / `deleteMany` on ordinary to-many relations — item 3, and cheap now that both sides' child-statement sequencing exists — then the nested `update` / `updateMany` / `upsert` family, which #65 itself flags as the place to stop and decide whether the whole tree is in scope. (#83 covers those operands for *implicit many-to-many* only, where there is a join table rather than a foreign key.)

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -245,6 +245,7 @@ await List.create({
 | `create` | Writes new related rows. One statement each. |
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
 | `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
+| `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -260,10 +261,16 @@ Three things follow from that, and all three are load-bearing:
   A `createMany` writes its rows in one statement and they are *each* scoped.
 - **A failure anywhere rolls the whole thing back**, including the parent row.
 
-Everything else in Prisma's nested grammar — `connectOrCreate`, `set`, `disconnect`, `update`,
-`updateMany`, `upsert`, `delete`, `deleteMany` — is refused, by name and with the reason. They share
-one property: each writes rows that already exist, which needs a scoping pass of its own rather than
-the child's `onCreate`. `skipDuplicates` is not implemented on `createMany` at any level.
+`connectOrCreate` is worth one more line, because a scoped-away row makes it take the *create*
+branch rather than raise: a row your policies hide reads as absent, so the call writes its own — the
+same answer you would get if it truly did not exist. That is deliberate. `connect` raising where
+`connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
+tenant.
+
+Everything else in Prisma's nested grammar — `set`, `disconnect`, `update`, `updateMany`, `upsert`,
+`delete`, `deleteMany` — is refused, by name and with the reason. They share one property: each
+writes rows that already exist, which needs a scoping pass of its own rather than the child's
+`onCreate`. `skipDuplicates` is not implemented on `createMany` at any level.
 
 ### Paging a relation
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -315,9 +315,14 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 Everything else in Prisma's nested grammar — `set`, `disconnect`, `update`, `updateMany`, `upsert`,
-`delete`, `deleteMany` — is refused, by name and with the reason. They share one property: each
-writes rows that already exist, which needs a scoping pass of its own rather than the child's
-`onCreate`. `skipDuplicates` is not implemented on `createMany` at any level.
+`delete`, `deleteMany` — is refused, by name and with the reason. The line is **which rows an
+operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
+one you identified by unique key) and writes either a whole new row or one foreign key the ORM
+chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
+name; `update` names its row but writes your columns to it, which needs the child's `onUpdate` and
+the scope-escape guard run over the payload.
+
+`skipDuplicates` is not implemented on `createMany` at any level.
 
 ### Paging a relation
 
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -41,7 +41,12 @@ import { matchUniqueKey } from "./unique";
  * back the parent row too.
  */
 
-const SUPPORTED = new Set(["connect", "create", "createMany"]);
+const SUPPORTED = new Set([
+  "connect",
+  "connectOrCreate",
+  "create",
+  "createMany",
+]);
 
 /**
  * The operands still refused, and what each would take.
@@ -53,9 +58,6 @@ const SUPPORTED = new Set(["connect", "create", "createMany"]);
  * key, and none of it has to reason about what was there before.
  */
 const REFUSED: Record<string, string> = {
-  connectOrCreate:
-    `It is an upsert against the child, which needs the read-then-write ` +
-    `fallback 'upsert' uses when 'on conflict' cannot express the target.`,
   set: `It replaces the whole set, so it has to disconnect what is there now.`,
   disconnect: `It clears a foreign key on rows this call did not name.`,
   delete: `It deletes rows this call did not name.`,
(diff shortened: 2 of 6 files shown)


nstfkc:
Retargeted at `feat/orm` and merged the base — pushed as 06ba086. This was stacked on #75's branch so the diff would be only the new operand; now that #75 has landed the stack is unnecessary, and the PR stands on its own.

No conflicts: nothing else touched the nested-write operand table.

942 unit tests, template suite green on SQLite (498) and Postgres 16 (753), with only the pre-existing `generated.test.ts` linkage probe failing.

Worth noting what #75's merge changes about the remaining work on #65: `set` / `disconnect` / `delete` / `deleteMany` on ordinary to-many relations — item 3 — no longer need any stacking, since both sides' child-statement sequencing is now on `feat/orm`.

nstfkc:
## Review — nested `connectOrCreate` (#65, item two)

Read the full diff against the merge base with `feat/orm` (`d38b367`) — 6 files, +563.

The scoping analysis is the strongest part and it is the part I would have gone at hardest. Routing both branches through the child's own `$exec` un-pre-scoped, in **opposite directions** — an unscoped update would re-parent another tenant's row, an unscoped read would let a hit reveal that a key exists — is right, and the observation that the two together form a probe is the one that is easy to miss: `connect` raising where `connectOrCreate` succeeds confirms the row is there. Falling back to *create* on a scoped-away hit is what closes that, and it gives the caller the same answer they would get if the row genuinely did not exist. Verifying it by pre-scoping the lookup and watching the test fail is the right standard for a property whose failure is silent.

"A hit ignores `create` entirely" is the semantic I would have assumed wrong, and naming the create payload `SHOULD-NOT-APPEAR` so an upserting implementation fails rather than passes is the difference between a test and a test that discriminates.

### The file's stated rule no longer describes the file

`nested-writes.ts`'s header still reads:

> Everything else in Prisma's nested-write grammar — **`connectOrCreate`**, `set`, `disconnect`, `update`, `upsert`, `delete`, `deleteMany`, `updateMany` — throws `UnsupportedQueryError` naming the operation *and the reason*; see `REFUSED`. **They share one property, and it is the line this file draws: each writes rows that already exist**, which needs a scoping pass of its own rather than the child's `onCreate`.

Two things are now wrong with that, and the second matters more than the first.

`connectOrCreate` is in `SUPPORTED` twenty lines below and implemented 250 lines down, so the list is stale. That is a one-word fix.

But the *rule* is also broken, and by this PR specifically. On the foreign side a hit is an `"update"` — the PR body says so plainly, that Prisma repoints the existing row rather than duplicating it, and the code issues exactly that. So a supported operand now writes a row that already exists, which is the property the header offers as the dividing line between what is implemented and what is refused. The line is real and worth having; it just is not that line any more.

That matters because it is the rule the next operand gets judged against — #75 built its whole `REFUSED` table on it, with a per-entry reason derived from it. Someone reaching for `update` or `set` next will read a criterion that this PR has already crossed, and either conclude those are fine or conclude the file is untrustworthy. Worth restating what the actual boundary is now: my reading of the code is that it is *which rows can be named* — every supported operand either creates a row or repoints one the caller identified by unique key, and every refused one acts on rows the call did not name (`set` and `disconnect` clear whatever is there, `deleteMany` / `updateMany` match a predicate). That reading also explains #83's `set`, where the fix was precisely to scope the delete to rows the caller can see.

### Smaller

- Validating the operand at plan time rather than after the parent row is written is the same call `createMany` made, and it is right for the same reason — a refusal that arrives mid-transaction has to unwind a write that should never have started.
- `findUnique` rather than `findUniqueOrThrow`, because a miss is the other branch here rather than an error, is a one-line difference that would have been easy to copy wrong from the `connect` path directly below it.

I read the implementation rather than running the database probes this round, so the differential and policy claims are taken from the diff and the tests rather than independently reproduced; the scoping structure is visible in the code and matches the description.

nstfkc:
Pushed as 8f1b9e3. The stale list was a one-word fix; the rule underneath it was the finding, and you are right that **this PR is what broke it**.

I went to restate it as "which rows can be named" and found that reading does not quite hold either — which is worth saying, because the whole point of the criterion is that the next operand gets judged against it.

**Nested `update` is `{ where, data }`. It names its row.** So "acts on rows the call did not name" explains `set`, `disconnect`, `delete` and `deleteMany`, but not `update`. It is refused for a second reason: *whose columns* it writes. Caller-supplied data needs the child's `onUpdate` and the scope-escape guard run over the payload, where a `connect` writes one foreign key the ORM itself chose.

So the header now states both halves:

- every supported operand **names its rows**, so it goes through the child's own `findUnique` / `create` / `update` and the child's policies decide reachability — "whatever is there" has no lookup to hang a scope on;
- every supported operand writes either a whole new row through the child's `create`, where `onCreate` applies, or **one foreign key the ORM chose**.

`REFUSED`'s `update` and `updateMany` entries carried the old criterion verbatim and now carry the accurate one, with a comment at `update` saying why the obvious phrasing is wrong — it is the entry most likely to be copied for the next operand. `docs/orm.md` repeated the same sentence and is corrected with it.

Your framing of why this matters is the part I would keep: a stale criterion is worse than none, because it is the one the next reader applies.

Noted on reading rather than running this round — the differential and policy claims are reproducible with `TZ=UTC bun --bun vitest run app/models` in `templates/saas-starter`, and the pre-scoping check is the one I would re-run first, since its failure is silent.

942 unit tests, template suite green on SQLite (498).

nstfkc:
### Re-reviewed `8f1b9e3` — the restated rule is better than the one I proposed

I offered "which rows can be named". That is only half, and `update` is the counter-example that shows it: `update: { where, data }` **does** name its row by unique key, so a rule about naming would classify it as supportable — and it is refused for a different reason entirely.

The two-half form handles every entry, which I checked against the sets rather than taking on trust:

| operand | names its rows | writes | verdict |
| --- | --- | --- | --- |
| `connect` | unique key | one FK the ORM chose | supported |
| `connectOrCreate` | unique key | one FK, or a whole row through the child's `create` | supported |
| `create` / `createMany` | new rows | whole rows, `onCreate` applies | supported |
| `set` / `disconnect` / `delete` / `deleteMany` | **no** — "whatever is there" | — | refused, half one |
| `updateMany` | no | caller columns | refused, both |
| `update` | **yes** | **caller columns** | refused, half two |
| `upsert` | yes | caller columns | refused, half two |

The per-entry reasons now say which half each fails, and `update`'s in particular — *"the child's `onUpdate` and the scope-escape guard have to run over the payload — a pass this does not have yet"* — names the missing machinery rather than the missing feature, which is the thing that tells a reader whether it is a wait or a rewrite.

"`set`, `disconnect`, `delete`, `deleteMany` … act on rows the *call* did not name, and 'whatever is there' has no lookup to hang a scope on" also retroactively explains #83's `set` fix: the answer there was to give it a lookup — read the links, filter through the child's `findMany`, delete only those — which is precisely making it name its rows. The rule and that fix agree, which is the sign it is the real boundary rather than a description of the current code.

942 tests pass. Nothing outstanding from me on this PR.

PR #95, opened 2026-07-28
Composite relations: every read surface joins on all of the fields

### LINKED ISSUE #67: ORM: multi-field (composite) relations in `include`

nstfkc:
Closes #67, whose groundwork — the refusal that reached every surface — was #79.

`@relation(fields: [tenantId, orderId], references: [tenantId, id])` is the tenant-scoped composite-key style, and an application using it could not `include` at all. `Link` now carries `parentFields` / `childFields` rather than one of each, and the six read surfaces widen with it.

## The refusal's structure is what made this small

#79 established that all seven surfaces resolve their link through **one function**, precisely so a one-field assumption could not be reached by any path. That property inverted cleanly: `correlate()` gaining a conjunction fixes three surfaces at once — a relation filter, a `_count` and an `orderBy` — because hoisting the correlation into one function is what #58 did.

`composite-relations.test.ts` predicted this in its own header — *"when the feature lands this file inverts: the refusals become assertions about emitted SQL, and the list of surfaces is already here"* — and it has, with the surface list reused rather than rewritten.

## Three decisions worth naming

- **An `OR` of `AND`s, not a tuple `in`.** Postgres has `(a, b) in ((…),(…))`; SQLite does not, so a tuple would mean a new dialect method and two spellings of one predicate. `OR: [{ a: 1, b: 2 }, …]` is a shape both dialects already compile, through the `where` compiler every other filter goes through — so it inherits the parameter accounting, the plan key and the injection rules instead of needing its own. It costs the same placeholders either way, so `ParameterLimitError` fires proportionally sooner on a composite relation. That is the honest behaviour: the ceiling is on placeholders and a composite key uses more of them.
- **The stitch key is `JSON.stringify` of the normalised values.** Not concatenation — `("ab", "c")` and `("a", "bc")` are the same string, and these are user data, so parents would collect each other's children *silently*, since both queries succeed. `keyOf` already solved `Date` and `Bytes` for the single-field case, so this reuses it rather than restating it. The seed carries `(1, "a")` and `(1, "ab")` for exactly this, and there is a unit test with three parents across that boundary.
- **Nested writes still refuse**, through a `singleFieldLink` call rather than a shared resolver. Reading across a composite relation is a wider correlation; writing through one would contribute that many foreign-key columns to an insert, which is different work. The narrowing is a function nobody can reach past — there is no single-field property on `Link` to index into — which is #79's own argument applied one level down.

## Tests

**Ten differential cases** against Prisma on both dialects (batched on SQLite, lateral on Postgres), over a schema whose seed makes a *partial* join wrong rather than empty: the code `"a"` exists in two tenants, so joining on the code alone pulls in the other tenant's entries, and joining on the tenant alone pulls in `"ab"`'s. Either mistake returns rows, which is why these compare against Prisma rather than asserting counts. They cover to-many and to-one, a child `select` that omits both key fields, `some` / `none`, `_count`, `orderBy` through the relation, a `where` on the far side, and two levels of nesting.

**Plan-cache discrimination**, as #67 asks: two relations differing only in field *order* compile differently — which also catches a by-name pairing, since the sides pair positionally. Plus a length mismatch reported as a `MalformedRelationError` naming both sides, because that is the artifact disagreeing with the schema rather than a query anyone can fix.

## One thing found on the way

The differential harness's truncation list did not know about the new models. That is **silent on SQLite**, where every run gets a fresh temp database, and **loud on Postgres**, where the second run's seed collides with the first's rows — which is how I found it. Second time that has bitten this suite (the first was `Post` / `Tag` on #83), so the list now carries a comment saying so.

## Verification

931 unit tests. Template suite green on SQLite (501) and Postgres 16 (752), **identical across two consecutive Postgres runs** against one database, which is the check the isolation bug demands. `tsc` clean; oxlint clean.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -293,6 +293,19 @@ await List.create({
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
 | `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
 
+**A relation may join on more than one field.** `@relation(fields: [tenantId, orderId], references:
+[tenantId, id])` is the tenant-scoped composite-key style, and every read surface handles it: an
+`include` under either strategy, a relation filter, a `_count`, and an `orderBy` through the
+relation. The correlation becomes one equality per field; the batched strategy filters its children
+with an `OR` of `AND`s rather than a tuple `in`, because that is one shape both dialects already
+compile.
+
+Two consequences worth knowing. A composite join uses one placeholder *per field per parent*, so
+SQLite's parameter ceiling arrives proportionally sooner on a wide `include` — `ParameterLimitError`
+still names it rather than letting the driver fail. And a **nested write** through a composite
+relation is refused: it would have to contribute that many foreign-key columns to the insert, which
+is not implemented. Write the child separately with its keys set.
+
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
 child holds it, nothing can be written until this row exists, so those run after — which is why the
--- a/packages/gemi/orm/compile/composite-relations.test.ts
+++ b/packages/gemi/orm/compile/composite-relations.test.ts
@@ -12,23 +12,23 @@ import { compileWrite } from "./write";
 /**
  * Multi-field relations — `@relation(fields: [a, b], references: [c, d])`.
  *
- * Not implemented (#67), and the point of this file is the *shape* of the
- * refusal rather than the feature:
+ * **This file has inverted, exactly as it said it would.** It used to assert
+ * that seven surfaces refuse a composite relation, on the reasoning that a
+ * one-field assumption reached by any path would join on the first field and
+ * silently return the wrong rows. #67 implements six of the seven, so those six
+ * now assert the SQL — *every* joined field appears in the correlation, which is
+ * the same property from the other side.
  *
- * 1. **Every surface refuses.** A one-field assumption reached by any path
- *    would join on the first field and silently return the wrong rows — the
- *    worst failure available here, because the query succeeds. Seven surfaces
- *    correlate over a relation, and each has its own compiler; they are safe
- *    only because all seven resolve their link through one function. That is a
- *    property to hold, not a coincidence to rediscover.
+ * The seventh is a nested write, and it still refuses: reading across a
+ * composite relation is a wider correlation, writing through one would have to
+ * contribute that many foreign-key columns to an insert. The refusal is now a
+ * `singleFieldLink` call rather than a shared resolver, so the narrowing is a
+ * function nobody can reach past — there is no single-field property on `Link`
+ * to index into.
  *
- * 2. **Each names the operation it came from.** They all used to report
- *    `(Model.include)` — a composite relation named in a `where`, in an
- *    `orderBy`, or in a nested `connect` on a `create` sent the reader to an
- *    `include` they had not written.
- *
- * When the feature lands this file inverts: the refusals become assertions
- * about emitted SQL, and the list of surfaces is already here.
+ * What has *not* changed is the reason the list is exhaustive: all seven still
+ * resolve their link through one function, so a new surface inherits composite
+ * support rather than having to grow it.
  */
 
 const sqlite = new SqliteDialect();
@@ -42,19 +42,22 @@ beforeEach(() => {
 
 afterEach(() => registry.clearRegistry());
 
-/** `[surface, run, model, operation]` — the seven paths that correlate. */
-const SURFACES: [string, () => unknown, string, string][] = [
+/**
+ * `[surface, run, fragments]` — the read paths that correlate, and the SQL each
+ * has to contain. Every one names *both* joined fields.
+ */
+const READ_SURFACES: [string, () => unknown, string[]][] = [
   [
-    "include, to-many",
+    "include, to-many (batched)",
     () => compileRead(ledger, "findMany", { include: { entries: true } }, sqlite),
-    "Ledger",
-    "findMany",
+    // The batched strategy adds nothing to the root statement; it selects both
+    // key fields so the second query can be filtered and stitched.
+    [`"tenantId"`, `"code"`],
   ],
   [
-    "include, to-one",
+    "include, to-one (batched)",
     () => compileRead(ledgerEntry, "findMany", { include: { ledger: true } }, sqlite),
-    "LedgerEntry",
-    "findMany",
+    [`"tenantId"`, `"ledgerCode"`],
   ],
   [
     "include under the lateral strategy",
@@ -66,8 +69,10 @@ const SURFACES: [string, () => unknown, string, string][] = [
         postgres,
         lateralStrategy,
       ),
-    "Ledger",
-    "findMany",
+    [
+      `"tenantId" = "Ledger"."tenantId"`,
+      `"ledgerCode" = "Ledger"."code"`,
+    ],
   ],
   [
     "a relation filter",
@@ -78,8 +83,10 @@ const SURFACES: [string, () => unknown, string, string][] = [
         { where: { entries: { some: { amount: 1 } } } },
         sqlite,
       ),
-    "Ledger",
-    "findMany",
+    [
+      `"_r0"."tenantId" = "Ledger"."tenantId"`,
+      `"_r0"."ledgerCode" = "Ledger"."code"`,
+    ],
   ],
   [
     "_count",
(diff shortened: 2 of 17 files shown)


nstfkc:
## Review — composite relations across the read surfaces (closes #67)

Read the full diff against the merge base (`d38b367`). 959 tests pass.

#79's structure paid off exactly as it predicted. Widening `correlate()` fixes three surfaces at once, and I checked the emitted SQL rather than the tests — all four correlated forms carry the conjunction:

```
relation filter   … exists (select 1 from "Ledger" as "_r0"
                    where "_r0"."tenantId" = "LedgerEntry"."tenantId"
                      and "_r0"."code"     = "LedgerEntry"."ledgerCode")
_count            … (select count(*) from "LedgerEntry" as "_c0"
                    where "_c0"."tenantId" = "Ledger"."tenantId"
                      and "_c0"."ledgerCode" = "Ledger"."code")
orderBy through   … (select "_o0"."code" from "Ledger" as "_o0"
                    where "_o0"."tenantId" = … and "_o0"."code" = …)
include (batched) … select "tenantId", "code", "title" from "Ledger"   both key fields
```

The `JSON.stringify(values.map(keyOf))` stitch key is the decision I would have most wanted argued, and it is: concatenation makes `("ab","c")` and `("a","bc")` the same string, both queries still succeed, and parents collect each other's children silently. Seeding `(1,"a")` and `(1,"ab")` so the fixture straddles that boundary is what makes the test discriminate rather than merely pass.

Narrowing nested writes through `singleFieldLink` — rather than a shared resolver with a single-field property to index into — is #79's own argument applied one level down, and it is the right shape.

### The composite child query reintroduces plan-cache churn that the single-field path was fixed to avoid

`plan.ts`'s `collapsedList` exists for this exact sentence, in its own words:

> every batched relation query is an `in` over the parent keys, so the list length is **the number of rows the parent query returned**, and a variably-sized parent set would churn the LRU

That collapse applies to `in` / `notIn` — `LIST_KEYS` — and the composite path builds an `OR` instead, so it does not reach it. Measured over four parent counts (2, 3, 10, 50):

```
sqlite    composite: 4 distinct keys      single: 4 distinct keys
postgres  composite: 4 distinct keys      single: 1 distinct key
```

So on Postgres a single-field `include` over a variable parent set is one plan forever, and a composite one is a plan per page size. Same query shape, same include, different relation arity. With the cache capped at 1000 and the key varying with a result-set size the caller does not control, that is the churn the collapse was written to prevent, arriving through the door it does not cover.

SQLite churns either way and always has — *"a coarser key cannot fix the SQLite side"* is already recorded. The new thing is the Postgres asymmetry.

I do not think collapsing the `OR` is the fix: unlike `= any($1)`, its text genuinely varies with length, so a shared key would hand a plan the wrong number of placeholders — the same trap the comment describes for SQLite. Which means the honest options are to accept and document it beside `collapsedList` (where a reader already goes to understand this), or to find a form where the parent keys ride in one parameter. The PR's decision section covers the *parameter ceiling* consequence of `OR` carefully and does not mention the *plan cache* one; given how much machinery here exists for the latter, it deserves a line either way.

### A smaller question about the same choice

The stated reason for `OR`-of-`AND`s over a tuple `in` is that a tuple *"would mean a new dialect method and two spellings of one predicate."* True — but `SqlDialect.inList` is already exactly that: a method that exists because *"SQLite and Postgres disagree about structure, not just spelling"*, with `= any($1)` on one side and expanded placeholders on the other. The seam is there and this is the case it was built for.

That does not make `OR` wrong — it is portable, it reuses the `where` compiler, and it inherits the parameter accounting and injection rules, which is worth a lot. But an N-branch `OR` and a tuple `IN` are not equivalent to a query planner on large parent sets, and the reason given rules out the alternative on a cost the architecture has already paid elsewhere. Worth restating as "portability and reuse beat plan quality here" rather than "a dialect method would be new", since the first is defensible on its own and the second is not quite accurate.

Nothing blocking. The surface coverage is the substance and it holds.

nstfkc:
Pushed as 4ed5a3e. Both findings are right, and the second one is a correction I should have made myself.

### The churn

Reproduced your measurement exactly:

```
sqlite    composite: 4 distinct keys    single: 4 distinct keys
postgres  composite: 4 distinct keys    single: 1 distinct key
```

And your reasoning about the fix holds: collapsing the `OR` would hand a plan the wrong number of placeholders, which is the trap `plan.ts` already describes for SQLite's expanded `in`. So I took the first of your two options, and put the note in **both** places a reader arrives from — beside `collapsedList`, which is where someone goes to understand this, and at the call site that builds the `OR`.

I also pinned it as a test rather than only a comment, because the *asymmetry* is the surprising half: same include, same query shape, different relation arity. A future change that closes it should have to update an assertion rather than quietly agree.

Filed as **#97** with the form that would actually fix it — `(a, b) in (select * from unnest($1::int[], $2::text[]))`, constant text regardless of parent count, so one plan. It wants the column types threaded to that call site and its own coverage on both dialects, which is why it is an issue rather than a commit here.

### The reason I gave for `OR` over a tuple `in`

You are right and I was not. I wrote that a tuple form "would mean a new dialect method and two spellings of one predicate" — and `SqlDialect.inList` **is** that method, existing precisely because the two databases disagree about structure rather than spelling. The seam is there and this is the case it was built for.

Restated as what it actually is: portability and reuse beat plan quality here. An `OR` inherits the parameter accounting, the plan key and the injection rules from the `where` compiler every other filter goes through, and an N-branch `OR` is not equivalent to a tuple `IN` to a planner on a large parent set — that is the cost being paid. Defensible on its own; "the seam does not exist" was not, and it would have sent the next reader looking for a constraint that is not there.

932 unit tests, template suite green on SQLite (501).

nstfkc:
### Re-reviewed `4ed5a3e` — documented in both places, and the follow-up is real

The behaviour is deliberately unchanged, which is the right call: my own reading was that collapsing the `OR` would be *wrong* rather than merely hard, since its text varies with branch count and a shared key would hand a plan the wrong number of placeholders. Re-measured on the new head, identical to before:

```
sqlite    composite: 4 keys    single: 4 keys
postgres  composite: 4 keys    single: 1 key
```

932 tests pass.

What changed is where a reader meets it. The note now sits beside `collapsedList` — *"It covers `in` / `notIn` and nothing else, which leaves one gap worth naming here rather than only where it happens"* — which is the function someone opens precisely when they are wondering why the cache is churning. Putting the measurement there rather than only at `childQuery` is what makes it findable from the symptom rather than from the cause.

**#97 is filed** — *"A composite relation's batched `include` churns the plan cache on Postgres"* — with the `unnest` form named as the candidate and the reason it is not guessed at here (it needs the column types and its own coverage on both dialects). I checked, because a code comment citing an issue number is worth exactly as much as the issue existing; my first lookup returned nothing and I re-ran rather than reporting it, which is the correct outcome of that particular near-miss.

The justification is also corrected accurately: *"the trade is portability and reuse against plan quality — not the absence of a seam"*, with `SqlDialect.inList` named as the seam that already exists. That was the smaller half of what I raised and it is the half that would have misled the next person deciding whether a tuple form is available.

PR #96, opened 2026-07-28
Nested `disconnect` and `delete`, which name the rows they touch

nstfkc:
#65's item three, minus the operands that do not fit the line this file draws. **Stacked on #94**, where that line is restated — hence the base, so the diff here is only the two new operands.

## Why these two and not the other three

#94's review established the boundary: every supported operand **names its rows**, so it goes through the child's own operations and the child's policies decide reachability. `set`, `deleteMany` and `updateMany` act on rows the *call* did not name — "whatever is there" has no lookup to hang a scope on — so they stay refused.

`disconnect` and `delete` both take a unique key. That is what makes them the tractable half, and it is a good sign the restated criterion is doing work rather than describing the past.

## The two differ on an unlinked row, and it is Prisma's difference

Measured before implementing:

```
disconnect a row linked elsewhere  ->  succeeds, changes nothing
delete     a row linked elsewhere  ->  raises "are not connected"
```

So both filter on the parent's key as well as the caller's, and only `delete` treats a miss as an error. That filter does two jobs at once: it reproduces the semantics, and it stops either operand from reaching a row belonging to a different parent.

Three refusals, each measured rather than assumed:

- **Under `create`** — Prisma reports them as an unknown argument, and it is right to: a row that does not exist yet has nothing linked to it.
- **`disconnect` on a required foreign key** — there is no value to leave behind. Named by column, so the message is not a not-null violation from inside a nested step.
- **`delete` through a to-one** — deleting the far row as a side effect of an unrelated update is a thing to implement deliberately, not to guess.

## Scoping

The child's own operations, as everywhere here: `updateMany`, `findFirst` and `deleteMany` all go through the child's `$exec` un-pre-scoped. A hidden row is reported by `delete` as **not connected** rather than as denied — the same answer a genuinely unlinked row gets, so the two are indistinguishable and neither confirms the row exists.

## I had the differential cases wrong first, and the way I found out is the point

They set their fixtures up in the test body. `expectSameWrite` **resets to the seeded state before each client runs** — that is the whole point of it — so the rows were gone by comparison time and every case ran against an empty `Account` table, passing whatever the code did.

I found it by removing the parent-key filter to check the cases discriminated, and watching them all still pass. The rows are in the seed now, two of them on a *different* user, and with the filter removed:

```
× disconnecting a row that is not linked is a no-op
× deleting a row that is not linked raises, as Prisma does
```

The seed carries a comment saying why, because the next case needing a fixture would otherwise be written the same wrong way.

## Tests

Six differential cases, three policy tests (including the hidden-row-reports-not-connected one), and the existing refusal table updated.

## Verification

942 unit tests. Template suite green on SQLite (507) and Postgres 16 (762), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

## One dependency worth naming

The policy tests pass a no-op scalar (`code: "ours"`) alongside the nested operand, because `Folder` has no `@updatedAt` and a relation-only `update` is refused on this base with *"At least one field must be updated"*. #83 makes that compile to a select of the same returning list; the template's `User` already has `@updatedAt`, which is why the differential cases need no such thing.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -293,6 +293,8 @@ await List.create({
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
 | `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
 | `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
+| `disconnect` | Clears the link. `true` on a to-one; a unique key, or a list of them, on a to-many. The column has to be nullable. |
+| `delete` | Deletes the named rows outright, not just the link. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -314,8 +316,15 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
-Everything else in Prisma's nested grammar — `set`, `disconnect`, `update`, `updateMany`, `upsert`,
-`delete`, `deleteMany` — is refused, by name and with the reason. The line is **which rows an
+`disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
+Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
+this parent, and the difference is Prisma's: `disconnect` succeeds and changes nothing, `delete`
+raises. Both filter on the parent's key as well as yours, so neither can reach a row belonging to
+somebody else — and a row your policies hide is not reachable either, which `delete` reports as "not
+connected" rather than as denied, since that is the same answer a genuinely unlinked row gets.
+
+Everything else in Prisma's nested grammar — `set`, `update`, `updateMany`, `upsert`, `deleteMany` —
+is refused, by name and with the reason. The line is **which rows an
 operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
 one you identified by unique key) and writes either a whole new row or one foreign key the ORM
 chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -316,6 +316,12 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
+> **Known gap (#98).** A child whose policy *scopes on the foreign key itself* — `scope: () => ({
+> folderId: mine })` — cannot use any relation operand that writes that key, including `connect`,
+> unless it also declares an `onUpdate`. The scope-escape guard reads the payload and cannot tell a
+> column you supplied from one the nested step put there. Pre-existing rather than new to
+> `disconnect`, and tracked separately because the decision affects `connect` too.
+
 `disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
 Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
 this parent, and the difference is Prisma's: `disconnect` succeeds and changes nothing, `delete`
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -72,7 +72,22 @@ const SUPPORTED = new Set([
   "connectOrCreate",
   "create",
   "createMany",
+  "disconnect",
+  "delete",
 ]);
+
+/**
+ * Operands that only exist on a statement with a row to act on.
+ *
+ * Prisma refuses `disconnect` and `delete` under a `create` — *"Unknown
+ * argument `disconnect`"* — and it is right to: there is nothing linked to a
+ * row that does not exist yet. Refused here rather than accepted and made a
+ * no-op, so a caller who wrote one on the wrong operation hears about it.
+ */
+const EXISTING_ROW_ONLY = new Set(["disconnect", "delete"]);
+
+/** Statements that insert a new row, so nothing is linked to it yet. */
+const CREATE_ONLY_STATEMENTS = new Set(["create", "createMany"]);
 
 /**
  * The operands still refused, and what each would take.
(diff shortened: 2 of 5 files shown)


nstfkc:
## Review — nested `disconnect` and `delete` (#65, item three)

Read the diff against its true base (`8f1b9e3`, #94) — 5 files, +455. 942 tests pass.

Choosing these two *because* they name their rows is the strongest thing here: #94's restated criterion immediately decided which half of item three was tractable, and it decided it correctly rather than being reverse-engineered from what was easy. A rule that does work on the next decision is a rule; one that only describes the last one is a summary.

The parent-key filter doing two jobs at once — reproducing Prisma's *"succeeds and changes nothing"* / *"raises are not connected"* split **and** preventing either operand from reaching another parent's row — is the kind of thing that usually needs two mechanisms.

I checked the guard I most expected to be missing, since both operands compile to `updateMany` / `deleteMany` and a non-unique filter there would silently hit many rows:

```
delete by unique id        ACCEPTED
disconnect by unique id    ACCEPTED
delete by NON-unique col   REFUSED   (Account.update.accounts.delete)
disconnect NON-unique col  REFUSED   (Account.update.accounts.disconnect)
under create               REFUSED
delete through a to-one    REFUSED
disconnect required FK     REFUSED
```

`matchUniqueKey` per entry at plan time, with the operand in the reported path. That is the check that makes `deleteMany` safe to use here.

**The fixture bug is the most valuable paragraph in the description.** `expectSameWrite` resetting to the seeded state before each client is exactly the sort of harness property that makes a test vacuous rather than failing, and finding it by *removing the parent-key filter and watching every case still pass* is the only method that would have. Putting the reason in the seed so the next fixture is not written the same way is the part that generalises.

### `disconnect` trips the scope-escape guard on a column the caller never wrote

A child whose policy scopes on the foreign key — a plausible "my accounts" scope — and which has no `onUpdate`:

```
scope on userId, no onUpdate  ->  ScopeEscapeError: Account.updateMany writes 'userId',
                                  which Account's policy also scopes on…
scope on userId, w/ onUpdate  ->  OK  { userId: null }
```

The caller wrote `disconnect: { id: 1 }`. `userId` is in `data` because *this step put it there*. `assertNoScopeEscape` reads `args.data` and cannot tell a column the caller supplied from one the ORM chose, so the error describes a write the caller did not make — the same shape as the `create`-scoped-as-insert message that #85 was filed for, and the `Account.create` one before it.

Two readings, and I do not think it is obvious which is right:

- **The guard is correct and the message is wrong.** Disconnecting genuinely moves the row out of the scope that selected it, which is what the guard exists to catch. Then the fix is wording: say that this `disconnect` clears a column the policy scopes on, and that an `onUpdate` is how the policy declares that intended — which is precisely what `softDeletes` already does for `deletedAt`, and the comment there says so.
- **The guard should not fire at all**, because its purpose is to stop a *caller* moving a row between tenants, and here the only column written is the one the ORM itself chose. That reading needs the step to mark its payload as ORM-authored, which is machinery this does not have.

Either way it is worth deciding rather than discovering: a model with that policy shape loses `disconnect` entirely today, and the error sends the reader to their own `data`, where there is nothing to find.

Nothing else outstanding. The three refusals are each measured against Prisma and each say what to do instead.

nstfkc:
Pushed as e77cdee. The finding is right and the diagnosis is exact — and it turns out to be **pre-existing**, which changed what I did about it.

`connect` has the same shape and the same failure on `feat/orm` today. Measured against a child scoped on `folderId` with no `onUpdate`:

```
connect     (already shipped)  ScopeEscapeError: Note.update writes 'folderId'…
disconnect  (this PR)          ScopeEscapeError: Note.updateMany writes 'folderId'…
```

That matters for your framing — *"a model with that policy shape loses `disconnect` entirely today"* — because it loses `connect` too, and has since that operand merged. So the decision is about an operand already in `feat/orm`, and settling it inside a PR that adds two more would settle it by accident.

**Filed as #98** with both of your readings and a recommendation. I lean to the second: the guard is answering a question about *caller-supplied* data, and a relation key is not that — the value is the ORM's, the parent was reached through the parent model's own scoping, and a child row the caller cannot see is already refused before this runs. What is left is the operand doing exactly what it means. The fallback, if the guard stays, is your first reading: a message that names the operand and says an `onUpdate` is how a policy declares this intended, which is what `softDeletes` already does for `deletedAt`.

Pinned rather than left as prose, and pinned across **both** operands, so the test records this as a property of the guard rather than of `disconnect`. When #98 lands it should flip to expecting success rather than being deleted — same shape as the churn pin on #95. `docs/orm.md` carries it as a known gap, since a model with that policy shape meets it before it meets the issue.

### On the guard you went looking for

`matchUniqueKey` per entry at plan time was the thing I most wanted checked, for the reason you name: both operands compile to `updateMany` / `deleteMany`, and a non-unique filter there hits many rows silently. Good to have it verified independently rather than from my tests.

942 unit tests, template suite green on SQLite (508).

### One disagreement, on #94's summary table

Your re-review there put `set` / `disconnect` / `delete` / `deleteMany` in one row as "**no** — whatever is there". That is right for `set` and `deleteMany`, and not for the other two: Prisma's to-many `disconnect` and `delete` take a *unique key*, which is exactly why this PR could implement them and why the criterion picked them out. Worth splitting that row if the table is ever used as the reference — as written it would tell a future reader that these two are out of reach for a reason that does not apply to them.

nstfkc:
### Re-reviewed `e77cdee` — and my framing was narrower than the bug

I raised this as an interaction with the operand this PR adds. You establish it is not, and you are right. Checked on merged `feat/orm`, with a child scoped on its own foreign key and no `onUpdate`:

```
connect     (data: { userId: 7 })     ->  ScopeEscapeError
disconnect  (data: { userId: null })  ->  ScopeEscapeError
ordinary    (data: { publicId: … })   ->  OK
```

`connect` shipped before this PR and fails identically, so it is a property of `assertNoScopeEscape` reading `args.data` without provenance — not of `disconnect`. Presenting it as new to this operand would have sent the fix to the wrong place, and pinning **both** operands in the test is what records that.

The handling is better than either reading I offered. I framed it as a choice between "guard right, message wrong" and "guard should not fire"; **pinning the current behaviour as deliberate while the decision is made elsewhere** is the third option, and the right one for a PR that did not cause it. Two details worth keeping:

- *"Pinned, not endorsed"*, with the instruction that the test should **flip to expecting success** when #98 lands rather than being deleted. A test asserting a known-wrong behaviour is a liability unless it says what should happen to it, and this one does.
- The docs entry says **which operands are affected** — `connect` included — rather than only the one in front of it. That is what stops the next reader concluding `connect` is safe.

#98 is filed and labelled, titled for the mechanism rather than the symptom: *"`assertNoScopeEscape` fires on relation keys the ORM wrote, not the caller"*. That is the right level — the fix is provenance on the payload, and both operands fall out of it.

17 nested-write policy tests pass. Nothing outstanding from me on this PR.

PR #99, opened 2026-07-28
The scope-escape guard judges the caller's columns, not the ORM's

### LINKED ISSUE #98: `assertNoScopeEscape` fires on relation keys the ORM wrote, not the caller
Found in review of #96. **Pre-existing — it is not introduced by that PR**, which is the first thing worth establishing, because the review framed it as a cost of the new operands.

## The failure

A child whose policy scopes on the foreign key — a plausible "my notes" scope — and which declares no `onUpdate`:

```ts
class Note extends Model {
  static $policies = [{ scope: () => ({ folderId: 1 }) }]
}
```

Every nested relation operand that writes that key is refused:

```
connect     (on feat/orm today)  ScopeEscapeError: Note.update writes 'folderId',
                                 which Note's policy also scopes on…
disconnect  (#96)                ScopeEscapeError: Note.updateMany writes 'folderId', …
```

The caller wrote `connect: { id: 1 }` or `disconnect: { id: 1 }`. `folderId` is in `data` because **the nested step put it there** — it is the relation's own key, and its value is the parent this statement is about. `assertNoScopeEscape` reads `args.data` and cannot tell a column the caller supplied from one the ORM chose, so the error describes a write the caller never made and sends them to their own `data`, where there is nothing to find.

That is the same shape as #85 (a nested `include` under `create` reporting an `Account.create` the caller had not asked for) and the `Account.create` message before it: a refusal that misnames its own origin.

## Why this is not simply "the guard is right"

It has a real reading in its favour: disconnecting *does* move the row out of the scope that selected it, which is what the guard exists to catch, and `softDeletes` already shows how a policy declares such an intent — it carries an `onUpdate` for exactly this reason, and says so.

But the guard's stated purpose is to stop a **caller** moving a row between tenants by naming a scoped column in `data`. A relation operand's key is not that: its value is chosen by the ORM, the parent was reached through the parent model's own scoping, and the child row was reached through the child's — an unreachable row is already refused before this. What is left is the operand doing precisely what it means.

The practical cost is that a model scoped on its foreign key loses `connect`, `connectOrCreate`, `disconnect` and `delete`'s sibling paths entirely, unless it adds an `onUpdate` whose only job is to satisfy a guard about a column it never writes.

## The two options, as the review framed them

1. **Keep the guard, fix the message.** Say that this operand clears or repoints a column the policy scopes on, and that an `onUpdate` is how a policy declares that intended. Cheap, and honest about what is happening.
2. **Do not fire for ORM-authored payloads.** Needs the nested step to mark its payload as such — machinery that does not exist. `markPreScoped` is the precedent for carrying ORM intent through `$exec`'s options, so the seam is not far away.

I lean to (2) with (1) as the fallback: the guard is answering a question about caller-supplied data, and this payload is not that. But it is a policy-layer decision that affects an operand which already shipped, so it belongs in its own change rather than inside a PR adding two more.

## Acceptance

- A model scoped on its foreign key can `connect` and `disconnect` without declaring an `onUpdate` — or, if (1) is chosen, is told precisely why it must.
- A caller naming a scoped column in `data` themselves is still refused, with a test that fails if the distinction is lost.
- `nested-write-policies.test.ts` pins the current behaviour in *"a foreign-key scope refuses relation operands today"*; that test should flip rather than be deleted.

nstfkc:
Closes #98. Found in review of #96, but **not caused by it** — `connect` has had this since it merged, which is why the fix belongs here on `feat/orm` rather than inside a PR adding more operands.

## The failure

A child whose policy scopes on its **own foreign key** — `{ folderId: mine }`, a plausible "my notes" scope — and which declares no `onUpdate`:

```
connect     ScopeEscapeError: Note.update writes 'folderId', which Note's policy also scopes on…
disconnect  ScopeEscapeError: Note.updateMany writes 'folderId', …          (#96)
```

The caller wrote `connect: { id: 1 }`. `folderId` is in `data` because **the nested step put it there** — its value is the parent this statement is about. `assertNoScopeEscape` reads `args.data` and cannot tell that from a caller naming the same column, so the error describes a write the caller never made and sends them to their own payload, where there is nothing to find. Same shape as #85's `Account.create` message: a refusal that misnames its own origin.

## Which reading, and why

Of the two the review offered, this takes "the guard should not fire". Its stated purpose is to stop a **caller** moving a row out of the scope that selected it by naming a scoped column in `data`. A relation key is not that:

- the value is the ORM's, not the caller's;
- the parent was reached through the parent model's own scoping;
- a child row the caller cannot see is **already refused** by the child's scope before this runs — a different mechanism, untouched here, and pinned by the existing tests.

What is left is the operand doing exactly what it means.

## The mechanism follows `markPreScoped` exactly

A module-private `Symbol`, **not exported from `orm/index.ts`**, so an application cannot forge it. An `ExecOptions` field would have been a public door for writing a scoped column past the guard — the thing iteration 6 spent three review rounds closing.

The new parameter is optional, and the direction is the point: **omitting it makes the guard stricter**, so a call site that forgets is a refused query rather than a silent escape. That is the opposite of #79's defaulted `operation`, where the default was the wrong answer — and it is why this one is allowed a default at all. The comment says so, because "some defaults are fine" is exactly the reasoning that erodes.

## Tests

Two, both directions, and the first verified red by removing the marker:

- a foreign-key-scoped child **can** now be connected, and the row lands on the right parent;
- a caller naming that same column in `data` is **still refused**.

The second is the one that matters: the marker lists the columns the ORM wrote, so anything else in the payload is judged exactly as before.

## Merge-order note

#96 pins the current refusal across `connect` *and* `disconnect`, deliberately, so that the behaviour was deliberate rather than accidental. Whichever lands second, that pin flips to expecting success — it was written to flip rather than to be deleted.

## Verification

935 unit tests. Template suite green on SQLite (493) and Postgres 16 (748), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -305,6 +305,10 @@ Three things follow from that, and all three are load-bearing:
 - **The child's policies apply.** Each nested row goes through the related model's own `$exec`, so
   its `onCreate` stamps the tenant column and its `scope` decides which rows a `connect` can reach.
   A `createMany` writes its rows in one statement and they are *each* scoped.
+  A relation operand writes one column — the relation's own key — and the
+  scope-escape guard knows the ORM wrote it, so a child scoped on its foreign
+  key can still be connected. A column *you* name in `data` is judged exactly as
+  before.
 - **A failure anywhere rolls the whole thing back**, including the parent row.
 
 Everything else in Prisma's nested grammar — `connectOrCreate`, `set`, `disconnect`, `update`,
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -40,7 +40,9 @@ import {
   applyRedaction,
   currentUser,
   isPreScoped,
+  markOrmAuthored,
   markPreScoped,
+  ormAuthoredFields,
   policiesFor,
   policyContext,
   type PolicyEntry,
@@ -582,7 +584,12 @@ export abstract class Model {
       // the same predicate twice; re-running `redact` on an already-redacted row
       // is a no-op.
       if (!preScoped) {
-        effective = applyPolicies(policies, policy, args);
+        effective = applyPolicies(
+          policies,
+          policy,
+          args,
+          ormAuthoredFields(options),
+        );
       }
     }
 
@@ -643,16 +650,20 @@ export abstract class Model {
       // grandchild folded anyway and the statement count was one lower than the
       // caller asked for. Found by the query-count test, which is the only thing
       // that could have found it: the results were identical either way.
-      exec: (model, operation, relationArgs, preScoped) =>
-        registry
+      exec: (model, operation, relationArgs, preScoped, ormAuthored) => {
+        const base = preScoped
+          ? markPreScoped({ strategy: options?.strategy })
+          : { strategy: options?.strategy };
+        return registry
           .get<typeof Model>(model)
           .$exec(
             operation as Operation,
             relationArgs,
-            (preScoped
-              ? markPreScoped({ strategy: options?.strategy })
-              : { strategy: options?.strategy }) as never,
-          ),
+            (ormAuthored && ormAuthored.length > 0
+              ? markOrmAuthored(base, ormAuthored)
+              : base) as never,
+          );
+      },
       // The one query with no model behind it — the implicit m-n join table —
       // resolves its connection here rather than reaching for the pool, so it
       // joins the transaction like everything else.
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -490,6 +490,10 @@ function planForeignSide(
           // to the child, and the child's scope decides which rows are reachable
           // — otherwise `connect` re-parents another tenant's row.
           false,
+          // ...and the column being written is *ours*, not the caller's. Without
+          // this, a child whose policy scopes on its foreign key is refused for
+          // a write it never made — see #98 and `ormAuthoredFields`.
+          [childField],
         );
       }
     },
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -113,6 +113,17 @@ export interface RelationExecutor {
     op: string,
     args: unknown,
     preScoped: boolean,
+    /**
+     * Columns of `args.data` this step wrote itself — a relation operand's own
+     * foreign key, whose value is the parent this statement is about.
+     *
+     * Only the scope-escape guard reads it, so that a child scoped on its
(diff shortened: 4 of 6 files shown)


nstfkc:
## Review — the scope-escape guard judges the caller's columns (closes #98)

Read the full diff against the merge base (`d38b367`). 935 tests pass.

This is a change to the policy layer, so I checked the bypass surface before anything else.

**The marker cannot be forged.** `Symbol()` rather than `Symbol.for()`, so it is not in the global registry, and it is not re-exported from `orm/index.ts`:

```
orm/index exports markOrmAuthored : undefined
global symbol registry hit        : false
```

Both properties match `markPreScoped`, which is the right precedent — and the reasoning for not putting it on `ExecOptions` is the same one iteration 6 spent three rounds on.

**The exemption is per column, not a blanket bypass.** The case I most wanted to fail is the fifth:

```
1 ORM wrote folderId (connect)        OK
2 caller wrote folderId, unmarked     ScopeEscapeError
3 caller wrote folderId + other col   ScopeEscapeError
4 ORM wrote folderId, caller title    OK
5 marked title, caller wrote folderId ScopeEscapeError
```

Marking one column does not exempt another, so a step that marks what it wrote cannot accidentally cover what the caller wrote alongside it.

**Row-level scoping is untouched**, which is the property that had to survive. The `preScoped` argument threads through exactly as before — `base` is still computed from it and `markOrmAuthored` layers on top — so a child row the caller cannot see is still refused by the child's own scope, by a different mechanism, before this guard runs at all. Separating "which rows are reachable" from "which columns may be written" is what makes the relaxation safe rather than a hole.

And the two cannot collide by name: `insertColumns` already refuses `'X' is set both directly and through a nested relation write`, so a caller cannot smuggle a column in under an ORM-authored name in the same payload. That pre-existing guard is load-bearing for this one and is worth knowing about together.

**The optional parameter is argued rather than assumed.** `preScoped` was deliberately made required because *"the two callers need opposite answers and the wrong one is silent in both directions"*; this one defaults, and the difference is real — omitting it means the columns are not exempted, so the guard fires and the query is refused. It fails closed. Saying so explicitly, rather than leaving "some defaults are fine" to be inferred, is the part that keeps #79's rule from eroding by analogy.

The reading chosen is the one I would have argued for, and the justification is better than the one I offered: not merely that the value is the ORM's, but that the parent was reached through the parent model's own scoping and the child row through the child's — so by the time this guard runs, both halves have already been decided elsewhere. That is what makes "the operand doing exactly what it means" the whole of what is left.

Nothing outstanding. The merge-order note is right too — #96's pin was written to flip to expecting success, and this is what flips it.

PR #100, opened 2026-07-28
A composite `include` is one plan on Postgres, through `unnest`

### LINKED ISSUE #97: A composite relation's batched `include` churns the plan cache on Postgres
Found in review of #95, which implements composite relations across the read surfaces.

## The asymmetry

`collapsedList` exists so that a batched relation query — *"an `in` over the parent keys, so the list length is the number of rows the parent query returned"* — does not mint one plan per parent count on Postgres, where `= any($1)` makes every length one SQL text.

It covers `in` / `notIn`. A relation joining on **more than one field** cannot be filtered with a single `in`, so the batched loader builds an `OR` of `AND`s instead, and that does not reach the collapse. Measured over parent counts 2, 3, 10, 50:

```
sqlite    composite: 4 distinct keys    single: 4 distinct keys
postgres  composite: 4 distinct keys    single: 1 distinct key
```

SQLite churns either way and always has — `plan.ts` already records that *"a coarser key cannot fix the SQLite side"*. The new thing is that Postgres, which is otherwise immune, is not immune for a composite relation. Same include, same query shape, different relation arity.

With the cache capped at 1000 and the key varying with a result-set size the caller does not control, this is the churn `collapsedList` was written to prevent, arriving through the door it does not cover.

## Why the obvious fix is wrong

Collapsing the `OR` the way `collapsedList` collapses an `in` would hand a plan the wrong number of placeholders — an `OR`'s text genuinely varies with its branch count, unlike `= any($1)`. That is the same trap `plan.ts` describes for SQLite's expanded `in`.

## What would fix it

A form where the parent keys ride in a **fixed** number of parameters, one array per column rather than one placeholder per parent:

```sql
select … from "LedgerEntry"
where ("tenantId", "ledgerCode") in (
  select * from unnest($1::int[], $2::text[])
)
```

Constant text regardless of parent count, so one plan entry — and the arrays bind as parameters, so nothing about the injection rules changes.

This is a **dialect method**, not a plan-key change: `SqlDialect.inList` is the precedent, and it exists for exactly this kind of structural disagreement. SQLite has no `unnest` and would keep the `OR`, which is consistent with it already churning.

The reason it is filed rather than folded into #95: it needs each column's SQL type for the casts, which the compiler has (`FieldSchema.type`) but does not currently thread to that call site, and it wants its own coverage on both dialects — including that the `OR` path stays correct where it is still used.

## Acceptance

- A composite batched `include` over 2, 3, 10 and 50 parents produces **one** plan key on Postgres.
- The emitted statement binds one parameter per joined column, not per parent.
- SQLite is unchanged and still correct.
- `plan.discrimination.test.ts` already pins the current behaviour in *"a composite parent filter churns on both dialects, unlike a single-field one"* — that test should flip to asserting the fix rather than being deleted.

nstfkc:
Closes #97. **Stacked on #95**, which added composite relations and measured the churn this removes — hence the base, so the diff here is only the fix.

## The gap

`collapsedList` keeps a batched relation query to one plan entry however many parents it has, because on Postgres `= any($1)` gives one SQL text for every length. A relation joining on more than one field cannot use it, so the loader built an `OR` of `AND`s whose text grows — and the plan key grew with it:

```
before   sqlite   composite: 4 keys    single: 4 keys
         postgres composite: 4 keys    single: 1 key
after    postgres composite: 1 key
```

`(a, b) in (select * from unnest($1::int[], $2::text[]))` binds **one array per column** rather than one placeholder per parent, so the text is fixed. SQLite has no such form, keeps the `OR`, and churns as it always has — `plan.ts` already records that a coarser key cannot fix that side, and the internal key is never emitted there to be collapsed wrongly.

## Behind the seam the review pointed at

Your note on #95 was that `SqlDialect.inList` already exists *because the two databases disagree about structure rather than spelling* — and this is the same disagreement one arity up. So `canBindCompositeIn` / `compositeIn` sit beside it rather than becoming a branch in the compiler.

## Three decisions

- **The dialect is asked, not told.** `canBindCompositeIn` takes the column types; Postgres answers for `Int`, `BigInt`, `String` and `Boolean` only, and anything else keeps the `OR`. A cast that does not round-trip exactly is *silently wrong rows*, not an error, so the map is deliberately short. `DateTime` is the instructive absence: this driver's own note records that a zoneless timestamp already decodes differently depending on which protocol the statement used, and a text array literal would add a second representation question to one that is already open — for a key type nobody joins on.
- **`$compositeIn`, because a Prisma field name cannot start with `$`.** A `__` prefix could collide with a column; this cannot. It is internal — only `plan-relations.ts` builds one — and a caller writing it by hand meets the same unknown-key treatment as anywhere else, since no schema has that field.
- **The collapse takes the tuple list and not the fields.** The joined columns are *identifiers* in the emitted SQL, so two relations joining on different ones — or on the same ones in a different **order**, which pairs the sides differently — must stay different entries.

## Tests

The churn test #95 added was written to flip rather than be deleted, and this is the flip: it asserts one plan on Postgres and many on SQLite. A companion pins the half the collapse must not take with it — three distinct field lists produce three keys, on both dialects.

All ten composite differential cases pass on Postgres with `unnest` actually in use, and SQLite still emits the `OR`. I checked both by reading the child query the loader builds rather than inferring it from green tests:

```
postgres  child where: {"$compositeIn":{"fields":["tenantId","code"],"values":[[1,"a"]]}}
sqlite    child where: {"OR":[{"tenantId":1,"code":"a"}]}
```

## Verification

933 unit tests. Template suite green on SQLite (501) and Postgres 16 (752), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

### DIFF
--- a/packages/gemi/orm/compile/composite-relations.test.ts
+++ b/packages/gemi/orm/compile/composite-relations.test.ts
@@ -279,6 +279,56 @@ describe("plan discrimination", () => {
   });
 });
 
+/**
+ * `$compositeIn` is internal, and that is enforced rather than claimed.
+ *
+ * The branch that compiles it sits **above** the field lookup in
+ * `compileWhere`, so a caller writing the key by hand used to be honoured. Not
+ * an injection — the fields are resolved against the schema, the columns go
+ * through `quoteIdent`, the values are bound — but an undocumented input
+ * surface with none of the operand validation its neighbours have, and one
+ * shape that compiled clean and deferred its failure to bind time.
+ *
+ * The operand now has to carry a module-private `Symbol` the planner attaches,
+ * which an application cannot reach. Same mechanism as `markPreScoped` and
+ * `markOrmAuthored`, for the same reason: a claim about who may do something is
+ * worth only as much as the thing that enforces it.
+ */
+describe("the composite-in key is the planner's alone", () => {
+  const attempt = (operand: unknown) => () =>
+    compileRead(ledgerEntry, "findMany", { where: { $compositeIn: operand } }, postgres);
+
+  test("a hand-written one is refused, however well formed", () => {
+    expect(
+      attempt({ fields: ["tenantId", "ledgerCode"], values: [[1, "a"]] }),
+    ).toThrow(UnsupportedQueryError);
+    expect(
+      attempt({ fields: ["tenantId", "ledgerCode"], values: [[1, "a"]] }),
+    ).toThrow(/not part of the query grammar/);
+  });
+
+  /**
+   * The three shapes that used to escape validation: two raw `TypeError`s from
+   * a destructure and a `.map`, and one that compiled with no `values` at all.
+   * Every one is now an `UnsupportedQueryError` naming the key.
+   */
+  test.each([
+    ["null", null],
+    ["fields not an array", { fields: "tenantId", values: [[1]] }],
+    ["no values at all", { fields: ["tenantId"] }],
+    ["a tuple of the wrong width", { fields: ["tenantId", "ledgerCode"], values: [[1]] }],
+  ])("a malformed one is refused with a path: %s", (_label, operand) => {
+    expect(attempt(operand)).toThrow(UnsupportedQueryError);
+    expect(attempt(operand)).not.toThrow(TypeError);
+  });
+
+  /** ...and the planner's own still compiles, on the dialect that binds it. */
+  test("the planner's own is accepted", () => {
+    const plan = compileRead(ledgerEntry, "findMany", { include: { ledger: true } }, postgres);
+    expect(plan.relations).toHaveLength(1);
+  });
+});
+
 describe("a nested write still refuses, and says why", () => {
   const run = () =>
     compileWrite(
--- a/packages/gemi/orm/compile/composite-relations.test.ts
+++ b/packages/gemi/orm/compile/composite-relations.test.ts
@@ -2,12 +2,13 @@ import { afterEach, beforeEach, describe, expect, test } from "vitest";
 
 import { PostgresDialect } from "../dialect/postgres";
 import { SqliteDialect } from "../dialect/sqlite";
-import { UnsupportedQueryError } from "../errors";
+import { UnknownFieldError, UnsupportedQueryError } from "../errors";
 import { ledger, ledgerEntry } from "../fixtures";
 import * as registry from "../registry";
 import { lateralStrategy } from "./lateral";
 import { compileRead } from "./read";
 import { compileWrite } from "./write";
+import { plannerCompositeIn } from "./where";
 
 /**
  * Multi-field relations — `@relation(fields: [a, b], references: [c, d])`.
@@ -298,13 +299,17 @@ describe("the composite-in key is the planner's alone", () => {
   const attempt = (operand: unknown) => () =>
     compileRead(ledgerEntry, "findMany", { where: { $compositeIn: operand } }, postgres);
 
-  test("a hand-written one is refused, however well formed", () => {
-    expect(
-      attempt({ fields: ["tenantId", "ledgerCode"], values: [[1, "a"]] }),
-    ).toThrow(UnsupportedQueryError);
-    expect(
-      attempt({ fields: ["tenantId", "ledgerCode"], values: [[1, "a"]] }),
-    ).toThrow(/not part of the query grammar/);
+  /**
+   * The **unknown-key treatment**, which is what the original comment promised.
+   * `UnsupportedQueryError` would say "does not support '$compositeIn' *yet*",
+   * a promise about a key that is deliberately not in the grammar — the word
+   * #82 and #88 have each already corrected once.
+   */
+  test("a hand-written one is not a field on the model", () => {
+    const operand = { fields: ["tenantId", "ledgerCode"], values: [[1, "a"]] };
+    expect(attempt(operand)).toThrow(UnknownFieldError);
+    expect(attempt(operand)).toThrow(/is not a field on model LedgerEntry/);
+    expect(attempt(operand)).not.toThrow(/yet/);
   });
 
   /**
@@ -317,9 +322,24 @@ describe("the composite-in key is the planner's alone", () => {
     ["fields not an array", { fields: "tenantId", values: [[1]] }],
     ["no values at all", { fields: ["tenantId"] }],
     ["a tuple of the wrong width", { fields: ["tenantId", "ledgerCode"], values: [[1]] }],
-  ])("a malformed one is refused with a path: %s", (_label, operand) => {
-    expect(attempt(operand)).toThrow(UnsupportedQueryError);
+  ])("a malformed one is refused rather than crashing: %s", (_label, operand) => {
+    // Every one of these used to be a raw `TypeError` from a destructure or a
+    // `.map`, or — for "no values at all" — a clean compile whose failure
+    // arrived at bind time.
+    expect(attempt(operand)).toThrow(UnknownFieldError);
     expect(attempt(operand)).not.toThrow(TypeError);
+  });
+
+  /**
+   * The planner's *own* malformed operand is a different audience: an internal
+   * invariant that does not hold is genuinely something the ORM does not
+   * support, and there "yet" is the honest word.
+   */
+  test("a malformed operand from the planner names the invariant", () => {
+    const planners = plannerCompositeIn(["tenantId"], [[1, "extra"]]);
+    expect(() =>
+      compileRead(ledgerEntry, "findMany", { where: { $compositeIn: planners } }, postgres),
+    ).toThrow(/the planner built/);
   });
 
   /** ...and the planner's own still compiles, on the dialect that binds it. */
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -10,7 +10,12 @@ import {
 import { assertPageArgument } from "./paginate";
 import { COUNT_KEY } from "../relation-filters";
 import * as registry from "../registry";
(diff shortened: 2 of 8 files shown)


nstfkc:
## Review — composite `include` through `unnest` (closes #97)

Read the diff against its true base (`4ed5a3e`, #95) — 7 files, +334. 933 tests pass.

**The collapse works, and keeps the half it must not take:**

```
postgres  $compositeIn: 1 key for 4 parent counts     (OR shape: 4)
sqlite    $compositeIn: 4 keys                        keeps the OR, as before
field ORDER stays distinct: true
```

That last line is the one worth having a test for: `["tenantId", "ledgerCode"]` and `["ledgerCode", "tenantId"]` pair the two sides differently, so collapsing on the tuple *list* while keeping the *fields* in the key is exactly the right split.

Putting `canBindCompositeIn` / `compositeIn` beside `inList` rather than branching in the compiler is the right home — it is the same structural disagreement one arity up. Asking the dialect about the **column types** rather than assuming is the part I would have expected to be skipped, and keeping the map to four exact-round-trip types is the conservative direction: a cast that does not round-trip is wrong rows, not an error. The `DateTime` absence with its reason attached is a better argument than an allowlist without one.

*A note on my own method:* my first check reported 4 keys on Postgres and I nearly wrote that the fix did not work. The probe was building `{ OR: [...] }` by hand — the shape the loader used to emit — so it was measuring the old path faithfully and the new one not at all. Re-running against `$compositeIn` gives the numbers above.

### The stated safety property for `$compositeIn` does not hold

> `$compositeIn`, because a Prisma field name cannot start with `$`. … It is internal — only `plan-relations.ts` builds one — and **a caller writing it by hand meets the same unknown-key treatment as anywhere else, since no schema has that field.**

It does not. The `COMPOSITE_IN` branch sits above the field lookup in `compileWhere`'s key loop, so a caller-supplied one is honoured rather than rejected:

```
where: { $compositeIn: { fields: ["id","email"], values: [[1,"a"],[2,"b"]] } }
  ->  OK, compiles
```

This is **not** an injection: `fields` are resolved against `schema.fields` and an unknown one raises `UnknownFieldError`, the columns go through `quoteIdent`, and the values are bound. What it is, is an undocumented input surface with no shape validation — and the file it lives in validates every other operand:

```
{ $compositeIn: null }                     TypeError: Cannot destructure property 'fields' from null
{ $compositeIn: { fields: "id", … } }      TypeError: fields.map is not a function
{ $compositeIn: { fields: ["id"] } }       OK — compiles with no `values` at all
```

Two raw `TypeError`s where every neighbouring operand raises `UnsupportedQueryError` with a path and a reason, and a third shape that **compiles clean** and defers its failure to bind time — which is the specific thing `assertCreateManyOperand` was written to avoid on #75, in this same file's words: *checked at plan time, so a misspelled key fails when the query is compiled rather than after the parent row has already been written.*

The fix is small either way, and which one depends on what is intended:

- if it is genuinely internal, the claim needs to become true — check the key *after* the field lookup, or refuse it when the operand did not come from the planner;
- if a caller reaching it is acceptable, it needs the operand validation every other `where` key has, and a line in the docs saying it exists.

What it should not stay is a documented-as-unreachable key that is reachable, since the next person will read the comment rather than test it. I did test it, which is the only reason this is in the review rather than in the code.

nstfkc:
Pushed as 6252931. You are right, and the claim was the problem rather than the code — I asserted a property instead of enforcing one, which is the thing this codebase has spent several PRs learning not to do.

Reproduced all four shapes exactly as you reported them, including the third:

```
{ $compositeIn: { fields: ["id","email"], values: [[1,"a"]] } }   COMPILED
{ $compositeIn: null }                                           TypeError: Cannot destructure 'fields'
{ $compositeIn: { fields: "id", values: [[1]] } }                TypeError: fields.map is not a function
{ $compositeIn: { fields: ["id"] } }                             COMPILED — no values at all
```

**I did both halves, because the choice you offered turned out to be a false one.**

The claim is now *enforced*: the operand has to carry a module-private `Symbol` the planner attaches. That is the third use of the pattern here after `markPreScoped` and `markOrmAuthored`, and it fits for the same reason — a claim about who may do something is worth exactly what enforces it. A `Symbol` is invisible to `Object.entries`, so it stays out of the plan key and cannot be written in JSON by someone who saw the key name in a stack trace. A caller now gets a refusal naming the key and saying what to write instead.

And the shape is validated at plan time *anyway*, for the planner's own operand. Below the provenance check those are the planner's bugs rather than a caller's — which is why they get a different sentence — but they belong where every other operand in the file is checked rather than at bind time. `assertCreateManyOperand`'s reasoning applies unchanged.

All four now:

```
caller-supplied, well formed   UnsupportedQueryError: … not part of the query grammar
null operand                   UnsupportedQueryError
fields not an array            UnsupportedQueryError
no values at all               UnsupportedQueryError
```

and the planner's path is unchanged — checked by reading the child query on both dialects rather than from green tests: `postgres → ["$compositeIn"]`, `sqlite → ["OR"]`.

Six tests covering the well-formed hand-written case, the three shapes that escaped, a tuple of the wrong width, and that the planner's own still compiles.

### On your note about the method

Worth saying that the near-miss cuts both ways: your first probe built `{ OR: [...] }` by hand and faithfully measured the path the fix removes. Mine was the mirror image — I asserted the key was unreachable from the shape of the code rather than trying it. Both are the same error, which is trusting a model of the system over the system, and only one of us caught it in time. Thank you for testing it.

939 unit tests, template suite green on SQLite (501) and Postgres 16 (752).

nstfkc:
### Re-reviewed `6252931` — the claim is true now rather than asserted

Every shape I used to get past the guard is refused, and the planner's own still works:

```
hand-written, well-formed   UnsupportedQueryError
hand-written, null          UnsupportedQueryError
hand-written, bad fields    UnsupportedQueryError
hand-written, no values     UnsupportedQueryError
planner-built               OK

plan keys for 2,3,10,50 parents : 1        the collapse survives
symbol in global registry       : false    unforgeable
```

A module-private `Symbol` on the operand is the right mechanism and the right precedent — it is the third thing in this codebase gated that way, after `markPreScoped` and `markOrmAuthored`, and it answers the question the `$` prefix could not: `$` makes the key un-collidable with a Prisma field, which is a different property from un-constructible by a caller. The PR originally conflated the two and now separates them.

The `no values` case is the one I would keep in the suite: it compiled clean before and deferred its failure to bind time, which is the shape `assertCreateManyOperand` exists to prevent — a validated operand and an unreachable key are two fixes, and this needed both.

The collapse surviving matters because a symbol key is invisible to `Object.entries`, so the marker cannot leak into the cache key. Worth knowing that is load-bearing rather than incidental.

939 tests pass.

One small thing, and I have now raised the same word three times so treat it as a tally rather than a request: `UnsupportedQueryError` renders as *"does not support '$compositeIn' **yet**"* for a key that is deliberately not in the grammar and never will be. #82 and #88 each corrected this once for their own case; whether this one wants `UnsupportedByDesignError` from #78 or simply the unknown-key error the original comment promised is a smaller question than the fix it is attached to.

nstfkc:
Pushed as the follow-up commit. Three raisings is enough — treating it as a tally is fair, and it turned out to have an answer that needs neither a new class nor #78's `UnsupportedByDesignError`.

**A caller's key now raises `UnknownFieldError`**, which is exactly what the original comment promised — *"the same unknown-key treatment as anywhere else"* — and which is simply accurate: `$compositeIn` is not a field on the model, and the message lists the ones that are. Every other non-field key already gets this.

**The planner's own malformed operand keeps `UnsupportedQueryError`**, and the split is the point. That is an internal invariant that does not hold — genuinely something the ORM does not support, where "yet" is the honest word for a state it should not be in. Two audiences, two errors, and the provenance check is already the line between them.

The tests assert the absence too, not just the new class:

```
expect(attempt(operand)).toThrow(UnknownFieldError)
expect(attempt(operand)).toThrow(/is not a field on model LedgerEntry/)
expect(attempt(operand)).not.toThrow(/yet/)
```

and there is a new case for a planner-built operand whose tuple width does not match its fields — the branch that was previously only reachable by hand, so it had no coverage of its own.

### On the two properties

Your separation is the thing I had conflated and would not have found alone: `$` makes the key **un-collidable with a Prisma field**; the `Symbol` makes it **un-constructible by a caller**. I had used the first to argue the second. They are different properties with different mechanisms, and the PR now says so in those terms.

Keeping the `no values` case for the reason you give is right — it compiled clean and deferred to bind time, which is the one of the four that a `TypeError` fix alone would have left in place.

940 unit tests, template suite green on SQLite (501).

nstfkc:
### Re-reviewed `e03bf5f` — and the split is a better answer than the one I asked for

I raised the word as a tally and expected at most a reworded string. Instead the operand now has **two regimes**, and they land in the right places:

```
caller, well-formed    UnknownFieldError: '$compositeIn' is not a field on model User
caller, null           UnknownFieldError: same
planner, well-formed   OK
planner, bad shape     UnsupportedQueryError
collapse still 1 key   true
```

A caller's key gets the unknown-key treatment — which is what the original comment promised before any of this, and it is *simply true*: `$compositeIn` is not a field on the model. That is better than the by-design category I had in mind, because it needs no new vocabulary at all; the accurate error already existed.

The line the diff draws — *"below here the operand **is** the planner's, so a bad shape is this ORM's bug"* — is the part worth keeping. It means the two failure modes are no longer the same code path answering two different questions: one is a caller naming something that does not exist, the other is an internal invariant. Separating them is what let the first one become accurate.

The residual "yet" is now only on the planner's own bad-shape path, which is unreachable without a gemi bug and is the right place for it to be least polished. I have raised that word three times and it has been acted on three times; I am not raising it a fourth.

940 tests pass, and the collapse this PR exists for is intact.

PR #101, opened 2026-07-28
Nested `update`, and the pass its refusal said was missing

nstfkc:
#65's item four, minus the two operands that still do not fit the line. **Stacked on #96.**

## The refusal was describing a gap in the wrong place

`REFUSED` said a nested `update` *"writes caller-supplied columns to a row that already exists, so the child's `onUpdate` and the scope-escape guard have to run over the payload — a pass this does not have yet."*

**The pass exists. It is one layer down.** Both live in `applyPolicies`, which the child's own `$exec` runs because a nested write step is not pre-scoped — the same property that stops `connect` reaching another tenant's row. Checked rather than argued, straight through the child's `updateMany`:

```
another tenant's row          ->  { count: 0 }
caller names a scoped column  ->  ScopeEscapeError
```

So what `update` needs beyond `disconnect` turns out to be **nothing**: the parent-key filter that keeps it off another parent's row is the same one, and the payload is the child's business — which is what routing it through the child's own operation means.

That is the second time this session that a `REFUSED` reason turned out to be stale rather than a real constraint, which is an argument for the entries naming *machinery* rather than *features* — this one named machinery, and that is exactly why it was checkable.

## What stays refused, and now for accurate reasons

- **`updateMany`** names a *predicate* rather than a row, so there is no lookup for the child's scope to narrow. That is the half of #94's line that has not moved.
- **`upsert`** needs a third thing neither half has: deciding which branch ran, from inside a nested step. Its old reason — "only the second half is implemented" — is no longer true now that both halves are.

## Both directions

The to-one form needed the foreign key in the statement's `RETURNING`, because the far row is identified by *this* row's column and the arguments do not carry its current value. It is the only owning-side operand that needs anything returned.

Prisma accepts both spellings there — `update: { … }` and `update: { data: { … } }` — measured, and accepting only the documented one would refuse a legal query.

## The harness caught the error type

A row that is not connected raises `RecordNotFoundError`, not `UnsupportedQueryError`. The differential harness compares the failure **kind** precisely so a refusal cannot pass as agreement with a miss, and this one was `other` against Prisma's `notFound` until it failed:

```
prisma threw but gemi threw — gemi ORM does not support 'data.accounts.update' yet…
expected { kind: 'other' } to deeply equal { kind: 'notFound' }
```

Nothing about that query is unsupported; the row is simply not there.

## Tests

Five differential cases — the named row, a list, the not-linked raise, and both to-one spellings — plus two policy tests asserting the claim this PR rests on: a nested update is scoped by the child, and its payload is judged by the child.

## Verification

942 unit tests. Template suite green on SQLite (515) and Postgres 16 (770), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -293,6 +293,8 @@ await List.create({
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
 | `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
 | `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
+| `disconnect` | Clears the link. `true` on a to-one; a unique key, or a list of them, on a to-many. The column has to be nullable. |
+| `delete` | Deletes the named rows outright, not just the link. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -314,8 +316,15 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
-Everything else in Prisma's nested grammar — `set`, `disconnect`, `update`, `updateMany`, `upsert`,
-`delete`, `deleteMany` — is refused, by name and with the reason. The line is **which rows an
+`disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
+Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
+this parent, and the difference is Prisma's: `disconnect` succeeds and changes nothing, `delete`
+raises. Both filter on the parent's key as well as yours, so neither can reach a row belonging to
+somebody else — and a row your policies hide is not reachable either, which `delete` reports as "not
+connected" rather than as denied, since that is the same answer a genuinely unlinked row gets.
+
+Everything else in Prisma's nested grammar — `set`, `update`, `updateMany`, `upsert`, `deleteMany` —
+is refused, by name and with the reason. The line is **which rows an
 operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
 one you identified by unique key) and writes either a whole new row or one foreign key the ORM
 chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -316,6 +316,12 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
+> **Known gap (#98).** A child whose policy *scopes on the foreign key itself* — `scope: () => ({
+> folderId: mine })` — cannot use any relation operand that writes that key, including `connect`,
+> unless it also declares an `onUpdate`. The scope-escape guard reads the payload and cannot tell a
+> column you supplied from one the nested step put there. Pre-existing rather than new to
+> `disconnect`, and tracked separately because the decision affects `connect` too.
+
 `disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
 Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
 this parent, and the difference is Prisma's: `disconnect` succeeds and changes nothing, `delete`
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -295,6 +295,7 @@ await List.create({
 | `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
 | `disconnect` | Clears the link. `true` on a to-one; a unique key, or a list of them, on a to-many. The column has to be nullable. |
 | `delete` | Deletes the named rows outright, not just the link. |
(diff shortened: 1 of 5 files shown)


nstfkc:
## Review — nested `update` (#65, item four)

Read the diff against its true base (`e77cdee`, #96) — 5 files, +360. 942 tests pass.

**The central claim holds.** "The pass exists, one layer down" is checkable and I checked it, straight through what the child's `updateMany` meets:

```
ordinary column, scope only    OK   where={"id":1,"AND":[{"orgId":7}]}  data={"label":"x"}
scoped column, no onUpdate     ScopeEscapeError
scoped column, with onUpdate   OK   data={"orgId":999,"touched":true}
onUpdate rewrites payload      OK   data={"label":"x","touched":true}
```

The child's scope is `AND`ed into the `where`, the escape guard judges the payload, and `onUpdate` runs over it. So `update` really does need nothing beyond `disconnect` — the parent-key filter is the same one and the payload is the child's business. Passing the caller's `data` through **unmarked** is what keeps that true, and it stays true when #99 lands: nothing here supplies `ormAuthored`, so the exemption never covers a caller's column.

The observation that this is the second stale `REFUSED` reason, and that it was checkable *because it named machinery rather than a feature*, is the kind of thing worth having written down — it makes the table a set of claims rather than a set of apologies.

### The two sides disagree about the kind of "there is nothing connected"

The foreign side raises `RecordNotFoundError`, and the comment explains why in detail — the harness compares failure **kind**, `other` against Prisma's `notFound` failed, and nothing about the query is unsupported.

The owning side, for the same condition, raises `UnsupportedQueryError`:

```ts
// planOwningSide, key === "update"
if (linked === null || linked === undefined) {
  throw new UnsupportedQueryError(
    `data.${relation.name}.update`, …,
    `this ${schema.name} has no '${relation.name}' to update — '${fkField}' is null. Prisma raises here too.`,
  )
}
```

`failureKind` maps `RecordNotFoundError` / `P2025` to `notFound` and everything else to `other`, so these two classify differently for what is semantically one thing: the row to update is not there. The message even says *"Prisma raises here too"* — which is the argument for matching its **kind**, not just the fact of raising.

Two consequences:

- It is the same defect the PR fixed on the other side, left in place on this one, and its own comment is the argument against it.
- The five differential cases cover *"the named row, a list, the not-linked raise, and both to-one spellings"* — the not-linked raise is the foreign side, and the to-one cases exercise the connected path. So nothing compares the owning-side miss against Prisma, which is exactly why it did not fail the way its twin did. Given the twin's history, `other` is unlikely to be the kind Prisma answers with; I have not run Prisma against it, so that is the thing to check rather than something I am asserting.

Also worth noting the wording: `UnsupportedQueryError` prefixes with *"does not support … yet"*, so a fully-supported operation reports itself as unimplemented because a row is missing. That is the vocabulary #82 and #88 have each already corrected once.

The fix looks like one word — `RecordNotFoundError` — plus a differential case for the owning-side miss, which is what would have caught it.

Nothing else outstanding. Accepting both to-one spellings because Prisma does, rather than only the documented one, is the right call and the sort of thing only measuring finds.

nstfkc:
Pushed as 6bf83e7. You are right, and it is the same defect I had just fixed on the other side — with its own comment as the argument against it.

**Measured, since you left it open rather than asserting it.** A user whose `organizationId` is null:

```
name: PrismaClientKnownRequestError
code: P2025
msg : depends on one or more records that were required but not found
```

P2025 → `notFound`, so the two sides classified one condition two ways.

Wrong twice over, and the second half would have outlived the first: `UnsupportedQueryError` prefixes *"does not support … yet"*, so a fully implemented operation reported itself as unimplemented because a row was missing. As you note, that vocabulary has already been corrected once each on #82 and #88 — this would have been the third.

### Why it survived is the more useful half

Your reading of the coverage is exact. The five cases were "the named row, a list, the not-linked raise, and both to-one spellings" — but the not-linked raise is the *foreign* side, and the to-one cases exercise the **connected** path. Nothing compared the owning-side miss against Prisma, so the twin's fix had nothing to carry it across.

That case now exists, and it fails against the old error with the same diff that caught the first one:

```
× updating a to-one that is not there raises, as Prisma does
  expected { kind: 'other' } to deeply equal { kind: 'notFound' }
```

Which is the point worth keeping: I fixed the foreign side *because a test failed*, and left the owning side wrong *because no test covered it*. The fix propagated exactly as far as the coverage did, not as far as the reasoning did — and the reasoning was written down in a comment three lines away.

942 unit tests, template suite green on SQLite (516) and Postgres 16 (771).

nstfkc:
### Re-reviewed `6bf83e7` — fixed, and you measured the half I flagged as unmeasured

Both misses now raise `RecordNotFoundError` — lines 434 and 671, the owning side and the foreign side, one condition with two spellings. 942 tests pass.

I said Prisma's kind was *"the thing to check rather than something I am asserting"*, and you checked it: P2025, *"depends on one or more records that were required but not found"*, which `failureKind` maps to `notFound`. That is the right response to a hedged claim — the hedge was there because I had not run Prisma, and the answer is now in the comment rather than in either of our heads.

**The differential case is the part that keeps it fixed.** Its own comment says why it did not exist: *"The connected to-one cases above never touch this branch."* That is the whole mechanism of the miss — the twin's fix did not carry over because nothing exercised the path, and the coverage gap and the defect had the same shape. Seeding a third user with no organisation so the branch is reachable at all is what makes the case real rather than nominal.

Recording that `UnsupportedQueryError` was *"wrong twice over"* — wrong kind for the harness, and wrong word for a fully-implemented operation — folds the vocabulary point in where it belongs rather than leaving it as a separate note. That is now three fixes on the same word (#82, #88, here), which is a reasonable argument that the default prefix is the thing to revisit rather than each call site.

PR #102, opened 2026-07-28
Every refusal says what to do instead, because `tsc` requires it

nstfkc:
Closes the second half of **#61**. The first half — separating "not yet" from "out of scope" — landed in #78 as `UnsupportedByDesignError`.

## The gap

`UnsupportedQueryError`'s `detail` was optional, and the call sites that omitted it were the highest-traffic ones: *an argument this operation does not accept*, which is the path a typo takes. They told the reader least likely to know why only **that** something was refused.

## Required, not conventional

For the reason five other parameters here are — `render`'s `origin`, `changedFields`' `schema`, `RelationExecutor.exec`'s `preScoped`, `resolveLink`'s `operation`, `assertPageArgument`'s reported name. Where omitting an argument produces a *worse result* rather than an error, `tsc` is the only thing that keeps a convention.

**And it found three sites a search did not.** I counted first with a script — 108 call sites, five without a detail — and the compiler answered **eight**. The three it added were `mode`, an unknown scalar operator, and a relation-node argument: all reachable from a typo, none matched by a crude arg-count. That gap is the argument for the type over the grep, and it is why this is a signature change rather than six edits.

## The details enumerate what *is* accepted

Which is the standard `UnknownFieldError` and `UnknownRelationError` already set — they list the valid names rather than only rejecting the invalid one:

```
'nope' yet (User.findMany). findMany takes include, omit, orderBy, select, skip, take, where.
'nope' yet (User.create). create takes data, include, omit, select.
'where.email.weird' yet (User.findMany). A scalar filter takes contains, endsWith, equals,
    gt, gte, in, lt, lte, mode, not, notIn, startsWith.
'mode: "loud"' yet (User.findMany). Prisma takes 'default' or 'insensitive'.
```

## The test pins the property, not the wording

A refusal has to be **more than its first sentence** — deliberately not "the message contains X". A call site added later with an empty string would satisfy the compiler and still fail here, which is exactly the hole a required parameter alone leaves. Verified by emptying one and watching it fail.

## One process note

`#61` was **unlabelled**, which is why it never appeared in any pass over the `orm` issues — and why four separate PRs rediscovered pieces of it. #82 removed the word from the dialect message, #88 recorded it as a category error and deferred, #100 and #101 each fixed their own call site. Each was right locally; none found the issue that owns the problem. It is labelled now.

The remaining piece of #61 is the *wording* of the prefix itself — "does not support 'x' **yet**" is still wrong for a refusal that is neither a gap nor a by-design feature decision (invalid input, an internal key, a missing row). That is a smaller question than the split #78 made and is best decided against the current messages rather than the ones the issue was written from, several of which have since shipped.

## Verification

959 unit tests. Template suite green on SQLite (509) and Postgres 16 (770), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

### DIFF
--- a/packages/gemi/orm/compile/aggregate.ts
+++ b/packages/gemi/orm/compile/aggregate.ts
@@ -515,6 +515,11 @@ function assertArgs(
       );
     }
 
-    throw new UnsupportedQueryError(key, schema.name, op);
+    throw new UnsupportedQueryError(
+      key,
+      schema.name,
+      op,
+      `${op} takes ${[...AGGREGATE_ARGS].sort().join(", ")}.`,
+    );
   }
 }
--- a/packages/gemi/orm/compile/index.ts
+++ b/packages/gemi/orm/compile/index.ts
@@ -35,7 +35,14 @@ export function compile(
     return compileRead(schema, op, args, dialect, strategy);
   }
   if (isWriteOperation(op)) return compileWrite(schema, op, args, dialect);
-  throw new UnsupportedQueryError(op, schema.name, op);
+  throw new UnsupportedQueryError(
+    op,
+    schema.name,
+    op,
+    `'${op}' is not an operation this ORM compiles — the reads, the writes, ` +
+      `'aggregate' and 'groupBy' are, and a raw statement goes through ` +
+      `'DB.query' or 'DB.execute'.`,
+  );
 }
 
 export { compileRead, isReadOperation };
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -561,6 +561,9 @@ function assertNodeArgs(
       `${node.as}.${key}`,
       schema.name,
       operation,
+      `A relation node takes ${[...(many ? RELATION_ARGS : TO_ONE_ARGS)]
+        .sort()
+        .join(", ")}.`,
     );
   }
 }
--- a/packages/gemi/orm/compile/read.test.ts
+++ b/packages/gemi/orm/compile/read.test.ts
@@ -9,7 +9,9 @@ import {
   UnsupportedQueryError,
 } from "../errors";
 import { USER_COLUMNS, account, mapped, user } from "../fixtures";
+import { compile } from "./index";
 import { compileRead } from "./read";
+import { compileWrite } from "./write";
 import { compileWhere } from "./where";
 import * as registry from "../registry";
 import type { RelationStrategy } from "./plan-relations";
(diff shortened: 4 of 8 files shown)


nstfkc:
## Review — every refusal says what to do instead (second half of #61)

Read the full diff against the merge base (`1ec25d3`). 959 tests pass.

Both halves of the mechanism check out, and they are genuinely two halves:

**`tsc` requires it.** Omitting the argument is a compile error, supplying it is not:

```
probe.ts(3,20): error TS2554: Expected 4 arguments, but got 3.
```

**The test catches what `tsc` cannot.** A required parameter still admits `""`, so I emptied one detail and ran the suite:

```
× an argument the read does not take
FAIL  orm/compile/read.test.ts > every refusal says what to do instead
```

Pinning *"the refusal is more than its first sentence"* rather than matching wording is the right shape — it survives every rewording and fails on the one thing that matters. That pairing is what makes this a mechanism rather than a sweep.

### The grep-versus-compiler argument is correct, and I proved it the hard way

The claim is that a script found five sites and the compiler found eight. I recounted independently and got **108 total, 8 without a detail** — the compiler's number, with the three extras exactly where the description says: `where.ts` twice (`mode`, an unknown scalar operator) and `plan-relations.ts` (a relation-node argument).

My *first* count said 44. It treated a backtick as an opening bracket that never closed, so every argument list containing a template literal stopped splitting and read as too short. That is the same class of error that made the original script say five, arrived at from the opposite direction — and it took me two attempts to count a thing the type system answers exactly. I would not have believed the argument as strongly if I had got it right first time.

### On the process note — it applies to how I have been reading this repo

> `#61` was **unlabelled**, which is why it never appeared in any pass over the `orm` issues — and why four separate PRs rediscovered pieces of it.

That is also why *I* kept rediscovering it. My pass over this work filters on the `orm` label, so #61 was invisible to me throughout; I raised the same word on #88, #100 and #101 as three local observations and explicitly declined to raise it a fourth time — without ever finding the issue that owned it. Three of those were fixed at the call site, which is the outcome you get when the reviewer and the author are both working from the same incomplete index.

Labelling it is the fix. Worth noting for anything else that predates the labelling convention: an unlabelled issue is invisible to every filtered pass, and the cost is not that it goes unfixed but that it gets fixed repeatedly and incompletely.

### The remainder

The description names what is left — the prefix wording for refusals that are neither a gap nor a feature decision — and leaves it in #61 rather than folding it in. That is the right call, and it is now tracked in the one place that owns the problem rather than in three PR threads.

PR #103, opened 2026-07-28
`InvalidArgumentError`: the argument exists, the value cannot mean anything

nstfkc:
Completes **#61**. Stacked on #102, which made the guidance sentence required — hence the base, so the diff here is only the new class and its conversions.

## The taxonomy

```
not implemented yet     UnsupportedQueryError      wait for a release
decided against         UnsupportedByDesignError   change the call        (#78)
implemented, bad value  InvalidArgumentError       fix the value
```

The third is what "yet" was wrong about **four separate times** — #82, #88, #100 and #101 each corrected it at their own call site before #61 was found to own the problem.

`take: "-2"` is the clearest case. It said *"gemi ORM does not support 'take' yet"* — wrong on both halves: it does, and there is nothing to wait for. It sent the caller to a changelog when the fix was one character in their own code. Now:

```
Invalid 'take' (User.findMany). Expected an integer, got "-2".
```

## This note asked for it, and is resolved by it

#88 left the error type deliberately wrong in `paginate.ts`:

> **The error type is wrong here, and deliberately left wrong until #78 lands.** … Validation is a fourth category again — not unimplemented, not a design refusal of a feature, but invalid input — so the fix is a category, not a reworded string, and inventing one here while #78 is in flight would be the second copy this function exists to avoid.

#78 landed the by-design half; this is the third class it predicted. The marker has been **replaced by what it asked for** rather than deleted — which is what a deferral note is for.

## How the 36 sites were chosen

By their own detail. A refusal already saying what a valid value looks like — *"Expected an integer"*, *"Expected 'asc' or 'desc'"*, *"Expected an object of fields"* — is by definition one where the argument exists and the value does not fit. That made the selection checkable rather than a judgement call per site.

## Extends `UnsupportedQueryError`

So every existing `catch` and every test asserting the base class is unaffected — **959 tests passed before a single assertion was touched**. That property is what lets a taxonomy be added to a shipped ORM without a breaking change: the specific classes are for the reader, the base class is for the handler. Same choice `UnsupportedByDesignError` made.

## Tests

Six cases pinning which class each kind of refusal gets, and a second set pinning the property that matters more than any wording:

```
a bad value for take    never says 'yet'
a bad value for skip    never says 'yet'
a bad value for orderBy never says 'yet'
a bad value for mode    never says 'yet'
```

That single sentence is what sent four PRs to four call sites. Asserting it directly is cheaper than asserting each message, and it catches the next one.

## Verification

970 unit tests. Template suite green on SQLite (509) and Postgres 16 (770), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint down to one warning (the pre-existing `path` in `where.ts`) from two, since the conversion left two imports unused — one of them reachable only from the comment this PR resolves.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -245,6 +245,26 @@ On SQLite the error names the dialect and says so. If a batch is too large for o
 split inside a transaction, and `skipDuplicates` survives the split: the counts sum, and a conflict
 in a later chunk does not roll back an earlier one, because `do nothing` is not an error.
 
+### What a refusal tells you
+
+Three classes, and which one you get says what to do next:
+
+| Error | Means | Do |
+| --- | --- | --- |
+| `UnsupportedQueryError` | not implemented **yet** | wait for a release, or use what the message names |
+| `UnsupportedByDesignError` | decided against | change the call — the message says to what |
+| `InvalidArgumentError` | the argument exists, the value cannot mean anything | fix the value |
+
+The third is the one worth knowing about. `take: "-2"` used to report that the ORM "does not support
+`take` yet" — wrong on both halves: it does, and there is nothing to wait for. It now says
+`Invalid 'take' (User.findMany). Expected an integer, got "-2".`
+
+All three subclass `UnsupportedQueryError`, so `catch (e) { if (e instanceof UnsupportedQueryError) }`
+still catches every one. The specific classes are for the person reading the message.
+
+Every refusal carries a second sentence saying what to do instead — that is a required argument
+rather than a convention, so it cannot be dropped by a call site added later.
+
 ### Per-call options
 
 A second parameter, not a key inside `args` — intersecting Prisma's own arg types with a
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -262,6 +262,11 @@ The third is the one worth knowing about. `take: "-2"` used to report that the O
 All three subclass `UnsupportedQueryError`, so `catch (e) { if (e instanceof UnsupportedQueryError) }`
 still catches every one. The specific classes are for the person reading the message.
 
+One edge is deliberate: an argument that is not in the grammar **at all** — a typo — keeps
+`UnsupportedQueryError` and its "yet", because the same check also refuses real Prisma arguments this
+ORM has not implemented, and nothing there can tell the two apart. The sentence immediately after it
+lists what the operation *does* take, which is what tells you a typo is a typo.
+
 Every refusal carries a second sentence saying what to do instead — that is a required argument
 rather than a convention, so it cannot be dropped by a call site added later.
 
--- a/packages/gemi/orm/compile/aggregate.ts
+++ b/packages/gemi/orm/compile/aggregate.ts
@@ -1,5 +1,9 @@
 import type { SqlDialect } from "../dialect";
-import { UnknownFieldError, UnsupportedQueryError } from "../errors";
+import {
+  InvalidArgumentError,
+  UnknownFieldError,
+  UnsupportedQueryError,
+} from "../errors";
 import type { QueryPlan } from "../plan";
 import type { FieldSchema, ModelSchema, ScalarType } from "../schema";
 import {
@@ -272,7 +276,7 @@ function aggregateTerms(schema: ModelSchema, op: string, args: any): Term[] {
       requested === null ||
       Array.isArray(requested)
     ) {
-      throw new UnsupportedQueryError(
+      throw new InvalidArgumentError(
         kind,
         schema.name,
         op,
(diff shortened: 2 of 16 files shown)


nstfkc:
## Review — `InvalidArgumentError` (completes #61)

Read the diff against its true base (`92ee1f1`, #102) — 16 files, +199. 970 tests pass.

The taxonomy holds at every point I probed, including the property that makes it safe to add to a shipped ORM:

```
take: "-2"      InvalidArgumentError      base-catchable  no "yet"
take: 1.5       InvalidArgumentError      base-catchable  no "yet"
skip: -1        InvalidArgumentError      base-catchable  no "yet"
orderBy: "up"   InvalidArgumentError      base-catchable  no "yet"
mode: "loud"    InvalidArgumentError      base-catchable  no "yet"
distinct        UnsupportedByDesignError  base-catchable  no "yet"
```

Both new classes are `instanceof UnsupportedQueryError`, so *"the specific classes are for the reader, the base class is for the handler"* is true rather than intended — and "959 tests passed before a single assertion was touched" is the right way to have demonstrated it.

Selecting the 36 sites **by their own detail** is the part I would not have thought of. A refusal already saying *"Expected an integer"* is by construction one where the argument exists and the value does not fit, so the criterion is checkable per site rather than a judgement call — and it only became available because #102 made that sentence mandatory first. The two PRs in that order are one mechanism.

Resolving #88's deferral marker by *replacing it with the thing it asked for* is the strongest signal here. That note said the fix was a category rather than a reworded string and declined to invent one mid-flight; the marker is gone from `paginate.ts` and `InvalidArgumentError` is imported in its place. A deferral that gets discharged rather than rotting is rare enough to name.

Asserting *"never says 'yet'"* directly, rather than matching four messages, is also the right level: that single word sent four PRs to four call sites, so testing the word is testing the actual failure.

### The one place the word survives, and whether that is intended

An argument that is not in the grammar at all still takes the base class:

```
{ nope: 1 }  ->  gemi ORM does not support 'nope' yet (User.findMany).
                 findMany takes include, omit, orderBy, select, skip, take, where.
```

By the taxonomy's own table that reads as *"wait for a release"*, and no release will ever add `nope`. But the same code path also refuses real Prisma arguments that gemi genuinely has not implemented, and it cannot tell a typo from one — so "yet" is right for half of what reaches it and wrong for the other half, with no information available to split them.

I am raising it as an observation about where the taxonomy's edge falls, not as a fifth request about that word: #102's mandatory detail already lists what the operation *does* take, so the sentence immediately after the wrong one tells the reader `nope` is not coming. That is probably sufficient, and splitting the case would mean knowing Prisma's full argument grammar per operation, which is a real cost for a small gain. Worth a line in the taxonomy saying so, so the next reader does not re-derive it.

Nothing outstanding. #61 is a good example of the thing #102's post-mortem identified — four PRs each fixed a symptom correctly, and the issue that owned the problem produced a fix that generalises.

nstfkc:
Pushed as 47c3193 — taking it as the observation it was, and writing the edge down rather than moving it.

Your analysis of why it cannot be split is the part I have recorded, because it is the part that would otherwise be re-derived: **the information needed to tell a typo from an unimplemented Prisma argument does not exist where the decision is made.** The same check refuses both, so "yet" is right for half of what reaches it. Splitting them means carrying Prisma's full argument grammar per operation — a large amount of duplicated schema for a small gain, and a second definition free to drift from the generated client, which is the thing every other typing decision here avoids.

And you named what makes it tolerable: #102's mandatory `detail` means the sentence immediately after the wrong one lists what the operation *does* take. A caller who typed `nope` learns it is not coming from the very next clause, without the class having to know which case it is in. The two halves of #61 turn out to cover for each other.

Pinned as a test as well as prose, so the edge is a decision rather than an oversight — it asserts the message keeps "yet" **and** names what is accepted, so a change to either half fails:

```
an argument outside the grammar keeps 'yet', and says what is taken
```

### On the ordering

> Selecting the 36 sites **by their own detail** … only became available because #102 made that sentence mandatory first. The two PRs in that order are one mechanism.

That is a better description of it than mine. I split them for reviewability and only noticed afterwards that the split was load-bearing: before #102 the criterion did not exist, because a site with no detail could not be classified by one. Had I done them as a single PR I would have hand-picked the sites and called it judgement.

971 unit tests, template suite green on SQLite (509).

nstfkc:
### Re-reviewed `47c3193` — the edge is a decision now

I offered that as an observation rather than a request, and asserting it is more than I asked for: `{ nope: 1 }` keeps "yet" **and** is pinned to say what the operation takes, so the mitigation is tested rather than assumed.

The note records the part that is easy to lose — that the same check refuses a typo and a genuinely unimplemented Prisma argument, and nothing at that point can tell them apart, so splitting it would mean carrying Prisma's argument grammar per operation. Saying *"'the third class did not cover this one' is the obvious first reading and it is not the interesting part"* is the sentence that stops the next reader filing it.

That completes #61 as far as I can see it: three classes, a required guidance sentence, and the one place the taxonomy deliberately stops written down with its reason.

PR #104, opened 2026-07-28
Nested `set`, which means "replace the set I can see"

nstfkc:
The last of #65's item three. **Stacked on #101.**

## The operand that appears to fail the line

`set` detaches whatever is currently linked before attaching what was named — so it acts on rows the *call* did not name, which is the half of #94's boundary that `disconnect` and `delete` pass and `updateMany` does not.

**#83 already answered it.** Its implicit many-to-many `set` had the identical problem, and the fix was to give the disconnect a *lookup*: read the linked rows through the child's own `findMany`, and clear only the ones the caller can see. The refusal's stated reason — *"it replaces the whole set, so it has to disconnect what is there now"* — was accurate about the difficulty and stale about the answer.

So `set` means **"replace the set I can see"**: a row the caller cannot see stays attached rather than being silently detached. That is the same choice `disconnect` makes one operand over, and the opposite of what an unscoped clear would do.

## Three behaviours the name does not suggest

Measured against Prisma before implementing — none of them guessable:

```
set [one of two]                the unnamed row's key becomes null
set [another parent's row]      it is repointed here, exactly as `connect` does
set [a row that does not exist] silently ignored — no error
```

The last decides an implementation detail: the attach half is an `updateMany` rather than the `update` a nested `connect` uses, because Prisma does not raise and matching it means not raising either.

## Tests

Four differential cases, one per behaviour above plus `set: []`.

Two policy tests, and the second is the one that keeps the first honest: with **no** policy on the child, `set` clears everything — Prisma's `set` exactly. Without it, "scope everything away" would pass the scoped test alone.

The scoped one is verified red by making the clear unscoped and watching the other tenant's row lose its link:

```
× set replaces only the links the caller can see
```

## Inherited limitation

`set` writes the child's foreign key on both halves, so it inherits **#98** exactly as `connect` does: a child whose policy scopes on that column is refused by the scope-escape guard. **#99** fixes that for every relation operand at once, and this needs no change when it lands.

## Housekeeping

`set` comes out of the refused list in `write.test.ts`, and the "a refusal explains what the operand would take" case is repointed at `updateMany` — whose reason now carries the line on its own: it names a *predicate* rather than a row.

## Verification

941 unit tests. Template suite green on SQLite (522) and Postgres 16 (777), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

## What is left on #65

`updateMany` and `upsert`. The first names a predicate rather than a row, which is the half of the boundary that has not moved; the second needs to know which branch ran from inside a nested step. Both reasons are stated in `REFUSED` and neither has been shown stale.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -296,6 +296,7 @@ await List.create({
 | `disconnect` | Clears the link. `true` on a to-one; a unique key, or a list of them, on a to-many. The column has to be nullable. |
 | `delete` | Deletes the named rows outright, not just the link. |
 | `update` | Writes your columns to the named row. The child's own `onUpdate` and scope rules apply to the payload. |
+| `set` | Replaces the whole set — detaches what is linked now, attaches what you name. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -336,8 +337,13 @@ is judged by the child's policies. Naming a column the child scopes on is refuse
 would be at the top level. On a to-one both spellings work, `update: { … }` and
 `update: { data: { … } }`, because Prisma accepts both.
 
-Everything else in Prisma's nested grammar — `set`, `updateMany`, `upsert`, `deleteMany` — is
-refused, by name and with the reason. The line is **which rows an
+`set` is the one supported operand that acts on rows you did **not** name, so it means *replace the
+set I can see*: it detaches the linked rows your policies let you read, and leaves the rest attached.
+With no policy on the child that is Prisma's `set` exactly. Two of its behaviours are worth knowing
+because the name does not suggest them — it will repoint a row belonging to another parent, exactly
+as `connect` does, and it silently ignores a named row that does not exist.
+
+Everything else in Prisma's nested grammar — `updateMany`, `upsert`, `deleteMany` — is refused, by name and with the reason. The line is **which rows an
 operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
 one you identified by unique key) and writes either a whole new row or one foreign key the ORM
 chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -75,6 +75,7 @@ const SUPPORTED = new Set([
   "disconnect",
   "delete",
   "update",
+  "set",
 ]);
 
 /**
@@ -85,7 +86,7 @@ const SUPPORTED = new Set([
  * row that does not exist yet. Refused here rather than accepted and made a
  * no-op, so a caller who wrote one on the wrong operation hears about it.
  */
-const EXISTING_ROW_ONLY = new Set(["disconnect", "delete", "update"]);
+const EXISTING_ROW_ONLY = new Set(["disconnect", "delete", "update", "set"]);
 
 /** Statements that insert a new row, so nothing is linked to it yet. */
 const CREATE_ONLY_STATEMENTS = new Set(["create", "createMany"]);
@@ -100,7 +101,6 @@ const CREATE_ONLY_STATEMENTS = new Set(["create", "createMany"]);
  * key, and none of it has to reason about what was there before.
  */
 const REFUSED: Record<string, string> = {
-  set: `It replaces the whole set, so it has to disconnect what is there now.`,
   deleteMany: `It deletes rows this call did not name.`,
   updateMany:
     `It writes caller-supplied columns to rows this call did not name — a ` +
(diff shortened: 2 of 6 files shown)


nstfkc:
## Review — nested `set` (the last of #65's item three)

Read the diff against its true base (`6bf83e7`, #101) — 6 files, +225.

The argument is right and it is the same one #83 discovered: `set` looks like it fails #94's boundary, and the answer is to give the detach a lookup so it *does* name its rows. "Replace the set I can see" is the correct reading, and a hidden row staying attached is the same choice `disconnect` makes one operand over.

The three measured behaviours are all ones I would have got wrong from the name — particularly that a nonexistent row is **silently ignored**, and that this is what forces `updateMany` rather than the `update` a nested `connect` uses. That is an implementation detail falling out of a measurement rather than a preference.

**Both halves are scoped, and I checked the one that matters.** The attach is `updateMany` on the caller's key, and `assertNamedRows` calls `matchUniqueKey` per entry at plan time — so `set: [{ role: "admin" }]` cannot repoint every matching row, which is the failure an `updateMany` on an unvalidated filter would otherwise allow. The clear is un-pre-scoped, so the child's own policy narrows it.

The second policy test — no policy on the child, `set` clears everything, Prisma's `set` exactly — is what keeps the first honest, and it is the same pin #96 used for both operands. Without it, "scope everything away" passes the scoped test.

### The `findMany` is not the mechanism its comment claims, and is not load-bearing

```ts
// Everything currently linked *that this caller can see*. Not
// pre-scoped, which is the whole mechanism: the child's own policies
// decide what "currently linked" means for this caller.
const linked = await executor.exec(relation.model, "findMany", …, false)

if (linked.length > 0) {
  await executor.exec(relation.model, "updateMany",
    { where: { [childField]: parentKey }, data: { [childField]: null } }, false)
}
```

The result is used only as `linked.length > 0`. The clear that follows re-derives the same set from the same `where` and is *itself* un-pre-scoped, so the scoping the comment credits to the `findMany` is actually done by the `updateMany` — which would apply the child's policy identically with the read removed.

Removing it entirely and letting the clear run unconditionally:

```
baseline                          21 passed
findMany + guard deleted          21 passed
```

Which is what you would expect: an `updateMany` matching nothing is a no-op, so the guard cannot change an outcome. What it changes is cost — it saves one statement when *nothing* is linked and adds one whenever something is, and a parent with existing links is the case `set` exists for. That is the wrong way round on a repo that counts statements deliberately, and the same shape as #83's `set`, where the fix was to collapse N statements into one.

Two ways to resolve it, and the choice is about what the comment should say:

- **Drop it.** The behaviour is identical, one round trip cheaper in the common case, and the scoping story becomes "the clear is un-pre-scoped", which is both simpler and true.
- **Keep it and use it** — clear by the ids it returned rather than re-deriving the set. That makes the comment accurate and makes the read pay for itself, at the cost of an `in` list bounded by the link count.

Either is fine; the current shape is the one where the comment describes the first and the code does neither.

Nothing else outstanding. The #98 note is accurate — `set` inherits it exactly as `connect` does, and #99 fixes both at once with no change needed here.

nstfkc:
### Merge-order note for the nested-write stack, now that #83 has landed

Posting on this one because it is the top of the stack and therefore the branch that inherits the reconciliation — but it applies to #94, #96 and #101 equally, and none of the four can see it from inside itself.

`feat/orm` (`4c87d21`) now carries #83's implicit many-to-many path, which routes **before** the ordinary operand dispatch — `planOne` checks `link.join` first and returns, so the m2m branch never consults `SUPPORTED` at all. It has its own set:

```
feat/orm   JOIN_TABLE_SUPPORTED = { connect, disconnect, set, create }
```

None of the four stacked branches reference that constant — they were all cut before #83 merged, so it does not exist in their trees:

```
feat/orm-connect-or-create   0 refs
feat/orm-disconnect-delete   0 refs
feat/orm-nested-update       0 refs
feat/orm-nested-set          0 refs
```

So once the stack lands, the two paths diverge silently:

| operand | ordinary relation | implicit m-n |
| --- | --- | --- |
| `connect` | ✓ | ✓ |
| `create` | ✓ | ✓ |
| `disconnect` | ✓ (#96) | ✓ |
| `set` | ✓ (#104) | ✓ |
| `connectOrCreate` | ✓ (#94) | ✗ |
| `createMany` | ✓ | ✗ |
| `delete` | ✓ (#96) | ✗ |
| `update` | ✓ (#101) | ✗ |

Nothing fails to compile and no test notices — a caller writing `update` through an implicit m-n just gets `planJoinTable`'s refusal listing four operands, while the identical spelling works on an ordinary relation one line away.

**Some of those gaps are right and should stay.** `delete` through a join table means deleting the far row rather than the link, which is different work and is exactly the distinction `disconnect` versus `delete` draws elsewhere. `createMany` needs the join rows as well as the child rows. Those are decisions; they are just not currently *stated* anywhere, because the branch that would state them does not know the other set exists.

Two things worth doing on whichever merges last:

- Reconcile the two sets deliberately, with the asymmetries named — the `REFUSED` table is the established shape for that, and it already carries a reason per operand.
- Note that `set` now has **two implementations**: #83's join-table one and #104's. They agree semantically, which is the good news — #83's scoped delete and this PR's "replace the set I can see" landed on the same rule from opposite directions — but they are separate code paths with one name, and the thing that keeps them in step today is that both were reasoned to the same place rather than sharing anything.

Recording it here rather than as a comment on all four.

nstfkc:
### The divergence is live on `feat/orm` now

#94 merged, so the split I flagged above is no longer hypothetical. On `origin/feat/orm` (`4280351`):

```
SUPPORTED              connect, connectOrCreate, create, createMany
JOIN_TABLE_SUPPORTED   connect, disconnect, set, create
```

Compiled against the template's own models, an ordinary relation and an implicit m-n now disagree in **both** directions:

```
m-n   tags.disconnect          ACCEPTED
ord.  accounts.disconnect      refused
m-n   tags.set                 ACCEPTED
ord.  accounts.set             refused
m-n   tags.connectOrCreate     refused
ord.  accounts.connectOrCreate ACCEPTED
```

The first two are the surprising ones: `disconnect` and `set` work *through a join table* today and are refused on an ordinary relation, which is backwards from what anyone would guess and is what a caller meets on `feat/orm` right now. #96 and #104 close that half when they land; `connectOrCreate` and `createMany` on the m-n side have nothing queued.

Nothing here is broken — every case is a clear refusal rather than a wrong result. It is a coherence question, and the reason to raise it now is that it is only visible from outside all four branches.

PR #105, opened 2026-07-28
Nested `updateMany` and `deleteMany`, and a criterion that was too strong

nstfkc:
The last of #65 except `upsert`. **Stacked on #104.**

## The criterion I restated on #94 was too strong

Both operands were refused on the reading that every supported one *"names its rows"*, and these name a **predicate**. The way to see the flaw is #101: `update` is implemented, and it routes through the child's own `updateMany` to get its policies. So the question the line is really asking is whether there is a `where` the child's scope can be `AND`ed into — and a predicate is exactly that.

Both operations are in `SCOPABLE`, so the child's scope narrows which rows match. `updateMany` additionally gets the child's `onUpdate` and the scope-escape guard over its payload, by the same mechanism #101 verified directly.

That makes five of the six `REFUSED` entries that turned out to be describing missing machinery which existed one layer down. `upsert` is now the only one left, and its reason has not moved: it needs to know which branch ran from inside a nested step, which neither half of it provides.

## The parent-key filter is load-bearing

Prisma applies both to *this parent's* children only — measured before implementing. Without the parent key in the filter, `deleteMany: {}` empties the table. All four differential cases fail with it removed:

```
× updateMany writes this parent's matching rows
× updateMany with an empty where takes every row of this parent
× deleteMany takes the filter directly
× deleteMany with an empty filter stops at this parent
```

## The two operands are shaped differently

The detail most likely to be got backwards, and not visible from the names:

```
updateMany: { where: { … }, data: { … } }
deleteMany: { … }                            the filter directly
```

Validated at plan time, with a message that says which shape the *other* one takes.

## Tests

Four differential cases against Prisma, one per behaviour above. The seeded `acc-theirs` belongs to another user and survives all four, which is what the parent-key filter is for.

## Note on the error class

The operand validation raises `UnsupportedQueryError` because that is what this branch has — both are argument *validation* and belong in #103's `InvalidArgumentError`. They need no edit when it lands: its conversion selects sites by whether the detail already says what a valid value looks like, and these say `Expected an object with 'where' and 'data'`. Adding two new "yet"-bearing validation refusals would otherwise undo work #103 has just finished.

## Verification

939 unit tests. Template suite green on SQLite (526) and Postgres 16 (781), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean on every file this touches.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -297,6 +297,8 @@ await List.create({
 | `delete` | Deletes the named rows outright, not just the link. |
 | `update` | Writes your columns to the named row. The child's own `onUpdate` and scope rules apply to the payload. |
 | `set` | Replaces the whole set — detaches what is linked now, attaches what you name. |
+| `updateMany` | Writes your columns to this parent's rows matching a filter. |
+| `deleteMany` | Deletes this parent's rows matching a filter — the filter goes directly under the key, not inside a `where`. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -343,7 +345,11 @@ With no policy on the child that is Prisma's `set` exactly. Two of its behaviour
 because the name does not suggest them — it will repoint a row belonging to another parent, exactly
 as `connect` does, and it silently ignores a named row that does not exist.
 
-Everything else in Prisma's nested grammar — `updateMany`, `upsert`, `deleteMany` — is refused, by name and with the reason. The line is **which rows an
+`updateMany` and `deleteMany` take a **filter** rather than a key, and it applies to this parent's
+rows only. Their operands are shaped differently and it is easy to get backwards: `updateMany` wraps
+its filter in `where` and carries a `data` beside it, while `deleteMany` *is* the filter.
+
+Only `upsert` in Prisma's nested grammar is refused, by name and with the reason. The line is **which rows an
 operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
 one you identified by unique key) and writes either a whole new row or one foreign key the ORM
 chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -76,6 +76,8 @@ const SUPPORTED = new Set([
   "delete",
   "update",
   "set",
+  "updateMany",
+  "deleteMany",
 ]);
 
 /**
@@ -86,7 +88,14 @@ const SUPPORTED = new Set([
  * row that does not exist yet. Refused here rather than accepted and made a
  * no-op, so a caller who wrote one on the wrong operation hears about it.
  */
-const EXISTING_ROW_ONLY = new Set(["disconnect", "delete", "update", "set"]);
+const EXISTING_ROW_ONLY = new Set([
+  "disconnect",
+  "delete",
+  "update",
+  "set",
+  "updateMany",
+  "deleteMany",
+]);
 
 /** Statements that insert a new row, so nothing is linked to it yet. */
 const CREATE_ONLY_STATEMENTS = new Set(["create", "createMany"]);
@@ -101,11 +110,6 @@ const CREATE_ONLY_STATEMENTS = new Set(["create", "createMany"]);
  * key, and none of it has to reason about what was there before.
  */
 const REFUSED: Record<string, string> = {
-  deleteMany: `It deletes rows this call did not name.`,
-  updateMany:
-    `It writes caller-supplied columns to rows this call did not name — a ` +
-    `predicate, not a key — so there is no lookup for the child's scope to ` +
-    `narrow. 'update' names its row and is implemented.`,
   upsert:
     `It is 'update' and 'connectOrCreate' at once and needs a third thing ` +
     `neither has: deciding which branch ran, from inside a nested step.`,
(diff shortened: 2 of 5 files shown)


nstfkc:
## Review — nested `updateMany` and `deleteMany` (the last of #65 bar `upsert`)

Read the diff against its true base (`b6497eb`, #104) — 5 files, +207.

**The correction to the criterion is right, and it is a correction to something I restated.** I wrote on #94 that every supported operand "names its rows"; #101 already falsified that by implementing `update`, which routes through the child's `updateMany`. The real question is whether there is a `where` the child's scope can be `AND`ed into — and a predicate is exactly that. Five of six `REFUSED` entries turning out to describe machinery that existed one layer down is a strong argument for the entries naming machinery rather than features, which is what made them checkable.

The differently-shaped operands (`updateMany: { where, data }` versus `deleteMany` taking the filter directly) is the detail I would have got backwards, and validating it with a message that names the *other* shape is the useful form of that refusal.

**The parent-key filter works.** Driving the planned step with a stand-in executor:

```
deleteMany {}             ->  where { userId: 42 }
updateMany { where … }    ->  where { userId: 42 }, data { … }
```

So `deleteMany: {}` takes this parent's children rather than the table, which is the failure the four differential cases are pinned against.

### The parent key *overrides* the caller's filter rather than conjoining it

```ts
const where = {
  ...(filter as object),
  [childField]: parent[parentField],
}
```

Spread first, parent key last — so a caller naming the foreign key has it silently replaced:

```
deleteMany { userId: 9 }  ->  where { userId: 42 }
```

The caller asked to delete this parent's children *that belong to user 9* — an empty set. What runs deletes **every one of this parent's children**. The direction is the destructive one, and nothing reports it.

The override is also the only reason the filter is safe: get the spread order backwards and `deleteMany: { userId: 9 }` reaches another parent's rows entirely. So one line is doing two jobs, and it is doing the second by silently discarding user input.

`{ AND: [filter, { [childField]: parentKey }] }` does both jobs without either hazard — it cannot be overridden in either direction, and a caller's `userId: 9` narrows to nothing the way they asked. That is also the shape `withScope` already uses to add a policy's predicate without displacing the caller's, for the same reason and with the reasoning written out there.

Worth measuring against Prisma before deciding, since the harness is the arbiter here and I have not run it: my expectation is that Prisma conjoins (its nested `deleteMany` takes a child `WhereInput`, and the parent restriction is additional), which would make the current behaviour a silent divergence in the deleting direction. The four differential cases do not cover a filter naming the foreign key, which is why nothing caught it — the same shape as #101's owning-side miss, where the uncovered case and the defect were the same case.

Contrast with #96, where the identical spread appears: there the caller supplies a *unique key* validated by `matchUniqueKey`, so a filter naming the foreign key is both unusual and constrained. Here the filter is an arbitrary predicate, which is what makes the collision reachable.

### Smaller

The note on the error class is right and worth keeping — raising `UnsupportedQueryError` for the operand validation, and observing that #103's conversion selects by whether the detail already says what a valid value looks like, so these need no edit when it lands. Adding two new "yet"-bearing validation refusals would have undone #103 the week it finished.

nstfkc:
Taking the merge-order note from #104 here, since this is now the top of the stack — pushed as 255d598.

**The divergence was real and none of the four branches could see it.** `planOne` checks `link.join` first and returns, so the join-table path never consults `SUPPORTED`; the two sets drift with nothing failing to compile and no test noticing. A caller writing `update` through an implicit m-n would have got a refusal listing four operands while the identical spelling worked on an ordinary relation one line away.

Reconciled by **what an operand is about**, which turned out to be the whole split rather than a list of exceptions:

```
connect  create  disconnect  set              the link      — both paths
connectOrCreate  createMany  delete  update
updateMany  deleteMany                        the far row   — ordinary only
```

A join table holds two foreign keys and no columns of its own, so there is no row here for the second group to act on. Your `delete` example is the one I made explicit, because it is the case where the same word means two things: through a join table it would delete the far row rather than the link, which is what `disconnect` already does — the same distinction those two draw on an ordinary relation.

`JOIN_TABLE_REFUSED` carries a reason per operand in `REFUSED`'s shape, and the message now says the operand **means something different here** rather than that it is unimplemented. Asserted as well, since nothing else can see it: six refused with a reason, four not refused.

**On the two `set` implementations** — written down where either would be edited. You are right that the good news and the risk are the same fact: they agree because both were reasoned to "replace the set I can see" independently, and nothing keeps them in step but that.

### Two conflicts, both resolved by keeping both sides

`nested-writes.ts`'s imports, and the seed — where #90's `_count` accounts and this stack's `acc-mine-*` / `acc-theirs` rows are each load-bearing for their own cases. Taking either side alone would have silently weakened the other's tests, which is the failure this seed already carries a comment about. It now says which rows exist for which reason.

1025 unit tests, template suite green on SQLite (586) and Postgres 16 (851).

nstfkc:
### Re-reviewed `255d598` — the reconciliation is done well; the other finding is still open

**The operand split is right, and the framing is better than my table.** I listed exceptions; *"the link versus the far row"* is the actual rule, and it predicts every entry rather than enumerating them. Verified against the template's own m-n:

```
m-n accepted: connect, create, disconnect, set
m-n refused : connectOrCreate, createMany, delete, update, updateMany, deleteMany
```

Four and six, matching the assertion. `JOIN_TABLE_REFUSED` saying the operand **means something different here** rather than that it is unimplemented is the right message — `delete` through a join table really would be `disconnect` under another name, and that is the case where the same word means two things.

Resolving both merge conflicts by keeping both sides of the seed, and recording which rows exist for which reason, is the right call — #90's `_count` rows and this stack's `acc-*` rows are each load-bearing, and taking either side alone would have weakened the other's cases silently.

### Still open: the parent key overrides the caller's filter

Unchanged on this head — same probe, driving the planned step:

```
deleteMany { userId: 9 }  ->  where { userId: 42 }
```

The caller asked for this parent's children *belonging to user 9*, which is nothing. What runs deletes every one of this parent's children. That is not addressed in `255d598` and is not mentioned in the reply, so I am re-flagging rather than assuming it was declined — the head moving without it could easily read as handled.

The suggestion stands: `{ AND: [filter, { [childField]: parentKey }] }` keeps the parent restriction unforgeable *and* lets the caller's predicate narrow, which is what `withScope` already does for a policy's fragment for the same reason. And the thing worth measuring first is whether Prisma conjoins here — if it does, the current shape is a silent divergence in the deleting direction, which is the one that matters.

1025 unit tests pass on this head.

nstfkc:
### Re-reviewed `92d1ebf` — the cherry-pick closes the window

`conjoin` is on this branch now (4 call sites), so the destructive shape never reaches `feat/orm` regardless of which order the stack lands in. My original probe against this head:

```
deleteMany {}             ->  where { userId: 42 }
deleteMany { userId: 9 }  ->  where { AND: [{ userId: 9 }, { userId: 42 }] }
updateMany { where … }    ->  where { AND: [ … ] }, data { … }
```

Which is the point of raising merge order at all — the fix existing one branch up is not the same as the fix existing, if the branches can land separately.

PR #106, opened 2026-07-29
Join-table `connectOrCreate`, and a rule I invented that Prisma contradicts

nstfkc:
Follows the merge-order note on #104, and corrects the reconciliation I pushed for it one commit earlier. **Stacked on #105.**

## The rule was wrong, and I did not check it

#105 reconciled the two operand sets and explained the split as *"the link versus the far row"* — operands about the pair work through a join table, operands about the far row do not. Tidy, and false. Measured against Prisma on the template's own models:

```
update           OK   the far row is updated
delete           OK   the far row is deleted, not just the pair
deleteMany       OK
updateMany       OK
connectOrCreate  OK
createMany       Unknown argument — Prisma refuses it here too
```

`delete` through a join table **does** mean deleting the far row, which is the opposite of what the rule predicted. Five of the six gaps are *unimplemented* rather than principled; only `createMany` is a genuine refusal, and it is Prisma's rather than this ORM's.

That is the fifth `REFUSED`-style reason in this series to turn out stale — and the first I wrote myself, one commit after documenting the pattern. The lesson the other four taught is that a refusal naming **machinery** is checkable and one naming a **principle** is not. "The link versus the far row" is a principle, and it survived exactly as long as it took to run it.

## What this adds

`connectOrCreate`, because #83 already had both of its halves — `connect` inserts a pair for an existing row, `create` writes the row and its pair — so it is those two selected by a lookup and needed no new machinery. `findUnique` rather than `findUniqueOrThrow`, so a scoped-away hit takes the create branch: the same answer the ordinary-relation operand gives, closing the same probe.

The other four now say what they are: the far row is reachable only by reading the pairs first, which that path does not do yet. Each names the missing hop rather than a principle, so the next person can check it the way I should have.

## Verification

1024 unit tests. Template suite green on SQLite (586) and Postgres 16 (851), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

## One thing I could not settle, flagged rather than claimed

The m-n differential cases from #83 call a `seedTags()` helper **inside the test body**, and `expectSameWrite` resets to the seeded state before each client runs — which is the pattern that made four of my own cases vacuous on #96. Removing `seedTags` leaves four of the eight still failing, so at least half genuinely depend on it; I could not determine within a reasonable probe whether the other four are comparing against an empty `Tag` table.

Worth someone checking with fresh eyes — the decisive test is whether the tags exist at comparison time, not whether the case passes. I am recording it as a question rather than a finding because I have not proven it either way.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -422,12 +422,11 @@ as `connect` does, and it silently ignores a named row that does not exist.
 rows only. Their operands are shaped differently and it is easy to get backwards: `updateMany` wraps
 its filter in `where` and carries a `data` beside it, while `deleteMany` *is* the filter.
 
-**An implicit many-to-many accepts a narrower set**, and the split is by what an operand is *about*:
-`connect`, `create`, `disconnect` and `set` act on the **link** and work on both. `connectOrCreate`,
-`createMany`, `delete`, `update`, `updateMany` and `deleteMany` act on the far **row**, and a join
-table has two foreign keys and no row of its own — so each is refused there with which of those it
-is. `delete` is the clearest: through a join table it would mean deleting the far row rather than
-the link, which is what `disconnect` already does.
+**An implicit many-to-many accepts a narrower set** — `connect`, `connectOrCreate`, `create`,
+`disconnect` and `set`. The four that are missing (`update`, `updateMany`, `delete`, `deleteMany`)
+reach the far row *through* the pairs, which that path does not do yet; Prisma implements all four,
+so these are gaps rather than decisions and the refusals say so. `createMany` is the exception:
+Prisma does not offer it through a join table either.
 
 Only `upsert` in Prisma's nested grammar is refused, by name and with the reason. The line is **which rows an
 operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -1152,29 +1152,44 @@ function planForeignSide(
  * What an implicit many-to-many accepts, and how it differs from
  * {@link SUPPORTED} — which is the ordinary-relation set.
  *
- * **The two are reconciled deliberately.** `planOne` checks `link.join` first
- * and returns, so this path never consults `SUPPORTED` at all: the sets can
+ * **The two must be reconciled deliberately.** `planOne` checks `link.join`
+ * first and returns, so this path never consults `SUPPORTED` at all: the sets
  * drift without anything failing to compile and without a test noticing, and
  * they did — #83 landed this one while four branches were adding to the other,
  * none of which could see it from inside itself.
  *
- * The asymmetries that remain are decisions rather than omissions, and are
- * named in {@link JOIN_TABLE_REFUSED} beside the reasons. In short: the
- * operands that are *about the link* work here, and the ones that are about the
- * far **row** do not, because a join table has two foreign keys and no row of
- * its own to act on.
+ * **The gaps are unimplemented, not by design, and that is measured rather than
+ * reasoned.** An earlier version of this comment claimed the split was "the
+ * link versus the far row" — operands about the pair work here, operands about
+ * the far row do not. It is a tidy rule and it is wrong. Prisma offers five of
+ * the six through an implicit m-n:
  *
- *     connect  create  disconnect  set        both paths
- *     connectOrCreate  createMany  delete  update  updateMany  deleteMany
- *                                                  ordinary relations only
+ *     update           OK   the far row is updated
+ *     delete           OK   the far row is deleted, not just the pair
+ *     deleteMany       OK
+ *     updateMany       OK
+ *     connectOrCreate  OK
+ *     createMany       Unknown argument — Prisma refuses it here too
+ *
+ * So `delete` through a join table really does mean deleting the far row, which
+ * is the opposite of what that rule predicted. Only `createMany` is a genuine
+ * refusal, and it is Prisma's rather than ours.
+ *
+ * `connectOrCreate` is implemented here because #83 already had both of its
+ * halves — `connect` inserts a pair for an existing row, `create` writes the
+ * row and its pair — so it is those two selected by a lookup. The remaining
+ * four reach the far row *through* the pairs, which is a second hop this path
+ * does not have yet; {@link JOIN_TABLE_REFUSED} says so in those terms rather
+ * than inventing a principle.
  *
  * `set` is the one operand with **two implementations** — this file's
  * join-table version from #83 and the ordinary-relation one — and they agree
- * because both were reasoned to "replace the set I can see" independently,
- * not because they share anything. Worth knowing when either is changed.
+ * because both were reasoned to "replace the set I can see" independently, not
+ * because they share anything. Worth knowing when either is changed.
  */
 const JOIN_TABLE_SUPPORTED = new Set([
   "connect",
+  "connectOrCreate",
   "disconnect",
   "set",
   "create",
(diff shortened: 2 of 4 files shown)


nstfkc:
Fixed here rather than on #105, since this branch is on top of it and carries the reconciliation — pushed as 463b88c.

**You were right to re-flag it.** I answered the reconciliation half of your note on #105 and did not address this at all; the head moving without it reads exactly as you say, like it was handled.

And measuring first was the right instruction, because it settles the direction rather than the style:

```
deleteMany { userId: <other user> }  ->  nothing deleted, count unchanged
```

Prisma conjoins. Mine deleted every one of the parent's children — the caller asked for "this parent's children belonging to user 2", which is nothing, and got "all of this parent's children" instead. Silent, and in the deleting direction.

`AND` for the reason you gave: the restriction becomes **unforgeable** — no key the caller picks can replace it — while their predicate still narrows. That `withScope` already does exactly this for a policy fragment, and for the same reason, is the argument I should have found before writing a spread.

Applied to all three `where` sites: `updateMany` / `deleteMany`, `update`, and `disconnect` / `delete`. The last two take a unique key, so the collision needs a model whose foreign key is part of one — rarer, not impossible, and the rule is cheaper to hold everywhere than to re-derive per operand.

**Not** applied to the two `data` sites, where the overwrite is deliberate and documented: the foreign key there is the ORM's to set, and Prisma leaves it out of the nested input type entirely.

Two differential cases, both verified red against the spread:

```
× deleteMany with a filter on the foreign key deletes nothing
× updateMany with a filter on the foreign key writes nothing
```

Neither existed, which is why the shape survived three operands — I wrote the same merge three times and tested the case it handles rather than the case it breaks.

1024 unit tests, template suite green on SQLite (588) and Postgres 16 (853).

nstfkc:
## Review — join-table `connectOrCreate`, and the rule that was wrong (#106)

Read the diff against its true base (`255d598`, #105).

**I praised that rule without checking it, which is the same mistake in the other chair.** On #105 I called "the link versus the far row" *"better than my table — it predicts every entry rather than enumerating them"*. It predicted, and the prediction was wrong for five of six. Measuring it is what settled it, and the lesson you draw — that a refusal naming **machinery** is checkable and one naming a **principle** is not — is the right generalisation. I should have applied it when I read the principle rather than agreeing with it because it was tidier than mine.

`delete` through a join table meaning the far row, not the pair, is the specific thing I got wrong: I offered it as the example of a *correct* asymmetry on #104, and it is the one Prisma contradicts most directly.

`connectOrCreate` as "#83's two halves selected by a lookup" is right, and `findUnique` rather than `findUniqueOrThrow` — so a scoped-away hit creates instead of raising — closes the same probe the ordinary-relation operand closes. That symmetry is worth having.

### The `seedTags` question — settled, and it is worse than "four of eight"

You asked for fresh eyes on whether those cases compare against an empty `Tag` table. They do, and it is decidable statically:

```
seedTags()           = differential.reset()  +  prisma.tag.createMany([red, blue, green])
expectSameWrite(…)   = this.reset() → prisma runs → this.reset() → gemi runs → compare
reset()              = clearAll()  +  options.seed(prisma)
clearAll()           deletes Tag          (differential.ts, prisma.tag.deleteMany)
seed()               creates no tags      (0 mentions of "tag" in the seed body)
```

So `expectSameWrite`'s **first** `reset()` deletes the three tags `seedTags` just made, and nothing recreates them. Both clients run against an empty `Tag` table in every case.

Classifying the cases by whether they need a tag to already exist:

```
NEEDS existing tags   connect attaches existing rows
NEEDS existing tags   connect and create together
NEEDS existing tags   disconnect removes one pair and leaves the row
NEEDS existing tags   set replaces the whole set
NEEDS existing tags   set to nothing clears them
NEEDS existing tags   connecting the same pair twice is a no-op
NEEDS existing tags   the relation writes from the other side too
self-sufficient       create writes the child and the pair
```

Seven, not four. Each of those seven has both clients fail to find `label: "red"`, the harness records *both threw, same kind*, and the case passes having compared two failures.

**Why your experiment read as "four still fail".** `seedTags` does two jobs and only one of them survives: the `differential.reset()` at the top, and the `createMany` that gets wiped. Deleting the helper removes the reset as well, which changes test isolation — so what you measured was the reset going missing, not the tags. That is why the count did not come out at seven or zero.

The fix is the one #96 already found in this same harness: a fixture the comparison must see belongs in `seed`, which `reset()` re-runs, not in the test body, which it discards. Moving the three tags into `seed` makes all seven real, and the case that has been passing on an agreed failure will start comparing what it was written to compare.

I would treat that as blocking for this PR only insofar as `connectOrCreate` is one of the operands whose coverage is affected — its own new case will be vacuous by the same mechanism unless the seed moves.

nstfkc:
### Re-reviewed `463b88c` — the `AND` fix is right, and you measured the half I could not

```
deleteMany {}             ->  where { userId: 42 }                          unchanged
deleteMany { userId: 9 }  ->  where { AND: [{ userId: 9 }, { userId: 42 }] }
updateMany { where … }    ->  where { AND: [ … ] }, data { … }
```

The empty filter still produces a bare parent restriction rather than a pointless `AND`, which is the detail I would have expected to be lost.

I flagged Prisma's behaviour as *"the thing to measure rather than something I am asserting"*, and the comment now carries the measurement — `deleteMany { userId: <other> } -> nothing deleted`. That is the second time in this stack a hedge of mine has come back as a measured line in the code, which is the right disposal for one.

**All three parent-restriction sites converted, not just the one I found.** `conjoin` is at 764 (`updateMany`/`deleteMany`), 844 (`disconnect`/`delete`) and 901 — and I checked what was left: the remaining spreads at 967, 1022 and 1121 are `data:` payloads on *creates*, where the ORM sets a foreign key on a new row and there is nothing to conjoin with. Correct to leave. 1024 tests pass.

### Merge-order: #105 still carries the bug

```
conjoin refs on feat/orm-nested-many : 0
```

The fix lives here, one branch above the PR that introduced the shape. The stack has been merging bottom-up, so if #105 goes in first the destructive behaviour is live on `feat/orm` until #106 follows — with `deleteMany: { userId: 9 }` deleting every one of that parent's children in the window between.

Either land them together, or cherry-pick `conjoin` down to #105. Not urgent if they merge in one sitting; worth knowing if they do not.

### Still open

The `seedTags` finding from my last comment is untouched on this head — the seed still creates no tags (`0` mentions), and `seedTags` is still called from the test bodies (`9` references). Seven of the eight m-n cases are still comparing two agreed failures against an empty `Tag` table, and `connectOrCreate`'s new case joins them. Flagging only so the two findings do not get conflated — this commit answered the other one.

nstfkc:
Both items done. The `seedTags` finding is fixed on #107 (the top of the stack), and `conjoin` is cherry-picked down to #105.

### The m-n cases were vacuous, and you were right that I could not show it

Every one of the eight did its setup in the test body — the `seedTags()` helper *and* a `post.create` — and `expectSameWrite` resets to the **seeded** state before each client runs. So the rows were gone by comparison time, both clients failed identically, and the cases agreed on a *failure* rather than on a result. My earlier probe was inconclusive because I went looking through the harness; the decisive evidence is just that the seed creates no tags and the helper is called from the bodies, which is what you gave.

Fixed the way #96 was: the tags and one already-linked post move into the seed, the helper goes, and the seed comment says why — beside the accounts that are there for the same reason.

**Now verified to discriminate**, which is what was missing before. Breaking the join-table `disconnect` so it clears every pair rather than the named one:

```
× disconnect removes one pair and leaves the row
```

Before this it passed against that break.

One case needed rewriting rather than repointing: "the relation writes from the other side too" created its own post. It now connects the seeded post to `green` — the seeded tag with no posts — so it still has something to change.

### The merge window

Cherry-picked, since merge order is not mine to decide and the window is a *delete*. #105 now carries `conjoin` and its two differential cases and nothing else from this branch; the merge back up resolved to this branch's copy as the superset.

That leaves the stack consistent bottom-up: `deleteMany: { userId: 9 }` deletes nothing on #105, #106 and #107 alike.

1024 unit tests, template suite green on SQLite (591) with only the pre-existing linkage probe failing.

PR #107, opened 2026-07-29
Nested `upsert`, and `REFUSED` is empty

nstfkc:
Completes **#65**. Stacked on #106.

## The sixth stale reason

`upsert`'s entry said it *"needs a third thing neither half has: deciding which branch ran, from inside a nested step."* It does not — the lookup decides the branch, exactly as `connectOrCreate` has decided one since #94.

That is the sixth `REFUSED` entry in this series to describe machinery that was already there. The pattern across all six: a reason naming **machinery** is checkable, and one naming a **principle** is not. Five named a missing pass that the child's own `$exec` already ran; this one named a missing decision that a `findFirst` already makes. The one I wrote myself on #105 named a principle and was wrong within a commit.

## Scoped to this parent, which is what the operand means

Measured, because it decides the behaviour rather than merely constraining it:

```
upsert a row belonging to another parent
  ->  Unique constraint failed on the fields: (`publicId`)
```

Prisma looks only among **this** parent's children, does not find it, takes the *create* branch, and collides. So the conjunction #106 added is what produces the right branch here — the same mechanism answering a second question, not a safety net bolted on.

## `REFUSED` is empty, and stays

Kept rather than deleted: the next operand added here should have to write its reason down, and the table is where six wrong ones were findable.

## Tests

Three differential cases — the hit, the miss, and the collision. The third is the one that separates "scoped to this parent" from "found globally"; without it an implementation that looked up the row globally and updated it would pass the first two.

## Not implemented

The **to-one** form. Its far row is identified by *this* row's foreign key, so an absent one means creating the far row and then writing back to a parent that has already been inserted — a shape none of the other operands need. Recorded as what it is rather than left to be rediscovered.

## Verification

1024 unit tests. Template suite green on SQLite (591) and Postgres 16 (856), with only the pre-existing `generated.test.ts` linkage probe failing. `tsc` clean; oxlint clean.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -372,6 +372,7 @@ await List.create({
 | `set` | Replaces the whole set — detaches what is linked now, attaches what you name. |
 | `updateMany` | Writes your columns to this parent's rows matching a filter. |
 | `deleteMany` | Deletes this parent's rows matching a filter — the filter goes directly under the key, not inside a `where`. |
+| `upsert` | Finds the row **among this parent's**, updates it, or creates it. |
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -428,12 +429,21 @@ reach the far row *through* the pairs, which that path does not do yet; Prisma i
 so these are gaps rather than decisions and the refusals say so. `createMany` is the exception:
 Prisma does not offer it through a join table either.
 
-Only `upsert` in Prisma's nested grammar is refused, by name and with the reason. The line is **which rows an
-operand can name, and whose columns it writes**: everything supported names its rows (a new one, or
-one you identified by unique key) and writes either a whole new row or one foreign key the ORM
-chose. `set`, `disconnect`, `delete`, `deleteMany` and `updateMany` act on rows the call did not
-name; `update` names its row but writes your columns to it, which needs the child's `onUpdate` and
-the scope-escape guard run over the payload.
+`upsert` looks for its row **among this parent's** rather than globally, which is Prisma's own
+semantics and is the detail that decides what it does: naming a row that exists but belongs
+elsewhere takes the *create* branch and collides on the unique key, rather than updating somebody
+else's row.
+
+Every operand in Prisma's nested grammar is now implemented for an ordinary relation. Ones that act
+on rows already linked to this one — `disconnect`, `delete`, `update`, `updateMany`, `deleteMany`,
+`set`, `upsert` — are available on an `update` and refused on a `create`, which has nothing linked
+to it yet; Prisma reports them as unknown arguments there too.
+
+Two rules hold across all of them. The **foreign key is the ORM's to set**, so a nested row naming
+it is describing a different parent than the call is, and it is overwritten. And the **parent
+restriction is conjoined, never merged** — a filter naming the foreign key column is narrowed by the
+restriction rather than replacing it, so `deleteMany: { userId: <someone else> }` deletes nothing
+rather than everything.
 
 `skipDuplicates` is not implemented on `createMany` at any level.
 
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -507,6 +507,58 @@ function planOwningSide(
     return;
   }
 
+  /**
+   * `upsert` through a to-one, refused **by name**.
+   *
+   * The far row is identified by *this* row's foreign key, so an absent one
+   * means creating the far row and then writing back to a parent that has
+   * already been inserted — a shape no other operand here needs, and the
+   * reason this side is not implemented.
+   *
+   * Named rather than left to fall through, which is what it did: with no
+   * branch of its own it reached the `connect` handling below and reported
+   * `'where' yet (Organization.update.organization.connect)` — a different
+   * operand, a different model, and a claim that `{ id: 1 }` is not a unique
+   * field when it is. The real failure was that a `{ where, create, update }`
+   * operand was being read as a connect key.
+   *
+   * That is the shape #85 was filed for and #101 fixed on its own path: a
+   * refusal that misnames its origin sends the reader to a query they did not
+   * write.
+   */
+  /**
+   * The list operands, refused by name on a to-one.
+   *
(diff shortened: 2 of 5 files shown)


nstfkc:
## Review — nested `upsert`, and `REFUSED` is empty (completes #65)

Read the diff against its true base (`463b88c`, #106).

**The to-many form is right**, and I traced both branches through the planned step rather than reading them:

```
hit   findFirst  { AND: [{ publicId: "a1" }, { userId: 42 }] }
      updateMany { AND: [{ publicId: "a1" }, { userId: 42 }] }
miss  findFirst  { AND: [ … ] }
      create     { organizationId: 1, userId: 42 }
```

Both conjoined to this parent, both un-pre-scoped so the child's policies decide. The observation that #106's `conjoin` is what *produces the correct branch* here — Prisma looks only among this parent's children, misses, creates, and collides — is the good kind: one mechanism answering a second question rather than a guard bolted on. The collision case is the right third test, since without it a global lookup passes the first two.

`REFUSED` is genuinely empty (I checked the declaration, having first mis-matched it with a regex that ran into the following interfaces). Keeping the table rather than deleting it, so the next operand has to write its reason down, is right — six wrong reasons were findable there precisely because it existed.

### The to-one form is not refused — it falls through to `connect`

The description says it is *"recorded as what it is rather than left to be rediscovered"*. It is not recorded anywhere: `upsert` is in `SUPPORTED`, and `planOwningSide` handles `createMany`, `create`, `disconnect`, `delete`, `update` and `connectOrCreate` — no `upsert` branch — so it reaches the trailing `connect` handling. What a caller gets:

```ts
User.update({ where: { id: 1 }, data: {
  organization: { upsert: { where: { id: 1 }, create: {…}, update: {…} } },
}})
```
```
gemi ORM does not support 'where' yet (Organization.update.organization.connect).
update.organization.connect needs a unique field. Organization declares: id, publicId.
Use findFirst to query on anything else.
```

Every part of that is about a different query:

- the caller wrote `upsert`; the message says `connect`;
- the model is reported as `Organization`, not the `User` whose `data` this is;
- *"needs a unique field … declares: id, publicId"* — but `{ id: 1 }` **is** one. The real failure is that the operand is `{ where, create, update }` and the connect path reads the whole object as the key, so the extra members break `matchUniqueKey`;
- it suggests `findFirst`, which has nothing to do with it.

This is the shape #85 was filed for and #101 fixed on its own path — a refusal that misnames its own origin — and this PR is the one that would otherwise close the series clean. A named refusal in `planOwningSide` costs one branch and can say the true thing, which the description already articulates: the far row is identified by *this* row's foreign key, so an absent one means creating it and writing back to a parent that has already been inserted.

Worth a test alongside it. The to-many cases would not have caught this, and neither would `REFUSED` being empty — an operand can be in `SUPPORTED` and still be unimplemented on one side, which is a gap the table's own shape does not cover.

nstfkc:
### Still reproducing on `7934c04`

The rebase brought #105's cherry-picked `conjoin` up, but the to-one `upsert` is untouched — `planOwningSide` still has no `upsert` branch, and the message is unchanged:

```
gemi ORM does not support 'where' yet (Organization.update.organization.connect).
update.organization.connect needs a unique field. Organization declares: id, publicId.
Use findFirst to query on anything else.
```

Flagging only so the head moving does not read as addressed — the delta here was the merge, not this. The finding stands as written: `upsert` is in `SUPPORTED` but implemented on one side only, so it falls through to `connect`'s validator and reports a different operand, a different model, and a problem the caller does not have.

nstfkc:
Fixed and pushed. You are right, and the phrase "recorded as what it is" was true of the PR body and false of the code — which is the worse of the two places for it to be true.

Reproduced exactly as you have it: `upsert` is in `SUPPORTED`, `planOwningSide` had no branch, so it reached the trailing `connect` handling. Your reading of the four wrong clauses is the useful part — particularly that `{ id: 1 }` **is** a unique field and the real failure is a `{ where, create, update }` operand being read whole as a connect key. Refused by name now, saying the thing the description claimed.

### Your last paragraph found three more

> an operand can be in `SUPPORTED` and still be unimplemented on one side, which is a gap the table's own shape does not cover

That is the finding rather than the note. The two dispatches are separate, so `REFUSED` cannot cover this — the operand is not refused, it is supported on the other side. A walk over `SUPPORTED` against both a to-one and a to-many, asserting the refusal names the operand the caller wrote, caught:

```
organization.set          reported connect
organization.updateMany   reported connect
organization.deleteMany   reported connect
```

All three are list operands on a single-key relation, which Prisma does not offer either. They say so now.

One deliberate omission in that walk: it does not assert the **model**. `matchUniqueKey` reports the child whose key is missing — `Account.update.accounts.disconnect` — which is a separate question from whether the operand is named, and behaves the same on both sides. Asserting it would have folded two findings into one test and made neither clear. Worth its own look if it is worth anything.

Verified by removing the `upsert` branch and watching the walk fail.

1046 unit tests, template suite green on SQLite (591) and Postgres 16 (856).

nstfkc:
### Re-reviewed `13e444d` — fixed, and the finding generalised further than I took it

All four now refuse by their own name, on the side the caller actually wrote:

```
upsert      ->  'data.organization.upsert'
set         ->  'data.organization.set'
updateMany  ->  'data.organization.updateMany'
deleteMany  ->  'data.organization.deleteMany'
```

I reported one operand. You found that `set`, `updateMany` and `deleteMany` had the identical shape — in `SUPPORTED`, no branch on the to-one side, falling through to `connect` and reporting `connect` back to a caller who wrote something else. That is the general form of what I described as a structural gap and did not go looking for; three more instances is the difference between fixing an instance and fixing the class.

The to-one reason is also the honest one rather than a restatement: the far row is identified by *this* row's foreign key, so an absent one means creating it and then writing back to a parent that has already been inserted — a shape no other operand needs. That is a missing hop, which is checkable, rather than a principle, which the last six taught is not.

1046 tests pass.

Worth recording somewhere durable, since it is now a demonstrated failure mode rather than a hypothesis: **`SUPPORTED` is per-operand and the implementations are per-side**, so membership does not imply either side handles it, and an unhandled one falls through to whichever branch happens to be last rather than to an error. `REFUSED` being empty says nothing about it. Four operands sat in that gap at once, and the only reason any of them surfaced is that a caller-facing message named the wrong operand.

PR #109, opened 2026-07-29
A nested unique-key refusal reported the child's model and a made-up operation

### LINKED ISSUE #108: A nested unique-key refusal reports the child model and a synthesized operation
Found while adding the both-sides operand walk on #107.

## The failure

```ts
User.update({ where: { id: 1 }, data: { accounts: { disconnect: { nope: 1 } } } })
```

```
gemi ORM does not support 'where' yet (Account.update.accounts.disconnect).
update.accounts.disconnect needs a unique field. Account declares: id, publicId.
Use findFirst to query on anything else.
```

The caller wrote `User.update`. The refusal reports:

| field | value | should be |
| --- | --- | --- |
| `model` | `Account` | `User` — whose `data` this operand is in |
| `operation` | `update.accounts.disconnect` | `update` |
| `argument` | `where` | `data.accounts.disconnect.where` |

`operation` is the one that matters most: it is a **synthesized string**, not one of the thirteen operations, and `UnsupportedQueryError` documents these fields as structured and inspectable. An application branching on `error.operation` gets something that will never match.

It also reads oddly as prose — *"update.accounts.disconnect needs a unique field"* — and the suggestion is wrong for a nested operand: `findFirst` has nothing to do with a nested `disconnect`.

## Why it is the same line as #79 and #85

Both established that a refusal must name where it came from. #79 made `resolveLink`'s `operation` **required** because a defaulted one reported `(Order.include)` for a query that was a `where`; #85 was filed because a nested `include` under a `create` raised about `Account.create`. #101 fixed the same shape on its own path.

`matchUniqueKey`'s `op: string` is that defect one layer down: it is passed a path where an operation belongs, so it cannot report either correctly. The twelve call sites each build their own string — `` `${operation}.${relation.name}.${key}` `` — which is exactly the ad-hoc composition those PRs replaced elsewhere.

## The fix

Take the origin as a triple rather than a string:

```ts
matchUniqueKey(child, where, {
  model: schema.name,          // the caller's model
  operation,                   // the caller's operation
  argument: `data.${relation.name}.${key}.where`,
})
```

`tsc` enforces it across all twelve sites, which is the same mechanism #102 used for the mandatory `detail` — and there it found three sites a search had missed.

The child's name still belongs in the *message*, because it is whose keys are being listed: *"Account declares: id, publicId"* is the useful half and should stay.

## Acceptance

- A nested unique-key refusal reports `model` = the caller's model, `operation` = the caller's operation, and the operand path in `argument`.
- The message still names the child and its unique keys.
- The "use X instead" suggestion is right for a nested operand rather than saying `findFirst`.
- A test asserting the structured fields, not the message text — the fields are the part an application can act on.

nstfkc:
Closes #108. Stacked on #107 (`feat/orm-nested-upsert`).

## The defect

`matchUniqueKey`'s third parameter was a `string`, and it was doing three jobs. Root callers passed the operation. Nested callers passed a *path* into the same slot, and the model came from the child whose keys were being looked up:

```ts
User.update({ where: { id: 1 }, data: { accounts: { disconnect: { nope: 1 } } } })
```

refused with

| field | was | should be |
| --- | --- | --- |
| `model` | `Account` | `User` |
| `operation` | `update.accounts.disconnect` | `update` |
| `argument` | `where` | `data.accounts.disconnect` |

`UnsupportedQueryError` documents those three as structured and inspectable. Two were wrong. `operation` was not one of the thirteen, so an application branching on it could never match; `model` named a query the caller never wrote.

This is the same shape as #79 (`resolveLink`'s `operation`) and #85, one layer down.

## The change

A `RefusalOrigin` triple — `model`, `operation`, `argument` — so a caller **cannot** put a path where an operation belongs; the type rejects it. All twelve call sites converted.

The child is still named in the message, because it is whose keys are being listed. Only the structured half moves:

```
gemi ORM does not support 'data.accounts.disconnect' yet (User.update).
Account needs a unique field here. It declares: id, provider_providerAccountId.
Name it by one of those, or reach it through the Account model directly.
```

The trailing sentence is now conditional on nesting: at the root the advice is `findFirst`/`deleteMany`, which is right there and wrong under a `data.` path — you cannot `findFirst` inside a nested `disconnect`.

## Tests

A table over the six operands that refuse at compile time, asserting the **fields**, not the message text — the message is allowed to change and these are not. Plus a root case, to pin that the unnested path is unchanged.

One deliberate omission from that table: `connect` on the foreign side. It validates its key inside the `after` step rather than at plan time, so it does not refuse during compile at all. Its origin triple is converted with the rest, but the refusal arrives after the parent row is written — which contradicts this file's own "checked at plan time so a misspelled key fails when the query is compiled" rule. That is a separate defect from this one; noted in the test file rather than fixed here.

## Verification

- 1052 unit tests, `tsc --noEmit` clean, `oxlint` clean (the two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -1132,7 +1132,11 @@ function chunkedCreateMany(
  */
 function findThenWrite(schema: ModelSchema, args: any, op: Operation): boolean {
   try {
-    const key = matchUniqueKey(schema, args?.where, op);
+    const key = matchUniqueKey(schema, args?.where, {
+      model: schema.name,
+      operation: op,
+      argument: "where",
+    });
     return upsertAbsentConflictKey(schema, args, key).length > 0;
   } catch {
     return false;
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -1276,6 +1276,15 @@ function planForeignSide(
 
   // `connect` on this side means "point that existing row at me", which is an
   // update of the child's foreign key — not something this row's insert can do.
+  //
+  // Checked here rather than in the step below, for the reason `assertNamedRows`
+  // gives: the step runs after the parent row is written, so a key the child
+  // does not declare unwinds an insert that should never have happened. The
+  // owning side has always checked its `connect` at plan time; this side had
+  // not, which made the same operand answer differently depending on which
+  // table the foreign key is on (#110).
+  assertNamedRows(schema, relation, child, operand, key, operation);
+
   out.after.push({
     relation: relation.name,
     operation: "connect",
@@ -1284,11 +1293,6 @@ function planForeignSide(
       if (!parent) return;
 
       for (const item of listOf(at(args))) {
-        matchUniqueKey(child, item, {
-          model: schema.name,
-          operation,
-          argument: `data.${relation.name}.${key}`,
-        });
         await executor.exec(
           relation.model,
           "update",
@@ -1456,6 +1460,16 @@ function planJoinTable(
     );
   }
 
+  // `connect`, `disconnect` and `set` name existing rows by a unique key, and
+  // the step below resolves them one at a time — after the parent row exists,
+  // since a pair needs both ends. That made a key the child does not declare
+  // refuse from mid-transaction, while the *same* operand on an ordinary
+  // relation refuses from the compiler (#110). `create` names no existing row
+  // and `connectOrCreate` is checked above, so those two are not this.
+  if (key === "connect" || key === "disconnect" || key === "set") {
+    assertNamedRows(schema, relation, child, operand, key, operation);
+  }
+
   const join = link.join!;
   const parentField = link.parentField;
   const childField = link.childField;
@@ -1591,12 +1605,10 @@ function planJoinTable(
          * confirm a row the caller cannot see.
          */
         if (key === "connectOrCreate") {
+          // The `where` was validated at plan time by
+          // `assertConnectOrCreateOperand`, which the branch at the top of this
+          // function runs for both relation kinds.
           const entry = item as Record<string, unknown>;
-          matchUniqueKey(child, entry.where, {
-            model: schema.name,
-            operation,
-            argument: `data.${relation.name}.connectOrCreate.where`,
-          });
 
           const hit = (await executor.exec(
             relation.model,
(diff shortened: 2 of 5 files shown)


nstfkc:
## Review — nested unique-key refusals report the wrong origin (closes #108)

Read the diff against its true base (`13e444d`, #107). 1052 tests pass.

The defect is real and the fix is the right shape. Verified the structured fields directly, since the whole point is that they are inspectable rather than that the message reads well:

```
disconnect  model=User  operation=update  argument=data.accounts.disconnect
delete      model=User  operation=update  argument=data.accounts.delete
update      model=User  operation=update  argument=data.accounts.update.where
set         model=User  operation=update  argument=data.accounts.set
upsert      model=User  operation=update  argument=data.accounts.upsert.where
root        model=User  operation=update  argument=where
```

`operation` is one of the thirteen again rather than a synthesized path, so an application branching on it can match; `model` is the one the caller queried; the path moved to `argument` where it belongs. The root case being unchanged is the part worth having pinned.

Making it a `RefusalOrigin` triple rather than fixing twelve call sites is the difference between this recurring and not — a caller *cannot* put a path where an operation belongs now. That is the sixth parameter in this codebase made type-enforced for that reason, and the lineage back to #79 and #85 is accurate.

Asserting the **fields** rather than the message text is right for the same reason the "more than its first sentence" test on #102 was: the message is allowed to change and these are not.

Keeping the child's name in the *message* while moving only the structured half is the correct split — it is the child's keys being listed, and the reader needs to know that. And making the trailing advice conditional on nesting catches something I would have missed: `findFirst` is sound advice at the root and nonsense under a `data.` path.

### The disclosed omission is real, and worse in one respect than described

Confirmed independently — a foreign-side `connect` with a bad key is not refused at compile time at all:

```
connect(f)  ->  ACCEPTED (no compile-time refusal)
```

and only refuses when the step runs, at which point the triple is right:

```
step refused -> model=User operation=update argument=data.accounts.connect
```

So the disclosure is accurate. The part worth adding: because it compiles, the plan is **cached**. A misspelled `connect` key is a valid plan-cache entry, so every subsequent call with that shape re-opens the transaction, writes the parent, runs the step and unwinds — rather than failing once at compile. The plan cache keys on argument shape, and `{ nope: 1 }` is a shape, so this is not a one-off cost.

Naming it in the test file rather than fixing it here is the right call for scope — it is a different defect with a different fix. Worth an issue rather than only a comment, on the evidence of this session: an unfiled note on a merged branch has not survived twice already.

PR #111, opened 2026-07-29
Three nested operands checked their unique key after writing the parent row

### LINKED ISSUE #110: Three nested operands validate their unique key at run time, after the parent row is written
`assertNamedRows` states the rule this file follows:

> Validated at plan time for the reason every operand here is: a refusal that arrives mid-transaction has to unwind a parent row that should never have been written.

Three call sites do not follow it. Structurally located by brace-tracking every `matchUniqueKey` call in `nested-writes.ts` and asking whether it sits inside an `async run()` body:

| line | path |
| --- | --- |
| 1287 | foreign-side `connect` |
| 1595 | join-table `connectOrCreate.where` |
| 1627 | join-table `connect` / `disconnect` / `set` |

1595 turns out to be **redundant** — `assertConnectOrCreateOperand` already covers it at plan time. The other two are real gaps.

## Measured

Compiling each operand with a key the child does not declare, and asking whether `compileWrite` refuses:

| relation kind | operand | plan time |
| --- | --- | --- |
| owning (FK on this row) | `connect` | refused |
| foreign (FK on child) | `connect` | **accepted** |
| foreign | `disconnect` / `set` / `delete` / `update` / `upsert` / `updateMany` / `deleteMany` / `connectOrCreate` | refused |
| m-n (join table) | `connect` / `disconnect` / `set` | **accepted** |
| m-n | `connectOrCreate` | refused |

So the *same operand* gets two different answers depending on which side of the relation the foreign key is on — `connect` is checked at plan time on the owning side and at run time on the foreign side, and `disconnect`/`set` are checked at plan time on an ordinary relation and at run time through a join table. Nothing distinguishes those cases; it is an omission, not a decision.

Prisma refuses all of them with a `PrismaClientValidationError` raised client-side, before any statement is sent.

## Consequence

Multi-statement writes open an implicit transaction, so the parent row **is** rolled back — this is not data loss, and I want to be precise about that rather than overstate it. What it costs:

- A statically-invalid query does a `BEGIN` + parent `INSERT` + `ROLLBACK` before refusing. An application that puts request input into a `connect` key turns a validation error into a write transaction per request.
- `compileWrite` succeeds, so the query is treated as supported and its plan is cached. Anything that validates by compiling sees nothing wrong.
- The refusal arrives from inside a nested step rather than from the compiler, so it is raised at a different point in the call than every neighbouring operand's.

## Plan-time validation is sound here

Worth stating, because it is the assumption the existing asserts already rest on: the plan key carries the operand's key *names* and collapses only its values.

```
A: sqlite:batched:User:update:{data:{accounts:{connect:{id:number}}},where:{id:number}}
B: sqlite:batched:User:update:{data:{accounts:{connect:{nope:number}}},where:{id:number}}
```

`{id: 1}` and `{id: 999}` share a plan; `{id: 1}` and `{nope: 1}` do not. So a check that runs once per shape cannot accept one call's key on another call's behalf.

## Acceptance

- The two real gaps validated at plan time, through the existing `assertNamedRows` rather than a new helper.
- The redundant run-time check at 1595 removed, and the ones the plan-time asserts now shadow removed with it — a second check that can no longer fire reads as though it can.
- A test that walks *both* sides of a relation and the join table for every operand that names a row by unique key, asserting each refuses at compile time. The per-operand tables above were what turned "one omission" into three; a structural test is what keeps the next operand from re-opening it.

Found while writing #109, which converted all twelve of these call sites and noted the foreign-side `connect` one in passing without measuring how far it went.

nstfkc:
Closes #110. Stacked on #109.

`assertNamedRows` states the rule this file follows:

> Validated at plan time for the reason every operand here is: a refusal that arrives mid-transaction has to unwind a parent row that should never have been written.

I located every `matchUniqueKey` call in `nested-writes.ts` by brace-tracking whether it sits inside an `async run()` body, rather than reading for it. Three did not follow the rule. One — join-table `connectOrCreate.where` — turned out redundant, since `assertConnectOrCreateOperand` already covers it at plan time. The other two were real.

## Measured

Compiling each operand with a key the child does not declare, and asking whether `compileWrite` refuses:

| relation kind | operand | before | after |
| --- | --- | --- | --- |
| owning (FK here) | `connect` | refused | refused |
| foreign (FK on child) | `connect` | **accepted** | refused |
| foreign | `disconnect` / `set` / `delete` / `update` / `upsert` / `connectOrCreate` | refused | refused |
| m-n (join table) | `connect` / `disconnect` / `set` | **accepted** | refused |
| m-n | `connectOrCreate` | refused | refused |

The same operand answered differently depending on which table holds the foreign key. Nothing distinguishes those cases — it was an omission, not a decision. Prisma refuses all of them client-side with a `PrismaClientValidationError`, before any statement is sent.

## What it actually cost

Multi-statement writes open an implicit transaction, so the parent row **was** rolled back. This was not data loss, and I would rather say so than overstate it. What it cost:

- A statically-invalid query did a `BEGIN` + parent `INSERT` + `ROLLBACK` before refusing. An application that puts request input into a `connect` key turned a validation error into a write transaction per request.
- `compileWrite` succeeded, so the query counted as supported and its plan was cached.
- The refusal arrived from inside a nested step rather than from the compiler — a different point in the call than every neighbouring operand's.

## The fix

`assertNamedRows` called at plan time on both paths. No new helper: the existing one already does exactly this, which is part of why the gap reads as an omission.

The two run-time checks these now shadow are removed. A check that can no longer fire reads as though it can, and this repo has had several stale guards found that way.

**Plan-time checking is sound here**, which is the assumption the existing asserts already rested on and worth pinning: the plan key carries the operand's key *names* and collapses only its values.

```
sqlite:batched:User:update:{data:{accounts:{connect:{id:number}}},where:{id:number}}
sqlite:batched:User:update:{data:{accounts:{connect:{nope:number}}},where:{id:number}}
```

`{ id: 1 }` and `{ id: 999 }` share a plan; `{ id: 1 }` and `{ nope: 1 }` do not. A check that runs once per shape cannot accept one call's key on another call's behalf.

## Tests

A table over all three paths — to-one, to-many, join table — for every operand that names an existing row by unique key. `create`/`createMany` name no existing row and `updateMany`/`deleteMany` take a filter rather than a unique key; those four are absent deliberately, and the comment says so.

Walked as a table because the per-path tables are what turned "one omission" into three.

**Verified against the unfixed source**: exactly four of the thirteen cases fail, and they are the four measured gaps. The nine that pass either way are covering the paths that were already correct, not padding.

Also corrected a comment in the neighbouring both-sides walk that #109 made false — it explained a deliberate omission by saying `matchUniqueKey` reports the child's model, which is no longer true.

## Verification

- 1065 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -1276,6 +1276,15 @@ function planForeignSide(
 
   // `connect` on this side means "point that existing row at me", which is an
   // update of the child's foreign key — not something this row's insert can do.
+  //
+  // Checked here rather than in the step below, for the reason `assertNamedRows`
+  // gives: the step runs after the parent row is written, so a key the child
+  // does not declare unwinds an insert that should never have happened. The
+  // owning side has always checked its `connect` at plan time; this side had
+  // not, which made the same operand answer differently depending on which
+  // table the foreign key is on (#110).
+  assertNamedRows(schema, relation, child, operand, key, operation);
+
   out.after.push({
     relation: relation.name,
     operation: "connect",
@@ -1284,11 +1293,6 @@ function planForeignSide(
       if (!parent) return;
 
       for (const item of listOf(at(args))) {
-        matchUniqueKey(child, item, {
-          model: schema.name,
-          operation,
-          argument: `data.${relation.name}.${key}`,
-        });
         await executor.exec(
           relation.model,
           "update",
@@ -1456,6 +1460,16 @@ function planJoinTable(
     );
   }
 
+  // `connect`, `disconnect` and `set` name existing rows by a unique key, and
+  // the step below resolves them one at a time — after the parent row exists,
+  // since a pair needs both ends. That made a key the child does not declare
+  // refuse from mid-transaction, while the *same* operand on an ordinary
+  // relation refuses from the compiler (#110). `create` names no existing row
+  // and `connectOrCreate` is checked above, so those two are not this.
+  if (key === "connect" || key === "disconnect" || key === "set") {
+    assertNamedRows(schema, relation, child, operand, key, operation);
+  }
+
   const join = link.join!;
   const parentField = link.parentField;
   const childField = link.childField;
@@ -1591,12 +1605,10 @@ function planJoinTable(
          * confirm a row the caller cannot see.
          */
         if (key === "connectOrCreate") {
+          // The `where` was validated at plan time by
+          // `assertConnectOrCreateOperand`, which the branch at the top of this
+          // function runs for both relation kinds.
           const entry = item as Record<string, unknown>;
-          matchUniqueKey(child, entry.where, {
-            model: schema.name,
-            operation,
-            argument: `data.${relation.name}.connectOrCreate.where`,
-          });
 
           const hit = (await executor.exec(
             relation.model,
@@ -1621,14 +1633,9 @@ function planJoinTable(
         }
 
         // `connect`, `disconnect` and `set` all name existing rows by a unique
-        // key. Resolved through the child's own `$exec`, so its policies decide
-        // which rows exist — otherwise a connect by any unique key reaches
-        // every tenant's.
-        matchUniqueKey(child, item, {
-          model: schema.name,
-          operation,
-          argument: `data.${relation.name}.${key}`,
-        });
+        // key — checked at plan time above. Resolved through the child's own
+        // `$exec`, so its policies decide which rows exist — otherwise a
+        // connect by any unique key reaches every tenant's.
         const found = (await executor.exec(
           relation.model,
           "findUniqueOrThrow",
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -915,6 +915,87 @@ describe("nested writes", () => {
   });
 
   /**
+   * **A key the child does not declare is refused by the compiler, on every
+   * path that names a row by one.**
+   *
+   * `assertNamedRows` states the rule: "Validated at plan time for the reason
+   * every operand here is: a refusal that arrives mid-transaction has to unwind
+   * a parent row that should never have been written."
+   *
+   * Three sites did not follow it — foreign-side `connect`, and join-table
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — three nested operands validated their key at run time (closes #110)

Read the diff against its true base (`829a470`, #109). 1065 tests pass.

Every path in the table now refuses at compile time, verified rather than read:

```
foreign connect     refused at compile      (was accepted)
m-n connect         refused at compile      (was accepted)
m-n disconnect      refused at compile      (was accepted)
m-n set             refused at compile      (was accepted)
foreign disconnect  refused at compile
owning connect      refused at compile
```

And I reproduced your brace-tracking check independently: **5 `matchUniqueKey` calls, 0 inside an `async run()` body**. So the run-time validators are genuinely gone rather than shadowed, which is the part worth confirming — a guard that can no longer fire reads as though it can, and this repo has found several that way.

**The soundness assumption is the load-bearing one, and it holds.** Plan-time checking is only valid if the plan key separates key *names* while collapsing values, or a cached plan compiled for a valid key would serve an invalid one:

```
key names distinguished : true     {id: 1} vs {publicId: "x"}  -> different keys
values collapsed        : true     {id: 1} vs {id: 999}        -> same key
```

Stating that as the reason rather than leaving it implicit is what makes the change checkable — it is the assumption every other plan-time assert in the file already rested on without saying so.

**Finding the sites by brace-tracking rather than by reading** is the right method and, on this session's evidence, better than the alternative: I have had three sloppy regexes nearly produce false claims in these reviews, twice on this exact file. A mechanical criterion that answers "is this inside a `run()`" beats judgement applied twelve times.

The impact section is the part I would most want kept. *"This was not data loss, and I would rather say so than overstate it"* — followed by what it actually cost: a `BEGIN` + `INSERT` + `ROLLBACK` per request for a statically-invalid query, a cached plan for a query that cannot succeed, and a refusal arriving from a different point in the call than every neighbouring operand's. That is more useful than a severity label, and it is what makes the redundant third site worth mentioning rather than quietly dropping.

PR #113, opened 2026-07-29
Three refusals reported a model that was not the caller's

### LINKED ISSUE #112: Three refusals report a model that is not the caller's, and one of them has no test at all
#108 fixed this for `matchUniqueKey`. It is the only place it was fixed. Sweeping all **143** `UnsupportedQueryError` / `InvalidArgumentError` construction sites in `orm/` and checking what lands in the `model` slot turns up three more.

## 1. `assertDisconnectable` reports the child on the foreign side

Three call sites, and they do not agree about what `owner` means:

| line | passes | correct? |
| --- | --- | --- |
| 420 | `schema` — the caller's | yes |
| 925 | `child` | **no** |
| 1027 | `child` | **no** |

Measured, with a required `Account.userId`:

```
User.update({ data: { accounts: { disconnect: { id: 1 } } } })
  → model=Account operation=update argument=data.accounts.disconnect
```

The caller wrote `User.update`. `operation` and `argument` are right; only `model` is wrong, and only on the foreign side — the owning side reports `User` correctly.

The parameter is named `owner`, which is the honest name for what the function needs: the schema whose column is being checked, since the message says *which column* is required. That is a genuinely different thing from the caller's model, and the function currently has no way to say both. Hence one parameter doing two jobs — the same shape as #108.

**It has no test.** Nothing in the repo matches the refusal's message, on any of the three sites. That is how the disagreement survived; it also means the fixtures cannot currently reach it, because none of them has a required foreign key on a to-many child.

## 2. `soft-deletes.ts` passes the literal string `"soft delete"` as the model

```ts
throw new UnsupportedQueryError("data", "soft delete", operation, …)
```

That is not a model name and never can be. The real one is available: `softDelete(model)` receives the model class, and every generated model carries `static $schema = schema.Account`, so `model.$schema.name` is right there. The `{ update(args): Promise<T> }` structural parameter type is what hides it.

## 3. `resolveStrategy` passes `"Model"` and `"$exec"`

```ts
throw new UnsupportedQueryError(`strategy: …`, "Model", "$exec", …)
```

`"Model"` is the base class, not the caller's model, and `$exec` is not one of the thirteen operations. One caller — `Model.ts:593` — and both the schema and the operation are in scope there.

## Deliberately not included

- `sql.ts` passes `"DB"` / `"sql"`. There genuinely is no model: `DB.sql` is the raw-SQL surface. A placeholder is the right answer when nothing else exists, and this is that case.
- `policy.ts:719` and `:745` pass `context.model` and `context.operation`, which are correct. They put the *operation* in the `argument` slot, which reads as "the whole query rather than one argument". Fine as is.

## Acceptance

- All three report the caller's model and a real operation.
- `assertDisconnectable` says both things instead of conflating them: the caller's model in the structured field, the column's owner in the message. Reuse the `RefusalOrigin` triple from #108 rather than inventing a second convention, and make it required so `tsc` enforces it.
- A test for `assertDisconnectable` on **all three** call sites — it has none, and the two wrong ones are exactly the two nothing exercises. This needs a fixture with a required foreign key on a to-many child, which the repo does not currently have.
- A test for the soft-delete refusal asserting the model is the real one.

nstfkc:
Closes #112. Stacked on #111.

#108 fixed this for `matchUniqueKey`, and only there. I swept all **143** `UnsupportedQueryError` / `InvalidArgumentError` construction sites in `orm/` and checked what lands in the `model` slot. Three more were wrong.

## 1. `assertDisconnectable` reported the child on the foreign side

Three call sites, disagreeing about what the first parameter meant:

| line | passed | correct? |
| --- | --- | --- |
| 420 | `schema` — the caller's | yes |
| 925 | `child` | **no** |
| 1027 | `child` | **no** |

Measured, with a required `Account.userId`:

```
User.update({ data: { accounts: { disconnect: { id: 1 } } } })
  → model=Account operation=update argument=data.accounts.disconnect
```

`operation` and `argument` were right; only `model` was wrong, and only on the foreign side — so one function answered two different ways depending on which branch you reached.

The parameter is named `owner`, and that is the honest name for what the *message* needs: it says which column is required, and the column lives on whichever side holds the foreign key. That is a genuinely different question from whose query this is, and one parameter cannot answer both — the same shape as #108. Now two do: `owner` for the message, a required `RefusalOrigin` for the fields.

The message now qualifies the column — `'Account.userId' is required` rather than `'userId'` — because on the foreign side it is not a column of the model the caller named.

## 2. `soft-deletes.ts` passed the literal `"soft delete"` as the model

Not a model name, and never can be. The real one was in hand: `softDelete(model)` receives the model class and every generated model carries `static $schema = schema.User`. The `{ update(args): Promise<T> }` structural parameter type was hiding it rather than the value being absent — so the fix is to widen the type, not to plumb a new argument.

## 3. `resolveStrategy` passed `"Model"` and `"$exec"`

A base class, and something that is not one of the thirteen operations. Its one caller (`Model.ts:593`) has both in scope. Made **required**, so the next caller cannot quietly reintroduce a placeholder — the same `tsc`-enforces-the-convention pattern as #102.

## Deliberately unchanged

- `sql.ts` passes `"DB"`. `DB.sql` is the raw-SQL surface and there genuinely is no model. A placeholder is the right answer when nothing else exists, and this is that case.
- `policy.ts:719` and `:745` put the *operation* in the `argument` slot, which reads as "the whole query rather than one argument". Their model and operation are already the caller's.

## Coverage

**Two of the three had no test at all**, which is how they stayed wrong:

- `assertDisconnectable` matched nothing in the repo, on any of its three sites. The fixtures could not reach it either — none has a required foreign key on a to-many child, which is exactly the configuration the two broken sites serve. The test clones one.
- `softDelete` / `softDeleteMany` were untested entirely. The new file covers the refusal *and* the delegation, since a refusal test alone would pass against a version that refuses everything.
- `resolveStrategy` was untested. The new file covers the dialect defaults and the override as well as the refusal.

**Verified against the unfixed source**: every model assertion fails. One nuance worth stating — the owning-side `disconnect` case fails on the *qualified column name*, not the model, which was already correct there. It is kept because the divergence between the two branches was the defect, and a test on only the broken side would let them drift apart again.

## Verification

- 1077 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -590,7 +590,10 @@ export abstract class Model {
     // `defaultStrategy`, and either way it reaches the plan key — two strategies
     // emit different SQL for the same arguments, so sharing a plan between them
     // would run one request's statement for the other's.
-    const strategy = resolveStrategy(options?.strategy, dialect);
+    const strategy = resolveStrategy(options?.strategy, dialect, {
+      model: schema.name,
+      operation: op,
+    });
 
     // `createMany` past the driver's parameter ceiling, which used to raise.
     //
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -8,7 +8,7 @@ import {
   relatedSchema,
   resolveLink,
 } from "./plan-relations";
-import { matchUniqueKey } from "./unique";
+import { matchUniqueKey, type RefusalOrigin } from "./unique";
 
 /**
  * Nested writes: `connect`, `connectOrCreate`, shallow `create`, and
@@ -417,7 +417,11 @@ function planOwningSide(
    * of an unrelated update is the kind of thing to implement deliberately.
    */
   if (key === "disconnect") {
-    assertDisconnectable(schema, relation, fkField, operation);
+    assertDisconnectable(schema, relation, fkField, {
+      model: schema.name,
+      operation,
+      argument: `data.${relation.name}.disconnect`,
+    });
 
     if (operand !== true) {
       throw new UnsupportedQueryError(
@@ -922,7 +926,11 @@ function planForeignSide(
 
   if (key === "set") {
     assertNamedRows(schema, relation, child, operand, "set", operation);
-    assertDisconnectable(child, relation, childField, operation);
+    assertDisconnectable(child, relation, childField, {
+      model: schema.name,
+      operation,
+      argument: `data.${relation.name}.set`,
+    });
 
     out.after.push({
       relation: relation.name,
@@ -1024,7 +1032,13 @@ function planForeignSide(
 
   if (key === "disconnect" || key === "delete") {
     const deleting = key === "delete";
-    if (!deleting) assertDisconnectable(child, relation, childField, operation);
+    if (!deleting) {
+      assertDisconnectable(child, relation, childField, {
+        model: schema.name,
+        operation,
+        argument: `data.${relation.name}.disconnect`,
+      });
+    }
 
     assertNamedRows(schema, relation, child, operand, key, operation);
 
@@ -1918,23 +1932,32 @@ function assertNamedRows(
  * is the only thing that knows, so the refusal says which column and why rather
  * than letting the database report a not-null violation from inside a nested
  * step.
+ *
+ * `owner` and `caller` are separate parameters because they are separate
+ * questions, and one parameter answering both is what went wrong. The column
+ * lives on whichever side holds the foreign key — the caller's row on a to-one,
+ * the child's on a to-many — and that is what the *message* has to name. The
+ * *structured* model is always the caller's. Passing `owner` for both meant a
+ * `User.update` reported `model = "Account"` on the foreign side while the
+ * owning side reported `User`, from the same function (#112).
  */
 function assertDisconnectable(
   owner: ModelSchema,
   relation: RelationSchema,
   fieldName: string,
-  operation: string,
+  caller: RefusalOrigin,
 ): void {
   const field = owner.fields[fieldName];
   if (field && field.nullable) return;
 
   throw new UnsupportedQueryError(
-    `data.${relation.name}.disconnect`,
-    owner.name,
-    operation,
-    `'${fieldName}' is required, so there is no value to leave behind — a ` +
-      `disconnected row would have to be deleted or repointed instead. Prisma ` +
-      `does not offer 'disconnect' on a required relation either.`,
+    caller.argument,
+    caller.model,
+    caller.operation,
+    `'${owner.name}.${fieldName}' is required, so there is no value to leave ` +
+      `behind — a disconnected row would have to be deleted or repointed ` +
+      `instead. Prisma does not offer 'disconnect' on a required relation ` +
+      `either.`,
   );
 }
 
--- a/packages/gemi/orm/compile/strategy.test.ts
+++ b/packages/gemi/orm/compile/strategy.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, test } from "vitest";
+
+import { PostgresDialect } from "../dialect/postgres";
+import { SqliteDialect } from "../dialect/sqlite";
+import { UnsupportedQueryError } from "../errors";
+import { lateralStrategy } from "./lateral";
+import { batchedStrategy } from "./plan-relations";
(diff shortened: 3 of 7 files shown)


nstfkc:
## Review — three refusals reported a model that was not the caller's (closes #112)

Read the diff against its true base (`a5ef6b4`, #111). 1077 ORM tests pass; the 7 non-ORM failures in a full package run (router, `parseTranslation`) are present on the base too.

**The sweep is exactly right.** I checked it with the TypeScript AST rather than a grep, since the claim is a count:

```
construction sites, non-test, orm/   143      (matches)
string literal in the model slot     base 7  ->  head 5
remaining 5                          all "DB" in sql.ts
```

So the only literals left are the ones the PR defends, and the defence is correct — `DB.sql` has no model, and a placeholder is the right answer when nothing else exists.

Worth noting what that scan *cannot* do: it finds two of the three. `assertDisconnectable` passed `owner.name` — a variable, structurally indistinguishable from every correct site — so no scan for placeholders reaches it. It needed the question "what lands in the slot", asked per site. That is the one that had two wrong call sites out of three, and the mechanical check would have declared the file clean.

Verified against the unfixed source: **7 model assertions fail** across the three new files (3 in `write.test.ts`, 4 in the two new ones), and pass after. The point about keeping the already-correct owning-side case is the right call — the two branches disagreeing *was* the defect, so a test on only the broken side lets them drift again.

### The `set` site's `argument` was wrong too, not just its `model`

The table says *"`operation` and `argument` were right; only `model` was wrong"*. That holds for the `disconnect` sites. It does not hold for line 925. Measured on the base:

```
base   foreign disconnect   model=Account   argument=data.accounts.disconnect
base   foreign set          model=Account   argument=data.accounts.disconnect   <-- caller wrote `set`
head   foreign set          model=User      argument=data.accounts.set
```

The whole base message for a `set`:

> gemi ORM does not support **'data.accounts.disconnect'** yet (**Account**.update). '**userId**' is required…

Three fields describing a query nobody wrote: an operand the caller did not use, on a model that is not theirs, naming an unqualified column that is not on it. That is the #85 shape — a refusal misnaming its own origin — and it is the more confusing of the two failures, because `disconnect` at least named the operand correctly. Your fix corrects it (the new test asserts `data.accounts.set`), so this is only the description under-claiming; I would rather it claimed it, because it is the better example of why one parameter could not answer both questions.

### `?? "unknown"` is a placeholder, in the slot this PR exists to clear

> *a string that is not a model name can never be what an application matches on*

`"unknown"` is such a string, and it is reachable — `softDelete` is exported from `orm/index.ts`, and `$schema` being optional means this compiles today:

```ts
softDelete({ update: async (args) => …  })   // tsc: clean
  -> refusal with model = "unknown"           // measured
```

The stated reason is that the parameter is structural. It can stay structural *and* require `$schema` — that is the same convention this PR applies to `resolveStrategy` one file over, and to `render`'s `origin` and `changedFields`' `schema` before it. Making it required:

```
tsc --noEmit                          clean
orm tests                             1077 passed
softDelete({ update })                error TS2345: Property '$schema' is missing
```

Nothing in the repo passes a schema-less model — the one real caller is `softDelete(GuardedUser)`, and every generated model carries it, as the comment says. So the fallback is not covering a case that exists; it is covering the case the required form would reject at the call site. (I could not typecheck the template from my tree — its `gemi` resolves through `node_modules` — so that is on the `packages/gemi` graph plus reading the caller.)

### `origin.operation: string` still admits `"$exec"`

The complaint about `resolveStrategy` was two-part: `"Model"` is not a model, and `"$exec"` *"is not one of the thirteen operations"*. Making the parameter required fixes the first — a model name is a dynamic string, so required is the strongest available guard. The second is stronger than that, because the union is right there in `plan.ts`:

```
origin: { model: string; operation: string }     operation: "$exec"  -> compiles
origin: { model: string; operation: Operation }  operation: "$exec"  -> TS2322
```

`$exec`'s own `op` is already `Operation`, so the call site needs no change:

```
tsc --noEmit    clean
orm tests       1077 passed
```

That turns the sentence in the description into something `tsc` enforces, which is the argument the PR itself makes for the required parameter. (`Operation` has fifteen members now — `aggregate` and `groupBy` joined since that count was written.)

None of this is blocking. The sweep is sound and the three defects are real, and finding the third one is the part a checker could not have done for you.

Issue #114, opened 2026-07-29
45 of 143 refusal sites are never constructed by any test


nstfkc:
The last three defects I found — #108, #110, #112 — were all in code no test reached. #112's `assertDisconnectable`, `softDelete` and `resolveStrategy` had *no test at all* between them, on any call site. That is a pattern worth measuring rather than continuing to find one function at a time.

## Method

`@vitest/coverage-v8` is not installed, and line coverage would answer a slightly different question anyway. Instead I temporarily instrumented `UnsupportedQueryError`'s constructor to record the first `orm/` stack frame that is not `errors.ts` — the construction site — and ran both suites.

Then diffed that against every `new UnsupportedQueryError` / `UnsupportedByDesignError` / `InvalidArgumentError` in `orm/`, found by parsing rather than grepping for messages.

**143 sites. 98 reached. 45 never constructed**, by the 1077 unit tests and the 592 template tests together.

| file | unreached |
| --- | --- |
| `compile/nested-writes.ts` | 10 |
| `compile/select.ts` | 8 (all of them) |
| `compile/write.ts` | 7 |
| `compile/group-by.ts` | 5 |
| `compile/orderBy.ts` | 4 |
| `compile/where.ts` | 4 |
| `compile/aggregate.ts` | 2 |
| `compile/order-relation.ts` | 2 |
| `compile/index.ts`, `compile/plan-relations.ts`, `compile/read.ts` | 1 each |

## Worked example: `select.ts`, where all eight are unreached

Probed each by triggering it through the public API and reading the construction site off the error's own stack:

| line | refusal | reachable |
| --- | --- | --- |
| 28 | `select` + `include` | **no** |
| 41 | `select` + `omit` | yes |
| 59 | `select` is not an object | yes |
| 107 | `select.x` is not a boolean | yes |
| 161 | `omit` is not an object | yes |
| 179 | `omit` names a relation | yes |
| 193 | `omit.x` is not a boolean | yes |
| 213 | `omit` omits every field | yes |

Seven are reachable and answer correctly — so "unreached" is not "wrong", and I would rather report that than imply otherwise.

The eighth cannot fire. `resolveSelection` has four callers and **every one of them checks `select` + `include` first**:

| caller | pre-empted by |
| --- | --- |
| `read.ts:206` | `read.ts:341` |
| `write.ts:1277` | `write.ts:1411` |
| `plan-relations.ts:660` | `plan-relations.ts:506` |
[...]

PR #115, opened 2026-07-29
select.ts's refusals, none of which any test reached

nstfkc:
Part of #114. Stacked on #113.

## Why this file

#108, #110 and #112 were all in code no test reached — #112's `assertDisconnectable`, `softDelete` and `resolveStrategy` had **no test at all** between them, on any call site. Three for three is a pattern worth measuring rather than continuing to find one function at a time.

`@vitest/coverage-v8` is not installed, and line coverage answers a slightly different question anyway. So I temporarily instrumented `UnsupportedQueryError`'s constructor to record the first `orm/` stack frame that is not `errors.ts`, ran both suites, and diffed against every construction site in `orm/` found by parsing rather than by grepping messages.

**143 sites. 98 reached. 45 never constructed.** All eight of `select.ts`'s were in the 45 — the only file where *every* refusal was unreached, which is why this starts there. Full per-file breakdown in #114.

## Seven were reachable and correct

| line | refusal | reachable |
| --- | --- | --- |
| 28 | `select` + `include` | **no** |
| 41 | `select` + `omit` | yes |
| 59 | `select` is not an object | yes |
| 107 | `select.x` is not a boolean | yes |
| 161 | `omit` is not an object | yes |
| 179 | `omit` names a relation | yes |
| 193 | `omit.x` is not a boolean | yes |
| 213 | `omit` omits every field | yes |

Unreached is not the same as wrong, and I would rather report that than imply otherwise. No defect here. They are tested now because the three above are the argument for it — each asserts the structured fields, not just that something threw.

## The eighth cannot fire

`resolveSelection` has four callers and **every one checks `select` + `include` first**:

| caller | pre-empted by |
| --- | --- |
| `read.ts:206` | `read.ts:341` |
| `write.ts:1277` | `write.ts:1411` |
| `plan-relations.ts:660` | `plan-relations.ts:506` |
| `lateral.ts:291` | `plan-relations.ts:506` |

**This is not a stale guard to delete**, and I want to be careful about that, because #111 deleted two shadowed checks and the reasoning does not carry over. There the shadowed check was in the same path with identical output. Here the callers say it *better* — `plan-relations.ts` reports `'accounts: select + include'` with the relation path, which the generic one has no way to know — and this one is the floor under a fifth caller.

What it lacked was anything recording the difference between defensive and dead. So:

- the comment names which caller pre-empts each path, so the next reader does not re-derive it;
- the test calls `resolveSelection` **directly**, which is the honest way to reach it — going through the public API would test whichever caller got there first;
- two more cases pin the pre-emption itself, including the relation-path spelling. If a caller's own check is removed, those change and the direct one does not.

## Verification

Re-measured with the instrumentation after: all eight reached. The unit-suite unreached total drops **49 → 41**, which is 45 → 37 counting the template suite.

- 1088 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

The instrumentation was reverted; it is a recipe, not committed code.

### DIFF
--- a/packages/gemi/orm/compile/select.test.ts
+++ b/packages/gemi/orm/compile/select.test.ts
@@ -1,0 +1,159 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { SqliteDialect } from "../dialect/sqlite";
+import { UnsupportedQueryError } from "../errors";
+import { account, organization, user } from "../fixtures";
+import * as registry from "../registry";
+import { compileRead } from "./read";
+import { resolveSelection } from "./select";
+import { compileWrite } from "./write";
+
+const sqlite = new SqliteDialect();
+
+/**
+ * `select.ts`'s refusals — every one of which was unreached by any test in the
+ * repo, unit or template (#114).
+ *
+ * Found by instrumenting `UnsupportedQueryError`'s constructor to record its
+ * call site and diffing that against every construction site in `orm/`: 143
+ * sites, 98 reached, 45 not. All eight of this file's were in the 45.
+ *
+ * Seven turned out reachable and correct, which is worth stating plainly —
+ * unreached is not the same as wrong. The eighth is the interesting one, and
+ * it has its own case at the bottom.
+ */
+describe("resolveSelection refuses", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    registry.register("User", class { static $schema = user });
+    registry.register("Account", class { static $schema = account });
+    registry.register("Organization", class { static $schema = organization });
+  });
+
+  const read = (args: unknown) => {
+    try {
+      compileRead(user, "findMany", args as never, sqlite);
+      return null;
+    } catch (error) {
+      return error as UnsupportedQueryError;
+    }
+  };
+
+  const every = Object.fromEntries(
+    Object.keys(user.fields).map((name) => [name, true]),
+  );
+
+  test.each([
+    [
+      "select with omit",
+      { select: { id: true }, omit: { name: true } },
+      "select + omit",
+      /describe different column lists/,
+    ],
+    ["select is an array", { select: [1] }, "select", /Expected an object/],
+    [
+      "a selected field is not a boolean",
+      { select: { id: 1 } },
+      "select.id",
+      /Expected true or false/,
+    ],
+    ["omit is an array", { omit: [1] }, "omit", /Expected an object of fields/],
+    [
+      "omit names a relation",
+      { omit: { accounts: true } },
+      "omit.accounts",
+      /is a relation, not a column/,
+    ],
+    [
+      "an omitted field is not a boolean",
+      { omit: { name: 1 } },
+      "omit.name",
+      /Expected true or false/,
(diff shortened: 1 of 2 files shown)


nstfkc:
## Review — `select.ts`'s refusals, none of which any test reached (part of #114)

Read the diff against its true base (`003a229`, #113). 1088 tests pass.

**I reproduced the measurement independently** rather than taking the numbers, since the numbers are the argument. Same recipe — instrument `UnsupportedQueryError`'s constructor to record the first `orm/` frame that is not `errors.ts`, enumerate the sites by AST:

```
construction sites, orm/, non-test        143      (matches)
distinct sites reached, unit suite  base   94
                                    head  102
unreached                           base   49  ->  head 41
```

49 → 41 exactly as claimed, and the diff of the two reached sets is 8 sites, all of them in `select.ts`. Nothing else changed reach in either direction.

**The eighth really cannot be reached, and I went looking for a way.** Across the whole ORM suite the line-28 throw fires exactly once, from `select.test.ts` itself — no caller. Then a fuzz over 2 dialects × 12 operations × 19 argument shapes (root and nested `select`+`include`, buried under valid outer nodes, each relation kind, both strategies):

```
resolveSelection entered   356 times
  read.ts:206              120
  write.ts:1277            108
  plan-relations.ts:660    108
  lateral.ts:291            20      <- valid shapes included so lateral actually folds
line 28 reached              0
```

The control matters: the first version of that fuzz never reached `lateral:291` at all, so its zero would have meant nothing. With all four call sites exercised the zero is worth something.

**Not deleting it is the right call, and the distinction from #111 is the real one.** There the shadowed check produced identical output on the same path; here `plan-relations.ts:506` says strictly more — `'accounts: select + include'` — and the generic site has no relation path to name. Recording *which* caller pre-empts each path, in the comment, is what makes it defensive rather than dead, and it is the thing that was missing. Testing `resolveSelection` directly is right for the same reason: through the public API you would be testing whichever caller got there first, and the two pre-emption cases pin that separately.

### `select.ts` has nine refusals, not eight — and the ninth was already reached

The table's eight are lines 28, 41, 59, 107, 161, 179, 193, 213 on the base. The AST finds a ninth:

```
select.ts:119   "select" / "At least one field must be selected."
```

It is absent from the table, and it is the one site in the file the base suite **did** reach — it is the sole `select.ts` entry in my base reached-set. So:

> *All eight of `select.ts`'s were in the 45 — the only file where every refusal was unreached*

is eight of nine, and the file was not one where every refusal was unreached.

None of this changes the conclusion — eight unreached in one file is still the right place to start, and the +8 is exactly right. What it changes is the confidence the sentence carries into #114, since "the only file where…" is the kind of claim a later change gets checked against. And the omission is the failure mode the methodology section says it avoided: the site list came from parsing, but this table did not — a message-shaped reading finds the seven that describe a malformed argument and skips the one that describes an empty result.

Cheap to settle for the rest of #114: the same AST pass that produced the 143 also produces the per-file counts, so the breakdown can come from the parse rather than being assembled beside it.

PR #117, opened 2026-07-29
A to-one whose foreign key is on the child was planned as a list

### LINKED ISSUE #116: A to-one whose foreign key is on the child is planned as if it were a list
`planForeignSide` contains exactly **one** `relation.kind` check — the `createMany` guard at line 1156, whose comment claims it is reachable. It is: a one-to-one whose foreign key lives on the child, which is Prisma's

```prisma
model User    { profile Profile? }
model Profile { userId Int @unique   user User @relation(fields: [userId], references: [id]) }
```

**No fixture in the repo has that shape**, and no generated template model does either. So the branch is unreached (#114), and — because it is the only `kind` check in the function — every *other* operand on such a relation is planned as though the child were a list.

## Measured

Built the missing shape (`kind: "one"`, empty `from`/`to`, resolved through the child's opposing relation) and compiled every operand on both sides of a to-one:

| operand | owning to-one (key here) | foreign to-one (key on child) |
| --- | --- | --- |
| `create` | compiles | compiles |
| `connect` | compiles | compiles |
| `connectOrCreate` | compiles | compiles |
| `update` | compiles (bare data) | **refused** — "Expected an object with 'where' and 'data'" |
| `delete` | refused, by name | **refused** — "Expected an object naming a unique field" |
| `upsert` | refused, by name | **refused** — wants `{ where, create, update }` |
| `set` | refused, by name | refused (via `assertDisconnectable`) |
| `createMany` | refused, by name | refused, by name — the one `kind` check |
| `updateMany` | refused, by name | **compiles** |
| `deleteMany` | refused, by name | **compiles** |

## What Prisma actually accepts

Read off the generated client rather than reasoned about — `@prisma/client` 6.19.2, `.prisma/client/index.d.ts`:

```ts
export type UserUpdateOneWithoutAccountsNestedInput = {
  create?; connectOrCreate?; upsert?; disconnect?; delete?; connect?; update?
}
export type AccountUpdateManyWithoutUserNestedInput = {
  create?; connectOrCreate?; upsert?; createMany?; set?; disconnect?;
  delete?; connect?; update?; updateMany?; deleteMany?
}
```

A to-one input has **no `createMany`, `set`, `updateMany` or `deleteMany` key at all**. Note also `disconnect?: UserWhereInput | boolean` and `delete?: UserWhereInput | boolean` — a boolean on a to-one, not a unique-key object — and `update` takes the data directly.

So the table splits into two different problems:

**Accepts what Prisma rejects.** `updateMany` and `deleteMany` compile on a foreign-side to-one. Prisma has no such key there. This is the half that can do damage rather than merely annoy, and it is the half the owning side already refuses by name — #107 did that work for `planOwningSide` and it was never done for the other side.

**Refuses what Prisma accepts, with a misleading reason.** `update`, `delete` and `upsert` are refused because the operand does not look like the to-many spelling — "Expected an object with 'where' and 'data'" — when the real answer is that a to-one takes the bare data, or a boolean. The refusal blames the caller's syntax for what is actually an unimplemented relation shape, which is the kind of message this repo has repeatedly found worth fixing (#101, #107).

## Reads are fine

Worth recording, since I checked: `include` plans correctly on both strategies, the relation plan carries `kind: "one"` so the shaper produces an object rather than an array, `where: { profile: { is/isNot } }` compiles, `orderBy` compiles, and `_count` on it is refused by name. The gap is writes only.

## Acceptance

- A fixture with the shape, since its absence is the root cause and every item below needs it.
- `updateMany`, `deleteMany` and `set` refused by name on a foreign-side to-one — the operand does not exist on that input, so the refusal should say so rather than the operand happening to fail some other assertion.
- `update`, `delete` and `upsert` refused with a reason that names the relation shape rather than the operand's spelling, until they are implemented.
- A test that walks *both* sides of a to-one for every operand, the way #113's walk did — the two sides disagreeing is the defect, and only a both-sides table catches that.

nstfkc:
Closes #116. Stacked on #115.

#114 found `nested-writes.ts:1156` unreached — a `relation.kind` guard whose comment claims it is reachable. It is, and chasing why nothing reached it turned up something larger.

`planForeignSide` contains **exactly one** `relation.kind` check: that one. The shape that reaches it is Prisma's one-to-one with the key on the child —

```prisma
model User    { profile Profile? }
model Profile { userId Int @unique   user User @relation(fields: [userId], references: [id]) }
```

— and **no fixture in the repo has it**, nor does any generated template model. So this side was only ever reached with `kind: "many"`, and being the only `kind` check, every other operand planned the child as a list.

## Measured

Built the shape and compiled every operand on both sides of a to-one:

| operand | owning (key here) | foreign (key on child) |
| --- | --- | --- |
| `create` / `connect` / `connectOrCreate` | compiles | compiles |
| `update` | compiles (bare data) | **refused** — "Expected an object with 'where' and 'data'" |
| `delete` | refused by name | **refused** — "Expected an object naming a unique field" |
| `upsert` | refused by name | **refused** — wants `{ where, create, update }` |
| `set` | refused by name | refused, but via `assertDisconnectable` |
| `createMany` | refused by name | refused by name — the one `kind` check |
| `updateMany` | refused by name | **compiles** |
| `deleteMany` | refused by name | **compiles** |

## What Prisma accepts

Read off the generated client (`@prisma/client` 6.19.2), not reasoned about:

```ts
export type UserUpdateOneWithoutAccountsNestedInput = {
  create?; connectOrCreate?; upsert?; disconnect?; delete?; connect?; update?
}
```

**No `createMany`, `set`, `updateMany` or `deleteMany` key at all** — the to-many input has all eleven. So the table splits in two:

**Accepts what Prisma rejects.** `updateMany` and `deleteMany` compiled. This is the half that can act rather than merely annoy, and it is the half `planOwningSide` already refuses by name — #107 did that work for one side and it was never done for the other. Now refused in the same words.

**Refuses what Prisma has, for the wrong reason.** `update`, `delete` and `upsert` were refused for not matching the to-many spelling, when the real answer is that the shape is unimplemented here. The refusal now names the shape. `set` is the same story with a subtler cause: it *did* refuse, but through `assertDisconnectable` — `'Profile.userId' is required` — which is true and beside the point. The walk asserts the reason rather than that something threw, so it catches that.

## The fixture

Its absence is the root cause, so it is added: `profile`, plus `userWithProfile` carrying the far side (`kind: "one"` with empty `from`/`to`, resolved through the child's opposing relation). Kept as a separate export rather than folded into `user`, because several tests assert `user`'s emitted column list verbatim.

## Tests

A table over **both sides**, because the two disagreeing is the defect and a one-sided test cannot see it — the same shape as #113's walk.

**Verified against the unfixed source: six of the ten cases fail**, and they are the six measured gaps. The four that pass are `createMany` (already refused) and the three operands that genuinely work, so the table is covering rather than padding.

## Reads are fine

Checked, and worth recording since the fixture now makes it testable: `include` plans on both batched and lateral, the relation plan carries `kind: "one"` so the shaper returns an object rather than an array, `where: { profile: { is / isNot } }` and `orderBy` compile, and `_count` on it is refused by name. The gap was writes only.

## Verification

- 1098 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -704,6 +704,54 @@ function planForeignSide(
   out.keyFields.push(parentField);
 
   /**
+   * **A to-one whose key is on the child**, which this function otherwise plans
+   * as though the child were a list.
+   *
+   * It had exactly one `kind` check — `createMany`'s, further down — because no
+   * fixture had the shape, so nothing ever reached this side with `kind: "one"`
+   * (#116). Everything else took the to-many spelling: `updateMany` and
+   * `deleteMany` *compiled*, and `update` / `delete` / `upsert` were refused
+   * for looking wrong rather than for being unimplemented here.
+   *
+   * Measured against the generated client rather than reasoned about. Prisma's
+   * to-one nested input is
+   *
+   *     { create, connectOrCreate, upsert, disconnect, delete, connect, update }
+   *
+   * with no `createMany`, `set`, `updateMany` or `deleteMany` key at all — so
+   * the first group below is refused for the same reason `planOwningSide`
+   * refuses it, and says so in the same words. The second group exists on that
+   * input but not here yet, and the refusal now says *that* instead of
+   * complaining about the operand's shape.
+   */
+  if (relation.kind !== "many") {
+    if (key === "set" || key === "updateMany" || key === "deleteMany") {
+      throw new UnsupportedQueryError(
+        `data.${relation.name}.${key}`,
+        schema.name,
+        operation,
+        `'${relation.name}' is a to-one: there is a single ${child.name} row, ` +
+          `so there is no set of rows for '${key}' to act on. Prisma does not ` +
+          `offer it on a to-one either. Use 'connect', 'disconnect' or ` +
+          `'update'.`,
+      );
+    }
+
+    if (key === "update" || key === "delete" || key === "upsert") {
+      throw new UnsupportedQueryError(
+        `data.${relation.name}.${key}`,
+        schema.name,
+        operation,
+        `'${relation.name}' is a to-one whose foreign key lives on ` +
+          `${child.name}, and '${key}' is not implemented for that shape yet. ` +
+          `On a to-one Prisma takes the data directly rather than the ` +
+          `{ where, data } form this side expects. Reach the row through the ` +
+          `${child.name} model directly, or use 'connect' and 'disconnect'.`,
+      );
+    }
+  }
+
+  /**
    * `disconnect` and `delete` on this side both act on rows the caller named by
    * unique key — which is what makes them expressible at all, and what puts
    * them on the supported side of this file's line.
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -9,7 +9,17 @@ import {
   UnknownFieldError,
   UnsupportedQueryError,
 } from "../errors";
-import { account, bare, organization, post, reading, tag, user } from "../fixtures";
+import {
+  account,
+  bare,
+  organization,
+  post,
+  profile,
+  reading,
+  tag,
+  user,
+  userWithProfile,
+} from "../fixtures";
 import * as registry from "../registry";
 import { compileWrite } from "./write";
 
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — a to-one whose foreign key is on the child was planned as a list (closes #116)

Read the diff against its true base (`af8ed23`, #115). 1098 tests pass; six of the ten new cases fail against the unfixed source, and they are the six gaps.

**The diagnosis is the valuable part.** One unreached `kind` guard → it is the *only* `kind` guard on that side → no fixture has the shape → therefore every other operand on that side was only ever exercised as a list. That chain turns "a line no test covers" into "a whole branch of the compiler was never reached with the argument it exists to distinguish", and it is the first thing #114 has produced that was not visible from the coverage number alone.

Verified by walking every operand on both sides, on both branches:

```
base   create/connect/connectOrCreate  compiles      head  compiles
base   set                             assertDisconnectable    head  refused by name
base   createMany                      refused by name         head  refused by name
base   updateMany                      COMPILES                head  refused by name
base   deleteMany                      COMPILES                head  refused by name
```

The fixture being the root cause is right, and adding `userWithProfile` beside `user` rather than folding it in is the correct call — several tests assert `user`'s emitted column list verbatim.

**The read claim holds** — I checked it rather than taking it, since the fixture makes it newly testable:

```
include: { profile: true }   batched/sqlite   relations=[profile:one]
                             lateral/pg       relations=[profile:one]
```

### The table understates its own finding: five operands compiled, not two

`update`, `delete` and `upsert` are marked **refused** on the foreign side. That is true of the *to-one* spelling — `update: { bio: "x" }` fails "Expected an object with 'where' and 'data'". It is not true of the to-many spelling, which compiled and issued statements. Driving the planned steps on the base with a stand-in executor:

```
update      Profile.findFirst  {AND:[{id:1},{userId:42}]}  -> RecordNotFoundError
delete      Profile.findFirst  {AND:[{id:1},{userId:42}]}  -> threw
upsert      Profile.findFirst  {AND:[{id:1},{userId:42}]}  -> miss -> Profile.create
updateMany  Profile.updateMany where {userId: 42}
deleteMany  Profile.deleteMany where {userId: 42}
```

So the split is not two-that-act and three-that-annoy. **Five** operands reached the database on a relation whose Prisma input has no `updateMany`, `deleteMany` or `set` key, and takes `update`/`delete`/`upsert` in a different shape entirely. The `upsert` row is the one I would most want in the table: a miss on a to-one creates a second `Profile` for a user that already has one, which the `@unique` on `userId` then rejects — an error from the database about a constraint, for a query the compiler should have refused.

Your fix already covers all of this — the new guard is on `key` before any spelling check, so the head refuses all seven regardless of how they are written, and I confirmed that. This is the description selling the change short, which matters only because the table is the artifact #114 accumulates and "refused" in a cell reads as safe.

Cheap to settle in general: the walk that produced my numbers passes each operand in **both** spellings. A refusal that depends on which one the caller used is exactly the confusion this PR is fixing, so the table is worth measuring the same way — otherwise the next such row records the spelling rather than the shape.

PR #118, opened 2026-07-29
The remaining 36 refusals nothing reached, and the one nothing can

### LINKED ISSUE #114: 45 of 143 refusal sites are never constructed by any test

nstfkc:
Closes #114. Stacked on #117.

#114 instrumented `UnsupportedQueryError`'s constructor, ran both suites, and diffed the recorded call sites against every construction site in `orm/`: **145 sites, 45 never constructed**. #115 took `select.ts`'s eight; #116 came out of chasing one in `nested-writes.ts`. This is the remaining 36.

## Why the ordinary tests miss them

Every one is a **shape** guard — the argument is the wrong *type*, not a name the schema does not know. The existing tests pass well-formed arguments and vary the names, so they exercise `UnknownFieldError` thoroughly and these not at all.

## Triage

I built an input for each and read the construction site off the error's own stack, rather than trusting the input to land where I aimed it. That check earned its keep immediately — three of my first attempts hit a *different* guard than intended:

| aimed at | actually hit | why |
| --- | --- | --- |
| `where.ts:387` (compound key) | `where.ts:125` `UnknownFieldError` | no fixture has a compound unique; `membership`'s compound `@@id` is the one that works |
| `aggregate.ts:314` | `aggregate.ts:279` | 314 is `count`'s `select`, not an aggregate's |
| `nested-writes.ts:1515` | `write.ts:1019` | `tags` is on `post`, not `user` |

**35 are reachable and answer correctly.** No defect among them — worth stating plainly rather than implying a haul. The value is that they are pinned now, and the three defects #114 was filed off the back of (#108, #110, #112, all in unreached code) are the argument for pinning them.

Asserted on `argument` rather than message text, because `argument` is what identifies *which* guard fired — several share a message like "Expected an object."

## The 36th cannot fire

`parseOrderBy` refuses a relation key when called without an `OrderContext`. Four of its five callers pass one. The fifth is `lateral.ts:375` — and the lateral strategy **declines** a node whose `orderBy` names a relation before folding it, so that call is only ever reached with an ordering the strategy already accepted.

Measured rather than assumed, and both halves are asserted:

```
include: { accounts: { orderBy: { user: { name: "asc" } } } }  → no `lateral` in the SQL
include: { accounts: { orderBy: { id: "asc" } } }              → `lateral` present
```

Like `select.ts:28` in #115, this is a **floor rather than dead code** — what a future context-free caller would land on. It is reached by direct call, with a comment saying which caller pre-empts it, so the distinction is recorded instead of re-derived.

## Result

Re-measured with the instrumentation after: **145 sites, 0 unreached**, down from 45.

- 1136 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

The instrumentation stays a recipe, not committed code — the method is written up in #114 for the next time this is worth asking.

### DIFF
--- a/packages/gemi/orm/compile/refusals.test.ts
+++ b/packages/gemi/orm/compile/refusals.test.ts
@@ -1,0 +1,232 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { PostgresDialect } from "../dialect/postgres";
+import { SqliteDialect } from "../dialect/sqlite";
+import { UnsupportedQueryError } from "../errors";
+import {
+  account,
+  membership,
+  organization,
+  post,
+  profile,
+  tag,
+  userWithProfile,
+} from "../fixtures";
+import * as registry from "../registry";
+import { compileAggregate } from "./aggregate";
+import { compileGroupBy } from "./group-by";
+import { compile } from "./index";
+import { lateralStrategy } from "./lateral";
+import { parseOrderBy } from "./orderBy";
+import { compileRead } from "./read";
+import { compileWrite } from "./write";
+
+const sqlite = new SqliteDialect();
+const postgres = new PostgresDialect();
+
+/**
+ * The refusals nothing reached.
+ *
+ * #114 instrumented `UnsupportedQueryError`'s constructor, ran both suites, and
+ * diffed the recorded call sites against every construction site in `orm/`:
+ * **145 sites, 36 never constructed**. `select.ts`'s eight were taken in #115
+ * and `nested-writes.ts`'s `kind` guard led to #116; this is the remaining 36.
+ *
+ * Every one is a shape guard — the argument is the wrong *type* rather than
+ * naming something unknown, which is why the ordinary tests miss them: they
+ * pass well-formed arguments. Triaged by building an input for each and reading
+ * the construction site off the error's own stack.
+ *
+ * **35 are reachable and answer correctly.** No defect. Worth saying plainly
+ * rather than implying a haul — the value is that they are pinned now, and the
+ * three defects #114 was filed off the back of are the argument for pinning
+ * them. The 36th has its own case at the bottom.
+ *
+ * Asserted on `argument` rather than the message text, because `argument` is
+ * what identifies which guard fired — several share a message.
+ */
+describe("shape guards", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    for (const [name, schema] of [
+      ["User", userWithProfile],
+      ["Account", account],
+      ["Organization", organization],
+      ["Post", post],
+      ["Tag", tag],
+      ["Profile", profile],
+      ["Membership", membership],
+    ] as const) {
+      registry.register(name, class { static $schema = schema });
+    }
+  });
+
+  const refusal = (run: () => unknown) => {
+    try {
+      run();
+      return null;
+    } catch (error) {
+      return error as UnsupportedQueryError;
+    }
+  };
+
+  const read = (args: unknown, op = "findMany") => () =>
+    compileRead(userWithProfile, op as never, args as never, sqlite);
+  const write = (args: unknown, op = "update") => () =>
+    compileWrite(userWithProfile, op as never, args as never, sqlite);
+  const groupBy = (args: unknown) => () =>
+    compileGroupBy(userWithProfile, "groupBy" as never, args as never, sqlite);
+
+  // [what it guards, the query, the `argument` it reports, a fragment of why]
+  const CASES: [string, () => unknown, string, RegExp][] = [
+    // --- ordering ---------------------------------------------------------
+    ["orderBy is not an object", read({ orderBy: 5 }), "orderBy", /Expected an object or an array of objects/],
+    [
+      "orderBy long form has an unknown nulls",
+      read({ orderBy: { name: { sort: "asc", nulls: "sideways" } } }),
+      `orderBy.name.nulls: "sideways"`,
+      /Expected "first" or "last"/,
+    ],
+    ["orderBy value is neither form", read({ orderBy: { name: 5 } }), "orderBy.name", /Expected "asc", "desc", or \{ sort, nulls \}/],
+    [
+      "a relation _count orders by neither direction",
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the remaining 36 refusals, and the one nothing can reach (closes #114)

Read the diff against its true base (`0e1cda0`, #117) — one file, all tests. I re-ran the whole measurement rather than taking it.

```
construction sites, orm/, non-test    145      (144 UnsupportedQueryError + 1 UnsupportedByDesignError)
distinct sites reached, unit suite    141
unit tests                           1136
```

**The `parseOrderBy` claim holds**, checked rather than read — this is the one assertion in the PR that a reader cannot verify by inspection, since it depends on the strategy declining before the call:

```
include: { accounts: { orderBy: { user: { name: "asc" } } } }   lateral ABSENT
include: { accounts: { orderBy: { id: "asc" } } }               lateral PRESENT
```

Asserting **both** halves is what makes it worth anything — the first line alone is satisfied by a strategy that declines everything. Same shape as #115's line 28, and the same right conclusion: a floor rather than dead code, with the pre-empting caller named in a comment so it is recorded instead of re-derived.

Asserting on `argument` rather than message text is the right choice for the reason given, and reading the site off the error's own stack rather than trusting the input to land where you aimed is the part I would keep. Three of your first attempts hitting a different guard is not an embarrassing detail — it is the measurement of how unreliable "I wrote an input for this line" is, and it is why the triage table can be believed.

### The four with no unit-level pin

`0 unreached` is over both suites, and it holds. But the split is worth recording alongside the number, because the unit suite reaches 141 of 145 and the other four are pinned **only** by a suite that needs a database:

```
provenance.ts:187           save.test.ts               "carries no provenance"
Model.ts:209                save.test.ts               "came from Account, not User"
Model.ts:227                save.test.ts               "fetched without its primary key"
nested-writes.ts:1130       writes.differential.test.ts:1344
                            data: { accounts: { delete: { publicId: "acc-theirs" } } }
```

I checked each individually rather than trusting the aggregate, and the last one took the longest — its message appears in no test in either package, so a message-grep says it is unreached and the argument-level assertion is what actually covers it. Which is your own point about `argument` versus message, arriving from the other direction.

All four are runtime guards inside `run()` bodies or `Model.save`, so this is a property of what they are, not an oversight. Worth saying because the thesis of #114 is that unreached refusals are where defects hide, and these four are the ones whose pin breaks first: a template run needs migrations, a generated client, and a working `gemi-orm-generator` — I lost a while to a stale one reaching for exactly this measurement, which is the third time this session that artifact has produced a wrong answer.

If #114 is closing, the number worth carrying forward is the pair — `145 sites, 0 unreached, 4 of them unit-unreachable` — rather than the single figure. It is the one a future regression would show up in first.

PR #120, opened 2026-07-29
The other 16 error classes, which the coverage measurement never looked at

### LINKED ISSUE #119: The refusal-coverage measurement covered 1 of the ORM's 18 error classes
#114 measured `UnsupportedQueryError` and its subclass, and #118 closed it at **0 unreached**. But `errors.ts` declares **18** error classes, and the other 16 were never measured. Same method, wider net.

## Measured

Instrumented all 18 constructors at once — every class ends with `this.name = "..."`, so one substitution reaches them all — and ran both suites.

**188 construction sites across 18 classes. 14 unreached.**

| class | unreached |
| --- | --- |
| `MalformedRelationError` | 6 |
| `UnknownFieldError` | 3 |
| `MissingModelSchemaError` | 2 |
| `ReturningUnsupportedError` | 1 |
| `MissingRequiredValueError` | 1 |
| `ModelNotRegisteredError` | 1 |

## Triaged

All 14 reachable, and correct. Constructing an input for each is the work, and for the `MalformedRelationError` six it means building a **deliberately malformed schema** — which is the point of them: they guard against the Prisma generator emitting something inconsistent, not against caller input, so nothing that goes through the public API can reach them.

| site | reached by |
| --- | --- |
| `correlate.ts:105` | a relation whose `from` names a column that does not exist |
| `plan-relations.ts:736` | neither side of the relation holds the foreign key |
| `plan-relations.ts:765` | the child declares 0, or 2, relations with that `relationName` |
| `plan-relations.ts:860` | a join table connecting two models, neither of them the parent |
| `plan-relations.ts:883` | an implicit m-n whose end has a composite primary key |
| `plan-relations.ts:900` | a primary key naming a field the model does not declare |

Two are worth calling out.

**`ReturningUnsupportedError` (`write.ts:111`) is unreachable by design today.** Both dialects support `RETURNING`; the guard exists so MySQL is a named gap rather than a silent one, which its own comment says. Reached in a test with a stub dialect — the honest way, since inventing a third real dialect to test it would be absurd.

**`ModelNotRegisteredError` (`registry.ts:29`) is pre-empted.** `relatedSchema` asks `registry.has` and raises `UnregisteredRelationTargetError` — which names the parent, the relation *and* the registered models — before `registry.get` can raise the blunter one. Third instance of this shape, after `select.ts:28` (#115) and `parseOrderBy`'s relation guard (#118). Like both, it is a floor rather than dead code.

## Not a defect, but checked

`keyColumn` (`write.ts:1310`) takes `schema.primaryKey[0]` — only the first field of a composite key, which looked wrong. It is not: both callers use it to *count* rows (`returning <col>` on `createMany`/`updateMany`/`deleteMany`, and the constant-false select for an empty list), never to identify one. Any single column serves.

## Acceptance

- Tests for all 14, asserting the class and the diagnostic — for the malformed-schema six, the message is the entire value, since an author hitting one has a broken generated artifact and nothing else to go on.
- The two special cases documented where they are, not only in a test: which caller pre-empts `registry.ts:29`, and that `write.ts:111` waits on a dialect.
- Re-measure to confirm, the way #118 did.

nstfkc:
Closes #119. Stacked on #118.

#114 measured `UnsupportedQueryError` and #118 closed it at 0 unreached. But `errors.ts` declares **18** classes, and the other 16 were never asked. Same method, wider net — every class ends with `this.name = "..."`, so one substitution instruments all of them at once.

**188 construction sites across 18 classes. 14 unreached**, six of them `MalformedRelationError`.

| class | unreached |
| --- | --- |
| `MalformedRelationError` | 6 |
| `UnknownFieldError` | 3 |
| `MissingModelSchemaError` | 2 |
| `ReturningUnsupportedError` | 1 |
| `MissingRequiredValueError` | 1 |
| `ModelNotRegisteredError` | 1 |

## The six are a different animal

Unlike the shape guards in `refusals.test.ts`, these do not guard caller input — they guard the **generated artifact**. Nothing that goes through the public API can reach one, which is why no test did. Testing them means building a deliberately broken schema, and that is also what makes them worth testing: an author who hits one has a stale artifact and the message is the entire diagnosis.

Each is now reached by the specific malformation it describes:

| guard | malformation |
| --- | --- |
| `correlate.ts:105` | a link whose `from` names a column that is not there |
| `plan-relations.ts:736` | neither side holds the foreign key |
| `plan-relations.ts:765` | the child declares 0, or 2, relations with that `relationName` |
| `plan-relations.ts:860` | a join table connecting two models, neither of them the parent |
| `plan-relations.ts:883` | an implicit m-n whose end has a composite primary key |
| `plan-relations.ts:900` | a primary key naming a field the model does not declare |

## Two are not ordinary gaps

**`ReturningUnsupportedError` is unreachable by design.** Both dialects support `RETURNING`; the guard exists so MySQL is a *named* gap rather than a silent one — its own comment says so. A stub dialect is the honest way to reach it, rather than inventing a third real one.

**`fieldOf` is pre-empted**, and is the one site still unreached after this. Its only callers build `upsert`'s `on conflict` target, and two checks get there first with no path between them:

- a `where` selecting on a key the `create` does not set is refused (`write.ts:645`);
- satisfying that means putting the phantom name in `create`, where the data compiler refuses it as unknown.

Both halves are asserted rather than described. `registry.get`'s guard is pre-empted the same way, by `UnregisteredRelationTargetError` — which names the parent, the relation *and* what is registered, strictly more than the blunter one.

That makes **four floors** now: `select.ts:28` (#115), `parseOrderBy`'s relation guard (#118), and these two. Each is a floor rather than dead code, and each is pinned so the distinction is recorded instead of re-derived by the next reader.

## Checked, not a defect

`keyColumn` takes `schema.primaryKey[0]` — only the first field of a composite key, which looked wrong. It is not: both callers use it to **count** rows (`returning <col>` on the counting writes, and the constant-false select for an empty list), never to identify one. Any single column serves. Worth recording that it was checked.

## Verification

Re-measured with the instrumentation: **188 sites, 1 unreached** — the pre-empted `fieldOf`.

- 1153 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/compile/malformed-schema.test.ts
+++ b/packages/gemi/orm/compile/malformed-schema.test.ts
@@ -1,0 +1,352 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { SqliteDialect } from "../dialect/sqlite";
+import {
+  MalformedRelationError,
+  MissingModelSchemaError,
+  MissingRequiredValueError,
+  ReturningUnsupportedError,
+  UnknownFieldError,
+  UnregisteredRelationTargetError,
+} from "../errors";
+import { account, organization, post, tag, user } from "../fixtures";
+import { Model } from "../Model";
+import * as registry from "../registry";
+import { compileRead } from "./read";
+import { compileWrite } from "./write";
+
+const sqlite = new SqliteDialect();
+
+/**
+ * The construction sites nothing reached, across the **other** 16 error classes.
+ *
+ * #114 measured `UnsupportedQueryError` only, and #118 closed it at 0 unreached
+ * — but `errors.ts` declares 18 classes. Instrumenting all of them (every one
+ * ends with `this.name = "..."`, so one substitution reaches them all) gives
+ * **188 sites, 14 unreached** (#119).
+ *
+ * Most of them are `MalformedRelationError`, and those are a different animal
+ * from the shape guards in `refusals.test.ts`. They do not guard caller input:
+ * they guard the **generated artifact**, so nothing that goes through the
+ * public API can reach one. Testing them means building a deliberately broken
+ * schema, which is also what makes them worth testing — an author who hits one
+ * has a stale artifact and the message is the entire diagnosis.
+ */
+describe("a malformed generated artifact", () => {
+  const register = (pairs: [string, unknown][]) => {
+    registry.clearRegistry();
+    for (const [name, schema] of pairs) {
+      registry.register(name, class { static $schema = schema });
+    }
+  };
+
+  beforeEach(() => registry.clearRegistry());
+
+  /** `Account.user.from` naming a column `Account` does not declare. */
+  test("a link naming a column that is not there", () => {
+    const broken = {
+      ...account,
+      relations: {
+        ...account.relations,
+        user: { ...account.relations.user, from: ["nosuchcol"] },
+      },
+    };
+    register([["User", user], ["Account", broken]]);
+
+    expect(() =>
+      compileRead(
+        user,
+        "findMany",
+        { where: { accounts: { some: { id: 1 } } } } as never,
+        sqlite,
+      ),
+    ).toThrow(MalformedRelationError);
+    expect(() =>
+      compileRead(
+        user,
+        "findMany",
+        { where: { accounts: { some: { id: 1 } } } } as never,
+        sqlite,
+      ),
+    ).toThrow(/joins on 'nosuchcol', which is not a field on Account/);
+  });
+
+  /** Both sides with an empty `from` — a link with no key on either end. */
+  test("neither side holding the foreign key", () => {
+    const orphan = {
+      ...account,
+      relations: {
+        ...account.relations,
+        user: { ...account.relations.user, from: [], to: [] },
+      },
+    };
+    register([["User", user], ["Account", orphan]]);
+
+    expect(() =>
+      compileWrite(
+        user,
+        "update",
+        { where: { id: 1 }, data: { accounts: { connect: { id: 1 } } } } as never,
+        sqlite,
+      ),
+    ).toThrow(/neither User.accounts nor Account.user holds the foreign key/);
+  });
+
+  // The opposing relation is found by `relationName`, so both "none" and "more
+  // than one" are the same failure: the planner cannot tell which side it is
+  // looking at.
+  test.each([
+    [
+      "no opposing relation",
+      { ...account, relations: {} },
+      /Account declares no relation named 'AccountToUser'/,
+    ],
+    [
+      "two opposing relations",
+      {
+        ...account,
+        relations: {
+          ...account.relations,
+          user: { ...account.relations.user, from: [], to: [] },
+          user2: {
+            name: "user2",
+            model: "User",
+            kind: "one" as const,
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the other 16 error classes (closes #119)

Read the diff against its true base (`a3584df`, #118) — one file, all tests. Re-ran the measurement with the same substitution trick, which is the right one: every constructor ends in `this.name = "…"`, so one regex instruments all eighteen.

```
error classes in errors.ts               18       (matches)
construction sites, orm/, non-test      188       (matches)
distinct sites reached, unit suite      164  ->  177
newly reached                            13
unit tests                                     1153
```

The 13 newly reached include the six malformation guards **exactly** as tabled — `correlate.ts:105`, `plan-relations.ts:736 / 765 / 860 / 883 / 900` — plus `ReturningUnsupportedError` at `write.ts:111` and `registry.ts:29`. And `write.ts:1318` is `fieldOf`, the one still unreached, so the accounting lands where the description says it does.

**Guarding the artifact rather than the caller is a real distinction and I had not seen it drawn here before.** It also explains the shape of the whole #114 series in retrospect: the shape guards needed a malformed *argument*, these need a malformed *schema*, and the reason nothing reached them is that no test had a reason to build one. Reaching each by the specific malformation it describes — rather than one broken fixture tripping whichever guard fires first — is what makes the six pins independent.

### The `fieldOf` floor survives a harder test than the one you gave it

The pre-emption is argued from caller input: a `where` on a key the `create` does not set is refused, and satisfying that puts the phantom name in `create` where the data compiler refuses it. True — but this PR's own thesis invites the follow-up, since `key` comes from `matchUniqueKey` over the schema's declared uniques, never from the caller. So a phantom name there means a **malformed artifact**, and the obvious question is whether `fieldOf` is a seventh guard of the kind you just pinned six of.

It is not. Breaking the artifact the same way — `uniques: [["ghost"]]` with no `ghost` field — and reading the site off the stack rather than the message:

```
where { ghost }, create without ghost   ->  write.ts:645    (upsert cannot express it)
where { ghost }, create with ghost      ->  write.ts:1019   (unknown field)
                  ... and on both dialects, and with update: {}
fieldOf (write.ts:1318)                 ->  never reached
```

So it holds against the artifact break too, which is a stronger statement than the description makes and the one that actually settles it. Worth putting in the comment beside it — "no caller can reach this" is the weaker claim; "no schema can either, because two checks stand in front of it on every path" is what a future reader needs to not re-derive.

That is four floors now, each pinned rather than described. The pattern is worth keeping.

### The rest

`ReturningUnsupportedError` via a stub dialect is right — inventing a third real dialect to reach a guard whose purpose is to name MySQL as a gap would be testing the wrong thing. The `keyColumn` note checks out: both callers use it to count rows, never to identify one, so `primaryKey[0]` is sound. Recording that it was checked and found fine is more useful than silence, since the next reader will have the same reaction.

Eleven sites remain outside the unit suite on this branch and ten are runtime guards the template suite covers — the same 4-becomes-10 shape as #118, and the same caveat: that half of the number depends on a run with a database and a current generator.

PR #122, opened 2026-07-29
`divide`, the one arithmetic operator the differential never compared

### LINKED ISSUE #121: The differential never compared 'divide', the one arithmetic operator with a dialect-dependent edge
Several defects in this series survived because no differential case covered them — #101's owning-side miss and #105's parent-restriction overwrite both said so explicitly. So rather than keep noticing that after the fact, I measured what the differential harness actually compares.

## Method

Instrumented `expectSame` and `expectSameWrite` — the harness's two entry points — to record the model, operation and argument keys of every comparison, then ran both dialects.

**98 distinct (model, operation, argument-key-set) combinations, across 15 operations and 7 models.**

Per operation, distinct argument-key sets: `findMany` 23, `aggregate` 15, `groupBy` 8, `count` 8, `findFirst` 6, then `upsert`/`update`/`findUnique`/`delete`/`create` 3 each, and the rest 1–2.

## What is covered

Walking every key at every depth rather than only the top level:

- **All 12 `where` operators** — `contains`, `endsWith`, `equals`, `gt`, `gte`, `in`, `lt`, `lte`, `mode`, `not`, `notIn`, `startsWith`.
- All five relation filters (`some`, `every`, `none`, `is`, `isNot`) and all three logical operators (`AND`, `OR`, `NOT`).
- Every top-level read and write argument the ORM implements.

Two Prisma arguments never appear anywhere: `cursor` and `distinct`. **That is correct, not a gap** — both are in `PERMANENTLY_REFUSED` with measured justifications, so there is nothing to compare: gemi refuses where Prisma accepts, by decision.

## The gap

`ARITHMETIC` declares four operators — `increment`, `decrement`, `multiply`, `divide`. The differential table has cases for the first three, on consecutive lines. **`divide` is absent.**

It is also the one of the four with an edge that behaves differently per dialect:

- **Division by zero**: Postgres raises `division_by_zero`; SQLite yields `null` — and `globalRole` is not nullable, so the two dialects fail in different ways.
- **A remainder**: whether an `Int` divided into a non-integer truncates, rounds, or becomes a float is the database's answer, and both clients have to give the same one.

## Measured

Added three cases — plain, with a remainder, and by zero — and ran them.

**They pass on both dialects.** No defect: gemi and Prisma agree, including the by-zero case, on SQLite and Postgres 16 alike. Worth stating plainly rather than implying a find.

The value is that the last uncovered arithmetic operator is now covered, and that the coverage of the differential surface is now a measured fact rather than an assumption. The measurement is the artifact; the gap it found was small.

## Acceptance

- `divide` in the differential table beside its three siblings, with the two edge cases that make it different from them.

nstfkc:
Closes #121. Stacked on #120.

#101 and #105 both survived because no differential case covered them, and both said so in as many words. Rather than keep noticing that after the fact, I measured what the harness actually compares — instrumenting `expectSame` and `expectSameWrite` to record model, operation and argument keys, then running both dialects.

## What the differential reaches

**98 distinct (model, operation, argument-key-set) combinations, across 15 operations and 7 models.**

| operation | distinct arg-key sets |
| --- | --- |
| `findMany` | 23 |
| `aggregate` | 15 |
| `groupBy` / `count` | 8 each |
| `findFirst` | 6 |
| `upsert` / `update` / `findUnique` / `delete` / `create` | 3 each |
| `updateMany` / `findUniqueOrThrow` / `createMany` | 2 each |
| `findFirstOrThrow` / `deleteMany` | 1 each |

Walking every key at every depth rather than only the top level:

- **all 12 `where` operators** — `contains`, `endsWith`, `equals`, `gt`, `gte`, `in`, `lt`, `lte`, `mode`, `not`, `notIn`, `startsWith`;
- all five relation filters (`some`, `every`, `none`, `is`, `isNot`), all three logical operators;
- every top-level read and write argument the ORM implements.

`cursor` and `distinct` never appear. **That is correct, not a gap** — both are in `PERMANENTLY_REFUSED` with measured justifications, so there is nothing to compare: gemi refuses where Prisma accepts, by decision. I checked rather than assuming they were an oversight.

## The gap

`ARITHMETIC` declares four operators. The differential table had cases for `increment`, `decrement` and `multiply` on consecutive lines — and none for `divide`.

It is also the one of the four whose edges are dialect-dependent:

- **division by zero** — Postgres raises `division_by_zero`, SQLite yields `null`, and `globalRole` is not nullable, so the two dialects fail in different ways;
- **a remainder** — whether an `Int` divided into a non-integer truncates, rounds, or becomes a float is the database's answer, not the ORM's.

## Measured

Three cases: plain, with a remainder, and by zero.

**They pass on both dialects.** gemi and Prisma agree throughout, including the by-zero case, on SQLite and Postgres 16 alike. No defect — worth saying plainly rather than implying a find.

What changes is that the last uncovered arithmetic operator is covered, and that the differential's reach is a measured fact rather than an assumption. The measurement is the artifact here; the gap it turned up was small, which is itself the useful result.

## Verification

- Template suite: 595 passed on SQLite, 860 on Postgres 16 (up 3 on each — the new cases).
- `tsc --noEmit` and `oxlint` unchanged and clean.

### DIFF
--- a/templates/saas-starter/app/models/writes.differential.test.ts
+++ b/templates/saas-starter/app/models/writes.differential.test.ts
@@ -198,6 +198,24 @@ const CASES: Case[] = [
   ["update with multiply", "update", {
     where: { id: 3 }, data: { globalRole: { multiply: 3 } },
   }],
+  // `divide` was the one arithmetic operator the differential never compared,
+  // sitting in the table beside three that it did. It is also the one with an
+  // edge that behaves differently per dialect, hence the two cases below it.
+  ["update with divide", "update", {
+    where: { id: 1 }, data: { globalRole: { divide: 2 } },
+  }],
+  // An Int divided into a non-integer: whether the result truncates, rounds or
+  // becomes a float is a database question, and both clients have to answer it
+  // the same way.
+  ["update with divide leaving a remainder", "update", {
+    where: { id: 2 }, data: { globalRole: { divide: 3 } },
+  }],
+  // Division by zero: Postgres raises, SQLite yields null — and `globalRole` is
+  // not nullable, so the two dialects fail differently. Both clients still have
+  // to fail the *same* way as each other on whichever dialect is running.
+  ["update with divide by zero", "update", {
+    where: { id: 1 }, data: { globalRole: { divide: 0 } },
+  }],
   ["update matching nothing", "update", {
     where: { id: 99999 }, data: { name: "x" },
   }],


nstfkc:
## Review — `divide`, the one arithmetic operator the differential never compared (closes #121)

Read the diff against its true base (`ddf7a9c`, #120). Ran the file on SQLite: 137 pass. I did not have a Postgres 16 to hand, so the Postgres half of your numbers is yours, not mine.

**The gap is real and the checks around it hold.** `ARITHMETIC` declares four operators; the base's differential table contains the string `divide` zero times. And the `cursor` / `distinct` absence is correctly *not* a gap — both are in `PERMANENTLY_REFUSED` with the reasoning written out, so there is nothing to compare. Checking that rather than assuming an oversight is the right instinct, and it is the part of the method that generalises.

**I read the two by-zero errors rather than trusting the verdict**, because `failureKind` collapses everything outside P2002/P2025 to `other`, and its own comment says agreement there is the weakest thing the harness can claim:

```
gemi   : NOT NULL constraint failed: User.globalRole
prisma : Null constraint violation on the fields: (`globalRole`)
```

Genuinely the same failure, so the case is honest agreement rather than two unrelated errors both landing in `other`. Worth knowing, since that bucket is where a vacuous pass would hide.

### The plain `divide` case cannot fail

`globalRole` seeds as `0` for user 1, `1` for user 2, `2` for user 3. So:

```
increment 2  on id 1    0 -> 2
decrement 1  on id 2    1 -> 0
multiply  3  on id 3    2 -> 6
divide    2  on id 1    0 -> 0      <- starting value
```

It divides **zero**. The expected result is the value the row already had, so the case cannot distinguish a working `divide` from `multiply`, from a no-op, or from `set: 0`. Verified by breaking it — compiling `divide` as `*`:

```
update with divide                     PASSES
update with divide leaving a remainder FAILS
update with divide by zero             FAILS
```

Your other two cases catch it, so the operator is covered and this is not a hole in the suite. But the plain case is the one that names the ordinary use, and right now the two edge cases are carrying it — which is the reverse of how the table reads.

One character fixes it. On `id: 3` the plain case is `2 / 2 = 1`, distinct from the starting value and from every other operator's result on that row:

```
plain case on id 3, divide compiled as `*`     all three FAIL
plain case on id 3, divide compiled as `/`     137 passed
```

Both measured. That also makes the three cases a real progression — a clean division, a truncating one, and the dialect-dependent edge — rather than two edges and a no-op.

The `0 / 0` in the by-zero case is fine either way: SQLite nulls it and Postgres raises, and both are the intended failure. Worth leaving on `id: 1` deliberately if you want the zero numerator exercised somewhere, now that the plain case no longer does it by accident.

nstfkc:
## Refinement — what the plain `divide` case is actually failing to do

Re-checked my comment above at full-suite scope while auditing my own measurements on this stack. The finding stands but my framing of its cost was off.

I wrote that *"the two edge cases are carrying the operator while the case that names the ordinary use is inert"*. The operator is not carried by them — it is already pinned by the unit suite. Compiling `divide` as `*`:

```
orm suite    2 failed
  × divide reads the column and writes it back      compile/write.test.ts
  × every shape in the table produces its own plan  plan.writes.discrimination.test.ts
```

So a broken `/` does not reach anyone through this table either way.

**What the plain case fails to do is its own job.** A differential case exists to compare gemi against Prisma, and `globalRole` seeds as `0` for user 1, so `0 / 2 = 0` on both sides — agreement that would survive the operator being anything at all, including one Prisma and gemi disagreed about. The two edge cases do compare something; this one compares zero against zero.

That is a smaller claim than "the operator is uncovered", and it is the accurate one. The fix is unchanged and still worth it — `id: 3` makes it `2 / 2 = 1`, a value distinct from the starting value and from every other operator's result on that row, so the case compares what it was written to compare. Verified both ways: head 137 pass, and with `divide` compiled as `*` all three cases fail rather than two.

PR #124, opened 2026-07-29
The plan cache's invariants, as properties rather than cases

### LINKED ISSUE #123: The plan cache's two invariants are tested case-by-case, never as properties
The plan cache has produced two defects, in opposite directions:

- **#92** — a `having` threshold reached the key, so 200 thresholds made 200 entries. A value leaking into the key is unbounded growth.
- **#100** — `$compositeIn` was reachable by callers. Two different queries sharing a key means one runs the other's statement.

Each was fixed with a test for *that case*. Neither left anything asserting the **property**, so the next instance arrives the same way — silently, and found by hand.

## The invariants

1. **Same key ⇒ same SQL.** No two argument shapes share an entry unless they compile identically. The failure is severe and silent.
2. **A bound value never reaches the key.** Varying every parameter leaves the key unchanged.
3. **...but a literal position does**, because it decides the statement. `LITERAL_KEYS` documents this at length for `omit`: `{ password: true }` and `{ password: false }` both shape to `boolean`, so sharing an entry would hand the public endpoint the column the admin screen kept.

## Measured

Built a corpus of ~115 argument shapes — every `where` operator, the relation filters, the logical operators, `select`/`omit`/`include`/`_count`, ordering forms, pagination, the write operands, aggregates and groups — and checked all three on both dialects.

**118 distinct keys, 0 collisions, 0 leaks.** The cache is sound for this corpus; the point is that nothing was asserting it.

## One correction to my own assumption

I initially expected `take` and `skip` to be literal, since they change which rows come back. They are not, and should not be: they compile to `limit ?` / `offset ?` with the number **bound**, so two page sizes are one plan. A paginated endpoint would otherwise compile a plan per page. Now asserted in that direction, with the present-versus-absent case still differing.

## The test was vacuous on its first run

Worth recording, because it is the third time in this series.

The corpus's two `having` entries named a column that was not in `by`, which is refused — so they never compiled, were silently dropped by a `try/catch`, and the #92 case they existed to cover tested nothing. Reintroducing #92 into `plan.ts` did **not** fail the test.

Making the drops loud immediately surfaced three more entries that never compiled: `mode: "insensitive"` on SQLite, a per-parent `take` inside a to-many under the batched strategy, and an `upsert` whose `create` did not set the conflict target.

So the corpus needs a fourth assertion of its own: **every entry compiles, or is refused for a named reason**. Two are real dialect and strategy limits and are listed as such; a listed reason that stops firing fails the test too, so the list cannot go stale.

## Acceptance

- The three invariants as corpus-driven tests on both dialects.
- The corpus-compiles check, with named exceptions rather than silent drops.
- Verified by reintroducing both historical bugs and confirming each fails the relevant invariant.

nstfkc:
Closes #123. Stacked on #122.

The plan cache has produced two defects, in opposite directions:

- **#92** — a `having` threshold reached the key, so 200 thresholds made 200 entries. Unbounded growth.
- **#100** — `$compositeIn` was reachable by callers. Two queries sharing a key means one runs the other's statement.

Each was fixed with a test for *that case*. Neither left anything asserting the **property**, so the next instance arrives the same way — silently, and found by hand.

## Three invariants

Over a corpus of ~115 argument shapes — every `where` operator, the relation filters, the logical operators, `select`/`omit`/`include`/`_count`, the ordering forms, pagination, the write operands, aggregates and groups:

1. **Same key ⇒ same SQL.** No two shapes share an entry unless they compile identically.
2. **A bound value never reaches the key.** Varying every parameter leaves it unchanged.
3. **...but a literal position does.** `omit: { password: true }` and `{ password: false }` both shape to `boolean`; sharing an entry would hand the public endpoint the column the admin screen kept — the hazard `LITERAL_KEYS` documents at length.

**118 distinct keys, 0 collisions, 0 leaks**, on both dialects.

## A correction to my own assumption

I expected `take` and `skip` to be literal, since they change which rows come back. They are not, and should not be — they compile to `limit ?` / `offset ?` with the number **bound**, so two page sizes are one plan. A paginated endpoint would otherwise compile a plan per page. Now asserted in that direction, with present-versus-absent still differing.

## The test was vacuous on its first run

Worth stating plainly, because it is the third time in this series after #96 and #106.

The corpus's two `having` entries named a column that was not in `by` — which is refused — so they never compiled, were silently dropped by a `try/catch`, and the #92 case they existed to cover tested nothing. **Reintroducing #92 into `plan.ts` did not fail the test.**

Making the drops loud immediately surfaced three more entries that never compiled:

| entry | why |
| --- | --- |
| `mode: "insensitive"` | SQLite has no case-insensitive LIKE for non-ASCII; Prisma refuses it there too |
| `include: { accounts: { take: 2 } }` | a per-parent `take` needs the lateral strategy; these compile with batched |
| `upsert` whose `create` omits the conflict target | `Model.upsert` runs that as a read and a write instead |

The third was a corpus error and is fixed. The first two are real limits, so there is now a **fourth test**: every corpus entry compiles, or is refused for a *named* reason. A listed reason that stops firing fails too, so the list cannot go stale as the limits lift.

## Verification

Both historical bugs reintroduced and confirmed caught:

- `having` no longer collapsing → **invariant 2 fails** on both dialects;
- `omit` dropped from `LITERAL_KEYS` → **invariant 3 fails** on both dialects.

- 1163 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/plan-key.invariants.test.ts
+++ b/packages/gemi/orm/plan-key.invariants.test.ts
@@ -1,0 +1,385 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { compileAggregate } from "./compile/aggregate";
+import { compileGroupBy } from "./compile/group-by";
+import { compileRead } from "./compile/read";
+import { compileWrite } from "./compile/write";
+import { PostgresDialect } from "./dialect/postgres";
+import { SqliteDialect } from "./dialect/sqlite";
+import {
+  account,
+  membership,
+  organization,
+  post,
+  profile,
+  tag,
+  userWithProfile,
+} from "./fixtures";
+import { planKey } from "./plan";
+import * as registry from "./registry";
+
+const sqlite = new SqliteDialect();
+const postgres = new PostgresDialect();
+const DIALECTS = [
+  ["sqlite", sqlite],
+  ["postgres", postgres],
+] as const;
+
+/**
+ * The plan cache's two invariants, asserted over a corpus rather than case by
+ * case.
+ *
+ * Both have been broken before, in opposite directions:
+ *
+ *   - **#92** — a `having` threshold reached the key, so 200 thresholds made
+ *     200 entries. A value leaking into the key is unbounded growth.
+ *   - **#100** — `$compositeIn` was reachable by callers, which is the other
+ *     direction: two different queries sharing a key means one runs the
+ *     other's statement.
+ *
+ * Each was fixed with a test for that case. Neither left anything asserting the
+ * *property*, so the next instance would arrive the same way. These do:
+ *
+ *   1. same key ⇒ same SQL — no two shapes share an entry unless they compile
+ *      identically;
+ *   2. a bound value never reaches the key — varying every parameter leaves it
+ *      unchanged;
+ *   3. ...but a *literal* position does, because it decides the statement.
+ *
+ * The corpus is the thing to extend when a new argument is implemented. It is
+ * deliberately broad rather than minimal: the cost of a case is one line, and
+ * the failure mode being guarded is silent.
+ */
+const READS: unknown[] = [
+  {},
+  { where: { id: 1 } },
+  { where: { id: 2 } },
+  { where: { name: "a" } },
+  { where: { name: { contains: "a" } } },
+  { where: { name: { startsWith: "a" } } },
+  { where: { name: { endsWith: "a" } } },
+  { where: { name: { equals: "a" } } },
+  { where: { name: { not: "a" } } },
+  { where: { id: { gt: 1 } } },
+  { where: { id: { gte: 1 } } },
+  { where: { id: { lt: 1 } } },
+  { where: { id: { lte: 1 } } },
+  { where: { id: { in: [1, 2] } } },
+  { where: { id: { in: [1, 2, 3] } } },
+  { where: { id: { notIn: [1] } } },
+  { where: { name: { contains: "a", mode: "insensitive" } } },
+  { where: { AND: [{ id: 1 }, { name: "a" }] } },
+  { where: { OR: [{ id: 1 }, { name: "a" }] } },
+  { where: { NOT: { id: 1 } } },
+  { where: { accounts: { some: { id: 1 } } } },
+  { where: { accounts: { every: { id: 1 } } } },
+  { where: { accounts: { none: { id: 1 } } } },
+  { where: { organization: { is: { id: 1 } } } },
+  { where: { organization: { isNot: { id: 1 } } } },
+  { select: { id: true } },
+  { select: { id: true, name: true } },
+  { omit: { name: true } },
+  { include: { accounts: true } },
+  { include: { organization: true } },
+  { include: { accounts: { select: { id: true } } } },
+  { include: { accounts: { where: { id: 1 } } } },
+  { include: { accounts: { take: 2 } } },
+  { include: { accounts: { orderBy: { id: "asc" } } } },
+  { include: { accounts: { orderBy: { id: "desc" } } } },
+  { include: { _count: { select: { accounts: true } } } },
+  { orderBy: { id: "asc" } },
+  { orderBy: { id: "desc" } },
+  { orderBy: { name: { sort: "asc", nulls: "first" } } },
+  { orderBy: { name: { sort: "asc", nulls: "last" } } },
+  { orderBy: [{ id: "asc" }, { name: "desc" }] },
+  { orderBy: { accounts: { _count: "asc" } } },
+  { take: 1 },
+  { take: 5 },
+  { skip: 1 },
+  { skip: 1, take: 2 },
+];
+
+const WRITES: [string, unknown][] = [
+  ["create", { data: { email: "a@b.c" } }],
+  ["create", { data: { email: "a@b.c", name: "n" } }],
+  ["create", { data: { email: "a@b.c" }, select: { id: true } }],
+  [
+    "create",
+    { data: { email: "a@b.c", accounts: { create: { provider: "p", providerAccountId: "x" } } } },
+  ],
+  ["create", { data: { email: "a@b.c", organization: { connect: { id: 1 } } } }],
+  ["createMany", { data: [{ email: "a@b.c" }] }],
+  ["createMany", { data: [{ email: "a@b.c" }, { email: "d@e.f" }] }],
+  ["update", { where: { id: 1 }, data: { name: "x" } }],
+  ["update", { where: { id: 1 }, data: { globalRole: { increment: 1 } } }],
+  ["update", { where: { id: 1 }, data: { globalRole: { decrement: 1 } } }],
+  ["update", { where: { id: 1 }, data: { globalRole: { multiply: 2 } } }],
+  ["update", { where: { id: 1 }, data: { globalRole: { divide: 2 } } }],
+  ["update", { where: { id: 1 }, data: { globalRole: { set: 2 } } }],
+  ["update", { where: { id: 1 }, data: { accounts: { connect: { id: 1 } } } }],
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the plan cache's invariants as properties (closes #123)

Read the diff against its true base (`0e32db6`, #122). 1163 tests pass.

**Properties over cases is the right move here**, and the corpus-plus-`vary` shape is the cheap version of it — one line per case, and the thing being guarded is silent. The fourth test is the part I would keep hardest: *"or is refused for a **named** reason"*, with each named reason asserted to still fire, is what stops the corpus rotting. Loud drops are the whole difference between this and the version that was vacuous.

Confirmed both stated verifications by re-breaking the code:

```
omit removed from LITERAL_KEYS             invariant 3 fails, both dialects
having subtree made literal                invariant 2 fails, both dialects
```

The `take`/`skip` correction is right and worth having written down — I would have assumed the same thing, and asserting present-versus-absent alongside it is what keeps the assertion from being satisfied by a key that ignores pagination entirely.

### Reintroducing #92 itself does not fail the test

The `having` break above is the coarse form. The narrow one — neutralising the `inHaving` flag, which is the mechanism #92 actually added — leaves **all 10 green**:

```
if (key === "having") return canonicalShape(value, false, collapseLists, false);
                                                                        ^^^^^
  -> plan-key.invariants.test.ts   10 passed
```

It is a real reintroduction; the keys move exactly where #92 lives:

```
having { globalRole: { _count: { gt: 1 } } } vs { gt: 200 }
  head            same key = true
  #92 reintroduced same key = false      <- 200 thresholds, 200 entries
```

Two things hide it, and the second is the interesting one:

1. **The corpus's only `having` entries compare a plain column** — `having: { globalRole: { gt: N } }` — whose key collapses with or without the flag. The aggregate form the flag exists for is not in the corpus. (`having: { _count: { globalRole: … } }` is refused — `'_count' is not a field on model User` — so the spelling that compiles is `{ globalRole: { _count: … } }`.)

2. **Adding it is not enough**, because `vary`'s own `LITERAL` set lists `_count`, `_avg`, `_sum`, `_min`, `_max` unconditionally. That is precisely the conflation #92 removed from `plan.ts` — an aggregate kind is a projection in a `select` and a comparison in a `having` — so `vary` returns the threshold untouched and the leak is invisible to the very test written to catch it. The helper reproduces the bug it guards.

Both fixed together, measured in both directions:

```js
// corpus
{ by: ["globalRole"], _count: true, having: { globalRole: { _count: { gt: 1 } } } },
{ by: ["globalRole"], _count: true, having: { globalRole: { _count: { gt: 200 } } } },

// vary(), mirroring plan.ts's own distinction
const literal = LITERAL.has(key) && !(inHaving && AGG.has(key));
```
```
head                10 passed
#92 reintroduced    invariant 2 fails, both dialects
```

Worth doing narrowly rather than by dropping the aggregate kinds from `LITERAL` altogether — `mode` sits in that set too, and varying it to `"ZZZZ"` changes the statement, so a blanket relaxation would report a false leak.

This does not undermine the PR: the invariants are the right ones and the other three breaks are caught. It is that #92 is one of the two defects named as the motivation, and the corpus does not currently reach the shape it lived in.

nstfkc:
## Follow-up — six of `LITERAL_KEYS`' thirteen members can be deleted with the suite green

Went back over this one at the top of the stack (`fe9679b`, #134) and measured each member of `LITERAL_KEYS` individually — delete it, run all 1186 tests:

```
orderBy         9 failed        _count           1 failed
select          3 failed        skipDuplicates   1 failed
omit           10 failed        by               2 failed
include         2 failed
                                distinct      1186 passed
                                mode          1186 passed
                                _avg          1186 passed
                                _sum          1186 passed
                                _min          1186 passed
                                _max          1186 passed
```

`distinct` is fine — it is in `PERMANENTLY_REFUSED`, so no corpus entry can compile and there is nothing to collide. The other five are gaps.

### `mode` is the sharp one, and its pair is the reason

It is *in* the invariant-3 list, so it looks covered:

```ts
["mode", { where: { name: { contains: "a" } } },
         { where: { name: { contains: "a", mode: "insensitive" } } }],
```

Those two differ by the **presence** of a key, which distinguishes them whether or not `mode` is literal. The pair passes with `mode` deleted from `LITERAL_KEYS`. The pair that would not is same key, different value — and that case is a live invariant-1 violation, measured on Postgres:

```
{ contains: "a", mode: "insensitive" }   vs   { contains: "a", mode: "default" }
  head            same key = false
  mode not literal same key = true      same SQL = false
                                        A: where "name" ilike $1
                                        B: where "name"  like $1
```

Two shapes sharing an entry while compiling differently — the failure the first property exists to catch, passing it. A case-insensitive search runs the case-sensitive statement, or the reverse, depending on which compiled first.

### The four aggregate kinds are the same shape

`_avg`, `_sum`, `_min` and `_max` have no corpus entry that switches a field off, so nothing distinguishes them from `_count`, which does. (This is the note I have been carrying as untracked since #78 — it belongs here rather than on its own, because this is the test that now owns the property.)

### Five corpus lines close all five

```js
// READS
{ where: { name: { contains: "a", mode: "default" } } },

// AGGREGATES — same keys, one switched off, which is what makes it a collision
{ _sum: { globalRole: true, id: false } },
{ _avg: { globalRole: true, id: false } },
{ _min: { globalRole: true, id: false } },
{ _max: { globalRole: true, id: false } },
```
```
head             10 passed        (all five, no false positives)
mode deleted     invariant 1 FAILS, postgres   (sqlite refuses `mode` — correct)
_sum deleted     invariant 1 FAILS, both
_avg deleted     invariant 1 FAILS, both
_min deleted     invariant 1 FAILS, both
_max deleted     invariant 1 FAILS, both
```

My first attempt at the aggregate entries used `{ globalRole: true, id: true }` and caught nothing — different key sets, so the shapes differ regardless. Exactly the mistake the `mode` pair makes, which is worth recording: for a `LITERAL_KEYS` member the distinguishing case is always **same keys, different values**, because a key that is merely present or absent is already structural.

That, plus the `having` gap in my earlier comment, makes a mechanical check worth having: every member of `LITERAL_KEYS` should have a corpus pair that collides when it is removed. It is the same "each exemption still needs its exemption" discipline #127 applied to the allow-lists, and it would have found all six of these without anyone thinking of the cases.

nstfkc:
## Retraction — the `having` half of my review above is wrong

I claimed that reintroducing #92 — neutralising the `inHaving` flag — leaves this branch green, and built a suggested fix on it. **That is not true, and the error is mine.** I ran the mutation against `orm/plan-key.invariants.test.ts` only rather than the suite. Re-run properly on this same head:

```
inHaving neutralised, full orm suite    1 failed
  × a having threshold        plan.discrimination.test.ts
```

`plan.discrimination.test.ts` already carries exactly the case I proposed adding:

```js
having: { globalRole: { _count: { gt: 1 } } }   vs   { gt: 9999 }
```

So the corpus does reach the shape #92 lived in — through a different file, which is why a scoped run could not see it. No corpus line and no `vary` change is needed for that, and the `vary` `LITERAL` observation, while accurate about `vary`, is not load-bearing: the behaviour is pinned elsewhere.

This is the third time in this series that a boundary has been the thing that was wrong, and this time it is mine — the same error #137 corrected in its own method, and the same one I flagged on #147. Measuring inside a scope that excludes the answer produces a confident wrong number, whichever side of the review it comes from.

**What survives** is the follow-up comment below it, which was measured against the full suite and still reproduces on the current top of stack (`ac5bd52`):

```
mode removed from LITERAL_KEYS       1287 passed
_avg / _sum / _min / _max removed    1287 passed each
```

`mode` is the one worth acting on, for the reason given there — its invariant-3 pair is distinguished by key *presence*, so it passes with `mode` non-literal, while `mode: "insensitive"` against `mode: "default"` is a genuine invariant-1 collision (`ilike` versus `like` sharing an entry). That half stands.

PR #126, opened 2026-07-29
The injection invariant, tested without a heuristic

### LINKED ISSUE #125: The 'every value is bound' invariant is tested by a heuristic that cannot catch a spliced string
Invariant 2 is the whole of the ORM's injection argument. `compile/fragment.ts` states it: safety does not come from a tagged template, because an ORM query's *shape* is dynamic. It comes from two rules — identifiers only ever come from the generated schema, and **every value is a bound parameter**.

There is a test for the second rule, `no value reaches the SQL text` in `write.test.ts`. It covers seven write statements and works like this:

```ts
const withoutIdentifiers = statement.replace(/"[^"]*"/g, "");
expect(withoutIdentifiers).not.toMatch(/\d/);
```

Strip double-quoted spans (identifiers), then assert no digit survives.

## The weakness

**It cannot catch a spliced string, by construction.** SQL string literals are single-quoted, so they are not stripped — and a spliced value only fails the assertion if it happens to contain a digit. `'a@b.c'`, `'; drop table "User"; --'`, `'%_wildcards%'` all pass.

It also covers writes only. Reads are where a `where` clause assembles operator fragments — `contains`, `startsWith`, `in`, the relation filters, the logical combinators — which is the larger surface and the one an untrusted value actually reaches.

## A property instead

Compile the same *shape* twice with different values; the text must be **byte-identical**. That is type-agnostic and needs no heuristic — anything reaching the text shows up as a difference, whatever it is.

`vary()` from the plan-key corpus (#123) already does exactly the required mutation: replace every scalar that is bound as a parameter, leave structure alone. So the corpus is reusable, and should be shared rather than duplicated.

## Two vacuity traps this has to avoid

Both are live, and one of them already bit in #123.

1. **A compiler that dropped every value would pass.** Two shapes compile alike because neither carries anything. So the values have to be asserted *present in the binding* as well.
2. **A compiler that escaped rather than bound would pass** the "no hostile value in the text" check while corrupting data. So the hostile value has to be asserted to survive binding **unchanged**.

## Measured

Zero leaks: 100 statements built from five hostile values — a quote-and-comment, an identifier break, LIKE metacharacters, a backslash, and a non-ASCII character — across both dialects. No defect.

One dialect difference worth recording: `in` binds as one placeholder per member on SQLite and as a **single array parameter** on Postgres (`{"11","22","33"}`), so an assertion on the binding has to be shape-agnostic.

## Acceptance

- The corpus extracted so both invariant suites share one, since a corpus that drifts between two copies is worse than a short one.
- The byte-identical property over the whole corpus, on both dialects.
- Both vacuity guards above.
- Verified by splicing a value into `where.ts`'s equality and confirming the properties fail.

nstfkc:
Closes #125. Stacked on #124.

Invariant 2 is the whole of the ORM's injection argument — `compile/fragment.ts` states it: safety does not come from a tagged template, because an ORM query's shape is dynamic. It comes from identifiers only ever coming from the generated schema, and **every value being a bound parameter**.

The existing test for the second rule covers seven write statements and works by heuristic:

```ts
const withoutIdentifiers = statement.replace(/"[^"]*"/g, "");
expect(withoutIdentifiers).not.toMatch(/\d/);
```

## Why that is weak

**It cannot catch a spliced string, by construction.** SQL string literals are single-quoted, so they survive the strip — and a spliced value only fails if it happens to contain a digit. `'a@b.c'`, `'; drop table "User"; --'` and `'%_wildcards%'` all pass it.

It also covers writes only, while `where` — where operator fragments, relation filters and logical combinators are assembled — is both the larger surface and the one an untrusted value actually reaches.

## A property instead

Compile the same **shape** twice with different values; the text must be **byte-identical**. Type-agnostic, no heuristic: anything reaching the text is a difference, whatever its type.

`vary()` from #123 already performs exactly that mutation, so the corpus is extracted to `corpus.ts` and shared by both invariant suites rather than duplicated — a corpus that drifts between two copies is worse than a short one.

## Two vacuity guards

The property alone is satisfiable by a broken compiler in two different ways, and #123 already showed this series is prone to exactly that:

| a compiler that... | passes the property because... | guarded by |
| --- | --- | --- |
| drops every value | two shapes compile alike, carrying nothing | values asserted **present in the binding** |
| escapes instead of binding | the hostile value is absent from the text | hostile value asserted to **survive binding unchanged** |

## Measured

**Zero leaks.** 100 statements built from five hostile values — a quote-and-comment, an identifier break, LIKE metacharacters, a backslash, a non-ASCII character — across both dialects. No defect; the invariant holds.

One dialect difference worth recording: `in` binds one placeholder per member on SQLite and a **single array parameter** on Postgres (`{"11","22","33"}`). My first binding assertion was not shape-agnostic and failed on Postgres for that reason rather than a real one.

## Verification

Spliced the value into `where.ts`'s equality and **three of the four properties fail** on both dialects.

An earlier attempt spliced a *constant* rather than the value and tripped only the binding guard — worth noting, because for the byte-identical property to see an injected bug the bug has to vary with the input. That is the right sensitivity here: a constant in the text is not an injection.

- 1172 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).


nstfkc:
## Review — the injection invariant, tested without a heuristic (closes #125)

Read the diff against its true base (`ddeda3c`, #124). 1172 tests pass.

**The critique of the old test is exactly right**, and byte-identical-under-`vary` is the right replacement: type-agnostic, no pattern to be clever about, and it fails on anything that reaches the text. Extending it past writes to `where` matters more than the heuristic swap, since that is where the operator fragments and combinators are assembled.

Confirmed the headline by splicing the value into `where.ts`'s equality —

```ts
if (typeof operand === "string") return sql(`${column} = '${operand}'`);
```
```
the same shape with different values compiles alike   FAILS  both dialects
...and every value is present in the binding          FAILS  both dialects
a hostile value never appears in the statement        FAILS  both dialects
a hostile value is bound unchanged                    passes
```

Three of four, on both dialects, as stated. The note about the earlier attempt splicing a *constant* is worth keeping — "the bug has to vary with the input for a byte-identical property to see it" is the honest limit of the technique, and stating it is better than a reader discovering it.

**The first vacuity guard holds.** I built the compiler from row one — binds a constant, drops the value — and *"every value is present in the binding"* catches it on both dialects. Nothing else does, so the guard is load-bearing.

### The second vacuity guard does not

Row two says an escaping compiler is caught by *"hostile value survives binding unchanged"*. I built exactly that — quote-escaping inside `where.ts`'s `encoded`, the classic sanitise-instead-of-parameterise:

```ts
return (args) => {
  const v = dialect.encode(locate(args), field);
  return typeof v === "string" ? v.replace(/'/g, "''") : v;
};
```
```
binding.invariants.test.ts    8 passed
```

All four properties pass. The corruption is real — every `'` in a `where` value silently doubled — and nothing sees it, because the fourth property tests **one shape**:

```ts
const args = { where: { id: 1 }, data: { name: value } };
compileAny("update", args, dialect);
```

That is a write-`data` value, which never goes through `where.ts` at all. So the guard covers the surface the third property already sweeps eleven shapes of, and does not cover the one this PR opens by arguing is *"both the larger surface and the one an untrusted value actually reaches"*.

Giving it the same shapes as its neighbour closes it, measured both ways:

```ts
["update",    { where: { id: 1 }, data: { name: value } }],
["findMany",  { where: { name: value } }],
["findMany",  { where: { OR: [{ name: value }, { email: value }] } }],
["updateMany",{ where: { name: value }, data: { name: value } }],
```
```
head                8 passed
escaping compiler   a hostile value is bound unchanged FAILS, both dialects
```

Worth keeping to plain equality rather than reusing the third property's list verbatim — `contains` binds `%value%`, so `toContain(value)` would fail on correct code.

### The shared corpus carries #124's blind spot into a second suite

Extracting to `corpus.ts` is right, and one corpus beats two that drift. It does mean the gap I raised on #124 now costs twice: `vary`'s `LITERAL` still lists `_count`/`_avg`/`_sum`/… unconditionally, and the only `having` entries compare a plain column. Reintroducing #92 passes **both** suites:

```
plan-key.invariants + binding.invariants    18 passed
```

For this PR specifically that is not only a plan-key concern — a `having` threshold spliced into the *text* would be invisible to the byte-identical property for the same reason, since `vary` hands the threshold back untouched. The fix from #124 lands in one place now, which is the upside of the extraction.

nstfkc:
## Retraction — the "#92 passes both suites" claim in my review above

My last section said reintroducing #92 leaves both invariant suites green (`18 passed`) and that extracting the corpus therefore carried a blind spot into a second suite. **The measurement was wrong**: I ran the mutation against `orm/plan-key.invariants.test.ts` and `orm/binding.invariants.test.ts` only, not the suite. On this same head:

```
inHaving neutralised, full orm suite    1 failed
  × a having threshold        plan.discrimination.test.ts
```

The shape is covered — by a third file, which a two-file run could not see. So the extraction to `corpus.ts` carries no such gap, and that paragraph should be disregarded. Retracted on #123 as well, where I first made the claim.

Everything above it was measured against the full suite and stands unchanged — the splice reproducing three of four properties, the first vacuity guard being load-bearing, and the second one passing an escaping compiler because the fourth property tests a single write-`data` shape and never a `where`.

nstfkc:
## Correction — the escaping compiler *is* caught, just not by the guard credited with it

Following my retraction above, I re-ran every measurement in that review at full-suite scope rather than per-file. One more needs correcting, and it is the main finding.

I wrote that an escaping compiler — quote-doubling inside `where.ts`'s `encoded` — passes all four properties, and concluded that the second vacuity guard does not hold. Rebuilt from a pristine checkout of this head, with no edits of mine in the tree:

```
escaping compiler, orm/binding.invariants.test.ts    8 passed
escaping compiler, full orm suite                    1 failed
  × never interpolates a value into the text         compile/read.test.ts:743
```

So the sanitise-instead-of-parameterise compiler does not reach production unnoticed. A pre-existing read test catches it.

**What survives is the narrower half**, and it is still worth acting on: the guard the table credits does *not* catch it. `a hostile value is bound unchanged` tests exactly one shape — `{ where: { id: 1 }, data: { name: value } }` — which never goes through `where.ts`, so all eight cases in that file pass under an escaping `where`. The table's row two claims a guard that is not doing that job; something else is.

That changes what the fix is for. Widening the fourth property to `where` shapes is worth doing so the guard performs the role the table assigns it — and I verified it does — but it closes a gap in the *argument*, not in the suite. My "the corruption is real and nothing sees it" was wrong.

Two corrections in two reviews, from one cause: running a mutation against the file I was reading rather than the suite. That is the same error #137 corrected in its own harness and the one I flagged on #147, and it produces a confident wrong number in exactly the direction that makes a review sound more useful than it is. I have re-checked the rest at full-suite scope; the findings on #127, #129, #131 and #139 stand as measured, and #131's is stronger than I showed — with a pristine test file, the two cycle files sit in the generated directory and all six tests pass, including the zero-line diff.

PR #128, opened 2026-07-29
Invariants 1 and 4, enforced rather than stated

### LINKED ISSUE #127: Invariants 1 and 4 are stated in comments and enforced by nothing
The ORM has four stated invariants. Two of them are properties of the *directory* rather than of any function, and neither is enforced:

- **1. `Model.$exec` is the single door to the database.**
- **4. The dialect seam — the compiler asks the dialect, it does not test it.** `plan.ts` puts it plainly: asking the dialect is what keeps `if (dialect === "postgres")` out of the compiler.

`runtime-isolation.test.ts` already establishes the pattern for this — it asserts that no runtime file references `@prisma/*`, precisely because "a single stray import while adding an operation would quietly re-introduce the boundary the project exists to remove." The same argument applies to both of these, and neither has the test.

## Measured

**Invariant 1** holds. `.unsafe(` appears at exactly two sites, both in `Model.ts`:

- `$exec`'s own execution;
- the executor it hands to nested-write steps — `query: (text, values) => conn.unsafe(text, values)`.

`nested-writes.ts` builds the implicit-m-n pair statements, which are the one query with no model behind them, and **never runs them**: `runPairStatement` calls `executor.query`, so they join the caller's transaction like everything else. `context.ts` touches `.begin(` for transaction plumbing, which is not query execution.

**Invariant 4** holds, with two exceptions. `dialect.name` is *compared* in exactly two files, and both choose a **strategy** rather than emit SQL:

| file | comparison |
| --- | --- |
| `compile/strategy.ts` | lateral on Postgres, batched elsewhere |
| `compile/lateral.ts` | the strategy declining a node the dialect cannot express |

Everywhere else `dialect.name` only *appears* — inside a refusal's message (`fragment.ts`, `where.ts`, `write.ts`) or in the plan key (`plan.ts`), which must differ per dialect. Neither decides what SQL comes out.

## Why an allow-list rather than a prohibition

Both invariants have real exceptions, so a bare `grep` for the pattern would either fail today or be written loosely enough to be useless. A list makes each exception a decision someone wrote down, with the reason next to it.

It needs a second half to stay honest: **each entry must still contain the thing it excuses**. An exception that stops being needed should be deleted, not left standing as permission for the next person.

## Acceptance

- Both invariants as directory-wide tests, in the style of `runtime-isolation.test.ts`.
- Allow-lists with the reason for each entry, plus the staleness check above.
- Verified by introducing each violation and confirming the relevant test fails.

nstfkc:
Closes #127. Stacked on #126.

The ORM has four stated invariants. Two are properties of the **directory** rather than of any function, and neither was enforced:

- **1. `Model.$exec` is the single door to the database.**
- **4. The dialect seam** — `plan.ts` puts it plainly: asking the dialect is what keeps `if (dialect === "postgres")` out of the compiler.

`runtime-isolation.test.ts` already establishes both the pattern and the argument — it asserts no runtime file references `@prisma/*`, because "a single stray import while adding an operation would quietly re-introduce the boundary the project exists to remove." That reasoning applies unchanged here, and neither invariant had the test.

## Both hold today

**One door.** `.unsafe(` appears at exactly two sites, both in `Model.ts`: `$exec`'s own execution, and the executor it hands to nested-write steps.

The implicit many-to-many pair statements are the one query with no model behind them — and `nested-writes.ts` builds them without ever running them. `runPairStatement` calls `executor.query`, so they join the caller's transaction like everything else rather than getting their own door. `context.ts` touches `.begin(` for transaction plumbing, which is not query execution.

**The seam.** `dialect.name` is *compared* in exactly two files, and both choose a **strategy** rather than emit SQL:

| file | comparison |
| --- | --- |
| `compile/strategy.ts` | lateral on Postgres, batched elsewhere |
| `compile/lateral.ts` | the strategy declining a node the dialect cannot express |

Everywhere else the name only *appears* — in a refusal's message (`fragment.ts`, `where.ts`, `write.ts`) or in the plan key (`plan.ts`), which must differ per dialect. Neither decides what SQL comes out, so the check is for a *comparison*, not a mention: `=== dialect.name`, a `switch` on it, or an `instanceof` on a concrete dialect class.

## Allow-list, not prohibition

Both invariants have real exceptions, so a bare `grep` would either fail today or be written loosely enough to be useless. A list makes each exception a decision someone wrote down, with the reason beside it.

That needs a second half to stay honest, so there is one: **each entry must still contain the thing it excuses**. An exception that stops being needed gets deleted rather than left standing as permission for the next person.

## Verification

Each violation introduced into `where.ts` and confirmed caught:

- a second `.unsafe(` caller → **"only Model.ts reaches the driver"** fails;
- a `dialect.name === "postgres"` comparison → **"only the strategy chooser tests the dialect's name"** fails.

- 1179 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/architecture.test.ts
+++ b/packages/gemi/orm/architecture.test.ts
@@ -1,0 +1,131 @@
+import { readFileSync, readdirSync } from "node:fs";
+import { join, relative } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * Two of the ORM's stated invariants, asserted as properties of the directory
+ * rather than of any one file — the same shape as `runtime-isolation.test.ts`,
+ * and for the same reason: a single line added while implementing an operation
+ * is how either of them would go, and no unit test would notice.
+ *
+ *   **1. `Model.$exec` is the single door to the database.**
+ *   **4. The dialect seam — the compiler asks the dialect, it does not test it.**
+ *
+ * Both are enforced here by allow-list rather than by prohibition, because both
+ * have real exceptions. A list makes an exception a decision someone wrote
+ * down; a bare `grep` for the pattern would either fail today or be useless.
+ */
+function sourceFiles(dir: string): string[] {
+  const out: string[] = [];
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const path = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      out.push(...sourceFiles(path));
+    } else if (entry.name.endsWith(".ts") && !entry.name.endsWith(".test.ts")) {
+      // Tests are not runtime code, and this one has to name both patterns in
+      // order to assert against them.
+      out.push(path);
+    }
+  }
+  return out;
+}
+
+const ROOT = import.meta.dirname;
+const FILES = sourceFiles(ROOT).map((path) => [relative(ROOT, path), path] as const);
+const read = (path: string) => readFileSync(path, "utf8");
+
+describe("the ORM's seams", () => {
+  test("the scan covers the whole directory", () => {
+    expect(FILES.length).toBeGreaterThan(20);
+  });
+
+  /**
+   * **One door to the driver.**
+   *
+   * `conn.unsafe` is where a statement stops being data and starts being a
+   * query. Everything the ORM does around a query — the plan cache, the policy
+   * layer, the ambient transaction, the error translation, the slow-transaction
+   * timer — hangs off `$exec` calling it. A second caller elsewhere would get
+   * none of that, and would look completely ordinary in review.
+   *
+   * The implicit many-to-many join table is the one statement with no model
+   * behind it, so it cannot be compiled through a model's plan. It does *not*
+   * get its own door: `$exec` hands the nested-write steps an executor closed
+   * over the connection it already resolved, so the pair statements join the
+   * caller's transaction like everything else. `nested-writes.ts` builds that
+   * SQL and never runs it.
+   */
+  const DRIVER_DOOR = ["Model.ts"];
+
+  test("only Model.ts reaches the driver", () => {
+    const callers = FILES.filter(([, path]) => /\.unsafe\(/.test(read(path))).map(
+      ([name]) => name,
+    );
+
+    expect(callers.sort()).toEqual(DRIVER_DOOR);
+  });
+
+  test("...and the join-table statements go through the executor, not the pool", () => {
+    const nested = read(join(ROOT, "compile/nested-writes.ts"));
+
+    // It builds pair SQL...
+    expect(nested).toMatch(/delete from \$\{quoted\(join\.table\)\}|insert into \$\{quoted\(join\.table\)\}/);
+    // ...and hands it to the executor rather than running it.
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — invariants 1 and 4, enforced rather than stated (closes #127)

Read the diff against its true base (`b91eaa5`, #126). 1179 tests pass. Both stated violations reproduce:

```
a second `.unsafe(` in where.ts            "only Model.ts reaches the driver" fails
dialect.name === "postgres" in where.ts    "only the strategy chooser tests..." fails
```

**Allow-list plus a staleness check is the right shape**, and the second half is what makes it more than a snapshot — an exemption that stops being needed gets deleted rather than left standing as permission. Following `runtime-isolation.test.ts` rather than inventing a form is right too; these are the same kind of claim.

### The seam check tests for comparison syntax, not for deciding on the dialect

The invariant is that the compiler *asks* the dialect rather than *recognising* it. The regex looks for the recognising **syntax**. Five ways to decide without it — each appended to `where.ts`, each returning a different operator per dialect, all seven tests still green:

```
OPS[dialect.name]                                    7 passed   <- a per-dialect table
dialect.name == "postgres"                           7 passed   <- loose equality
const { name } = dialect; name === "postgres"        7 passed
const d = dialect.name; d === "postgres"             7 passed
dialect.constructor.name === "PostgresDialect"       7 passed
```

The last three I would not report on their own — they are contrived, and no textual guard catches everything. **The first two are not.** A record keyed by dialect name is the idiomatic way to write a dialect difference and is at least as likely as an `if`, and it is the one form that decides the SQL while containing no comparison at all. Loose equality is a typo away from the case already covered.

Both close for two characters and one alternative, with no new allow-list entries:

```js
/dialect\.name\s*[=!]==?|[=!]==?\s*dialect\.name|switch\s*\(\s*dialect\.name|\[\s*dialect\.name\s*\]|instanceof\s+(Postgres|Sqlite)Dialect/
                     ^                  ^                                    ^^^^^^^^^^^^^^^^^^^^^^^
```
```
head             7 passed          (no new exemptions needed)
OPS[dialect.name]  seam test FAILS
loose equality     seam test FAILS
```

Measured in all three directions. It keeps your "comparison, not mention" line — a lookup keyed by the name is a comparison written as an index, not a mention — so the allow-list stays at two files rather than growing to include `fragment.ts`, `where.ts`, `write.ts` and `plan.ts`, which is the trade you deliberately avoided.

### The scan boundary is load-bearing and unstated

`sourceFiles` walks `orm/` only, and `.unsafe(` appears in three files in the package:

```
orm/Model.ts                  allowed
facades/DB.ts                 outside the scan
database/DatabaseManager.ts   outside the scan
```

Both exclusions are correct — `DB` is the sanctioned raw-SQL surface, and the manager owns the pool — but "only Model.ts reaches the driver" reads as a claim about the repository, and what makes it true is the directory the walk starts in. One sentence in the comment saying the invariant is *"the single door for ORM queries; `DB` is the declared escape hatch"* would put the boundary where the next reader finds it, alongside the join-table exception you already explain that way.

PR #130, opened 2026-07-29
Invariants 2 and 3, the structural halves

### LINKED ISSUE #129: Invariants 2 and 3 have structural halves nothing checks
#127 covered invariants 1 and 4. `plans/orm/README.md` lists **six**, and two more have structural halves that no test enforces.

## Invariant 2's other half

The value half — "two calls with the same shape and different values must produce byte-identical SQL" — is now `binding.invariants.test.ts` (#125).

The structural half is the last sentence of the same invariant: args stay plain data until compile, which "lets the entire compiler be unit-tested with no database at all". `relation-filters.ts` states the operative rule in passing:

> `compile/` may not import `policy.ts` (invariant 2 — the compiler is a pure function of the arguments)

Nothing checks it. And the reason it matters is specific: policies rewrite the **arg tree** before the compiler sees it, which is what makes a scope a two-line `AND` into `args.where` rather than SQL string surgery. A compiler that could consult a policy would invite the surgery back.

**Measured — it holds.** `compile/**` imports eight modules from outside itself: `defaults`, `dialect`, `errors`, `plan`, `registry`, `relation-filters`, `schema`, `shape`. Following those transitively reaches eleven modules in total, and **none** of them touch the database.

The check has to be transitive. A one-hop version is the one that passes while being wrong: `compile/x.ts` importing a helper that imports `Model.ts` reads as clean in review and drags the whole runtime in behind it.

## Invariant 3

> The result-shaping stage is `static $shape` on the model base, not a module-level function. Subclassing is the extension mechanism: a future `ActiveRecordModel` base overrides `$shape` to build instances, and **every model extending it gets that with zero changes to the twelve operations**.

The clause in bold only holds while `$exec` is the thing calling it.

**Measured — it holds.** `plan.shape(` appears at exactly two sites: `Model.ts:929` (the base's `$shape`) and `active-record.ts:57` (the one override). And `$exec` shapes at exactly one place, `this.$shape(plan, rows)`.

The failure mode is quiet and partial, which is what makes it worth pinning: an operation that called `plan.shape(rows)` directly would still return correct data — it would just hand back plain objects on a model whose other eleven operations return instances.

## One deliberate strictness

The import check does not distinguish `import type`. A type-only import is erased and would not actually drag the runtime in, so this is stricter than the letter of the invariant — but it matches what `runtime-isolation.test.ts` already does for `@prisma/*`, and a type-only import is one keyword away from a value import.

## Acceptance

- Both as directory-wide tests, alongside the two from #127.
- The import check computed transitively, with the seed count asserted so an empty scan cannot pass.
- Verified by introducing each violation: a direct `policy.ts` import, a *transitive* one through `shape.ts`, and an operation shaping directly.

nstfkc:
Closes #129. Stacked on #128.

#127 covered invariants 1 and 4. `plans/orm/README.md` lists **six**, and two more have structural halves that nothing enforces.

## Invariant 2's other half

The value half — "two calls with the same shape and different values must produce byte-identical SQL" — is `binding.invariants.test.ts` (#126).

The structural half is the last sentence of the same invariant: args stay plain data until compile, which "lets the entire compiler be unit-tested with no database at all". `relation-filters.ts` states the operative rule in passing:

> `compile/` may not import `policy.ts` (invariant 2 — the compiler is a pure function of the arguments)

The reason is specific: policies rewrite the **arg tree** before the compiler sees it, which is what makes a scope a two-line `AND` into `args.where` rather than SQL string surgery. A compiler that could consult a policy would invite the surgery back.

**It holds.** `compile/**` imports eight modules from outside itself — `defaults`, `dialect`, `errors`, `plan`, `registry`, `relation-filters`, `schema`, `shape` — and following those transitively reaches eleven in total. None touch the database.

**The check is transitive**, because a one-hop version is the one that passes while being wrong: `compile/x.ts` importing a helper that imports `Model.ts` reads as clean in review and drags the whole runtime in behind it. That is also the case my sensitivity check exercises.

## Invariant 3

> The result-shaping stage is `static $shape` on the model base, not a module-level function. Subclassing is the extension mechanism: a future `ActiveRecordModel` base overrides `$shape` to build instances, and **every model extending it gets that with zero changes to the twelve operations**.

The bold clause only holds while `$exec` is the thing calling it.

**It holds.** `plan.shape(` appears at exactly two sites — `Model.ts` (the base's `$shape`) and `active-record.ts` (the one override) — and `$exec` shapes at exactly one place, `this.$shape(plan, rows)`.

The failure mode is quiet and *partial*, which is what makes it worth pinning: an operation calling `plan.shape(rows)` directly still returns correct data. It just hands back plain objects on a model whose other eleven operations return instances.

## One deliberate strictness

The import check does not distinguish `import type`. A type-only import is erased and would not really drag the runtime in, so this is stricter than the letter of the invariant. I kept it that way because it matches what `runtime-isolation.test.ts` already does for `@prisma/*`, and a type-only import is one keyword away from a value import — but it is a choice, not an oversight.

## Verification

Each violation introduced and confirmed caught:

| violation | result |
| --- | --- |
| `compile/where.ts` importing `policy.ts` directly | **fails** the import check |
| `shape.ts` importing `Model.ts` — reached *transitively* from the compiler | **fails** it too |
| `$exec` calling `plan.shape(rows)` instead of the static | **fails** the shaping check |

- 1182 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/architecture.test.ts
+++ b/packages/gemi/orm/architecture.test.ts
@@ -1,5 +1,5 @@
-import { readFileSync, readdirSync } from "node:fs";
-import { join, relative } from "node:path";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
+import { dirname, join, normalize, relative } from "node:path";
 import { describe, expect, test } from "vitest";
 
 /**
@@ -8,8 +8,15 @@ import { describe, expect, test } from "vitest";
  * and for the same reason: a single line added while implementing an operation
  * is how either of them would go, and no unit test would notice.
  *
- *   **1. `Model.$exec` is the single door to the database.**
- *   **4. The dialect seam — the compiler asks the dialect, it does not test it.**
+ *   **1. One choke point** — `Model.$exec` is the only thing that touches the
+ *       database.
+ *   **2. Compile is pure** — and in particular, "lets the entire compiler be
+ *       unit-tested with no database at all". `relation-filters.ts` states the
+ *       operative half: `compile/` may not import `policy.ts`.
+ *   **3. `shape` is a static on the model class** — subclassing is the
+ *       extension mechanism, so shaping has to route through it.
+ *   **4. The dialect seam** — the compiler asks the dialect, it does not test
+ *       it.
  *
  * Both are enforced here by allow-list rather than by prohibition, because both
  * have real exceptions. A list makes an exception a decision someone wrote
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — invariants 2 and 3, the structural halves (closes #129)

Read the diff against its true base (`bf07408`, #128). 1182 tests pass. All three stated violations reproduce:

```
compile/where.ts importing policy.ts        import check FAILS
shape.ts importing Model.ts (transitive)    import check FAILS
$exec calling plan.shape(rows)              shaping check FAILS
```

**Making it transitive is the whole value**, and the reasoning is right: a one-hop check is the version that passes while being wrong. The `import type` decision is a genuine choice made for a stated reason and matching the precedent next door, which is the right way to make it.

### Invariant 3's pair is the better design, and it is worth saying why

The shaping check does not look for a forbidden spelling — it asserts `this.$shape(` occurs in `Model.ts` **exactly once**. That is a positive fact, so it survives evasion in a way pattern-matching does not. I tried to route around it:

```ts
((f) => f(rows as unknown[]))(plan.shape.bind(plan))     shaping check FAILS
```

Still caught, because the required call went missing rather than a banned one appearing. Contrast the dialect-seam check on #128, where I found five spellings that decide on the dialect and slip past the pattern. Same file, two approaches, and this one is the transferable lesson: assert the thing that must be **present**, not the absence of the thing that must not be.

### The edge pattern misses re-exports and side-effect imports

The walk matches `import … from "…"` only. Two kinds of edge are invisible, both verified against `shape.ts`:

```ts
export { Model } from "./Model";     10 passed   <- re-export
import "./Model";                    10 passed   <- side-effect import
```

Either drags `Model.ts` into the compiler's closure while the check reports it clean.

**To be fair about the size of this: the current closure is unaffected.** I computed it both ways and they are identical — 29 modules, zero edges missed today — so the eleven-module result in the description is correct as measured.

What makes it worth fixing anyway is where the gap sits. `compile/index.ts` is one of the **seeds**, and it is composed almost entirely of `export … from` lines. A single `export { applyPolicies } from "../policy"` added there would be exactly the "one line added while implementing an operation" this test exists to catch, and it would not be caught.

One alternative and one union, measured in all three directions:

```js
/^\s*(?:import|export)\s+(?:type\s+)?[^;]*?from\s+"([^"]+)"|^\s*import\s+"([^"]+)"/gm
        ^^^^^^^^^^^^^^^                                      ^^^^^^^^^^^^^^^^^^^^^^^
const resolved = resolve(current, match[1] ?? match[2]);
```
```
head                 10 passed      (closure unchanged, no new exemptions)
export … from        import check FAILS
import "…"           import check FAILS
```

The re-export half is the one I would take; the bare-import half is cheap enough to come along.

PR #134, opened 2026-07-29
`track`'s relationKeys parameter never changed the outcome

### LINKED ISSUE #133: track()'s relationKeys parameter never changes the outcome, and its test is vacuous
Invariant 5 is the last of the six I had not looked at. It holds — provenance is off by default (`options?.track === true`, strict) and `ActiveRecordModel` is the per-model half, since it hydrates through `Model.wrap`, which tracks. But looking at it turned up something else.

## The parameter does nothing

`track(row, schema, relationKeys)` builds the snapshot like this:

```ts
const skip = new Set(relationKeys);
for (const name of Object.keys(source)) {
  // Relations are not this row's columns to write, and an attached child array
  // in a snapshot would make every `save` look dirty.
  if (skip.has(name)) continue;
  if (!(name in schema.fields)) continue;
  snapshot[name] = source[name];
}
```

**Every name `skip` excludes is excluded again on the next line.** A relation is never a field — Prisma will not let a model declare a scalar and a relation under one name — so the second check subsumes the first entirely.

Measured: tracking a row carrying `accounts`, `organization` and `_count` produces a byte-identical snapshot with the relation keys passed and with `[]`.

## What it cost

Two things, both about reading rather than behaviour.

**The two call sites look like they disagree.** `Model.wrap` passes `[]`; `$exec` passes `plan.relations.map(r => r.as)`. A reader comparing them would reasonably conclude an instance from `wrap` is snapshotted differently from one from a query — and specifically that `wrap` has a bug, since it is the one that looks like it forgot. Neither is true.

**The reason is attached to the line that is not doing the work.** The comment explaining why relations must stay out of a snapshot sits on the `skip` check, not on the field check that actually enforces it.

## And the test is vacuous

```ts
test("relation keys are excluded", () => {
  const row = { id: 1, email: "a@b.c", accounts: [] as unknown[] };
  track(row, user, ["accounts"]);
  expect(provenanceOf(row)?.snapshot).toEqual({ id: 1, email: "a@b.c" });
});
```

Changing `["accounts"]` to `[]` — **it still passes.** The test named after the parameter never exercised it; the test immediately below it ("keys that are not fields are ignored") is what covers the real mechanism.

Fourth vacuous test in this series, after #96, #106 and #123.

## Acceptance

- The parameter removed, and the reason moved onto the check that performs the exclusion. `track` is exported from `orm/index.ts`, so this is a signature change to the public surface — worth stating, though the argument was inert.
- The test replaced by one over every shape a relation read attaches: a to-many array, a to-one object, a `_count`, and all three at once.
- A test stating the assumption the removal rests on — that no relation name is also a field name.
- Verified by deleting the field check and confirming the new cases fail.

nstfkc:
Closes #133. Stacked on #132.

Invariant 5 was the last of the six I had not looked at. **It holds** — provenance is off by default (`options?.track === true`, strict equality) and `ActiveRecordModel` is the per-model half, since it hydrates through `Model.wrap`, which tracks. Looking at it turned up something adjacent.

## The parameter does nothing

```ts
const skip = new Set(relationKeys);
for (const name of Object.keys(source)) {
  // Relations are not this row's columns to write, and an attached child array
  // in a snapshot would make every `save` look dirty.
  if (skip.has(name)) continue;
  if (!(name in schema.fields)) continue;
  snapshot[name] = source[name];
}
```

**Every name `skip` excludes is excluded again on the next line.** A relation is never a field — Prisma will not let a model declare a scalar and a relation under one name — so the second check subsumes the first entirely.

Measured: a row carrying `accounts`, `organization` and `_count` snapshots byte-identically with the relation keys passed and with `[]`.

## What it cost

Legibility, twice.

**The two call sites looked like they disagreed.** `Model.wrap` passed `[]`; `$exec` passed `plan.relations.map(r => r.as)`. A reader comparing them would reasonably conclude an instance from `wrap` is snapshotted differently from one from a query — and specifically that `wrap` had forgotten something. Neither is true.

**The reason sat on the wrong line.** The comment explaining why relations must stay out of a snapshot was attached to the `skip` check rather than to the field check that actually enforces it.

## The test was vacuous

```ts
test("relation keys are excluded", () => {
  const row = { id: 1, email: "a@b.c", accounts: [] as unknown[] };
  track(row, user, ["accounts"]);
  expect(provenanceOf(row)?.snapshot).toEqual({ id: 1, email: "a@b.c" });
});
```

Changing `["accounts"]` to `[]` — **it still passes.** The test named after the parameter never exercised it; the one immediately below covers the real mechanism.

Fourth vacuous test found in this series, after #96, #106 and #123.

## The change

- Parameter removed; the reason moved onto the field check that performs the exclusion.
- The test replaced by one over **every shape a relation read attaches** — a to-many array, a to-one object, a `_count`, and all three at once — rather than the single array it used.
- A test stating the assumption the removal rests on: no relation name is also a field name.

`track` is exported from `orm/index.ts`, so this changes a public signature. Worth stating plainly, though the argument it drops was inert.

## Verification

Deleting the field check fails **all four** new cases. The old test would have kept passing, on the strength of an argument that did nothing.

- 1186 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -304,7 +304,7 @@ export abstract class Model {
     Object.assign(instance, row);
     // Tracked on the instance, not the argument — `save()` on the instance has
     // to diff against the values it was constructed from.
-    track(instance, schema, []);
+    track(instance, schema);
 
     return instance;
   }
@@ -905,8 +905,7 @@ export abstract class Model {
       // override gets provenance for free — which is the same argument that put
       // everything else in `$exec`.
       if (options?.track === true) {
-        const relationKeys = (plan.relations ?? []).map((relation) => relation.as);
-        for (const row of rowsOf(result)) track(row, schema, relationKeys);
+        for (const row of rowsOf(result)) track(row, schema);
       }
 
       return result;
--- a/packages/gemi/orm/provenance.test.ts
+++ b/packages/gemi/orm/provenance.test.ts
@@ -22,14 +22,14 @@ describe("what gets snapshotted", () => {
   test("only the fields actually present on the row", () => {
     // A partial select: the row has two of the model's fields.
     const row = { id: 1, email: "a@b.c" };
-    track(row, user, []);
+    track(row, user);
 
     expect(provenanceOf(row)?.snapshot).toEqual({ id: 1, email: "a@b.c" });
   });
 
   test("the primary key is captured separately, at fetch time", () => {
     const row = { id: 7, email: "a@b.c" };
-    track(row, user, []);
+    track(row, user);
 
     // Captured rather than read at save time, so mutating `id` cannot move the
     // row a save targets — it would otherwise update a different record.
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — `track`'s `relationKeys` never changed the outcome (closes #133)

Read the diff against its true base (`59b64d8`, #132). 1186 tests pass.

**Both claims reproduce.** The old test passes identically with `[]`:

```
track(row, user, ["accounts"])  ->  20 passed
track(row, user, [])            ->  20 passed
```

...and deleting the field check fails every new case, which is what makes them the real cover:

```
a to-many array is excluded          FAILS
a to-one object is excluded          FAILS
a relation count is excluded         FAILS
all three at once is excluded        FAILS
keys that are not fields are ignored FAILS   (pre-existing)
```

**I checked the soundness structurally rather than taking the Prisma rule on trust**, since the removal rests on it. `relation.as` is `key` from `Object.keys(include)`/`Object.keys(select)`, and both loops require `schema.relations[key]` — `include` throws `UnknownRelationError` without it, `select` skips. So every name that reached `relationKeys` was a declared relation name, and for a well-formed schema that is never a field. The subsumption is total.

**And the parameter could not have protected the case it looked like it protected.** If a name ever were both, `$exec` would skip it and `Model.wrap` — passing `[]` — would not. So the two call sites already disagreed for the only input the argument could have mattered on. That is a stronger version of the legibility point: they did not merely *look* inconsistent, they *were*, over an empty set.

The public-signature change costs nothing findable — `track` has exactly two callers in the repo, both updated, and no template or doc calls it.

Fourth vacuous test in the series, said plainly. That count is becoming the most useful thing this series produces: three of the four were found by asking "would this fail if the thing it names were broken", and it is a cheap question to ask of a test at review time.

### The assumption test samples where the rest of the series sweeps

`no relation name is also a field name` iterates two schemas, `user` and `organization`. There are fourteen in `fixtures.ts`, and all fourteen have zero overlap:

```
14 schemas, overlap=0 for every one
user 2  account 2  organization 2  post 1  tag 1  ledger 1  ledgerEntry 1
category 2  membership 0  mapped 0  reading 0  bare 0  profile 1  userWithProfile 3
```

So widening it is free, and it matches the standard the rest of this run has held — #118 walked all 143 sites, #120 all 188, #130 the whole transitive closure. A two-element sample of an assumption a deletion rests on is the one place a spot check reads oddly.

Worth considering the stronger home too: the guarantee is *Prisma's*, so `generated.test.ts` — where #132 just established asserting on the emitted artifact — is closer to it than a hand-written fixture. Fixtures are precisely where an overlap could appear later without Prisma being involved, and that is the case that would silently put a relation into a snapshot and make every `save` look dirty.

PR #136, opened 2026-07-29
Mutation testing, and the Bytes branch nothing entered

### LINKED ISSUE #135: Mutation testing: 30 of 149 mutants survive the ORM suite
Four vacuous tests have turned up in this series — #96, #106, #123 and #133 — and every one was found *by accident*, while doing something else. The discipline that found them each time was the same: change the implementation, check the test notices. That is mutation testing done by hand, one case at a time.

So I ran it properly. No tool is installed and none is needed: the unit suite takes **1.7s**, which makes a mutant cost about two seconds.

## Method

One mechanical mutation at a time, six operators — `===`↔`!==`, `>=`→`>`, `<=`→`<`, `&&`↔`||` — applied to a single occurrence, suite run, source restored. A surviving mutant is behaviour no assertion depends on.

| file | mutants | survived |
| --- | --- | --- |
| `provenance.ts` + `shape.ts` | 16 | **7** |
| `policy.ts`, `plan.ts`, `registration.ts` | 133 | **23** |

~20% overall. That is a rate, not a verdict — see below.

## Not every survivor is a gap

Worth stating early, because it is the trap in this technique and I walked into it.

`plan.ts:305` — `collapsedList` returning `value.length === 0 ? "[]" : "[*]"` — survives. I assumed a coverage hole, added `in: []` to the invariant corpus, re-ran, and it **still survived**. Inverting the condition only swaps two labels: empty lists still get one key and non-empty lists another, so nothing observable changes. It is an **equivalent mutant**.

`provenance.ts:92` is the same story: `if (row === null || typeof row !== "object") return undefined` guards a `WeakMap.get`, and `WeakMap.get` on a primitive already returns `undefined` rather than throwing. The line is a fast path, not a correctness check.

A survivor is a *question*, not a defect.

## The real gap this found

`provenance.ts`'s `same()` — the dirty check's value comparison. Its own comment states the consequence:

> without it, every save of a row carrying a timestamp would rewrite that timestamp

**Three mutants inside the `Bytes` branch survived because nothing ever entered it.** `Uint8Array` appears nowhere in `provenance.test.ts`. Prisma maps `Bytes` to `Uint8Array`, so every fetched value is a fresh view and `!==` is always true — without the byte-wise comparison every `save` of a row carrying one rewrites it, and no test would have noticed the branch being deleted.

A fourth survived on the Date branch: `a instanceof Date && b instanceof Date` could be `||` with the suite still green, because no test compares a Date against a non-Date. Under that mutation, assigning a string over a fetched `Date` calls `getTime` on the string and throws from inside a dirty check.

**The behaviour is correct** — I measured all of it before writing anything. The tests were simply absent.

Two more survived on `untracked`'s message, which names what the caller passed so that "the query did not ask for tracking" and "this is a copy" have different fixes. A diagnostic nobody asserts can quietly start saying the wrong thing.

## Acceptance

- The `Bytes` branch covered: equal views, a differing byte, both length directions, and a view into a larger buffer (`byteOffset`/`byteLength` have to be honoured).
- The Date branch covered against non-Dates, both directions.
- `untracked`'s shape naming covered.
- The two equivalent mutants recorded **in the source**, so the next person to run this does not chase them.
- Re-measured: `provenance.ts` from 7 survivors to 1, and that one is equivalent.

The 23 in `policy.ts` / `plan.ts` are not in this pass. The harness is a dozen lines and the method is written down here.

nstfkc:
Closes #135. Stacked on #134.

Four vacuous tests have turned up in this series — #96, #106, #123, #133 — and every one was found **by accident**. The discipline that caught them was always the same: change the implementation, check the test notices. That is mutation testing done by hand, one case at a time.

So I ran it properly. No tool is installed and none is needed: the unit suite takes **1.7s**, which makes a mutant cost about two seconds.

## Method and result

One mutation at a time, six operators (`===`↔`!==`, `>=`→`>`, `<=`→`<`, `&&`↔`||`), suite run, source restored.

| file | mutants | survived |
| --- | --- | --- |
| `provenance.ts` + `shape.ts` | 16 | **7** |
| `policy.ts`, `plan.ts`, `registration.ts` | 133 | **23** |

## Not every survivor is a gap — and I walked into that first

`collapsedList` returning `value.length === 0 ? "[]" : "[*]"` survives. I assumed a coverage hole, added `in: []` to the invariant corpus, re-ran — and it **survived again**. Inverting the condition only swaps two labels: empty lists still get one key and non-empty another, so nothing observable changes. An **equivalent mutant**.

`provenanceOf`'s null guard is the same story: it fronts a `WeakMap.get`, and `WeakMap.get` on a primitive already returns `undefined` rather than throwing. A fast path, not a correctness check.

Both are now recorded **in the source**, so the next person to run this does not chase them. A survivor is a question, not a defect.

## The real find

`same()` — the dirty check's value comparison — whose own comment states the consequence:

> without it, every save of a row carrying a timestamp would rewrite that timestamp

**Three mutants inside its `Bytes` branch survived because nothing ever entered it.** `Uint8Array` appears nowhere in `provenance.test.ts`. Prisma maps `Bytes` to `Uint8Array`, so every fetched value is a fresh view and `!==` is always true — without the byte-wise comparison, every `save` of a row carrying one rewrites it, and deleting the branch outright would have gone unnoticed.

A fourth survived on the Date branch: `a instanceof Date && b instanceof Date` could be `||`, because no test compares a Date against a non-Date. Under that mutation, assigning a string over a fetched `Date` calls `getTime` on the string and throws from inside a dirty check — a long way from where the caller made the mistake.

**The behaviour is correct.** I measured all of it before writing anything: equal views are clean, a differing byte is dirty, both length directions are dirty, a view into a larger buffer honours `byteOffset`/`byteLength`, and a Date replaced by a string is dirty rather than an error. The tests were absent, not the logic.

Two more survived on `untracked`'s message, which names what the caller passed so that "the query did not ask for tracking" and "this is a copy of one that did" have different fixes. A diagnostic nobody asserts can quietly start saying the wrong thing.

## Also

`in: []` and `notIn: []` stay in the corpus regardless of the equivalent mutant — an empty list compiles to a constant-false predicate rather than `= any($1)`, a distinct statement the corpus did not reach.

## Verification

Re-ran the harness: **`provenance.ts` from 7 survivors to 1**, and that one is the equivalent mutant now documented in the source.

The 23 in `policy.ts` / `plan.ts` are **not** in this pass — the harness is a dozen lines and the method is written up in #135.

- 1201 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/corpus.ts
+++ b/packages/gemi/orm/corpus.ts
@@ -26,6 +26,12 @@ export const READS: unknown[] = [
   { where: { id: { in: [1, 2] } } },
   { where: { id: { in: [1, 2, 3] } } },
   { where: { id: { notIn: [1] } } },
+  // An empty list keeps its own plan key: `in: []` compiles to a constant-false
+  // predicate rather than `= any($1)`, which is a different statement. Found by
+  // mutation testing — `collapsedList`'s `length === 0` could be inverted with
+  // nothing noticing, because the corpus had no empty list in it.
+  { where: { id: { in: [] } } },
+  { where: { id: { notIn: [] } } },
   { where: { name: { contains: "a", mode: "insensitive" } } },
   { where: { AND: [{ id: 1 }, { name: "a" }] } },
   { where: { OR: [{ id: 1 }, { name: "a" }] } },
--- a/packages/gemi/orm/provenance.test.ts
+++ b/packages/gemi/orm/provenance.test.ts
@@ -7,6 +7,7 @@ import {
   provenanceOf,
   resnapshot,
   track,
+  untracked,
 } from "./provenance";
 
 /**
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — mutation testing, and the `Bytes` branch nothing entered (closes #135)

Read the diff against its true base (`fe9679b`, #134). 1201 tests pass. I ran my own mutation pass rather than taking the numbers — same six operators, comment-aware so that `!==` inside a doc comment is not counted as a site:

```
base (#134)   17 mutants,  8 survived
head (#136)   17 mutants,  1 survived
```

The survivors on the base land exactly where the description says:

```
provenance.ts:221  && -> ===   ArrayBuffer.isView(a) && ArrayBuffer.isView(b)   |
provenance.ts:224  !== -> ===  left.length !== right.length                     |  the Bytes branch
provenance.ts:226  !== -> ===  left[i] !== right[i]                             |
provenance.ts:217  && -> ||    a instanceof Date && b instanceof Date              the Date branch
provenance.ts:186  x3          row === null || row === undefined                   untracked's message
provenance.ts:87   || -> &&    provenanceOf's null guard                           equivalent
```

(8 against your 7 is operator set — I apply `||`→`&&` as well as `&&`→`||`, which adds one on `untracked`. Same picture.)

**The central claim holds, tested the strongest way available.** Rather than mutating an operator I deleted the whole `Bytes` comparison:

```
base    1186 passed                     nothing notices
head    2 failed
          equal bytes in different views are not dirty
          a view into a larger buffer compares only its own window
```

A branch whose own comment states the consequence — *every save of a row carrying a timestamp would rewrite that timestamp* — removable in full without a red test. That is the best single argument in this PR and it is worth keeping in the description.

**I checked both equivalent mutants rather than accepting them**, since "equivalent" is exactly the claim a tired reviewer waves through:

- `provenanceOf` — under `&&`, `null` and a primitive both fall past the guard to `provenance.get(...)`, and `WeakMap.get` returns `undefined` for both rather than throwing. Nothing observable changes.
- `collapsedList` — inverting maps empty to `"[*]"` and non-empty to `"[]"`. Still a bijection over the same two partitions, so no pair of shapes changes its relationship to any other. Equivalent.

Both sound. And **the corpus lines are right independently of that**, as you argue — the three forms are three different statements:

```
in [1,2]  ->  where "id" = any ($1)
in []     ->  where false
notIn []  ->  where true
```

**"A survivor is a question, not a defect"** is the sentence I would keep, and recording the two answers *in the source* rather than in the PR is what makes it stick. So is walking into `collapsedList` first and saying so — adding `in: []`, re-running, finding it survives *again* is the step that separates a coverage hole from an equivalent mutant, and it is the one most people skip.

### The harness is now the thing stated and enforced by nothing

#127, #129 and #131 all made the same argument: an invariant that lives in a comment and is checked by no test is one line away from being false. The mutation harness is currently a rule that lives in an issue. Four vacuous tests were found by accident, this pass found four more in one file, and the next person to add a branch to `same()` has nothing telling them to re-run it.

The suite is 1.7s, which is what makes this unusual — a scoped run is seconds, not a nightly job. Something like the allow-list shape already used twice in this series would fit: a committed harness plus a list of files whose survivor count must stay at its recorded number, with the equivalent mutants named. That turns "I ran this once" into the same kind of standing check `runtime-isolation.test.ts` is, and it is the only way the 23 in `policy.ts` and `plan.ts` do not become a number nobody re-measures.

PR #138, opened 2026-07-29
The orderBy list branch, and a correction to #135's method

### LINKED ISSUE #137: 16 of policy.ts's mutants survive both suites, and the orderBy list branch is one
Continues #135, which left `policy.ts`'s survivors untriaged — and corrects its method.

## The method was wrong, and the number with it

#135's harness ran only `vitest run orm`. Policy behaviour is largely asserted in the **template's** suite, so a mutant surviving the unit run was not yet evidence of anything.

Re-checked all 20 `policy.ts` survivors against both suites: **16 survive both**; 4 were killed by the template suite all along. The clearest is `isPreScoped`'s `[PRE_SCOPED] === true` — inverting it makes every ordinary query read as pre-scoped and skip its policies. Unit suite: green. Template suite: 15 failures.

That is worth stating loudly, because a survivor list is exactly the kind of artifact that gets treated as a defect list.

## The one worth acting on

`scopeRelationOrderings` takes an **array** `orderBy` through its own branch, and neither suite entered it. Prisma accepts `orderBy: [{ a: "asc" }, { b: "desc" }]`, and `policy.test.ts` covered only the object form.

## And a claim I had to withdraw

My first reading was that inverting that branch's `if (scoped !== entry) changed = true` would discard a scope and return the unscoped ordering — an information leak, since ordering by an unscoped relation count ranks rows by a number computed over rows the caller cannot see.

**Measured, and that is not what happens.** Under the inversion the entries still come back scoped; the only difference is whether the array is rebuilt. The flag is an allocation choice, not a correctness guard, and I would rather record that than ship the more dramatic version.

The mutant is still worth killing — an uncovered branch is an uncovered branch — but by the assertion that can actually see it. `toEqual` passes either way; only identity distinguishes "returned as it was" from "rebuilt equal", which is precisely why the branch stayed uncovered.

## Acceptance

- The array form covered: every entry scoped, a plain column passing through untouched, and a two-entry list so the branch is not exercised as a single-element special case.
- The `changed` flag pinned by identity, which is the assertion that kills the mutant.
- The other 15 survivors left open, with the corrected method recorded so the next pass re-checks against both suites first.

nstfkc:
Closes #137. Stacked on #136.

Continues #135, which left `policy.ts`'s survivors untriaged — and **corrects its method**.

## The method was wrong, and the number with it

#135's harness ran only `vitest run orm`. Policy behaviour is largely asserted in the **template's** suite, so a mutant surviving the unit run was not yet evidence of anything, and I reported it as though it were.

Re-checked all 20 `policy.ts` survivors against both suites:

| | count |
| --- | --- |
| survive both suites | **16** |
| killed by the template suite alone | **4** |

The clearest of the four is `isPreScoped`'s `[PRE_SCOPED] === true`. Inverting it makes every ordinary query read as pre-scoped and skip its policies — unit suite green, template suite **15 failures**.

A survivor list is exactly the kind of artifact that gets mistaken for a defect list, so the correction is on #135 as well as here.

## The one worth acting on

`scopeRelationOrderings` takes an **array** `orderBy` through its own branch, and neither suite entered it. Prisma accepts `orderBy: [{ a: "asc" }, { b: "desc" }]`; `policy.test.ts` covered only the object form.

## A claim I withdrew while writing it

My first reading was that inverting that branch's `if (scoped !== entry) changed = true` would discard the scope and return the unscoped ordering — an information leak, since ordering by an unscoped relation count ranks rows by a number computed over rows the caller cannot see.

**Measured, and that is not what happens.** Under the inversion the entries still come back scoped; the only difference is whether the array is rebuilt. The flag is an allocation choice, not a correctness guard.

I went looking for the leak, instrumented the branch to confirm it was even entered, and found the returned value identical either way. The comment in the test now says that, rather than the more dramatic version I first wrote.

The mutant is still worth killing — an uncovered branch is an uncovered branch — but by the assertion that can actually see it. `toEqual` passes either way; only **identity** distinguishes "returned as it was" from "rebuilt equal", which is precisely why the branch stayed uncovered.

## The tests

- every entry of a list is scoped, with a plain column passing through untouched and two entries so the branch is not exercised as a single-element special case;
- the `changed` flag pinned by identity — this is the one that kills the mutant.

The other 15 survivors are left open, with the corrected method recorded so the next pass re-checks against both suites first.

- 1203 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -820,6 +820,86 @@ describe("applyNestedPolicies", () => {
     expect(out.orderBy.accounts).toEqual({ _count: "asc", where: seen });
   });
 
+  /**
+   * The **array** form of `orderBy`, which Prisma accepts and nothing covered.
+   *
+   * `scopeRelationOrderings` takes a list through its own branch, and neither
+   * suite entered it: mutating that branch's `changed` comparison left both
+   * green (#135).
+   *
+   * I expected the flag to be load-bearing and it is not — measured, because
+   * the first version of this comment claimed a scope could be dropped. Under
+   * the inversion the entries still come back scoped; what changes is only
+   * whether the array is rebuilt. So the flag is an allocation choice, and the
+   * two tests below split accordingly: this one pins that every entry is
+   * scoped, and the next pins the flag itself.
+   *
+   * Two entries, so the branch is exercised as a list rather than a
+   * single-element special case, and one of them a plain column so an untouched
+   * entry is asserted to come through intact.
+   */
+  test("every entry of an orderBy list is scoped, not just the object form", () => {
+    const reportsOperation: ModelPolicy = {
+      scope: (context) => ({ seenOperation: context.operation }),
+    };
+
+    const out = applyNestedPolicies(
+      {
+        relations: {
+          accounts: { model: "Account", kind: "many" as const },
+          posts: { model: "Post", kind: "many" as const },
+        },
+      },
+      {
+        orderBy: [
+          { name: "asc" },
+          { accounts: { _count: "asc" } },
+          { posts: { _count: "desc" } },
+        ],
+      },
+      USER,
+      false,
+      (model: string) =>
+        model === "Account" || model === "Post"
+          ? {
+              policies: [reportsOperation],
+              schema: { name: model, relations: {} },
+            }
+          : undefined,
+    );
+
+    const seen = { seenOperation: "findMany" };
+    expect(out.orderBy).toEqual([
+      // A plain column is not a relation and is returned untouched.
+      { name: "asc" },
+      { accounts: { _count: "asc", where: seen } },
+      { posts: { _count: "desc", where: seen } },
+    ]);
+  });
+
+  /**
+   * ...and a list whose entries name no policied relation is returned **as it
+   * was**, rather than as an equal copy.
+   *
+   * This is the assertion that kills the mutant. The `changed` flag exists so
+   * that an array nothing rewrote is not rebuilt per query, and identity is the
+   * only thing that can see that — `toEqual` passes either way, which is
+   * exactly why the branch went uncovered.
+   */
+  test("an orderBy list with nothing to scope is not rebuilt", () => {
+    const orderBy = [{ name: "asc" }, { email: "desc" }];
+
+    const out = applyNestedPolicies(
+      { relations: { accounts: { model: "Account", kind: "many" as const } } },
+      { orderBy },
+      USER,
+      false,
+      () => undefined,
+    );
+
+    expect(out.orderBy).toBe(orderBy);
+  });
+
   test("an unregistered relation target is left for the planner to report", () => {
     const args = { include: { unknown: true } };
     const out = applyNestedPolicies(


nstfkc:
## Review — the `orderBy` list branch, and a correction to #135's method (closes #137)

Read the diff against its true base (`e793f66`, #136). 1203 tests pass.

**The method correction is the important half**, and it lands on me too — I have a mutation harness running against these same files that also ran `vitest run orm` only, so my numbers for `policy.ts` are subject to exactly the error you are correcting. Putting the retraction on #135 as well as here is right; a survivor list that circulated as a defect list would have been worse than not measuring.

The array branch of `scopeRelationOrderings` really was unentered, and the identity assertion really is what kills the mutant — `toEqual` passes either way.

### The withdrawn claim was correct, and the new test is why it looked otherwise

You withdrew the leak on the grounds that *"under the inversion the entries still come back scoped"*. That is true of the array you measured with, and only of that array. The new test's list is

```js
[{ name: "asc" }, { accounts: { _count: "asc" } }, { posts: { _count: "desc" } }]
```

Under `if (scoped === entry) changed = true`, the **plain column** is the entry that is *not* rewritten — so it sets `changed`, `return changed ? out : orderBy` returns the rebuilt `out`, and every relation comes back correctly scoped. The plain column is masking the mutant.

Drop it and the flag is load-bearing:

```
orderBy: [{ accounts: … }, { posts: … }]          (relations only)
  head    2/2 scoped   [{"accounts":{"_count":"asc","where":{…}}}, {"posts":{"_count":"desc","where":{…}}}]
  mutant  0/2 scoped   [{"accounts":{"_count":"asc"}},             {"posts":{"_count":"desc"}}]
```

The caller's original array comes back, unscoped — which is the leak you first wrote and then removed. So `changed` is not an allocation choice, and the comment now committed in `policy.test.ts` says something false about the code.

The sentence that explains the test is the one that disarms it: *"two entries, so the branch is exercised as a list rather than a single-element special case, and one of them a plain column so an untouched entry is asserted to come through intact."* Both reasons are good. Their combination is what makes the case blind, because an untouched entry is exactly what the inverted flag needs.

One more case, verified both ways:

```
an orderBy list of only policied relations is still scoped
  head    97 passed
  mutant  FAILS
```

### The same branch exists one function down, and is still uncovered

`scopeRelationFilters` has a byte-identical array branch — `changed`, `.map`, `return changed ? out : where`. Mutating it survives the entire suite:

```
policy.ts:1284   !== -> ===     1203 passed
```

And it fails the same way, on a `where` rather than an ordering:

```
AND: [{ name: "x" }, { accounts: { some: … } }]      mixed    -> scoped under both  (masked)
AND: [{ accounts: { some: … } }, { posts: { some: … } }]
  head    {"accounts":{"some":{"id":1,"AND":[{scope}]}}}, {"posts":{"some":{"id":2,"AND":[{scope}]}}}
  mutant  {"accounts":{"some":{"id":1}}},                 {"posts":{"some":{"id":2}}}
```

This is the more serious of the two: a dropped scope in an `orderBy` ranks rows by a count over rows the caller cannot see; a dropped scope in a `where` **matches** them. An `AND` of two relation filters is an ordinary query, and it is the shape no test uses.

Neither is a live bug — the code is right as written. What makes them worth fixing together is that "the flag is an allocation choice" is an invitation to simplify `return changed ? out : orderBy`, and one of the two simplifications a reader might reach for is the leak.

PR #140, opened 2026-07-29
The policy layer's malformed-input passthrough, which nothing asserted

### LINKED ISSUE #139: The policy layer's malformed-input passthrough is stated in a comment and untested
Continues #137, which left 15 `policy.ts` mutants surviving both suites.

## The contract

`scopeRelationOrderings` states it:

> Malformed, and the compiler reports it with the message it has.

The policy layer runs **before** the compiler, so when it meets a nested argument that is not the shape it expects — an array where a node belongs, a number, a string — its job is to do **nothing** and let the compiler produce the refusal. The same guard is repeated at seven places: the include walk, the ordering walk, the `_count` walk, its `select`, and two `undefined`/`false` skips.

None of them was tested. Every one could have its `||` turned into an `&&` with both suites green.

## Why it matters

Under that change the policy layer walks into an array where it expects a node, and the caller gets whatever that produces instead of the compiler's precise refusal. This repo has fixed that shape of failure repeatedly — #101 and #107 were both refusals that named the wrong thing, and the standing complaint is that it sends the reader to a query they did not write.

## Measured, and most of them are equivalent

Nine malformed shapes, all confirmed to pass through unchanged. Then each of the 15 survivors re-tested against the new assertions:

| | count |
| --- | --- |
| real gaps, now killed | **3** — the include walk, the ordering node, and `_count.select` |
| equivalent mutants | the rest |

The equivalent ones are guards **subsumed by a later check**, which is the same pattern as `select.ts:28` (#115) and `provenanceOf`'s null guard (#135). `scopeCounts`'s node guard is the clearest: inverting it lets an array through, but the loop that follows does `Object.keys([])` and finds no relation, so the output is byte-identical. Verified directly rather than argued — `{ _count: [] }` and `{ _count: [1, 2] }` produce the same result mutated and unmutated.

That is worth recording as a *rate*: on this file, most surviving mutants are not coverage gaps. A survivor list is a list of questions.

## Acceptance

- The seven guards covered by a table over the malformed shapes each one handles, asserted as **passthrough** — the refusal is the compiler's to make, and asserting on it here would move the contract.
- `false` and `undefined` named apart from the malformed cases: they are Prisma's way of saying "not this relation", not a mistake, and they are skipped by a different guard.
- The remaining survivors identified as equivalent, with the mechanism named, so the next pass does not re-triage them.

nstfkc:
Closes #139. Stacked on #138.

Continues #137, which left 15 `policy.ts` mutants surviving both suites.

## The contract

`scopeRelationOrderings` states it:

> Malformed, and the compiler reports it with the message it has.

The policy layer runs **before** the compiler, so when it meets a nested argument that is not the shape it expects — an array where a node belongs, a number, a string — its job is to do **nothing** and let the compiler produce the refusal. The same guard appears at seven places: the include walk, the ordering walk, the `_count` walk, its `select`, and two `undefined`/`false` skips.

None of them was tested. Every one could have its `||` turned into an `&&` with both suites green.

Under that change the policy layer walks into an array where it expects a node, and the caller gets whatever that produces instead of the compiler's refusal. That is the shape of failure #101 and #107 both fixed: an error naming a query the reader did not write.

## Measured — and most of the survivors are equivalent

Nine malformed shapes, all confirmed to pass through unchanged. Then each of the 15 survivors re-tested against the new assertions:

| | count |
| --- | --- |
| real gaps, now killed | **3** — the include walk, the ordering node, `_count.select` |
| equivalent mutants | the remaining 12 |

The equivalent ones are guards **subsumed by a later check** — the same pattern as `select.ts:28` (#115) and `provenanceOf`'s null guard (#135). `scopeCounts`'s is the clearest: inverting it lets an array through, but the loop that follows does `Object.keys([])`, finds no relation, and returns the node untouched.

Verified rather than argued — `{ _count: [] }` and `{ _count: [1, 2] }` produce byte-identical results mutated and unmutated.

**That is worth recording as a rate.** On this file most surviving mutants are not coverage gaps — the second time in three passes that the survivor list has been mostly questions rather than answers. It is the main cost of the technique and the thing to say out loud when reporting one.

## The tests

- A table over the malformed shapes each guard handles, asserted as **passthrough**. Not by catching an error: the refusal belongs to the compiler, and asserting on it here would move the contract.
- `false` and `undefined` named apart from the malformed cases — they are Prisma's way of saying "not this relation", not a mistake, and a different guard skips them.
- Deep equality rather than identity, since the walk may copy the surrounding object on its way past; what must not change is the malformed node.

## Verification

- 1212 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -900,6 +900,69 @@ describe("applyNestedPolicies", () => {
     expect(out.orderBy).toBe(orderBy);
   });
 
+  /**
+   * **A malformed nested argument is left alone for the compiler to report.**
+   *
+   * The rule is stated in `scopeRelationOrderings` — "Malformed, and the
+   * compiler reports it with the message it has" — and repeated as a guard at
+   * six other places: the include walk, the `_count` walk, its `select`, the
+   * ordering walk, and two `undefined`/`false` skips.
+   *
+   * Nothing tested any of them. Every one of those guards could have its `||`
+   * turned into an `&&` with both suites still green (#137), and under that
+   * change the policy layer walks into an array where it expects a node — so
+   * the caller gets whatever that produces instead of the compiler's precise
+   * refusal. This repo has fixed that shape of failure repeatedly (#101, #107):
+   * an error that names the wrong thing sends the reader to a query they did
+   * not write.
+   *
+   * Asserted as passthrough rather than by catching an error, because the
+   * policy layer's job here is to do *nothing* — the refusal is the compiler's
+   * to make, and asserting on it here would move the contract.
+   */
+  describe("a malformed nested argument is passed through untouched", () => {
+    const scopes: ModelPolicy = { scope: () => ({ tenant: 1 }) };
+
+    const walk = (args: unknown) =>
+      applyNestedPolicies(
+        { relations: { accounts: { model: "Account", kind: "many" as const } } },
+        args as never,
+        USER,
+        false,
+        (model: string) =>
+          model === "Account"
+            ? { policies: [scopes], schema: { name: "Account", relations: {} } }
+            : undefined,
+      );
+
+    test.each([
+      ["an include node that is an array", { include: { accounts: [] } }],
+      ["an include node that is a number", { include: { accounts: 5 } }],
+      ["an include node that is a string", { include: { accounts: "yes" } }],
+      ["an orderBy node that is an array", { orderBy: { accounts: [] } }],
+      ["an orderBy node that is a number", { orderBy: { accounts: 5 } }],
+      ["a _count that is an array", { include: { _count: [] } }],
+      ["a counted relation that is an array", { include: { _count: { select: { accounts: [] } } } }],
+    ])("%s", (_label, args) => {
+      // Deep equality, not identity: the walk may copy the surrounding object
+      // on its way past. What must not change is the malformed node itself.
+      expect(walk(args)).toEqual(args);
+    });
+
+    /**
+     * `false` and `undefined` are not malformed — they are Prisma's way of
+     * saying "not this relation" — and they are skipped by a different guard
+     * that was equally untested. Same assertion, different reason, so they are
+     * named apart rather than folded into the table above.
+     */
+    test.each([
+      ["false", { include: { accounts: false } }],
+      ["undefined", { include: { accounts: undefined } }],
+    ])("an include node that is %s is skipped, not scoped", (_label, args) => {
+      expect(walk(args)).toEqual(args);
+    });
+  });
+
   test("an unregistered relation target is left for the planner to report", () => {
     const args = { include: { unknown: true } };
     const out = applyNestedPolicies(


nstfkc:
## Review — the malformed-input passthrough (closes #139)

Read the diff against its true base (`9c9db0a`, #138). 1212 tests pass.

**The contract is the right one to pin.** "The policy layer runs before the compiler, so on a shape it does not expect it does nothing" is a real invariant, and asserting it as *passthrough* rather than by catching an error is correct — catching here would move the contract to the wrong layer.

I mutated all seven guard sites on both branches. Three flip from surviving to killed, so the coverage claim holds:

```
                                                          #138      #140
policy.ts:1054  nodeArgs in the include walk              SURVIVED  killed
policy.ts:1167  the ordering node                         SURVIVED  killed
policy.ts:1233  nodeArgs inside the _count walk           SURVIVED  killed
```

My own pass over `policy.ts` also reproduces your survivor set in shape — including the `isPreScoped` trio at 895–897 that #137 showed the template suite kills, which is a neat confirmation that the unit-only method overstates. (My harness has that same defect; I am reporting only *killed* results from it, where unit-only is sufficient evidence.)

### The third gap is not `_count.select`, and `_count.select` is not equivalent

The table names the three as *"the include walk, the ordering node, `_count.select`"*. Measured, the third is **1233** — the per-entry `nodeArgs` guard *inside* the `_count` loop, which handles `_count: { select: { accounts: [1,2] } }`. The `select` guard itself is lines 1211–1212, and **both still survive on this branch**.

That matters because they are not equivalent mutants. Inverting either one:

```
include: { _count: { select: null } }

TypeError: null is not an object (evaluating 'Object.keys(source)')
  at scopeCounts (orm/policy.ts:1222)
  at applyNestedPolicies (orm/policy.ts:1023)
```

A raw `TypeError` thrown from the policy layer, which is precisely the failure the PR opens by describing — *"the caller gets whatever that produces instead of the compiler's refusal"*. The passthrough table covers `"oops"`, `5`, `[]`, `[1,2]` for this position, and all four are genuinely subsumed by the loop that follows. `null` is the one that is not, because `Object.keys(null)` throws where `Object.keys(5)` returns `[]`.

So two of the twelve filed as equivalent are real, and one case closes both:

```js
test("a null _count.select passes through instead of crashing", () => { … })
```
```
head        106 passed
1211 mutant FAILS
1212 mutant FAILS
```

### On the rate

> most surviving mutants are not coverage gaps … the main cost of the technique and the thing to say out loud

Agreed, and it is the right thing to report. This is also the hazard in it: a 12-of-15 equivalence rate is reached by classifying twelve, and the two that were wrong are the two where `null` behaves differently from every other malformed value. `scopeCounts`'s clearest case was verified by probing (`{ _count: [] }` and `{ _count: [1,2] }` byte-identical) — the twelve did not each get that, and the two that failed are exactly the ones a per-mutant probe would have caught.

Cheap rule that would have found both: for a guard whose disjuncts are *type* and *null*, the `null` disjunct is never subsumed by a later `Object.keys`, `Object.entries` or spread. It is the one shape that throws rather than iterating empty.

nstfkc:
## Review of `1c034fb` — the two shapers were never compared on failure

New commit on this branch since my last comment, so reviewing it separately. 1219 tests pass.

**The argument is right and the tests are load-bearing.** "Two implementations that agree when things work and disagree when they do not have not been shown to agree" is the same reasoning `failureKind` carries in the differential harness, and it applies here for a sharper reason: the compiled shaper *inlines* what the interpreted one calls, so the failure paths are genuinely separate code.

Verified by making them disagree — the interpreted shaper swallowing `DecodeError`:

```
a BigInt that is not an integer, sqlite     FAILS
a BigInt that is not an integer, postgres   FAILS
a Json that will not parse, sqlite          FAILS
a DateTime that is not one, sqlite          FAILS
```

**The sqlite `Bytes` row is load-bearing too, and its justification is exact.** `needsDecode` on sqlite falls through to `default: false` for `Bytes`, so it is a pure passthrough there — and injecting a divergence confined to the passthrough branch is caught by that one row and nothing else:

```
reading with bytes, sqlite   FAILS
                             1 failed | 23 passed
```

The prior table did carry `"reading with bytes, postgres"` only, so the gap was real.

### Two of the six failure rows cannot fail

> All three decoders that can refuse are covered on both dialects

Not on Postgres. Measured:

```
                 sqlite                postgres
BigInt 3.5       THREW DecodeError     THREW DecodeError
Json "{not json" THREW DecodeError     ok  -> "{not json"
DateTime bad     THREW DecodeError     ok  -> "not a date"
```

Postgres's `decode` has no `DateTime` case at all — it reaches `default: return value` — and its `Json` case returns the value verbatim, for the reason its own comment gives at length: Bun parses `json`/`jsonb`, and re-parsing was the bug that comment records. So on Postgres exactly one decoder can refuse, and it is `BigInt`.

That is why the divergence mutation above turns **four** rows red rather than six: the two postgres rows are success comparisons wearing failure labels, and they would stay green under any failure-path divergence whatsoever.

**Coverage is still complete** — every decoder that can refuse is covered on the dialect where it can — so this is the description overstating rather than a gap, and I would not hold the PR for it. It is worth correcting because the next person extending the table will go looking for a Postgres `Json` failure that cannot exist, and conclude the table is missing a case rather than that the decoder cannot throw.

One observation the new data makes visible, and I do not think it is a defect: a malformed `DateTime` is a `DecodeError` on sqlite and a raw string on postgres, where the field is typed `Date`. Unreachable through a model read — Bun returns a `Date` for a timestamp column, which is exactly why postgres has no case — but this test file is now the first place that asymmetry is written down, and it is worth one line saying it is driver-guaranteed rather than checked.

Carried forward from my earlier comment on `5393135`, since `policy.ts` is untouched here: `include: { _count: { select: null } }` still throws `TypeError: null is not an object` out of `scopeCounts`, and the mutants at 1211–1212 still survive.

PR #142, opened 2026-07-29
The two shapers were asserted to succeed alike, never to fail alike

### LINKED ISSUE #141: The two shapers are asserted to succeed alike, never to fail alike
`shape.ts` has two implementations of one thing, and says why the equivalence test matters:

> Exported because the two implementations must produce *identical* output, and a test can only assert that if it can obtain both. **That test is what stands between this optimisation and a shape divergence visible only in production.**

The test exists and is good — a table of row shapes across both dialects, compared by deep equality *and* by `JSON.stringify` so key **order** is pinned too, plus the relation cases and the aliasing hazard.

## What it does not do

It only ever compares **successful output**.

`DecodeError` is raised from inside `dialect.decode`, and the compiled shaper **inlines** what the interpreted one calls — `d.decode(row[3], f[3])` emitted into generated source, versus a closure calling the same method. A divergence there would be one path throwing while the other returned something plausible, on exactly the malformed input that made it matter.

This repo already carries the lesson elsewhere: `differential.ts` compares failure *kind*, not only values, because two clients that agree when things work and disagree when they do not have not been shown to agree. The same argument applies to the two shapers, and was not applied.

Three decoders can refuse — `BigInt` (`BigInt(3.5)` throws a RangeError), `Json` (unparseable), `DateTime` — and none was compared across the two paths, on either dialect.

## One dialect gap in the table

`reading` — the fixture carrying `Bytes` — is in the table for **postgres only**. On SQLite `Bytes` has no decode branch at all and falls through untouched, so a divergence confined to the passthrough side would not have shown.

## Measured

No divergence: all seven cases agree today, on both dialects. No defect — the assertions were simply absent.

Sensitivity checked in both directions, because an equivalence test is worthless if it cannot see one side breaking:

| break | result |
| --- | --- |
| the **compiled** shaper stops decoding | 5 existing cases fail |
| the **interpreted** shaper stops decoding | 4 existing cases fail, **plus the new failure-equivalence cases** |

The second row is the point: with only the old table, breaking the interpreted path is caught by the output cases; the failure cases catch it *as a failure divergence*, which is the specific thing that would otherwise reach production as "it throws on my machine".

## Acceptance

- Failure equivalence for all three refusing decoders, on both dialects.
- `reading` added to the output table on **sqlite**, not just postgres.
- Verified by breaking each shaper's decode in turn.

nstfkc:
Closes #141. Stacked on #140.

`shape.ts` keeps two implementations of one thing, and says what protects them:

> Exported because the two implementations must produce *identical* output, and a test can only assert that if it can obtain both. **That test is what stands between this optimisation and a shape divergence visible only in production.**

That test exists and is good — a table of row shapes across both dialects, compared by deep equality *and* by `JSON.stringify` so key **order** is pinned too, plus the relation cases and the aliasing hazard.

## What it does not do

It only ever compares **successful output**.

`DecodeError` is raised from inside `dialect.decode`, and the compiled shaper **inlines** what the interpreted one calls — a generated `d.decode(row[3], f[3])` against a closure calling the same method. A divergence there is one path throwing while the other returns something plausible, on exactly the malformed input that made it matter.

The repo already carries this lesson: `differential.ts` compares failure *kind* rather than only values, because two clients that agree when things work and disagree when they do not have not been shown to agree. The same argument applies here and had not been applied. Three decoders can refuse — `BigInt` (`BigInt(3.5)` throws a RangeError), `Json` (unparseable), `DateTime` — and none was compared across the two paths, on either dialect.

## A dialect gap in the existing table

`reading` — the fixture carrying `Bytes` — was in the table for **postgres only**. On SQLite `Bytes` has no decode branch at all and falls through untouched, so a divergence confined to the passthrough side would not have shown. Added.

## Measured

**No divergence.** All seven cases agree today, on both dialects. The assertions were absent, not the behaviour.

## Sensitivity, in both directions

An equivalence test is worthless if it cannot see one side break, so I broke each:

| break | result |
| --- | --- |
| the **compiled** shaper stops decoding | 5 existing cases fail |
| the **interpreted** shaper stops decoding | 4 existing cases fail, **plus the new failure cases** |

The second row is the point. With only the old table, breaking the interpreted path is caught as an *output* difference; the new cases catch it as a **failure divergence** — the specific thing that otherwise reaches production as "it throws on my machine".

- 1219 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/shape.test.ts
+++ b/packages/gemi/orm/shape.test.ts
@@ -4,11 +4,13 @@ import { afterEach, describe, expect, test } from "vitest";
 import { PostgresDialect } from "./dialect/postgres";
 import { SqliteDialect } from "./dialect/sqlite";
 import { mapped, reading, user } from "./fixtures";
+import type { FieldSchema } from "./schema";
 import {
   buildInterpretedShaper,
   buildRowShaper,
   setShaperCompilation,
   shaperCompilationAvailable,
+  type RowShaper,
   type ShapedRelation,
 } from "./shape";
 
@@ -159,6 +161,22 @@ describe("compiled and interpreted shapers agree", () => {
         },
       ],
     ],
+    // ...and the same shape on sqlite, where `Bytes` has no decode branch at
+    // all and falls through. The table covered it on postgres only, so a
+    // divergence confined to the passthrough side would not have shown.
+    [
+      "reading with bytes, sqlite",
+      reading,
+      sqlite,
+      [
+        {
+          id: 1,
+          at: 1_717_200_000_000,
+          digest: new Uint8Array([1, 2, 3]),
+          value: 1.5,
+        },
+      ],
+    ],
     ["a row missing every optional column", user, sqlite, [{ id: 1 }]],
     ["no rows at all", user, sqlite, []],
   ];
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the two shapers were never compared on failure (closes #141)

This commit was on #140 when I reviewed it, so [that comment](https://github.com/nstfkc/gemi/pull/140#issuecomment-5112269989) landed on the wrong PR. Repeating the substance here, plus the sensitivity table which is new.

1219 tests pass. **The argument is right and the tests are load-bearing.** "Two implementations that agree when things work and disagree when they do not have not been shown to agree" is the same reasoning `failureKind` carries in the differential harness, and it applies more sharply here because the compiled shaper *inlines* what the interpreted one calls — genuinely separate code on the failure path.

**The sqlite `Bytes` row earns its place, and the justification is exact.** `needsDecode` on sqlite falls through to `default: false` for `Bytes`, so it is a pure passthrough there. Injecting a divergence confined to the passthrough branch is caught by that one row and nothing else:

```
reading with bytes, sqlite   FAILS      1 failed | 23 passed
```

### The sensitivity table undersells the result

Measured both breaks:

```
compiled stops decoding      10 failed   6 existing  + 4 new failure cases
interpreted stops decoding    8 failed   4 existing  + 4 new failure cases
```

Two differences from the table. The compiled row is **6** existing, not 5 — the sixth is `reading with bytes, sqlite`, which this PR adds, so a count taken before adding it would have said 5. And the new failure cases fire on **both** breaks, not only the interpreted one, so the asymmetry the table draws — *"the second row is the point"* — is not the one that holds. The real asymmetry is smaller and still worth stating: the compiled break is loud either way, while the interpreted break shows up in four fewer existing cases, so the new ones carry proportionally more of it.

More sensitive than claimed, but the claim is the thing a later reader checks against.

### Two of the six failure rows cannot fail

> Three decoders can refuse — `BigInt`, `Json`, `DateTime` — and none was compared across the two paths, on either dialect.

On Postgres, one can. Measured:

```
                 sqlite                postgres
BigInt 3.5       THREW DecodeError     THREW DecodeError
Json "{not json" THREW DecodeError     ok  -> "{not json"
DateTime bad     THREW DecodeError     ok  -> "not a date"
```

Postgres's `decode` has no `DateTime` case — it reaches `default: return value` — and its `Json` case returns the value verbatim, for the reason its own comment records at length: Bun parses `json`/`jsonb`, and re-parsing was the bug. So `postgres/Json` and `postgres/DateTime` are success comparisons wearing failure labels, and they are exactly the two that stay green under every failure-path divergence I could inject.

**Coverage is still complete** — each decoder is compared on the dialect where it can refuse — so this is the description overstating rather than a gap, and I would not hold the PR for it. Worth correcting because the next person extending the table will hunt for a Postgres `Json` failure that cannot exist and conclude a case is missing rather than that the decoder cannot throw.

One observation the new data makes visible, which I do not think is a defect: a malformed `DateTime` is a `DecodeError` on sqlite and a raw string on postgres, where the field is typed `Date`. Unreachable through a model read — Bun returns a `Date` for a timestamp column, which is why postgres has no case — but this file is now the first place that asymmetry is written down, and it is worth a line saying it is driver-guaranteed rather than checked.

PR #144, opened 2026-07-29
Postgres had no identifier-quoting or placeholder test

### LINKED ISSUE #143: Postgres has no identifier-quoting or placeholder test, and it is the dialect where a shift is worse
`sqlite.test.ts` opens with `describe("identifiers and placeholders")`: quoting, escaping an embedded quote, refusing the parameter sentinel, and the positional `?`. `postgres.test.ts` has **no such block at all**.

## Why the missing one is the one that matters

The SQLite test says why the sentinel case is there:

> NUL is the parameter sentinel in `compile/fragment.ts`, so it is the one character that could shift a placeholder's position rather than merely produce broken SQL. Unreachable from a generated schema — asserted anyway, **so the invariant is unconditional rather than argued**.

It was conditional — on the dialect. And Postgres is the side where a shifted position is **worse**: SQLite's `?` is positional, so a shift produces a broken statement, while Postgres's `$7` silently addresses a *different value*. Same class of bug, one loud and one quiet, and the quiet one had no test.

`postgres.ts` implements the guard, with a comment pointing at the SQLite implementation. Only the assertion was missing.

## Also missing: the numbering itself

`placeholder(index)` returning `$1`, `$2` is the dialect seam's clearest difference, and the whole `Fragment` sentinel machinery exists because placeholders cannot be rendered until a fragment's position in the finished statement is fixed.

It is exercised *indirectly* — `write.test.ts` asserts `values ($1, $2, $3, $4, $5, $6)` — but the dialect contract itself was never asserted, so a change to it fails somewhere else with a message about a statement rather than about numbering.

## Measured

Postgres behaves correctly: `"id"`, `"we""ird"`, NUL refused at the end, in the middle and alone, and `placeholder(0) === "$1"` through `placeholder(41) === "$42"`. No defect — the assertions were absent.

Sensitivity checked both halves:

| break | result |
| --- | --- |
| remove the NUL guard from `postgres.ts` | 3 cases fail |
| make `placeholder` zero-based | 3 cases fail, plus an in-list case |

## Acceptance

- The block `sqlite.test.ts` has, on Postgres: quoting, escaping, the sentinel in three positions.
- Placeholder numbering asserted **against SQLite's answer for the same index**, because "they differ" is the property — a shared implementation would pass a test that only looked at one.

nstfkc:
Closes #143. Stacked on #142.

`sqlite.test.ts` opens with `describe("identifiers and placeholders")` — quoting, escaping an embedded quote, refusing the parameter sentinel, and the positional `?`. **`postgres.test.ts` has no such block at all.**

## The missing one is the one that matters more

The SQLite test says why the sentinel case is there:

> NUL is the parameter sentinel in `compile/fragment.ts`, so it is the one character that could shift a placeholder's position rather than merely produce broken SQL. Unreachable from a generated schema — asserted anyway, **so the invariant is unconditional rather than argued**.

It was conditional — on the dialect. And Postgres is the side where a shifted position is **worse**: SQLite's `?` is positional, so a shift produces a broken statement, while Postgres's `$7` silently addresses a *different value*. Same class of bug, one loud and one quiet, and the quiet one had no test.

`postgres.ts` implements the guard already, with a comment pointing at the SQLite implementation. Only the assertion was missing.

## Numbering, too

`placeholder(index)` → `$1`, `$2` is the dialect seam's clearest difference, and the whole `Fragment` sentinel machinery exists because a placeholder cannot be rendered until its fragment's position in the finished statement is fixed.

It is exercised *indirectly* — `write.test.ts` asserts `values ($1, $2, $3, $4, $5, $6)` — but the dialect contract itself was never asserted, so a change to it fails somewhere else, with a message about a statement rather than about numbering.

Asserted here **against SQLite's answer for the same index**, because "they differ" is the property: a shared implementation would pass a test that only looked at one.

## Measured

Postgres behaves correctly throughout — `"id"`, `"we""ird"`, NUL refused at the end, in the middle and alone, `$1` through `$42`. **No defect**; the assertions were absent.

Sensitivity checked on both halves:

| break | result |
| --- | --- |
| remove the NUL guard from `postgres.ts` | 3 cases fail |
| make `placeholder` zero-based | 3 cases fail, plus an in-list case |

- 1227 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/dialect/postgres.test.ts
+++ b/packages/gemi/orm/dialect/postgres.test.ts
@@ -3,8 +3,10 @@ import { describe, expect, test } from "vitest";
 import { user } from "../fixtures";
 import { compileRead } from "../compile/read";
 import { PostgresDialect } from "./postgres";
+import { SqliteDialect } from "./sqlite";
 
 const postgres = new PostgresDialect();
+const sqlite = new SqliteDialect();
 
 /**
  * The array parameter behind `= any($1)`.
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — Postgres had no identifier-quoting or placeholder test (closes #143)

Read the diff against its true base (`1c034fb`, #142). 1227 tests pass, and the base's `postgres.test.ts` genuinely has no such block — four describes, none of them about identifiers or placeholders.

**The NUL half is a real hole, and bigger than the sensitivity table shows.** I deleted the guard from `postgres.ts` on both branches:

```
base (#142)   1219 passed        the guard is deletable in silence
head (#144)   3 failed           the three new cases
```

Nothing anywhere caught it before — not the compiler tests, not the template suite's Postgres run. So this is the case the SQLite comment describes as *"asserted anyway, so the invariant is unconditional rather than argued"*, and it was argued rather than asserted on the dialect where a shifted placeholder is silent. That framing — `?` breaks the statement, `$7` addresses a different value — is the right reason to care, and it is why this is worth more than the line count suggests.

The three positions are also an improvement on the file being mirrored: `sqlite.test.ts` tests the sentinel only at the end. The asymmetry now runs the other way, and backporting `in the middle` / `on its own` costs one `test.each`.

### The numbering half was already covered — thoroughly

The table says *"make `placeholder` zero-based → 3 cases fail, plus an in-list case"*. Measured:

```
base (#142)   51 failed
head (#144)   54 failed        ...the 3 new ones
```

Fifty-one tests already fail on the base — the insert placeholder lists, the in-list binding, `numbers its placeholders`, pagination, `mode: insensitive`. Zero-based numbering is one of the best-covered behaviours in the repo; it just is not covered *by name*.

Your prose already makes the correct argument for adding these anyway:

> a change to it fails somewhere else, with a message about a statement rather than about numbering

That is a legibility claim and it stands — 51 failures pointing at insert statements is a worse diagnosis than one saying `index 0 is $1`. The table is what muddles it, by presenting the new tests as the thing that catches the break. Worth restating as "51 → 54, and the three are the ones that name the cause", because the next person deciding whether this test is load-bearing will check the table rather than the prose.

Asserting Postgres against SQLite's answer for the same index is the right shape for it — "they differ" being the property is exactly the thing a shared implementation would slip past.

Nothing outstanding from me on this one.

PR #146, opened 2026-07-29
docs/orm.md listed groupBy and aggregate as out of scope

### LINKED ISSUE #68: ORM: decide `omit`, `distinct` and `cursor` — implement, or make the refusal permanent

### LINKED ISSUE #145: docs/orm.md lists groupBy and aggregate as out of scope, and documents both with examples
`docs/orm.md`'s **Not in scope** list opens with "Stated so you can plan around them rather than discover them", and contains:

> - **No `groupBy` or `aggregate`.** These land in [Raw SQL](#raw-sql), which exists so that "not implemented" has an answer rather than a shrug.

The same document has a `### Aggregates` section at line 162 and a `### groupBy` section at line 207, both with worked examples, and the `groupBy` one runs to four documented rules and a named Prisma divergence.

## Measured

Both compile:

```
aggregate   select count(*) as "_count_0" from "User"
groupBy     select "name" as "_by_0", count(*) as "_count_0" from "User" group by …
```

Both are also compared against Prisma by the differential harness — 15 distinct argument-key sets for `aggregate` and 8 for `groupBy`.

The bullet is stale. It predates the implementation and was never removed.

## Why this is worse than the failure #68 fixed

#68 was the mirror image: the docs said *Not in scope* while the runtime said *"yet"*. `read.test.ts` now guards that direction, and its comment names it — "which is how `docs/orm.md` came to list an argument under *Not in scope* while the runtime said 'yet'."

This drift is the other way, and it is worse in one specific respect. A runtime refusal that contradicts the docs gets corrected the moment someone runs the query. **"Refused, but actually works" has nothing to bump into** — the reader believes it, goes and writes raw SQL for a query the ORM would have answered, and never finds out.

## The rest of the list checks out

Verified against the runtime rather than assumed:

| entry | state |
| --- | --- |
| no identity map / unit of work | accurate |
| no lazy loading | accurate |
| no multi-field relations | accurate — refused as *"yet"* on all four correlated surfaces (`include`, relation filter, `_count`, `orderBy`) |
| no migrations / schema DSL | accurate |
| no `distinct` | accurate — `UnsupportedByDesignError` |
| no `cursor` | accurate — `UnsupportedByDesignError` |

## Acceptance

- The stale bullet removed.
- A test asserting the section never names an operation the ORM implements — the direction nothing guarded. Matched on the backticked spelling, since that is how the section names an API surface; prose like "no lazy loading" names a feature and has nothing to check against.
- The operation list in that test **verified** against the compiler's own predicates rather than asserted, so it cannot name something that does not exist.
- The opposite drift guarded too: `distinct` and `cursor` must stay listed while they are still refused.

nstfkc:
Closes #145. Stacked on #144.

`docs/orm.md`'s **Not in scope** list opens with *"Stated so you can plan around them rather than discover them"*, and contained:

> - **No `groupBy` or `aggregate`.** These land in [Raw SQL](#raw-sql), which exists so that "not implemented" has an answer rather than a shrug.

The same document has a `### Aggregates` section and a `### groupBy` section, both with worked examples — the second running to four documented rules and a named Prisma divergence.

## Measured

Both compile:

```
aggregate   select count(*) as "_count_0" from "User"
groupBy     select "name" as "_by_0", count(*) as "_count_0" from "User" group by …
```

and both are compared against Prisma by the differential harness — 15 distinct argument-key sets for `aggregate`, 8 for `groupBy`. The bullet predates the implementation and was never removed.

## Worse than the failure #68 fixed

#68 was the mirror image: docs saying *Not in scope* while the runtime said *"yet"*. `read.test.ts` guards that direction now, and its comment names it exactly.

This drift runs the other way, and is worse in one specific respect. A runtime refusal that contradicts the docs gets corrected the moment someone runs the query. **"Refused, but actually works" has nothing to bump into** — the reader believes it, goes and writes raw SQL for a query the ORM would have answered, and never finds out.

## The rest of the list checks out

Verified against the runtime rather than assumed:

| entry | state |
| --- | --- |
| no identity map / unit of work | accurate |
| no lazy loading | accurate |
| no multi-field relations | accurate — refused as *"yet"* on all four correlated surfaces |
| no migrations / schema DSL | accurate |
| no `distinct` | accurate — `UnsupportedByDesignError` |
| no `cursor` | accurate — `UnsupportedByDesignError` |

## The guard

The section may never name an operation the ORM implements. Matched on the **backticked** spelling, since that is how the section names an API surface — prose like "no lazy loading" names a *feature* and has nothing to check against.

The operation list in the test is **verified against the compiler's own predicates** rather than asserted, so it cannot quietly name something that does not exist. That check earned its keep immediately: it caught my own list, because `groupBy` has its own predicate rather than falling under the aggregate one.

And the opposite drift is guarded too — `distinct` and `cursor` must stay listed while they are still refused.

## Verification

| change | result |
| --- | --- |
| put the stale bullet back | `aggregate` and `groupBy` cases fail |
| delete `distinct` from the list | the still-listed case fails |

- 1260 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1121,8 +1121,6 @@ Stated so you can plan around them rather than discover them:
   from. This one *is* pending rather than declined; it needs a composite key comparison on both
   sides (a tuple `in` for the batched strategy, a conjunction for the lateral one).
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
-- **No `groupBy` or `aggregate`.** These land in [Raw SQL](#raw-sql), which exists so that "not
-  implemented" has an answer rather than a shrug.
 - **No `distinct`, and this one is deliberate rather than pending.** Prisma applies it **in
   memory** — its query log shows no `DISTINCT` at all, so `take` neither reduces the rows pulled
   nor paginates by group. Reproducing that faithfully would mean reading the whole result set and
--- a/packages/gemi/orm/docs-scope.test.ts
+++ b/packages/gemi/orm/docs-scope.test.ts
@@ -1,0 +1,102 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+import { isAggregateOperation } from "./compile/aggregate";
+import { isGroupByOperation } from "./compile/group-by";
+import { isReadOperation } from "./compile/read";
+import { isWriteOperation } from "./compile/write";
+import type { Operation } from "./plan";
+
+/**
+ * `docs/orm.md`'s **Not in scope** list, checked against the runtime.
+ *
+ * `read.test.ts` already guards one direction: `cursor` and `distinct` raise
+ * `UnsupportedByDesignError` rather than "yet", and its comment names why —
+ * "which is how `docs/orm.md` came to list an argument under *Not in scope*
+ * while the runtime said 'yet'."
+ *
+ * Nothing guarded the **other** direction, and the doc had drifted that way
+ * instead: it listed *"No `groupBy` or `aggregate`. These land in Raw SQL"*
+ * while the same document documents both with worked examples, the compiler
+ * implements both, and the differential harness compares both against Prisma.
+ *
+ * That is worse than the failure #68 fixed. "Refused, but actually works" tells
+ * a reader to go and write raw SQL for a query the ORM would have answered, and
+ * unlike a runtime refusal there is nothing to bump into that corrects them.
+ *
+ * The section is prose, so this does not try to parse the reasons — only to
+ * assert that it never names an operation the ORM implements.
+ */
+const DOC = join(import.meta.dirname, "../../../docs/orm.md");
+
+/**
+ * The operations the ORM implements. Listed here because `Operation` is a type
+ * and erased — but every entry is then **verified** against the compiler's own
+ * predicates, so this cannot quietly name something that does not exist.
+ */
+const IMPLEMENTED: Operation[] = [
+  "findUnique",
+  "findUniqueOrThrow",
+  "findFirst",
+  "findFirstOrThrow",
+  "findMany",
+  "count",
+  "aggregate",
+  "groupBy",
+  "create",
+  "createMany",
+  "update",
+  "updateMany",
+  "upsert",
+  "delete",
+  "deleteMany",
+];
+
+describe("docs/orm.md and the runtime agree", () => {
+  const source = readFileSync(DOC, "utf8");
+
+  const notInScope = (() => {
+    const start = source.indexOf("\n## Not in scope");
+    expect(start, "the Not in scope section moved or was renamed").toBeGreaterThan(0);
+    const end = source.indexOf("\n## ", start + 1);
+    return source.slice(start, end === -1 ? undefined : end);
+  })();
+
+  test("the list is found, and is not empty", () => {
+    expect(notInScope.length).toBeGreaterThan(200);
+  });
+
+  /** The list itself has to be real operations, or the check below means little. */
+  test.each(IMPLEMENTED)("%s is an operation the compiler recognises", (operation) => {
+    expect(
+      isReadOperation(operation) ||
+        isWriteOperation(operation) ||
+        isAggregateOperation(operation) ||
+        isGroupByOperation(operation),
+    ).toBe(true);
+  });
+
+  /**
+   * The assertion. An implemented operation named in this section is a
+   * contradiction a reader has no way to resolve — and the reader who believes
+   * it writes raw SQL for a query that already works.
+   *
+   * Matched on the backticked spelling, which is how the section names an API
+   * surface. Prose like "no lazy loading" is deliberately not matched: it names
+   * a feature rather than an operation, and there is nothing to check it
+   * against.
+   */
+  test.each(IMPLEMENTED)("%s is not listed as out of scope", (operation) => {
+    expect(notInScope).not.toContain(`\`${operation}\``);
+  });
+
+  /**
+   * ...and the two that genuinely are refused by design stay named, so removing
+   * them from the doc while leaving the runtime refusing is caught too. That is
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — `docs/orm.md` listed `groupBy` and `aggregate` as out of scope (closes #145)

Read the diff against its true base (`5ef75b3`, #144). 1260 tests pass, and both stated verifications reproduce:

```
put the stale bullet back      aggregate / groupBy "is not listed as out of scope"  FAIL
delete `distinct` from the list  "distinct is still listed"                          FAIL
```

**The asymmetry argument is the right one and I had not seen it put this way.** A runtime refusal contradicting the docs is self-correcting — someone runs the query. *"Refused, but actually works"* has nothing to bump into: the reader believes it, writes raw SQL for a query the ORM would have answered, and never learns otherwise. That is a better reason for this guard than "the docs should be accurate", and it is why guarding only the #68 direction left the worse half open.

**Verifying `IMPLEMENTED` against the compiler's own predicates is the detail that makes it hold up.** A hand-list of operation names would rot exactly the way the bullet did; a list that must satisfy `isReadOperation || isWriteOperation || isAggregateOperation || isGroupByOperation` cannot. That it caught your own list — `groupBy` needing its own predicate — is the kind of thing that only shows up if the check exists.

### The scoping to operations is load-bearing, and the comment does not say so

The stated rule is that the section may never name an implemented surface, matched on the backticked spelling. That rule is false in general. The section already contains, legitimately:

```
_count, include, orderBy, take, where, sql
```

— all implemented, four of them literally in `READ_ARGS.findMany`, every one present because the section is *describing* something else: "a relation you did not `include`", "use `take` with a `where` on the last row's sort key".

The guard works because `IMPLEMENTED` holds the fifteen `Operation` values, none of which is ever used descriptively here. The comment explains why unbackticked prose is not matched — *"it names a feature rather than an operation"* — but that is the argument for excluding `no lazy loading`, not the argument for excluding `` `include` ``, which is backticked and is an API surface.

It matters because the neighbouring test matches backticked **arguments** for the opposite purpose, so the file uses one syntactic signal two ways, and the symmetric-looking extension is wrong. Measured:

```js
const IMPLEMENTED_ARGS = ["where", "orderBy", "take", "skip", "select", "include", "omit", "_count"];
```
```
where     FAILS      orderBy  FAILS      take  FAILS
include   FAILS      _count   FAILS      5 failed | 36 passed
```

Five false positives immediately, on a section that is correct as written.

One sentence in the comment fixes it: operations are never named descriptively in this section, arguments routinely are, so the check is deliberately restricted to `Operation` and must not be widened to arguments without a different matching rule. Without that, the next person to extend this reaches for the obvious generalisation and has to work out from five red tests why it cannot be done — or, worse, edits the prose to satisfy it.

The rest of the *Not in scope* audit checks out against what I have measured across this series — `distinct` and `cursor` in `PERMANENTLY_REFUSED` with the stated reasoning, multi-field relations refused as "yet".

PR #148, opened 2026-07-29
StaleSchemaArtifactError had no test, and #119 never saw it

### LINKED ISSUE #147: StaleSchemaArtifactError has no test, and the coverage measurement never saw it
Verifying `docs/orm.md`'s **Dialects** section turned this up sideways. The section claims every model operation raises `UnsupportedDialectError` on MySQL and that `ormSupports` is exported — both accurate, both covered. But `UnsupportedDialectError` is declared in `orm/dialect/index.ts`, and I had enumerated the ORM's error classes out of `errors.ts`.

## The measurement was of a file, not of the ORM

`class \w*Error extends` across `orm/` finds **twenty** classes. #119's pass saw eighteen. The two it missed:

| class | file |
| --- | --- |
| `UnsupportedDialectError` | `orm/dialect/index.ts` |
| `StaleSchemaArtifactError` | `orm/schema.ts` |

`UnsupportedDialectError` is well covered — `dialect/unsupported.test.ts` has six tests, including one that the predicate and the resolver cannot disagree. **`StaleSchemaArtifactError` had no test at all.** Correction posted on #119.

## What it guards

The generated `index.ts` calls `assertSchemaArtifactVersion(ARTIFACT_VERSION)` at import time, so a mismatch is caught at registration rather than on the first query. The situation it exists for is ordinary: upgrading gemi without re-running `prisma generate`.

Two things were unasserted and both are load-bearing:

- **The comparison is `!==`, not `<`.** An artifact from a *newer* gemi is as much a mismatch as one from an older gemi. A check that only looked for "older" would accept an artifact this version cannot read.
- **The committed template artifact actually satisfies the handshake.** Every unit assertion about the guard passes just as well against a generator emitting a constant this package has never seen — the two packages have to be tied together or the handshake is only tested against itself.

## Acceptance

- Both directions of the mismatch, plus the accepted case.
- The message asserted to carry both versions and the command, since an author who hits this has a stale directory and nothing else to go on.
- The cross-package check: the template's `ARTIFACT_VERSION` equals `SCHEMA_ARTIFACT_VERSION`.
- Verified by making the comparison one-directional (2 cases fail) and by bumping the constant without regenerating (the cross-package case fails, which is exactly the situation the guard exists for).

nstfkc:
Closes #147. Stacked on #146.

Verifying `docs/orm.md`'s **Dialects** section turned this up sideways. The section claims every model operation raises `UnsupportedDialectError` on MySQL and that `ormSupports` is exported — **both accurate, both covered**. But `UnsupportedDialectError` is declared in `orm/dialect/index.ts`, and #119 enumerated the ORM's error classes out of `errors.ts`.

## The measurement was of a file, not of the ORM

`class \w*Error extends` across `orm/` finds **twenty** classes. #119's pass saw eighteen:

| class | file | covered? |
| --- | --- | --- |
| `UnsupportedDialectError` | `orm/dialect/index.ts` | yes — six tests, including that the predicate and the resolver cannot disagree |
| `StaleSchemaArtifactError` | `orm/schema.ts` | **no test at all** |

So "188 sites across 18 classes" was the count of a file. Correction posted on #119, with the fix for the next pass: enumerate by pattern across `orm/`, not by file.

## What it guards

The generated `index.ts` calls `assertSchemaArtifactVersion(ARTIFACT_VERSION)` at **import time**, so a mismatch is caught at registration rather than on the first query. The situation it exists for is ordinary — upgrading gemi without re-running `prisma generate`.

Two things were unasserted, and both are load-bearing:

**The comparison is `!==`, not `<`.** An artifact from a *newer* gemi is as much a mismatch as one from an older gemi. A check that only looked for "older" would accept an artifact this version cannot read.

**The committed template artifact actually satisfies the handshake.** Every unit assertion about the guard passes just as well against a generator emitting a constant this package has never seen — the two packages have to be tied together, or the handshake is only tested against itself.

## Verification

| break | result |
| --- | --- |
| make the comparison `<` instead of `!==` | 2 cases fail |
| bump `SCHEMA_ARTIFACT_VERSION` without regenerating | the cross-package case fails — exactly the situation the guard exists for |

- 1266 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/packages/gemi/orm/schema.test.ts
+++ b/packages/gemi/orm/schema.test.ts
@@ -1,0 +1,89 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+import {
+  assertSchemaArtifactVersion,
+  SCHEMA_ARTIFACT_VERSION,
+  StaleSchemaArtifactError,
+} from "./schema";
+
+/**
+ * The generated artifact's version handshake, which had no test.
+ *
+ * It was missed by the coverage measurement in #119 for a specific reason worth
+ * recording: that pass enumerated error classes out of `errors.ts`, and this
+ * one is declared in `schema.ts`. Two of the ORM's twenty live outside that
+ * file — this and `UnsupportedDialectError` — so "18 classes" was the count of
+ * a file rather than of the ORM. `UnsupportedDialectError` turned out to be
+ * well covered by `dialect/unsupported.test.ts`; this one was not covered at
+ * all.
+ *
+ * What it guards is ordinary rather than exotic: upgrading gemi without
+ * re-running `prisma generate`. The generated `index.ts` calls this with its own
+ * version literal at import time, so the mismatch is caught at registration —
+ * before any query — and the message says which way round the versions are and
+ * what to run.
+ */
+describe("the generated artifact's version", () => {
+  test("the current version is accepted", () => {
+    expect(() => assertSchemaArtifactVersion(SCHEMA_ARTIFACT_VERSION)).not.toThrow();
+  });
+
+  /**
+   * Both directions. An artifact from a *newer* gemi is as much a mismatch as
+   * one from an older gemi, and the check is `!==` rather than `<` precisely so
+   * that downgrading is caught too — a comparison that only looked for "older"
+   * would pass an artifact this version cannot read.
+   */
+  test.each([
+    ["older", SCHEMA_ARTIFACT_VERSION - 1],
+    ["newer", SCHEMA_ARTIFACT_VERSION + 1],
+    ["much older", 0],
+  ])("an artifact from a %s gemi is refused", (_label, version) => {
+    expect(() => assertSchemaArtifactVersion(version)).toThrow(StaleSchemaArtifactError);
+  });
+
+  /**
+   * The message is the whole diagnosis — an author who hits this has a stale
+   * directory and nothing else to go on — so it has to carry both versions and
+   * the command.
+   */
+  test("the message names both versions and what to run", () => {
+    let error: StaleSchemaArtifactError | null = null;
+    try {
+      assertSchemaArtifactVersion(7);
+    } catch (raised) {
+      error = raised as StaleSchemaArtifactError;
+    }
+
+    expect(error).toBeInstanceOf(StaleSchemaArtifactError);
+    expect(error!.name).toBe("StaleSchemaArtifactError");
+    // What was found, what is wanted, and the fix.
+    expect(error!.message).toContain("version 7");
+    expect(error!.message).toContain(`version ${SCHEMA_ARTIFACT_VERSION}`);
+    expect(error!.message).toContain("prisma generate");
+    expect(error!.message).toContain("app/models/generated");
+  });
+
+  /**
+   * ...and the template's committed artifact actually satisfies it.
+   *
+   * The unit assertions above would all pass with a generator that emitted a
+   * constant this file has never seen. This is the half that ties the two
+   * packages together — and it is the same handshake the generated `index.ts`
+   * performs at import time.
+   */
+  test("the committed artifact declares the version this gemi reads", () => {
+    const generated = readFileSync(
+      join(import.meta.dirname, "../../../templates/saas-starter/app/models/generated/schema.ts"),
+      "utf8",
+    );
+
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — `StaleSchemaArtifactError` had no test, and #119 never saw it (closes #147)

**First, my own correction.** My review of #119 independently "confirmed" 18 classes and 188 sites — by enumerating from `errors.ts`, which is the same boundary the pass used. Reproducing a measurement inside its own scope validates nothing, and that is exactly the failure I have been flagging in test design all series: a check that shares the assumption it is checking. Measured properly now:

```
class \w*Error extends, across orm/, non-test    20
  errors.ts                                      18
  schema.ts        StaleSchemaArtifactError       1
  dialect/index.ts UnsupportedDialectError        1
```

So the corrected figure for #119 is **190 sites across 20 classes**, and the *"0 unreached"* headline was really **1 unreached** — `UnsupportedDialectError` was already covered, `StaleSchemaArtifactError` was not. It is named in zero test files on the base.

Both sensitivity rows reproduce exactly:

```
!== changed to <                     an artifact from a newer gemi is refused   FAIL
                                     the message names both versions            FAIL
SCHEMA_ARTIFACT_VERSION bumped       the committed artifact declares …          FAIL
```

**The cross-package assertion is the idea worth keeping.** *"Every unit assertion about the guard passes just as well against a generator emitting a constant this package has never seen"* is right, and it is the same class of hazard as the differential harness comparing two clients against each other rather than against a fixture. A handshake tested only against itself is not tested. The second row above is what proves it fires.

The `!==` versus `<` point is the kind that reads as pedantry until stated: an artifact from a *newer* gemi is a mismatch, and a `<` check would import it and then read fields that may not mean what it thinks.

### The boundary moved from file to directory, and the directory is now the unexamined one

The fix recorded for the next pass is *"enumerate by pattern across `orm/`, not by file"*. Within `orm/` that pattern is complete — I checked for a class extending `Error` without the `Error` suffix and there is none, so nothing else hides from it.

One directory over there are two more:

```
database/dialect.ts   UnknownDatabaseUrlError    covered (database/dialect.test.ts)
database/dialect.ts   MissingDatabaseUrlError    named in no test file
```

`MissingDatabaseUrlError` is thrown from `DatabaseManager`'s constructor, and it has the same shape as the class this PR is about — deletable in silence:

```
guard removed, orm + database suites    1294 passed
```

I would not widen this PR for it. The point is that the correction replaces one scope with another, and `orm/` is a scope that was inherited rather than chosen — the same shape as #127's `.unsafe(` scan, which walks `orm/` while `facades/DB.ts` and `database/DatabaseManager.ts` also reach the driver. Twice now the boundary has been the thing that was wrong, and both times it was a directory picked for convenience. Worth stating in the recorded method as "across `orm/`, **and** `database/`, which owns the connection", so the next pass does not have to rediscover it a third time.

nstfkc:
## Correction to my own suggestion above

I ended that comment recommending the recorded method read *"across `orm/`, **and** `database/`, which owns the connection"*. That is right for one measurement and wrong for the other, and lumping them was the same kind of imprecision I was pointing at.

Measured on #149's head:

```
database/dialect.ts   UnknownDatabaseUrlError    exported from gemi/database, covered
database/dialect.ts   MissingDatabaseUrlError    exported from gemi/database, no test
docs claiming a complete error catalogue         orm.md only — "typed error from `gemi/orm`"
no Errors section under any other doc
```

So the two enumerations want different boundaries:

- **The coverage measurement** (#119, corrected by #147) should include `database/`. The question there is "does every error class have a test", and `MissingDatabaseUrlError` has none — deleting its guard leaves 1294 green across both suites. That one is a genuine gap, and it sits in the same constructor as the un-awaited `ready` I raised on #91.

- **The docs Errors guard** (#149) should **not**. Its claim is scoped by the sentence it enforces — *"Every failure is a typed error from `gemi/orm`"* — and there is no equivalent sentence for `gemi/database`, so a `database/` class is undocumented rather than mis-documented. Widening that guard would demand rows for classes no document promises.

Both classes are exported from `gemi/database`, so the `declared ⊆ exported` property I suggested on #149 already holds there too — it is only the test coverage that does not.

The general form, since this is the third time a boundary has been the thing that was wrong: the scope belongs to the *claim*, not to the directory. A coverage claim is about the code that ships; a documentation claim is about the surface the document names. Writing which one a given enumeration is answering would have settled all three cases without re-deriving them.

PR #150, opened 2026-07-29
The Errors table listed 12 of the ORM's 20 classes

### LINKED ISSUE #149: The Errors table lists 12 of the ORM's 20 classes, and omits the one the docs tell you to catch
`docs/orm.md`'s **Errors** section opens

> Every failure is a typed error from `gemi/orm`, not a driver string.

and then gives a table. A reader takes that as the catalogue — it is where you look to find out what there is to catch.

## Measured

All **20** error classes in `orm/` are exported from `gemi/orm`. The table lists **12**.

| exported but undocumented |
| --- |
| `UnsupportedByDesignError` |
| `UnsupportedDialectError` |
| `ReturningUnsupportedError` |
| `MissingRequiredValueError` |
| `DecodeError` |
| `MalformedRelationError` |
| `MissingModelSchemaError` |
| `UnregisteredRelationTargetError` |

Nothing is documented-but-missing, so the table is a subset rather than out of step.

## The costly one

`UnsupportedByDesignError`. The **Not in scope** section says the two deliberate refusals throw it, and that this is the point:

> which says *"and this is a decision rather than a gap"* rather than *"yet"* — so a refusal you can plan around reads differently from one that might lift next release.

That distinction is #68's entire subject, and acting on it means catching the subclass. The place a reader looks for the class did not mention it. The reader is told the difference matters, then not told how to observe it.

Two more are named elsewhere in the same document and absent here: `UnsupportedDialectError` (**Dialects**) and `ReturningUnsupportedError`, which is the same MySQL gap seen from the write path.

## Why it drifted, and the guard

The same way #145 did: the section was written once and the code kept moving. `StaleSchemaArtifactError` is *in* the table, so the table has been maintained — just not completely.

The guard derives the class list from **the module's own exports**, filtered by `value.prototype instanceof Error` rather than by name, so a class that stopped extending `Error` — and therefore stopped being catchable the documented way — drops out rather than being counted. There is no second list to keep in step, which is the failure mode the operations check in the same file had to work around by verifying its hardcoded list against the compiler's predicates.

## Acceptance

- The eight added, grouped so the registry/artifact family reads as one answer rather than three.
- A test asserting every exported `Error` subclass appears in the table.
- Verified in both directions: removing `UnsupportedByDesignError` from the table fails, and adding a new exported error class without documenting it fails — which is how this drifted in the first place.

nstfkc:
Closes #149. Stacked on #148.

`docs/orm.md`'s **Errors** section opens

> Every failure is a typed error from `gemi/orm`, not a driver string.

and then gives a table. A reader takes that as the catalogue — it is where you look to find out what there is to catch.

## Measured

All **20** error classes in `orm/` are exported from `gemi/orm`. The table listed **12**:

| exported but undocumented |
| --- |
| `UnsupportedByDesignError` |
| `UnsupportedDialectError` |
| `ReturningUnsupportedError` |
| `MissingRequiredValueError` |
| `DecodeError` |
| `MalformedRelationError` |
| `MissingModelSchemaError` |
| `UnregisteredRelationTargetError` |

Nothing was documented-but-missing, so the table was a *subset* rather than out of step.

## The costly one

`UnsupportedByDesignError`. The **Not in scope** section says the deliberate refusals throw it, and that this is the point:

> which says *"and this is a decision rather than a gap"* rather than *"yet"* — so a refusal you can plan around reads differently from one that might lift next release.

That distinction is #68's entire subject, and acting on it means catching the subclass. The place a reader looks for the class did not mention it — told the difference matters, then not told how to observe it.

Two more are named elsewhere in the same document and were absent here: `UnsupportedDialectError` (**Dialects**) and `ReturningUnsupportedError`, the same MySQL gap seen from the write path.

## The guard

It drifted the way #145 did — written once, code kept moving. `StaleSchemaArtifactError` *is* in the table, so it has been maintained, just not completely.

The check derives the class list from **the module's own exports**, filtered by `value.prototype instanceof Error` rather than by name — so a class that stopped extending `Error`, and therefore stopped being catchable the documented way, drops out rather than being counted. There is no second list to keep in step, which is the failure mode the operations check in the same file has to work around by verifying its hardcoded list against the compiler's predicates.

## Verification

| change | result |
| --- | --- |
| remove `UnsupportedByDesignError` from the table | its case fails |
| add a new exported error class without documenting it | its case fails — which is how this drifted originally |

- 1287 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -922,7 +922,13 @@ Every failure is a typed error from `gemi/orm`, not a driver string.
 | `ModelNotRegisteredError` / `UnregisteredPolicyClassError` | Registry problems (see [Setup](#your-model-class)). |
 | `RelationDepthExceededError` | An include tree past `MAX_RELATION_DEPTH`. |
 | `ParameterLimitError` | A statement exceeding the dialect's parameter ceiling. |
+| `MissingRequiredValueError` | A write leaving a required column with no value and no default. |
+| `DecodeError` | A column the driver returned that cannot be read as its declared type. |
+| `UnsupportedByDesignError` | A subclass of `UnsupportedQueryError` for the arguments under [Not in scope](#not-in-scope) — it says *decision*, where the parent says *not yet*. Catch this one to tell them apart. |
+| `UnsupportedDialectError` | A model operation on a dialect with no compiler — MySQL and MariaDB. See [Dialects](#dialects). |
+| `ReturningUnsupportedError` | A write on a dialect without `RETURNING`, which is the same gap seen from the write path. |
 | `StaleSchemaArtifactError` | Generated files predate the running gemi. Re-run `prisma generate`. |
+| `MalformedRelationError` / `MissingModelSchemaError` / `UnregisteredRelationTargetError` | The generated artifact and the registry disagree — the same family as the two above, and the same fix. |
 
 ## Performance
 
--- a/packages/gemi/orm/docs-scope.test.ts
+++ b/packages/gemi/orm/docs-scope.test.ts
@@ -6,6 +6,7 @@ import { isAggregateOperation } from "./compile/aggregate";
 import { isGroupByOperation } from "./compile/group-by";
 import { isReadOperation } from "./compile/read";
 import { isWriteOperation } from "./compile/write";
+import * as orm from "./index";
 import type { Operation } from "./plan";
 
 /**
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — the Errors table listed 12 of 20 (closes #149)

Read the diff against its true base (`fa2aede`, #148). 1287 tests pass. All 20 classes are exported from `gemi/orm` — that matches the 20 I enumerated on #147 exactly, so *declared* and *exported* are the same set today, which is worth knowing for the finding below.

**Deriving the list from the module's own exports is the right call**, and the contrast you draw with the operations check in the same file is the useful part: that one needs a hardcoded list verified against the compiler's predicates because `Operation` is a type and erased; this one has a runtime source, so there is nothing to keep in step. Filtering by `value.prototype instanceof Error` rather than by name is the detail that makes it a *catchability* check rather than a naming convention.

`UnsupportedByDesignError` is the right one to lead with. The **Not in scope** section tells a reader the decision/gap split is the point and that acting on it means catching the subclass, and the catalogue did not name the subclass. Removing its row reproduces:

```
UnsupportedByDesignError is in the table    FAIL
```

### The second sensitivity row holds only as literally worded

> add a new exported error class without documenting it → its case fails

True — with the word *exported* doing the work. `orm/index.ts` re-exports the error classes **by name, one per line**, so adding a class to `errors.ts` does not export it. Measured:

```
new class in errors.ts + added to index.ts, undocumented      FAILS
new class in errors.ts, not added to index.ts, undocumented   54 passed
class declared and thrown in schema.ts, never exported        54 passed
```

The invisible case is the likelier accident, because it is *one* omission rather than two. And it is worse than a documentation gap: a class that is thrown but not exported cannot be caught by `instanceof` at all, which is the section's own opening sentence — *"Every failure is a typed error from `gemi/orm`"* — being false rather than incomplete.

The guard asserts `exported ⊆ documented`. That sentence also claims `thrown ⊆ exported`, and nothing checks it. #147 is the evidence that it is reachable: two of the twenty live outside `errors.ts`, in `schema.ts` and `dialect/index.ts`, and each had to be remembered separately in `index.ts`. Both were. Nothing would have said so if they had not been.

Cheap to close, in this PR's own idiom and using #147's corrected enumeration:

```js
const declared = files.flatMap((f) =>
  [...read(f).matchAll(/^\s*(?:export\s+)?class (\w+) extends \w*Error\b/gm)].map((m) => m[1]),
);
expect(declared.filter((n) => !exported.has(n))).toEqual([]);
```
```
today                                    1 passed
a declared, thrown, unexported class     FAILS — names the class
```

Green now precisely because declared and exported are both 20, so it costs nothing to add and it pins the fact that makes the table's derivation trustworthy in the first place.

PR #152, opened 2026-07-29
A bare JSON scalar diverges from Prisma, not just between dialects

### LINKED ISSUE #151: A bare JSON scalar diverges from Prisma on Postgres, and the docs call it a dialect difference
`docs/orm.md`'s **`Json` columns** section says:

> A bare JSON number or boolean is refused on Postgres. … SQLite has no such limit. **This is the one shape where the two dialects disagree**, and it fails loudly rather than storing the wrong thing.

True, and not the sharp part.

## Measured

Added the shape to the differential table and ran it against Postgres 16:

```
User.create({"data":{"email":"…","metadata":42}}):
  prisma returned but gemi threw —
  column "metadata" is of type jsonb but expression is of type integer
```

**Prisma stores it. gemi raises.** So the divergence a caller meets is from *Prisma*, not between the two dialects — and parity with Prisma is this ORM's premise. A reader porting code hits it; a reader reading the docs is told only that SQLite and Postgres differ.

The same section's opening sentence made it worse, listing "a number" among the values a `Json` column "round-trips whatever you give it".

## The behaviour itself is right

`dialect/postgres.ts` sets out the trade at length, and I am not proposing to undo it. The encoder that accepted bare scalars serialised first, which stored `42` as the jsonb **string** `"42"` — confirmed there with `jsonb_typeof` — and only looked correct because the decoder re-parsed it on the way out. The value was wrong in the database the whole time. Raising beats that.

What was missing is that the trade was **invisible**. The differential table had objects, an empty object, an array and a JSON *string* — every Json shape where the two agree. The one shape that does not agree was the one nobody wrote.

## Why the test is a unit test

I wrote the end-to-end version first and it perturbed the suite: the harness's Postgres runs share one database, and a file that writes through both clients into it made two unrelated nested-write cases fail — reproducibly, and cleanly gone when the file was removed. Weakening a shared harness to hold a regression test for a documented decision is the wrong trade, so the assertion sits on the encoder, which is where the behaviour lives and what would have to change.

The Prisma-side half of the comparison is recorded above, where it was measured.

## Acceptance

- The docs corrected: the divergence is from Prisma, with the reason, and the opening sentence no longer promises a bare number round-trips.
- The encoder's behaviour pinned per shape on both dialects, including that a bare scalar reaches Postgres as its own type — which is *why* `jsonb` refuses it.
- The string case pinned, since not turning `"42"` into `42` is the corruption the raw binding exists to prevent.
- Verified by making the encoder serialise again: five cases fail.

nstfkc:
Closes #151. Stacked on #150.

`docs/orm.md`'s **`Json` columns** section said:

> A bare JSON number or boolean is refused on Postgres. … SQLite has no such limit. **This is the one shape where the two dialects disagree**, and it fails loudly rather than storing the wrong thing.

True, and not the sharp part.

## Measured

Added the shape to the differential table and ran it against Postgres 16:

```
User.create({"data":{"email":"…","metadata":42}}):
  prisma returned but gemi threw —
  column "metadata" is of type jsonb but expression is of type integer
```

**Prisma stores it. gemi raises.** The divergence a caller meets is from *Prisma*, not between the two dialects — and parity with Prisma is this ORM's premise. Someone porting code hits it; someone reading the docs is told only that SQLite and Postgres differ.

The section's opening sentence made it worse, listing "a number" among the values a `Json` column "round-trips whatever you give it".

## The behaviour is right, and is not changed here

`dialect/postgres.ts` sets out the trade at length. The encoder that accepted bare scalars serialised first, which stored `42` as the jsonb **string** `"42"` — confirmed there with `jsonb_typeof` — and only looked correct because the decoder re-parsed it on the way out. The value was wrong in the database the whole time. Raising beats that.

What was missing is that the trade was **invisible**. The differential table had objects, an empty object, an array and a JSON *string* — every Json shape where the two agree. The one shape that does not agree was the one nobody wrote.

## Why this is a unit test

A decision rather than convenience, and worth stating.

I wrote the end-to-end version first and it perturbed the suite. The harness's Postgres runs share one database, and a file that writes through both clients into it made two unrelated nested-write cases fail — reproducibly, and cleanly gone when the file was removed (865 passed without it, 867/2 with it). Weakening a shared harness to hold a regression test for a documented decision is the wrong trade, so the assertion sits on the encoder, which is where the behaviour lives and what would have to change. The Prisma half of the comparison is recorded in #151, where it was measured.

## What is pinned

- Every `Json` shape's encoding, on both dialects.
- That a bare scalar reaches Postgres **as its own type** — which is *why* `jsonb` refuses it. If that ever starts returning a string, the refusal is gone and the docs need changing with it.
- That `"42"` is not turned into `42` — the corruption the raw binding exists to prevent.
- That `null` stays `null` rather than becoming the JSON text `"null"`.

**Verified** by making the encoder serialise again: five cases fail.

- 1302 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).
- Template suite: 600 passed on SQLite, 865 on Postgres 16.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1067,16 +1067,22 @@ matters.
 
 ## `Json` columns
 
-A `Json` column round-trips whatever you give it: an object, an array, a string, a number, `null`.
-The value is stored as JSON and comes back as the same JavaScript value, on both dialects and
-whether it is read at the root or nested inside an `include`.
+A `Json` column round-trips whatever you give it: an object, an array, a string, `null` — and a
+number or boolean on SQLite, with the one exception below. The value is stored as JSON and comes
+back as the same JavaScript value, on both dialects and whether it is read at the root or nested
+inside an `include`.
 
 Two things are worth knowing:
 
 - **A bare JSON number or boolean is refused on Postgres.** `metadata: 42` raises rather than being
   stored, because the driver binds it as an integer and the column is `jsonb`. Wrap it — `{ value:
-  42 }` — or store it as a string. SQLite has no such limit. This is the one shape where the two
-  dialects disagree, and it fails loudly rather than storing the wrong thing.
+  42 }` — or store it as a string. SQLite has no such limit.
+
+  **This is the one shape where gemi diverges from Prisma**, which stores it on both dialects, so it
+  is worth knowing if you are porting code rather than writing it fresh. It is a trade rather than
+  an oversight: the encoder that accepted it serialised first, which stored `42` as the jsonb
+  *string* `"42"` — and only looked correct because the decoder re-parsed it on the way out, so the
+  value was wrong in the database the whole time. Failing loudly beats storing the wrong thing.
 - **A string is a string.** `metadata: "42"` stores the JSON string `"42"`, not the number, and
   `metadata: '{"a":1}'` stores that text as a string rather than as an object. If you want an
   object, pass an object.
--- a/packages/gemi/orm/json-encode.test.ts
+++ b/packages/gemi/orm/json-encode.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, test } from "vitest";
+
+import { PostgresDialect } from "./dialect/postgres";
+import { SqliteDialect } from "./dialect/sqlite";
+import type { FieldSchema } from "./schema";
+
+const postgres = new PostgresDialect();
+const sqlite = new SqliteDialect();
+
+const metadata = {
+  name: "metadata",
+  column: "metadata",
+  type: "Json",
+  nullable: true,
+  isId: false,
+  isUpdatedAt: false,
+} as FieldSchema;
+
+/**
+ * How a `Json` value is bound, and the one shape where that **diverges from
+ * Prisma**.
+ *
+ * `dialect/postgres.ts` hands `Json` over raw, for a measured reason: the
+ * encoder that serialised first stored `42` as the jsonb *string* `"42"` —
+ * confirmed with `jsonb_typeof` — and only looked correct because the decoder
+ * re-parsed it on the way out. Raising beats storing the wrong thing, and that
+ * trade is not what this file questions.
+ *
+ * What it pins is the consequence, which nothing asserted. Binding raw means a
+ * bare number or boolean reaches Postgres as its own type, and `jsonb` will not
+ * take it:
+ *
+ *     column "metadata" is of type jsonb but expression is of type integer
+ *
+ * **Prisma stores it.** Measured against Prisma 6.19 and Postgres 16 through the
+ * differential harness: `metadata: 42` succeeds under Prisma on both dialects
+ * and raises under gemi on Postgres. So the divergence a caller meets is from
+ * *Prisma*, not between the two dialects as `docs/orm.md` described it — which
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — a bare JSON scalar diverges from Prisma, not just between dialects (closes #151)

Read the diff against its true base (`ac5bd52`, #150). 1302 tests pass.

I checked #151's premises before this landed, and all three hold: the opening sentence did list "a number" among the values a `Json` column round-trips, three lines above a bullet saying it is refused; and the seed carries exactly four Json shapes — `{plan…}`, `{}`, `[]`, `"42"` — every one a shape where the two agree. The characterisation was exact.

**The reframing is the right one.** "The two dialects disagree" is true and answers a question nobody asks; "gemi diverges from Prisma, and here is the trade" is the sentence a porting reader needs, and putting the `jsonb_typeof` finding in the docs rather than only in a commit message is what makes it read as a decision instead of a bug.

**The `why a unit test` section is the part I would keep.** Reporting that the end-to-end version made two unrelated nested-write cases fail — with the counts, and that removing the file cleaned it up — is more useful than silently choosing the unit test, because the perturbation is a fact about the harness that outlives this change.

Sensitivity re-measured. Making the encoder serialise again:

```
8 failed | 1294 passed
  object / array / string / bare number / bare boolean handed unchanged — postgres
  a number reaches postgres as its own type
  a boolean reaches postgres as its own type
  a JSON string is not turned into a number
```

Eight, not the five claimed — the third sensitivity count in this run that reads low (#141 said 5 existing where 6 fail, #143 said 3 where 51 already did). Always in the PR's favour, and always the number a later reader checks the test against.

### The comment block in the differential table describes the approach that was abandoned

```js
/**
 * Bare JSON scalars — the one shape `docs/orm.md` says the two dialects disagree on…
 *
 * The harness compares failure *kind*, so these assert the interesting thing
 * without the test needing to know which dialect it is on: whatever Postgres
 * does, gemi and Prisma have to do the same.
 */
// The two bare-scalar shapes are *not* here: on Postgres gemi raises where
// Prisma succeeds… pinned as such in `json-scalars.test.ts` instead.
["create with a wrapped JSON number", "create", { … metadata: { value: 42 } }],
```

Three problems, all in eighteen lines:

- The block is titled **Bare JSON scalars** and sits above a case that is not one.
- Its closing paragraph gives the argument for *including* them — *"whatever Postgres does, gemi and Prisma have to do the same"* — and the next line says they are excluded precisely because gemi and Prisma do **not** do the same. Two comments in direct contradiction, three lines apart.
- **`json-scalars.test.ts` does not exist.** The file is `json-encode.test.ts`, and nothing else in the repo mentions that name.

The fossil is the same failure this series has been correcting all the way through — #133's reason attached to the line that was not performing the exclusion, #137's "allocation choice" describing a flag that guards a scope. A reader who trusts the first paragraph goes looking for bare-scalar rows, then follows a pointer to a file that is not there.

What the case that *is* there needs is its own two lines: the wrapped number is the documented workaround, both clients agree on it, and pinning it is what stops the advice in the docs going stale. That is a good reason, and it is not the one currently written above it.

PR #154, opened 2026-07-29
Where a nested scope lands in the lateral subquery

### LINKED ISSUE #153: Nothing asserts where a nested scope lands in the lateral subquery
`docs/orm.md` lists six paths that carry a child's policies, and gives the folded relation its **own row**:

> | a folded relation (the `lateral` strategy) | the scope lands *inside* the subquery |

It is a separate row because it is a separate mechanism. `policy.test.ts` states the hazard:

> A nested read under the batched strategy gets its policies by recursing through the child's own `$exec`. A lateral join has no such call, so the child's policies would never be consulted and the subquery would be unscoped — **a cross-tenant read through a new door.**

## What is covered, and what is not

Covered: `applyNestedPolicies` is unit-tested to produce a scoped *argument* tree, and the template's policy suite runs against Postgres — where lateral is the default — so a scope that vanished outright fails there as wrong rows.

Not covered: **the placement.** A predicate that moved from the subquery to the outer `where` still filters. The failure is not "no filtering", it is filtering the *parents* by a child's column instead of the children — which surfaces downstream as a row-count difference in a database test rather than as the shape change it is.

## Measured

The scope does land inside, correctly:

```sql
left join lateral (
  select coalesce(json_agg(json_build_object(…)), '[]'::json) as "data"
  from "Account"
  where "Account"."userId" = "User"."id" and "Account"."organizationId" = $1
) as "__lat_accounts" on true
```

Correlation and scope together, inside; nothing about `Account` in the outer statement. No defect.

## Acceptance

- The subquery body extracted by paren balancing rather than by regex, and asserted to hold both the correlation and the scope.
- The outer statement asserted to hold **neither** — which is the half that catches a move rather than a drop.
- A caller's own filter on the same relation asserted to land there too, conjoined rather than replacing the scope — the shape #105 fixed for nested writes, checked here on the read path.
- `asSystem` asserted to leave nothing to place, so "the scope is inside" cannot pass by a scope that is unconditionally present.
- Verified by two independent breakages: making `applyNestedPolicies` a no-op, and dropping the subquery's filter.

nstfkc:
Closes #153. Stacked on #152.

`docs/orm.md` lists six paths that carry a child's policies, and gives the folded relation its **own row**:

> | a folded relation (the `lateral` strategy) | the scope lands *inside* the subquery |

A separate row because it is a separate mechanism. `policy.test.ts` states the hazard:

> A nested read under the batched strategy gets its policies by recursing through the child's own `$exec`. A lateral join has no such call, so the child's policies would never be consulted and the subquery would be unscoped — **a cross-tenant read through a new door.**

## What was covered, and what was not

**Covered:** `applyNestedPolicies` is unit-tested to produce a scoped *argument* tree, and the template's policy suite runs against Postgres — where lateral is the default — so a scope that vanished outright fails there as wrong rows.

**Not covered: the placement.** A predicate that moved from the subquery to the outer `where` still filters. The failure is not "no filtering" — it is filtering the *parents* by a child's column instead of the children, which surfaces downstream as a row-count difference in a database test rather than as the shape change it is.

## Measured

```sql
left join lateral (
  select coalesce(json_agg(json_build_object(…)), '[]'::json) as "data"
  from "Account"
  where "Account"."userId" = "User"."id" and "Account"."organizationId" = $1
) as "__lat_accounts" on true
```

Correlation and scope together, inside; nothing about `Account` in the outer statement. **No defect.**

## The assertions

- The subquery body extracted by **paren balancing**, not regex — the body contains parentheses of its own.
- The outer statement asserted to contain neither the child's scope column nor a `where` at all. **That half is what catches a move rather than a drop.**
- A caller's own filter asserted to land there too, *conjoined* rather than replacing the scope — the shape #105 fixed for nested writes, checked here on the read path.
- `asSystem` asserted to leave nothing to place, so "the scope is inside" cannot pass by a scope that is unconditionally present.

## Verification

Two independent breakages:

| break | result |
| --- | --- |
| `applyNestedPolicies` made a no-op | 3 cases fail |
| the subquery's filter dropped | 2 cases fail |

My first attempt at the second mutation deleted a `node.where` that had already been consumed and changed nothing — worth recording, because a sensitivity check that does not move the needle proves nothing about the test.

- 1307 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).
- Template suite: 600 passed on SQLite, 865 on Postgres 16.

### DIFF
--- a/packages/gemi/orm/lateral-scope.test.ts
+++ b/packages/gemi/orm/lateral-scope.test.ts
@@ -1,0 +1,151 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { lateralStrategy } from "./compile/lateral";
+import { compileRead } from "./compile/read";
+import { PostgresDialect } from "./dialect/postgres";
+import { account, organization, user } from "./fixtures";
+import { applyNestedPolicies, type ModelPolicy } from "./policy";
+import * as registry from "./registry";
+
+const postgres = new PostgresDialect();
+
+/**
+ * **Where a nested scope lands in the emitted SQL under the lateral strategy.**
+ *
+ * `docs/orm.md` lists six paths that carry a child's policies and gives the
+ * folded relation its own row — "the scope lands *inside* the subquery" —
+ * because it is a distinct mechanism rather than another instance of the same
+ * one. `policy.test.ts` says why: a batched nested read gets its policies by
+ * recursing through the child's own `$exec`, and "a lateral join has no such
+ * call, so the child's policies would never be consulted and the subquery would
+ * be unscoped — a cross-tenant read through a new door."
+ *
+ * Two things were tested, and one was not.
+ *
+ * `applyNestedPolicies` is unit-tested to produce a scoped *argument* tree, and
+ * the template's policy suite runs against Postgres, where lateral is the
+ * default — so an outright dropped scope fails there already, as wrong rows.
+ *
+ * What nothing asserted is the **placement**. A predicate that moved from the
+ * subquery to the outer `where` still filters, so the failure is not "no
+ * filtering" — it is filtering the *parents* by a child's column instead of the
+ * children. That reads downstream as a row-count difference in a database test
+ * rather than as the shape change it is, and this is the cheaper place to catch
+ * it.
+ *
+ * Compile-level and dialect-only: no database, because the question is what SQL
+ * comes out.
+ */
+describe("a nested scope under the lateral strategy", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    for (const [name, schema] of [
+      ["User", user],
+      ["Account", account],
+      ["Organization", organization],
+    ] as const) {
+      registry.register(name, class { static $schema = schema });
+    }
+  });
+
+  const tenant: ModelPolicy = {
+    scope: () => ({ organizationId: 7 }),
+    onCreate: (_context, data) => data,
+  };
+
+  /** The arg tree as `$exec` hands it to the compiler: policies already applied. */
+  const scoped = (args: unknown) =>
+    applyNestedPolicies(
+      user as never,
+      args as never,
+      { id: 1 },
+      false,
+      ((model: string) =>
+        model === "Account"
+          ? { policies: [tenant], schema: { name: "Account", relations: {} } }
+          : undefined) as never,
+    );
+
+  const compiled = (args: unknown) =>
+    compileRead(user, "findMany", scoped(args) as never, postgres, lateralStrategy).text;
+
+  /** The subquery body, from `left join lateral (` to its matching paren. */
+  const subquery = (sql: string) => {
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — where a nested scope lands in the lateral subquery (closes #153)

Read the diff against its true base (`026afdd`, #152). 1307 tests pass, and both sensitivity rows are exact:

```
applyNestedPolicies made a no-op    3 failed
the subquery's filter dropped       2 failed
```

**The distinction is the right one to pin.** "A predicate that moved from the subquery to the outer `where` still filters" — so the failure is filtering *parents* by a child's column rather than filtering nothing, and it surfaces downstream as a row count rather than as the shape change it is. That is why the template's Postgres run does not cover it, and the argument is worth having in the file.

Checked the helper rather than trusting it, since the assertions rest on it — at depth 1 it splits 466 characters of subquery from 344 of outer statement, both with substance, so neither `toContain` nor `not.toContain` is passing on an empty string. The ` =` suffix on the inner assertion is load-bearing too: `"Account"."organizationId"` appears in the subquery's `json_build_object` as a projection, and without the suffix the test would pass on the selected column rather than the predicate.

**Two things I would keep.** *"the policy rewrote the argument tree, or this asserts nothing"* is a vacuity guard written in from the start rather than added after one was found — four of those turned up by accident earlier in this series, and this is the first PR to open with one. And recording that the first mutation attempt deleted an already-consumed `node.where` and changed nothing: a sensitivity check that does not move the needle proves nothing, and saying so is what stops the next reader trusting it.

### The distinction is unavailable one level down

Every case is depth 1, and the helper cannot go deeper. `subquery()` takes the first `left join lateral (` and paren-balances to its close — but a nested fold puts the grandchild's lateral *inside* the parent's. Measured on `include: { accounts: { include: { user: true } } }` with a policy on both models:

```
laterals emitted                    2
outer lateral starts at             316
inner lateral starts at             738
subquery() span                     1093 chars — contains the inner lateral
```

So at depth 2, "the scope is inside the subquery" is satisfied whether the grandchild's scope sits in its own lateral or in its parent's. The second is the exact failure this PR exists to catch: a `User` predicate in the `Account` subquery filters *accounts* by a column of the user they belong to, and every assertion here still passes.

**No defect today** — I checked before writing this:

```
"User"."globalRole" =           at 1274   (inside the inner lateral)
"Account"."organizationId" =    at 1377   (the outer lateral's own where)
```

Both correctly placed, and the argument tree is scoped at both levels.

The fix is small because the aliases are already unique — `suffixed("__lat_", as, depth, index)` exists to stop an inner node shadowing an ancestor. Selecting the span by alias rather than by first-match makes the helper depth-agnostic, and one depth-2 case would then assert what the depth-1 ones do: that each scope is inside *its own* subquery and in no other.

Worth doing here rather than later, because the helper is the thing a follow-up would copy, and first-match-plus-balance reads as correct until there are two.

PR #156, opened 2026-07-29
DB.sql joins the ambient transaction on SQLite

### LINKED ISSUE #155: DB.sql joins the ambient transaction on SQLite, and the docs say it never does
`docs/orm.md`'s **Raw SQL** section states it plainly and unconditionally:

> `DB.sql` is unchanged and still Bun's own tagged template… It does **not** join the ambient transaction — use `DB.query` when that matters.

## Measured, on both dialects

Inside a `Model.transaction` that then throws, a `DB.sql` update against a seeded row:

| dialect | after the rollback |
| --- | --- |
| Postgres 16 | `"escaped"` — the write **survived** |
| SQLite | `null` — the write **rolled back** |

A control in the same test confirms the statement itself works outside a transaction on both, so this is the transaction and not the statement.

The mechanism is `DatabaseManager`: SQLite gets a **single connection**, so a `DB.sql` statement issued inside an open transaction is on that same connection and therefore inside it. Postgres has a pool, so the statement takes a different connection and escapes.

So the doc is right about Postgres and wrong about SQLite.

## The direction is the bad one

A `DB.sql` write inside a transaction **rolls back correctly in development on SQLite and survives the rollback in production on Postgres.**

That is the same hazard the **Transactions** section already documents at length for a different door — "a bug that passes in development on SQLite and takes out the transaction in production on Postgres" — and this one is reached without doing anything the docs warn against. The reader who follows the current text does the opposite of what it intends: told `DB.sql` never joins, they test on SQLite, watch it roll back, and conclude the docs are being over-careful.

## Why it was not caught

`raw-sql.test.ts` has one `DB.sql` test — "DB.sql is untouched, and still Bun's own tagged template" — which asserts the template *works*. The transactional claim, which is the reason the paragraph exists, had nothing.

## Acceptance

- The claim corrected to say which dialect does what, with the reason, and the awkward direction named.
- Both behaviours pinned per dialect, asserted rather than skipped on the inconvenient one — the point is that they differ.
- `DB.execute` asserted to roll back on **both**, so the first test cannot pass by way of a transaction that never opened.

nstfkc:
Closes #155. Stacked on #154.

`docs/orm.md`'s **Raw SQL** section stated it plainly and unconditionally:

> `DB.sql` is unchanged and still Bun's own tagged template… It does **not** join the ambient transaction — use `DB.query` when that matters.

## Measured

Inside a `Model.transaction` that then throws, a `DB.sql` update against a seeded row:

| dialect | after the rollback |
| --- | --- |
| Postgres 16 | `"escaped"` — the write **survived** |
| SQLite | `null` — the write **rolled back** |

A control in the same run confirms the statement works outside a transaction on both, so this is the transaction and not the statement.

The mechanism is `DatabaseManager`: SQLite gets a **single connection**, so a `DB.sql` statement issued inside an open transaction is on that same connection and therefore inside it. Postgres has a pool, so the statement takes a different connection and escapes.

**The doc is right about Postgres and wrong about SQLite.**

## The direction is the bad one

A `DB.sql` write inside a transaction **rolls back correctly in development on SQLite and survives the rollback in production on Postgres.**

That is the same hazard the **Transactions** section already documents at length for a different door — "a bug that passes in development on SQLite and takes out the transaction in production on Postgres" — reached here without doing anything the docs warn against.

The current text makes it worse rather than neutral: a reader told `DB.sql` never joins will test on SQLite, watch it roll back, and conclude the warning is over-careful.

## Why it was not caught

`raw-sql.test.ts` has exactly one `DB.sql` test — *"DB.sql is untouched, and still Bun's own tagged template"* — which asserts the template **works**. The transactional claim, which is the reason that paragraph exists, had nothing.

## The change

- The doc says which dialect does what, with the reason, and names the awkward direction.
- Both behaviours pinned **per dialect**, asserted rather than skipped on the inconvenient one — the point is that they differ.
- `DB.execute` asserted to roll back on **both**, so the first test cannot pass by way of a transaction that never opened.

- 1307 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).
- Template suite: 602 passed on SQLite, 869 on Postgres 16.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1062,8 +1062,17 @@ same applies to identifiers and sort directions, which cannot be parameters in S
 is the same for those, and it is about where the value came from, not what it is.
 
 `DB.sql` is unchanged and still Bun's own tagged template, for a single self-contained statement
-that needs none of this. It does **not** join the ambient transaction — use `DB.query` when that
-matters.
+that needs none of this. **Whether it joins the ambient transaction depends on the dialect**, so use
+`DB.query` when that matters:
+
+- On **Postgres** it does not. The pool hands it a different connection, so a `Model.transaction`
+  that rolls back leaves a `DB.sql` write behind.
+- On **SQLite** it does, because `DatabaseManager` gives SQLite a single connection and the
+  statement is therefore inside the open transaction.
+
+The direction is the awkward one: a `DB.sql` write inside a transaction rolls back correctly in
+development on SQLite and survives the rollback in production on Postgres. `DB.query` and
+`DB.execute` join the transaction on both.
 
 ## `Json` columns
 
--- a/templates/saas-starter/app/models/raw-sql.test.ts
+++ b/templates/saas-starter/app/models/raw-sql.test.ts
@@ -369,6 +369,87 @@ function suite(label: string, url?: string) {
       expect([...rows]).toHaveLength(1);
     });
 
+    /**
+     * **`DB.sql` does not join the ambient transaction**, which the docs state
+     * and nothing asserted — only that the template still works.
+     *
+     * It is the consequential half. A write issued through it inside a
+     * `Model.transaction` is on a different connection, so a rollback does not
+     * take it back: the transaction reports failure and the row is still there.
+     *
+     * Worth pinning in both directions rather than as a caveat, because the two
+     * calls look interchangeable at the call site and only one of them is safe
+     * to mix with a transaction.
+     */
+    /**
+     * **`DB.sql` and the ambient transaction — and the two dialects disagree.**
+     *
+     * `docs/orm.md` says it plainly and unconditionally: "It does **not** join
+     * the ambient transaction — use `DB.query` when that matters."
+     *
+     * Measured, that holds on Postgres and **not** on SQLite. `DatabaseManager`
+     * gives SQLite a single connection, so a `DB.sql` statement issued inside a
+     * `Model.transaction` is on the same connection and therefore inside the
+     * transaction — it rolls back with everything else. Postgres has a pool, so
+     * the statement takes a different connection and escapes.
+     *
+     * The direction matters. A developer who tests on SQLite watches a `DB.sql`
+     * write roll back correctly and ships it; on Postgres the rollback leaves
+     * that write behind. It is the same hazard the transactions section already
+     * warns about for a caught statement — "passes in development on SQLite and
+     * takes out the transaction in production on Postgres" — reached by a
+     * different door.
+     *
+     * Both halves are asserted per dialect rather than skipping the one that is
+     * inconvenient: the point is that they genuinely differ.
+     */
+    test("whether a DB.sql write survives a rollback depends on the dialect", async () => {
+      await expect(
+        Model.transaction(async () => {
+          await DB.sql`update "User" set "name" = ${"escaped"} where "email" = ${"ada@x.test"}`;
+          throw new Error("rolled back");
+        }),
+      ).rejects.toThrow("rolled back");
+
+      const after = await DB.query<{ name: string | null }>(
+        sql`select "name" from "User" where "email" = ${"ada@x.test"}`,
+      );
+
+      if (url) {
+        // Postgres: a different connection from the pool, so the transaction
+        // never contained it and the rollback did not reach it.
+        expect(after[0].name).toBe("escaped");
+        await DB.execute(
+          sql`update "User" set "name" = null where "email" = ${"ada@x.test"}`,
+        );
+      } else {
+        // SQLite: one connection, so it was inside the transaction after all.
+        expect(after[0].name).toBeNull();
+      }
+    });
+
+    /**
+     * ...and `DB.query` / `DB.execute`, which the docs point to instead, roll
+     * back on **both**. Without this the test above says only that something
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — `DB.sql` joins the ambient transaction on SQLite (closes #155)

Read the diff against its true base (`b0e86ba`, #154). 18 pass on SQLite; I have no Postgres 16, so that column of your table is yours.

**The SQLite half reproduces and the finding is real.** The doc said "does not join the ambient transaction" unconditionally, and on SQLite it does. The mechanism checks out structurally — `DatabaseManager` gives SQLite a single `SQL` instance, so a statement issued inside an open transaction is on the same connection by construction.

**The direction is the part worth having in the docs**, and the rewrite says it: rolls back correctly in development on SQLite, survives the rollback in production on Postgres. That is the same hazard the transactions section already documents for a caught statement, reached without doing anything the docs warn against — and the old text actively primed the reader to dismiss it, since testing on SQLite confirms the reassuring half.

### The control the description claims is not in the file

> A control in the same run confirms the statement works outside a transaction on both

There are exactly two `DB.sql` call sites in `raw-sql.test.ts`: a `select` at line 367, and the `update` inside the rolled-back transaction at 409. No `DB.sql` write outside a transaction exists.

It matters more on SQLite than the existing `DB.execute` control covers. That control is aimed at the Postgres branch — its comment says so, *"a transaction that never opened would satisfy it too"* — and Postgres asserts a **positive** value, `toBe("escaped")`. SQLite asserts `toBeNull()`, and the seeded row starts with no name:

```ts
await User.create({ data: { email: "ada@x.test", globalRole: 0 } });   // name is null
...
expect(after[0].name).toBeNull();
```

So the assertion carrying the novel claim is an absence assertion against a column that is absent to begin with. It is satisfied by the write rolling back — and equally by the write never taking effect, for any reason that leaves no error: a quoting difference, a statement the driver accepts and discards, a future change to how `DB.sql` binds.

I want to be precise about what I did and did not show. I made the update match no row and the suite still passed — but that mutates the test rather than the system, so it demonstrates nothing a control could catch either. What I can show is that the ambiguity is unresolved by anything in the file, and that resolving it is free:

```ts
test("a DB.sql write outside a transaction lands", async () => {
  await DB.sql`update "User" set "name" = ${"landed"} where "email" = ${"ada@x.test"}`;
  expect((await DB.query(sql`select "name" from "User" where "email" = ${"ada@x.test"}`))[0].name)
    .toBe("landed");
});
```
```
19 passed | 1 skipped
```

Green on this head, so `DB.sql` writes do land — which is exactly the fact the SQLite assertion needs and does not currently have.

### The abandoned framing is still in the file

Two doc comments sit stacked above the one test, at lines 373 and 385. The first opens:

> **`DB.sql` does not join the ambient transaction**, which the docs state and nothing asserted… A write issued through it inside a `Model.transaction` is on a different connection, so a rollback does not take it back: the transaction reports failure and the row is still there.

That is the claim this PR disproves, stated as fact, twelve lines above the block that corrects it. Both were added in this diff.

Second PR running with a fossil of the abandoned approach left in place — #151 left a pointer to `json-scalars.test.ts`, a file that does not exist. Worth a pass over the comment before merge rather than a note: the whole value of these blocks is that a reader trusts them, and a reader who stops at the first one leaves with the belief the PR exists to remove.

nstfkc:
## Follow-up — the dangling-pointer half of this is mechanically checkable, and there is a second one

I flagged the `json-scalars.test.ts` pointer on #151 and the stale block here as a pattern. The pointer half does not need judgement, so I checked it across the repo.

Naively — every backticked `*.ts` reference in a comment — it flags 20, and 18 are legitimate: a user's `app/config/broadcast.ts`, placeholders like `compile/x.ts` in prose, the hypothetical `Post.ts` in the bootstrap-cycle explanation. Those are the repo talking about somebody else's file.

Narrow it to references ending `.test.ts` or `.test-d.ts` — the subset where the repo is talking about *itself* — and it separates cleanly:

```
481 files scanned
  packages/gemi/orm/Model.ts:426                          `policy.plan-cache.test.ts`
  templates/saas-starter/app/models/writes.differential.test.ts:151  `json-scalars.test.ts`
-- 2 dangling references, 0 false positives
```

**The first is new, and it is in source rather than a test.** `Model.ts:426` sits in the middle of the comment explaining why scoping has to run *before* `getOrCompile`:

> two users with the same query shape and different scopes would share one plan, and one of them would run the other's SQL. That is a cross-tenant data leak produced by a caching bug… and `policy.plan-cache.test.ts` pins both halves

No such file. The reader most likely to follow that pointer is the one about to reorder those two lines, which is the reason the paragraph exists.

**The claim itself is pinned** — I checked before reporting. `plan.discrimination.test.ts:757` carries exactly it: *"two whose scopes differ in shape must not"* share a plan, alongside the same-shape-one-plan cases. So this is a wrong name rather than a missing test, and the fix is one identifier.

Worth having as a standing check rather than a fix, in the shape #127 and #131 already use. It needs no allow-list today, which is the unusual part — the narrowing to test files does the work an allow-list would otherwise do, and the eighteen legitimate references stay out of it by construction rather than by being listed. A comment that names a sibling test is a claim about this repository, and it is the only kind of file reference that can be checked without knowing what the reader's app looks like.

PR #158, opened 2026-07-29
Merging #95 and #100 made the multi-field-relations entry stale

### LINKED ISSUE #157: Merging #95 and #100 made the multi-field-relations entry stale
`docs/orm.md`'s **Not in scope** list says:

> **No multi-field relations.** `@relation(fields: [a, b], references: [c, d])` is refused wherever a relation is correlated — `include` under either strategy, a relation filter, `_count`, an `orderBy` through a relation, and nested writes.

**Four of those five stopped being true** when #95 ("Composite relations: every read surface joins on all of the fields") and #100 ("A composite `include` is one plan on Postgres, through `unnest`") were merged into `feat/orm`. The entry did not move with them.

## Measured on `feat/orm` at b4884c3

`ledgerEntry.ledger` joins on `tenantId` **and** `ledgerCode`:

| surface | state |
| --- | --- |
| `include`, batched | works — the plan carries `parentFields: ["tenantId", "code"]` |
| `include`, lateral | works — both columns inside the subquery |
| relation filter | works — `exists (… where "_r0"."tenantId" = … and "_r0"."ledgerCode" = …)` |
| `_count` | works — both columns in the correlated subquery |
| `orderBy` through a relation | works — both columns |
| **nested write** | **still refused**, by name: *"entries joins on 2 fields"* |

So the entry should be narrowed to the nested-write case, not deleted.

## Why this one is different from #145

#145 was the same *shape* — a *Not in scope* entry naming something implemented — but it had been wrong since it was written. This one **became** wrong, during this session, because the boundary moved underneath it.

That is the case a test catches and review does not: nothing in #95 or #100 touched this paragraph, so there was no diff for a reviewer to notice it against.

It is also not caught by the guard added in #146, which asserts the section never names an implemented **operation**. "Multi-field relations" is a capability, not one of the fifteen, so there is nothing for that check to match on. A capability needs its own assertion, which is what this adds.

## Acceptance

- The entry narrowed to nested writes, with the read surfaces described as working.
- Each of the five surfaces asserted to correlate on **both** fields — one field would still compile and would silently return the wrong rows, which is the failure the original refusal existed to prevent.
- The nested-write refusal asserted **by name**, so when it lands this test fails and prompts moving the entry again.
- Verified by truncating the composite link to its first field: four of the six cases fail.

nstfkc:
Closes #157. **Based on `feat/orm` directly**, not on the review stack — it corrects something the recent merges caused, so it should be mergeable on its own.

`docs/orm.md`'s **Not in scope** list says:

> **No multi-field relations.** … is refused wherever a relation is correlated — `include` under either strategy, a relation filter, `_count`, an `orderBy` through a relation, and nested writes.

**Four of those five stopped being true** when #95 and #100 merged. The entry did not move with them.

## Measured on `feat/orm` at b4884c3

`ledgerEntry.ledger` joins on `tenantId` **and** `ledgerCode`:

| surface | state |
| --- | --- |
| `include`, batched | works — plan carries `parentFields: ["tenantId", "code"]` |
| `include`, lateral | works — both columns inside the subquery |
| relation filter | works — `exists (… "_r0"."tenantId" = … and "_r0"."ledgerCode" = …)` |
| `_count` | works — both columns |
| `orderBy` through a relation | works — both columns |
| **nested write** | **still refused**, by name: *"entries joins on 2 fields"* |

So the entry is **narrowed**, not deleted.

## Why this differs from #145

#145 was the same shape — a *Not in scope* entry naming something implemented — but it had been wrong since it was written. **This one became wrong during the session**, because the boundary moved underneath it.

That is the case a test catches and review does not: nothing in #95 or #100 touched this paragraph, so there was no diff to notice it against.

It is also **not** caught by #146's guard, which asserts the section never names an implemented *operation*. A multi-field relation is a capability rather than one of the fifteen, so there is nothing for that check to match. A capability needs its own assertion, which is what this adds.

## The test

Each surface asserts **both** fields. One field would still compile and would silently return the wrong rows — the exact failure the original refusal existed to prevent.

The nested-write refusal is asserted **by name**, so when it lands this test fails and prompts moving the entry again.

**Verified** by truncating the composite link to its first field: four of the six cases fail.

- 1032 unit tests, `tsc --noEmit` clean, `oxlint` clean (its two warnings are pre-existing in `where.ts`).
- Template suite on this base: 571 passed on SQLite, 427 passed on Postgres 16 — no failures on either.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1081,12 +1081,12 @@ Stated so you can plan around them rather than discover them:
   an identity map is worse than none.
 - **No lazy loading.** A relation you did not `include` is absent, not a proxy that queries when
   touched.
-- **No multi-field relations.** `@relation(fields: [a, b], references: [c, d])` is refused wherever
-  a relation is correlated — `include` under either strategy, a relation filter, `_count`, an
-  `orderBy` through a relation, and nested writes — rather than joining on the first field and
-  returning plausible wrong rows. The error names the fields on both sides and the operation it came
-  from. This one *is* pending rather than declined; it needs a composite key comparison on both
-  sides (a tuple `in` for the batched strategy, a conjunction for the lateral one).
+- **No multi-field relations in a nested write.** `@relation(fields: [a, b], references: [c, d])`
+  now works everywhere a relation is *read* — `include` under either strategy, a relation filter,
+  `_count`, and an `orderBy` through a relation all correlate on every field. What is still refused
+  is reaching one from a nested write, and the error names the fields on both sides and the
+  operation it came from, rather than joining on the first field and writing plausible wrong rows.
+  Pending rather than declined.
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
 - **No `groupBy` or `aggregate`.** These land in [Raw SQL](#raw-sql), which exists so that "not
   implemented" has an answer rather than a shrug.
--- a/packages/gemi/orm/composite-surfaces.test.ts
+++ b/packages/gemi/orm/composite-surfaces.test.ts
@@ -1,0 +1,137 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { lateralStrategy } from "./compile/lateral";
+import { compileRead } from "./compile/read";
+import { compileWrite } from "./compile/write";
+import { PostgresDialect } from "./dialect/postgres";
+import { SqliteDialect } from "./dialect/sqlite";
+import { UnsupportedQueryError } from "./errors";
+import { ledger, ledgerEntry } from "./fixtures";
+import * as registry from "./registry";
+
+const sqlite = new SqliteDialect();
+const postgres = new PostgresDialect();
+
+/**
+ * Which surfaces a **multi-field relation** reaches, stated as a table.
+ *
+ * `docs/orm.md`'s *Not in scope* list said a composite relation "is refused
+ * wherever a relation is correlated — `include` under either strategy, a
+ * relation filter, `_count`, an `orderBy` through a relation, and nested
+ * writes". Four of those five stopped being true when #95 and #100 landed, and
+ * the entry did not move with them.
+ *
+ * That is the direction #145 was about: a *Not in scope* list is written so a
+ * reader can plan around it, and one that names something already implemented
+ * sends them to write raw SQL for a query that works. It drifted here because
+ * the boundary moved rather than because the text was wrong when written —
+ * which is exactly the case a test catches and review does not.
+ *
+ * `ledgerEntry.ledger` joins on `tenantId` **and** `ledgerCode`, so every case
+ * below asserts *both* columns appear. One would be enough for the query to
+ * compile and would silently return the wrong rows, which is the failure the
+ * refusal existed to prevent.
+ */
+describe("a relation that joins on two fields", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    registry.register("Ledger", class { static $schema = ledger });
+    registry.register("LedgerEntry", class { static $schema = ledgerEntry });
+  });
+
+  /** Both sides of the composite key, as the compiler spells them. */
+  const correlatesOnBoth = (sql: string) => {
+    expect(sql).toMatch(/"tenantId"/);
+    expect(sql).toMatch(/"ledgerCode"/);
+  };
+
+  test("include, batched", () => {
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — merging #95 and #100 made the multi-field-relations entry stale (closes #157)

Based on `feat/orm` rather than the stack, so I reviewed it against `b4884c3` — its own diff is the two files claimed. Note the branch is now ~8 commits behind its base: #96, #101 and #102 have landed since, and all three touch `nested-writes.ts`, where the one surface this PR says is *still refused* lives.

**That turns out not to matter, which is worth stating because it is the first thing to check for a PR whose subject is staleness.** Applying the new test file to `feat/orm` at `279778a`:

```
composite-surfaces.test.ts    6 passed
```

So all five read surfaces still work and the nested-write refusal survived the three merges. The correction is accurate on the base it will merge into, not only on the one it was measured against.

Truncating the link to its first field reproduces exactly:

```
4 failed | 2 passed
```

### The two that survive truncation are the finding

`correlatesOnBoth` matches the bare identifiers:

```ts
expect(sql).toMatch(/"tenantId"/);
expect(sql).toMatch(/"ledgerCode"/);
```

`LedgerEntry` *projects* `ledgerCode` as a column, so that is satisfied by the select list whether or not the join correlates on it. The two surfaces that pass under truncation — `include, lateral` and `a relation filter` — are the two whose SQL names the column somewhere other than the predicate.

Which is the failure the file's own comment describes: *"One would be enough for the query to compile and would silently return the wrong rows."* On those two surfaces, that is what the test currently permits.

The fix is the same one #153 uses — assert the predicate, not the identifier — with one wrinkle I got wrong first and had to measure:

```
lateral          where "LedgerEntry"."tenantId" = "Ledger"."tenantId"
                   and "LedgerEntry"."ledgerCode" = "Ledger"."code"
relation filter  exists (… where "_r0"."tenantId" = "LedgerEntry"."tenantId"
                   and "_r0"."code" = "LedgerEntry"."ledgerCode")
```

The second field is spelled from whichever side is on the left — `ledgerCode` in one, `code` in the other — so a single `/"ledgerCode"\s*=/` fails on head. Allowing either:

```ts
expect(sql).toMatch(/"tenantId"\s*=/);
expect(sql).toMatch(/"(?:ledgerCode|code)"\s*=/);
```
```
head         6 passed
truncated    6 failed        (was 4)
```

All six surfaces then detect the truncation, which is what the six were written to do.

### Two smaller notes

**The #145 comparison is a forward reference.** `docs-scope.test.ts` is not on `feat/orm` — it arrives with the stack — so on this base there is no guard at all, and "not caught by #145's guard" is true of the stack rather than checkable here. Worth a word, since the point of basing this on `feat/orm` is that it stands alone.

**Three lines below the corrected entry**, the same list still carries *"**No `groupBy` or `aggregate`.** These land in Raw SQL"* — the bullet #145 exists to remove, still wrong on `feat/orm`, and wrong in precisely the way this PR's own rationale describes. #145 handles it on the stack, so this is a judgement call rather than a defect; but a reader of `feat/orm` today sees one entry corrected beside one that is not, and this PR is already editing that list.

nstfkc:
## Follow-up — this and the stack's docs guards are compatible, in either order

Since this is deliberately based off-stack and edits the same **Not in scope** list that #145 and #149 add guards for, I checked whether the two collide. They do not.

Composing the stack's `docs/orm.md` (which carries #145's removal of the `groupBy`/`aggregate` bullet and #149's completed Errors table) with this PR's rewritten multi-field bullet, then running both guards:

```
docs-scope.test.ts    54 passed
```

So the edits are disjoint in substance: #145 removes a bullet this PR does not touch, #149 adds table rows, and this PR narrows a bullet neither guard inspects — it names a capability, and both guards match operation and error-class names. Whichever lands first, the other still passes.

Worth recording because the guards are the thing that would otherwise have to be re-derived at merge time, and "based on `feat/orm` so it is mergeable on its own" is only true if that holds.

My first attempt at this check was worthless and I nearly posted it: I combined this PR's `docs/orm.md` with the stack's *test* file and got eight failures, every one of them the stack's own doc changes being absent rather than a real interaction. Composing a branch's tests with another branch's sources measures the gap between the branches, not the merge — the same scoping error I have already had to retract twice on this stack, in a new costume.

The gap the PR identifies stands either way: neither guard covers a capability, which is why `composite-surfaces.test.ts` is the right shape for it rather than another row in the operations list.

PR #160, opened 2026-07-29
feat/orm does not typecheck: one call site missed when `detail` became required

### LINKED ISSUE #159: feat/orm does not typecheck: #102 made `detail` required and one call site was missed
`feat/orm` at 279778a fails `tsc`:

```
orm/compile/group-by.ts(674,11): error TS2554: Expected 4 arguments, but got 3.
```

Verified on a clean checkout of `origin/feat/orm`, not on a working branch.

## How it got there

A semantic merge conflict — the kind neither PR can see.

- **#92** added `throw new UnsupportedQueryError(key, schema.name, op)` in `groupBy`'s `assertArgs`, when `detail` was optional.
- **#102** made `detail` **required** — the second half of #61 — and fixed the call sites on its own branch.

Each typechecked alone. `group-by.ts` was not touched by #102 and `errors.ts` was not touched by #92, so there was no textual conflict to resolve and no diff for either reviewer to look at. The break exists only in the combination.

## Why it matters beyond the error

Every branch stacked on `feat/orm` inherits it, so `tsc` is now failing for reasons unrelated to whatever a contributor is working on — which is exactly the condition under which people start ignoring it.

I found it while merging `feat/orm` into #103 to clear that PR's conflict: the merge produced one `tsc` error, and both files involved turned out byte-identical to `feat/orm`.

## The fix

Supply the detail, in the wording `read.ts` already uses for the same refusal:

```ts
`${op} takes ${[...GROUP_BY_ARGS].sort().join(", ")}.`
```

That is the case #61's second half was actually about — an argument the operation does not accept is the path a typo arrives on, and naming the set is the whole of the answer.

## Acceptance

- `tsc` clean on `feat/orm`.
- A test on the **message**, not only the signature. `tsc` catches the arity; nothing would have caught a `detail` that said nothing useful, which is the failure #102 set out to prevent.

nstfkc:
Closes #159. **Based on `feat/orm` directly** — it fixes the integration branch, so it should not sit behind the review stack.

`feat/orm` at 279778a fails `tsc`:

```
orm/compile/group-by.ts(674,11): error TS2554: Expected 4 arguments, but got 3.
```

Verified on a **clean checkout of `origin/feat/orm`**, not on a working branch.

## How it got there

A semantic merge conflict — the kind neither PR can see.

| | |
| --- | --- |
| **#92** | added `throw new UnsupportedQueryError(key, schema.name, op)` in `groupBy`'s `assertArgs`, while `detail` was optional |
| **#102** | made `detail` **required** — the second half of #61 — and fixed the call sites on its own branch |

Each typechecked alone. #102 never touched `group-by.ts` and #92 never touched `errors.ts`, so there was no textual conflict to resolve and **no diff for either reviewer to look at**. The break exists only in the combination.

## Why it matters beyond the error

Every branch stacked on `feat/orm` inherits it, so `tsc` now fails for reasons unrelated to whatever a contributor is working on — which is exactly the condition under which people stop reading it.

I found it while merging `feat/orm` into #103 to clear that PR's conflict: the merge produced exactly one `tsc` error, and both files involved turned out **byte-identical to `feat/orm`**, which is what pointed at the base rather than at the merge.

## The fix

The wording `read.ts` already uses for the same refusal:

```ts
`${op} takes ${[...GROUP_BY_ARGS].sort().join(", ")}.`
```

That is the case #61's second half was actually about — an argument the operation does not accept is the path a typo arrives on, and naming the set is the whole of the answer.

## The test asserts the message, not the arity

`tsc` catches the signature. Nothing would have caught a `detail` that said *nothing useful*, which is the failure #102 set out to prevent — so the test checks that the refusal names the argument and the set.

- `tsc --noEmit` **clean** on this branch, 1044 unit tests, `oxlint` clean.
- Template suite: 595 passed on SQLite.

### DIFF
--- a/packages/gemi/orm/compile/group-by.test.ts
+++ b/packages/gemi/orm/compile/group-by.test.ts
@@ -373,4 +373,32 @@ describe("shaping", () => {
   test("no groups is an empty array, not null", () => {
     expect(plan({ by: ["globalRole"], _count: true }).shape([])).toEqual([]);
   });
+
+  /**
+   * An argument `groupBy` does not take, refused **with the set it does**.
+   *
+   * This one call site was left with three arguments when #102 made `detail`
+   * required, so `feat/orm` stopped typechecking the moment that merged
+   * alongside `groupBy`. Neither PR could see it: #92 added the call, #102
+   * changed the signature, and each typechecked on its own branch.
+   *
+   * `tsc` is what catches the signature; this is what catches the *message*,
+   * which is the thing #61's second half was actually about — an argument the
+   * operation does not accept is the path a typo takes, and naming the set is
+   * the whole of the answer.
+   */
+  test("an unknown argument names the ones groupBy takes", () => {
+    let message = "";
+    try {
+      plan({ by: ["globalRole"], _count: true, nope: 1 } as never);
+    } catch (error) {
+      message = (error as Error).message;
+    }
+
+    expect(message).toContain("'nope'");
+    // The set, sorted, exactly as `read.ts` words the same refusal.
+    expect(message).toMatch(/groupBy takes .*\bby\b.*/);
+    expect(message).toContain("having");
+    expect(message).toContain("orderBy");
+  });
 });
--- a/packages/gemi/orm/compile/group-by.ts
+++ b/packages/gemi/orm/compile/group-by.ts
@@ -671,7 +671,15 @@ function assertArgs(schema: ModelSchema, op: string, args: any): void {
   for (const key of Object.keys(args).sort()) {
     if (args[key] === undefined) continue;
     if (GROUP_BY_ARGS.has(key)) continue;
-    throw new UnsupportedQueryError(key, schema.name, op);
+    // The same wording `read.ts` uses for an argument an operation does not
+    // take, and for the reason #61's second half gives: this is the path a
+    // typo arrives on, so naming the set is the whole of the answer.
+    throw new UnsupportedQueryError(
+      key,
+      schema.name,
+      op,
+      `${op} takes ${[...GROUP_BY_ARGS].sort().join(", ")}.`,
+    );
   }
 
   if (args.by === undefined) {


nstfkc:
## Review — `feat/orm` does not typecheck (closes #159)

Reproduced on a clean checkout of `origin/feat/orm` at `279778a`, exactly one error:

```
orm/compile/group-by.ts(674,11): error TS2554: Expected 4 arguments, but got 3.
```

And on this branch: **0 errors**, 1044 tests. I also ran the base's suite — `feat/orm` passes 1032 — so the breakage really is type-level only, which the description says and which is worth having confirmed rather than assumed.

**The diagnosis is the valuable part.** #92 added the call while `detail` was optional; #102 made it required and fixed the sites on its own branch. Neither touched the other's file, so there was no textual conflict and no diff for either reviewer to read. That is a class of failure that is invisible to every check in this repository *by construction* — it does not exist in either branch.

**Asserting the message rather than the arity is right.** `tsc` owns the signature, so a test that counted arguments would duplicate the compiler; what nothing owned is a `detail` that satisfies the type and says nothing useful, which is the failure #102 existed to prevent. Naming the accepted set is the same answer `read.ts` gives, so the two refusals now read alike.

### Nothing checks the integration branch

`.github/workflows/` holds `docs.yml` — push, docs deploy — and `publish.yml`, manual. Neither runs `tsc` or the suite, and neither triggers on `pull_request`. So:

- every "tsc clean" in every PR body in this series is a claim about that branch in isolation;
- `feat/orm` itself is checked by nobody;
- and this defect sat there until someone merged `feat/orm` into #103 for an unrelated conflict and read one error carefully enough to notice both files were byte-identical to the base.

That last sentence is the argument. It took an accident and a careful reader to surface a one-line type error on the branch everything else is stacked on — and the PR is right that the real cost is not the error but that `tsc` stops being worth running once it fails for reasons unrelated to your work.

This series has added roughly a dozen tests to catch drift *within* a branch: the plan-key invariants, the architecture seams, the docs guards, the coverage measurements. The one failure class none of them can see is the one that only exists after a merge, and it is the cheapest to check — `bunx tsc --noEmit` plus the suite, on push to `feat/orm` and on `pull_request`. The methodology this series has been applying to test coverage says measure the thing rather than assume it; here the thing has never been measured except by hand, once, by luck.

Worth filing separately rather than widening this PR — it fixes the instance correctly and should land as it is.

nstfkc:
## Follow-up on `92e48a8` — the merge is clean, and the base is still broken

The new commit is a base refresh (`origin/feat/orm` merged in through #106). Given that this PR exists because of a semantic merge conflict, the merge is worth checking rather than assuming:

```
fix/orm-groupby-detail @ 92e48a8    tsc clean, 1048 passed
```

No new breakage from the merge itself.

**`feat/orm` is another matter.** It has moved on to `1fc5d6f`, and the error is still there:

```
origin/feat/orm @ 1fc5d6f
  orm/compile/group-by.ts(674,11): error TS2554: Expected 4 arguments, but got 3.
  1121 tests passed
```

The line is unchanged — `throw new UnsupportedQueryError(key, schema.name, op)`, three arguments.

Since #159 was filed against `279778a`, **ten pull requests have merged onto that branch**: #103, #104, #105, #106, #107, #109, #111, #113, #115, #117. Every one of them onto a base that does not typecheck, and none of them saw it, because each was measured on its own branch.

That is the argument in my review above, no longer as a mechanism but as a count. It is also why this PR is behind again — six of those ten landed after its last sync — so the one-line fix is now chasing a branch it was written to unblock.

Nothing here changes my read of the change itself; it still looks right and should land. The measurement just makes the follow-up sharper: a `tsc --noEmit` on push to `feat/orm` would have failed ten times by now, each time naming the one file, and the first failure would have cost a minute.

PR #162, opened 2026-07-29
CI: typecheck, tests and lint, on the merge result as well as the proposal

### LINKED ISSUE #161: Nothing runs typecheck or tests on any branch, which is why #159 survived six merges
`feat/orm` stopped typechecking at #102 and is still broken now, at fea6fbb — across **#101 through #109**, six merges.

Measured on a clean checkout of the current head:

| check | result |
| --- | --- |
| `tsc --noEmit` | **fails** — `group-by.ts(674,11): Expected 4 arguments, but got 3` |
| unit tests | 1088 pass |
| template tests | 604 pass, 32 skipped |
| `oxlint` | 1 warning, 0 errors |

**Everything green except the one thing that was broken.** That is why nobody noticed: a contributor running the suite sees a clean board.

## The repository runs no checks

`.github/workflows/` holds `docs.yml` (Pages deploy) and `publish.yml` (npm publish). Nothing runs `tsc`, `vitest` or `oxlint` on a push or a pull request.

There is also no `typecheck` script — the command exists only as a raw `bunx tsc --noEmit -p tsconfig.json` that a contributor has to know.

## Why a per-PR check would not have been enough on its own

#159 is a **semantic** merge conflict. #92 added a three-argument `UnsupportedQueryError` while `detail` was optional; #102 made `detail` required and fixed the call sites on its own branch. Neither touched the other's file — no textual conflict, no diff for either reviewer, and each branch typechecked alone.

It only exists once both are on the same branch, so the check has to run **on the result of merging**, not only on the proposal. That is the argument for running on pushes to the integration branches as well as on pull requests.

## Proposed, and it is the maintainer's call

A single job doing what I have been doing by hand each iteration: install, build the binaries (the generator is resolved by bin name, and `prisma generate` needs it), generate the template artifact, **typecheck**, unit tests, template tests, lint.

Two things stated rather than hidden:

- **It will fail on `feat/orm` until #160 merges.** That is the point rather than a snag — the branch is broken and nothing currently says so.
- **Postgres is not included.** The differential harness runs against it when `TEST_POSTGRES_URL` is set and skips loudly otherwise, so a run without it is honest about covering SQLite only. A service container is a sensible next step and a separate decision.

`packageManager` says `bun@1.1.3`, which is well behind what the repo is developed against, so the workflow takes `latest` rather than pinning to something stale. Worth reconciling separately.

## Acceptance

- A workflow running the sequence above on pull requests and on pushes to `main` and `feat/orm`.
- A `typecheck` script, so the command has one spelling for CI and for people.

nstfkc:
Closes #161. Based on `feat/orm` directly.

**This one is a proposal rather than a fix** — it adds a workflow, which is a decision about how the repository is run, not just about the ORM. The evidence for it is concrete, but the call is yours.

## The evidence

`feat/orm` stopped typechecking at #102 and is still broken at fea6fbb — across **#101 through #109**, six merges. On a clean checkout of the current head:

| check | result |
| --- | --- |
| `tsc --noEmit` | **fails** — `group-by.ts(674,11): Expected 4 arguments, but got 3` |
| unit tests | 1088 pass |
| template tests | 604 pass, 32 skipped |
| `oxlint` | 1 warning, 0 errors |

**Everything green except the one thing that was broken.** That is why it survived: a contributor running the suite sees a clean board.

`.github/workflows/` currently holds `docs.yml` and `publish.yml`. Nothing runs `tsc`, `vitest` or `oxlint` on a push or a pull request.

## Why per-PR checking would not have been enough

#159 is a **semantic** merge conflict. #92 added a three-argument `UnsupportedQueryError` while `detail` was optional; #102 made `detail` required and fixed the call sites on its own branch. Neither touched the other's file — no textual conflict, no diff for either reviewer, and **each branch typechecked alone**.

It exists only once both are on the same branch. So the check has to run on the *result of merging*, which is why this triggers on pushes to `main` and `feat/orm` as well as on pull requests.

## What it runs

The sequence I have been running by hand each iteration: install → build the binaries (the generator is resolved by bin name, and `prisma generate` needs it) → generate the template artifact → **typecheck** → unit tests → template tests → lint.

Also adds a `typecheck` script, so the command has one spelling for CI and for people rather than being a raw `bunx` invocation you have to know.

## Two things stated rather than hidden

- **It will fail on `feat/orm` until #160 merges.** That is the point rather than a snag — the branch is broken and nothing currently says so.
- **Postgres is not included.** The differential harness runs against it when `TEST_POSTGRES_URL` is set and skips loudly otherwise, so a run without it is honest about covering SQLite only. A service container is a sensible next step and a separate decision.

`packageManager` says `bun@1.1.3`, well behind what the repo is developed against, so the workflow takes `latest` rather than pinning to something stale — worth reconciling separately.

## Verification

Every step was run locally against this branch. `bun run typecheck` reproduces the `feat/orm` failure exactly; `bun run lint` exits 0; the two suites pass at 1088 and 604. The workflow file was checked for tab characters and step structure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,76 @@
+name: CI
+
+# Typecheck, tests and lint for the ORM and the template that exercises it.
+#
+# **Why this exists.** `feat/orm` stopped typechecking at #102 and stayed broken
+# across six merges (#101–#109) with every test green the whole time. The break
+# was a *semantic* merge conflict: #92 added a three-argument
+# `UnsupportedQueryError` while `detail` was optional, #102 made `detail`
+# required and fixed the call sites on its own branch. Neither touched the
+# other's file, so there was no textual conflict, no diff for a reviewer to look
+# at, and nothing that runs on the result of merging them.
+#
+# That is a class a per-PR check cannot cover on its own — it only appears once
+# both are on the same branch — which is why this runs on pushes to the
+# integration branches as well as on pull requests.
+#
+# Postgres is deliberately not here yet. The differential harness runs against
+# it when `TEST_POSTGRES_URL` is set and skips loudly otherwise, so a run
+# without it is honest about covering SQLite only. Adding a service container is
+# a sensible next step and a separate decision.
+on:
+  pull_request:
+  push:
+    branches: [main, feat/orm]
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          # Not pinned to `packageManager` (bun@1.1.3), which is behind what the
+          # repo is actually developed against.
+          bun-version: latest
+
+      - name: Install
+        run: bun install --frozen-lockfile
+
+      # The ORM generator is resolved by bin name, and in this repo that symlink
+      # points at built output — so `prisma generate` below needs it to exist.
+      - name: Build the binaries
+        run: bun run build:bin
+        working-directory: packages/gemi
+
+      # The template's committed artifact is what its suite runs against, and
+      # `generated.test.ts` asserts a second run produces a zero-line diff.
+      - name: Generate the template's model artifact
+        run: bunx prisma generate
+        working-directory: templates/saas-starter
+
+      # The step that would have caught #159. Kept before the tests, because a
+      # type error is cheaper to read than a suite that still passes around it.
+      - name: Typecheck
+        run: bun run typecheck
+        working-directory: packages/gemi
+
+      - name: Unit tests
+        run: bun --bun vitest run orm
+        working-directory: packages/gemi
+
+      - name: Template tests
+        run: bun --bun vitest run app/models
+        working-directory: templates/saas-starter
+        env:
+          TZ: UTC
+
+      - name: Lint
+        run: bun run lint
+        working-directory: packages/gemi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,8 +61,17 @@ jobs:
         run: bun run typecheck
         working-directory: packages/gemi
 
+      # `orm database`, not `orm`. The package has 47 test files and only 32
+      # have "orm" in their path — `database/DatabaseManager.test.ts` is the
+      # ORM's own SQLite foreign-key pragma test and would have been skipped by
+      # the narrower filter.
+      #
+      # Not the whole package: `app/`'s router tests fail on `feat/orm` today
+      # and are unrelated to this. Widening to everything would put CI red for a
+      # reason nobody here can act on, which is how a red board stops being
+      # read. Reported separately.
       - name: Unit tests
-        run: bun --bun vitest run orm
+        run: bun --bun vitest run orm database
         working-directory: packages/gemi
 
       - name: Template tests
--- a/packages/gemi/package.json
+++ b/packages/gemi/package.json
@@ -41,6 +41,7 @@
   },
   "scripts": {
     "lint": "oxlint",
+    "typecheck": "tsc --noEmit -p tsconfig.json",
     "build": "NODE_ENV=production bun run build:core && bun run build:bin && bun run build:types && bun run build:client && bun run build:plugin && bun run build:client-types",
     "build:core": "bun ./scripts/build.ts",
     "build:bin": "bun build --outdir=./dist/bin --target=bun --external=vite --external=react-dom --external=react/jsx-runtime --external=bun --external=sharp --sourcemap ./bin/gemi.ts ./bin/orm-generator.ts && bun ./scripts/prepare-bin.ts",


nstfkc:
**Corrected the filter, and found the reason to.**

The job ran `bun --bun vitest run orm`. That matches **32 of the package's 47 test files** — and among the 15 it skipped is `database/DatabaseManager.test.ts`, which is the ORM's own SQLite foreign-key pragma test. The job I proposed would have gone green while never running it.

Now `vitest run orm database`: 34 files, 1233 tests, passing.

**Deliberately not widened to the whole package.** `bun --bun vitest run` in `packages/gemi` is currently **7 failed | 1246 passed** — all seven in `app/`, which holds superseded copies of three router modules that `services/router/` also implements. Nothing imports the `app/` ones. Running everything would put CI red for a reason nobody reviewing an ORM change can act on, and a board that is red for reasons nobody can act on stops being read. Filed as #163; the filter can go once that is resolved.

Also merged current `feat/orm` into the branch. Re-verified on the new head: `typecheck` still reproduces the #159 failure (expected until #160 lands), unit tests 1233, template tests 607, lint clean.

nstfkc:
## Review — CI: typecheck, tests and lint (closes #161)

Read the diff against its merge-base (`16af1a3`) — two files, `+77`. Verified the parts that decide whether the workflow does what it says.

**The core argument is right and the trigger set follows from it.** A semantic merge conflict exists only in the combination, so a `pull_request`-only workflow cannot see it; running on pushes to the integration branches is what makes this cover #159's class rather than only its instance. Stating that in the workflow file, where the next person to prune triggers will read it, is the right place for it.

Verified locally on this branch:

```
bun run typecheck   orm/compile/group-by.ts(674,11): error TS2554   (reproduces exactly)
bun run lint        exit 0
template suite      604 passed | 32 skipped
```

### The generate step defeats the test it cites

The workflow runs `bunx prisma generate` before the suites, with this reason:

> The template's committed artifact is what its suite runs against, and `generated.test.ts` asserts a second run produces a zero-line diff.

That test snapshots the working tree *before its own* regeneration, not `HEAD`:

```ts
const before = snapshot();
execFileSync("bunx", ["prisma", "generate"], …);
expect(snapshot()).toEqual(before);
```

So a stale committed artifact is repaired by the workflow step before the test starts, and the test then compares the repaired file against itself. Measured, with a hand edit standing in for staleness:

```
test alone                             1 failed  — "running prisma generate again produces a zero-line diff"
prisma generate, then the test         2 passed
```

The step turns the one check on the committed artifact into a no-op, and the sentence justifying it is the thing it breaks.

**And it is not needed.** On a clean checkout of this commit, with the binaries built and nothing generated:

```
604 passed | 32 skipped
```

The same numbers the description reports. The artifact is committed, so the suite has it; `build:bin` is genuinely required — `generated.test.ts` asserts the bin is linked and runs the generator itself — but the generate step only removes the evidence.

### Two smaller things

**The `oxlint` row does not match.** The table says *1 warning, 0 errors*; measured on this branch:

```
packages/gemi   89 warnings, 0 errors   (73 of them eslint(no-unused-vars))
repo root      116 warnings, 0 errors
```

It exits 0 either way, so the workflow step passes and nothing is blocked — which is why this is a note rather than a finding. But every PR in this series has reported "oxlint clean, its two warnings pre-existing in `where.ts`", and that has not matched measurement for a while. Worth reconciling once, since a number nobody can reproduce stops being read.

**The merge count is already out of date.** *"#101 through #109, six merges"* was measured at `fea6fbb`; `feat/orm` has taken **eight more** since this branch's base, and still fails the same way at `dba9a7e`. That strengthens the case rather than weakening it — I would just take the count from the branch at merge time rather than from the write-up, since this PR's whole subject is a number going stale while nobody looks.

Not blocking. Dropping one step and correcting one row would leave this doing exactly what it claims.

nstfkc:
## Follow-up on `6d491b1` — the filter fix, and the generate step is still there

Two changes since my review: a base merge through #130, and the `orm` → `orm database` filter. Re-measured both on this head.

**The filter change is right, and the reason for stopping there is right too.** The six files it still excludes are exactly the ones you name:

```
app/createComponentTree.test.ts        services/router/createFlatViewRoutes.test.ts
app/createFlatViewRoutes.test.ts       services/router/createRouteManifest.test.ts
app/createRouteManifest.test.ts        utils/parseTranslation.test.tsx
```

Router and translation, failing on `feat/orm` and unrelated to any of this. "A red board stops being read" is the correct trade, and saying so in the file is what stops someone widening it later and reverting to a permanently red run.

**The numbers in the comment do not match.** It says 47 test files, 32 with `orm` in the path. Measured through vitest, which is what the filter actually is:

```
vitest run                 52 files   6 failed | 46 passed
vitest run orm             36 files   all pass
vitest run orm database    38 files   1233 tests, all pass
vitest run database         2 files   28 tests
```

So the change recovers **2 files and 28 tests**, not 15. The 15 is the gap between the old filter and the *whole package*, and the commit deliberately closes only the part of it that belongs here — which is the right decision, described by a number that reads as though the filter now covers fifteen more files. `DatabaseManager.test.ts` being one of the two is the whole point and survives the correction intact.

**The generate step is unchanged**, and the finding reproduces on this head — the base merge did not affect it:

```
test alone                       1 failed — "running prisma generate again produces a zero-line diff"
prisma generate, then the test   2 passed
```

Still the only check on whether the committed artifact is current, still turned into a no-op by the step above it, and still not needed for the suite to run.

Worth noting the shape of both of these together: this PR exists because a number nobody re-measured stayed wrong through ten merges, and two of its own numbers are off in the same direction — the reported figure being from a different method than the one the tool uses. Not a criticism of the change, which is good and should land; it is the argument for the workflow printing its own counts rather than the write-up carrying them.

PR #165, opened 2026-07-29
Three exports nothing outside their file uses, and one dead local

### LINKED ISSUE #164: Three exports nothing outside their file uses, and one dead local
`orm/` has been at "1 warning" for the whole of this series and I kept reporting it as pre-existing rather than looking at it. It is a dead local, and looking at it turned up three dead exports beside it.

## The local

`compile/where.ts:201`, in `compileRelationFilter`:

```ts
const path = `where.${relation.name}`;
```

Never read. The identical line thirty lines down **is** used — it feeds the `argument` of three refusals — so this one is a leftover from when those moved. Removing it takes `oxlint packages/gemi/orm` from 1 warning to **0**.

Worth checking rather than assuming, because the interesting version would have been the opposite: a refusal that *should* carry the path and does not, which is the #108 / #112 shape. It is not that — this function delegates its refusals to the one that computes its own path.

## The exports

Scanning `orm/` for exported names referenced nowhere outside their own file:

| file | export |
| --- | --- |
| `relation-filters.ts` | `MANY_FILTER_OPERATORS` |
| `relation-filters.ts` | `ONE_FILTER_OPERATORS` |
| `corpus.ts` | `LITERAL` |

162 exported names; those three. The two operator tuples are used only by the function directly beneath them. `LITERAL` is mine — I exported it when extracting the shared corpus and never needed to.

None is reachable from `orm/index.ts`, so none was public. They were just wider than they had to be.

## Why it matters at all

`orm/index.ts` is a curated surface — the `PRE_SCOPED` symbol is deliberately absent from it so an application cannot forge one. A module-level `export` is a smaller version of the same decision: it is how a helper becomes something another file may reach for. Exporting one nothing imports invites a reach that was never intended, and it is the same class as the dead `relationKeys` parameter removed in #136 — surface that implies a use it does not have.

## Acceptance

- The local and the three exports removed; `oxlint` at 0 warnings for `orm/`.
- A check in `architecture.test.ts` asserting no export is referenced only inside its own file, with an allow-list for a deliberate exception. Test files count as references — an export existing *for* a test is a real reason, and `isTracked` says so in as many words.
- Verified by re-exporting one and confirming the check names it.

nstfkc:
Closes #164. Based on `feat/orm` directly.

`orm/` has sat at "1 warning" for this whole series and I kept reporting it as *pre-existing* rather than looking at it.

## The local

`compile/where.ts:201`, in `compileRelationFilter`:

```ts
const path = `where.${relation.name}`;
```

Never read. The identical line thirty lines down **is** used — it feeds the `argument` of three refusals — so this one is a leftover from when those moved. Removing it takes `oxlint packages/gemi/orm` from 1 warning to **0**.

Worth checking rather than assuming, because the interesting version would have been the opposite: a refusal that *should* carry the path and does not, which is the #108 / #112 shape. It is not that — this function delegates its refusals to the one that computes its own path.

## The exports

Exported names referenced nowhere outside their own file — three of 162:

| file | export |
| --- | --- |
| `relation-filters.ts` | `MANY_FILTER_OPERATORS` |
| `relation-filters.ts` | `ONE_FILTER_OPERATORS` |
| `corpus.ts` | `LITERAL` |

The two tuples are used only by the function directly beneath them. `LITERAL` is mine — exported when I extracted the shared corpus, never needed.

**None is reachable from `orm/index.ts`**, so none was public. They were just wider than they had to be.

## Why it matters at all

`orm/index.ts` is a curated surface — the `PRE_SCOPED` symbol is deliberately absent from it so an application cannot forge one. A module-level `export` is a smaller version of the same decision: it is how a helper becomes something another file may reach for. Exporting one nothing imports invites a reach that was never intended — the same class as the dead `relationKeys` parameter removed in #136: surface implying a use it does not have.

## The check

Lives with the other structural ones in `architecture.test.ts`, with an allow-list so a deliberate exception is something someone writes down rather than a silent pass.

**Test files count as references.** An export existing *for* a test is a real reason, and `isTracked` says so in as many words.

A name declared in two files is skipped rather than guessed at — a false positive here would read as a real finding, which is worse than a miss.

**Verified** by re-exporting one of the tuples: the check names it exactly.

- `oxlint` on `orm/`: **0 warnings**, 0 errors.
- 1312 unit tests, template suite 611 on SQLite.
- `tsc` shows only the known #159 failure, which #160 fixes.

### DIFF
--- a/packages/gemi/orm/architecture.test.ts
+++ b/packages/gemi/orm/architecture.test.ts
@@ -224,6 +224,69 @@ describe("the ORM's seams", () => {
   });
 
   /**
+   * **Nothing is exported that only its own file uses.**
+   *
+   * `orm/index.ts` is the public surface and is curated — the `PRE_SCOPED`
+   * symbol is deliberately absent from it so an application cannot forge one.
+   * A module-level `export` is a smaller version of the same decision: it is
+   * how a helper becomes something another file may reach for, and exporting
+   * one that nothing imports widens the surface for nothing.
+   *
+   * Three had drifted that way — `relation-filters.ts`'s two operator tuples,
+   * used only by the function directly beneath them, and `corpus.ts`'s
+   * `LITERAL`, which I exported when extracting that corpus and never needed
+   * to. None was reachable from `index.ts`, so none was public; they were just
+   * wider than they had to be.
+   *
+   * Test files count as references, because an export existing *for* a test is
+   * a real reason — `isTracked` says so in as many words.
+   */
+  const MAY_BE_EXPORTED_UNUSED: string[] = [
+    // Nothing today. An entry here is a deliberate exception with a reason
+    // beside it, not a place to put whatever this test happens to flag.
+  ];
+
+  test("no export is referenced only inside its own file", () => {
+    const declared = new Map<string, string>();
+    const sources = new Map<string, string>();
+
+    for (const [name, path] of FILES) {
+      const content = read(path);
+      sources.set(name, content);
+      if (name === "index.ts") continue;
+
+      for (const match of content.matchAll(
+        /^export (?:async )?function (\w+)|^export const (\w+)|^export class (\w+)/gm,
+      )) {
+        const exported = match[1] ?? match[2] ?? match[3];
+        // A name declared in two files is ambiguous to a text scan; skip rather
+        // than guess, since a false positive here reads as a real finding.
+        declared.set(exported, declared.has(exported) ? "" : name);
+      }
+    }
+
+    // Tests are references: an export that exists for a test is justified.
+    for (const path of readdirSync(ROOT, { recursive: true, withFileTypes: true })) {
+      if (!path.name.endsWith(".test.ts")) continue;
+      const full = join((path as { parentPath?: string }).parentPath ?? ROOT, path.name);
+      sources.set(`test:${full}`, read(full));
+    }
+
+    const unreferenced: string[] = [];
+    for (const [name, home] of declared) {
+      if (home === "" || MAY_BE_EXPORTED_UNUSED.includes(name)) continue;
+
+      const used = [...sources.entries()].some(
+        ([file, content]) =>
+          file !== home && new RegExp(`\\b${name}\\b`).test(content),
+      );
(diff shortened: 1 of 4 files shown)


nstfkc:
## Review — three exports nothing outside their file uses (closes #164)

Read the diff against its merge-base (`7590ceb`). Verified end to end:

```
base   oxlint orm/    1 warning, 0 errors
head   oxlint orm/    0 warnings, 0 errors
head   orm database   1312 passed
re-export one tuple   × no export is referenced only inside its own file
                        "relation-filters.ts exports MANY_FILTER_OPERATORS, and nothing else uses it"
```

**Checking the dead local rather than assuming it is the good part.** *"The interesting version would have been the opposite: a refusal that should carry the path and does not"* — that is the #108/#112 shape, it is what a duplicated `const path = …` most plausibly means, and ruling it out is what separates deleting a line from deleting evidence.

**I probed the two ways this check could be weaker than it looks, and both are empty.** It matches names as text over whole file contents, and this repository has more prose in its comments than most have in code — so an export mentioned only in a comment elsewhere would read as used. And a name declared twice is skipped rather than guessed at, which could hide a lot. Measured over `orm/`:

```
exports scanned              159
skipped as ambiguous           1   (`sql`)
"used" only inside a comment   0
```

So the duplicate-name skip costs one name, and the comment blind spot is theoretical rather than actual. 159 plus the three removed is your 162, which lines up.

**The empty allow-list with a comment explaining what an entry would mean** is the right starting state, and better than omitting the list: the next person to hit this has somewhere to write a reason instead of deciding between deleting a working export and deleting the test.

### One clarification that settles something I raised elsewhere

I noted on #161 that the series' reported `oxlint` figure did not match measurement. This PR explains it — the figure is `orm/`-scoped:

```
oxlint orm       1 warning     (what the series has been reporting)
oxlint           89 warnings   (what `bun run lint` runs, and what #161's Lint step will run)
```

Both are right for their scope, so the reported number was never wrong. What is worth reconciling is that #161's workflow runs the *wider* one, so the first CI run will print 89 against a series that has been saying 1, and the person reading it will not know which is the anomaly. Scoping that step to `orm` — or saying in the step why it is wider — would keep the two readings from colliding.

Nothing outstanding on this one.

PR #167, opened 2026-07-29
InvalidArgumentError is exported but missing from the Errors table

### LINKED ISSUE #166: feat/orm has a failing test: InvalidArgumentError is exported but not in the Errors table
`feat/orm` at 22eacb8 has a failing test:

```
× InvalidArgumentError is in the table
  Tests  1 failed | 1338 passed (1339)
```

That is the guard from #149/#150 doing its job — and it is failing because of **my own PR**, in the same way #159 happened.

## How

#103 added `InvalidArgumentError` and exported it. #150 added a check that every exported `Error` subclass appears in `docs/orm.md`'s **Errors** table.

#150's branch was cut before #103 merged, so on that branch the class did not exist and the check passed. On the merge result it does exist, is exported, and is not in the table. **Neither PR could see it**: #103 never touched the check, #150 never touched the class, each was green alone.

Same class as #159 — a break that exists only in the combination — and it is the second instance in this series. #162 exists precisely for this and would have caught both the moment they landed.

## The document is not wrong, exactly

`docs/orm.md` *does* describe `InvalidArgumentError`, in the refusal-taxonomy table #103 added:

> | `InvalidArgumentError` | the argument exists, the value cannot mean anything | fix the value |

What it is missing from is the **Errors** catalogue — the section that opens "Every failure is a typed error from `gemi/orm`" and is where a reader looks to find out what there is to catch. Described in one place, absent from the index.

That distinction is exactly what the check tests, and why it checks that section rather than the whole file.

## Acceptance

- The row added, saying what distinguishes it from its sibling: the argument exists and the *value* is the problem, so the message says what a good value looks like rather than that the argument is unsupported.
- No change to the check. It was right.

nstfkc:
Closes #166. Based on `feat/orm` directly — it fixes a failing test on the integration branch.

`feat/orm` at 22eacb8:

```
× InvalidArgumentError is in the table
  Tests  1 failed | 1338 passed (1339)
```

That is the guard from #149/#150 doing its job — and it is failing because of **my own PR**, in the same way #159 happened.

## How

| | |
| --- | --- |
| **#103** | added `InvalidArgumentError` and exported it |
| **#150** | added a check that every exported `Error` subclass appears in the **Errors** table |

#150's branch was cut before #103 merged, so on that branch the class did not exist and the check passed. On the merge result it exists, is exported, and is not in the table.

**Neither PR could see it.** #103 never touched the check; #150 never touched the class; each was green alone. Same class as #159 — a break that exists only in the combination — and the **second instance in this series**. #162 exists precisely for this and would have caught both the moment they landed.

## The document is not wrong, exactly

`docs/orm.md` *does* describe the class, in the refusal-taxonomy table #103 added:

> | `InvalidArgumentError` | the argument exists, the value cannot mean anything | fix the value |

What it is missing from is the **Errors** catalogue — the section that opens *"Every failure is a typed error from `gemi/orm`"* and is where a reader looks to find out what there is to catch. Described in one place, absent from the index.

That distinction is what the check tests, and why it reads that section rather than the whole file.

## The change

One row, saying what distinguishes it from its sibling: the argument exists and the **value** is the problem, so the message says what a good value looks like rather than that the argument is unsupported.

**No change to the check.** It was right.

- 1339 unit tests pass (was 1 failed | 1338 passed).
- Template suite: 611 on SQLite. `oxlint` unchanged.
- `tsc` still shows the separate #159 failure, which #160 fixes.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -967,6 +967,7 @@ Every failure is a typed error from `gemi/orm`, not a driver string.
 | `MissingRequiredValueError` | A write leaving a required column with no value and no default. |
 | `DecodeError` | A column the driver returned that cannot be read as its declared type. |
 | `UnsupportedByDesignError` | A subclass of `UnsupportedQueryError` for the arguments under [Not in scope](#not-in-scope) — it says *decision*, where the parent says *not yet*. Catch this one to tell them apart. |
+| `InvalidArgumentError` | The other subclass: the argument exists and the *value* cannot mean anything — `take: "-2"`. Says what a good value looks like, rather than that the argument is unsupported. |
 | `UnsupportedDialectError` | A model operation on a dialect with no compiler — MySQL and MariaDB. See [Dialects](#dialects). |
 | `ReturningUnsupportedError` | A write on a dialect without `RETURNING`, which is the same gap seen from the write path. |
 | `StaleSchemaArtifactError` | Generated files predate the running gemi. Re-run `prisma generate`. |


nstfkc:
## Review — `InvalidArgumentError` is exported but missing from the Errors table (closes #166)

The failure reproduces, and on a newer head than the one measured — `feat/orm` has moved to `b5511e5` since:

```
× InvalidArgumentError is in the table      1 failed | 1365 passed
```

The one row fixes it there too, not only at `22eacb8`:

```
1366 passed
```

Worth saying because this branch has moved under three of the last four PRs, and a docs row is the kind of fix that could have been overtaken.

**The distinction the description draws is the right one and it is what the guard encodes.** The class *is* described, in the refusal-taxonomy table #103 added — so "the document is wrong" would be false. What it is missing from is the catalogue that opens *"Every failure is a typed error from `gemi/orm`"*, which is where a reader looks for what there is to catch. Described in one place, absent from the index, and the check reads that section rather than the file for exactly that reason. Adding a row and leaving the check alone is the correct shape.

**Second instance of the class, and #161 would have caught it.** Its Unit tests step is `bun --bun vitest run orm database` in `packages/gemi`, which is the suite that fails here — so it would have gone red on the merge that created the break, naming the test. #159 was the same shape and is now fixed: `tsc --noEmit` on `feat/orm` is **clean** as of this head, so the first of the two is closed.

### I checked whether the mass merge left anything else, and it did not

Roughly twenty branches landed on `feat/orm` in one go, which is the condition both of these breaks came out of. With this PR's row applied to `b5511e5`:

```
tsc --noEmit                       clean
vitest run orm database            1366 passed
vitest run   (whole package)       7 failed | 1425 passed
                                     app/createComponentTree, createFlatViewRoutes, createRouteManifest
                                     services/router/{createFlatViewRoutes,createRouteManifest}
                                     utils/parseTranslation
template  vitest run app/models    614 passed | 32 skipped, 0 failed
```

The six failing files are the router and translation tests that were already failing before any of this and that #161 deliberately excludes. Nothing else came out of the merge — so this one row is the whole remaining cost of landing the stack, which is a better outcome than two instances in a row would suggest.

Nothing outstanding from me.

PR #169, opened 2026-07-29
Composite-in's field lookup had no test, and paired()'s empty guard is a floor

### LINKED ISSUE #168: Re-measuring construction-site coverage on the merged composite work: one gap, one floor
Re-ran the construction-site coverage from #114 against `feat/orm` at 24c836b, now that the composite-relation work (#95, #100) and everything through #160 has merged.

**21 error classes, 195 construction sites** — up from 20 and 188. Three genuinely unreached, after discounting two my own instrumentation missed (I recorded through a `globalThis` hook I never defined for the two classes outside `errors.ts`; both are covered, verified in #147/#148).

| site | verdict |
| --- | --- |
| `write.ts:1325` (`fieldOf`) | floor, already documented in #120 |
| `plan-relations.ts:835` (`paired`, no join fields) | **floor** — newly measured |
| `where.ts:570` (`compileCompositeIn`, unknown field) | **gap** |

## The gap

`assertCompositeInOperand` checks that a planner-built `$compositeIn` has a non-empty array of string fields and tuples of matching length — four guards, all tested. It does **not** check that each name is a field on the model. The only thing that does is the lookup in `compileCompositeIn`, and that one had no test.

It is reachable exactly the way its siblings are, through the exported `plannerCompositeIn` helper:

```
plannerCompositeIn(["nosuchfield"], [[1]])
  → UnknownFieldError: 'nosuchfield' is not a field on model LedgerEntry. Known fields: …
```

`UnknownFieldError` rather than the "planner built" wording, and that is right: the *shape* is fine and the **name** is wrong, which is what the caller-facing path reports for the same mistake.

## The floor

`paired()` refuses a relation naming no join fields. Fed both malformed shapes — empty `from`/`to` on both sides, and a lopsided one-to-two pairing — through an `include` and a relation filter, all four paths stop earlier: `resolveLink` raises `MalformedRelationError` naming both sides, or the length check catches the mismatch.

Fifth instance of this shape, after `select.ts:28`, `parseOrderBy`'s relation guard, `registry.get` and `fieldOf`. Kept rather than deleted, for the reason the others are, and now commented so the next coverage run does not re-triage it — I have now spent effort on `fieldOf` twice.

## Acceptance

- The missing test, beside the four siblings it belongs with.
- The floor commented with what pre-empts it and how that was measured.
- Verified by removing the field lookup: the new test names it.

nstfkc:
Closes #168. Based on `feat/orm` directly.

Re-ran the construction-site coverage from #114 against `feat/orm` at 24c836b, now that the composite-relation work (#95, #100) and everything through #160 has merged.

**21 error classes, 195 construction sites** — up from 20 and 188. Three genuinely unreached, after discounting two my own instrumentation missed: I recorded through a `globalThis` hook I never defined for the two classes outside `errors.ts`, and both are in fact covered (verified in #147/#148).

| site | verdict |
| --- | --- |
| `write.ts:1325` (`fieldOf`) | floor, already documented in #120 |
| `plan-relations.ts:835` (`paired`) | **floor** — newly measured |
| `where.ts:570` (`compileCompositeIn`) | **gap** |

## The gap

`assertCompositeInOperand` checks a planner-built `$compositeIn` has a non-empty array of string fields and tuples of matching length — four guards, all tested. It does **not** check that each name is a field on the model. The only thing that does is the lookup in `compileCompositeIn`, and that one had no test.

Reachable exactly the way its siblings are, through the exported `plannerCompositeIn` helper:

```
plannerCompositeIn(["nosuchfield"], [[1]])
  → UnknownFieldError: 'nosuchfield' is not a field on model LedgerEntry. Known fields: …
```

`UnknownFieldError` rather than the "planner built" wording, and that is right: the **shape** is fine and the **name** is wrong, which is what the caller-facing path reports for the same mistake.

## The floor

`paired()` refuses a relation naming no join fields. I fed both malformed shapes — empty `from`/`to` on both sides, and a lopsided one-to-two pairing — through an `include` and a relation filter. All four paths stop earlier, at `resolveLink`'s `MalformedRelationError` or the length check below it.

Fifth instance of this shape, after `select.ts:28`, `parseOrderBy`'s relation guard, `registry.get` and `fieldOf`. Kept rather than deleted, for the reason the others are — and now **commented with what pre-empts it**, because I have re-triaged `fieldOf` twice and that is twice too many.

## Verification

Removing the field lookup fails the new test by name.

- `tsc --noEmit` clean. 1366 unit tests pass, up one.

Two known conditions inherited from the base, both already fixed in open PRs: the `InvalidArgumentError is in the table` failure (#167) and the single `oxlint` warning in `where.ts` (#165).

### DIFF
--- a/packages/gemi/orm/compile/composite-relations.test.ts
+++ b/packages/gemi/orm/compile/composite-relations.test.ts
@@ -342,6 +342,32 @@ describe("the composite-in key is the planner's alone", () => {
     ).toThrow(/the planner built/);
   });
 
+  /**
+   * ...and a planner operand naming a field the model does not have.
+   *
+   * The shape checks above pass it — `["nosuchfield"]` is a non-empty array of
+   * strings with matching tuples — so the only thing that catches it is the
+   * field lookup in `compileCompositeIn`, and that lookup was the one guard in
+   * this family with no test. Found by re-running the construction-site
+   * coverage from #114 against the merged composite work.
+   *
+   * `UnknownFieldError` rather than the "planner built" wording, and that is
+   * right: the shape is fine and the *name* is wrong, which is the same thing
+   * the caller-facing path reports for the same mistake.
+   */
+  test("a planner operand naming a field that is not there", () => {
+    const planners = plannerCompositeIn(["nosuchfield"], [[1]]);
+
+    expect(() =>
+      compileRead(ledgerEntry, "findMany", { where: { $compositeIn: planners } }, postgres),
+    ).toThrow(UnknownFieldError);
+
+    // The known fields are listed, which is what makes it actionable.
+    expect(() =>
+      compileRead(ledgerEntry, "findMany", { where: { $compositeIn: planners } }, postgres),
+    ).toThrow(/is not a field on model LedgerEntry/);
+  });
+
   /** ...and the planner's own still compiles, on the dialect that binds it. */
   test("the planner's own is accepted", () => {
     const plan = compileRead(ledgerEntry, "findMany", { include: { ledger: true } }, postgres);
--- a/packages/gemi/orm/compile/plan-relations.ts
+++ b/packages/gemi/orm/compile/plan-relations.ts
@@ -832,6 +832,15 @@ function paired(
   operation: string,
 ): Link {
   if (parentFields.length === 0) {
+    // **No caller reaches this.** A relation with no fields on either side is
+    // refused earlier — `resolveLink` raises `MalformedRelationError` naming
+    // both sides — and a lopsided pairing is refused by the length check below
+    // it. Measured by feeding both shapes through an `include` and a relation
+    // filter: all four paths stop before here.
+    //
+    // Kept as the floor under a future caller that builds a `Link` some other
+    // way, not deleted — the same distinction `select.ts`'s `select + include`
+    // guard records, and for the same reason.
     throw new UnsupportedQueryError(
       relation.name,
       parent.name,


nstfkc:
## Review — composite-in's field lookup had no test (closes #168)

Re-ran the measurement independently. **21 classes, 195 construction sites** — matches exactly.

The branch is based on `24c836b`, and `feat/orm` has since moved to `00d5075` with both of the conditions you list already fixed (#167's row, #165's warning). Applied to that head instead:

```
1368 passed        (baseline 1367, so the +1 is the new case)
```

So it lands clean on the current head, not only on the one measured.

Sensitivity reproduces by name:

```
field lookup removed   × a planner operand naming a field that is not there
```

**`UnknownFieldError` rather than the planner wording is the right call**, and the reason given is the one that matters: the shape is fine and the name is wrong, so this is the same mistake the caller-facing path reports, and reporting it the same way is what keeps a reader from thinking they have hit a different class of problem.

### The floor claim holds, on more surfaces than you measured

Instrumented the `paired()` throw and ran the suite: **0 hits across 1368 tests**, so nothing reaches it — consistent with the coverage number. Then tried to reach it deliberately.

The comment says *"an `include` and a relation filter: all four paths stop before here"*. Correlated surfaces are five, and #157 established that list — so I ran all of them with both sides of the relation emptied:

```
include batched   MalformedRelationError
include lateral   MalformedRelationError
relation filter   MalformedRelationError
_count            MalformedRelationError
orderBy rel       MalformedRelationError
```

Every one stops at `resolveLink`, `paired()` still at 0 hits. The claim is right and is stronger than stated — worth saying five rather than four in the comment, since `_count` and `orderBy` are exactly where #116 found a surface behaving differently from its neighbours, and "four paths" invites the next reader to wonder which two were left out.

**One detail worth recording, because it is how this claim would be got wrong.** My first probe emptied `from`/`to` on the *parent's* relation only, and all five surfaces compiled happily — because that is the ordinary spelling of a foreign-side relation, resolved through the child's opposing one. The malformed shape needs **both** sides empty. A reachability claim tested with the first shape would have concluded the guard is reachable-but-silent rather than pre-empted, which is the opposite error and a more expensive one.

Fifth instance of the floor pattern, and commenting what pre-empts it is the right response to having re-triaged `fieldOf` twice. That is the thing that makes these cheap to skip next time.

nstfkc:
CI is red here for the reason it is red on `feat/orm` itself — `gemi-orm-generator: not found`, which is the workflow bug fixed in #171, not anything in this change. Confirmed from the run log: the only failing step is *Generate the template's model artifact*.

CI passes on #171's branch, so this goes green once that lands.

PR #171, opened 2026-07-29
CI: link the ORM generator, which `bun install` does not leave behind

### LINKED ISSUE #170: CI is red: prisma generate cannot find gemi-orm-generator on a fresh checkout
The workflow from #162 is failing on `feat/orm` and on every PR since it merged:

```
Error: Generator "gemi-orm-generator" failed:
/bin/sh: 1: gemi-orm-generator: not found
```

My bug, and the first run of that workflow is what found it — which is the job doing what it was added for.

## What actually happens

`prisma generate` resolves the generator **by bin name**, and `bun install` does not leave a usable one. Two behaviours combine, both measured locally:

- **At install time the target does not exist.** `packages/gemi/dist/bin/orm-generator.js` is produced by `bun run build:bin`, which the workflow runs *after* `bun install`. Bun does not create a link to a file that is not there.
- **A second install does not fix it.** With the lockfile satisfied, `bun install` reports `no changes` and relinks nothing. Deleting the link and re-running install leaves it deleted — confirmed for both the root and the template.

Interesting detail: re-running install *did* recreate `gemi` but not `gemi-orm-generator`. Bun appears to link only the bin matching the package name into a nested workspace.

## Not only a CI problem

A fresh clone hits exactly this. `bun install` then `bunx prisma generate` in the template fails the same way — CI just ran it on a clean machine, which is the first time anything had.

My own environment hid it: I created the link by hand early in this series, hit the same "not linked" error, and treated it as a local quirk rather than a repository one.

## Two links, not one

`generated.test.ts` asserts the binary exists at `templates/saas-starter/node_modules/.bin/gemi-orm-generator` — that is where a real installation puts it. The root link is what `prisma generate` resolves through.

Linking only the root makes `prisma generate` work and leaves that test failing, which is what running the workflow's steps in order locally showed before pushing.

## Acceptance

- The link created in both places after `build:bin`, with the reason for each.
- Verified by removing both links and running the workflow's sequence: generation succeeds, 1367 unit tests, 614 template tests.

nstfkc:
Closes #170. Based on `feat/orm` directly — CI is red on the integration branch.

The workflow from #162 has failed on `feat/orm` and every PR since it merged:

```
Error: Generator "gemi-orm-generator" failed:
/bin/sh: 1: gemi-orm-generator: not found
```

**My bug** — and the first run of that workflow is what found it, which is the job doing exactly what it was added for.

## What happens

`prisma generate` resolves the generator **by bin name**, and `bun install` does not leave a usable one. Two behaviours combine, both measured locally:

| | |
| --- | --- |
| **At install time the target does not exist** | `dist/bin/orm-generator.js` is produced by `build:bin`, which the workflow runs *after* `bun install`. Bun does not link a file that is not there. |
| **A second install does not fix it** | With the lockfile satisfied, `bun install` reports `no changes` and relinks nothing. Deleting the link and re-running install leaves it deleted. |

One detail worth recording: re-running install **did** recreate `gemi` but not `gemi-orm-generator`, so bun appears to link only the bin matching the package name into a nested workspace.

## Not only a CI problem

A fresh clone hits exactly this — `bun install` then `bunx prisma generate` in the template fails the same way. CI simply ran it on a clean machine, which is the first time anything had.

**My own environment hid it.** I created that link by hand early in this series after hitting the same error, and treated it as a local quirk rather than a repository one.

## Two links, not one

`generated.test.ts` asserts the binary exists at `templates/saas-starter/node_modules/.bin/gemi-orm-generator` — where a real installation puts it. The **root** link is what `prisma generate` resolves through.

Linking only the root makes generation work and leaves that test failing. I found that by running the workflow's steps in order locally *before* pushing, rather than by watching the next run go red.

## Verification

From a state with **neither** link present, running the workflow's sequence: generation succeeds, **1367** unit tests, **614** template tests, `typecheck` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,31 @@ jobs:
         run: bun run build:bin
         working-directory: packages/gemi
 
+      # `bun install` does not leave a usable `gemi-orm-generator` behind, and
+      # `prisma generate` resolves the generator **by bin name**.
+      #
+      # Two behaviours combine. At install time the target does not exist yet —
+      # `dist/bin/` is produced by the step above — so the link is not created.
+      # And a second `bun install` does not fix it: with the lockfile satisfied
+      # it reports "no changes" and relinks nothing, which is measurable
+      # locally by deleting the link and re-running install.
+      #
+      # So the link is made here, after the file exists. Not a CI quirk: a fresh
+      # clone hits exactly this, which is what the first run of this workflow
+      # discovered.
+      # Both places, and each for its own reason. The root one is what
+      # `prisma generate` resolves through; the template one is what
+      # `generated.test.ts` asserts exists, because that is where a real
+      # installation would put it.
+      - name: Link the ORM generator
+        run: |
+          for dir in node_modules templates/saas-starter/node_modules; do
+            mkdir -p "$dir/.bin"
+            ln -sf "$PWD/packages/gemi/dist/bin/orm-generator.js" \
+                   "$dir/.bin/gemi-orm-generator"
+          done
+          echo "$PWD/node_modules/.bin" >> "$GITHUB_PATH"
+
       # The template's committed artifact is what its suite runs against, and
       # `generated.test.ts` asserts a second run produces a zero-line diff.
       - name: Generate the template's model artifact


nstfkc:
## Review — link the ORM generator (closes #170)

**The fix can work, and I checked the parts that decide whether it does.** `prepare-bin.ts` gives the built file `#!/usr/bin/env bun` and `chmod 0o755`, so a bare `ln -sf` to it yields something `prisma generate` can execute — the step does not need a wrapper.

**"Two links, not one" is necessary rather than belt-and-braces.** `generated.test.ts:123` resolves the binary against the *template* root:

```ts
const bin = join(ROOT, "node_modules/.bin/gemi-orm-generator");
```

so linking only the root would make generation succeed and leave that assertion failing — exactly as described. Finding that by running the sequence locally before pushing, rather than by watching the next run go red, is the right instinct given the last one cost a red integration branch.

`bunx prisma generate` is also safe from a floating CLI: the lockfile pins `prisma@6.19.2`, and `--frozen-lockfile` installs it, so the step resolves the pinned version.

**What I could not verify.** I tried to reproduce the original failure from a state with neither link and hit a different one — my scratch `node_modules` is hand-assembled, so `bunx` reached Prisma 7.9.1 and failed at schema validation (`P1012`, the datasource `url` removal) long before any generator lookup. That says nothing about your diagnosis; it just means the end-to-end half is yours, not mine. The mechanism you describe is consistent with everything I *can* see.

The disclosure is the part worth keeping: *"my own environment hid it — I created that link by hand early in this series and treated it as a local quirk rather than a repository one."* That is the same trap that cost me several wrong measurements in this series, from the opposite direction — a stale prebuilt generator in my scratch trees producing phantom failures. A hand-made link is invisible precisely because it works.

### The step above this one still defeats the test it cites

Unchanged from my note on #161, and now live rather than theoretical — before this fix, CI never got as far as generating anything:

```
# The template's committed artifact is what its suite runs against, and
# `generated.test.ts` asserts a second run produces a zero-line diff.
- name: Generate the template's model artifact
```

That test snapshots the working tree *before its own* regeneration, so a stale committed artifact is repaired by this step and the test then compares the repaired file against itself:

```
test alone                       1 failed — "running prisma generate again produces a zero-line diff"
prisma generate, then the test   2 passed
```

With the links now created, the generate step will actually run, so from this PR onward CI cannot detect a stale artifact. The step is also not needed for the suite — the artifact is committed, and `614 passed | 32 skipped` with only `build:bin` and the links in place. Removing it restores the check and costs nothing.

Worth doing in this PR rather than separately, since it is the step immediately below the one being added and the two are now doing opposite things.

PR #173, opened 2026-07-29
CI: run the differential's Postgres half, which ran nowhere

### LINKED ISSUE #172: CI covers SQLite only, so the differential's Postgres half runs nowhere
The workflow added in #162 runs the template suite on SQLite. `differential.ts` runs against Postgres only when `TEST_POSTGRES_URL` is set — so the Postgres half of the harness runs when a person remembers to export a variable, and otherwise nowhere.

That harness is the repository's actual correctness claim: every query put through Prisma *and* through gemi against one database, compared on returned rows **and** on table contents. Half of it not running is half the claim unchecked.

## What is missed

- **The lateral relation strategy is the Postgres default.** An entire strategy — the one `plans/orm/benchmarks.md` justifies, and the one #154 is about — is exercised only by hand.
- **Every dialect divergence measured in this series.** `skipDuplicates` being Postgres-only, `in` binding as one array parameter, the `jsonb` refusal of bare scalars in #151, `DB.sql` escaping a transaction in #155. All of those are Postgres-side facts.
- The harness skips **loudly** rather than silently, which is the right design — but loud only helps someone watching, and nothing was.

## Measured

Ran the exact sequence against Postgres 16: **877 passed, 379 skipped** (the skips are the SQLite half, which the harness refuses to run against a client generated for the other provider).

## Why a separate job

`@prisma/client` is generated for one provider at a time, and the harness refuses the dialect it was not generated for rather than skipping — deliberately, so a mis-generated client cannot look like a pass. The two dialects therefore cannot share a checkout.

Three details the sequence needs, each learned the hard way in this series:

- **No `?schema=public`** in the URL. Bun's SQL client rejects it as an unrecognised parameter, which reads as a connection failure rather than a bad URL.
- **`--no-file-parallelism`.** The harness truncates shared tables between cases, so parallel files clobber each other's fixtures — measured, and it fails with row-count differences unrelated to the code.
- **A health check on the service.** Without it the first query races the server's startup.

The provider is switched with `sed` in the job rather than committed, so the repository stays on SQLite for everyone else.

## Acceptance

- A `postgres` job beside the existing one, which is renamed `sqlite` so the matrix is legible.
- Verified by running the job's exact sequence locally against Postgres 16.

nstfkc:
Closes #172. Stacked on #171, because the job needs that PR's generator link to get as far as `prisma generate`.

The workflow from #162 covers SQLite. `differential.ts` runs against Postgres only when `TEST_POSTGRES_URL` is set — so that half ran when a person remembered to export a variable, and otherwise **nowhere**.

That harness is the repository's actual correctness claim: every query put through Prisma *and* through gemi against one database, compared on returned rows **and** on table contents. Half of it not running is half the claim unchecked.

## What was missed

- **The lateral relation strategy is the Postgres default.** An entire strategy — the one `plans/orm/benchmarks.md` justifies, and the subject of #154 — was exercised only by hand.
- **Every dialect divergence measured in this series is Postgres-side**: `skipDuplicates` being Postgres-only, `in` binding as one array parameter, the `jsonb` refusal of bare scalars (#151), `DB.sql` escaping a transaction (#155).

The harness skips **loudly** rather than silently, which is the right design — but loud only helps someone who is watching, and nothing was.

## Why a separate job

`@prisma/client` is generated for one provider at a time, and the harness **refuses** the dialect it was not generated for rather than skipping — deliberately, so a mis-generated client cannot look like a pass. The two dialects cannot share a checkout.

Three details the sequence needs, each learned the hard way in this series:

| detail | why |
| --- | --- |
| no `?schema=public` in the URL | Bun's SQL client rejects it as an unrecognised parameter, which reads as a connection failure rather than a bad URL |
| `--no-file-parallelism` | the harness truncates shared tables between cases, so parallel files clobber each other's fixtures — measured, and it fails with row-count differences unrelated to the code |
| a health check on the service | without it the first query races the server's startup |

The provider is switched with `sed` inside the job rather than committed, so the repository stays on SQLite for everyone else. The existing job is renamed `sqlite` so the matrix is legible.

## Verification

Ran the job's exact sequence locally against Postgres 16 — same URL shape, same flags, same order: **877 passed, 379 skipped**. The skips are the SQLite half, which the harness refuses to run against a client generated for the other provider.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,9 +186,20 @@ jobs:
         run: bunx prisma db push --skip-generate
         working-directory: templates/saas-starter
 
+      # `-t postgres` selects the dialect-labelled suites, and it is required
+      # rather than tidy. With `TEST_POSTGRES_URL` set, the harness makes the
+      # *SQLite* half **fail** rather than skip — deliberately, so a client
+      # generated for the wrong provider cannot look like a pass. Run unfiltered,
+      # this job reports "877 passed" and still exits 1, because three suites
+      # error at collection.
+      #
+      # So this job runs the Postgres-labelled suites and the `sqlite` job runs
+      # the rest. The dialect-agnostic tests are that job's business; nothing is
+      # left uncovered between them.
+      #
       # Serially: the harness truncates shared tables between cases, so parallel
       # files clobber each other's fixtures. Measured — the parallel run fails
       # with row-count differences that have nothing to do with the code.
       - name: Differential and policy suites
-        run: bun --bun vitest run app/models --no-file-parallelism
+        run: bun --bun vitest run app/models --no-file-parallelism -t postgres
         working-directory: templates/saas-starter
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ concurrency:
 
 jobs:
   check:
+    name: sqlite
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — run the differential's Postgres half (closes #172)

No Postgres 16 here, so the run is yours. What I could check is whether the job's reasoning matches the harness, and it does — every claim in the description is written down in `differential.ts` already:

```
:23   "The skip is deliberately loud: a silently skipped dialect reads as a passing one."
:461  "…refuses any other protocol outright."
:469  TZ=UTC TEST_POSTGRES_URL=postgres://... vitest run --no-file-parallelism
:488  assertPrismaSpeaks(prisma, provider)
```

The job reproduces the harness's **own documented invocation**, flags included. That is the right relationship between a workflow and a suite — the workflow is not inventing a way to run it, it is running the way the file says to.

**The separate-job argument holds for a checkable reason.** `assertPrismaSpeaks` is empirical: it puts `select 1` through the client and reads the protocol error. So the two dialects genuinely cannot share a checkout, and — the part I went looking for — **a silently failed `sed` cannot produce a false pass**. It would leave a sqlite client, and the first query would raise the "generated for the wrong provider" error. The job goes red.

The `sed` itself is correct against the real file: `^  provider = "sqlite"` matches only line 19, leaving the `prisma-client-js` and `gemi-orm-generator` provider lines untouched, and line count unchanged.

### Two notes, neither blocking

**Name the cause when the `sed` misses.** It cannot cause a wrong pass, but it can cause a confusing failure: the job would report "@prisma/client is generated for the wrong provider, set `datasource db { provider }`" when the truth is that this step already tried to and the pattern moved. One line after it —

```sh
grep -q '^  provider = "postgresql"' prisma/schema.prisma
```

— turns that into a failure naming the step that failed. This series has spent a lot of effort on refusals that name their own origin; a workflow step is the same argument.

**The generate step means something different in each job.** Here `bunx prisma generate` is genuinely required — the provider just changed and the client must be re-emitted. In the `sqlite` job it is not, and it is what defeats `generated.test.ts`'s zero-line-diff check (my note on #170). Worth saying so in one of the two comments, because the steps now look identical and the obvious tidy-up is to unify them — which would either break generation here or restore the masking there.

Renaming the existing job `sqlite` is the small thing that makes the matrix readable, and worth having done in the same change.

nstfkc:
**The first version of this job could not pass, and CI caught it.**

Run unfiltered, the Postgres job reports `877 passed | 379 skipped` and still exits 1. With `TEST_POSTGRES_URL` set the harness makes the **SQLite** half *fail* rather than skip — deliberately, so a client generated for the wrong provider cannot look like a pass — so three suites error at collection and the run is red.

**My local check hid it.** I grepped the output for the test counts and read "877 passed", which is exactly what CI reported too. The suite-level failures are on a separate line I was not looking at. The lesson is about the filter rather than the harness: a verification that only reads the numbers it expects will confirm whatever it is shown.

Fixed with `-t postgres`, which selects the dialect-labelled suites: **622 passed, 634 skipped, exit 0**. The dialect-agnostic remainder is the `sqlite` job's business, so nothing falls between them.

Now green on GitHub — `postgres: success`, `sqlite: success` — which is the verification that matters here, since the whole point of this PR is a job that runs somewhere I cannot fully simulate.

nstfkc:
## Follow-up on `f4a5231` — the `-t postgres` filter reaches what it needs to

`-t` matches a substring of the **full test path**, not a file, so the question is whether nested describes under a dialect-labelled suite are selected. Measured with the sqlite analogue, which I can run:

```
app/models unfiltered      614 passed |  32 skipped  (646)
app/models -t sqlite       557 passed |  89 skipped
```

557 of 646 selected, so ancestor matching works — the inner describes come along with their suite.

**And it reaches the headline case.** The justification for this job is that the lateral strategy is the Postgres default, so the tests that exercise it are what must run. `differential.test.ts`:

```
1018  function suite(label, url) {
1050      describe("composite relations", …)      <- inside it
1483  suite("differential vs prisma — sqlite");
1486  suite("differential vs prisma — postgres", POSTGRES_URL);
```

On a Postgres run the full path is `differential vs prisma — postgres > composite relations > …`, which contains the word. Same for `skipDuplicates` and the implicit-m-n describes, which sit inside `writes vs prisma — postgres`. So the filter does not cost the thing the job exists for.

**The coverage claim holds too**, and for a reason worth writing down: the `sqlite` job runs `app/models` **unfiltered**, not `-t sqlite`. So the dialect-agnostic suites run there, the labelled ones run here, and nothing falls between. If that job ever gains a `-t sqlite` for symmetry, the 83 tests that match neither label would stop running anywhere.

### The correction is the part I would keep

The description's verification says *"877 passed, 379 skipped"*; this commit says the same command *"reports '877 passed' and still exits 1, because three suites error at collection"*. So the original number was read off the pass line while the command was failing — and you caught it and said so rather than quietly adjusting the flag.

That is the series' own lesson landing on its own verification, which is the useful place for it to land.

**One thing nothing enforces.** The filter depends on every dialect-labelled suite containing the word. All eight do today, in two spellings — `— postgres` and `(postgres)` — but a suite added as, say, `writes against pg` would silently stop running in this job, and a dialect-agnostic one named `postgres-style ordering` would be pulled in. Given how much of this series is guards against exactly that kind of drift, a check that every `suite(` label ends in a known dialect tag would cost a few lines and would fail loudly the first time someone deviates.

Issue #163, opened 2026-07-29
packages/gemi's suite is red: app/ holds superseded router copies whose tests fail


nstfkc:
Not an ORM issue — found while checking what a CI job should run, and reported rather than fixed because the fix is a judgement call outside that scope.

## Measured on `feat/orm` at cc44188

`bun --bun vitest run` in `packages/gemi`:

```
Tests  7 failed | 1246 passed | 1 skipped (1254)
FAIL  app/createComponentTree.test.ts
FAIL  app/createFlatViewRoutes.test.ts
FAIL  app/createRouteManifest.test.ts
```

All seven failures are in `app/`. The suite has been red for however long that has been true.

## Why nobody sees it

Three modules exist **twice**:

| | |
| --- | --- |
| `app/createFlatViewRoutes.ts` | tests **fail** |
| `services/router/createFlatViewRoutes.ts` | tests pass |

Same for `createRouteManifest` and `createComponentTree`. The `app/` copies appear superseded — nothing imports `app/createFlatViewRoutes`, and `services/router/` is where the working implementations live.

Nobody runs the whole package suite. Filtered runs are the norm, and every filter people use — `vitest run orm`, or a path while working — excludes `app/`.

## The thing that made me look

I proposed a CI job (#161 / #162) running `bun --bun vitest run orm`. That matches **32 of the package's 47 test files**. Among the 15 it skips is `database/DatabaseManager.test.ts`, which is the ORM's own SQLite foreign-key pragma test — so the job I proposed would have gone green while never running it.

Widened to `orm database` there, which is 34 files and passes. Deliberately **not** widened to the whole package, because that would put CI red for a reason unrelated to the change under review, and a board that is red for reasons nobody can act on stops being read.

## What I am not doing

Deleting or repairing the `app/` copies. Either could be right — they may be a migration in progress — and it is not mine to decide from an ORM loop. Once resolved, CI can drop the filter and run the package.

nstfkc:
## Measured, because this now gates the release — and it is not one problem

Raised as the blocker on #265: CI carries six `--exclude` flags to stay green,
those exclusions are `next`'s CI today, and they become **`main`'s** when #273
merges. Ran the suite without them against `next` at `932266b`.

**5 files fail, and they fail for three different reasons.** The issue title and
#265 both describe them as *"superseded copies in `app/`"*. That is true of three
of them and false of the other two, which is the part that changes the decision.

### 1. Dead copies — 3 files, nothing imports them

[...]

PR #175, opened 2026-07-29
CI: hold `orm/` at zero warnings, since a new one would be the 89th

### LINKED ISSUE #174: A new warning in orm/ would be one of 89, so nothing would notice
CI runs `bun run lint`, which is `oxlint` over all of `packages/gemi`: **88 warnings, 0 errors**. It passes, correctly — but a warning introduced in `orm/` would be the 89th, and nothing distinguishes it from the 88 that were already there.

`orm/` itself is at **0 warnings** since #165 removed the last one. That is a state worth holding, and holding it is cheap.

## The asymmetry is the point

The rest of the package is not this series' to clean up. Widening the gate to the whole package would put CI red for reasons nobody reviewing an ORM change can act on — the same argument #163 makes about the router suite, and the reason the test job runs `orm database` rather than everything.

So: keep `bun run lint` for the errors it would catch across the package, and add a second step that holds `orm/` specifically at zero. `oxlint --deny-warnings` turns a warning there into a build failure instead of a line nobody reads.

## Measured

```
oxlint orm --deny-warnings          → exit 0
  ...with one dead local added      → exit 1
  ...restored                        → exit 0
```

The regression it catches is exactly the one #165 fixed: an unused variable that sat in `where.ts` for the whole of this series while I reported it as "pre-existing" every time.

## Acceptance

- A second lint step scoped to `orm/`, with `--deny-warnings`.
- The package-wide lint kept, and the reason for the split written down rather than left as an oddity.

nstfkc:
Closes #174. Stacked on #173.

CI runs `bun run lint`, which is `oxlint` over all of `packages/gemi`: **88 warnings, 0 errors**. It passes, correctly — but a warning introduced in `orm/` would be the **89th**, and nothing would distinguish it from the 88 already there.

`orm/` is at **0 warnings** since #165 removed the last one. That state is worth holding, and holding it is cheap.

## The asymmetry is deliberate

The rest of the package is not this series' to clean up. Widening the gate would put CI red for reasons nobody reviewing an ORM change can act on — the same argument #163 makes about the router suite, and the reason the test job runs `orm database` rather than everything.

So the package-wide lint stays, for the **errors** it would catch anywhere; a second step holds `orm/` specifically at zero.

## Measured

| | |
| --- | --- |
| `oxlint orm --deny-warnings` | exit **0** |
| ...with one dead local added | exit **1** |
| ...restored | exit **0** |

The regression it catches is exactly the one #165 fixed: an unused variable that sat in `where.ts` for the whole of this series while I reported it as "pre-existing" in every iteration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,8 +106,23 @@ jobs:
         env:
           TZ: UTC
 
+      # Two lints, because the package and the ORM are in different states.
+      #
+      # `bun run lint` is `oxlint` over the whole package: 88 warnings, 0
+      # errors, so it passes — and a new warning in `orm/` would be one of 89.
+      # Kept, because the errors it would catch are worth catching.
+      #
+      # `orm/` is at zero after #165, so it can be held there. `--deny-warnings`
+      # makes that a build failure rather than a line nobody reads. Scoped to
+      # `orm/` deliberately: the rest of the package is not this series' to
+      # clean up, and a check that is red for someone else's reasons stops
+      # being read — the argument #163 makes about the suite.
       - name: Lint
         run: bun run lint
+        working-directory: packages/gemi
+
+      - name: Lint the ORM at zero warnings
+        run: bunx oxlint orm --deny-warnings
         working-directory: packages/gemi
 
   # The other half of the differential harness, which is the repository's actual


nstfkc:
## Review — hold `orm/` at zero warnings (closes #174)

Verified both gates and the sensitivity:

```
oxlint orm --deny-warnings                exit 0
  ...with a dead local added              exit 1
  ...restored                             exit 0
bun run lint (whole package)              88 warnings, 0 errors, exit 0
```

88 rather than the 89 I measured earlier — #165 removed the one in `where.ts`, which is consistent and is the point of the PR.

**Keeping both, with the asymmetry stated, is right.** A gate that goes red for the router's warnings would be red for reasons nobody reviewing an ORM change can act on, which is the same argument #163 makes about the suite and the reason the test step is scoped too. The package-wide run stays for errors; the narrow one holds the state this series actually earned.

The regression it catches is the honest example to name — an unused local that sat in `where.ts` through the whole series while it was reported as "pre-existing" every iteration, including by me. A number carried forward in a write-up is not a check.

### The two gates in this workflow disagree about scope

The test step runs `orm database`; this one runs `orm`. `database/` is where `DatabaseManager` and `MissingDatabaseUrlError` live — both things this series has been in, and #147's correction concluded that the coverage boundary belongs there too.

It is already clean, so widening costs nothing:

```
oxlint database                    0 warnings, 0 errors   (7 files)
oxlint orm database --deny-warnings   exit 0              (86 files)
```

Making it `bunx oxlint orm database --deny-warnings` matches the scope the step twenty lines above already uses, and holds a directory that is at zero today rather than waiting for it to acquire a warning nobody is watching for. If the answer is that `database/` is a different series' responsibility, that is a fine answer — it is just worth being the same answer both steps give.

Nothing blocking.

PR #177, opened 2026-07-29
Keep the template's foreign keys indexed, which the docs promise

### LINKED ISSUE #176: Nothing keeps the template's foreign keys indexed, which the docs promise
`docs/orm.md` makes a promise on this template's behalf, in the **Counting a relation** section:

> The starter template indexes every foreign key for this reason, so a new application starts on the right side of it.

The reason is measured, and it is not a small one:

> On SQLite that one line is worth several times the query — enough to **flip** the advice: unindexed, `_count` is slower than loading the children and counting them in JavaScript; indexed, it is comfortably faster. The same applies to relation filters and relation orderings, which are the same shape.

A relation `_count`, a relation filter and a relation ordering all compile to a correlated subquery that runs **once per parent row**. Prisma declares no index for a relation's foreign key on either dialect, so unless the schema says so, there is none.

## Measured

Parsed the template's schema: **13 models, 0 unindexed foreign keys.** The promise holds today.

What it lacks is anything to keep it true. A model added later with a `@relation` and no `@@index` does not make the template slightly slower — it makes a documented recommendation **wrong**, silently, for every application generated from it.

That is the same drift as #157, where a *Not in scope* entry stopped being true because the boundary moved underneath it, and #164, where the export surface widened without anyone deciding to.

## Acceptance

- A test over `prisma/schema.prisma` asserting every `@relation(fields: [...])` is covered by an index.
- **Prefix** matching, because that is what a composite index serves: an index on `[a, b]` answers a lookup on `[a]`, and one on `[b]` does not.
- `@@index`, `@@unique`, `@@id` and a single-field `@unique`/`@id` all count — a foreign key that is also the primary key needs no separate index, which is the shape a one-to-one takes.
- A guard that the parse found models and foreign keys at all, so a schema that stops matching the pattern fails loudly rather than passing over nothing.

nstfkc:
Closes #176. Based on `feat/orm` directly.

`docs/orm.md` makes a promise on this template's behalf:

> The starter template indexes every foreign key for this reason, so a new application starts on the right side of it.

And the reason is measured, and not small:

> On SQLite that one line is worth several times the query — enough to **flip** the advice: unindexed, `_count` is slower than loading the children and counting them in JavaScript; indexed, it is comfortably faster.

A relation `_count`, a relation filter and a relation ordering all compile to a correlated subquery that runs **once per parent row**. Prisma declares no index for a relation's foreign key on either dialect, so unless the schema says so, there is none.

## Measured

**13 models, 0 unindexed foreign keys.** The promise holds today.

What it lacks is anything to keep it true. A model added later with a `@relation` and no `@@index` does not make the template slightly slower — it makes a **documented recommendation wrong**, silently, for every application generated from it.

Same drift as #157, where a *Not in scope* entry stopped being true because the boundary moved underneath it, and #164, where the export surface widened without anyone deciding to.

## What the check gets right

- **Prefix** matching, because that is what a composite index serves: an index on `[a, b]` answers a lookup on `[a]`, and one on `[b]` does not.
- `@@index`, `@@unique`, `@@id` **and** a single-field `@unique`/`@id` all count — a foreign key that is also the primary key needs no separate index, which is the shape a one-to-one takes.
- A guard that the parse found models and foreign keys **at all**, so a schema that stops matching the pattern fails loudly rather than passing over nothing.

## Verification

Deleting one `@@index` fails with the model and field named:

```
Account: @relation(fields: [userId]) has no index with that prefix
```

- Template suite: 616 passed on SQLite (up 2).
- 1367 unit tests, `typecheck` clean.

### DIFF
--- a/templates/saas-starter/app/models/foreign-key-indexes.test.ts
+++ b/templates/saas-starter/app/models/foreign-key-indexes.test.ts
@@ -1,0 +1,103 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * **Every relation's foreign key is indexed**, which `docs/orm.md` promises on
+ * this template's behalf.
+ *
+ * The promise is not decorative. A relation `_count`, a relation filter and a
+ * relation ordering all compile to a correlated subquery that runs **once per
+ * parent row**, so without an index each run scans the child table — and Prisma
+ * declares no index for a relation's foreign key on either dialect, so by
+ * default there is none.
+ *
+ * The docs put a number on it: on SQLite the index is worth enough to *flip*
+ * the advice. Unindexed, `_count` is slower than loading the children and
+ * counting them in JavaScript; indexed, it is comfortably faster. So a model
+ * added without one does not make the template a little slower — it makes a
+ * documented recommendation wrong.
+ *
+ * > The starter template indexes every foreign key for this reason, so a new
+ * > application starts on the right side of it.
+ *
+ * That sentence is what this asserts. It held when written and holds now; what
+ * it lacked was anything to keep it true as models are added, which is the same
+ * drift #157 caught in the prose and #164 in the export surface.
+ */
+const SCHEMA = join(import.meta.dirname, "../../prisma/schema.prisma");
+
+/** A model's body, keyed by name. */
+function models(source: string): Map<string, string> {
+  const out = new Map<string, string>();
+  for (const match of source.matchAll(/^model (\w+) \{(.*?)^\}/gms)) {
+    out.set(match[1], match[2]);
+  }
+  return out;
+}
+
+/** The local columns each `@relation(fields: [...])` uses. */
+function foreignKeys(body: string): string[][] {
+  return [...body.matchAll(/@relation\([^)]*fields:\s*\[([^\]]*)\]/g)].map((match) =>
+    match[1].split(",").map((name) => name.trim()).filter(Boolean),
+  );
+}
+
+/**
+ * Every column list the database will have an index for.
+ *
+ * `@@index`, `@@unique` and `@@id` all produce one, and so does a single-field
+ * `@unique` or `@id` — a foreign key that is also the primary key needs no
+ * separate index, which is the shape a one-to-one takes.
+ */
+function indexed(body: string): string[][] {
+  const lists: string[][] = [];
+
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — keep the template's foreign keys indexed (closes #176)

Verified the promise, the guard and its sensitivity:

```
docs/orm.md:595   "The starter template indexes every foreign key for this reason…"
schema            13 models, 8 foreign keys, all covered
guard             2 passed
@@index([userId]) deleted
  → × every relation's foreign key is covered by an index
    "Account: @relation(fields: [userId]) has no index with that prefix"
```

**The prefix rule is not just a comment — the template exercises it.** `LedgerEntry.ledger` joins on `[tenantId, ledgerCode]` and is covered by `@@index([tenantId, ledgerCode])`. Reversing that index:

```
@@index([ledgerCode, tenantId])
  → "LedgerEntry: @relation(fields: [tenantId, ledgerCode]) has no index with that prefix"
```

So the subtlest rule in the check is pinned by a real composite relation rather than by reasoning about one. That is worth more than the eight single-column cases, because the set-versus-prefix mistake is the one a reimplementation would make.

**The floor test is the right shape** — asserting the parse found models *and* at least one foreign key, so a schema that stops matching the regexes fails rather than passing over an empty list. Third PR in a row to open with a vacuity guard rather than acquire one.

The framing is right too: an unindexed foreign key here does not make the template slower, it makes a **documented recommendation wrong** for every application generated from it. That is why it belongs in the template's suite and not in a benchmark.

### One false positive, and it is the kind this check is meant not to have

Prisma allows per-field modifiers in an index, and the parse does not strip them:

```
@@index([userId(sort: Desc)])
  → "Account: @relation(fields: [userId]) has no index with that prefix"
```

That index *does* cover `userId`. The template does not use modifiers today, so nothing is failing — but the failure it would produce tells the reader the opposite of the truth, and points at an index sitting three lines away. A miss is cheaper than that.

One line in `indexed()` closes it, and I checked both directions:

```ts
.map((name) => name.trim().replace(/\(.*$/, ""))
```
```
unmodified schema        2 passed
with a sort modifier     2 passed
```

The same applies to `length:` on a `String` key, which is the modifier a template is most likely to acquire first.

Nothing else outstanding.

nstfkc:
CI is red here for the same reason it is red on `feat/orm` itself — `gemi-orm-generator: not found`, the workflow bug fixed in #171 — not anything in this change. This goes green once that lands.

PR #179, opened 2026-07-29
The generator was never tested on a composite relation

### LINKED ISSUE #178: The generator is never tested on a composite relation, and a truncation there is silent
Composite relations landed in #95 and #100, and the compiler is thoroughly tested for them. The **generator** that produces the artifact those tests stand in for is not.

Every relation in `emit.test.ts` declares exactly one field:

```ts
relationFromFields: ["authorId"],
relationToFields: ["id"],
```

## Why that gap is quiet rather than loud

Three things would each have caught a truncation, and none of them can:

- The **compiler's** composite tests build their own `ModelSchema` fixtures by hand, so they never run the generator.
- The **starter template** declares no composite relation — measured: 16 relations, 0 composite — so its suite exercises nothing here.
- The **artifact diff** test asserts a second `prisma generate` matches the first. A generator that consistently drops a field is consistently wrong, and the diff stays zero.

So a generator keeping only the first field would emit `from: ["tenantId"]`, every suite would stay green, and an application with a composite relation would correlate on **one column** and return plausible wrong rows — which is exactly what the compiler's own refusal was written to prevent, back when the shape was unsupported.

## Measured

The generator is correct: a two-field relation comes through as `from: ["tenantId", "ledgerCode"]`, `to: ["tenantId", "code"]`, with the far side empty and the composite primary key intact.

And the gap is real. Truncating `emit.ts` to `slice(0, 1)`:

```
bin/orm/emit.test.ts     1 failed | 26 passed
orm database             1 failed | 1367 passed
```

**Only the new test notices.** 1367 others pass.

## Acceptance

- A generator test for a two-field relation, asserting **both** sides — `from` and `to` are positional pairs, so a truncation or a reorder on either is the same bug.
- The far side asserted empty, since it is resolved through the owning side at plan time and is where a missing key would *not* show up.
- The composite primary key it references asserted too, being the same shape.

nstfkc:
Closes #178. Based on `feat/orm` directly.

Composite relations landed in #95 and #100, and the compiler is thoroughly tested for them. The **generator** that produces the artifact those tests stand in for is not — every relation in `emit.test.ts` declares exactly one field:

```ts
relationFromFields: ["authorId"],
relationToFields: ["id"],
```

## Why the gap is quiet rather than loud

Three things would each have caught a truncation, and **none of them can**:

| | |
| --- | --- |
| the compiler's composite tests | build their own `ModelSchema` fixtures by hand — they never run the generator |
| the starter template | declares no composite relation (16 relations, 0 composite), so its suite exercises nothing here |
| the artifact diff test | asserts a second `prisma generate` matches the first, which a *consistently* wrong generator satisfies |

So a generator keeping only the first field would emit `from: ["tenantId"]`, every suite would stay green, and an application with a composite relation would correlate on **one column** and return plausible wrong rows — exactly what the compiler's own refusal was written to prevent, back when the shape was unsupported.

## Measured

The generator is **correct**: a two-field relation comes through as `from: ["tenantId", "ledgerCode"]`, `to: ["tenantId", "code"]`, far side empty, composite primary key intact.

The gap is in what would notice if it stopped being. Truncating `emit.ts` to `slice(0, 1)`:

```
bin/orm/emit.test.ts     1 failed | 26 passed
orm database             1 failed | 1367 passed
```

**Only the new test notices.**

## What it asserts

- **Both sides.** `from` and `to` are positional pairs, so a truncation *or* a reorder on either is the same bug.
- **The far side empty**, since it is resolved through the owning side at plan time — and is therefore where a missing key would *not* show up.
- **The composite primary key** it references, being the same shape.

- `typecheck` clean, 1368 unit tests, `oxlint orm --deny-warnings` exit 0.

### DIFF
--- a/packages/gemi/bin/orm/emit.test.ts
+++ b/packages/gemi/bin/orm/emit.test.ts
@@ -351,6 +351,79 @@ describe("buildModelSchemas()", () => {
     const [, user] = buildModelSchemas([USER, POST]);
     expect(user.relations.posts.joinTable).toBeUndefined();
   });
+
+  /**
+   * **A relation that joins on more than one field**, which every other case
+   * here declares with exactly one.
+   *
+   * The compiler handles this shape and is well tested for it — but against
+   * *hand-written* fixtures, and the starter template declares no composite
+   * relation, so nothing exercised the generator that produces the real
+   * artifact.
+   *
+   * The failure that leaves open is the quiet one. A generator that kept only
+   * the first field would emit `from: ["tenantId"]`, every compiler test would
+   * still pass because they build their own schemas, the template suite would
+   * still pass because it has no such relation — and an application with one
+   * would correlate on a single column and return plausible wrong rows. That is
+   * precisely what the compiler's own refusal was written to prevent, so it is
+   * worth checking that the input reaches it intact.
+   *
+   * Both sides are asserted: `from` and `to` are positional pairs, and a
+   * truncation or a reorder on either side is the same bug.
+   */
+  test("carries every field of a composite relation, on both sides", () => {
+    const LEDGER = model({
+      name: "Ledger",
+      fields: [
+        field({ name: "tenantId" }),
+        field({ name: "code", type: "String" }),
+        field({
+          kind: "object",
+          name: "entries",
+          type: "Entry",
+          isList: true,
+          relationName: "EntryToLedger",
+          relationFromFields: [],
+          relationToFields: [],
+        }),
+      ],
+      primaryKey: { name: null, fields: ["tenantId", "code"] } as never,
+    });
+
+    const ENTRY = model({
+      name: "Entry",
+      fields: [
+        field({ name: "id", isId: true }),
+        field({ name: "tenantId" }),
+        field({ name: "ledgerCode", type: "String" }),
+        field({
+          kind: "object",
+          name: "ledger",
+          type: "Ledger",
+          isRequired: true,
+          relationName: "EntryToLedger",
+          relationFromFields: ["tenantId", "ledgerCode"],
+          relationToFields: ["tenantId", "code"],
+        }),
+      ],
+    });
+
+    const [entry, ledger] = buildModelSchemas([ENTRY, LEDGER]);
+
+    // The owning side names both of its own columns and both of the target's,
+    // in declaration order — the pairing is positional.
+    expect(entry.relations.ledger.from).toEqual(["tenantId", "ledgerCode"]);
+    expect(entry.relations.ledger.to).toEqual(["tenantId", "code"]);
+
+    // The far side names neither, and is resolved through this one at plan
+    // time — which is why a truncation here would not show up as a missing key.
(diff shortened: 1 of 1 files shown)


nstfkc:
CI is red here for the same reason it is red on `feat/orm` itself — `gemi-orm-generator: not found`, the workflow bug fixed in #171 — not anything in this change. This goes green once that lands.

nstfkc:
## Review — the generator was never tested on a composite relation (closes #178)

The generator is correct and a unit test on it is worth having — fast, precise, names the field, needs no database and no regeneration. But I could not reproduce the "why the gap is quiet" table, and all three rows come apart the same way.

**The template does declare a composite relation.**

```
prisma/schema.prisma:235
  ledger Ledger @relation(fields: [tenantId, ledgerCode], references: [tenantId, code])
```

8 foreign keys, 1 composite — the same one #176's index guard covers. And the differential seed was written for exactly this failure:

> the *same* ledger code appears in two tenants, so a join that used only the first field — **or only the second** — returns the wrong entries rather than failing

**So the template suite does exercise it.** Truncating `emit.ts` to `slice(0, 1)`, regenerating the artifact, and running the template suite:

```
10 failed | 604 passed | 32 skipped     app/models/differential.test.ts
```

**And the artifact diff test catches it too** — this is row three, and the reasoning there is inverted. It does not compare two generator runs; it compares the **committed artifact** against one run. A consistently wrong generator therefore disagrees with what is committed and fails:

```
truncated generator, artifact left as committed
  → 1 failed     app/models/generated.test.ts
```

So *"only the new test notices"* holds for the **unit** suites, which is what was measured (`emit.test.ts` and `orm database`). The template suite notices in both configurations — through the diff test if nothing regenerated first, through the differential if something did.

### The interaction is worth more than the correction, and it revises something I argued for

Those two paths are mutually exclusive, and which one fires depends on the CI step I have twice asked to be removed:

| | `generated.test.ts` | differential |
| --- | --- | --- |
| no pre-regenerate (a local run) | **catches it** | passes — artifact still correct |
| pre-regenerate (CI, since #170) | masked | **catches it** |

I argued on #161 and #170 that the `prisma generate` step should go because it turns the diff test into a no-op. That is still true — but I did not see that the step is also what makes the *differential* sensitive to the generator at all. Removing it restores one check and removes the other, rather than being free as I said.

The safe version is the step plus a `git diff --exit-code` on `app/models/generated` immediately after it, which catches a stale or wrongly-generated artifact directly instead of relying on whichever of the two happens to fire. That is a better suggestion than the one I gave twice.

None of this argues against the PR. The unit test is the right place for this — it fails in milliseconds naming `from`/`to`, where the template paths fail with ten row-count differences or a file diff. It is the justification that needs correcting, not the test.

nstfkc:
See the correction on #178 — two of the three reasons given in this PR were wrong. The template *does* declare a composite relation, and the artifact's zero-diff test *does* catch a truncating generator; my "only the new test notices" measurement was run against the package suite, which excludes the template. The test remains worth having for the one reason that holds, and the comment it left behind is corrected in a follow-up.

PR #181, opened 2026-07-29
docs/llms-full.txt carried half the ORM page

### LINKED ISSUE #180: docs/llms-full.txt carries half the ORM page, including a claim that omit is unsupported
`docs/llms-full.txt` is "every page concatenated for one-shot LLM context", served verbatim at a stable URL by the docs workflow. `plans/orm/README.md` records the intent: `orm.md` and `orm-rows-and-entities.md` were unreachable from the four index files "until the doc pass. **Both are now in all four.**"

The ORM page's embedded copy had drifted to **626 lines against the source's 1209** — roughly half.

## What it still said

```
## Not in scope
- **No identity map and no unit of work.**
- **No lazy loading.**
- **No `omit`.** Use `select`, or a policy's `redact`.
- **No migrations, no schema DSL.**
```

`omit` is implemented and `orm.md` documents it under its own heading. The embedded copy also predates the `Aggregates` and `groupBy` sections, `Json` columns, `Dialects`, the composite-relation paragraphs, the error taxonomy, and every correction in this series — #145, #149, #151, #155, #157, #166.

## Worse than the same drift in `orm.md`

For the reason #145 gives about "refused, but actually works": a wrong runtime refusal gets corrected the moment someone runs the query. **This file's audience is tools**, which repeat what they are given, confidently, and never bump into the runtime that would correct them. An assistant reading it would tell you `omit` is unsupported and to reach for a policy's `redact` instead.

## Checked rather than assumed

I compared every `.md` page against its embedded copy. Only the ORM page had drifted materially. `orm-rows-and-entities.md` differed **only** in title casing — `## Rows and Entities` against the source's `# Rows and entities` — and my first pass reported it as "not embedded" because the matcher keyed on the title; worth saying, because that was a false alarm I nearly wrote up.

`README.md`, `cli.md` and `configuration.md` also differ. Those are outside this series and are left alone rather than swept in.

## Acceptance

- The ORM page replaced with its source, and the sibling page's heading normalised so both match exactly.
- A test asserting both are embedded **verbatim**, comparing exactly rather than loosely — a "close enough" check is what let this reach half a page.

nstfkc:
Closes #180. Based on `feat/orm` directly.

`docs/llms-full.txt` is "every page concatenated for one-shot LLM context", served **verbatim** at a stable URL by the docs workflow. `plans/orm/README.md` records the intent: `orm.md` and `orm-rows-and-entities.md` were unreachable from the four index files "until the doc pass. **Both are now in all four.**"

The ORM page's embedded copy had drifted to **626 lines against the source's 1209** — roughly half.

## What it still said

```
## Not in scope
- **No identity map and no unit of work.**
- **No lazy loading.**
- **No `omit`.** Use `select`, or a policy's `redact`.
- **No migrations, no schema DSL.**
```

`omit` is implemented and `orm.md` documents it under its own heading. The copy also predates `Aggregates`, `groupBy`, `Json` columns, `Dialects`, the composite-relation paragraphs, the error taxonomy, and every correction in this series — #145, #149, #151, #155, #157, #166.

## Worse than the same drift in `orm.md`

For the reason #145 gives about *"refused, but actually works"*: a wrong runtime refusal is corrected the moment someone runs the query. **This file's audience is tools**, which repeat what they are given, confidently, and never bump into the runtime that would correct them. An assistant reading it would tell you `omit` is unsupported and to reach for a policy's `redact`.

## Checked rather than assumed

I compared **every** `.md` page against its embedded copy. Only the ORM page had drifted materially.

`orm-rows-and-entities.md` differed **only** in title casing — `## Rows and Entities` against the source's `# Rows and entities` — and my first pass reported it as *"not embedded"* because the matcher keyed on the title. That was a false alarm I nearly wrote up; the sibling's heading is normalised here so both match exactly.

`README.md`, `cli.md` and `configuration.md` also differ. Those are outside this series and are **left alone** rather than swept in.

## The guard

Compares **exactly**, not loosely — a "close enough" check is what let this reach half a page. Verified by editing a heading in `orm.md`: the test fails.

- `typecheck` clean, 1370 unit tests, `oxlint orm --deny-warnings` exit 0.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2743,7 +2743,7 @@ generator gemi {
 Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 - `schema.ts` — runtime metadata: tables, columns, types, defaults, relations.
-- `models.ts` — a typed base class per model, carrying the thirteen operations.
+- `models.ts` — a typed base class per model, carrying the fourteen operations.
 - `index.ts` — registers every model by name.
 
 **The output is committed on purpose.** Diffs stay reviewable and CI needs no codegen step.
@@ -2792,10 +2792,10 @@ read. It can only see modules you hand it, so it does not replace the `register`
 
 ## Querying
 
-Thirteen operations, with Prisma's argument types verbatim:
+Fourteen operations, with Prisma's argument types verbatim:
 
 ```
-findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count
+findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count   aggregate   groupBy
 create     createMany  update             updateMany   upsert              delete   deleteMany
 ```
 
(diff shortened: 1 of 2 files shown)


nstfkc:
## Review — `docs/llms-full.txt` carried half the ORM page (closes #180)

Measured the drift independently at the base:

```
docs/orm.md                1209 lines
embedded `## ORM` section   624 lines
```

(626 by your boundary, 624 by mine — the two lines are the heading and its blank. Same finding.)

**The `omit` claim is exactly as bad as described.** It is in `READ_ARGS` for all five reads and documented at `orm.md:257` under its own heading, while the embedded copy said:

> - **No `omit`.** Use `select`, or a policy's `redact`.

After the fix the section is 1211 lines and no longer says it.

**The argument for why this is worse than the same drift in `orm.md` is the right one**, and it is the sharpest version of #145's point I have seen in this series: a wrong refusal in the docs is corrected the moment a reader runs the query, and this file's readers never run anything. An assistant would not merely fail to mention `omit` — it would actively talk someone out of it, and the substitute it offers is a policy, which is a different thing with different semantics.

**The guard compares exactly, and I checked it is not fooled by either kind of edit:**

```
heading changed in orm.md    × orm.md is embedded verbatim
a paragraph reflowed         × orm.md is embedded verbatim
untouched                      57 passed
```

### I ran your sweep independently, and it holds

Every `.md` page against its embedded copy, at the base:

```
orm.md                     1209L src / 624L embedded    <- the drift
orm-rows-and-entities.md   NOT EMBEDDED under its own title  <- the casing false alarm, reproduced
README.md                    88L src /  84L embedded
authorization, broadcasting, cli, configuration, cron, middleware
                           differ, identical line counts     <- formatting only
15 others                  exact
```

So *"only the ORM page had drifted materially"* is right. The `orm-rows-and-entities` alarm reproduces precisely — it reports as missing because the matcher keys on the title and the embedded heading is `## Rows and Entities` against the source's `# Rows and entities`. Reporting that as a near-miss rather than quietly normalising it is worth more than the fix, because the next person to write a matcher over these files will key on the title too.

**One page has a real size difference and is not called out as such.** `README.md` is 88 lines against 84 embedded — the only *other* page where the difference is content rather than whitespace, and it is listed with `cli.md` and `configuration.md` as "left alone". Leaving it is the right call for this series; being the only one of the three that is not formatting is worth the half-sentence, so whoever picks up the rest knows which one is not cosmetic.

**One latent fragility.** `embedded()` slices to the next `\n---`. Neither ORM page contains a bare `---` today — I checked, zero in both — so it works. But a horizontal rule is an ordinary thing to add to a 1209-line page, and the day someone does, this test fails with a truncated-diff mismatch rather than saying what happened. Keying the end on the *next* `\n## ` heading instead would be immune to that and costs nothing.

PR #183, opened 2026-07-29
The operation count is documented three different ways, all wrong

### LINKED ISSUE #182: The ORM's operation count is documented three different ways, all wrong
`Operation` declares **fifteen** operations. The repository states the count in fourteen places, in three different wrong numbers:

| count | where |
| --- | --- |
| **twelve** | `active-record.ts` (×3), `active-record.test.ts`, `architecture.test.ts`, `orm-rows-and-entities.md`, `plans/orm/README.md`, `plans/orm/08-eloquent-doorway.md` |
| **thirteen** | `strategy.ts`, `strategy.test.ts`, `llms.txt`, `plans/orm/README.md` |
| **fourteen** | `docs/orm.md` (×2) |

Each was right when written. None moved when `count`, `aggregate` and `groupBy` arrived — twelve plus those three is fifteen.

**I contributed to it.** #144 wrote "not one of the thirteen operations" in `strategy.test.ts`, copied out of the comment beside it without checking.

## The list itself is fine

`docs/orm.md`'s code block names all fifteen. Only the prose introducing it says "Fourteen". That is the milder failure — but it is the sentence a reader meets first, and a number that has drifted three times will drift again.

## The fix, and what it is guarded on

The counts are replaced with phrasing that cannot drift — "every operation", "the operations" — except in `orm.md`'s introduction, where the number is doing real work and is now correct.

The **guard is on the list, not the number**. A wrong count is one word out of date; a wrong list is a documented surface that does not exist, or an existing one nobody is told about. It asserts the code block names exactly the operations the compiler implements, checked against the same predicate-verified set #146 uses.

`plans/` is deliberately untouched. That directory records what was believed at each iteration, and says so: a superseded claim is "left standing above with this correction beneath it rather than quietly rewritten, because the gap between what an invariant promised and what it delivered is the useful part."

## Acceptance

- Every stale count in `docs/` and `packages/gemi/orm/` corrected or reworded.
- `llms.txt` and `llms-full.txt` brought along, since they carry the same sentences.
- A test asserting the documented list equals the implemented set.
- Verified by deleting `groupBy` from the block: the guard names it.

nstfkc:
Closes #182. Stacked on #181, because `llms.txt` and `llms-full.txt` carry the same sentences and that PR is already rewriting one of them.

`Operation` declares **fifteen** operations. The repository states the count in fourteen places, in three different wrong numbers:

| count | where |
| --- | --- |
| **twelve** | `active-record.ts` (×3), `active-record.test.ts`, `architecture.test.ts`, `orm-rows-and-entities.md`, `plans/` (×2) |
| **thirteen** | `strategy.ts`, `strategy.test.ts`, `llms.txt`, `plans/` |
| **fourteen** | `docs/orm.md` (×2) |

Each was right when written. None moved when `count`, `aggregate` and `groupBy` arrived — twelve plus those three is fifteen.

**I contributed to it.** #144 wrote *"not one of the thirteen operations"* in `strategy.test.ts`, copied out of the comment beside it without checking.

## The list itself is fine

`docs/orm.md`'s code block names all fifteen. Only the prose introducing it says "Fourteen". That is the milder failure — but it is the sentence a reader meets first, and a number that has drifted three times will drift again.

So the counts become phrasing that cannot drift — "every operation", "the operations" — except in `orm.md`'s introduction, where the number does real work and is now correct.

## The guard is on the list, not the number

A wrong count is one word out of date. A wrong **list** is a documented surface that does not exist, or an existing one nobody is told about.

It asserts the code block names exactly the operations the compiler implements, checked against the same predicate-verified set #146 uses — so the list cannot quietly name something that is not an operation either.

**Verified** by deleting `groupBy` from the block: the guard names it (and the verbatim check from #181 catches it too).

## `plans/` is deliberately untouched

That directory records what was believed at each iteration, and says so:

> The claim is left standing above with this correction beneath it rather than quietly rewritten, because the gap between what an invariant promised and what it delivered is the useful part.

- `typecheck` clean, 1372 unit tests, `oxlint orm --deny-warnings` exit 0.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2743,7 +2743,7 @@ generator gemi {
 Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 - `schema.ts` — runtime metadata: tables, columns, types, defaults, relations.
-- `models.ts` — a typed base class per model, carrying the fourteen operations.
+- `models.ts` — a typed base class per model, carrying every operation.
 - `index.ts` — registers every model by name.
 
 **The output is committed on purpose.** Diffs stay reviewable and CI needs no codegen step.
@@ -2792,7 +2792,7 @@ read. It can only see modules you hand it, so it does not replace the `register`
 
 ## Querying
 
-Fourteen operations, with Prisma's argument types verbatim:
+Fifteen operations, with Prisma's argument types verbatim:
 
 ```
 findMany   findFirst   findFirstOrThrow   findUnique   findUniqueOrThrow   count   aggregate   groupBy
@@ -4057,7 +4057,7 @@ There is an `ActiveRecordModel` base in the ORM that makes every query return in
 overriding one method. It ships as a **documented example rather than a supported tier**, for
 exactly the reason above plus one more: the return *types* do not change, so the methods are
 invisible to TypeScript at the call site even though they exist at runtime. Fixing that means
-conditional return types across all twelve operations, which is the complexity the plain-object
+conditional return types across every operation, which is the complexity the plain-object
 default was chosen to avoid.
 
 Use it if you want to see the seam work. Reach for `wrap` if you want the guarantee.
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -28,7 +28,7 @@ Each link below points to a self-contained Markdown page. For a single file cont
 
 ## Data
 
-- [ORM](https://nstfkc.github.io/gemi/orm.md): the Prisma-typed, gemi-executed query layer — the Prisma generator block, the thirteen operations, relations and the batched/lateral strategies, ambient transactions, policies that apply to nested reads, soft deletes.
+- [ORM](https://nstfkc.github.io/gemi/orm.md): the Prisma-typed, gemi-executed query layer — the Prisma generator block, every operation, relations and the batched/lateral strategies, ambient transactions, policies that apply to nested reads, soft deletes.
 - [Rows & Entities](https://nstfkc.github.io/gemi/orm-rows-and-entities.md): plain objects by default, opt-in provenance with track + save, and wrap for methods on a record.
 
 ## Auth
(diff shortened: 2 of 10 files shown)


nstfkc:
## Review — the operation count, documented three ways (closes #182)

`Operation` declares fifteen, and the guard is sensitive in both directions:

```
1372 unit tests
groupBy removed from the block
  × it names exactly the operations the compiler implements   -   "groupBy"
  × orm.md is embedded verbatim                                   (#180's check, also)
```

**Guarding the list rather than the number is right**, and the reason given is the one this series has now shown three times — #145, #149 and #180 were each a documented surface that did not match the code, where the count is only ever one word stale. Checking against the same predicate-verified set #146 uses means the list cannot name a non-operation either, which is the half a hand-written list would lose.

Leaving `plans/` alone is right, and quoting the directory's own rule for why is better than asserting it.

### One stale count remains, and it is the one the enumeration missed

Re-swept every count-word within a line of "operation" on this head:

```
fourteen    0
fifteen     2   docs/orm.md, docs/llms-full.txt        <- both correct now
twelve      3   plans/ ×3                              <- deliberate
thirteen    2   plans/ ×1                              <- deliberate
                packages/gemi/orm/compile/unique.ts    <- not deliberate
```

```
orm/compile/unique.ts:13
  * that is not one of the thirteen. `UnsupportedQueryError` documents those
```

Source, in `orm/`, and the only stale count left outside `plans/`. I named it on #182 before this PR was written, so the miss is in the table rather than in the sweep.

**Which is this PR's own thesis, happening inside it.** The argument for guarding the list is that a number fixed by enumeration will drift again — and these numbers were fixed by enumeration, from the issue's table, and the one occurrence absent from that table survived. The eleven files changed are exactly the eleven listed.

Nothing will catch it either: the new guard reads the docs block against the compiler's set, not source comments. So this one has no check behind it, which is the argument for using the same phrasing-that-cannot-drift here as everywhere else — *"an operation that is not one of them"* — rather than correcting `thirteen` to `fifteen` and waiting for the sixteenth.

Worth folding into this PR rather than a follow-up, since it closes the issue that names the problem and would otherwise close it with an instance still open.

PR #184, opened 2026-07-29
Correct the reasoning left behind by #179

nstfkc:
Based on `feat/orm` directly. Corrects a comment I wrote, not behaviour.

#179 added a generator test for composite relations and justified it on **two claims that are both false**. Re-measured on `feat/orm` at f207120:

| claim | reality |
| --- | --- |
| "the starter template declares no composite relation" | it declares `LedgerEntry.ledger @relation(fields: [tenantId, ledgerCode], references: [tenantId, code])`, with `@@id([tenantId, code])` on `Ledger` |
| "the artifact diff test … a consistently wrong generator satisfies" | truncating `emit.ts` to `slice(0, 1)` and rebuilding **fails** `running prisma generate again produces a zero-line diff` |

The composite relation arrived with the work that merged during this series. The "16 relations, 0 composite" figure I quoted was taken much earlier and I carried it forward without re-running it.

And the "**only the new test notices** — 1367 others pass" measurement was `vitest run orm database`. That is the package suite and **excludes the template**, so it could not have seen the diff test. I reported a scoped result as a general one.

## The test stays

The remaining reason holds and is enough on its own: the compiler's composite tests build their own `ModelSchema` fixtures and never run the generator, so this is the only **direct** check of it. It also fails differently — naming the field that went missing, rather than reporting that regeneration produced a diff, which is slower to read and which a stale committed artifact could mask.

What does not stay is the comment arguing for it, because it would have told the next reader two things that are not true.

Corrections are posted on #178 and #179 as well, so the record is not only in the source.

- `typecheck` clean, 1371 unit tests, `oxlint orm --deny-warnings` exit 0.
- Template suite: 616 on SQLite.

### DIFF
--- a/packages/gemi/bin/orm/emit.test.ts
+++ b/packages/gemi/bin/orm/emit.test.ts
@@ -356,18 +356,24 @@ describe("buildModelSchemas()", () => {
    * **A relation that joins on more than one field**, which every other case
    * here declares with exactly one.
    *
-   * The compiler handles this shape and is well tested for it — but against
-   * *hand-written* fixtures, and the starter template declares no composite
-   * relation, so nothing exercised the generator that produces the real
-   * artifact.
+   * **Correcting the reason this was added (#178, #179).** It was justified on
+   * two claims that are both false: that the starter template declares no
+   * composite relation, and that the artifact's zero-diff test would not catch
+   * a truncating generator. The template declares
+   * `LedgerEntry.ledger @relation(fields: [tenantId, ledgerCode], …)`, and
+   * truncating `emit.ts` to the first field **does** fail that test — the
+   * committed artifact holds the right value, so regenerating diverges from it.
    *
-   * The failure that leaves open is the quiet one. A generator that kept only
-   * the first field would emit `from: ["tenantId"]`, every compiler test would
-   * still pass because they build their own schemas, the template suite would
-   * still pass because it has no such relation — and an application with one
-   * would correlate on a single column and return plausible wrong rows. That is
-   * precisely what the compiler's own refusal was written to prevent, so it is
-   * worth checking that the input reaches it intact.
+   * The measurement behind "only this test notices" was `vitest run orm
+   * database`, which is the package suite and excludes the template. Reported
+   * as though it were general, it was not.
+   *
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — correct the reasoning left behind by #179

Comment-only, one file, `+17 −11`; 1371 unit tests and 35 in `bin/` unchanged. Both corrections match what I measured, and the correction is complete on the two claims it names.

**Recording the scope error explicitly is the part worth keeping.** *"That is the package suite and excludes the template, so it could not have seen the diff test. I reported a scoped result as a general one."* That is the same mistake I made twice on this stack — mutating against one file rather than the suite, which cost two retractions — and naming the mechanism rather than the instance is what makes it transferable. It is now the third correction in this series to come down to *what did the command I ran actually cover*.

**The surviving reason is the right one to keep**, and it is stronger than the two it replaces: the compiler's composite tests build their own fixtures and never run the generator, so this is the only direct check. Failing by naming the field beats failing with a regeneration diff, for the reason given.

### One path the correction leaves out, and it is the one CI takes

The new comment says the zero-diff test catches a truncating generator, and that *"a stale committed artifact could mask"* it. Both true. What is not said is what happens when the artifact is repaired *before* the test runs — which is exactly what CI does, since #170 added `prisma generate` ahead of the suites:

```
truncated generator, artifact left as committed   1 failed   generated.test.ts
truncated generator, artifact regenerated first  10 failed   differential.test.ts
```

So the two checks are mutually exclusive, and the workflow selects the second. The comment now describes the configuration a developer runs locally and not the one that gates the branch — where the differential's composite seed is the backstop, and it is only a backstop because the template happens to declare `LedgerEntry.ledger`. Remove that relation and CI catches neither.

Half a sentence covers it: the diff test catches this locally, the differential catches it in CI, and neither is the direct check that this test is.

That also revises something I said twice — I argued on #161 and #170 that the `prisma generate` step should go. It should not, on its own: removing it restores the diff test and removes the differential's sensitivity. `git diff --exit-code` on `app/models/generated` immediately after the step keeps both.

Nothing outstanding otherwise. Correcting a comment rather than quietly deleting it, and posting the correction on the closed issues too, is the right shape for this.

PR #186, opened 2026-07-29
Setup's generator-ordering reason does not hold

### LINKED ISSUE #185: Setup says the generator block must come after client, and gives a reason that does not hold
`docs/orm.md`'s **Setup** section opens:

> Add the gemi generator to `schema.prisma`, *after* the `client` block — the emitted bases type-import `@prisma/client`, and Prisma runs generators in declaration order.

The premise is true and the conclusion does not follow.

## Measured

- **The generator never imports `@prisma/client`.** `bin/orm-generator.ts` and `bin/orm/emit.ts` import only `@prisma/generator-helper` — the protocol Prisma hands the models over. There is nothing for the client to be ready *for*.
- **The emitted import is a type import.** `import type { Prisma } from "@prisma/client"` is erased at build and only has to resolve when someone typechecks — by which point one `prisma generate` has produced both outputs, whatever order it ran them in.
- **Swapping the blocks produces byte-identical output.** `schema.ts`, `models.ts` and `index.ts` all unchanged, and generation succeeds.

## Why it is worth correcting rather than leaving

The advice is harmless; the *reason* is not. It tells a reader that gemi's output depends on the client existing first, which would make ordering a thing to check when something goes wrong — and it is not. Someone debugging a real generation problem could spend time on it.

I am stating what I measured and what I did not: I did not run this on a machine where `@prisma/client` had never been generated. That cannot change the answer, because the generator does not read the client at all — which is the fact worth pinning rather than the experiment.

## Acceptance

- The sentence corrected to say the order is convention, with the actual reason.
- A test asserting the generator imports nothing from `@prisma/client`, and that the emitted base imports `Prisma` **as a type only** — never as a value, which would make it a runtime dependency. That is the fact the corrected sentence rests on, so if it ever stops being true the ordering advice comes back and this fails first.

nstfkc:
Closes #185. Based on `feat/orm` directly.

`docs/orm.md`'s **Setup** section opens:

> Add the gemi generator to `schema.prisma`, *after* the `client` block — the emitted bases type-import `@prisma/client`, and Prisma runs generators in declaration order.

The premise is true and **the conclusion does not follow**.

## Measured

| | |
| --- | --- |
| the generator never imports `@prisma/client` | `orm-generator.ts` and `emit.ts` import only `@prisma/generator-helper`, the protocol Prisma hands the models over — there is nothing for the client to be ready *for* |
| the emitted import is a **type** import | erased at build; it only has to resolve when someone typechecks, by which point one `prisma generate` has produced both outputs whatever order it ran them in |
| swapping the blocks | produces **byte-identical** `schema.ts`, `models.ts` and `index.ts`, and generation succeeds |

## Why correct it rather than leave it

The advice is harmless; the **reason** is not. It tells a reader that gemi's output depends on the client existing first, which makes ordering a thing to check when something goes wrong — and it is not. Someone debugging a real generation problem could spend time there.

**What I measured and what I did not:** I did not run this on a machine where `@prisma/client` had never been generated. That cannot change the answer, because the generator does not read the client at all — which is the fact worth pinning rather than the experiment.

## The guard

Asserts the generator imports nothing from `@prisma/client`, and that the emitted base imports `Prisma` **as a type only** — never as a value, which would make it a runtime dependency. If either stops being true, the ordering advice comes back and this fails first.

Verified by adding a value import to `orm-generator.ts`: the test names the file.

## Two process notes

- `emitArtifacts` returns an object keyed by filename, not an array. My first draft assumed an array; the run caught it.
- Editing `orm.md` **failed the verbatim guard from #181** until `llms-full.txt` was re-synced — the guard working on its own author, which is the best evidence I can offer that it was worth adding.

- `typecheck` clean, 1374 unit tests, `oxlint orm --deny-warnings` exit 0.
- Template suite: 616 on SQLite.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2726,8 +2726,11 @@ and never ships in a bundle.
 
 ## Setup
 
-Add the gemi generator to `schema.prisma`, *after* the `client` block — the emitted bases
-type-import `@prisma/client`, and Prisma runs generators in declaration order.
+Add the gemi generator to `schema.prisma`. Conventionally after the `client` block, though the
+order does not matter: the generator reads its models from Prisma's generator protocol and never
+imports `@prisma/client` itself, so it emits the same three files either way. The emitted bases do
+`import type { Prisma }`, but that is erased at build and only has to resolve when you typecheck —
+by which point one `prisma generate` has produced both.
 
 ```prisma
 generator client {
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -31,8 +31,11 @@ and never ships in a bundle.
 
 ## Setup
 
-Add the gemi generator to `schema.prisma`, *after* the `client` block — the emitted bases
-type-import `@prisma/client`, and Prisma runs generators in declaration order.
+Add the gemi generator to `schema.prisma`. Conventionally after the `client` block, though the
+order does not matter: the generator reads its models from Prisma's generator protocol and never
+imports `@prisma/client` itself, so it emits the same three files either way. The emitted bases do
+`import type { Prisma }`, but that is erased at build and only has to resolve when you typecheck —
+by which point one `prisma generate` has produced both.
 
 ```prisma
 generator client {
--- a/packages/gemi/bin/orm/emit.test.ts
+++ b/packages/gemi/bin/orm/emit.test.ts
@@ -1,3 +1,6 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
 import type { DMMF } from "@prisma/generator-helper";
 import { describe, expect, test } from "vitest";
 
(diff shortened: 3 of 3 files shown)


nstfkc:
## Review — Setup's generator-ordering reason does not hold (closes #185)

All three claims reproduce independently.

```
orm-generator.ts / emit.ts imports    node:*, @prisma/generator-helper, local  — no @prisma/client
the emitted base                      import type { Prisma } from "@prisma/client"
blocks swapped, regenerated           schema.ts, models.ts, index.ts all byte-identical; generation succeeds
```

I swapped the two `generator` blocks in the template, re-ran `prisma generate`, and `diff -r` over the whole generated directory reported no difference. So the conclusion really does not follow from the premise.

**Both halves of the guard fire, and name the right thing:**

```
value import added to orm-generator.ts   × bin/orm-generator.ts does not import @prisma/client
emitted import made a value import       × the emitted models file imports Prisma as a type only
```

Guarding the *imports* rather than the sentence is the right choice — the sentence is now correct and could go stale again, whereas an import that changes is the thing that would make the ordering advice true.

**Separating "what I measured and what I did not" is the right instinct**, and I hit the same wall from a different direction: my scratch tree shares `node_modules` with the real checkout, so I could not remove a generated client without touching it. The reasoning is what carries this — a process that never reads the client cannot depend on when it was written — and stating that rather than the experiment is correct.

The `emitArtifacts` shape note and the #180 guard catching its own author on the first edit are both worth keeping.

### The same sentence is still in two other places, one of them the file a reader is looking at

`templates/saas-starter/prisma/schema.prisma:10`:

```prisma
// Declared after `client` on purpose: the emitted bases
// type-import `@prisma/client`, and Prisma runs generators in declaration
// order. The output is committed.
```

That is the corrected claim, verbatim, in the schema every new application is generated from — and the one a reader has open while doing the thing the Setup section describes. Correcting `orm.md` and leaving it means the two now disagree.

Third time in this series that a claim fixed by enumeration left the instance outside the enumeration: #182's count sweep missed `unique.ts`, #178's table missed the template's composite relation, and this one misses the schema. The new guard cannot cover it either — it asserts imports, not prose.

**And one worth reconciling rather than deleting.** `bin/orm-generator.ts:23` says:

> The generated `models.ts` type-imports `@prisma/client`, so the client generator **has to have run**.

against the new docs text:

> that is erased at build and **only has to resolve when you typecheck**

Both are about the same fact and they read differently. That comment is also the better argument — it is why `assertPrismaClientGenerator` exists, which checks the client generator is *declared* rather than trusting order, and that assertion is unaffected by any of this. Worth aligning the wording so the file that got it right is not the one that looks stale.

nstfkc:
## Follow-up — how much surface the "corrected in one place" problem actually has

I have now flagged this shape three times in three PRs — #182's sweep missing `unique.ts`, #178's table missing the template's composite relation, and this one missing `schema.prisma`. Rather than flag a fourth, I measured how much of it there is.

Normalising punctuation and looking for **nine-word runs of prose** shared between `docs/orm.md` and any non-generated source file:

```
source files sharing prose with orm.md          44
  packages/gemi/orm/policy.ts                   51 distinct runs
  packages/gemi/orm/Model.ts                    40
  templates/.../foreign-key-indexes.test.ts     38
  packages/gemi/orm/compile/select.ts           35
  packages/gemi/orm/compile/plan-relations.ts   35
  …
  templates/saas-starter/prisma/schema.prisma    the stale one this PR leaves
```

Shared *code examples* were excluded; these are sentences. `policy.ts` and `orm.md` share fifty-one of them.

**That is a feature, not a defect.** The docs quote the implementation's own reasoning, which is why a reader gets the same explanation whichever they open, and it is a large part of what makes this codebase readable. It is also exactly why a corrected claim needs a *sweep* rather than an enumeration: the same sentence is, by design, in more than one place, and the copies are not adjacent.

**One pair is guarded and forty-four are not.** #180 added a verbatim check for `orm.md` ↔ `llms-full.txt`, and it has already caught its own author twice — including on this PR, per your process note. Nothing covers `orm.md` ↔ source comments, and the three misses in three PRs are all in that gap.

I am not proposing a check for it — comparing prose across 44 files would produce more false positives than findings, and this repo's stated preference is that a false positive reads as a real finding. The cheaper answer is procedural and you are already most of the way there: when a claim is corrected, grep the distinctive phrase rather than the enumerated locations. `"runs generators in declaration"` would have returned `schema.prisma` and `orm-generator.ts` in one command, and both are in this PR's blast radius.

Nothing here changes my review above; the schema comment is still the concrete item.

PR #188, opened 2026-07-29
The benchmark report says Postgres both was and was not measured

### LINKED ISSUE #187: The benchmark report says Postgres both was and was not measured
`plans/orm/benchmarks.md` contains, three lines apart:

> 3. **Deliverable 2 (lateral + `json_agg`) was not evaluated**: Postgres was not measured, and it is the only dialect where the case can be made. Set `BENCH_POSTGRES_URL`.

> - Postgres was measured over loopback, so every Postgres round trip here is optimistic — see the note below.

and, in *Where these numbers should not be trusted*:

> The Postgres point read and depth-2 include come out at 0.82× and 0.94× because a ~150µs loopback round trip varies by more than the difference being measured…

The data says the first one is right:

```
$ grep -c "| postgres" plans/orm/benchmarks.md
0
$ grep -o '"dialect": "[a-z]*"' plans/orm/benchmarks.json | sort | uniq -c
   9 "dialect": "sqlite"
```

Nine results, all SQLite, and no Postgres row in any table. The 0.82×/0.94× figures appear nowhere in the data they are presented as summarising.

### Why it happened

The report is generated by `templates/saas-starter/app/models/bench/run.ts`, and its header says "Regenerate rather than editing". One Postgres claim *is* generated conditionally:

```ts
...(anchors.postgres === undefined
  ? ["3. **Deliverable 2 (lateral + `json_agg`) was not evaluated**: Postgres", ...]
  : lateralConclusion(results, anchors.postgres, statementCounts.postgres ?? {})),
```

The rest are static strings in the same array, emitted whether or not Postgres ran:

- the positional-row comparison — "17% faster on Postgres", "p95 near double p50 on Postgres"
- the transaction cost — "+350µs on Postgres"
- the policy overhead — "within run-to-run variance on Postgres"
- the sub-1.0× ratios — "0.82× and 0.94×"
- "Postgres was measured over loopback"
- the `_count` fixture caveat, which explains how to read Postgres rows the table does not contain

### Why it matters

`docs/orm.md` deliberately does not repeat any benchmark number, and says why:

> a number copied into prose drifts from its source the first time the source is re-run, which happened three times inside that report before its own sentences were derived rather than written.

These are the sentences that never made that transition. The report is cited as the evidence in three places — `docs/orm.md` twice (for "what the measurement does **not** establish", and for "the reasoning behind the lateral default") and from source in `lateral.ts` and `strategy.ts` — so a reader who follows the citation lands on a document that contradicts itself, with no way to tell which half is current.

Worth noting the two source citations are *accurate*: both cite statement counts (1, 2, 3 for no-include, depth-2, depth-3, present at `benchmarks.md:84-85`) and both say explicitly "counted rather than timed". The problem is confined to the report's own prose.

### Fix

Gate each of them on whether Postgres produced results, the way the deliverable-2 caveat already is, and bring the committed report in line with the data it was generated from.

The guard belongs on the committed artifacts rather than the generator — `main()` does I/O, and the property worth protecting is the one a reader sees: no table row, and no sentence citing a figure, may name a dialect absent from `benchmarks.json`.

Leaves open: the 0.82×/0.94× ratios are still written rather than derived when Postgres *does* run. Gating removes the contradiction; deriving them from `results` is the other half.

nstfkc:
Closes #187.

`plans/orm/benchmarks.md` claimed both sides of the same question, three lines apart — *"Deliverable 2 (lateral + `json_agg`) **was not evaluated**: Postgres was not measured"* and *"**Postgres was measured over loopback**, so every Postgres round trip here is optimistic"* — and quoted "the Postgres point read and depth-2 include come out at 0.82× and 0.94×", while the data behind it was nine results, every one `"dialect": "sqlite"`, with no Postgres row in any table.

Only the first was generated conditionally. The rest were static strings in `run.ts`'s report array: the positional-row comparison, the transaction cost, the policy overhead, the sub-1.0× ratios, "Postgres was measured over loopback", and the `_count` caveat explaining how to read rows the table does not have.

Each is now gated on `measuredPostgres`, the same way the deliverable-2 caveat already was, and the committed report is regenerated in line with its data. Every remaining mention of Postgres is about its absence.

## Why this one is worth a test

`docs/orm.md` refuses to repeat any benchmark number and says why:

> a number copied into prose drifts from its source the first time the source is re-run, which happened three times inside that report before its own sentences were derived rather than written.

These are the sentences that never made that transition — the same failure, in the document held up as the reason to avoid it. And the report is cited as the evidence three times: `docs/orm.md` for "what the measurement does **not** establish" and for "the reasoning behind the lateral default", plus `lateral.ts` and `strategy.ts` from source.

Both source citations turned out to be **accurate** and are untouched: they cite statement counts (present at `benchmarks.md:84-85`) and both say "counted rather than timed". The defect was confined to the report's prose.

## The guard

`packages/gemi/orm/benchmarks-report.test.ts` asserts the property against the committed artifacts rather than the generator — `main()` does I/O, and what matters is what a reader sees, whichever code path emitted it:

- no table row names a dialect absent from `benchmarks.json`
- no sentence citing a figure (`0.82×`, `+350µs`, `17%`) or saying "was measured" names a dialect with no data — with lines about *absence* excluded by their own wording, since those are what the report should say

**Verified to fail on the report as it stood**, naming the offending line:

```
× no sentence claims a postgres figure unless that dialect ran
+   "  overhead. The Postgres point read and depth-2 include come out at 0.82×",
```

It runs under CI's existing `orm` filter.

## Checks

- `bun --bun vitest run orm database` — 44 files, 1377 passed
- `bunx tsc --noEmit -p tsconfig.json` — clean, both `packages/gemi` and the template

## Not fixed here

The 0.82×/0.94× ratios are still *written* rather than derived when Postgres does run. Gating removes the contradiction; deriving them from `results` is the remaining half, and I have left it out rather than guess at numbers I cannot currently measure.

### DIFF
--- a/packages/gemi/orm/benchmarks-report.test.ts
+++ b/packages/gemi/orm/benchmarks-report.test.ts
@@ -1,0 +1,106 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * `plans/orm/benchmarks.md` and its `.json` have to agree about which dialects
+ * were actually measured.
+ *
+ * The report is generated — its own header says "Regenerate rather than
+ * editing" — and `docs/orm.md` sends readers to it twice, once for "exact
+ * multipliers, the method, and what the measurement does **not** establish"
+ * and once for "the reasoning behind the lateral default". `lateral.ts` and
+ * `strategy.ts` cite it from source as well.
+ *
+ * The committed report claimed both sides of the same question. Three lines
+ * apart it said *"Deliverable 2 (lateral + `json_agg`) **was not evaluated**:
+ * Postgres was not measured"* and *"**Postgres was measured over loopback**,
+ * so every Postgres round trip here is optimistic"*, and it quoted a Postgres
+ * point read and depth-2 include "at 0.82× and 0.94×" — while `benchmarks.json`
+ * held nine results, every one of them `"dialect": "sqlite"`, and the tables
+ * contained no Postgres row at all.
+ *
+ * Only the first of those was generated conditionally. The rest were static
+ * strings in the report array, emitted whether or not Postgres ran — which is
+ * exactly the failure `docs/orm.md` describes as the reason the numbers are not
+ * repeated there: "a number copied into prose drifts from its source the first
+ * time the source is re-run, which happened three times inside that report
+ * before its own sentences were derived rather than written." These were the
+ * sentences that never made that transition.
+ *
+ * A reader cannot resolve a report that contradicts itself, and the half that
+ * is wrong is the confident half. So this asserts the property directly against
+ * the committed artifacts rather than against the generator: whatever code path
+ * emitted a claim, a claim about a dialect that has no data is caught here.
+ */
+const PLANS = join(import.meta.dirname, "../../../plans/orm");
+
+/** The dialects the run actually produced results for. */
+const measured = (() => {
+  const results = JSON.parse(
+    readFileSync(join(PLANS, "benchmarks.json"), "utf8"),
+  ) as Array<{ dialect?: string }>;
+
+  expect(Array.isArray(results), "benchmarks.json is not an array").toBe(true);
+  expect(results.length).toBeGreaterThan(0);
+
+  return new Set(results.map((result) => result.dialect).filter(Boolean));
+})();
+
+const report = readFileSync(join(PLANS, "benchmarks.md"), "utf8");
+
+describe("plans/orm/benchmarks.md reports only dialects it measured", () => {
+  test("the data names at least one dialect", () => {
+    expect(measured.size).toBeGreaterThan(0);
+  });
+
+  /**
+   * Every dialect appearing in a table row has to be one the data knows about.
+   * The rows are the part a reader trusts most, so a row for an unmeasured
+   * dialect is the worst version of this.
+   */
+  test("no table row names a dialect the data does not have", () => {
+    const rows = report
+      .split("\n")
+      .filter((line) => /^\|\s*(sqlite|postgres)\b/.test(line))
+      .map((line) => line.split("|")[1].trim());
+
+    expect(rows.length, "no dialect rows found — the table format moved").toBeGreaterThan(0);
+
+    for (const dialect of new Set(rows)) {
+      expect(measured, `benchmarks.md has ${dialect} rows`).toContain(dialect);
+    }
+  });
+
+  /**
+   * ...and the prose has to hold the same line.
+   *
+   * A line is read as *asserting a measurement* when it names the dialect
+   * alongside a figure (`0.82×`, `+350µs`, `17%`) or says it "was measured".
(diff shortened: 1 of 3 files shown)


nstfkc:
## Review — the benchmark report says Postgres both was and was not measured (closes #187)

The contradiction is real, and the data settles it:

```
plans/orm/benchmarks.json     9 entries, dialect: sqlite ×9, postgres ×0
plans/orm/benchmarks.md:177   "Deliverable 2 … was not evaluated: Postgres…"
                    :195-6   "The Postgres point read and depth-2 include come out at 0.82× and 0.94×"
                    :200     "Postgres was measured over loopback"
```

After the fix, every Postgres mention in the report is about its absence — I checked all five, and they are "require a second run", "not in this run's data", "was not evaluated", "Set `BENCH_POSTGRES_URL`", "was not measured".

The guard reproduces both ways, naming the line:

```
fixed report      4 passed
report as it was  × no sentence claims a postgres figure unless that dialect ran
                    "overhead. The Postgres point read and depth-2 include come out at 0.82×"
```

And it applies to the current head — `feat/orm` has moved to `8b67cce` since this branched, and the three files land clean there: **1377 passed**.

**Asserting against the committed artifacts rather than the generator is the right call**, and the reason is the one that matters: `main()` does I/O, and the property is about what a reader sees, whichever path emitted it. It also means the check survives the report being hand-edited, which is how these sentences got there.

**Excluding absence-lines by their own wording is well judged.** I tried to slip a claim past it with a negation — *"Postgres was measured at 0.82x, not estimated"* — and it was caught. The pattern is narrower than "contains a negation", which is what I expected to find wrong.

### The figure pattern is narrower than the prose it guards

`figure` matches `µs`, `×`, `%` or the literal "was measured". Three ordinary ways to write the same claim slip through, each stating a Postgres number with no Postgres data:

```
- Postgres point reads come out at 0.82x on this hardware.        4 passed
- The Postgres include costs about 150 microseconds per trip.     4 passed
- Postgres depth-2 include runs in 3.4 ms.                        4 passed
```

`ms` and an ASCII `x` are not exotic. They matter here more than they would elsewhere because of what the PR itself says under *Not fixed here* — the ratios are still **written rather than derived**, so the sentences at risk are hand-typed prose, which is exactly where a unit variant appears.

Widening the units closes all three with no false positive on the corrected report:

```js
const figure =
  /\d+(\.\d+)?\s*(µs|us|ms|×|x|%)\b|\d+(\.\d+)?\s*(micro|milli)?seconds?\b|\bwas measured\b/i;
```
```
corrected report   4 passed
each of the three  × no sentence claims a postgres figure unless that dialect ran
```

Measured in all four directions.

**The framing is the best part of this PR** and worth keeping verbatim: these are the sentences that never made the transition `docs/orm.md` describes — *"a number copied into prose drifts from its source the first time the source is re-run"* — inside the very document held up as the reason for the rule. And checking the two source citations and finding them accurate, rather than sweeping them in, is the right restraint.

PR #190, opened 2026-07-29
The docs quote a Postgres provenance cost the benchmarks never measured

### LINKED ISSUE #189: The docs quote a Postgres provenance cost the benchmarks never measured
`docs/orm-rows-and-entities.md`, in *It is off by default, and that is a measurement*:

> Provenance costs a `WeakMap` insert and a snapshot clone per row. On a 1 000-row read that is **about +100% on SQLite and +40% on Postgres** — see `plans/orm/benchmarks.md`.

The cited report's provenance table has exactly one row:

```
| Dialect | Rows | off µs | on µs | added |
| --- | --: | --: | --: | --: |
| sqlite | 1000 | 418.9 | 833.0 | 99% |
```

The SQLite half is right — 99%, quoted as "about +100%". The Postgres half is a number that appears nowhere in the file it cites:

```
$ grep -rn "40%" plans/
$ grep -o '"dialect": "[a-z]*"' plans/orm/benchmarks.json | sort | uniq -c
   9 "dialect": "sqlite"
```

The committed run was SQLite-only — the same root cause as #187, where the report itself was found asserting Postgres figures it had no data for.

### Why it is worth fixing rather than ignoring

The sentence's entire job is to help a reader decide whether to set `track: true`. `+40% on Postgres` reads as "this is much cheaper on your production database" — a conclusion the measurement supports in *neither* direction, because Postgres was never run. A reader who acts on it turns tracking on for a large read on the strength of a number nobody measured.

The page is also embedded verbatim in `docs/llms-full.txt`, which is served at a stable URL for one-shot LLM context, so a tool repeats the figure confidently with no runtime to bump into and correct it.

`docs/orm.md` quotes no benchmark number at all, and says why:

> a number copied into prose drifts from its source the first time the source is re-run, which happened three times inside that report before its own sentences were derived rather than written.

This page *does* quote them, which is defensible — the magnitude is the point of the section — but quoting means the numbers need checking, and nothing was checking them.

### Fix

Drop the unbacked figure, and guard both directions: a figure carrying a unit and attributed to a dialect must name a dialect the report measured, **and** the provenance percentage must match the provenance table's own `added` column within the rounding the prose deliberately uses.

### Checked and found correct, for the record

The rest of the page holds up and should not be touched:

- the `'email' was not fetched by the query this row came from` error text matches `provenance.ts:163`
- `ActiveRecordModel` is exported from `orm/index.ts:2`, so "there is an `ActiveRecordModel` base in the ORM" is true
- the unchanged-row no-op (`provenance.test.ts:325`) and the post-save refresh (`Model.ts:306`) both exist

nstfkc:
Closes #189.

`docs/orm-rows-and-entities.md` told readers that `track: true` costs "**about +100% on SQLite and +40% on Postgres** — see `plans/orm/benchmarks.md`". That report's provenance table has one row:

```
| sqlite | 1000 | 418.9 | 833.0 | 99% |
```

The SQLite half is right. The Postgres half appears nowhere in the file it cites — `grep -rn "40%" plans/` finds nothing — because the committed run was SQLite-only: nine results, all `"dialect": "sqlite"` (the same root cause as #187).

It matters more than a wrong digit. The sentence exists so a reader can decide whether to set the flag, and "+40% on Postgres" reads as *this is much cheaper on your production database* — which the measurement supports in neither direction, since Postgres was never run. The page is embedded verbatim in `docs/llms-full.txt`, so a tool repeats it with nothing to bump into.

## The guard

`docs/orm.md` quotes no benchmark number at all and says why — "a number copied into prose drifts from its source the first time the source is re-run, which happened three times inside that report". This page *does* quote them, which is defensible: the magnitude is the section's point. But quoting means the numbers need checking, so `docs-benchmark-figures.test.ts` checks both directions rather than just deleting the number:

- a figure carrying a unit and attributed to a dialect (`+40% on Postgres`, `833µs on SQLite`) must name a dialect present in `benchmarks.json`
- the provenance percentage must match the provenance table's own `added` column, within 5 points — the rounding the prose deliberately uses, since "about +100%" for a measured 99% is the honest way to quote it and a doc forced to say "+99%" would be stale after any re-run

Plus a test that the docs quote *at least one* figure, so the guard cannot pass by matching nothing.

**Verified to fail on the page as it stood**, naming the quote:

```
× +40% on Postgres names a dialect the report measured
× the provenance figure matches the provenance table

AssertionError: orm-rows-and-entities.md quotes "+40% on Postgres", but
benchmarks.json has no postgres result: expected [ 'sqlite' ] to include 'postgres'
```

The pattern is narrow on purpose. Both pages discuss dialect *behaviour* constantly — "a `Date` binds as milliseconds on SQLite", "fine on SQLite; on Postgres: current transaction is aborted" — and none of that is a measurement. Only a number with a unit is.

`docs/llms-full.txt` is re-synced, so the verbatim-embedding guard from #181 stays green.

## Checked and left alone

The rest of the page holds up: the `was not fetched` error text matches `provenance.ts:163`, `ActiveRecordModel` is exported from `index.ts:2`, and the unchanged-row no-op and post-save refresh both have tests.

## Checks

- `bun --bun vitest run orm database` — 44 files, 1377 passed

Independent of #188 (that one fixes the report's own prose; this one fixes a doc that cites it), so the two can merge in either order.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3956,8 +3956,8 @@ describing exactly what the operation accepts.
 ### It is off by default, and that is a measurement
 
 Provenance costs a `WeakMap` insert and a snapshot clone per row. On a 1 000-row read that is
-**about +100% on SQLite and +40% on Postgres** — see `plans/orm/benchmarks.md`. So it is not a
-flag to set out of habit; on a large read it is a real decision.
+**about +100% on SQLite** — see `plans/orm/benchmarks.md`, which has no Postgres measurement to
+set beside it. So it is not a flag to set out of habit; on a large read it is a real decision.
 
 ### Provenance is tied to object identity
 
--- a/docs/orm-rows-and-entities.md
+++ b/docs/orm-rows-and-entities.md
@@ -49,8 +49,8 @@ describing exactly what the operation accepts.
 ### It is off by default, and that is a measurement
 
 Provenance costs a `WeakMap` insert and a snapshot clone per row. On a 1 000-row read that is
-**about +100% on SQLite and +40% on Postgres** — see `plans/orm/benchmarks.md`. So it is not a
-flag to set out of habit; on a large read it is a real decision.
+**about +100% on SQLite** — see `plans/orm/benchmarks.md`, which has no Postgres measurement to
+set beside it. So it is not a flag to set out of habit; on a large read it is a real decision.
 
 ### Provenance is tied to object identity
 
--- a/packages/gemi/orm/docs-benchmark-figures.test.ts
+++ b/packages/gemi/orm/docs-benchmark-figures.test.ts
@@ -1,0 +1,132 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * A benchmark figure quoted in the docs has to exist in the benchmarks.
+ *
+ * `docs/orm-rows-and-entities.md` said provenance costs "**about +100% on
+ * SQLite and +40% on Postgres** — see `plans/orm/benchmarks.md`". The report's
+ * provenance table has exactly one row:
+ *
+ *     | sqlite | 1000 | 418.9 | 833.0 | 99% |
+ *
+ * The SQLite half is right. The Postgres half is a number that appears nowhere
+ * in the file it cites — `grep -rn "40%" plans/` finds nothing — because the
+ * whole report was generated from a SQLite-only run (see #187).
+ *
+ * `docs/orm.md` refuses to quote a benchmark number at all, and says why:
+ * "a number copied into prose drifts from its source the first time the source
+ * is re-run". This page does quote them, which is defensible — the section's
+ * point is that the default was chosen by measurement, and a reader deciding
+ * whether to set `track: true` wants the magnitude. But quoting means the
+ * numbers need checking, and nothing was checking them.
+ *
+ * The cost is concrete rather than cosmetic: `+40% on Postgres` reads as "this
+ * is cheaper on your production database", which is the opposite of a decision
+ * the report supports. This page is also embedded verbatim in
+ * `docs/llms-full.txt`, so a tool repeats it confidently with nothing to bump
+ * into.
+ */
+const ROOT = join(import.meta.dirname, "../../..");
+
+/** The ORM's two documentation pages — the ones `llms-full.txt` embeds. */
+const PAGES = ["orm.md", "orm-rows-and-entities.md"];
+
+/** The dialects the committed run actually produced results for. */
+const measured = new Set(
+  (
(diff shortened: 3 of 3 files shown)


nstfkc:
## Review — the docs quote a Postgres provenance cost the benchmarks never measured (closes #189)

The claim checks out at every step:

```
base  orm-rows-and-entities.md:52   "about +100% on SQLite and +40% on Postgres"
      grep -rn "40%" plans/          nothing
      benchmarks.json                9 results, dialect: sqlite ×9
```

And both stated failures reproduce with the exact messages:

```
× +40% on Postgres names a dialect the report measured
    orm-rows-and-entities.md quotes "+40% on Postgres", but benchmarks.json
    has no postgres result: expected [ 'sqlite' ] to include 'postgres'
× the provenance figure matches the provenance table
```

**Quoting the number and checking it, rather than deleting it, is the right call**, and the reason given is the one that decides it — the section exists so a reader can weigh `track: true`, and the magnitude is the content. The 5-point tolerance is the detail that makes it survivable: a doc forced to say "+99%" is stale after the next run, and "about +100%" for a measured 99 is how a person would honestly write it.

The rewritten sentence is better than a deletion for a reason worth naming: *"which has no Postgres measurement to set beside it"* tells a reader the comparison is unavailable, where dropping the clause would have left them assuming it was unremarkable.

Checking the surrounding claims and leaving them — the `was not fetched` text against `provenance.ts:163`, the `index.ts:2` export — is the right restraint.

### The pattern catches the sentence that was there, and close variants only

`FIGURE` requires a number, one of `%`/`µs`/`×`, then literally `on <Dialect>`. Three ordinary rewrites of the same false claim pass:

```
On Postgres, provenance adds about +40%.            4 passed
Postgres pays roughly 40% for the snapshot clone.   4 passed
Provenance costs about +40 ms on Postgres.          4 passed
Provenance costs about +40% on Postgres.            2 failed ← only this one
```

**The units half is cheap and safe** — the same gap I raised on #187's sibling guard an hour ago, and the same fix:

```js
const FIGURE = /([+~]?\s*\d+(?:\.\d+)?)\s*(%|µs|us|ms|×|x)\s+on\s+(SQLite|Postgres)/gi;
```
```
corrected pages    4 passed      (no false positive)
"+40 ms on Postgres"  1 failed
```

**The word-order half is not, and the reason is interesting enough to record.** I tried an order-agnostic version and it fails on *this PR's own corrected sentence* — because that sentence names Postgres precisely to say it was not measured, so a figure and a dialect appear together in a line that is correct. Any bidirectional matcher needs #187's absence-exclusion to work, which is an argument for the two guards sharing one, rather than for widening this one alone.

So: take the units, leave the word order, and if the two benchmark guards are ever merged, that is where the absence rule earns its second use.

PR #192, opened 2026-07-29
The 'policies' link in the rows-and-entities page goes to route middleware

### LINKED ISSUE #191: The 'policies' link in the rows-and-entities page goes to route middleware
`docs/orm-rows-and-entities.md:107`, in *Everything else still applies*:

> `save` compiles through the same path as any other write, so [policies](./authorization.md) scope it, `@updatedAt` stamps it, an open transaction contains it…

The link resolves — `docs/authorization.md` exists — so a link checker sees nothing wrong. The problem is what is on the other end.

`authorization.md` is about **route-level** authorization: role middleware, applying it to routes, shared authorization logic, authorization errors. It contains the word "policy" exactly once, at line 110, and only to say the opposite of what the reader came for:

> gemi has no `Gate` facade or policy auto-discovery.

Model policies — the thing that actually scopes a `save` — are documented in `docs/orm.md#policies` (line 750), with a subsection at line 837 on how they propagate through relations.

So a reader working out what constrains their `save` clicks **policies** and lands on HTTP middleware.

### Why it reads as a slip rather than a decision

`docs/orm.md:1211` links the same file, correctly, in *Related*:

> **[Authorization](./authorization.md)** — route-level authorization, which policies complement

That sentence draws exactly the distinction the other link misses, so the split between the two pages is understood; one link just points the wrong way.

### Fix

Retarget it to `./orm.md#policies`, and guard it.

Worth noting a generic link checker would not have caught this and should not be the fix on its own — I wrote one to be sure. Both ordinary checks (target file exists, anchor resolves to a real heading) **pass** on the broken link. What is wrong is that the label promises a concept the target does not document, so the check has to read the label:

```
× a link labelled 'policies' goes to the section documenting them

AssertionError: orm-rows-and-entities.md sends "policies" to authorization.md,
which has no Policies section
```

The two ordinary checks are still worth adding alongside it — `orm.md` has four in-page anchors that nothing currently verifies.

nstfkc:
Closes #191.

`docs/orm-rows-and-entities.md` said, of `save`:

> `save` compiles through the same path as any other write, so [policies](./authorization.md) scope it

The link resolves, so a link checker sees nothing wrong. But `authorization.md` documents **route-level** role middleware, and the only time it says "policy" is to state that gemi has *no* policy auto-discovery. Model policies — the thing that actually scopes a `save` — are `docs/orm.md#policies`, a section of its own with a subsection on how they propagate through relations.

So a reader working out what constrains their `save` clicked **policies** and landed on HTTP middleware.

`orm.md:1211` links the same file correctly — "route-level authorization, which policies complement" — which is what makes this a slip rather than a disagreement about where the content belongs.

## The guard, and why it reads the label

`docs-links.test.ts` adds three checks. Two are the ordinary ones: every relative target exists, and every anchor resolves to a real heading. Both are worth having — `orm.md` has four in-page anchors nothing currently verifies — but **neither would have caught this**, and I confirmed that rather than assuming it. Against the broken link:

```
  ✓ [policies](./authorization.md) points at a file that exists
  ✓ ...all anchors resolve
  × a link labelled 'policies' goes to the section documenting them

AssertionError: orm-rows-and-entities.md sends "policies" to authorization.md,
which has no Policies section
```

The file was there and the page had headings. What was wrong is that the link promised a concept the target does not document, and only a check that reads the *label* — the promise the reader is acting on — can see that.

Deliberately not a general "every label must be a heading somewhere" rule: most link text is a sentence fragment, and that check would be noise rather than a guard. `policies` is pinned because it is the concept most likely to be sent to `authorization.md` by reflex — that page owns the word for the HTTP layer while the ORM's meaning of it lives in `orm.md`.

`docs/llms-full.txt` is re-synced so the verbatim-embedding guard from #181 stays green.

## Checks

- `bun --bun vitest run orm database` — 44 files, 1394 passed

Independent of #188 and #190; the three touch different files and can merge in any order.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -4014,7 +4014,7 @@ comes back too.
 
 ### Everything else still applies
 
-`save` compiles through the same path as any other write, so [policies](./authorization.md)
+`save` compiles through the same path as any other write, so [policies](./orm.md#policies)
 scope it, `@updatedAt` stamps it, an open transaction contains it, and two saves touching the
 same columns share one cached plan.
 
--- a/docs/orm-rows-and-entities.md
+++ b/docs/orm-rows-and-entities.md
@@ -104,7 +104,7 @@ comes back too.
 
 ### Everything else still applies
 
-`save` compiles through the same path as any other write, so [policies](./authorization.md)
+`save` compiles through the same path as any other write, so [policies](./orm.md#policies)
 scope it, `@updatedAt` stamps it, an open transaction contains it, and two saves touching the
 same columns share one cached plan.
 
--- a/packages/gemi/orm/docs-links.test.ts
+++ b/packages/gemi/orm/docs-links.test.ts
@@ -1,0 +1,111 @@
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * Where the ORM's documentation sends readers, checked.
+ *
+ * `docs/orm-rows-and-entities.md` said, of `save`:
+ *
+ * > `save` compiles through the same path as any other write, so
+ * > [policies](./authorization.md) scope it
+ *
+ * The link resolves — `authorization.md` exists — so nothing was wrong with it
+ * in the way a link checker understands. But that page documents route-level
+ * role middleware, and the only time it says "policy" is to state that gemi has
+ * **no** policy auto-discovery. Model policies are in `docs/orm.md#policies`, a
+ * section of its own.
+ *
+ * So a reader learning what `save` is scoped by clicked "policies" and landed
+ * on HTTP middleware. `orm.md` links the same file correctly two hundred lines
+ * later — "route-level authorization, which policies complement" — which is
+ * what makes this a slip rather than a disagreement about where the content
+ * belongs.
+ *
+ * Hence the third test below. The first two are the ordinary checks, and are
+ * worth having for the anchors, but neither would have caught this: the target
+ * existed and the page had headings. What was wrong is that the link promised a
+ * concept the target does not document, and only a check that reads the label
+ * can see that.
+ */
+const DOCS = join(import.meta.dirname, "../../../docs");
+
+/** The ORM's own pages — the two `llms-full.txt` embeds. */
+const PAGES = ["orm.md", "orm-rows-and-entities.md"];
+
+/** GitHub's heading slug: lowercased, punctuation dropped, spaces hyphenated. */
+const slug = (heading: string) =>
+  heading
+    .trim()
+    .toLowerCase()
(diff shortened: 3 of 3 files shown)


nstfkc:
## Review — the 'policies' link goes to route middleware (closes #191)

The finding checks out at every step:

```
authorization.md headings   Role-based access, Defining role middleware, Applying it,
                            Shared authorization logic, Authorization errors, Related
its only "policy" mention   :110 "gemi has no `Gate` facade or policy auto-discovery"
orm.md:750                  ## Policies          <- what the reader wanted
```

And the claim that the ordinary checks would not have caught it reproduces exactly — against the broken link, **only** the label check fires:

```
✓ points at a file that exists
✓ all anchors resolve
× a link labelled 'policies' goes to the section documenting them
    orm-rows-and-entities.md sends "policies" to authorization.md,
    which has no Policies section
```

**Reading the label is the insight**, and stating why the general form is not worth having — most link text is a sentence fragment — is what keeps it a guard rather than noise. Confirming the two ordinary checks *don't* catch it, rather than assuming, is what makes the third one's existence justified.

**The fix is now itself protected**, which I checked because it introduces an anchor dependency where there was none. Renaming `## Policies` in `orm.md`:

```
× [policies](./orm.md#policies) points at a heading that exists
× a link labelled 'policies' goes to the section documenting them
```

Both halves fire. The link cannot rot in either direction.

### What widening the scan would cost — measured, since the scope was chosen without it

`PAGES` is the two ORM pages. Across all of `docs/` there are 212 relative links:

```
targets that do not exist      0        <- widening this check is free
anchors that do not resolve    7        <- README, broadcasting ×2, cli ×2, configuration, project-structure
'policies'-labelled links      1        <- the corrected one; no straggler anywhere
```

So two of the three checks could widen at no cost, and the label check especially: the reflex it guards is *more* likely in a page thinking about HTTP than in the ORM pages, and there is currently nothing for it to trip on. The anchor check is the one that should stay narrow — seven pre-existing failures in five other pages is the "red for someone else's reasons" trade #163 and #174 already settled.

Those seven are real, not slug-mismatch artefacts. `./configuration.md#apppreloadts` is linked from `cli.md:28` and `project-structure.md:77`, and `configuration.md` has no heading that slugs to it. Outside this series, but worth knowing they exist rather than discovering them the day someone widens the scan.

**One thing to note before that day.** `headings()` matches `^#+\s+` without tracking fences, so a `#` comment inside a code block counts as a heading — `configuration.md` has several (`# openssl rand …`, `# RESEND_API_KEY=`). Harmless today: both ORM pages have **zero** such lines, so the current scope is unaffected. It would matter on a widened scan, where an anchor could "resolve" to a shell comment — a false pass rather than a false failure, which is the worse direction.

PR #194, opened 2026-07-29
The benchmark report's sub-1.00× ratios are still quoted from memory

### LINKED ISSUE #193: The benchmark report's sub-1.00× ratios are still quoted from memory
The half #187 named and #188 deliberately left open.

#188 gated the report's Postgres sentences on whether Postgres ran, so the report no longer says "Postgres was measured over loopback" three lines under "Postgres was not measured". But one of the sentences it gated still carries its numbers:

```
overhead. The Postgres point read and depth-2 include come out at 0.82×
and 0.94× because a ~150µs loopback round trip varies by more than the
difference being measured…
```

Gating decides *whether* that sentence prints. It does not make the two ratios belong to the run that prints them — a run that measured Postgres would be handed `0.82×` and `0.94×` whatever it actually measured. This is the last written-rather-than-derived measurement left in `bench/run.ts`, and the report's header says "Regenerate rather than editing", which only means something if regenerating changes the numbers.

### Second problem: it is Postgres-only for no reason

The caveat is about ratios below 1.00× — gemi appearing *faster* than hand-written SQL, which the report says is never a real result. Nothing about that is dialect-specific. A SQLite scenario that dips under 1.00× is the same observation, and a reader seeing only Postgres named is left to guess whether SQLite is exempt.

### Fix

Compute the list from `results`: every scenario whose `gemi.total.p50 / raw.p50` is below 1, named with its actual ratio.

No committed result is below 1.00× — the nine range from 1.18× to 1.98× — so the generated report is byte-identical today and this changes nothing until Postgres is actually measured. The derivation was checked against synthetic input rather than assumed: handed results at those ratios it produces

```
postgres point read by pk at 0.82×, postgres depth-2 include, 100 parents at 0.94×
```

which is the sentence it replaces, reconstructed from the data that sentence claimed to describe.

### Guard

Extend #188's test rather than adding another, because the dialect check there is necessary and not sufficient — it catches a figure attributed to a dialect with no data, not a stale figure attributed to the right one.

Restricted to ratios below 1.00×: that is the one conclusion the report explicitly says is never real, so a sub-1 ratio in prose is either a live measurement worth reading or a leftover. Table cells are skipped, since they *are* the data and checking them against it proves nothing.

nstfkc:
Closes #193. The follow-up #188 named and deliberately left open.

#188 gated the report's Postgres sentences on whether Postgres ran, so it no longer says "Postgres was measured over loopback" three lines under "Postgres was not measured". But one gated sentence still carried its numbers:

> The Postgres point read and depth-2 include come out at **0.82×** and **0.94×**

Gating decides *whether* that prints. It does not make the ratios belong to the run printing them — a run that did measure Postgres would be handed those two figures whatever it measured. The gate fixes the contradiction; only deriving the figure fixes the figure. This was the last written-rather-than-derived measurement in `bench/run.ts`, and "Regenerate rather than editing" only means something if regenerating changes the numbers.

Deriving also stops the caveat being Postgres-only. It is about ratios below 1.00× — gemi appearing *faster* than hand-written SQL, which the report says is never real — and nothing about that is dialect-specific. A SQLite scenario that dips under 1.00× now gets named the same way.

## The report is unchanged, and that is checked rather than assumed

No committed result is below 1.00× — the nine range from 1.18× to 1.98× — so the generated output is byte-identical and `plans/orm/benchmarks.md` is not in this diff. That is evidence the refactor is faithful, not evidence it does nothing: run against synthetic results at the old ratios, the derivation produces

```
postgres point read by pk at 0.82×, postgres depth-2 include, 100 parents at 0.94×
```

— the sentence it replaces, reconstructed from the data that sentence claimed to describe.

## Guard

Extends #188's test rather than adding another, because the dialect check there is necessary and not sufficient: it catches a figure attributed to a dialect with **no** data, not a stale figure attributed to the right one.

Restricted to ratios below 1.00×, which is the precise claim worth pinning — that is the one conclusion the report says is never real, so a sub-1 ratio in prose is either a live measurement or a leftover. Ratios above 1 are ordinary results and quoting them is not this failure mode. Table cells are skipped: they *are* the data, so checking them against it proves nothing.

Verified by re-injecting the original sentence:

```
× every sub-1.00× ratio in the prose is one the run measured

AssertionError: the report quotes 0.82×, which no result in benchmarks.json
produces: expected [ '1.88', '1.18', '1.98', '1.80' ] to include '0.82'
```

## Checks

- `bun --bun vitest run orm database` — 1381 passed
- `bunx tsc --noEmit -p tsconfig.json` — clean, both `packages/gemi` and the template

Touches `benchmarks-report.test.ts` (added by #188, now merged) and `bench/run.ts`; independent of #190 and #192.

### DIFF
--- a/packages/gemi/orm/benchmarks-report.test.ts
+++ b/packages/gemi/orm/benchmarks-report.test.ts
@@ -103,4 +103,47 @@ describe("plans/orm/benchmarks.md reports only dialects it measured", () => {
       expect(claims, `${dialect} has no data, but the report states figures for it`).toEqual([]);
     },
   );
+
+  /**
+   * ...and a ratio the prose quotes has to be a ratio the run produced.
+   *
+   * The dialect check above is necessary and not sufficient. The report said
+   * "the Postgres point read and depth-2 include come out at **0.82× and
+   * 0.94×**"; gating that sentence on whether Postgres ran stops it appearing
+   * beside "Postgres was not measured", but a run that *did* measure Postgres
+   * would still have been handed those two numbers whatever it measured. The
+   * gate fixes the contradiction; only deriving the figure fixes the figure.
+   *
+   * Restricted to ratios **below 1.00×**, which is the precise claim worth
+   * pinning: "gemi came out faster than hand-written SQL here" is the one
+   * conclusion the report explicitly says is never real, so a sub-1 ratio in
+   * prose is either a live measurement worth reading or a leftover. Ratios
+   * above 1 are ordinary results and quoting them is not this failure mode.
+   *
+   * Table cells are skipped — they *are* the data, so checking them against
+   * the data they were rendered from proves nothing.
+   */
+  test("every sub-1.00× ratio in the prose is one the run measured", () => {
+    const ratios = new Set(
+      (
+        JSON.parse(
+          readFileSync(join(PLANS, "benchmarks.json"), "utf8"),
+        ) as Array<{ raw?: { p50: number }; gemi: { total: { p50: number } } }>
+      )
+        .filter((result) => (result.raw?.p50 ?? 0) > 0)
+        .map((result) => (result.gemi.total.p50 / result.raw!.p50).toFixed(2)),
+    );
+
+    const quoted = report
+      .split("\n")
+      .filter((line) => !line.trimStart().startsWith("|"))
+      .flatMap((line) => [...line.matchAll(/\b0\.\d{2}×/g)].map((m) => m[0]));
+
+    for (const ratio of quoted) {
+      expect(
+        [...ratios],
+        `the report quotes ${ratio}, which no result in benchmarks.json produces`,
+      ).toContain(ratio.replace("×", ""));
+    }
+  });
 });
--- a/templates/saas-starter/app/models/bench/run.ts
+++ b/templates/saas-starter/app/models/bench/run.ts
@@ -150,6 +150,35 @@ async function main() {
   // not have run is a bug in the generator, not in the report.
   const measuredPostgres = anchors.postgres !== undefined;
 
+  /**
+   * Scenarios that came out *below* hand-written SQL, named from `results`
+   * rather than remembered.
+   *
+   * The sentence about them was the last written-rather-than-derived
+   * measurement in this file. #187 gated it on whether Postgres ran, which
+   * stopped it contradicting a SQLite-only run — but a run that *did* measure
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — the sub-1.00× ratios are still quoted from memory (closes #193)

The distinction this rests on is the right one and worth stating as plainly as it is here: **gating decides whether a sentence prints; it does not make the number belong to the run printing it.** #187 removed the contradiction and left the figure, and a Postgres run would have been handed `0.82×` and `0.94×` whatever it measured.

Verified the surrounding claims:

```
committed results with a raw baseline    4 of 9
their gemi/raw p50 ratios                1.88, 1.18, 1.98, 1.80
any below 1.00                           no
plans/orm/benchmarks.md in this diff     no
```

So the output really is byte-identical and the report is legitimately absent from the change. (Small thing: "the nine range from 1.18× to 1.98×" is four of the nine — the other five carry no `raw` baseline to form a ratio from. The substance holds.)

**The guard works in both directions**, which is the part I checked rather than assumed:

```
report as committed                                    5 passed
original sentence re-injected                          × no sentence claims a postgres figure …
                                                       × every sub-1.00× ratio … is one the run measured
a sub-1 result added to benchmarks.json, quoted        5 passed
```

That last row is the one that matters for the first real Postgres run — a genuine 0.82× has to be quotable, and it is. Restricting to sub-1.00 is well judged for the reason given: above 1.00 is an ordinary result, below it is the one conclusion the report says is never real, so a stray one is either live or a leftover.

Generalising the caveat off Postgres is right too. It was never a dialect fact.

### The third guard in this file with the same character class

I re-injected the sentence with an ASCII `x` first and it passed — my probe was wrong, not the guard. But it makes the pattern concrete:

```
"… come out at 0.82× and 0.94×."     both checks fire
"… come out at 0.82x and 0.94x."     5 passed
```

Same claim, one character apart. This is now the third guard in `benchmarks-report.test.ts` and its sibling to key on `%|µs|×` — I raised it on #187 and #189, and this PR extends the same file without widening it. The units are worth fixing once, in the shared place, rather than a fourth time:

```js
(%|µs|us|ms|×|x)
```

I verified on #189 that widening the units alone introduces no false positive on the corrected pages; the word-order half is the one that needs care, for the reason I gave there.

It matters slightly more here than in the earlier two, because the sentence this PR removes was itself hand-typed prose — and hand-typed prose is where `x` gets used instead of `×`.

PR #196, opened 2026-07-29
Three more places the operation count never moved

### LINKED ISSUE #195: Three more places the operation count never moved
#182 found the ORM's operation count wrong in three places at once — "twelve" in `active-record.ts`, "thirteen" in `strategy.ts` and `llms.txt`, "fourteen" in `docs/orm.md` — against a union of **fifteen**. #183 fixed those three and put the guard on the *list* rather than the number:

> So the guard is on the **list**, not the number: a count can only be wrong by one word, while a missing operation is a documented surface that does not exist, or an existing one nobody is told about.

That is right about severity and wrong about recurrence. Three more survived the sweep, because they say "operations" without listing any and are therefore invisible to a list-based guard:

```
packages/gemi/bin/orm/emit.ts:439   the twelve operations return plain objects
plans/orm/README.md:218             zero changes to the twelve operations
plans/orm/README.md:491             the thirteen operations
```

Same drift from the same cause: each was right when written, and none moved when `count`, `aggregate` and `groupBy` arrived.

### The one in `emit.ts` is load-bearing

It is not a passing mention — it is the argument for why a deliberately imprecise generated type is safe:

> The honest cost: `new UserModel()` now type-checks as having columns it does not have. These classes are constructed by `wrap` and by nothing else — the twelve operations return plain objects — so the lie is confined to a constructor call no application has a reason to make.

The conclusion is reached by enumerating what returns plain objects. A reader checking that reasoning against the fifteen operations that exist finds three unaccounted for, and cannot tell whether the argument still holds or whether the newer operations are the exception.

### Deliberately not fixed

`plans/orm/08-eloquent-doorway.md:89` also says "twelve operations". That file records what was planned at iteration 8, where twelve was true. Correcting it would be rewriting a dated record to say something it did not say.

`plans/orm/README.md` **is** in scope: it is a living document, written in the present tense and updated as the design moves — line 496 reads "Both are now in all four".

### Fix

Correct the three, and guard the number as well as the list.

The guard derives its expected count from the operations block in `docs/orm.md`, which `docs-scope.test.ts` already pins to the compiler's own predicates — so the two chain rather than duplicating a hand-maintained list, and adding an operation updates both.

The list guard stays. Neither subsumes the other: a count can be right while an operation is missing from the list, and the list can be complete while three files disagree about how many it holds.

nstfkc:
Closes #195.

#182 found the operation count wrong in three places at once; #183 fixed them and guarded the **list** rather than the number, reasoning that "a count can only be wrong by one word, while a missing operation is a documented surface that does not exist".

Right about severity, wrong about recurrence. Three more survived that sweep, because they say "operations" without listing any and so are invisible to a list-based guard:

```
packages/gemi/bin/orm/emit.ts   the twelve operations return plain objects
plans/orm/README.md:218         zero changes to the twelve operations
plans/orm/README.md:491         the thirteen operations
```

Same drift, same cause: each was right when written, and none moved when `count`, `aggregate` and `groupBy` arrived.

## The `emit.ts` one is load-bearing

It is the argument for why a deliberately imprecise generated type is safe:

> These classes are constructed by `wrap` and by nothing else — the twelve operations return plain objects — so the lie is confined to a constructor call no application has a reason to make.

The conclusion is reached by enumerating what returns plain objects. A reader checking it against the fifteen that exist finds three unaccounted for, and cannot tell whether the argument still holds or whether the newer operations are the exception.

## Deliberately not fixed

`plans/orm/08-eloquent-doorway.md:89` also says "twelve operations" and is left alone: it records what was planned at iteration 8, where twelve was true, and correcting it would be rewriting a dated record to say something it did not say.

`plans/orm/README.md` **is** in scope — a living document, present tense, updated as the design moves (line 496: "Both are now in all four"). I also left its "a future `ActiveRecordModel` base" wording untouched: that is a separate claim about framing, not about the count.

## The guard

It derives the expected count from the operations block in `docs/orm.md`, which `docs-scope.test.ts` already pins to the compiler's own predicates — so the two chain rather than duplicating a hand-maintained list, and adding an operation updates both.

Written before the fix, and it reported exactly the three:

```
× packages/gemi/bin/orm/emit.ts: "twelve operations"
× plans/orm/README.md: "twelve operations"
× plans/orm/README.md: "thirteen operations"

AssertionError: packages/gemi/bin/orm/emit.ts says "twelve operations",
but the ORM implements 15: expected 'twelve' to be 'fifteen'
```

...while passing the two that are already right (`docs/orm.md`, `llms-full.txt`).

The list guard stays: neither subsumes the other, since a count can be right while an operation is missing from the list, and the list can be complete while three files disagree about how many it holds. `.test.ts` files are excluded from the sweep, because a test about a wrong number has to be able to quote it.

## Checks

- `bun --bun vitest run orm database` — 1391 passed
- the `emit.ts` change is a source comment, not emitted: `grep "operations" templates/saas-starter/app/` finds nothing, and the artifact tests pass

Independent of #192 and #194.

### DIFF
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -436,7 +436,7 @@ function wrapOperation(model: string): string {
  *
  * The honest cost: `new UserModel()` now type-checks as having columns it does
  * not have. These classes are constructed by `wrap` and by nothing else — the
- * twelve operations return plain objects — so the lie is confined to a
+ * fifteen operations return plain objects — so the lie is confined to a
  * constructor call no application has a reason to make.
  */
 function instanceShape(model: string): string {
--- a/packages/gemi/orm/operation-count.test.ts
+++ b/packages/gemi/orm/operation-count.test.ts
@@ -1,0 +1,122 @@
+import { readFileSync, readdirSync, statSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * Wherever the ORM's operation count is spelled out, it has to be the count.
+ *
+ * #182 found the number wrong in three places at once — "twelve" in
+ * `active-record.ts`, "thirteen" in `strategy.ts` and `llms.txt`, "fourteen" in
+ * `docs/orm.md` — against a union of fifteen. #183 fixed those three and put a
+ * guard on the **list** rather than the number, reasoning that "a count can only
+ * be wrong by one word, while a missing operation is a documented surface that
+ * does not exist".
+ *
+ * That reasoning was right about severity and wrong about recurrence. Three more
+ * survived the sweep, because they say "operations" without listing any:
+ *
+ *   packages/gemi/bin/orm/emit.ts   the twelve operations return plain objects
+ *   plans/orm/README.md:218         zero changes to the twelve operations
+ *   plans/orm/README.md:491         the thirteen operations
+ *
+ * Each was right when written and none moved when `count`, `aggregate` and
+ * `groupBy` arrived — the same drift, from the same cause, in the places the
+ * list-based guard cannot see. The one in `emit.ts` is load-bearing: it is the
+ * argument for why a deliberately imprecise instance type is safe, and it
+ * reaches that conclusion by enumerating what returns plain objects.
+ *
+ * So this guards the number, and the list guard stays. Neither subsumes the
+ * other: a count can be right while an operation is missing from the list, and
+ * the list can be complete while three files disagree about how many it holds.
+ */
+
+/** The count comes from the block `docs-scope.test.ts` pins to the compiler. */
+const IMPLEMENTED_COUNT = (() => {
+  const source = readFileSync(
+    join(import.meta.dirname, "../../../docs/orm.md"),
+    "utf8",
+  );
+  const at = source.search(/^\w+ operations, with Prisma's argument types verbatim:$/m);
+  expect(at, "the operations block moved or was reworded").toBeGreaterThan(0);
+  const fence = source.indexOf("```", at);
+  const end = source.indexOf("```", fence + 3);
+  return source
+    .slice(fence + 3, end)
+    .split(/\s+/)
+    .filter(Boolean).length;
+})();
+
+const NUMBER_WORDS = [
+  "zero", "one", "two", "three", "four", "five", "six", "seven", "eight",
+  "nine", "ten", "eleven", "twelve", "thirteen", "fourteen", "fifteen",
+  "sixteen", "seventeen", "eighteen", "nineteen", "twenty",
+];
+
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — three more places the count never moved (closes #195)

The diagnosis is right and the correction to #182's reasoning is the useful part: *"right about severity, wrong about recurrence"*. A list guard cannot see a sentence that says "operations" without listing any, and that is exactly the shape all of these take.

Verified the three: `emit.ts` no longer says "twelve", and `plans/orm/README.md`'s lines 218 and 491 now say "fifteen". The `emit.ts` one is the load-bearing one for the reason given — its argument is *reached by enumerating* what returns plain objects, so a reader checking it against fifteen finds three unaccounted for and cannot tell whether the conclusion survives.

Deriving the expected count from the operations block that `docs-scope.test.ts` already pins to the compiler's predicates is the right chaining. Keeping both guards, with the reason why neither subsumes the other, is right too.

### Two more survive, for the same structural reason

`COUNT` is `\b(<word>)\s+operations\b` — a count word **immediately** followed by the plural noun. Two phrasings in the tree are not that:

```
packages/gemi/orm/compile/unique.ts:12
  … the child's model, and an operation that is not one of the thirteen.

plans/orm/README.md:186
  `Model.$exec(op, args)` is the only place that touches the database.
  The twelve public operations are one-line delegations to it.
```

The first puts the number after the noun and in the singular; the second puts an adjective between. The guard passes — **7 passed** on this head — with both still there.

`README.md:186` is in scope by this PR's own rule, and it is not an incidental line: it is invariant 1 under *"Every iteration must preserve these"*, which is the most load-bearing prose in `plans/`. `unique.ts:12` is the one I raised on #182 before #183 was written and again on #45; this is the third sweep it has survived, each time because the pattern and the phrasing did not line up.

**A widened pattern needs care — I got it wrong first.** Matching any count word within 60 characters of "operation" floods, because `NUMBER_WORDS` contains "one" and "three" and prose is full of them near that noun. Allowing up to two intervening words is also too loose: it turns *"not one of the operations"* — the drift-proof wording #185 deliberately introduced — into a claim that there is one.

Two narrow additions do it, and I checked them across `packages/gemi`, `docs/` and `plans/orm/README.md`:

```js
/\b(WORDS)\s+(?:\w+\s+)?operations\b/i      // at most one adjective
/\bone of the\s+(ELEVEN_AND_UP)\b/i         // the idiom, real counts only
```
```
matches that disagree with "fifteen":  2
  compile/unique.ts     thirteen   "one of the thirteen"
  plans/README.md:186   twelve     "twelve public operations"
false positives:        0
```

Restricting the second to eleven-and-up is what keeps *"one of the operations"* out of it.

Fourth time in this series that a guard's pattern has been narrower than the prose it protects — after the three benchmark-figure unit sets. Worth noticing as the shape rather than the instance: each was written against the sentence that was wrong, and each passes on the next phrasing of the same mistake.

PR #198, opened 2026-07-29
The design record still calls shipped API future

### LINKED ISSUE #197: The design record still calls shipped API future
`plans/orm/README.md` is the ORM's design record — where the six invariants come from, and cited by name from `strategy.ts`, `lateral.ts`, `nested-writes.ts` and `order-relation.ts`. It is a living document, not a dated one: it carries a *"Where it actually is now"* section, and line 496 reads "Both are now in all four".

Two of its invariants still describe shipped API as forthcoming.

**Invariant 3** (line 216):

> Subclassing is the extension mechanism: **a future `ActiveRecordModel`** base overrides `$shape` to build instances…

**Invariant 5** (lines 283, 289):

> `$shape` is the place a `WeakMap<row, { model, pk, snapshot }>` **would be** populated, enabling **a future `User.save(row)`** that diffs against the snapshot…
>
> Iterations 1–7 only need the seam to exist; **iteration 8 fills it in.**

Both shipped:

- `ActiveRecordModel` is exported from `packages/gemi/orm/index.ts:2`
- `Model.save` is a static at `packages/gemi/orm/Model.ts:204`
- `docs/orm-rows-and-entities.md` documents `track`/`save`/`wrap` as three working levels

And the same document's own status table, line 33, says iteration 8 landed:

| PR | Branch | What it adds |
| --- | --- | --- |
| #53 | `feat/orm-08-eloquent-doorway` | opt-in provenance, `save(row)`, `wrap`, the ORM auth adapter |

### Why it is worth fixing

The document disagrees with itself, and the half a reader is more likely to act on is the invariant. Someone reading invariant 5 to find out whether row provenance exists is told it is a seam awaiting a later iteration — and would go and build what is already there. The status table is a snapshot near the top; the invariants are the part quoted from source comments.

### Fix

Put the two claims in the present tense, and guard the class.

It is guardable rather than merely fixable because the claim names the API: a sentence calling `X` future can be checked against whether `X` is exported from `gemi/orm` or is a static on `Model`.

The pattern matches only ``a future `identifier` ``. "every future tier becomes a rewrite" sits two lines from one of these and names no API — matching loosely on the word would flag it and teach the next person to route around the guard.

### Note

The invariant-3 line is two lines from #196's change, so whichever merges second needs a one-line rebase. Kept as separate findings rather than merged into one PR.

nstfkc:
Closes #197.

`plans/orm/README.md` is the ORM's design record — where the invariants come from, cited by name from `strategy.ts`, `lateral.ts`, `nested-writes.ts` and `order-relation.ts`. It is a *living* document: it has a "Where it actually is now" section, and line 496 reads "Both are now in all four".

Two invariants still described shipped API as forthcoming:

```
invariant 3   a future `ActiveRecordModel` base overrides `$shape`
invariant 5   enabling a future `User.save(row)` that diffs against the snapshot …
              Iterations 1–7 only need the seam to exist; iteration 8 fills it in.
```

`ActiveRecordModel` is exported from `orm/index.ts:2`; `Model.save` is a static at `Model.ts:204`; and the document's own status table (line 33) records iteration 8 as landed — "#53 … opt-in provenance, `save(row)`, `wrap`".

So it contradicts itself, and the half a reader is more likely to act on is the invariant: someone reading invariant 5 to find out whether provenance exists is told it is a seam awaiting a later iteration, and would go build what is already there.

## The guard

Guardable rather than merely fixable, because the claim names the API — a sentence calling `X` future can be checked against whether `X` is exported from `gemi/orm` or is a static on `Model`.

Verified against the record as it stood, naming both:

```
× a future `ActiveRecordModel` is not something the ORM already ships
× a future `User.save(row)` is not something the ORM already ships

AssertionError: the design record calls ActiveRecordModel future,
but the ORM exports it
```

The pattern matches only ``a future `identifier` ``, deliberately. "every future tier becomes a rewrite" sits two lines from one of these and names no API; matching loosely on the word would flag it and teach the next person to route around the guard. Two anchor tests come with it, so a moved or reworded file fails loudly rather than sweeping up nothing and passing.

## Also checked this round, and found correct

Three things I expected to be gaps and were not — recording them so they are not re-investigated:

- **the differential harness covers all fifteen operations.** `aggregate` (20 cases) and `groupBy` (16) are compared against Prisma, contrary to what a grep for `.aggregate(` suggests — the harness takes the operation as a *string*, so the cases live in `CASES`/`PER_PARENT`/… tables.
- **invariant 6 is guarded**, by `generated.test.ts`'s "the artifact has no import graph", which quotes the invariant and asserts the generated artifact has no runtime imports between models.
- **invariant 5's off-by-default claim is guarded**, by `save.test.ts:86`.

## Checks

- `bun --bun vitest run orm database` — 1386 passed

Touches `plans/orm/README.md:216`, two lines from #196's change to line 218, so whichever merges second needs a one-line rebase. Kept separate because they are different findings.

### DIFF
--- a/packages/gemi/orm/plan-record.test.ts
+++ b/packages/gemi/orm/plan-record.test.ts
@@ -1,0 +1,92 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+import { Model } from "./Model";
+import * as orm from "./index";
+
+/**
+ * The design record does not call shipped API "future".
+ *
+ * `plans/orm/README.md` is the ORM's design record — the document the invariants
+ * come from, cited by name from `strategy.ts`, `lateral.ts`, `nested-writes.ts`
+ * and `order-relation.ts`. It is a living document: it has a "Where it actually
+ * is now" section, and line 496 reads "Both are now in all four".
+ *
+ * Two of its invariants still described shipped API as forthcoming:
+ *
+ *   invariant 3   a future `ActiveRecordModel` base overrides `$shape`
+ *   invariant 5   enabling a future `User.save(row)` that diffs against …
+ *                 Iterations 1–7 only need the seam to exist; iteration 8
+ *                 fills it in.
+ *
+ * `ActiveRecordModel` is exported from `orm/index.ts:2`. `Model.save` is a
+ * static at `Model.ts:204`. Iteration 8 landed, and the same document's own
+ * status table says so — "#53 … opt-in provenance, `save(row)`, `wrap`".
+ *
+ * So the document contradicts itself, and the half a reader is more likely to
+ * act on is the invariant: someone reading it to find out whether provenance
+ * exists is told it is a seam awaiting a later iteration, and would go and build
+ * what is already there.
+ *
+ * This is guardable rather than merely fixable because the claim names the API.
+ * A sentence calling `X` future can be checked against whether `X` exists.
+ */
+const RECORD = join(import.meta.dirname, "../../../plans/orm/README.md");
+
+/**
+ * `a future \`ActiveRecordModel\``, `a future \`User.save(row)\`` — a backticked
+ * identifier the prose puts in the future.
+ *
+ * Only this phrasing, deliberately. "every future tier becomes a rewrite" two
+ * lines below one of these names no API and is not a claim about what exists;
+ * matching loosely on the word would flag it and teach the next person to work
+ * around the guard.
+ */
+const FUTURE = /a future `([A-Za-z_][\w.]*)(\([^`]*\))?`/g;
+
+const claimed = [...readFileSync(RECORD, "utf8").matchAll(FUTURE)].map(
+  (match) => ({ name: match[1], quote: match[0] }),
+);
+
+/**
+ * Is the ORM already shipping this?
+ *
+ * Two forms appear: a bare export (`ActiveRecordModel`) and a call on a model
+ * (`User.save(row)`), which names a static on the model base rather than a
+ * member of any particular model.
+ */
+const ships = (name: string) => {
+  if (name in orm) return true;
+  const [, member] = name.split(".");
+  return Boolean(member) && typeof (Model as never as Record<string, unknown>)[member] === "function";
+};
+
+describe("plans/orm/README.md does not call shipped API future", () => {
+  /**
+   * The two the fix removed are the only two the phrasing ever had, so an empty
+   * sweep is the expected state — but a *silently* empty one would also be what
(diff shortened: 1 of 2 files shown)


nstfkc:
## Review — the design record still calls shipped API future (closes #197)

Both claims verified against the code they name:

```
README.md:216  "a future `ActiveRecordModel`"   →  orm/index.ts:2   export { ActiveRecordModel }
README.md:283  "a future `User.save(row)`"      →  orm/Model.ts:204 static async save<T>(row: T)
```

And the guard reproduces both ways, naming each:

```
fixed record   2 passed
base record    × a future `ActiveRecordModel` is not something the ORM already ships
               × a future `User.save(row)` is not something the ORM already ships
```

**Checking the claim against the module rather than against a list is what makes this guardable**, and it is a better shape than the count guards next to it: there is no second copy of the truth to keep in step — the export either exists or it does not.

### I went looking for stragglers and did not find any

The last four prose guards in this series each shipped with a pattern narrower than the prose, and each left an instance behind — I have flagged that on #187, #189, #193 and #195. So I swept the corrected record for every future-tense construction, not just `` a future `X` ``:

```
:140  `gemi upgrade` — free now, breaking later          versioning policy, no API claim
:218  As a free function, every future tier…             the sentence you defend, names nothing
:269  future n+1-avoiding variant…                       hypothetical strategy
:335  Anything not yet implemented…                      general
:460  (provenance, future computed fields)               genuinely unshipped — no computed-field API
:530  Anything in "Out of scope" belongs to a later…     general
```

Six phrases, none naming shipped API. `:460` is the one I expected to be a straggler and it holds: there is no computed-field surface in `orm/` at all. So this is the first of the five where the narrow pattern leaves nothing behind — and the argument for the narrowness is sound rather than lucky, since `:218` sits two lines from a corrected claim and would be flagged by any looser rule.

**The three "checked and found correct" items are worth keeping**, and I verified the counterintuitive one: `differential.test.ts` contains **zero** occurrences of `.aggregate(` and twenty of `"aggregate"`, sixteen of `"groupBy"`. The harness really does take the operation as a string, so a grep for the method call reports absence where there is coverage. That is the sort of thing that gets "found" as a gap twice before someone writes it down.

### It shares a paragraph with #195, and they merge cleanly

Both edit `plans/orm/README.md` inside invariant 3 — this PR at line 216, #195 at line 218, within overlapping diff context. Checked rather than assumed:

```
merge-tree 311e9e5 894c176   clean
merged text                  "the `ActiveRecordModel` … with zero changes to the fifteen operations"
```

Both fixes survive. No rebase needed in either order.

Nothing outstanding from me.

PR #200, opened 2026-07-29
The framework summary lists everything gemi ships except the ORM

### LINKED ISSUE #199: The framework summary lists everything gemi ships except the ORM
`docs/README.md`, `docs/llms.txt` and `docs/llms-full.txt` open with the same one-paragraph description of what gemi ships. It enumerates eleven capabilities and none of them is the query layer:

> It ships **everything a typical web application needs** — class-based routing, a type-safe network layer, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.

The sentence is a claim plus an enumeration, and the enumeration is what a reader checks the claim against. A typical web application needs a database layer. gemi ships one — two documentation pages, its own Prisma generator block, a differential suite against Prisma — and someone reading that list to work out whether they still have to wire up Drizzle or the Prisma client themselves concludes that they do.

That is the opposite of the sentence's promise, and of the subsystem's entire pitch.

### Where it costs most

`llms.txt` exists to be the short answer a tool reads before deciding what else to fetch. Cron and i18n are named in it; the ORM is not. So *"does gemi have an ORM?"* answers itself wrongly from the summary alone — and the two ORM pages listed twenty lines further down do not undo that, because the summary is what gets quoted.

### Related history

`plans/orm/README.md` records the same concern one level down:

> neither page was reachable from `docs/README.md`, `docs/index.html`, `docs/llms.txt` or `docs/llms-full.txt` until the doc pass. Both are now in all four.

The pages got linked. The sentence describing what the framework ships did not move with them.

### Fix

Name it in the list, in all three copies, and guard it.

The guard locates the summary by the first item in its list rather than by line number, so a reworded opening still finds the sentence instead of silently finding nothing. It also asserts the three agree on the capability list — which is what stops the next edit fixing one copy and leaving the other two, which is how this arose.

`docs/index.html` carries no such summary and is out of scope.

nstfkc:
Closes #199.

`docs/README.md`, `docs/llms.txt` and `docs/llms-full.txt` open with the same description of what gemi ships, enumerating eleven capabilities without a query layer among them:

> It ships **everything a typical web application needs** — class-based routing, a type-safe network layer, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.

The sentence is a claim plus an enumeration, and the enumeration is what a reader checks it against. A typical web application needs a database layer; gemi ships one, with two documentation pages, its own Prisma generator block and a differential suite against Prisma. Someone reading that list to work out whether they still have to wire up Drizzle or the Prisma client themselves concludes that they do — the opposite of the sentence's promise, and of the subsystem's whole pitch.

It costs most in `llms.txt`, whose purpose is to be the short answer a tool reads before deciding what else to fetch. Cron and i18n are named there and the ORM is not, so *"does gemi have an ORM?"* answers itself wrongly from the summary alone; the two ORM pages listed twenty lines below do not undo that, because the summary is the part that gets quoted.

`plans/orm/README.md` records the same concern one level down — neither page "was reachable from `docs/README.md`, `docs/index.html`, `docs/llms.txt` or `docs/llms-full.txt` until the doc pass". The pages got linked; the sentence saying what the framework ships did not move with them.

## The guard

It locates the summary by the first item in its list rather than by line number or heading, so a reworded opening still finds the sentence instead of silently finding nothing — and it asserts all three agree on the capability list, which is what stops the next edit fixing one copy and leaving the other two, which is how this arose.

**Verified against all three as they stood:**

```
× README.md names the ORM in it
× llms.txt names the ORM in it
× llms-full.txt names the ORM in it

AssertionError: llms.txt enumerates what gemi ships without naming the query layer
```

One correction worth recording: my first version of the agreement check compared the summaries to the end of the sentence and reported a disagreement that was not one — `llms.txt` closes "as one integrated stack" where the other two close "so you spend time on product code", which is deliberate. It now compares the list from its first item to its last.

`docs/index.html` carries no such summary and is not in scope.

## Also checked this round, and found correct

- **all 21 exported error classes are constructed in the runtime** — no documented-but-unreachable error. `UnsupportedQueryError` has 19 construction sites, `InvalidArgumentError` 12; the rarest have one each.

## Checks

- `bun --bun vitest run orm database` — 1391 passed

Touches only `docs/README.md`, `docs/llms.txt` and the summary paragraph of `docs/llms-full.txt` — no overlap with #192's edit further down that file, and none with #194, #196 or #198.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # gemi Documentation
 
-**gemi** is a batteries-included, full-stack TypeScript web framework built on **Bun**, **Vite**, and **React 19** with server-side rendering. It ships everything a typical web application needs — class-based routing, a type-safe network layer, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.
+**gemi** is a batteries-included, full-stack TypeScript web framework built on **Bun**, **Vite**, and **React 19** with server-side rendering. It ships everything a typical web application needs — class-based routing, a type-safe network layer, a Prisma-typed ORM, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.
 
 gemi is Laravel-inspired: an application is a **Kernel** owning a **container**, **service providers** register bindings into that container, runtime settings live in **`app/config/*.ts`**, routes are declared in **router classes**, request logic lives in **controllers**, and framework services are reached through **facades**. It is **not** Next.js — there is no file-based routing; every URL is mapped explicitly in a router.
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3,7 +3,7 @@
 
 ## gemi Documentation
 
-**gemi** is a batteries-included, full-stack TypeScript web framework built on **Bun**, **Vite**, and **React 19** with server-side rendering. It ships everything a typical web application needs — class-based routing, a type-safe network layer, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.
+**gemi** is a batteries-included, full-stack TypeScript web framework built on **Bun**, **Vite**, and **React 19** with server-side rendering. It ships everything a typical web application needs — class-based routing, a type-safe network layer, a Prisma-typed ORM, authentication and authorization, middleware, form validation, transactional email, background jobs, cron, object storage, real-time broadcasting, and i18n — so you spend time on product code instead of wiring libraries together.
 
 gemi is Laravel-inspired: an application is a **Kernel** owning a **container**, **service providers** register bindings into that container, runtime settings live in **`app/config/*.ts`**, routes are declared in **router classes**, request logic lives in **controllers**, and framework services are reached through **facades**. It is **not** Next.js — there is no file-based routing; every URL is mapped explicitly in a router.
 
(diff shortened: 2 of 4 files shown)


nstfkc:
## Review — the framework summary lists everything except the ORM (closes #199)

Verified at the base: all three copies enumerate eleven capabilities with no query layer, and the guard names each one.

```
base    × README.md names the ORM in it
        × llms.txt names the ORM in it
        × llms-full.txt names the ORM in it
fixed   7 passed   ("a type-safe network layer, a Prisma-typed ORM, authentication…")
```

**The `llms.txt` argument is the one that decides this.** That file exists to be the short answer a tool reads before deciding what else to fetch, and it names cron and i18n while omitting the query layer — so *"does gemi have an ORM?"* resolves wrongly from the summary without the tool ever reaching the two pages listed below it. Same shape as #179, and a stronger instance: there the embedded copy was stale, here the omission is in the sentence whose whole job is to be quoted.

**The agreement check is the part I would keep**, and it does what it says. Reverting one copy:

```
× llms.txt names the ORM in it
× all three carry the same capability list
```

Two failures, and the second is the one that matters — it is the first guard in this series aimed squarely at *one copy fixed, the others left*, which is the failure mode behind #182, #185, #187 and #195. Worth noting that it now exists, because the same shape keeps recurring and this is the pattern for handling it.

**Anchoring on the first list item rather than a heading survives a rewording loudly**, which I checked because "finds the sentence instead of silently finding nothing" is exactly the property that is usually asserted and not tested:

```
"class-based routing" → "controller-first routing" in all three
  × README.md carries the summary
  × llms.txt carries the summary
  × llms-full.txt carries the summary
```

It fails on the anchor before it fails on the content, which is the right order — a reader gets "the summary moved", not "the ORM is missing" from a sentence that no longer exists.

Recording the false start about the closing clause — `llms.txt` ends "as one integrated stack" where the others end "so you spend time on product code" — is worth keeping too. Comparing to the end of the sentence would have made a deliberate difference look like drift, and that is the failure mode that makes a guard get deleted rather than fixed.

### Composition

It touches `docs/llms-full.txt`, which #191 also rewrote, and it branched before the current head. Checked rather than assumed:

```
0238400 vs feat/orm            clean
0238400 vs #193, #195, #197    clean
full orm+database suite        1391 passed   (#191's verbatim guard still green)
```

Nothing outstanding.

PR #202, opened 2026-07-29
The template's 15 type assertions were never run

### LINKED ISSUE #201: The template's 15 type assertions are never run
`templates/saas-starter/app/models/wrap.test-d.ts` and `aggregate.test-d.ts` hold 15 type-level assertions. Nothing runs them — not CI, not any script, not a local `bun test`.

Vitest's default `include` is `**/*.test.ts`, which `.test-d.ts` does not match, and type tests need `--typecheck` to be collected at all. No invocation in the repository passes it. Measured:

```
$ bun --bun vitest list app/models              # 0 tests from .test-d.ts
$ bun --bun vitest list app/models --typecheck  # 15
```

CI's two template steps are `vitest run app/models` and `vitest run app/models --no-file-parallelism -t postgres`. Neither collects them.

### Why it matters

They are the only check on the compile-time half of the ORM's pitch, and the runtime suite cannot express what they assert:

```ts
// @ts-expect-error a partial row is missing columns a method could read
UserModel.wrap(partial);
```

That is the claim `docs/orm-rows-and-entities.md` makes in as many words:

> `wrap` requires a **complete** row. A partial one is a compile error

and it is the reason the page gives for why narrowing and behaviour can coexist at all — *"the type system, not a runtime check, is what guarantees it"*. The guarantee has been documented, and unverified, since iteration 8.

They are load-bearing rather than decorative. `@ts-expect-error` fails when the error it expects stops happening, so making `wrap` accept what it should refuse produces `TypeCheckError: Unused '@ts-expect-error' directive.` — checked rather than assumed.

### Why it was not simply switched on

A plain typecheck run of the template reports **276 type errors in its own app code**, unrelated to the ORM, and exits 1. That is the same shape as #163, where `packages/gemi`'s full suite is red for reasons outside this work.

`--typecheck.ignoreSourceErrors` scopes the run to errors inside the `.test-d.ts` files. With it, the 15 assertions pass and the run exits 0.

### Fix

Add a `test:types` script to the template and a CI step that runs it, and guard the wiring — the failure was that nothing ran them, and nothing would notice if the step were removed again.

The 276 template type errors are left alone: out of ORM scope, same as #163.

nstfkc:
Closes #201.

`wrap.test-d.ts` and `aggregate.test-d.ts` hold 15 type-level assertions and ran nowhere — not in CI, not from any script, not locally.

Vitest's default `include` is `**/*.test.ts`, which `.test-d.ts` does not match, and type tests need `--typecheck` to be collected at all. Nothing in the repository passed it. Measured rather than inferred:

```
vitest list app/models              →  0 tests from .test-d.ts
vitest list app/models --typecheck  → 15
```

## Why it matters

They are the only check on the compile-time half of the ORM's pitch, and the runtime suite cannot express what they assert:

```ts
// @ts-expect-error a partial row is missing columns a method could read
UserModel.wrap(partial);
```

That is the claim `docs/orm-rows-and-entities.md` makes in as many words — *"`wrap` requires a **complete** row. A partial one is a compile error"* — and the reason the page gives for why narrowing and behaviour can coexist: *"the type system, not a runtime check, is what guarantees it"*. Documented, and unverified, since iteration 8.

Load-bearing rather than decorative, checked rather than assumed: `@ts-expect-error` fails when the error it expects stops happening, so making `wrap` accept what it should refuse produces

```
```

## Why it could not simply be switched on

A plain typecheck run of the template reports **276 type errors in its own app code**, unrelated to the ORM, and exits 1 — the same shape as #163, where `packages/gemi`'s full suite is red for reasons outside this work. That is presumably why the step never existed.

`--typecheck.ignoreSourceErrors` scopes the run to errors inside the `.test-d.ts` files. With it the 15 pass and the run exits 0. The 276 are left alone: out of ORM scope, same call as #163.

## The guard

On the **wiring**, not the assertions — the failure was that nothing ran them, and nothing would have noticed if the step were removed again. It pins the script, the `ignoreSourceErrors` flag *and its reason*, the CI step, and that both files still contain assertions, so deleting them fails rather than passing over an empty set.

Verified against the pre-fix state:

```
× the template has a script that runs vitest in typecheck mode
× it reports only the assertions, not the template's unrelated errors
× CI runs it
```

## Checks

- `bun --bun vitest run orm database` — 1407 passed
- `bun run test:types` (the new step, as CI runs it) — 2 files, 15 passed, exit 0
- template runtime suite unchanged at 17 files / 616 tests, confirmed against `feat/orm`

One honest note: an early run of the template suite reported a collection failure. It did not reproduce across three consecutive runs and the counts match `feat/orm` exactly, so it was a flake rather than this change — recording it in case it resurfaces.

Touches `.github/workflows/ci.yml`, the template's `package.json`, and one new test file. No overlap with #194, #196, #198 or #200.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,22 @@ jobs:
         env:
           TZ: UTC
 
+      # `.test-d.ts` files do not match vitest's default `include`, so the step
+      # above collects none of them: 15 type assertions in `wrap.test-d.ts` and
+      # `aggregate.test-d.ts` ran nowhere, locally or here. They are the only
+      # check on the compile-time half of the ORM's pitch — that `wrap` refuses
+      # a partial row, and that a narrowed `select` narrows the result type.
+      #
+      # `ignoreSourceErrors` because the template's own app code has 276 type
+      # errors unrelated to the ORM. Without it this step reports them and
+      # fails, which is why it could not simply be switched on; with it, the
+      # scope is exactly the assertions in the `.test-d.ts` files.
+      - name: Template type tests
+        run: bun run test:types
+        working-directory: templates/saas-starter
+        env:
+          TZ: UTC
+
       # Two lints, because the package and the ORM are in different states.
       #
       # `bun run lint` is `oxlint` over the whole package: 88 warnings, 0
--- a/packages/gemi/orm/type-tests-run.test.ts
+++ b/packages/gemi/orm/type-tests-run.test.ts
@@ -1,0 +1,74 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * The template's type assertions are actually executed by something.
+ *
+ * `wrap.test-d.ts` and `aggregate.test-d.ts` hold 15 assertions — and ran
+ * nowhere. Vitest's default `include` is `**‍/*.test.ts`, which `.test-d.ts`
+ * does not match, so `vitest run app/models` collected none of them; there was
+ * no `--typecheck` invocation in any script or in CI. `vitest list app/models`
+ * reported 0 of them, and `--typecheck` reported 15.
+ *
+ * They are the only check on the compile-time half of the ORM's pitch. The
+ * runtime suite cannot express what they assert:
+ *
+ *     // @ts-expect-error a partial row is missing columns a method could read
+ *     UserModel.wrap(partial);
+ *
+ * which is the claim `docs/orm-rows-and-entities.md` makes in as many words —
+ * "`wrap` requires a **complete** row. A partial one is a compile error" — and
+ * the reason the page gives for why narrowing and behaviour can coexist at all.
+ *
+ * `@ts-expect-error` fails when the error it expects stops happening, so these
+ * are load-bearing rather than decorative: making `wrap` accept what it should
+ * refuse turns the directive into `Unused '@ts-expect-error' directive`. That
+ * was checked rather than assumed.
+ *
+ * This guards the wiring, not the assertions — the failure was that nothing ran
+ * them, and nothing would have noticed if the step were deleted again.
+ */
+const ROOT = join(import.meta.dirname, "../../..");
+
+const workflow = readFileSync(join(ROOT, ".github/workflows/ci.yml"), "utf8");
+const templatePackage = JSON.parse(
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — the template's 15 type assertions were never run (closes #201)

The core claim reproduces exactly:

```
vitest list app/models              0 tests from .test-d.ts
vitest list app/models --typecheck  15
bun run test:types                  2 files, 15 passed, exit 0
```

**And they are load-bearing, which is the part that decides whether this is worth wiring up.** Loosening the generated `wrap` signature to accept `object` — 13 models — turns three of them red with exactly the mechanism described:

```
Tests  3 failed | 12 passed (15)
```

So the file really is checking that `wrap` refuses a partial row, which is the claim `orm-rows-and-entities.md` makes in as many words and the only check on it.

**Guarding the wiring rather than the assertions is the right target** — the failure was that nothing ran them — and it fires at every point where it could be undone. I tried three:

```
--typecheck.ignoreSourceErrors removed   × it reports only the assertions…
the CI step removed                      × CI runs it
wrap.test-d.ts emptied                   × wrap.test-d.ts is still there
```

Pinning the flag *and its reason* is what stops the next person removing it to "clean up" and rediscovering the 276.

### The 276 are mostly not the template's app code

The count is exact, and none of them are in the `.test-d.ts` files, so `ignoreSourceErrors` scopes to precisely the assertions. But the attribution is off:

```
../../packages/gemi   229    (framework, reached through the template's tsconfig)
app/                   47    (the template's own code)
```

And `packages/gemi` typechecks **clean** under its own config — the 229 appear only when the same source is compiled with the template's settings, and they are overwhelmingly React type-identity errors (`ReactElement` vs `ReactNode`, `MemoExoticComponent`), which is what a `jsx`/`lib` difference produces.

That strengthens the decision rather than weakening it — five sixths of what the flag excludes is not even the template's to fix, and #163's precedent applies more cleanly than stated. But "the template's own app code has 276 type errors" sends the next reader to `app/`, where they will find 47 and conclude the number moved. Worth saying "276 through the template's tsconfig, 229 of them in framework source that is clean under its own", because the interesting fact is that two configs disagree about the same files — and that is a thing someone might want to chase separately.

Everything else checks out: template runtime suite unchanged, `tsc -p packages/gemi` clean, `oxlint orm --deny-warnings` clean.

PR #204, opened 2026-07-29
CI skipped eight passing test files to avoid six failing ones

### LINKED ISSUE #203: CI skips eight passing test files to avoid six failing ones
The unit step runs `vitest run orm database` — an inclusive path filter. The reason recorded in the workflow for not running the whole package is #163:

> Not the whole package: `app/`'s router tests fail on `feat/orm` today and are unrelated to this. Widening to everything would put CI red for a reason nobody here can act on, which is how a red board stops being read.

That reason is sound, and it covers **six** files. The filter excludes **fourteen**. The other eight pass and have never run in CI:

```
container/Container.test.ts                         passes
foundation/Application.test.ts                      passes
http/FileRoute.test.ts                              passes
support/withDefaults.test.ts                        passes
app/App.test.ts                                     passes
services/router/assertNoReservedRoutePaths.test.ts  passes
services/router/createComponentTree.test.ts         passes
services/router/createFlatApiRoutes.test.ts         passes
```

Measured against the full suite: **55 of the package's 61 test files are green, and CI runs 47.**

### The direction is the real problem

An inclusive path filter fails silently in one direction. A test file added outside those two words is skipped, and nothing reports it — which is exactly how #201's fifteen type assertions went unnoticed for as long as they did.

A named exclusion list is the opposite shape: anything added is covered by default, and narrowing coverage means deleting a line rather than merely not writing one.

### Fix

Run the package with the six known-red files excluded by name:

```
app/createComponentTree.test.ts
app/createFlatViewRoutes.test.ts
app/createRouteManifest.test.ts
services/router/createFlatViewRoutes.test.ts
services/router/createRouteManifest.test.ts
utils/parseTranslation.test.tsx      (fails at collection)
```

Excluded in the **workflow**, not in a vitest config — so a local `vitest run` still shows them. CI stays green without the failures becoming invisible to whoever could fix them.

That introduces one new way to be wrong: an exclusion that outlives the file it names matches nothing and reads as though it still excludes something, which is invisible in a passing run. So each is guarded to still exist.

These six remain #163's; this issue does not propose fixing them.

nstfkc:
Closes #203.

The unit step ran `vitest run orm database` — an inclusive path filter. The recorded reason for not running the whole package was that `app/`'s router tests fail on `feat/orm` and are unrelated to this work (#163), and a board red for a reason nobody here can act on stops being read.

That reason is sound and it covers **six** files. The filter excluded **fourteen** — the other eight pass and had never run:

```
container/Container.test.ts                         passes
foundation/Application.test.ts                      passes
http/FileRoute.test.ts                              passes
support/withDefaults.test.ts                        passes
app/App.test.ts                                     passes
services/router/assertNoReservedRoutePaths.test.ts  passes
services/router/createComponentTree.test.ts         passes
services/router/createFlatApiRoutes.test.ts         passes
```

**55 of the package's 61 test files are green; CI was running 47.**

## The direction matters more than the count

An inclusive filter fails silently in one direction: a test file added outside those two words is skipped and nothing reports it. That is exactly how #201's fifteen type assertions went unnoticed. Naming what is broken instead means anything added is covered by default, and narrowing coverage takes deleting a line rather than merely not writing one.

Measured rather than assumed — full suite is 6 failed / 55 passed, and the new command is **56 passed / 1474 tests** with the new guard included.

The six are excluded in the workflow rather than in a vitest config, so a local `vitest run` still shows them. CI stays green without the failures becoming invisible to the person who could fix them. They remain #163's; this does not propose fixing them.

## The guard, and two things it caught

Excluding by name introduces a new way to be wrong: an exclusion that outlives the file it names matches nothing, reads as though it still excludes something, and is invisible in a passing run. Each is checked to still exist — verified by renaming one:

```
AssertionError: CI excludes utils/renamed.test.tsx, which is not there
— the exclusion is stale
```

It also caught a mistake of mine while I was writing it. The first version read the whole workflow and asserted it no longer contained `vitest run orm database` — and failed, because the comment explaining the change *quotes the command it replaced*. It now strips comment lines first: a guard that cannot tell a command from a sentence about a command would be satisfied by deleting the explanation.

## Checks

- the exact CI command — 56 files, 1474 passed, 1 skipped, exit 0
- `bun --bun vitest run orm database` — 48 files, 1419 passed
- CI has three vitest invocations total; the postgres job runs the differential suite, not package units, so there is no second copy of this filter

Touches `.github/workflows/ci.yml`, which #202 also edits — #202 adds a step after *Template tests*, this rewrites *Unit tests* about ten lines above, so they should not collide, but whichever merges second is worth a glance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,17 +87,38 @@ jobs:
         run: bun run typecheck
         working-directory: packages/gemi
 
-      # `orm database`, not `orm`. The package has 47 test files and only 32
-      # have "orm" in their path — `database/DatabaseManager.test.ts` is the
-      # ORM's own SQLite foreign-key pragma test and would have been skipped by
-      # the narrower filter.
+      # Everything except a named red set, rather than a path filter.
       #
-      # Not the whole package: `app/`'s router tests fail on `feat/orm` today
-      # and are unrelated to this. Widening to everything would put CI red for a
-      # reason nobody here can act on, which is how a red board stops being
-      # read. Reported separately.
+      # This step used to run `vitest run orm database`, chosen so
+      # `database/DatabaseManager.test.ts` — the ORM's own SQLite foreign-key
+      # pragma test — was not skipped by the narrower `orm`. The reason given
+      # for not running the whole package was that `app/`'s router tests fail on
+      # `feat/orm` and are unrelated (#163), and a board that is red for a
+      # reason nobody here can act on stops being read.
+      #
+      # That reason covers six files. The filter excluded fourteen: `container`,
+      # `foundation`, `http`, `support`, `app/App` and three of the router
+      # tests all pass and were never run — 55 of the package's 61 files are
+      # green, and CI was running 47 of them.
+      #
+      # An inclusive filter also fails silently in one direction: a new test
+      # file outside those two words is skipped and nothing says so. That is how
+      # #201's type assertions went unnoticed. Naming what is broken instead
+      # means anything added is covered by default, and the exclusions have to
+      # be deleted for CI to stop being honest rather than merely left alone.
+      #
+      # These six are #163. They are excluded here rather than in a vitest
+      # config so a local `vitest run` still shows them: CI stays green without
+      # the failures becoming invisible to the person who could fix them.
       - name: Unit tests
-        run: bun --bun vitest run orm database
+        run: |
+          bun --bun vitest run \
+            --exclude '**/app/createComponentTree.test.ts' \
+            --exclude '**/app/createFlatViewRoutes.test.ts' \
+            --exclude '**/app/createRouteManifest.test.ts' \
+            --exclude '**/services/router/createFlatViewRoutes.test.ts' \
+            --exclude '**/services/router/createRouteManifest.test.ts' \
+            --exclude '**/utils/parseTranslation.test.tsx'
         working-directory: packages/gemi
 
       - name: Template tests
--- a/packages/gemi/orm/ci-coverage.test.ts
+++ b/packages/gemi/orm/ci-coverage.test.ts
@@ -1,0 +1,84 @@
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * CI runs the whole package except a named red set, and the names are real.
+ *
+ * The unit-test step used to be `vitest run orm database` — an *inclusive* path
+ * filter. The stated reason for not running everything was that `app/`'s router
+ * tests fail on `feat/orm` and are unrelated to this work (#163).
+ *
+ * That reason covers six files. The filter excluded fourteen: `container`,
+ * `foundation`, `http`, `support`, `app/App` and three of the router tests all
+ * pass and were never run. 55 of the package's 61 test files are green; CI was
+ * running 47.
+ *
+ * The direction matters more than the count. An inclusive filter fails silently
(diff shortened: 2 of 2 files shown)


nstfkc:
## Review — CI skipped eight passing files to avoid six failing ones (closes #203)

Every number reproduces:

```
full suite              6 failed | 56 passed  (62 files)
the new CI command      56 passed, 1474 tests, 1 skipped, exit 0
bun --bun vitest run orm database   48 files, 1419 tests
```

Eight more files and fifty-five more tests, for a command that is longer and says what it is doing.

**The argument about direction is the right one and is the general form of #201.** An inclusive filter fails silently when something is added outside it; a named exclusion fails loudly when something is added inside it. That is the difference between "nobody wrote the line" and "somebody deleted the line", and it is worth more than the eight files.

Excluding in the workflow rather than in a vitest config is the detail I would keep — a local `vitest run` still shows the six, so they stay visible to the person who could fix them while CI stays green.

**The comment-stripping correction is a good catch on yourself.** A guard that cannot tell a command from a sentence quoting the command would have been satisfied by deleting the explanation, which is the worst possible way for that check to pass. I have made the same class of mistake twice on this branch.

### The exclusions are checked for outliving their *file*, not their *failure*

Renaming one fires, exactly as described:

```
× utils/parseTranslation.test.tsx still exists
  CI excludes utils/parseTranslation.test.tsx, which is not there — the exclusion is stale
```

But replacing an excluded file's contents with a passing test does not:

```
utils/parseTranslation.test.tsx → a single trivially-passing test
  orm/ci-coverage.test.ts   9 passed
```

The file exists, so the staleness check is satisfied, and the file stays out of CI permanently. That is the same shape as the filter this PR removes — eight passing files skipped with nothing saying so — one step further in. And it is not hypothetical: these six are #163's to fix, and on the day they are fixed nothing will prompt removing their names.

**The assertable invariant is available and exactly true today.** I diffed the two sets:

```
failing files (full suite)   6
excluded in ci.yml           6
identical                    yes (modulo vitest's collection-failure formatting)
```

So *"the set CI excludes is the set that fails"* holds now, and would fire in both directions — a renamed file and a fixed one. Running the full suite from inside a test to check it is heavy; the cheap version is a CI step that runs **only** the excluded six with `continue-on-error: true`. A fixed file then turns that step green, in a step nobody gates on, and the name gets removed by whoever notices — which is the same "visible to the person who could fix it" reasoning the workflow already uses for keeping them out of the config.

nstfkc:
## Composition with #201 — clean, and both guards hold on the merged workflow

Both this PR and #201 rewrite `.github/workflows/ci.yml`: this one replaces the Unit tests step, #201 appends a Template type tests step. That is the arrangement that produced #159 and #166, so worth checking rather than assuming.

```
#201 vs feat/orm      clean
#203 vs feat/orm      clean
#201 vs #203          clean
```

The merged workflow carries both — the six `--exclude` lines on the unit step, and `bun run test:types` as its own step. And each PR's guard still passes against the *other's* change, which is the half a per-branch run cannot see:

```
ci-coverage.test.ts + type-tests-run.test.ts   14 passed
merged CI unit command, run for real           59 files, 1488 passed, 1 skipped
tsc --noEmit                                   clean
```

Worth noting what the two do together: 59 files rather than the 56 this PR reports alone, because #201's guard file and its two `.test-d.ts` companions come along. Neither number is wrong — they are measured on different trees — but 59 is what CI will print, and a reader comparing the PR body to the first green run should know which is which.

No action needed; recording it so the merge order does not have to be reasoned about.

PR #206, opened 2026-07-29
The lateral strategy's tests ran in neither CI job

### LINKED ISSUE #205: The lateral strategy's tests run in neither CI job
The Postgres job runs:

```
bun --bun vitest run app/models --no-file-parallelism -t postgres
```

`-t` matches the test's full name and is **case-sensitive**. A Postgres-only suite whose name does not contain the exact string `postgres` is selected by nothing.

Three files are in that position. Because they gate themselves with

```ts
const RUN = POSTGRES_URL ? describe : describe.skip;
```

they are skipped in the SQLite job too — for want of a URL there, and for want of a matching name in the other. **24 tests run in neither job:**

| file | tests | suite name |
| --- | --: | --- |
| `lateral.test.ts` | 13 | `"lateral vs batched, on Postgres"`, `"strategy selection"` |
| `lateral.coercion.test.ts` | 7 | `"every scalar through a relation, lateral vs batched"` |
| `redact-strategies.test.ts` | 4 | `"redact on a nested model, both strategies"` |

Eight of them are missed by a capital P alone:

```
-t postgres  -> 0 tests from lateral.test.ts
-t Postgres  -> 8
```

### Why this one matters

Two of the three files are `LATERAL` + `json_agg`. That strategy is Postgres-only, it is the **default** on Postgres, and `plans/orm/README.md` accepts it on the grounds that

> The **full** nested differential matrix runs against it — every relation case in `differential.test.ts` … That is iteration 9's acceptance criterion 2.

The matrix ran nowhere.

The suites that *do* run announce their own absence — `⚠ Postgres differential tests did NOT run. Set TEST_POSTGRES_URL`. A `describe.skip` says nothing, so these three were silent in precisely the way that warning mechanism exists to prevent.

### A broken test was hiding behind it

`redact-strategies.test.ts` builds a fixture table named `Ledger` — which is also a real model in the template's `schema.prisma`, with `LedgerEntry` holding a foreign key to it. Its setup `DROP TABLE IF EXISTS "Ledger"` fails on Postgres:

```
PostgresError: cannot drop table "Ledger" because other objects depend on it
```

It also registers its fixture under the name `Ledger`, replacing the template's own model for the remainder of a `--no-file-parallelism` run. The test could never have passed on Postgres — which is why nobody noticed that it could not.

### Fix

Rename the four suites to carry the `— postgres` suffix the other Postgres suites already use, and rename the colliding fixture (rather than `DROP ... CASCADE`, which would take the real table with it).

Verified against a real Postgres 16: the job goes from 622 to 646 selected tests, all passing on a clean database.

Guard: every suite gated behind `POSTGRES_URL` must contain the filter string, compared case-sensitively, with the filter read out of the workflow rather than assumed.

nstfkc:
Closes #205.

The Postgres job runs `vitest run app/models --no-file-parallelism -t postgres`. `-t` matches the test's full name and is **case-sensitive**, so a Postgres-only suite whose name does not contain the exact string `postgres` is selected by nothing.

Three files were in that position, and because they gate themselves with `const RUN = POSTGRES_URL ? describe : describe.skip` they were skipped in the SQLite job too — for want of a URL there, and for want of a matching name in the other. **24 tests ran in neither:**

```
lateral.test.ts            13   "lateral vs batched, on Postgres" / "strategy selection"
lateral.coercion.test.ts    7   "every scalar through a relation, lateral vs batched"
redact-strategies.test.ts   4   "redact on a nested model, both strategies"
```

Eight were missed by a capital P alone — `-t postgres` collects 0 from `lateral.test.ts`, `-t Postgres` collects 8.

Two of the three files are `LATERAL` + `json_agg`: Postgres-only, the **default** strategy there, and the thing `plans/orm/README.md` accepts on the grounds that "the full nested differential matrix runs against it — that is iteration 9's acceptance criterion 2". It ran nowhere.

The suites that do run announce their absence — `⚠ Postgres differential tests did NOT run`. A `describe.skip` says nothing, so these three were silent in exactly the way that mechanism exists to prevent.

## Verified against a real Postgres, not assumed

Postgres 16 in Docker, following the job's own steps (flip the provider, `prisma generate`, `db push`):

| | selected | result |
| --- | --: | --- |
| before | 622 | pass |
| after | 646 | pass |

Exactly +24, all green on a clean database. The SQLite job is unchanged at 616 passed / 32 skipped.

## One of the 24 was broken

`redact-strategies.test.ts` builds a fixture table called `Ledger` — also a real model in the template's schema, with `LedgerEntry` holding a foreign key to it:

```
PostgresError: cannot drop table "Ledger" because other objects depend on it
```

It also registered its fixture under the name `Ledger`, over the template's own model, for the rest of a `--no-file-parallelism` run. I renamed the fixture rather than using `DROP ... CASCADE`, which would have taken the real table and `LedgerEntry`'s foreign key with it. It could never have passed on Postgres, which is why nobody noticed that it could not.

A second failure appeared on the first re-run and did **not** reproduce on a clean database: it was state left behind by the aborted run above, not this change.

## The guard

Every suite gated behind `POSTGRES_URL` must contain the filter string. It reads the filter out of the workflow rather than assuming `postgres`, and compares **case-sensitively**, because that is how `-t` compares — "on Postgres" reads as though it is covered and is not.

Verified against the original names:

```
× lateral.test.ts: lateral vs batched, on Postgres
× lateral.test.ts: strategy selection
```

## Checks

- `bun --bun vitest run orm database` — 48 files, 1416 passed
- template SQLite suite — 17 passed / 3 skipped, 616 tests, unchanged
- template Postgres suite — 13 files, 646 passed, 0 failed

The scratch Postgres container was removed and the template's schema and Prisma client restored to SQLite.

Touches three template test files and adds one guard — no overlap with #198, #200, #202 or #204.

### DIFF
--- a/packages/gemi/orm/postgres-suite-selection.test.ts
+++ b/packages/gemi/orm/postgres-suite-selection.test.ts
@@ -1,0 +1,92 @@
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * Every Postgres-gated suite is named so CI's filter selects it.
+ *
+ * The Postgres job runs `vitest run app/models --no-file-parallelism -t
+ * postgres`. The filter is not tidiness — without it the job fails, because the
+ * harness refuses the dialect its Prisma client was not generated for. But `-t`
+ * matches on the test's full name, and it is **case-sensitive**, so a
+ * Postgres-only suite whose name does not contain the exact string `postgres`
+ * is selected by nothing.
+ *
+ * Three files were in that position, and because they gate themselves with
+ *
+ *     const RUN = POSTGRES_URL ? describe : describe.skip;
+ *
+ * they were skipped in the SQLite job too — for want of a URL there, and for
+ * want of a matching name here. 24 tests ran in neither:
+ *
+ *     lateral.test.ts             13   "lateral vs batched, on Postgres"
+ *                                      "strategy selection"
+ *     lateral.coercion.test.ts     7   "every scalar through a relation, …"
+ *     redact-strategies.test.ts    4   "redact on a nested model, both …"
+ *
+ * Eight of those were missed by a capital P alone. `-t postgres` collected 0
+ * tests from `lateral.test.ts`; `-t Postgres` collected 8.
+ *
+ * The cost was concentrated: two of the three files are the `LATERAL` +
+ * `json_agg` strategy, which is Postgres-only and the *default* there, and
+ * whose acceptance criterion in `plans/orm/README.md` is that the full nested
+ * differential matrix runs against it. It ran nowhere.
+ *
+ * The suites that do run announce their absence — "⚠ Postgres differential
+ * tests did NOT run. Set TEST_POSTGRES_URL". A `describe.skip` says nothing, so
+ * these three were silent in exactly the way that mechanism exists to prevent.
+ */
+const MODELS = join(import.meta.dirname, "../../../templates/saas-starter/app/models");
+const WORKFLOW = join(import.meta.dirname, "../../../.github/workflows/ci.yml");
+
+const workflow = readFileSync(WORKFLOW, "utf8");
+
+/**
+ * The exact string the Postgres job filters on, read from the workflow rather
+ * than assumed — if the job changes filter, this test should follow it or fail,
+ * not keep checking a word nobody greps for any more.
+ */
+const filter = (() => {
+  const match = workflow.match(/vitest run app\/models[^\n]*-t (\S+)/);
+  expect(match, "the Postgres job's -t filter was not found").not.toBeNull();
+  return match![1];
+})();
+
+/** Files that gate a suite behind `POSTGRES_URL`, and the suite names they use. */
+const gated = readdirSync(MODELS)
+  .filter((file) => file.endsWith(".test.ts"))
+  .map((file) => ({ file, source: readFileSync(join(MODELS, file), "utf8") }))
+  .filter(({ source }) => /POSTGRES_URL \? describe : describe\.skip/.test(source))
+  .map(({ file, source }) => ({
+    file,
+    suites: [...source.matchAll(/^RUN\(\s*"([^"]+)"/gm)].map((match) => match[1]),
+  }));
+
+describe("Postgres-only suites are selected by the Postgres job", () => {
+  test("the filter was read from the workflow", () => {
+    expect(filter).toBe("postgres");
+  });
+
+  test("the gated files were found", () => {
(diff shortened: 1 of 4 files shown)


nstfkc:
## Review — the lateral strategy's tests ran in neither CI job (closes #205)

I have no Postgres either, so I checked this by **selection** rather than execution — which turns out to be enough, because selection is the whole defect. With `TEST_POSTGRES_URL` set to a dummy value so the suites are collected rather than skipped, `vitest list` never connects:

```
base    -t postgres   0 from lateral.test.ts, lateral.coercion.test.ts, redact-strategies.test.ts
        -t Postgres   8 from lateral.test.ts
after   -t postgres   13 + 7 + 4 = 24
total selected        622 → 646
```

Exactly the +24 in your table, and the capital-P claim is precise: eight tests differ by one character, and the other two files contain no form of the word at all, so neither filter reached them.

**"Skipped in both jobs for different reasons" is the part worth keeping.** `describe.skip` in the SQLite job for want of a URL, and unselected in the Postgres job for want of a matching name — neither of which is a failure, so the suites were invisible twice over. The contrast with the harness's own `⚠ Postgres differential tests did NOT run` is the right one: that mechanism exists precisely so a dialect cannot go missing quietly, and these three sat outside it.

That two of the three are `LATERAL` + `json_agg` — the Postgres default, and the strategy `plans/orm/README.md` accepts *on the grounds that* the nested differential matrix runs against it — is what makes this worth more than the test count.

**Finding a genuine failure among the 24 is the strongest evidence for the change**, and the handling is right: a fixture table named `Ledger` colliding with the template's own model, renamed rather than `DROP ... CASCADE`, which would have taken `LedgerEntry`'s foreign key with it. *"It could never have passed on Postgres, which is why nobody noticed that it could not"* is the sentence that justifies the whole PR.

Distinguishing the second failure as leftover state from the aborted run, rather than folding it in, is the right call too.

**The guard fires in both directions**, including the one that matters going forward:

```
suite renamed back to "on Postgres"      × lateral.test.ts: lateral vs batched, on Postgres
a brand-new POSTGRES_URL-gated suite     × newpg.test.ts: a brand new dialect suite
```

with a message that names the mechanism — *"the Postgres job's -t filter is case-sensitive, so this suite is selected by neither job"*. Reading the filter out of the workflow rather than hard-coding `postgres` is what keeps it honest if the job's flag changes, and comparing case-sensitively **because that is how `-t` compares** is the detail that would otherwise be softened into a `toLowerCase()` and lose the actual bug.

This is the check I suggested on #193 when I noted nothing enforced the labelling convention the filter depends on. It is a better version than what I described — I proposed asserting suite labels end in a known dialect tag; deriving the required string from the workflow itself removes the second copy.

Nothing outstanding.

PR #208, opened 2026-07-29
Catch a fixture table that shadows a real model with a different shape

### LINKED ISSUE #207: Nothing catches a fixture table that shadows a real model with a different shape
Follow-on from #205, which found `redact-strategies.test.ts` building a fixture table called `Ledger`:

```sql
CREATE TABLE "Ledger" (
  "id" SERIAL NOT NULL PRIMARY KEY,
  "bookId" INTEGER NOT NULL REFERENCES "Book"("id"),
  "label" TEXT NOT NULL,
  "secret" TEXT
)
```

The real `Ledger` in `schema.prisma` is `tenantId`, `code`, `title` with a composite primary key and a `LedgerEntry` holding a foreign key to it. #206 renames the fixture. Nothing stops the next one.

### The rule is about shape, not names

Worth stating carefully, because the obvious rule is wrong. `relations.many-to-many.test.ts` also creates `Post` and `Tag`, which are also real models — and that is fine, because its DDL mirrors them exactly (`id`/`title`, `id`/`label`). Dropping and recreating them changes nothing. Banning the names would fail a suite that is doing the right thing.

The defect is a shadowing table with a *different* shape.

### Why it is worth catching

The Postgres suites share one database and run with `--no-file-parallelism`. A fixture that replaces a real table with a differently-shaped one replaces it for every suite that runs after it — and `policies.test.ts` and `writes.differential.test.ts` both write to `Post` and `Tag` through the differential harness.

`Ledger` never got that far only because its drop failed outright:

```
PostgresError: cannot drop table "Ledger" because other objects depend on it
```

Had it dropped the dependants too, it would have taken `LedgerEntry`'s foreign key with it.

### Fix

A static comparison: any table a template test creates whose name matches a model in `schema.prisma` must declare the same columns.

Static on purpose. The `Ledger` mismatch survived for as long as the suite holding it ran nowhere at all (#205), and a check that needs a live database to notice would have been exactly as quiet as the one that was missing.

The known-shadowing set is pinned as `Post`, `Tag` rather than asserted empty, so a third arrives as a visible change rather than silently.

nstfkc:
Closes #207. **Stacked on #206** — based on its branch and targeting it, so this diff is the one new file. If #206 merges first this retargets cleanly to `feat/orm`.

#205 found `redact-strategies.test.ts` building a `Ledger` of `id`, `bookId`, `label`, `secret`, where the real `Ledger` is `tenantId`, `code`, `title` with a composite primary key and a `LedgerEntry` pointing at it. #206 renames that fixture; this turns the one-off into a rule.

## The rule is about shape, not names

Worth stating carefully, because the obvious rule is wrong. `relations.many-to-many.test.ts` also creates `Post` and `Tag`, and those *are* real models — but its DDL mirrors them exactly, `id`/`title` and `id`/`label`, so dropping and recreating them changes nothing. Banning the names would fail a suite doing the right thing.

I checked that rather than assuming it; the shape match is why only `Ledger` ever failed.

## Why it is worth catching

The Postgres suites share one database and run with `--no-file-parallelism`. A fixture that replaces a real table with a differently-shaped one replaces it for every suite that runs afterwards — and `policies.test.ts` and `writes.differential.test.ts` both write to `Post` and `Tag` through the differential harness.

`Ledger` never got that far only because the drop failed outright. Had it dropped its dependants too, it would have taken `LedgerEntry`'s foreign key with it.

## Static on purpose

The `Ledger` mismatch survived for as long as the suite holding it ran nowhere at all (#205). A check that needs a live database to notice would have been exactly as quiet as the one that was missing.

Verified against the pre-#206 fixture, which it reports by name and by column:

```
× redact-strategies.test.ts: Ledger

AssertionError: redact-strategies.test.ts creates a table called "Ledger",
which is a real model in schema.prisma, with different columns
expected { missing: [ 'tenantId', 'code', 'title' ], … } to deeply equal { missing: [], extra: [] }
```

The known-shadowing set is pinned as `Post`, `Tag` rather than asserted empty, so a third one arrives as a visible change here instead of silently.

## Checks

- `bun --bun vitest run orm database` — 49 files, 1422 passed
- template SQLite suite — 17 passed / 3 skipped, 616 tests, unchanged

### DIFF
--- a/packages/gemi/orm/fixture-shape.test.ts
+++ b/packages/gemi/orm/fixture-shape.test.ts
@@ -1,0 +1,113 @@
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * A template test that builds a table named after a real model builds the same
+ * table.
+ *
+ * The Postgres suites share one database, run with `--no-file-parallelism`, and
+ * several of them create their own tables with raw DDL. Two things follow from
+ * that, and only one of them was true.
+ *
+ * `relations.many-to-many.test.ts` creates `Post`, `Tag` and `_PostToTag`.
+ * Those are real models in `schema.prisma`, and the DDL mirrors them exactly —
+ * `id`/`title` and `id`/`label`. Dropping and recreating them is a no-op in
+ * shape, which is why it works and why it is fine.
+ *
+ * `redact-strategies.test.ts` created a `Ledger` of `id`, `bookId`, `label`,
+ * `secret`. The real `Ledger` is `tenantId`, `code`, `title` with a composite
+ * primary key and a `LedgerEntry` pointing at it. That test could not run at
+ * all — `cannot drop table "Ledger" because other objects depend on it` (#205)
+ * — and had it dropped the dependants too, it would have replaced a table two
+ * other suites write to with one of a different shape, and taken
+ * `LedgerEntry`'s foreign key with it.
+ *
+ * The name collision is not the defect on its own; the shape mismatch is. So
+ * that is what this checks, rather than banning the names — banning them would
+ * fail the many-to-many suite, which is doing the right thing.
+ *
+ * It is a static comparison on purpose. The failure it exists to catch was
+ * invisible for as long as the suite that would have hit it never ran, and a
+ * check that needs a database to notice would have been just as quiet.
+ */
+const TEMPLATE = join(import.meta.dirname, "../../../templates/saas-starter");
+const MODELS = join(TEMPLATE, "app/models");
+
+/** Every model in the Prisma schema, to its scalar field names. */
+const schemaModels = (() => {
+  const source = readFileSync(join(TEMPLATE, "prisma/schema.prisma"), "utf8");
+  const models = new Map<string, Set<string>>();
+
+  for (const match of source.matchAll(/^model\s+(\w+)\s*\{([^}]*)\}/gm)) {
+    const fields = new Set<string>();
+    for (const line of match[2].split("\n")) {
+      const field = line.trim().match(/^(\w+)\s+\w/);
+      // Relation fields carry a model type and no column; `@@id`/`@@index`
+      // start with `@` and are skipped by the same rule.
+      if (field && !/\[\]/.test(line)) fields.add(field[1]);
+    }
+    models.set(match[1], fields);
+  }
+
+  return models;
+})();
+
+/** Every `CREATE TABLE "X" (...)` in the template's tests, to its columns. */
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — catch a fixture table that shadows a real model (closes #207)

**The shape-not-names rule is the right one, and I checked the discrimination that justifies it** rather than taking it:

```
real Post   id, title       fixture Post   id, title       identical
real Tag    id, label       fixture Tag    id, label       identical
real Ledger tenantId, code, title    old fixture   id, bookId, label, secret
```

So banning the names really would have failed `relations.many-to-many.test.ts`, which is doing the right thing, and only `Ledger` ever differed. That is the distinction the whole check turns on.

Sensitivity reproduces against the pre-#205 fixture, with the columns named:

```
× redact-strategies.test.ts: Ledger
  expected { missing: [ 'tenantId', 'code', 'title' ], … } to deeply equal { missing: [], extra: [] }
```

**Static on purpose is the right call and the reason is the sharp one** — the mismatch survived precisely as long as the suite holding it ran nowhere, so a check needing a live database would have been exactly as quiet as the absence it replaces. Pinning `Post, Tag` as a known set rather than asserting empty is the same instinct.

### The DDL pattern is case-sensitive, in a PR stacked on one about case-sensitivity

```js
/CREATE TABLE (?:IF NOT EXISTS )?"(\w+)"\s*\(([^`]*?)\)\s*`/gs
```

Flags `gs`, no `i`. A differently-shaped `Ledger` written in lower case is invisible:

```
CREATE TABLE "Ledger"                  2 failed   ✓ caught
CREATE TABLE IF NOT EXISTS "Ledger"    2 failed   ✓ caught
create table "Ledger"                  6 passed   ✗
```

Same shape as the defect #205 fixes one commit down — eight tests lost to a capital P — with the polarity reversed: that filter needed lower case and got upper, this pattern needs upper and would miss lower. Worth saying because the two land together and a reader will meet the lesson twice.

It is not purely hypothetical either. All 25 `CREATE TABLE`s in `app/models` are upper case today, so nothing is missed now — but `packages/gemi/database/DatabaseManager.test.ts` writes `create table parent (…)` in lower case, so the convention is not universal in the repo, and the next fixture author has a precedent for either.

One character, verified both ways:

```
/…/gis    corrected tree      6 passed   (no false positive)
          lowercase shadow    2 failed
```

**A caught-myself note.** I first read the flags as `gis` and nearly reported the opposite — I had grepped a copy my own patch had already modified rather than the original. Third time this session that reading my instrumentation instead of the source produced a wrong answer, and the reason I re-checked was that the result contradicted an experiment I had already run.

PR #210, opened 2026-07-29
The recorded fix for the Json divergence does not work

### LINKED ISSUE #209: The recorded fix for the Json divergence does not work
`docs/orm.md` documents one deliberate divergence from Prisma: a bare JSON number or boolean is refused on Postgres, because the encoder that accepted it stored `42` as the jsonb **string** `"42"` and only looked correct because the decoder re-parsed it.

Both `dialect/postgres.ts` and `writes.coercion.test.ts` then record how to lift the boundary:

> Fixing it properly means serialising and emitting an explicit `::jsonb` cast, which has to reach the insert, the update's set clause and any `where` on a Json column.

**That does not work.** Measured through Bun 1.3.14 against Postgres 16:

| binding | value | result |
| --- | --- | --- |
| `values ($1)` | `42` | fails — `integer` vs `jsonb` (today's error) |
| `values ($1::jsonb)` | `42` | fails — cannot cast integer to jsonb |
| `values ($1::jsonb)` | `"42"` | `jsonb_typeof` → **string** ← the old bug |
| `values (to_jsonb($1))` | `42` | `jsonb_typeof` → **number** ✅ |
| `values ($1::text::jsonb)` | `"42"` | `jsonb_typeof` → **number** ✅ |

Bun serialises a JS string bound to a `jsonb` parameter, so pre-serialising encodes it twice. The prescribed remedy reproduces the exact silent mis-store the refusal exists to prevent — three paragraphs after the same comment explains that mis-store in detail.

Anyone picking this up would follow the note, watch `jsonb_typeof` answer `string`, and land back where the previous encoder was.

### Fix

Correct the note in both places and record the two forms that do work.

The rest of the assessment stands: either working form needs the column's type at the **placeholder**, so it still has to reach the insert, the update's set clause and any `where` on a Json column — a compiler change with its own differential pass. This issue does not propose making that change; it makes sure the change starts from a correct premise.

### Pinning it

This is a claim about the *driver*, not about this ORM, so nothing in the compiler would notice it changing. A Postgres-suite test asserts the failing form still fails (`string`) alongside the two that give `number` and `boolean` — the failing case included deliberately, since a test that pinned only the working forms would not have caught the claim that was actually made.

nstfkc:
Closes #209.

`Json` on Postgres refuses a bare number or boolean, and both `dialect/postgres.ts` and `writes.coercion.test.ts` recorded how to lift it:

> Fixing it properly means serialising and emitting an explicit `::jsonb` cast

Measured through Bun 1.3.14 against Postgres 16, that stores the jsonb **string** `"42"` — the silent mis-store the refusal exists to prevent, and the exact bug the surrounding comment spends a paragraph explaining was fixed. Bun serialises a JS string bound to a `jsonb` parameter, so pre-serialising encodes it twice.

## The measurement, across every shape

| binding | object | array | string | number | boolean | null |
| --- | --- | --- | --- | --- | --- | --- |
| `values ($1)` — today | ✓ | ✓ | ✓ | ✗ `integer` vs `jsonb` | ✗ | ✓ |
| `values ($1::jsonb)` serialised — **the recorded fix** | ✗ stored as `string` | ✗ | ✗ | ✗ | ✗ | ✗ |
| `values (to_jsonb($1))` raw | ✗ polymorphic | ✗ | ✗ | ✓ | ✓ | ✗ |
| **`values ($1::text::jsonb)` serialised** | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |

So the note telling the next person how to fix this leads them into the bug it warns about. There is exactly **one** form that carries every shape: `JSON.stringify` and bind through `$1::text::jsonb`.

The rest of the assessment stands and is unchanged: that form needs the column's type at the **placeholder**, so it still has to reach the insert, the update's set clause and any `where` on a Json column — a compiler change with its own differential pass. Still not done here; this makes sure it starts from a correct premise.

## Two commits, deliberately

The second commit corrects the first. I originally recorded **two** working forms, having measured `to_jsonb($1)` against a bare number and boolean — the case in front of me — and generalised. It fails on object, array, string and null.

That is the same mistake the branch is about, one step further along: a remedy written down after being checked against the case in hand rather than the cases it would meet. Kept as its own commit rather than amended, because the sequence is the point.

## How this was found

I set out to *implement* the fix, not to check it. A probe meant to confirm a shortcut failed, which sent me to test the recorded remedy itself — and later to test my own.

## Pinning it

This is a claim about the driver, not about this ORM, so nothing in the compiler would notice it changing. The test pins all three forms across all six shapes — the two that fail included, and value as well as `jsonb_typeof`, since the mis-store being replaced was a right-looking value under the wrong type.

It is named `— postgres` so the Postgres job selects it, which #206's guard now enforces.

## Checks

- Postgres 16, the job's own setup — 624 selected, all passing; the new test runs in the Postgres suite (12ms) and early-returns in the SQLite one
- template SQLite suite — 17 passed / 3 skipped, 617 tests
- `bun --bun vitest run orm database` — 50 files, 1424 passed

Scratch container removed; the template's schema and Prisma client restored to SQLite. Independent of #204, #206 and #208.

### DIFF
--- a/packages/gemi/orm/dialect/postgres.ts
+++ b/packages/gemi/orm/dialect/postgres.ts
@@ -260,9 +260,23 @@ export class PostgresDialect implements SqlDialect {
     // time, and anything reading that column *other than this ORM* saw a
     // string. A loud failure replaces a silent mis-store.
     //
-    // Fixing it properly means serialising and emitting an explicit `::jsonb`
-    // cast, which has to reach the insert, the update's set clause and any
-    // `where` on a Json column. Not done here.
+    // **Fixing it is not "serialise and cast", which is what this comment used
+    // to say.** Measured through Bun 1.3.14 against Postgres 16, that stores
+    // the jsonb *string* `"42"` — the exact silent mis-store the refusal above
+    // exists to prevent, because Bun serialises a JS string bound to a `jsonb`
+    // parameter, so pre-serialising encodes it twice:
+    //
+    //   values ($1)                 42        integer vs jsonb — the error above
+    //   values ($1::jsonb)          42        cannot cast integer to jsonb
+    //   values ($1::jsonb)          "42"      jsonb_typeof -> string   <- the old bug
+    //   values (to_jsonb($1))       42        jsonb_typeof -> number   <- works
+    //   values ($1::text::jsonb)    "42"      jsonb_typeof -> number   <- works
+    //
+    // Either working form needs the column's type at the *placeholder*, so it
+    // still has to reach the insert, the update's set clause and any `where` on
+    // a Json column. Not done here — but whoever does it should start from one
+    // of the last two rows, not the second. `writes.coercion.test.ts` pins them
+    // so the table above stays a measurement rather than a memory.
     return value;
   }
 
--- a/packages/gemi/orm/dialect/postgres.ts
+++ b/packages/gemi/orm/dialect/postgres.ts
@@ -269,14 +269,25 @@ export class PostgresDialect implements SqlDialect {
     //   values ($1)                 42        integer vs jsonb — the error above
     //   values ($1::jsonb)          42        cannot cast integer to jsonb
     //   values ($1::jsonb)          "42"      jsonb_typeof -> string   <- the old bug
-    //   values (to_jsonb($1))       42        jsonb_typeof -> number   <- works
-    //   values ($1::text::jsonb)    "42"      jsonb_typeof -> number   <- works
+    //   values (to_jsonb($1))       42        jsonb_typeof -> number
+    //   values ($1::text::jsonb)    "42"      jsonb_typeof -> number
     //
-    // Either working form needs the column's type at the *placeholder*, so it
-    // still has to reach the insert, the update's set clause and any `where` on
-    // a Json column. Not done here — but whoever does it should start from one
-    // of the last two rows, not the second. `writes.coercion.test.ts` pins them
-    // so the table above stays a measurement rather than a memory.
+    // **`to_jsonb($1)` is not the answer either**, though it looks like one from
+    // that row alone. It works for exactly the shapes Bun binds as a concrete
+    // type — a number and a boolean — and fails for the four that matter more,
+    // because an unannotated parameter leaves the polymorphic argument untyped:
+    //
+    //   to_jsonb($1)   {a:1} / [1,2] / "42" / null
+    //                  could not determine polymorphic type
+    //
+    // So there is one form, not two: `JSON.stringify` the value and bind it
+    // through `$1::text::jsonb`, which round-trips all six shapes with the right
+    // `jsonb_typeof` — object, array, string, number, boolean and null.
+    //
+    // It needs the column's type at the *placeholder*, so it still has to reach
+    // the insert, the update's set clause and any `where` on a Json column. Not
+    // done here. `writes.coercion.test.ts` pins every row above, the failing
+    // forms included, so this stays a measurement rather than a memory.
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


nstfkc:
## Review — the recorded fix for the Json divergence does not work (closes #209)

No Postgres here, so the five-row table is yours. What I could check holds at every point I could reach.

**The base really did say it.** `dialect/postgres.ts:263` on `0c01841`:

> Fixing it properly means serialising and emitting an explicit `::jsonb` cast

— three lines under a paragraph explaining that the old encoder's pre-serialisation stored the jsonb *string* `"42"` and that this was the bug. So the remedy note pointed at the failure the surrounding comment documents, which is the sharpest possible form of a stale instruction: it is contradicted by the text immediately above it.

**Pinning the failing form alongside the working ones is the decision that makes this test worth having**, and the reason given is exactly right — a test that asserted only `to_jsonb` and `::text::jsonb` would confirm two facts nobody was wrong about, and would not have caught the claim that was actually made. `expect(await typeOf("$1::jsonb", JSON.stringify(42))).toBe("string")` is the assertion that encodes the correction.

**It will actually run**, which is the half a comment cannot promise. Measured by selection with a dummy URL, so no database is needed:

```
vitest list app/models -t postgres   →  selects "the two bindings a jsonb fix could use — postgres"
```

So the naming follows #205's convention and the Postgres job picks it up. Worth noting the guard that enforces this is on #205's branch, not here — the claim is true today by construction rather than by check, and becomes checked when that merges.

**And the fixture does not shadow anything.** `JsonProbe` is not a model in `schema.prisma`, so it is outside the class of collision #207 exists to catch — which matters because this test creates and drops a table in the shared Postgres database that #205 showed runs everything serially.

Local checks reproduce: `orm database` 50 files / 1424 passed, `tsc --noEmit` clean.

**The framing is the part I would keep.** *"I set out to implement the fix, not to check it"* — the probe was meant to confirm a shortcut, failed, and that is what sent you to test the recorded remedy. A note nobody acts on is never wrong; this one was wrong for as long as it went unused, and only became visible when someone finally tried to follow it. That is worth stating because it is the argument for the whole series: the claims that rot are the ones nothing exercises.

Nothing outstanding from me.

nstfkc:
## Follow-up on `e07908e` — the correction is right, and it corrects a row I signed off on

I reviewed `f1ea450` and said the five-row table was yours to stand behind, since I have no Postgres. I did check it for internal coherence and reported it as consistent. This commit shows that one of those rows was wrong, and my "coherent" told nobody anything — `to_jsonb($1)` reads as a working form from the row that measures it, which is exactly why the row alone was not enough.

Worth being explicit about that: I verified everything reachable without a database and then described the rest as measured. The part I could not reach is the part that was wrong. That is the correct division of labour and it is also its limit.

**The corrected claim is the stronger one**, and the test now says why rather than asserting a conclusion:

```js
expect((await insert("to_jsonb($1)", 42)).kind).toBe("number");
expect((await insert("to_jsonb($1)", true)).kind).toBe("boolean");
for (const value of [{ a: 1 }, [1, 2], "42", null]) {
  await expect(insert("to_jsonb($1)", value)).rejects.toThrow(/polymorphic type/);
}
```

Pinning the two shapes it *does* carry alongside the four it does not is what stops the next reader adopting it from the same partial evidence — the same reasoning as keeping the `$1::jsonb` failure in the file, applied to a form that fails less obviously. Six shapes through `$1::text::jsonb`, asserting the value **and** `jsonb_typeof`, is the right coverage: the mis-store this replaces was a right-looking value under the wrong type, so checking either alone would have missed it.

**Composed against the merged branch**, now that #205 has landed and its naming guard is real rather than a forward reference:

```
patch onto feat/orm (2a1d097)          applies clean
postgres-suite-selection.test.ts       6 passed
orm database                           51 files, 1430 passed
-t postgres selects the renamed test   yes
```

So the rename from "the two bindings" to "the one binding" keeps the suffix that makes it selectable, and the guard that now enforces that is satisfied — the thing I could only assert by construction last time is checked this time.

Nothing outstanding.

PR #212, opened 2026-07-29
Postgres deployments need TZ=UTC and the docs never said so

### LINKED ISSUE #211: Postgres deployments need TZ=UTC and the docs never say so
`dialect/postgres.ts` carries a comment headed **KNOWN DIVERGENCE**:

> Prisma maps `DateTime` to `timestamp(3)` — no time zone — and stores UTC in it, but Bun's driver decodes that column differently depending on the *protocol* the statement used. […] Same row, same column, two instants, on any machine whose clock is not already UTC. […] Until it is fixed upstream, run the process with `TZ=UTC`.

`docs/orm.md` does not contain the string `TZ`, `UTC`, "timezone" or "time zone" anywhere.

What it *does* say, of the bare-JSON-number refusal:

> **This is the one shape where gemi diverges from Prisma**

So the more severe of the two divergences is recorded only where the people affected by it would never look, and the page that would have told them says there is nothing else to know.

### Reproduced against Postgres 16

Not taken from the comment — measured:

```
TZ=UTC                no parameters -> 2021-03-04T05:06:07.008Z    where id = $1 -> 05:06:07.008Z
TZ=America/New_York   no parameters -> 2021-03-04T10:06:07.008Z    where id = $1 -> 05:06:07.008Z
```

In ORM terms:

```ts
await User.findMany()                    // binds nothing -> simple protocol  -> WRONG instant
await User.findMany({ where: { id } })   // binds one     -> extended protocol -> correct
```

Whether a query binds a parameter is what selects the wire protocol, and that is not something a reader can guess. Two reads of the same row disagree by the machine's UTC offset, silently — nothing raises.

### Why no test caught it

CI and the suites run under `TZ=UTC`, which is the correct setting and also the one where both paths agree. This was found by reading the comment, not by a failure.

### Fix

Document it as the deployment requirement it is, and narrow the Json claim to "the one *write* shape", which is what it was accurate about.

Guard it so the documentation cannot quietly disappear again — including that the section says *which* queries are affected rather than only that some are, since that is the actionable half.

### Not proposed here

A behavioural fix. The dialect's assessment is that the ORM cannot correct it below the query, because the decoded value does not say which protocol produced it. Worth noting the same comment says "nothing *below the plan* does" — and the plan knows its own binder count, so a correction one level up may be possible. That deserves its own issue and its own differential pass.

nstfkc:
Closes #211.

`dialect/postgres.ts` carries a comment headed **KNOWN DIVERGENCE**: Prisma maps `DateTime` to `timestamp(3)`, and Bun's driver decodes that column differently depending on which wire protocol carried the statement — which depends on whether the query binds a parameter.

`docs/orm.md` did not contain `TZ`, `UTC`, "timezone" or "time zone" anywhere. What it *did* say, of the bare-JSON-number refusal, was that it is "**the one shape where gemi diverges from Prisma**".

So the more severe of the two divergences was recorded only where the people affected by it would never look, and the page that would have told them said there was nothing else to know.

## Reproduced, not quoted

```
TZ=UTC                no parameters -> 2021-03-04T05:06:07.008Z    where id = $1 -> 05:06:07.008Z
TZ=America/New_York   no parameters -> 2021-03-04T10:06:07.008Z    where id = $1 -> 05:06:07.008Z
```

In ORM terms:

```ts
await User.findMany()                    // binds nothing -> simple protocol   -> wrong instant
await User.findMany({ where: { id } })   // binds one     -> extended protocol -> correct
```

Same row, same column, two instants, off by the machine's UTC offset, and nothing raises. Which queries are affected is the part a reader cannot guess, so the new section leads with it.

The Json sentence is narrowed to "the one **write** shape" and points at the new section — it was accurate about writes and read as though it covered everything.

## No behaviour change

The ORM cannot correct it below the query: the decoded value alone does not say which protocol produced it. That is the dialect's own assessment and I have not revisited it here — though the same comment says "nothing *below the plan* does", and the plan knows its own binder count, so a fix one level up may exist. Recorded in #211 as its own piece of work rather than attempted alongside a docs change.

## Why nothing caught it

CI and both suites run under `TZ=UTC` — the correct setting, and also the one where the two paths agree. No test would ever have failed. It was found by reading the comment.

## The guard

Pins each divergence to the section documenting it, asserts the section names *which* queries are affected rather than only that some are, and checks the dialect still carries the `KNOWN DIVERGENCE` note the docs are derived from — so deleting the source of truth fails rather than leaving prose describing nothing.

Verified against the docs as they stood:

```
× the page documents the Postgres TZ requirement
× it says which queries are affected, not just that some are
× the Json refusal is not called the only divergence
```

`docs/llms-full.txt` is re-synced, so #181's verbatim guard stays green — it caught the drift as intended while I was writing this.

## Checks

- `bun --bun vitest run orm database` — 52 files, 1436 passed
- scratch Postgres container removed

Independent of #204, #208 and #210.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3828,8 +3828,10 @@ Two things are worth knowing:
   stored, because the driver binds it as an integer and the column is `jsonb`. Wrap it — `{ value:
   42 }` — or store it as a string. SQLite has no such limit.
 
-  **This is the one shape where gemi diverges from Prisma**, which stores it on both dialects, so it
-  is worth knowing if you are porting code rather than writing it fresh. It is a trade rather than
+  **This is the one *write* shape where gemi diverges from Prisma**, which stores it on both
+  dialects, so it is worth knowing if you are porting code rather than writing it fresh. (The other
+  divergence is on reads, and is a deployment requirement rather than an API difference — see
+  [Run Postgres deployments with `TZ=UTC`](#run-postgres-deployments-with-tzutc).) It is a trade rather than
   an oversight: the encoder that accepted it serialised first, which stored `42` as the jsonb
   *string* `"42"` — and only looked correct because the decoder re-parsed it on the way out, so the
   value was wrong in the database the whole time. Failing loudly beats storing the wrong thing.
@@ -3842,6 +3844,34 @@ Two things are worth knowing:
 > so nothing looked wrong from inside the ORM, but the column was wrong for anything else that read
 > it. Those rows now read back as strings. Re-seed development databases; there is no released
 > version affected.
+
+## Run Postgres deployments with `TZ=UTC`
+
+**A deployment requirement, not a preference.** Prisma maps `DateTime` to `timestamp(3)` — no time
+zone — and stores UTC in it. Bun's driver decodes that column differently depending on which wire
+protocol carried the statement, and *which protocol is used depends on whether the query binds a
+parameter*:
+
+```ts
+await User.findMany()                      // no parameters -> simple protocol
+await User.findMany({ where: { id } })     // one parameter -> extended protocol
+```
+
+The first comes back as zoneless text and is parsed as **local** time; the second comes back in
+binary and is correct. Same row, same column, two different instants — off by your machine's UTC
+offset. Measured against Postgres 16:
+
+```
+TZ=UTC                no parameters -> 2021-03-04T05:06:07.008Z    where id = $1 -> 05:06:07.008Z
+TZ=America/New_York   no parameters -> 2021-03-04T10:06:07.008Z    where id = $1 -> 05:06:07.008Z
+```
+
+Set `TZ=UTC` on any process that talks to Postgres and both paths agree. This is the setting the
+test suites and CI already run under.
+
+The ORM cannot correct it below the query: the decoded value alone does not say which protocol
+produced it. SQLite is unaffected — it stores `DateTime` as milliseconds and there is no text
+representation to reinterpret.
 
 ## Dialects
 
@@ -3905,7 +3935,6 @@ might lift next release.
   starter template still points at the Prisma adapter; switching it is a per-application call.
 - **[Authorization](./authorization.md)** — route-level authorization, which policies complement
   rather than replace.
-
 ---
 
 ## Rows and entities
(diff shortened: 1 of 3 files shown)


nstfkc:
## Review — Postgres deployments need `TZ=UTC` and the docs never said so (closes #211)

The omission is exact. On `2a1d097`, `docs/orm.md` contains:

```
TZ           0
UTC          0
timezone     0
time zone    0
```

while `dialect/postgres.ts:369` carries the whole thing under **KNOWN DIVERGENCE**, down to the two worked examples. And the page's only statement about divergence was `:1136` — *"This is the one shape where gemi diverges from Prisma"* — about the bare-JSON refusal. So the page a deployer reads asserted there was nothing else, three hundred lines from the section that would have mattered more.

Narrowing that to "the one **write** shape" and pointing at the new section is the right repair: the sentence was true of writes and read as a total.

**"Why nothing caught it" is the part worth keeping**, and it checks out — `TZ: UTC` appears three times in `ci.yml`, on both unit steps and the template step. So the suites run in exactly the configuration where the two protocol paths agree, and no test could have failed. It was found by reading the comment, which is the only way it could have been found.

The guard reproduces in both directions, and I probed the half that is easy to leave out:

```
base docs      × the page documents the Postgres TZ requirement
               × it says which queries are affected, not just that some are
               × the Json refusal is not called the only divergence
KNOWN DIVERGENCE renamed in the dialect
               × the dialect still carries the note the docs are derived from
```

That last one is what stops the prose outliving its source — the failure mode #189 and #193 were both instances of. Asserting the section names *which* queries are affected, rather than that some are, is the other half: "some queries are wrong" is unactionable, and `findMany()` versus `findMany({ where })` is the distinction a reader cannot derive.

**The note about a possible fix is well placed and I checked it against the source.** The comment says a `decode` cannot correct this because *"the value alone does not say which protocol produced it; nothing below the plan does"* — and the plan does know its own binder count, so the observation that a fix may exist one level up follows from the comment's own wording rather than from speculation. Filing it as #211's own work rather than attempting it beside a docs change is the right call; the docs are true either way.

`orm database` reproduces at 52 files / 1436 passed.

One thing I would add to the new section rather than change: it tells a deployer to set `TZ=UTC`, and the reason nothing caught this is that the test suites already do. Those are the same fact seen twice, and a sentence saying so — *the suites run under `TZ=UTC`, which is why no test fails and why your production process must too* — would make the requirement harder to read as optional.

PR #214, opened 2026-07-29
Warn once when a Postgres read will return the wrong instant

### LINKED ISSUE #213: Nothing tells a misconfigured process that its timestamps are wrong
The follow-up recorded in #211, which #212 documents but does not detect.

Bun decodes `timestamp(3)` over the simple query protocol as local time, and a query that binds no parameters takes that protocol. On a non-UTC machine:

```ts
await User.findMany()                    // -> 2021-03-04T10:06:07.008Z
await User.findMany({ where: { id } })   // -> 2021-03-04T05:06:07.008Z
```

Same row, same column, off by the machine's UTC offset. Nothing raises, and both values look plausible.

A reader who has seen the documentation sets `TZ=UTC` and never meets this. One who has not gets wrong timestamps and no signal at all — the kind of wrongness that reaches a report or an invoice before anyone notices.

### What can be done

Not a correction. The value is already decoded before anything in this package sees it, and it does not carry which protocol produced it — that is the dialect's own assessment and it holds.

What *is* available is the condition: `execute()` sees the dialect, the schema, the statement text and the bound values, and `values.length === 0` is exactly the discriminator. So the fault is detectable at the one point every statement passes through, even though it is not fixable there.

The project's stated trade, from the `Json` encoder that had the same shape of bug, is that **"a loud failure replaces a silent mis-store"**. A warning is the analogous move.

### Proposed shape

Warn at most once per process, when *all* of: Postgres, no bound parameters, a `select`, the model has a `DateTime` field, and the process clock is not UTC. A correctly configured deployment never sees it and pays one boolean.

**Not gated on `NODE_ENV`**, unlike the slow-transaction warning beside it — that one is a performance hint, worth having only where somebody is watching; this is a correctness fault, and production is where an unnoticed wrong timestamp costs something. Worth flagging as the debatable part of the proposal.

### Testing it

The helper should be pure and return the message rather than print it, taking the UTC offset as a parameter. CI runs under `TZ=UTC` — the configuration where the fault does not occur — so a test that reads the clock would be untestable in the one place it must run.

Each condition needs asserting separately: three of the four are cheap comparisons, and "no warning" is what a correctly configured process looks like, so a behavioural test cannot distinguish a working guard from a deleted one.

nstfkc:
Closes #213. The follow-up #211 records — #212 documents the requirement, this tells the process that is not meeting it.

Bun decodes `timestamp(3)` over the simple query protocol as local time, and a query that binds no parameters takes that protocol. On a non-UTC machine `User.findMany()` and `User.findMany({ where: { id } })` return **different instants for the same row**, off by the machine's UTC offset, with nothing raised and both values looking plausible.

A reader who has seen the documentation sets `TZ=UTC` and never meets this. One who has not gets wrong timestamps and no signal — the kind of wrongness that reaches a report before anyone notices.

## Not a correction, a signal

The value is already decoded before anything in this package sees it, and does not carry which protocol produced it. That is the dialect's own assessment and it holds.

But the *condition* is visible: `execute()` — the single point every statement runs through — has the dialect, the schema, the text and the bound values, and `values.length === 0` is exactly the discriminator. So the fault is detectable where it is not fixable.

The project's stated trade, from the `Json` encoder that had the same shape of bug, is that "a loud failure replaces a silent mis-store".

## The debatable part, stated plainly

**It is not gated on `NODE_ENV`**, unlike the slow-transaction warning beside it. That one is a performance hint, worth having only where somebody is watching; this is a correctness fault, and production is exactly where an unnoticed wrong timestamp costs something.

It fires at most once per process and only when every condition holds — Postgres, no bound parameters, a `select`, a `DateTime` column, and a clock that is not UTC — so a correctly configured deployment never sees it and pays one boolean. Easy to make dev-only if you would rather; it is one condition.

## Testing something CI cannot reproduce

CI runs under `TZ=UTC` — the configuration where the fault does not occur, and otherwise the one environment unable to test it. So the helper is pure, returns the message rather than printing it, and takes the UTC offset as a parameter.

Each condition is asserted separately, because three of the four are cheap comparisons a later edit could drop without any behavioural test noticing: "no warning" is what a correctly configured process looks like, so it cannot distinguish a working guard from a deleted one.

Verified end to end against Postgres 16 as well as by unit test:

```
TZ=UTC               two findMany() calls -> 0 warnings
TZ=America/New_York  two findMany() calls -> 1 warning
```

The second confirms both halves: that it fires, and that it fires once.

## Checks

- `bun --bun vitest run orm database` — 53 files, 1444 passed
- template suite — 17 passed / 3 skipped, 616 tests
- scratch container removed

Builds on #212's documentation but does not depend on it to merge.

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -20,6 +20,7 @@ import { resolveStrategy } from "./compile/strategy";
 import { matchUniqueKey } from "./compile/unique";
 import { upsertAbsentConflictKey } from "./compile/write";
 import { dialectFor, type SqlDialect } from "./dialect";
+import { protocolSkewWarning } from "./protocol-skew";
 import {
   MissingModelSchemaError,
   RecordNotFoundError,
@@ -956,6 +957,9 @@ async function runSteps(
   }
 }
 
+/** Set the first time `protocolSkewWarning` fires; see the call site below. */
+let warnedProtocolSkew = false;
+
 /**
  * Runs the statement and translates the failures that have a typed home.
  *
@@ -974,6 +978,24 @@ async function execute(
   text: string,
   values: unknown[],
 ): Promise<unknown> {
+  // Said once per process, and only for the configuration that silently returns
+  // the wrong instant — see `protocol-skew.ts`. Behind `warnedProtocolSkew` so
+  // the check disappears after it fires, and ordered so a correctly configured
+  // process pays one boolean.
+  if (!warnedProtocolSkew) {
+    const warning = protocolSkewWarning(
+      dialect.name,
+      schema,
+      text,
+      values,
+      new Date().getTimezoneOffset(),
+    );
+    if (warning) {
+      warnedProtocolSkew = true;
+      console.warn(warning);
+    }
+  }
+
   try {
     return await conn.unsafe(text, values);
   } catch (error) {
--- a/packages/gemi/orm/protocol-skew.test.ts
+++ b/packages/gemi/orm/protocol-skew.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, test } from "vitest";
+
+import { protocolSkewWarning } from "./protocol-skew";
+import type { ModelSchema } from "./schema";
+
+/**
+ * The conditions under which a `DateTime` read is silently wrong, asserted one
+ * at a time.
+ *
+ * Every one of them has to hold, and the reason to test them individually is
+ * that three of the four are cheap comparisons which a later edit could reorder
+ * or drop without any behavioural test noticing: the warning not firing is what
+ * a correctly configured process looks like, so "no warning" cannot distinguish
+ * a working guard from a deleted one.
+ *
+ * `offsetMinutes` is a parameter rather than read from the clock so these run
+ * the same in CI — which sets `TZ=UTC`, the configuration where the fault does
+ * not occur, and would otherwise be the one environment unable to test it.
+ */
+const field = (name: string, type: string) => ({
+  name,
+  column: name,
+  type: type as never,
+  nullable: false,
+  isId: false,
+  isUpdatedAt: false,
+});
+
+const withDates: ModelSchema = {
+  name: "User",
+  table: "User",
+  fields: {
+    id: field("id", "Int"),
+    createdAt: field("createdAt", "DateTime"),
+    updatedAt: field("updatedAt", "DateTime"),
+  },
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — warn once when a Postgres read will return the wrong instant (closes #213)

**The discriminator is sound where it is placed.** `execute()` receives the same `values` array it then hands to `conn.unsafe(text, values)`, so `values.length === 0` is literally "this statement binds no parameters" — the condition the dialect comment describes, evaluated at the only point where it is still knowable. Detecting a fault where it cannot be corrected is the right response to the dialect's assessment rather than a workaround for it.

**Each condition is separately pinned, and I checked that by removing them** rather than reading the test names:

```
drop the dialect check    1 failed
drop values.length > 0    1 failed
drop offsetMinutes === 0  1 failed
drop the select check     1 failed
drop columns.length === 0 1 failed
```

Exactly one each, which is what the reasoning asks for — "no warning" is what a correct process looks like, so a behavioural test cannot tell a working guard from a deleted one.

Making the helper pure and taking the offset as a parameter is the right answer to "CI runs in the one configuration that cannot reproduce this". And the `select` test is `/^\s*select/i` — case-insensitive, which is worth noting given three guards on this branch currently are not.

**Stating the `NODE_ENV` decision as debatable rather than settling it quietly** is right, and the argument is the correct shape: the slow-transaction warning is a performance hint, this is a correctness fault, and production is where an unnoticed wrong timestamp costs something.

### The once-per-process flag is the one part with no test

That argument depends on the warning being *once*. Removing the assignment:

```ts
if (warning) {
-  warnedProtocolSkew = true;
   console.warn(warning);
}
```
```
orm database   1444 passed
```

Nothing notices. Every qualifying read would warn, which for a warning deliberately left on in production is the difference between one line and a log full of them — and it is the specific claim that makes the not-gating-on-`NODE_ENV` case work.

You verified it end to end against Postgres, and that is real evidence — but it is evidence from the environment this PR argues CI cannot reproduce, so the once-ness is now as untested as the condition the helper was made pure to escape. The flag is the only stateful part of the change and it is the only part left outside that purity.

The same move fixes it: move the latch into `protocol-skew.ts` — a `protocolSkewWarningOnce()` closing over the boolean, with a reset for tests, or simply have the helper return `null` after its first non-null — and the second call becomes a unit assertion with no connection and no timezone. `Model.ts` then keeps the `console.warn` and nothing else.

`orm database` reproduces at 53 files / 1444 passed.

PR #216, opened 2026-07-29
Lift the Json boundary: bare numbers and booleans write on Postgres

### LINKED ISSUE #215: Lift the Json boundary now that the working form is known
#209 established which binding form actually stores a bare JSON scalar correctly on Postgres, and deliberately stopped there. This is the remaining half.

Today `metadata: 42` raises on Postgres — Bun binds it as an `integer` and the column is `jsonb` — while Prisma stores it. It is the last documented divergence from Prisma.

### The form, from #209's measurements

```
values ($1)                 42        integer vs jsonb
values ($1::jsonb)          "42"      jsonb_typeof -> string      <- the old silent mis-store
values (to_jsonb($1))       {a:1}     could not determine polymorphic type
values ($1::text::jsonb)    "42"      jsonb_typeof -> number      <- carries all six shapes
```

### Where the serialisation must not go

`encode`. It runs at **every** binding site, and a site that serialises without also emitting the cast is the second row above — the mis-store the refusal exists to prevent, arrived at silently.

Keeping the serialisation and the cast together means the failure mode of missing a binding site is that the site keeps binding raw: objects still work, a bare number is still refused, and the failure stays loud. That is the difference between an incomplete change and a regression.

### Scope

Every site that binds a *field value* needs it — the insert values clause, the update set clause, defaults, and `where` comparisons. `inList` is out: an `in` over a Json column would need `::jsonb[]` against a serialised array literal, which is a separate question.

`null` must stay SQL NULL rather than becoming the jsonb value `null` — `'null'::jsonb` is a JSON null, a different thing from an absent value, and would diverge from Prisma on every nullable Json column.

### Verification

The differential harness compares Json columns against Prisma directly, so a full Postgres run is the check. The round-trip test should assert `jsonb_typeof` as well as the value: the mis-store being replaced was a right-looking value under the wrong type, which is exactly how it survived.

nstfkc:
Closes #215. The last documented divergence from Prisma, and the half #209 deliberately stopped short of.

A bare JSON number or boolean was refused on Postgres because Bun binds it as its own type and the column is `jsonb`. The compiler now asks the dialect for a cast and serialises to match — `$1::text::jsonb` — the only one of the four forms measured in #209 that carries all six shapes:

```
values ($1)                 42        integer vs jsonb
values ($1::jsonb)          "42"      jsonb_typeof -> string
values (to_jsonb($1))       {a:1}     could not determine polymorphic type
values ($1::text::jsonb)    "42"      jsonb_typeof -> number
```

## The design choice worth reviewing

**The serialisation is deliberately not in `encode`.** That is the obvious place and the wrong one: `encode` runs at every binding site, so a site that serialises without *also* emitting the cast is the second row above — the silent mis-store the refusal existed to prevent.

Keeping both in `fieldParam` means the failure mode of missing a site is that the site keeps binding raw: objects still work, a bare number is still refused, and it stays loud. That is the difference between an incomplete change and a regression.

Eight sites converted — five in `write.ts` (insert values, update set clause, defaults) and three in `where.ts`. `inList` is left alone: an `in` over a Json column would need `::jsonb[]` against a serialised array literal, a different question nobody has asked.

`null` stays SQL NULL rather than becoming the jsonb value `null` — `'null'::jsonb` is a JSON null, a different thing from an absent value, and would diverge from Prisma on every nullable Json column.

The cast rides as text after the placeholder rather than in a parallel array, since the marker is replaced in place at render — text following it needs no bookkeeping and cannot fall out of step with the binders. It is the one string a dialect contributes to the statement that is not a parameter, so the interface says it must be a constant.

## Verification

Through the differential harness, which compares every one of these against Prisma:

| | before | after |
| --- | --: | --: |
| Postgres suite | 646 | **649 passed, 0 failed** |
| SQLite suite | 617 | **620 passed** |
| `packages/gemi` | 1436 | **1437 passed** |

The round-trip test asserts `jsonb_typeof` as well as the value, because the mis-store this replaces was a *right-looking value under the wrong type* — `"42"` reads back as `42` through this ORM and as a string through everything else, which is how it survived.

## Interaction with #212

#212 adds a guard asserting `docs/orm.md` calls the Json refusal "the one **write** shape where gemi diverges from Prisma". This PR removes that divergence, so that sentence is gone and the assertion will need dropping — whichever merges second. Flagging rather than pre-empting, since #212 is unreviewed and the wording may change.

`docs/llms-full.txt` is re-synced. Scratch container removed and the template restored to SQLite.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3817,25 +3817,20 @@ development on SQLite and survives the rollback in production on Postgres. `DB.q
 
 ## `Json` columns
 
-A `Json` column round-trips whatever you give it: an object, an array, a string, `null` — and a
-number or boolean on SQLite, with the one exception below. The value is stored as JSON and comes
-back as the same JavaScript value, on both dialects and whether it is read at the root or nested
-inside an `include`.
+A `Json` column round-trips whatever you give it: an object, an array, a string, a number, a
+boolean, `null`. The value is stored as JSON and comes back as the same JavaScript value, on both
+dialects and whether it is read at the root or nested inside an `include`.
 
-Two things are worth knowing:
+One thing is worth knowing:
 
-- **A bare JSON number or boolean is refused on Postgres.** `metadata: 42` raises rather than being
-  stored, because the driver binds it as an integer and the column is `jsonb`. Wrap it — `{ value:
-  42 }` — or store it as a string. SQLite has no such limit.
-
-  **This is the one shape where gemi diverges from Prisma**, which stores it on both dialects, so it
-  is worth knowing if you are porting code rather than writing it fresh. It is a trade rather than
-  an oversight: the encoder that accepted it serialised first, which stored `42` as the jsonb
-  *string* `"42"` — and only looked correct because the decoder re-parsed it on the way out, so the
-  value was wrong in the database the whole time. Failing loudly beats storing the wrong thing.
 - **A string is a string.** `metadata: "42"` stores the JSON string `"42"`, not the number, and
   `metadata: '{"a":1}'` stores that text as a string rather than as an object. If you want an
   object, pass an object.
+
+On Postgres the parameter is cast — `$1::text::jsonb` — and the value serialised to match, which
+is what lets a bare number or boolean through. Binding it raw makes the driver send an `integer` to
+a `jsonb` column, and serialising *without* the cast stores the jsonb string `"42"` instead of the
+number, which is worse than refusing it. SQLite needs neither: it stores JSON as text.
 
 > **Upgrading:** `Json` values written by a *pre-release* build of this ORM on Postgres were stored
 > as JSON **strings** rather than as objects — `{ a: 1 }` landed as `"{\"a\":1}"`. Reads undid it,
@@ -3905,7 +3900,6 @@ might lift next release.
   starter template still points at the Prisma adapter; switching it is a per-application call.
 - **[Authorization](./authorization.md)** — route-level authorization, which policies complement
   rather than replace.
-
 ---
 
 ## Rows and entities
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1122,25 +1122,20 @@ development on SQLite and survives the rollback in production on Postgres. `DB.q
 
 ## `Json` columns
 
-A `Json` column round-trips whatever you give it: an object, an array, a string, `null` — and a
-number or boolean on SQLite, with the one exception below. The value is stored as JSON and comes
-back as the same JavaScript value, on both dialects and whether it is read at the root or nested
-inside an `include`.
+A `Json` column round-trips whatever you give it: an object, an array, a string, a number, a
+boolean, `null`. The value is stored as JSON and comes back as the same JavaScript value, on both
+dialects and whether it is read at the root or nested inside an `include`.
 
-Two things are worth knowing:
+One thing is worth knowing:
 
-- **A bare JSON number or boolean is refused on Postgres.** `metadata: 42` raises rather than being
-  stored, because the driver binds it as an integer and the column is `jsonb`. Wrap it — `{ value:
-  42 }` — or store it as a string. SQLite has no such limit.
-
-  **This is the one shape where gemi diverges from Prisma**, which stores it on both dialects, so it
-  is worth knowing if you are porting code rather than writing it fresh. It is a trade rather than
-  an oversight: the encoder that accepted it serialised first, which stored `42` as the jsonb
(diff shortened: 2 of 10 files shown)


nstfkc:
**Rebased onto `feat/orm` after #212 merged, and the assertion it added is retired.**

This is the interaction flagged above. #212 guards that `docs/orm.md` calls the Json refusal "the one **write** shape where gemi diverges from Prisma"; this PR removes the refusal, so the sentence is gone and the assertion could not survive.

It is rewritten rather than deleted. The assertion has now moved twice:

1. *"the Json refusal is not called **the** only divergence"* — the page claimed it was the only one while the `TZ` fault went unmentioned (#211)
2. requiring the narrowed *"the one **write** shape"* (#212)
3. asserting the page claims **no** Json divergence at all (here)

The failure it exists to catch is a page describing a divergence that is not there — as wrong in this direction as it was in the other.

#212's `TZ=UTC` section is untouched and still reachable; only the cross-reference that lived inside the deleted bullet went with it.

Re-verified after the merge, not just before it:

- Postgres differential — 13 files, **649 passed, 0 failed**
- SQLite suite — 17 passed / 3 skipped, **620 tests**
- `packages/gemi` — 53 files, **1441 passed**
- typecheck and `oxlint --deny-warnings` clean

nstfkc:
## Review — the cast reaches the placeholder, and one branch of it is unpinned

Ran this against a real Postgres 16 rather than reading it. The fix works, and the new assertions catch its absence: the same suite against `feat/orm` (`86d3ccb`) fails eight cases with exactly the mis-store described — `jsonb_typeof` answering `string` for a number, a boolean, an object *and* an array — and passes 23 on this branch. So `$1::text::jsonb` is doing what the table says.

The design argument is the right one, and it is checkable rather than stylistic: I reverted the `where.ts` conversions and the three shapes that go through them still behave as before, which is the "an unconverted site keeps binding raw" claim holding under test rather than in prose.

### The `null` branch is load-bearing, and nothing pins it

`fieldParam`'s null guard has the longest comment in the file and no test. Removing it:

```
optjson IS NULL      false
jsonb_typeof         null        <- the jsonb value, not an absent one
```

Against a nullable `Json` column, written through this ORM, on Postgres. And the row is then unfindable by the ORM that wrote it, because `equals` compiles `{ optjson: null }` to `is null` — which a jsonb `null` does not satisfy. Written and read through gemi, in the same test, the write succeeds and the read returns nothing.

With the guard removed the whole tree stays green: 1413 in `packages/gemi/orm`, 23 in the Postgres coercion suite, 211 in the Postgres template run. Nothing anywhere writes `null` to a `Json` column *through gemi* — `Everything.payload` is `JSONB NOT NULL` in both DDLs, so the new five-shape test cannot reach it, and the one `settings: null` in the repository is inside `prisma.account.createMany`, a seed. That is Prisma's write exercising the decoder, which is why #187's "`Prisma.JsonNull` vs `Prisma.DbNull`" comment sits next to it — the read side of this distinction is covered and the write side is not.

This is the same failure the `jsonb_typeof` assertions were added for, one column over. A SQL NULL and a jsonb `null` both decode to JS `null`, so a round-trip comparison cannot tell them apart; only the database can be asked, and here nothing asks. A nullable `Json` column and `expect(sqlnull).toBe(true)` is the whole test.

(Caveat on the numbers: the differential suites could not run here — the Prisma client in this checkout is generated for sqlite and regenerating it writes outside the worktree — so that count is the non-Prisma Postgres suites. The grep above is what covers the differential half.)

### The three `where.ts` conversions have no test either

They work — I checked, since nothing else does:

```
where { payload: 7 }        finds the row      <- newly possible
where { payload: true }     finds the row      <- newly possible
where { payload: "42" }     finds the row
where { payload: {a:1} }    InvalidArgumentError: Invalid 'where.payload.a'
```

The last is pre-existing and identical on both dialects — an object operand is read as a filter object, so a `Json` column cannot be compared to an object at all. Loud, and not this PR's business, but worth knowing it is the boundary that remains after the write boundary is gone.

The first two are a capability this change adds beyond writes, and the description does not claim them. Worth one case, because "the cast reaches `where`" is a third of the converted sites and the whole suite passes without it.

### One smaller thing

`write.ts:1117` binds a field value through plain `param()` and is not among the eight. It is the foreign-key contribution path, so it is never a `Json` column and the omission is harmless — it is in fact the design working, since an unconverted site keeps binding raw. But "eight sites converted" reads as an enumeration of every site that binds a field, and there are nine. A clause naming the ninth and why it stays saves the next reader the search.

Nothing above is a reason not to merge. The null case is the one I would want before it, since it is the shape this PR exists to make impossible.

PR #219, opened 2026-07-29
Compare a Prisma enum against Prisma, and say how it sorts

### LINKED ISSUE #218: Enums are never compared against Prisma, and never documented
The generator has a dedicated branch for Prisma enums:

```ts
type: field.kind === "enum" ? "String" : scalarType(model, field),
...
if (field.kind === "enum") schema.enum = field.type;
```

`bin/orm/emit.test.ts:156` covers that mapping. Nothing runs it against a database.

- `templates/saas-starter/prisma/schema.prisma` contains **no enum**, so the differential harness has never compared one against Prisma — no read, no `where`, no `orderBy`, no `groupBy`.
- `docs/orm.md` mentions enums **zero** times.

The schema itself states the principle it is missing, in a comment on the columns added for exactly this reason:

> Json and Bytes exist here so the differential harness sees them.

### Not a connector limitation

My first assumption was that Prisma's SQLite connector rejects enums, which would have explained the absence. It does not — Prisma 6.19 validates an enum against `provider = "sqlite"` (`The schema is valid 🚀`). So there is nothing stopping this being covered.

### The thing coverage would have found

The two dialects **sort an enum differently**. Measured, for `enum OrganizationPlan { free pro enterprise }`:

```
postgres   free, pro, enterprise      (real enum type -> declaration order)
sqlite     enterprise, free, pro      (stored as TEXT -> alphabetical)
```

Both match Prisma on their own dialect, so it is parity rather than a gemi decision — the same call the codebase already documents for `contains` being case-sensitive on Postgres and insensitive on SQLite. But an application that sorts by an enum and is developed on SQLite will order differently in production, and nothing anywhere said so.

### Fix

Add an enum to the template schema with a migration, cover it in the differential suite, and document the ordering difference.

One trap to avoid: seeding every row to the same enum value — or to the column's default — makes the ordering, `in` and `groupBy` cases pass while discriminating nothing, which is how #124's `having` corpus went vacuous. The seed should use distinct non-default values, and that should itself be asserted.

nstfkc:
Closes #218.

The generator maps a Prisma `enum` field to `String` and records the enum's name beside it. `bin/orm/emit.test.ts` covers that mapping; nothing ran it against a database, because the template's schema carried no enum — so the differential harness had never seen one. The schema's own comment states the principle it was missing: *"Json and Bytes exist here so the differential harness sees them."*

Not a connector limitation, which was my first guess and wrong: Prisma 6.19 validates an enum against the sqlite provider.

## What coverage found

**The two dialects sort an enum differently** — and this is the part I nearly got wrong. I wrote "sorts alphabetically, not by declaration order" into the docs from memory, then measured it:

```
postgres   free, pro, enterprise      (real enum type -> declaration order)
sqlite     enterprise, free, pro      (TEXT -> alphabetical)
```

Both match Prisma on their own dialect, so it is parity rather than a gemi decision — the same call the codebase already makes for `contains`. But an application that sorts by an enum and is developed on SQLite orders differently in production, and nothing said so. `docs/orm.md` had no `Enums` section at all; it now has one, carrying the measured table rather than the sentence I first wrote.

## The change

`Organization.plan` with a default, so no existing insert changes, plus a migration. `Organization` is shadowed by no test fixture, so #208's guard stays quiet. Eight differential cases: read, `select`, equality, `not`, `in`, no-match, `orderBy`, `groupBy`.

## Guarding against a vacuous pass

The seed gives the two organizations **different** plans, neither of them the column's default. An enum seeded to one value compares equal under every ordering and grouping, so those cases would stay green while discriminating nothing — which is how #124's `having` corpus went vacuous.

A ninth test asserts that directly, on the rows `expectSame` hands back:

```ts
expect(new Set(plans).size).toBe(plans.length);
expect(plans).not.toContain("free");
```

## Checks

| | before | after |
| --- | --: | --: |
| Postgres differential | 649 | **657 passed, 0 failed** |
| SQLite suite | 620 | **626 passed** |
| `packages/gemi` | 1441 | **1450 passed** |

`bunx tsc --noEmit` exit 0, `bunx oxlint orm --deny-warnings` 0 warnings — both read in full this time, after last iteration's `tail -1` hid a warning.

Scratch container removed; the template's schema and Prisma client restored to SQLite. Independent of #204 and #216.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3815,6 +3815,32 @@ The direction is the awkward one: a `DB.sql` write inside a transaction rolls ba
 development on SQLite and survives the rollback in production on Postgres. `DB.query` and
 `DB.execute` join the transaction on both.
 
+## Enums
+
+A Prisma `enum` is a **string** at runtime, on both dialects, exactly as it is through Prisma:
+
+```ts
+await Organization.findMany({ where: { plan: { in: ["pro", "enterprise"] } } })
+```
+
+The generated schema records the enum's name beside the field, but the value you write and the
+value you read back are the plain member name. Equality, `not`, `in`, `orderBy` and `groupBy` all
+work on it and are compared against Prisma by the differential suite.
+
+One consequence worth stating, because it is the one thing that surprises: **`orderBy` on an enum
+sorts differently on the two dialects.** Postgres creates a real enum type and sorts by
+*declaration* order; SQLite stores the member as `TEXT` and sorts *alphabetically*. For
+`enum OrganizationPlan { free pro enterprise }`, measured:
+
+```
+postgres   free, pro, enterprise      (declaration order)
+sqlite     enterprise, free, pro      (alphabetical)
+```
+
+gemi matches Prisma on each dialect rather than inventing a uniformity Prisma does not have — the
+same call as `contains` being case-sensitive on Postgres and not on SQLite. If you need one order
+everywhere, sort in the application or store an explicit rank column.
+
 ## `Json` columns
 
 A `Json` column round-trips whatever you give it: an object, an array, a string, `null` — and a
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1120,6 +1120,32 @@ The direction is the awkward one: a `DB.sql` write inside a transaction rolls ba
 development on SQLite and survives the rollback in production on Postgres. `DB.query` and
 `DB.execute` join the transaction on both.
 
+## Enums
+
+A Prisma `enum` is a **string** at runtime, on both dialects, exactly as it is through Prisma:
+
+```ts
+await Organization.findMany({ where: { plan: { in: ["pro", "enterprise"] } } })
+```
+
+The generated schema records the enum's name beside the field, but the value you write and the
+value you read back are the plain member name. Equality, `not`, `in`, `orderBy` and `groupBy` all
+work on it and are compared against Prisma by the differential suite.
+
+One consequence worth stating, because it is the one thing that surprises: **`orderBy` on an enum
+sorts differently on the two dialects.** Postgres creates a real enum type and sorts by
+*declaration* order; SQLite stores the member as `TEXT` and sorts *alphabetically*. For
+`enum OrganizationPlan { free pro enterprise }`, measured:
+
+```
+postgres   free, pro, enterprise      (declaration order)
+sqlite     enterprise, free, pro      (alphabetical)
+```
+
+gemi matches Prisma on each dialect rather than inventing a uniformity Prisma does not have — the
+same call as `contains` being case-sensitive on Postgres and not on SQLite. If you need one order
+everywhere, sort in the application or store an explicit rank column.
+
 ## `Json` columns
 
 A `Json` column round-trips whatever you give it: an object, an array, a string, `null` — and a
--- a/templates/saas-starter/app/models/differential.test.ts
+++ b/templates/saas-starter/app/models/differential.test.ts
@@ -45,10 +45,16 @@ async function seed(prisma: PrismaClient) {
 
   // Two organizations so a to-one include has something to discriminate, and
   // so `organization.users` is a to-many with more than one member.
+  //
+  // Their `plan` values differ, and neither is the column's default. An enum
+  // seeded to one value compares equal under every ordering and every grouping,
+  // so the cases that exercise it would pass without discriminating anything —
+  // and `free` on both would additionally never distinguish the stored value
+  // from the default the migration writes.
   await prisma.organization.createMany({
     data: [
-      { publicId: "o1", name: "Acme", description: "the first one" },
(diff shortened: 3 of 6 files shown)


nstfkc:
## Review — the ordering table is right, and three of the eight cases run against a value the seed guarantees is absent

Checked the two factual claims rather than reading them.

**The ordering measurement holds.** Independently, on Postgres 16 with `CREATE TYPE p AS ENUM ('free','pro','enterprise')`, and on a plain text sort for the SQLite side:

```
postgres   free, pro, enterprise
sqlite     enterprise, free, pro
```

Exactly the table in the docs. Given you note you first wrote the opposite from memory, worth confirming the corrected version is the correct one.

**The #208 claim holds too** — no template test issues a `CREATE TABLE "Organization"`, and the guard keys on raw DDL, so adding a column cannot wake it. (`registration.test.ts` does `register("Organization", …)`, but that is a model registration, not a fixture table, and the guard does not look at those.)

### The seed's anti-vacuity property is what makes three cases inert

The seed gives `o1` `pro` and `o2` `enterprise`, and the ninth test asserts `expect(plans).not.toContain("free")`. Three of the eight cases filter on `free`. Against those two rows:

```
everything            (no where)         -> 2 rows  [Acme, Globex]
equality   plan = free                   -> 0 rows  []
not        plan <> free                  -> 2 rows  [Acme, Globex]
not-as-tautology (TRUE)                  -> 2 rows  [Acme, Globex]
in         (free, pro)                   -> 1 row   [Acme]
"no match" plan = enterprise             -> 1 row   [Globex]
```

**The `not` case cannot fail.** `<> 'free'` and an always-true predicate return the same two rows, so it is byte-identical to the `everything` case one line above it. `compileNot` emitting a tautology, dropping the predicate, or ignoring the operand would all pass. That is the one shape a negation test exists to catch, and this corpus cannot see it. `not: "pro"` returns one row and discriminates.

**The other two are label-swapped rather than missing.** The case named `equality` matches nothing, so it is the no-match test; the case named `no match` matches Globex, so it is the only equality-against-a-present-value test. Coverage of the two behaviours survives — but an auditor reading the labels to check for vacuity, which is the exact review this PR invites, is told the opposite of what runs.

The part worth drawing out: the guard and the corpus pull against each other. `not.toContain("free")` is the right assertion, and it is precisely what guarantees the `free`-filtered cases compare empty-to-empty or all-to-all. #124's corpus went vacuous because every row shared a value; this one goes vacuous because half the cases name the one value no row has. The ninth test cannot catch it, because it inspects the seed rather than the case results — the same distinction as asserting a fixture exists versus asserting a query discriminated.

A per-case assertion is the general fix and probably more than this needs. Pointing `not` at `pro`, and either swapping the two labels or moving `equality` to `pro` and keeping `free` as the deliberate no-match, costs three string edits and makes the labels true.

### Smaller

The docs say equality, `not`, `in`, `orderBy` and `groupBy` "are compared against Prisma by the differential suite". Literally true, and after the above it is worth the sentence being true in the sense a reader takes it.

`in: ["free", "pro"]` still discriminates — one row of two — so it is fine as it stands; `free` is just dead weight in the list.

Nothing here argues against the change. The column, the default-so-no-backfill choice, the migration's sqlite/`db push` split, and the decision to match Prisma per dialect rather than impose a uniform order all look right, and the ordering difference is a genuinely useful thing to have written down.

nstfkc:
### Follow-up: the `not` result reproduces on a real Postgres enum, and enums need no cast

Two things I checked after posting, both on Postgres 16 with an actual `CREATE TYPE ... AS ENUM`, since the first note's row counts came from a SQLite stand-in and the interesting dialect here is the other one.

**The `not` case returns both rows on Postgres too.** Sending the parameters the way a driver does — unspecified type, not `PREPARE q(text)`, which over-constrains it and refuses everything:

```
in    -> Acme       plan = any ($1)    '{free,pro}'
notIn -> Globex     plan <> all ($1)   '{free,pro}'
eq    -> Acme       plan = $1          'pro'
not   -> Acme       plan <> $1         'free'
not   -> Globex
```

`<> 'free'` returning the full table is not an artefact of my stand-in; it is what the shipped predicate does against the shipped seed, on the dialect where the enum is a real type.

**And the question #216 makes you ask has a good answer: no.** An enum does not need the treatment `Json` just got. `arrayLiteral` builds `{free,pro}` and hands it over as one parameter, so `= any ($1)` has to coerce a text array literal to `OrganizationPlan[]`, and `= $1` has to coerce a string to the enum — both by inference from the column, with no cast at the placeholder. Postgres does both, as above. That is worth knowing because it is exactly where `Json` failed, and the reason it does not fail here is that an enum has one unambiguous input syntax where `jsonb` had two.

The third row above is also `notIn`, which the eight cases do not include. It emits `<> all (...)`, a different operator from `not`'s `<>`, and it discriminates on this seed where `not` does not — so if you are already editing that list, it is a one-line addition that covers a real second code path rather than another spelling of the first.

PR #221, opened 2026-07-29
Assert that select and include narrow the result type

### LINKED ISSUE #220: Nothing asserts that select and include narrow the result type
The plain-object default exists to make one claim. `docs/orm-rows-and-entities.md` leads with it — the level table's first row is

| Level | What you get | What it costs |
| --- | --- | --- |
| POJOs *(default)* | Plain objects, **full `select` narrowing** | Nothing |

and gives it as the reason that level is the default:

> a narrowed row is a narrowed *type*, so there is nothing on it that could read a field the query did not fetch.

Nothing asserts it.

- `wrap.test-d.ts` constructs narrowed rows in order to **reject** them — it depends on narrowing without checking it.
- `aggregate.test-d.ts` covers the aggregate payloads.
- Neither claims that a narrowed *query* produces a narrowed *result*.

And until #201 the type tests ran nowhere at all, so the headline guarantee was unverified twice over.

### What is missing

Concretely, none of these is pinned:

```ts
const [user] = await UserModel.findMany({ select: { id: true } })
user.name                    // should be a compile error

const [u] = await UserModel.findMany({ include: { accounts: true } })
u.organization               // should be a compile error, not included

const [o] = await OrganizationModel.findMany({
  include: { users: { select: { email: true } } },
})
o.users[0].id                // should be a compile error, child narrowed
```

Only a type test can assert them: at runtime the object simply lacks the key and reading it is `undefined` rather than an error. The compile error *is* the feature.

### One thing worth knowing before writing them

`select` and `include` **together** are not a compile error, and it is tempting to assert that they are — the runtime refuses the pair and `docs/orm.md` documents the refusal.

Prisma does not reject it either. Checked against `Prisma.UserFindManyArgs` directly: a `@ts-expect-error` on the pair goes unused there too. Since the contract is "Prisma's argument types verbatim", tightening it in gemi would make valid Prisma code a compile error here.

That is also why `resolveSelection`'s runtime check is not redundant with the type system — the types cannot catch it, on either library. Worth asserting as parity rather than leaving implicit.

nstfkc:
Closes #220.

The plain-object default exists to make one claim, and nothing asserted it. `docs/orm-rows-and-entities.md` leads with it — "POJOs *(default)* | Plain objects, **full `select` narrowing** | Nothing" — and gives it as the reason that level is the default: *"a narrowed row is a narrowed type, so there is nothing on it that could read a field the query did not fetch."*

`wrap.test-d.ts` constructs narrowed rows in order to **reject** them and `aggregate.test-d.ts` covers the aggregate payloads, but no test claimed that a narrowed *query* has a narrowed *result* — the property both of those depend on. Until #201 the type tests ran nowhere at all, so it was unverified twice over.

Twelve assertions across `select`, `include`, a nested `select` inside an `include`, the default selection, and `findFirst`'s nullability as an axis separate from narrowing. Only a type test can make them: at runtime the object simply lacks the key and reading it is `undefined` rather than an error, so the compile error *is* the feature.

## One assertion I got wrong, and the correction is the interesting part

I asserted that `select` and `include` together are a compile error — the runtime refuses the pair and the docs document the refusal, so it seemed safe.

**It is not, and neither is it in Prisma.** Checked rather than assumed, by putting the same `@ts-expect-error` on `Prisma.UserFindManyArgs` and watching that directive go unused too:

```
FAIL  zz-prismacheck.test-d.ts > does Prisma's own args type reject select + include?
```

So gemi matches Prisma exactly, which is the contract — "Prisma's argument types verbatim" — and tightening it would make valid Prisma code a compile error here. The test now pins the **parity** instead, which also records why `resolveSelection`'s runtime check is not redundant with the type system: the types cannot catch it, on either library.

That check is the one in `select.ts` already documented as "the floor under a fifth caller". This gives it a second reason to exist.

## Load-bearing, verified

Removing a `select` so the field is present turns its directive into `Unused '@ts-expect-error' directive` — the suite discriminates rather than passing on anything.

## Checks

- `bun run test:types` — 3 files, **24 passed**, no type errors
- template runtime suite — 17 passed / 3 skipped, 620 tests
- `bun --bun vitest run orm database` — 54 files, 1449 passed

Adds one file and touches nothing else, so it is independent of #204 and #219.

### DIFF
--- a/templates/saas-starter/app/models/select.test-d.ts
+++ b/templates/saas-starter/app/models/select.test-d.ts
@@ -1,0 +1,126 @@
+import { expectTypeOf, test, describe } from "vitest";
+
+import { OrganizationModel, UserModel } from "./generated";
+
+/**
+ * **`select` and `include` narrow the result type**, which is the claim the
+ * plain-object default exists to make good on.
+ *
+ * `docs/orm-rows-and-entities.md` leads with it — the level table's first row is
+ * "POJOs *(default)* | Plain objects, full `select` narrowing | Nothing" — and
+ * gives it as the reason that level is the default: "a narrowed row is a
+ * narrowed *type*, so there is nothing on it that could read a field the query
+ * did not fetch."
+ *
+ * Nothing asserted it. `wrap.test-d.ts` constructs narrowed rows in order to
+ * reject them, and `aggregate.test-d.ts` covers the aggregate payloads, but no
+ * test claimed that a narrowed *query* has a narrowed *result* — the property
+ * both of those depend on. And until #201 the type tests ran nowhere at all, so
+ * the headline guarantee was unverified twice over.
+ *
+ * This can only be a type test. Every assertion below is about code that never
+ * runs: at runtime the object simply lacks the key, and reading it is
+ * `undefined` rather than an error. The compile error *is* the feature.
+ */
+describe("select narrows the result type", () => {
+  test("a selected column is present and correctly typed", async () => {
+    const [user] = await UserModel.findMany({ select: { id: true, email: true } });
+
+    expectTypeOf(user.id).toEqualTypeOf<number>();
+    // Nullability survives the narrowing rather than being widened away.
+    expectTypeOf(user.email).toEqualTypeOf<string | null>();
+  });
+
+  test("a column the query did not fetch is a compile error", async () => {
+    const [user] = await UserModel.findMany({ select: { id: true } });
+
+    // @ts-expect-error `name` was not selected, so it is not on the row
+    user.name;
+  });
+
+  test("selecting nothing but a relation still narrows the scalars away", async () => {
+    const [user] = await UserModel.findMany({
+      select: { accounts: { select: { id: true } } },
+    });
+
+    expectTypeOf(user.accounts[0].id).toEqualTypeOf<number>();
+
+    // @ts-expect-error the parent's own columns were not selected
+    user.id;
+    // @ts-expect-error and neither were the child's, beyond the one named
+    user.accounts[0].provider;
+  });
+
+  test("the default selection carries every scalar", async () => {
+    const [user] = await UserModel.findMany();
+
+    expectTypeOf(user.id).toEqualTypeOf<number>();
+    expectTypeOf(user.email).toEqualTypeOf<string | null>();
+  });
+
+  /**
+   * The narrowing has to survive the operation's own nullability, which is a
+   * separate axis: `findFirst` returns `T | null` whatever the selection is.
+   */
+  test("findFirst is nullable, and still narrowed", async () => {
+    const user = await UserModel.findFirst({ select: { id: true } });
+
+    expectTypeOf(user).toEqualTypeOf<{ id: number } | null>();
+  });
+});
+
+describe("include adds the relation to the type", () => {
+  test("an included to-many is an array of full rows", async () => {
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — four of the five narrowing directives are load-bearing; the fifth names a field that is not on the model

Ran the suite and mutated it rather than reading it. `bun run test:types` reproduces exactly: 3 files, 24 passed, no type errors.

**The load-bearing check, done for each directive rather than one.** Widening every query so the guarded field *is* present turns four of them into `Unused '@ts-expect-error' directive` — lines 37 (`user.name`), 48 (`user.id`), 85 (`user.organization`) and 96 (`org.users[0].id`). Those four discriminate, and the `findFirst` nullability case is a positive assertion that would fail on its own. That is the property this file needs most and it largely holds.

### `user.accounts[0].provider` cannot fail

`provider` is a field of **`SocialAccount`**, not of `Account`. `Account` is `id`, `publicId`, `organization`, `organizationId`, `organizationRole`, `userId`, `user`, `createdAt`, `updatedAt`, `deletedAt`, `settings`, `token`. So the directive fires because the property does not exist on the child type *at all*, not because the child was narrowed — and it fires equally when nothing is narrowed.

Demonstrated by removing the narrowing outright:

```
select: { accounts: { select: { id: true } } }   ->   24 passed
select: { accounts: true }                       ->   24 passed     <- child no longer narrowed, nothing notices
```

And the same mutation with a field `Account` actually has:

```
user.accounts[0].publicId   narrowed query      ->   24 passed          directive used
user.accounts[0].publicId   narrowing removed   ->   Unused '@ts-expect-error' directive at :50
```

So it is a one-word fix — any real `Account` column in place of `provider` — and the test then asserts what its own comment says it does.

Worth saying what this does and does not leave uncovered: `org.users[0].id` at line 96 *is* load-bearing, so a narrowed child under an `include` is pinned. What is not pinned is the child under a **`select`**, which is different machinery in the generated types and is the case this test exists for. The sibling assertion in the same test — `user.id` at 48, the parent's scalars going away — is load-bearing and unaffected.

### The parity test's comment describes the opposite of what the types do

```ts
// The result type follows `select`, which is the shape Prisma describes too.
expectTypeOf(rows).toBeArray();
```

It follows `include`. Measured on that exact query:

```ts
expectTypeOf(rows[0].id).toEqualTypeOf<number>();               // passes
expectTypeOf(rows[0].name).toEqualTypeOf<string | null>();      // passes — excluded by `select`
expectTypeOf(rows[0].accounts[0].publicId).toEqualTypeOf<string>();  // passes — the included relation, in full
```

All three pass, so the row carries a scalar `select` excluded and the relation `include` added — the full `include` shape, not the `select` one. `toBeArray()` pins none of it, so nothing catches the description drifting from the type.

I would keep the test and fix the sentence, because the reasoning around it is right and is the most valuable part of the PR: the pair type-checks in both libraries, gemi's contract is Prisma's args verbatim, and that is precisely why `resolveSelection`'s runtime refusal is not redundant. Since the stated purpose is that "a later change that tightens the generated types has to notice it is diverging", one of those three `expectTypeOf` lines is worth keeping alongside — it makes the assertion about the shape rather than about arity, which is what would actually notice.

Nothing here argues against the change. The gap it closes is real, the choice to verify the Prisma side rather than assume it is the right instinct, and four of five directives do the job.

PR #223, opened 2026-07-29
Prisma's Json null sentinels were stored as the object {}

### LINKED ISSUE #222: Prisma's Json null sentinels are stored as the object {}
A nullable `Json` column has **two** legal empty values, and Prisma makes the caller choose:

```
Prisma.DbNull     the column is SQL NULL
Prisma.JsonNull   the column holds the JSON value null
```

A bare `null` is a type error on a Json field — on Prisma *and* on gemi, which takes Prisma's argument types verbatim (verified with a `@ts-expect-error` on both). So the sentinels are what type-checks, and they are the input Prisma's own types point you at.

Nothing translates them. Both are ordinary objects with no enumerable properties, so every path that serialises a Json value turns them into the jsonb object `{}`. Measured against Postgres 16:

| written | `metadata is null` | `jsonb_typeof` | value |
| --- | --- | --- | --- |
| Prisma `DbNull` | true | null | — |
| Prisma `JsonNull` | false | null | `null` |
| **gemi `DbNull`** | **false** | **object** | **`{}`** |
| **gemi `JsonNull`** | **false** | **object** | **`{}`** |

Two different wrong rows, both plausible-looking, neither raising. Same on SQLite, where Json is serialised to text.

This is the silent mis-store the bare-scalar refusal existed to prevent, reached through the documented API. `differential.test.ts` already carries the warning about the *read* half of this distinction, in a seed comment:

> A JSON null, which is not the same as a SQL NULL — Prisma spells the difference `Prisma.JsonNull` vs `Prisma.DbNull`, and a decoder that conflates them returns the wrong one of two legal answers.

The write half had no such care and no coverage: the write differential only ever wrote an *object* to a Json column.

### Fix

Map the sentinels in both write paths — the Postgres cast from #216 and SQLite's `encode`.

Two constraints worth knowing before writing it:

- **Recognise by `toString`, not `instanceof`.** The ORM runtime may not import the Prisma client package, and `runtime-isolation.test.ts` enforces that against *comments* as well as imports. Prisma implements `toString` on these deliberately, returning `Prisma.DbNull` / `Prisma.JsonNull`; a constructor name is the other candidate and is what a minifier may rewrite.
- **Do not map `AnyNull`.** It means "either of the two" and is legal only in a filter. Prisma rejects it in a write, so mapping it would quietly accept something Prisma does not.

Coverage belongs in the write differential, which compares table contents — conflating the two again then fails there rather than passing quietly.

nstfkc:
Closes #222.

A nullable `Json` column has two legal empty values and Prisma makes the caller choose between them: `Prisma.DbNull` leaves the column SQL NULL, `Prisma.JsonNull` stores the JSON value `null`. A bare `null` is a type error on both libraries — checked, not assumed — so the sentinels are what type-checks here, and they are the input Prisma's own types point you at.

Nothing translated them. Both are ordinary objects with no enumerable properties, so every path that serialises a Json value turned them into the jsonb object `{}`:

```
PRISMA DbNull    -> sqlnull=true  typeof=null
PRISMA JsonNull  -> sqlnull=false typeof=null
GEMI   DbNull    -> sqlnull=false typeof=object   value={}      <- wrong
GEMI   JsonNull  -> sqlnull=false typeof=object   value={}      <- wrong
```

Two different wrong rows, both plausible-looking, neither raising — the silent mis-store the bare-scalar refusal existed to prevent, reached through the documented API. After:

```
GEMI   DbNull    -> sqlnull=true  typeof=null     matches Prisma
GEMI   JsonNull  -> sqlnull=false typeof=null     matches Prisma
```

## How it was found

Auditing my own #216. I had added a `null` branch to `fieldParam` and wanted to know whether it was covered — it was not, and following that thread showed the branch is a *floor* (a bare `null` does not type-check), which then raised the question of what the values that *do* type-check actually do.

## Two constraints the implementation had to meet

- **Recognised by `toString`, not `instanceof`.** The ORM runtime may not import the Prisma client package, and `runtime-isolation.test.ts` enforces that against **comments** as well as imports — which this learned the hard way when a doc comment naming the package failed the suite. Prisma implements `toString` on the sentinels deliberately; a constructor name is what a minifier may rewrite.
- **`AnyNull` is deliberately not mapped.** It means "either of the two" and is legal only in a filter. Prisma rejects it in a write, so mapping it would quietly accept something Prisma does not.

Both write paths are covered: the Postgres cast from #216, and SQLite's `encode`.

## Coverage, and a stale note it replaced

The cases go in the **write differential**, which compares table contents, so conflating the sentinels again fails there rather than passing quietly.

Two more went in alongside them — the bare JSON number and boolean. That table excluded them with a note saying gemi raised where Prisma stored; #216 lifted that, and the note pointed at a `json-scalars.test.ts` **that has never existed** (the boundary was pinned in `writes.coercion.test.ts`). Both the exclusion and its justification are gone.

## Checks

- Postgres differential — 13 files, **653 passed**, 0 failed
- SQLite suite — 17 passed / 3 skipped, **624 tests**
- `bun --bun vitest run orm database` — 54 files, **1450 passed**
- the ORM runtime still contains zero references to the Prisma client package

Scratch container removed; template restored to SQLite. Independent of #204, #219 and #221.

### DIFF
--- a/packages/gemi/orm/compile/cast.ts
+++ b/packages/gemi/orm/compile/cast.ts
@@ -1,4 +1,5 @@
 import type { SqlDialect } from "../dialect";
+import { jsonNullKind } from "../json-null";
 import type { FieldSchema } from "../schema";
 import { type Binder, type Fragment, param } from "./fragment";
 
@@ -42,6 +43,14 @@ export function fieldParam(
     // `JSON.stringify(null)` is `"null"`, and `'null'::jsonb` is a jsonb null,
     // which is a different thing from an absent value and would diverge from
     // Prisma on every nullable Json column.
-    return value === null || value === undefined ? null : JSON.stringify(value);
+    if (value === null || value === undefined) return null;
+
+    // Prisma's sentinels, which serialise to `{}` if left to `JSON.stringify`.
+    // `'null'::text::jsonb` is the JSON value; a bound `null` is SQL NULL.
+    const kind = jsonNullKind(value);
+    if (kind === "db") return null;
+    if (kind === "json") return "null";
+
+    return JSON.stringify(value);
   }, cast);
 }
--- a/packages/gemi/orm/dialect/sqlite.ts
+++ b/packages/gemi/orm/dialect/sqlite.ts
@@ -8,6 +8,7 @@ import {
   sql,
 } from "../compile/fragment";
 import { DecodeError } from "../errors";
+import { jsonNullKind } from "../json-null";
 import type { FieldSchema } from "../schema";
 import type { ConstraintViolation, SqlDialect } from "./index";
 
@@ -235,6 +236,15 @@ export class SqliteDialect implements SqlDialect {
         // `JSON.stringify` on every value is unambiguous: a string becomes
         // `"42"` with its quotes, which is what `JSON.parse` needs to hand back
         // a string. The mirror of `decode`, which parses unconditionally.
+        //
+        // Except Prisma's two null sentinels, which have no enumerable
+        // properties and would serialise to `{}` — see `json-null.ts`.
+        // `DbNull` is the column being NULL; `JsonNull` is the JSON value.
+        {
+          const kind = jsonNullKind(value);
+          if (kind === "db") return null;
+          if (kind === "json") return "null";
+        }
         return JSON.stringify(value);
       default:
         // BigInt passes through. Bun's driver truncates integers above 2^53 on
--- a/packages/gemi/orm/json-null.ts
+++ b/packages/gemi/orm/json-null.ts
@@ -1,0 +1,54 @@
+/**
+ * Prisma's two null sentinels for a `Json` column, recognised without importing
+ * Prisma.
+ *
+ * A nullable `Json` column has *two* legal empty values and Prisma makes the
+ * caller choose between them, because they are different rows:
+ *
+ *   Prisma.DbNull     the column is SQL NULL
+ *   Prisma.JsonNull   the column holds the JSON value `null`
+ *
+ * A bare `null` is a type error on both libraries — gemi takes Prisma's
+ * argument types verbatim, so the sentinels are what type-checks here too, and
+ * `differential.test.ts` already notes that "a decoder that conflates them
+ * returns the wrong one of two legal answers".
+ *
+ * Nothing translated them. Both are ordinary objects with no enumerable
+ * properties, so every path that serialises a Json value turned them into the
+ * jsonb object `{}` — a plausible-looking value, silently wrong, on both
+ * dialects and with nothing raised. That is the same silent mis-store the
+ * bare-scalar refusal existed to prevent.
+ *
+ * **Recognised by `toString`, not by `instanceof` or the constructor's name.**
+ * The ORM runtime may not import the Prisma client package at all —
+ * `runtime-isolation.test.ts` greps for the name, comments included — so
+ * `instanceof` is unavailable. Prisma implements
+ * `toString` on these deliberately, returning `Prisma.DbNull` and
+ * `Prisma.JsonNull`; a constructor name would be the other candidate and is the
+ * one a minifier is free to rewrite.
+ */
+export type JsonNullKind = "db" | "json";
+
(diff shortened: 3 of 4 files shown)


nstfkc:
## Review — the fix is right; the cases pin "not `{}`" but not which of the two nulls

Verified the detection mechanism against the installed client first, since everything rests on it:

```
DbNull   | String(): "Prisma.DbNull"   | JSON.stringify: {}
JsonNull | String(): "Prisma.JsonNull" | JSON.stringify: {}
AnyNull  | String(): "Prisma.AnyNull"  | JSON.stringify: {}
```

`toString` is real and stable, the `{}` mis-store is exactly as described, and leaving `AnyNull` out of the map is right — it has a tag too, so its absence is a decision the map records rather than an omission.

### The new cases cannot tell the two sentinels apart

> The cases go in the **write differential**, which compares table contents, so conflating the sentinels again fails there rather than passing quietly.

They do not. `expectSameWrite` compares `readTables(prisma, …)` — a read back **through Prisma** — and Prisma returns plain `null` for both forms. Measured, three rows written by raw SQL and read through `p.user.findMany`:

```
metadata = NULL     ->  value=null   isDbNull=false  isJsonNull=false
metadata = 'null'   ->  value=null   isDbNull=false  isJsonNull=false
metadata = '{}'     ->  value={}
```

The `{}` row survives the read, which is why the cases *do* catch the original defect — disabling `jsonNullKind` entirely takes the file from 7 failures to 9, both of them these. That much is load-bearing.

But the two `null` rows are indistinguishable after the read, so swapping either mapping is invisible. Whole file, SQLite:

```
no mapping at all (pre-PR)        9 failed   <- both new cases fail, correctly
"Prisma.JsonNull": "db"           7 failed   <- JsonNull stored as SQL NULL, nothing notices
"Prisma.DbNull":   "json"         7 failed   <- DbNull stored as JSON null, nothing notices
unmutated                         7 failed   <- baseline (the 7 are my stale client, not the PR)
```

So the property under test — *these are different rows* — is the one property the harness cannot see. An implementation that mapped both sentinels to the same thing would ship green, which is the shape of the bug being fixed rather than a distant hypothetical: it is what the code did before, one step less wrong.

This is the same collapse that made the `null` branch in `fieldParam` unpinnable, and the project already has the answer to it. `writes.coercion.test.ts` asserts `jsonb_typeof` precisely because a right-looking value under the wrong type reads back correctly through the ORM; the equivalent here is one raw read per case — `typeof("metadata")` on SQLite, `jsonb_typeof` on Postgres — asserting `null` vs `text`/`null`-the-jsonb-value. The differential comparison stays as it is; the raw assertion is what makes it discriminate.

(Measured on SQLite. The Postgres half runs the same `readTables`-through-Prisma path so I expect it identical, but I could not run it here — my client is generated for the pre-#219 schema, which is also what the 7 baseline failures are.)

### Smaller: the guard's comment describes something that is not happening

```ts
// Guarded because a plain object from `JSON.parse` has `Object.prototype`'s
// `toString`, and a caller's object may have thrown one of its own on.
const tag = Object.prototype.hasOwnProperty.call(value.constructor ?? {}, "name")
  ? String(value) : "";
```

Every function has an own `name`, so a plain object's constructor — `Object` — passes the guard. Plain objects reach `String(value)`, become `"[object Object]"`, and are rejected by the `SENTINELS` lookup, not by the guard.

What the guard *does* catch is a null-prototype object, where `value.constructor` is `undefined`:

```
plain {}             guard passes: true    String() -> "[object Object]"
Object.create(null)  guard passes: false   String() -> TypeError
```

So it is load-bearing — without it, `jsonNullKind(Object.create(null))` throws instead of returning `null` — for a reason the comment does not give. Worth swapping the comment for that one, since the next reader who works out that `Object` has an own `name` will otherwise conclude the guard is dead and delete it.

### Otherwise

The `where`-path consequence is already filed as #224 and scoped correctly there, including that it is not a regression from this PR — no need to hold this for it. Removing the stale exclusion note is a good catch, and the observation that it pointed at a `json-scalars.test.ts` which never existed is the kind of thing worth stating out loud.

PR #225, opened 2026-07-29
Filtering by Prisma.DbNull matched nothing

### LINKED ISSUE #224: Filtering by Prisma.DbNull silently matches nothing
Adjacent to #222/#223, which fixed the sentinels on the **write** path. The **filter** path is still wrong, and differently.

`Prisma.DbNull` and `Prisma.JsonNull` are legal in a `where` on a Json column — on Prisma *and* on gemi, which takes Prisma's argument types verbatim. Verified with `@ts-expect-error` on both:

```ts
await UserModel.findMany({ where: { metadata: { equals: Prisma.DbNull } } })   // type-checks
await UserModel.findMany({ where: { metadata: { equals: Prisma.JsonNull } } }) // type-checks
```

The compiler emits an equality comparison for them, where a bare `null` correctly becomes `IS NULL`:

```
equals: null       ->  where "meta" is null      correct
equals: DbNull     ->  where "meta" = ?          wrong
equals: object     ->  where "meta" = ?          correct
```

Since #223 binds `DbNull` as SQL NULL, `"meta" = NULL` is **never true** — the query returns zero rows whatever the table holds, with nothing raised. Prisma emits `IS NULL` and returns the rows whose column is SQL NULL.

Before #223 it bound the sentinel as the object `{}`, so it matched only rows holding `{}` — wrong in a different way. The write fix changed the flavour of the wrongness without addressing it, which is worth stating plainly: this is not a regression from #223, and #223 did not fix it either.

`JsonNull` is likely correct already — it binds `'null'`, and `= 'null'::jsonb` is the right comparison for a stored JSON null — but that should be asserted rather than assumed.

### Fix

`where.ts` has to treat `DbNull` the way it already treats a bare `null`: emit `IS NULL` rather than an equality against a bound parameter. `jsonNullKind` from #223 already recognises the sentinels without importing the Prisma client package, so the detection exists; only the null-comparison branch needs to consult it.

### Verification

Belongs in the read differential, which compares result sets — a filter that returns nothing is exactly what a comparison against Prisma catches and what a same-shape assertion does not. Seed rows with all three states (SQL NULL, JSON null, an object) so each filter discriminates; a seed with only one of them would let `IS NULL` and `= NULL` agree.

### How it was found

Auditing my own scoping decision in #223. I had deliberately not mapped `Prisma.AnyNull` on the grounds that it is filter-only — then checked whether the *other two* reach the filter path, which they do, through `fieldParam`.

nstfkc:
Closes #224. **Stacked on #223** — it needs `jsonNullKind` from that branch, so this targets it and the diff is the one-line fix plus its coverage. Retargets cleanly to `feat/orm` once #223 lands.

`equals` already carried the reasoning, for the case it did handle:

> `= ?` with a null parameter matches nothing in SQL, where Prisma means `is null`. This is the difference that would silently return wrong rows.

`Prisma.DbNull` means the same thing on a `Json` column and does not arrive as `null` — it is a sentinel object — so it fell through to the parameter path:

```
equals: null       ->  where "meta" is null      correct
equals: DbNull     ->  where "meta" = ?          wrong
equals: DbNull     ->  where "meta" is null      after
```

Since #223 binds the sentinel as SQL NULL, `= NULL` matched **nothing whatever the table held**, with nothing raised.

**Not a regression from #223.** Before it, the sentinel bound as the object `{}` and matched only rows holding `{}` — wrong differently. The write fix changed the flavour of the wrongness without addressing it, and it is worth being precise about that.

`JsonNull` deliberately keeps the parameter path: it is a genuine *value* comparison — the column holds the JSON `null`, and `= 'null'` is how you find it — so the two sentinels take different branches on purpose.

## The coverage detail that matters

The cases run against `Account.settings` rather than a new fixture, because the seed already carries **both** states there: `ac2` holds a JSON null, the other accounts leave the column SQL NULL.

That is load-bearing, not convenient. With only one of the two states present, `is null` and `= NULL` return the same empty set and the bug passes. My first attempt seeded a fresh row instead and broke two existing count assertions — which is itself the argument for using the states already in the seed.

Compared against Prisma rather than by row count, so a filter matching nothing fails here rather than agreeing with an equally empty expectation.

**Verified by reverting the one-line fix** — `DbNull finds the SQL NULLs` is the only test that fails.

## Checks

- Postgres differential — 13 files, **664 passed**, 0 failed
- SQLite suite — 17 passed / 3 skipped, **635 tests**
- `bun --bun vitest run orm database` — 54 files, **1450 passed**

Scratch container removed; template restored to SQLite.

### DIFF
--- a/packages/gemi/orm/compile/where.ts
+++ b/packages/gemi/orm/compile/where.ts
@@ -18,6 +18,7 @@ import {
   joinFragments,
   sql,
 } from "./fragment";
+import { jsonNullKind } from "../json-null";
 import { fieldParam } from "./cast";
 
 /**
@@ -820,7 +821,15 @@ function equals(
 ): Fragment {
   // `= ?` with a null parameter matches nothing in SQL, where Prisma means
   // `is null`. This is the difference that would silently return wrong rows.
-  if (operand === null) return sql(`${column} is null`);
+  //
+  // `Prisma.DbNull` means the same thing on a `Json` column and arrives as a
+  // sentinel object rather than as `null`, so it reached the parameter path
+  // below and compiled to `= ?` — matching nothing, whatever the table held.
+  // Its sibling `JsonNull` is a genuine *value* comparison and belongs there:
+  // the column holds the JSON `null`, and `= 'null'` is how you find it.
+  if (operand === null || jsonNullKind(operand) === "db") {
+    return sql(`${column} is null`);
+  }
   return concat(
     sql(`${column} = `),
     fieldParam(field, dialect, encoded(field, dialect, locate)),
--- a/packages/gemi/orm/compile/where.ts
+++ b/packages/gemi/orm/compile/where.ts
@@ -674,6 +674,29 @@ function compileFieldFilter(
   const { dialect } = context;
   const column = `${context.qualifier ?? ""}${dialect.quoteIdent(field.column)}`;
 
+  // A bare `Prisma.DbNull` is **not** shorthand for `equals`, and Prisma refuses
+  // it: `where: { settings: Prisma.DbNull }` is an invalid invocation there,
+  // spelled out under the offending key. Both sentinels are objects with no
+  // enumerable properties, so left alone they read as an operator map holding
+  // no operators — which compiles to `true`, and answered "every row" to a
+  // question about the null ones.
+  //
+  // Refused rather than interpreted, because guessing which of the two nulls a
+  // caller meant is the whole thing the sentinels exist to stop, and because
+  // matching Prisma means matching its refusals as well as its results.
+  const sentinel = jsonNullKind(value);
+  if (sentinel) {
+    throw new InvalidArgumentError(
+      `where.${field.name}`,
+      schema.name,
+      context.operation,
+      `Prisma.${sentinel === "db" ? "DbNull" : "JsonNull"} is not a filter on ` +
+        `its own — Prisma rejects it here too. Write it as an explicit ` +
+        `comparison: { ${field.name}: { equals: Prisma.` +
+        `${sentinel === "db" ? "DbNull" : "JsonNull"} } }.`,
+    );
+  }
+
   // A bare value is shorthand for `equals`. A `Date` is a value, not a filter
   // object, even though `typeof` cannot tell them apart.
   if (!isFilterObject(value)) {
@@ -795,7 +818,23 @@ function compileNot(
   locate: (args: any) => any,
   column: string,
 ): Fragment {
-  if (operand === null) return sql(`${column} is not null`);
+  // `DbNull` is the sentinel spelling of the same question, so it takes the
+  // same branch — see `equals`.
+  if (operand === null || jsonNullKind(operand) === "db") {
+    return sql(`${column} is not null`);
+  }
+
+  // `JsonNull` does not: it asks about a stored JSON *value*, so negating it is
+  // a value comparison. Handled here rather than by delegating, because the
+  // delegate refuses a bare sentinel — correctly, since Prisma does — and this
+  // one did not arrive bare. `{ not: Prisma.JsonNull }` is a valid Prisma
+  // filter and has to stay one.
+  if (jsonNullKind(operand) === "json") {
+    return concat(
+      sql("not "),
+      parenthesize(equals(column, field, operand, context.dialect, locate)),
+    );
+  }
 
   if (!isFilterObject(operand)) {
     return concat(
--- a/templates/saas-starter/app/models/differential.test.ts
+++ b/templates/saas-starter/app/models/differential.test.ts
@@ -1270,6 +1270,14 @@ function suite(label: string, url?: string) {
     test.each([
       ["DbNull finds the SQL NULLs", { equals: PrismaNS.DbNull }],
       ["JsonNull finds the JSON nulls", { equals: PrismaNS.JsonNull }],
+      // The shorthand, which is what Prisma's own documentation writes. It read
+      // the sentinel as an operator map with no operators and compiled to
+      // `where true` — the whole table, when asked for the null rows.
+      ["DbNull as the bare shorthand", PrismaNS.DbNull],
+      ["JsonNull as the bare shorthand", PrismaNS.JsonNull],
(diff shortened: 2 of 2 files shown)


nstfkc:
**Extended: the `equals` fix covered one spelling of three.** Checking the others against the compiler:

```
{ meta: { equals: DbNull } }  ->  is null            fixed in the first commit
{ meta: { not: DbNull } }     ->  not (true)         matched nothing
{ meta: DbNull }              ->  where true         matched EVERYTHING
```

The last is the shorthand, and it answered *every row* to a question about the null ones — worse than the bug this PR opened with. Both sentinels are objects with no enumerable properties, so read as an operator map they hold no operators, and an empty operator map compiles to `true`.

## The obvious fix was wrong, and the differential caught it

Treating the sentinel as a **value** in `isFilterObject` — the way a `Date` already is — turns the shorthand into `is null`. That looks right and is not: Prisma **rejects** that invocation outright.

```
AssertionError: prisma threw but gemi returned —
Invalid `prisma.account.findMany()` invocation:
  where: { settings: Prisma.DbNull }
                     ~~~~~~~~~~~~~
```

gemi would have been silently answering a query Prisma refuses. Caught by the two cases added for it, not by review — which is the argument for comparing against Prisma rather than asserting a row count.

So the bare sentinel is now **refused** with an `InvalidArgumentError` naming the explicit form. That matches Prisma's failure as well as its results, and guessing which of the two nulls a caller meant is the whole thing the sentinels exist to stop.

`{ not: DbNull }` takes the null branch; `{ not: JsonNull }` stays a value comparison, handled in `compileNot` rather than delegating — the delegate now refuses a bare sentinel, and this one did not arrive bare. `{ not: Prisma.JsonNull }` is valid Prisma and has to stay valid here.

## Re-verified

- Postgres differential — **668 passed**, 0 failed
- SQLite suite — **639 tests**
- `packages/gemi` — 54 files, **1450 passed**
- typecheck exit 0, `oxlint --deny-warnings` 0 warnings

Six spellings now covered, against a seed carrying both null states.

nstfkc:
## Review — every claim about Prisma checks out; the plan cache does not distinguish the two sentinels

I verified the parity claims against the installed client rather than reading them, because the whole design rests on them. All six shapes behave exactly as the PR says:

```
equals: DbNull     ACCEPTED -> the SQL NULL row
equals: JsonNull   ACCEPTED -> the JSON null row
bare DbNull        PrismaClientValidationError: Expected JsonNullableFilter, provided Enum
bare JsonNull      PrismaClientValidationError: (same)
not: DbNull        ACCEPTED -> everything not SQL NULL
not: JsonNull      ACCEPTED -> excludes the JSON null row *and* the SQL NULL one
```

The refusal is genuine parity, and `not: JsonNull` is the subtle one — `not (col = 'null')` drops SQL NULLs through three-valued logic, and that is what Prisma returns too, so the delegation is right for a reason that could easily have gone the other way.

**The seed claim holds as well**, which I doubted and was wrong about. Running the seed's exact `createMany` through Prisma:

```
ac2  stored='null'  sqlite_typeof=text    <- JSON null
ac4  stored=NULL    sqlite_typeof=null    <- SQL NULL
```

Both states really are present, so neither filter case is vacuous. That is a better position than #223's write cases, which cannot separate the two nulls at all — here the *result set* is observable, so the comparison discriminates.

### The plan key cannot tell the two apart, and this PR is what makes that matter

Both sentinels are objects with no enumerable properties, so `canonicalShape` renders them as `{}` — the same as any other empty object:

```
{ settings: { equals: Prisma.DbNull   } }  ->  {where:{settings:{equals:{}}}}
{ settings: { equals: Prisma.JsonNull } }  ->  {where:{settings:{equals:{}}}}
{ settings: { equals: {}              } }  ->  {where:{settings:{equals:{}}}}
```

One key, three queries — and after this change they no longer compile to the same SQL. `DbNull` becomes `is null` with **no parameter**; the other two stay `= ?`. Whichever runs first mints the plan and the rest reuse it. Against a table holding one row of each kind:

```
warm cache, DbNull first
  equals DbNull     -> sqlnull      correct
  equals JsonNull   -> sqlnull      wrong
  equals {}         -> sqlnull      wrong

warm cache, JsonNull first
  equals JsonNull   -> jsonnull     correct
  equals DbNull     -> (none)       wrong — #224's exact symptom, back again

clearPlanCache() before each
  equals DbNull     -> sqlnull
  equals JsonNull   -> jsonnull
  equals {}         -> obj          all correct
```

The refusal is reachable the same way. `{ settings: {} }` is a filter object with no operators, compiles to `true`, and shares the key `{where:{settings:{}}}` with a bare sentinel:

```
cold cache   bare DbNull  ->  THREW InvalidArgumentError
warm cache   settings: {} ->  sqlnull,jsonnull,obj
             bare DbNull  ->  sqlnull,jsonnull,obj    <- refusal bypassed, every row returned
```

Which is verbatim the failure the refusal's own comment describes: *"answered 'every row' to a question about the null ones."*

**Worth being precise about whose problem this is.** Before this PR all three operands compiled to identical SQL text and differed only in the bound parameter, so sharing one plan was correct — that is exactly what `VALUE_KEYS` is for, and #223 stays safe for the same reason (it changes the *parameter*, never the text). This change makes the emitted text depend on a value the key deliberately discards, which is the one thing a value key may not do. It is the same class as the note already in `planKey` about the relation strategy: *"different strategies emit different SQL for the same arguments … silent wrong results from a caching bug."*

**And the suite cannot see it**, which is why the six new cases pass. `expectSame` opens with `clearPlanCache()`, so every comparison compiles fresh and no two of them ever share an entry. The verification quoted in the PR — reverting the one-line fix and watching `DbNull finds the SQL NULLs` fail — is sound but runs under that same cold cache.

The fix is to make the sentinel part of the shape, since `jsonNullKind` already identifies it without importing Prisma. Something like returning `"DbNull"` / `"JsonNull"` instead of `{}` from `canonicalShape` for those two objects. A test needs two queries in one warm cache, which no existing case does — `plan.discrimination.test.ts` is where the sibling properties live and is the natural home.

I would treat this as blocking, not because the fix is wrong — it is right, and well argued — but because the failure it introduces is silent, order-dependent, and one of its three forms is the very bug this PR closes.

PR #227, opened 2026-07-29
Pin the generated API to the operations the compiler implements

### LINKED ISSUE #226: Nothing checks the generated API against the operations the compiler implements
There are three lists of operations in this repository, and only two of them are reconciled:

| list | where | pinned to |
| --- | --- | --- |
| the compiler | `isReadOperation` and its three siblings | — |
| the documentation | the block in `docs/orm.md` | the compiler, by `docs-scope.test.ts` (#183) |
| **the generator** | `READ_OPERATIONS` / `WRITE_OPERATIONS` in `bin/orm/emit.ts`, hand-written | **nothing** |

The third is the one an application actually calls.

An operation added to the compiler and left out of the emitter leaves `docs/orm.md` correct, the `docs-scope` guard green, and `User.groupBy` simply not existing. Of the three ways these lists can disagree, it is the only one no test can see — and the only one a reader meets as *"the documentation says this exists"*.

### Not hypothetical for this shape of code

#182 found the operation **count** wrong in three places at once; #195 found three more. Both for the same reason: each list is maintained by hand against a different copy of the truth. This is that divergence one level down, where the copies are the API surface rather than a number.

### Fix

Assert that the generated base declares a static for every operation the compiler implements, and nothing beyond them.

Read it from the committed artifact rather than by re-running the generator: the artifact is what the template compiles against, and `generated.test.ts` already guards that regenerating produces a zero-line diff — so the two together cover both directions.

### Two things to get right

- **Match both declaration spellings.** The emitter gives most operations a generic (`static findMany<T extends …>`) and gives `createMany`, `updateMany` and `deleteMany` none. A pattern assuming the generic reports those three as missing when they are not — a guard stricter than the thing it guards fails exactly like a real defect and is diagnosed differently.
- **Exclude `wrap` and the `$`-prefixed internals.** `wrap` is not an operation: it is level 3 in `docs/orm-rows-and-entities.md` and takes a row rather than query arguments.

nstfkc:
Closes #226.

There are three lists of operations here and only two were reconciled:

| list | pinned to |
| --- | --- |
| the compiler — `isReadOperation` and its siblings | — |
| the documentation — the block in `docs/orm.md` | the compiler, by `docs-scope.test.ts` (#183) |
| **the generator** — `READ_OPERATIONS` / `WRITE_OPERATIONS`, hand-written | **nothing** |

The third is the one an application calls. An operation added to the compiler and left out of the emitter leaves `docs/orm.md` correct, the `docs-scope` guard green, and `User.groupBy` simply not existing — the one failure of the three that no test could see, and the only one a reader meets as *"the documentation says this exists"*.

Not hypothetical for this shape of code: **#182** found the operation *count* wrong in three places at once and **#195** found three more, precisely because each list was maintained by hand against a different copy of the truth. This is that divergence one level down, where the copies are the API rather than a number.

Read from the committed artifact rather than by re-running the generator — the artifact is what the template compiles against, and `generated.test.ts` already guards that regenerating produces a zero-line diff, so the two together cover both directions.

**Verified by deleting `groupBy` from the artifact:**

```
× groupBy is declared on the generated base

AssertionError: the compiler implements groupBy, but the generated base
declares no static groupBy — an application cannot call it
```

## The first version was stricter than the thing it guards

It reported `createMany`, `updateMany` and `deleteMany` as missing. They are not: the emitter gives most operations a generic (`static findMany<T extends …>`) and those three none.

Worth recording because a guard that is too strict fails exactly like a real defect and is diagnosed differently — I went looking at the artifact before the test, which is the wrong end.

Two exclusions in the other direction, both deliberate: `$`-prefixed names are internals kept out of an application's way by that prefix, and `wrap` is not an operation at all — it is level 3 in `docs/orm-rows-and-entities.md` and takes a row rather than query arguments.

## Checks

- `bun --bun vitest run orm database` — 55 files, **1481 passed**
- template suite — 17 passed / 3 skipped, 629 tests

Adds one file and touches nothing else, so it is independent of #204, #223 and #225.

### DIFF
--- a/packages/gemi/orm/docs-scope.test.ts
+++ b/packages/gemi/orm/docs-scope.test.ts
@@ -4,8 +4,8 @@ import { describe, expect, test } from "vitest";
 
 import { isAggregateOperation } from "./compile/aggregate";
 import { isGroupByOperation } from "./compile/group-by";
-import { isReadOperation } from "./compile/read";
-import { isWriteOperation } from "./compile/write";
+import { READ_ARGS, isReadOperation } from "./compile/read";
+import { WRITE_ARGS, isWriteOperation } from "./compile/write";
 import * as orm from "./index";
 import type { Operation } from "./plan";
 
@@ -32,27 +32,24 @@ import type { Operation } from "./plan";
 const DOC = join(import.meta.dirname, "../../../docs/orm.md");
 
 /**
- * The operations the ORM implements. Listed here because `Operation` is a type
- * and erased — but every entry is then **verified** against the compiler's own
- * predicates, so this cannot quietly name something that does not exist.
+ * The operations the ORM implements, derived from the compiler's own tables.
+ *
+ * This was a hand-written list, verified entry by entry against the predicates.
+ * That direction is the weaker one: it catches a name that does not exist, and
+ * cannot catch an operation that exists and was never added — which is the
+ * failure a scope guard is for, and the one #182 and #195 kept finding in
+ * hand-written copies elsewhere.
+ *
+ * `op in READ_ARGS` is the whole of `isReadOperation`, so those keys are the
+ * compiler's record of itself. `aggregate` and `groupBy` are named because
+ * their predicates are literal comparisons with no table behind them.
  */
-const IMPLEMENTED: Operation[] = [
-  "findUnique",
-  "findUniqueOrThrow",
-  "findFirst",
-  "findFirstOrThrow",
-  "findMany",
-  "count",
+const IMPLEMENTED = [
+  ...Object.keys(READ_ARGS),
+  ...Object.keys(WRITE_ARGS),
   "aggregate",
   "groupBy",
-  "create",
-  "createMany",
-  "update",
-  "updateMany",
-  "upsert",
-  "delete",
-  "deleteMany",
-];
+] as Operation[];
 
 describe("docs/orm.md and the runtime agree", () => {
   const source = readFileSync(DOC, "utf8");
--- a/packages/gemi/orm/generated-surface.test.ts
+++ b/packages/gemi/orm/generated-surface.test.ts
@@ -1,0 +1,125 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+import { isAggregateOperation } from "./compile/aggregate";
+import { isGroupByOperation } from "./compile/group-by";
+import { isReadOperation } from "./compile/read";
+import { isWriteOperation } from "./compile/write";
+import type { Operation } from "./plan";
+
+/**
+ * **The generated model base exposes every operation the compiler implements.**
+ *
+ * There are three lists of operations in this repository and, until this, only
+ * two of them were reconciled:
+ *
+ *   the compiler        `isReadOperation` and its three siblings
+ *   the documentation   the block in `docs/orm.md`, pinned to the compiler by
+ *                       `docs-scope.test.ts` (#183)
+ *   the generator       `READ_OPERATIONS` / `WRITE_OPERATIONS` in
+ *                       `bin/orm/emit.ts`, hand-written, pinned to nothing
+ *
+ * The third is the one an application actually calls. An operation added to the
+ * compiler and left out of the emitter would leave `docs/orm.md` correct, the
+ * `docs-scope` guard green, and `User.groupBy` simply not existing — the one
+ * failure mode of the three that no test could see, and the only one a reader
+ * meets as "the documentation says this exists".
+ *
+ * That is not hypothetical for this shape of code: #182 found the operation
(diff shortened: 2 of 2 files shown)


nstfkc:
**Follow-up: the guard carried the bug it guards against.**

The first commit added a hand-maintained operation list — copied from `docs-scope.test.ts` — in a test whose entire purpose is catching hand-maintained operation lists that drift. That is #226, inside the fix for #226.

Both files now **derive** the set from the compiler:

```ts
const IMPLEMENTED = [
  ...Object.keys(READ_ARGS),
  ...Object.keys(WRITE_ARGS),
  "aggregate",
  "groupBy",
] as Operation[];
```

`op in READ_ARGS` is the whole of `isReadOperation`, and both tables were already exported, so their keys are the compiler's own record of what it implements. No production change was needed.

## It also fixes the direction the original list could not check

`docs-scope.test.ts` verified each hand-written entry against the predicates. That catches a name that does not exist — but never **an operation that exists and was never added to the list**, which is the failure a scope guard is for, and the one #182 and #195 kept finding in hand-written copies elsewhere.

**Verified by adding a sixteenth operation to `READ_ARGS` and nothing else:**

```
× the derived set is the whole surface
× findWhatever is declared on the generated base
× it names exactly the operations the compiler implements
```

Three assertions across the two files. Under the previous hand-written lists, none of the three would have noticed.

A count assertion comes with it, because two of the fifteen are reached by literal comparison rather than by a table — so a fifth kind of predicate added later would be invisible to the derivation. Fifteen is asserted rather than assumed, and the failure names what to look at.

## Re-checked

- `bun --bun vitest run orm database` — 55 files, **1482 passed**
- template suite — 17 passed / 3 skipped, 629 tests
- typecheck exit 0, `oxlint --deny-warnings` 0 warnings

nstfkc:
## Review — the derivation is right; the assertion is whole-file where the failure is per-model

Tree restored and green: 53 files, 1454 passed.

**The derivation checks out, and it is the part I would have got wrong.** `READ_ARGS` has six keys — the five finds plus `count` — and `WRITE_ARGS` seven, so with `aggregate` and `groupBy` named the set is exactly the fifteen in `Operation`, no gap in either direction. Deriving it beats what #226 proposed: the issue pointed at `READ_OPERATIONS` / `WRITE_OPERATIONS` as *the* generator list, and those two hold nine of the fifteen — the rest come from `BATCH_OPERATIONS` and the hardcoded `countOperation` / `aggregateOperation`. A guard built against the two named arrays would have reported six operations missing. Reading the artifact avoids all five sources, and making `docs-scope.test.ts` derive its list too is a real strengthening rather than a tidy-up: that file's hand-written array was the last copy in the chain.

The hardcoded `toBe(15)` is worth keeping, for what it is worth. It looks like the fourth copy of a number this repository has been burned by, but it is the only assertion of the count that is not derived from the same two tables — `operation-count.test.ts` reaches it through `docs/orm.md`, which `docs-scope` now pins to those tables, so a derived count here would be circular and could not catch the fifth-predicate-kind case the comment describes.

### The guard cannot see a per-model omission

`static ${operation}` is matched against the whole file, so it passes as long as **one** of the thirteen classes declares the operation. Removing `createMany` and `groupBy` from `AccountModel` alone:

```
removed createMany and groupBy from AccountModel only
remaining file-wide counts: createMany=12  groupBy=12
Tests  33 passed (33)
```

Green — while `AccountModel.createMany(...)` does not exist. That is verbatim the thing the failure message promises to catch: *"an application cannot call it."*

Not reachable through today's emitter, which appends every operation to every class unconditionally — I checked `emitModelsFile`, and the only per-operation conditional is `op.required` deciding a `?` on the args parameter. So this is not a live defect. It is worth fixing anyway, because the guard's entire reason to exist is a future change to that emitter, and per-operation variation in those tables is already a thing that exists. Matching within each `export class …Model` block, or asserting the occurrence count equals the number of model classes, closes it for about as many lines as the current version.

### The two directions are matched with different strictness

The positive test uses `\bstatic ${operation}\s*[<(]` with no anchor; the negative one uses `^\s*static (\w+)\s*[<(]` with `m`. So the "is it declared" half would also be satisfied by the name appearing in a comment or a string, and the "nothing beyond" half would not see such a thing. Today no occurrence in the artifact is anything but a declaration — I checked, every `static <name>` line is a real one — so it does not currently bite, and anchoring the positive one the same way is a two-character change that keeps the pair reading as a pair.

### Confirmed working

- Adding a `static bogus<T>(…)` to a model class fails *the base declares no operation the compiler does not implement*, naming it.
- Both declaration spellings are handled — the three batch operations have no generic, and `[<(]` covers them. The PR's note about its first version being stricter than the thing it guards is the same trap #226 warned about, and it is worth having written down.

Nothing above argues against merging. The per-model point is the one I would fix first, since it is the difference between the guard checking the property it states and checking a weaker one.

PR #229, opened 2026-07-29
The Errors table was checked in one direction only

### LINKED ISSUE #228: The Errors table is checked in one direction only
`docs-scope.test.ts` asserts that every exported `Error` subclass appears in `docs/orm.md`'s Errors table:

```ts
test.each(exported)("%s is in the table", (name) => {
  expect(errorsTable).toContain(`\`${name}\``);
});
```

Nothing asserts the converse. A row naming a class that no longer exists would stay in the table indefinitely.

That is the half a reader acts on. The section opens *"Every failure is a typed error from `gemi/orm`"*, and the table is where you look to find out what there is to catch — a row for a deleted error sends someone to write a `catch` for something no longer importable.

This is the same asymmetry the operations list carried until #226: a list verified in one direction reads as though it were verified in both. Both sets agree today, at 21, which is the moment to pin it rather than after they stop agreeing.

### Fix

Assert that every error name in the table is exported from `gemi/orm`.

Worth checking that it adds coverage rather than duplicating the existing assertion. **Renaming** a row trips both — the old assertion sees the name go missing, the new one sees the name that replaced it. The discriminating case is a row **added** with none removed; only the new assertion fails there.

nstfkc:
Closes #228.

`docs-scope.test.ts` asserts every exported `Error` subclass appears in the Errors table. Nothing asserted the converse, so a row naming a class that no longer exists would have stayed there.

That is the half a reader acts on: the section opens *"Every failure is a typed error from `gemi/orm`"*, and the table is where you look to find out what there is to catch. A row for a deleted error sends someone to write a `catch` for something no longer importable.

The same asymmetry the operations list carried until #226 — a list verified in one direction reads as though it were verified in both. Both sets agree today, at 21, which is when to pin it rather than after they stop agreeing.

## Verified that it adds coverage, not duplication

**Renaming** a row trips both assertions — the old one sees the name go missing, the new one sees the name that replaced it. So the discriminating case is a row **added** with none removed:

```
× the table names no error the ORM does not export

AssertionError: the Errors table names PhantomError, which gemi/orm does not
export: expected [ 'PhantomError' ] to deeply equal []
```

Only the new assertion fails there. Worth checking rather than assuming — a second assertion that fires only when an existing one already has adds nothing.

## Also verified this round, and found correct

Recording so they are not re-investigated:

- **All six of CI's excluded files still fail.** #204 merged this iteration, so the exclusion list is now the only thing keeping them out — and a *stale* exclusion would silently narrow coverage, which `ci-coverage.test.ts` cannot see (it checks the paths exist, not that they still fail). Five fail with assertions; `parseTranslation.test.tsx` still fails at collection. The list is accurate.
- **`feat/orm` is green** on the new whole-package filter after #204 landed.
- The `ScalarType` union and the generator's `SCALAR_TYPES` table agree exactly; the dialect `encode`/`decode` gaps are genuine pass-throughs already round-tripped by `writes.coercion`.

## Checks

- `bun --bun vitest run orm database` — 55 files, **1460 passed**

One assertion in an existing file; independent of #225 and #227.

### DIFF
--- a/packages/gemi/orm/docs-scope.test.ts
+++ b/packages/gemi/orm/docs-scope.test.ts
@@ -152,6 +152,37 @@ describe("docs/orm.md lists every error an application can catch", () => {
   test.each(exported)("%s is in the table", (name) => {
     expect(errorsTable).toContain(`\`${name}\``);
   });
+
+  /**
+   * ...and the other direction, which nothing checked.
+   *
+   * The assertion above catches an error class the table forgot. It cannot
+   * catch a row naming a class that no longer exists — and that is the half a
+   * reader acts on, because the table is where you look to find out what there
+   * is to catch. A row for a deleted error sends them to write a `catch` for
+   * something no longer importable.
+   *
+   * The same asymmetry the operations list had until #226: a hand-written list
+   * verified in one direction reads as though it were verified in both. Both
+   * sets happen to agree today, at 21 — which is exactly when to pin it, rather
+   * than after they stop agreeing.
+   */
+  test("the table names no error the ORM does not export", () => {
+    const named = [...errorsTable.matchAll(/`([A-Z]\w*Error)`/g)].map(
+      (match) => match[1],
+    );
+
+    expect(named.length, "no error names found — the table format moved").toBeGreaterThan(10);
+
+    const unknown = [...new Set(named)].filter(
+      (name) => !exported.includes(name),
+    );
+
+    expect(
+      unknown,
+      `the Errors table names ${unknown.join(", ")}, which gemi/orm does not export`,
+    ).toEqual([]);
+  });
 });
 
 /**


nstfkc:
## Review — correct, and the third number is clean too

Small and right. The new assertion is load-bearing: a `PhantomError` row fails it, naming the row. (`orm.md is embedded verbatim` fails alongside it, but only because editing the page desyncs `llms-full.txt` — that is my mutation method, not a second assertion covering the same ground.)

**Both sets agree at 21, and so does the one nobody checks.** The natural next question is what happens to an `Error` subclass that is neither exported nor documented — invisible to both directions, since the forward test iterates exports and the new one reads the table. Measured across every non-test `.ts` under `orm/`:

```
exported from gemi/orm      21
rows in the Errors table    21
declared in orm/ source     21
declared but not exported   none
```

So the loop is closed rather than merely symmetric, which is worth knowing because it is the assumption the pair rests on and nothing states it.

Two smaller things that came out of checking:

- **The `Error` suffix in the regex is doing real work**, not just being conservative. Dropping it to `` `([A-Z]\w*)` `` picks up `MAX_RELATION_DEPTH` and `RETURNING` from the same section and reports them as unknown. So the suffix is what makes a section-scoped match viable. It does mean a future error class not named `*Error` could have a stale row the reverse test cannot see — but the convention is 21 for 21, and the alternative is worse, so that reads as a trade to record rather than one to change.
- **No stale names outside the section either** — 21 error names in the `## Errors` block, 21 in the whole of `docs/orm.md`. And `llms-full.txt` is not a second unchecked copy: the same file already pins the ORM pages into it verbatim, so the table travels with them.

### The same asymmetry is still live in the thing you re-checked by hand

The PR records, in passing, that all six of CI's excluded files still fail — and why that matters now that #204 has merged. That check is the one `ci-coverage.test.ts` does not make. It asserts each excluded path **still exists**:

```ts
test.each(excluded)("%s still exists", (path) => { … existsSync … })
```

while the comment three lines below states the property that actually matters:

> The exclusions are #163's failures. If one is fixed it should be run, not left excluded — so the list is pinned to the reason, and the reason has to be findable.

...and then pins only that `#163` appears in the workflow. A file that has since been fixed stays excluded, coverage narrows, and nothing says so — which is this PR's own thesis, one file over: *a list verified in one direction reads as though it were verified in both*. The manual re-verification this round is the stand-in for the missing assertion, and it is the kind of stand-in that works exactly once.

Not this PR's job, and I would not hold it for that. But it is the strongest evidence that the pattern this PR is closing is worth closing generally, and #204 landing is what made the exclusion list load-bearing rather than redundant.

PR #231, opened 2026-07-29
Document the two Json nulls, which shipped without a word

### LINKED ISSUE #230: The two Json nulls shipped undocumented, and the docs offered a third that does not type-check
#223 taught the runtime to store `Prisma.DbNull` and `Prisma.JsonNull` correctly; #225 taught the filter path to read them. Neither told anybody.

```
$ grep -c "DbNull\|JsonNull" docs/orm.md docs/orm-rows-and-entities.md
docs/orm.md:0
docs/orm-rows-and-entities.md:0
```

They are the **only** way to write an empty `Json` column, and they are absent from both pages.

### The page offered something that does not work

> A `Json` column round-trips whatever you give it: an object, an array, a string, a number, a boolean, **`null`**.

A bare `null` does not type-check on a `Json` field — not on Prisma, and not here, because gemi takes Prisma's argument types verbatim. Verified with a `@ts-expect-error` on both.

So the documented way to write an empty Json value was a type error, and the real way was unmentioned. (That sentence arrived in #216, which is mine.)

### Why the distinction matters

The two sentinels produce **different rows**:

```
Prisma.DbNull     the column is SQL NULL
Prisma.JsonNull   the column holds the JSON value null
```

Both read back as `null` from JavaScript, so the difference is invisible from the application and entirely visible to anything else that reads the column — `psql`, a report, Prisma itself. `differential.test.ts`'s seed comment already calls conflating them "the wrong one of two legal answers".

### Fix

Document both, on writes and on filters, along with the `InvalidArgumentError` #225 added for the bare spelling — noting that Prisma rejects that spelling too, so it is parity rather than a gemi restriction.

Then pin the sentinels **by name**, so a shipped feature cannot go silent again.

### Worth knowing when writing the guard

An assertion that the page no longer offers a bare `null` will not fire if it matches ``a boolean, `null` `` on one line: the sentence wraps between "a" and "boolean". Match across the wrap, and check the guard against the page as it stands — if it does not fail there, it is not guarding anything.

nstfkc:
Closes #230.

#223 taught the runtime to store `Prisma.DbNull` and `Prisma.JsonNull` correctly and #225 taught the filter path to read them. Neither told anybody — both ORM pages contained **zero** mentions of either name, and they are the only way to write an empty `Json` column.

The page went further than silence. It said a `Json` column round-trips *"an object, an array, a string, a number, a boolean, `null`"* — and a bare `null` does **not** type-check on a `Json` field, on Prisma or here. So the documented way to write one was a type error and the real way was unmentioned. That sentence arrived in #216, which is mine.

The distinction is not cosmetic: the two sentinels produce **different rows**, both read back as `null` from JavaScript, and the difference is entirely visible to anything else reading the column — which is what `differential.test.ts`'s seed comment calls *"the wrong one of two legal answers"*.

Now documented on writes and on filters, including the `InvalidArgumentError` #225 added for the bare spelling — noting Prisma rejects that spelling too, so it is parity rather than a gemi restriction. Pinned **by name**, so the feature cannot go silent again.

## One of the four new assertions could not fail

Worth reporting, because the way it surfaced is the only reason it did.

`it does not offer a bare null` matched ``a boolean, `null` `` on a single line. The sentence wraps between "a" and "boolean", so the pattern never matched anything and passed on the old page exactly as happily as on the new one — a guard that cannot fail, in a commit about a guard that was missing.

It turned up because I ran the new guard against the page **as it stood** and only three of the four assertions failed. Now matched across the wrap, and verified to fail on the old page:

```
× it does not offer a bare null as a way to write one
```

## Checks

- `bun --bun vitest run orm database` — 55 files, **1463 passed**
- `docs/llms-full.txt` re-synced, so #181's verbatim guard stays green

## Also verified this round, and found correct

- Prisma's `UserCountArgs` rejects `omit` and `distinct`, so gemi's narrower `count` argument set matches Prisma — not an undocumented refusal.
- All six of CI's excluded files still fail, so #204's exclusion list is not stale.

Independent of #227 and #229.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3844,14 +3844,37 @@ everywhere, sort in the application or store an explicit rank column.
 ## `Json` columns
 
 A `Json` column round-trips whatever you give it: an object, an array, a string, a number, a
-boolean, `null`. The value is stored as JSON and comes back as the same JavaScript value, on both
+boolean. The value is stored as JSON and comes back as the same JavaScript value, on both
 dialects and whether it is read at the root or nested inside an `include`.
 
-One thing is worth knowing:
+Two things are worth knowing:
 
 - **A string is a string.** `metadata: "42"` stores the JSON string `"42"`, not the number, and
   `metadata: '{"a":1}'` stores that text as a string rather than as an object. If you want an
   object, pass an object.
+
+- **Empty is two values, and you have to say which.** `metadata: null` does not type-check —
+  Prisma's types do not allow it and gemi takes them verbatim — because a `Json` column has two
+  distinct empty states and guessing between them is what the sentinels exist to prevent:
+
+  ```ts
+  import { Prisma } from "@prisma/client"
+
+  await User.create({ data: { …, metadata: Prisma.DbNull } })    // the column is SQL NULL
+  await User.create({ data: { …, metadata: Prisma.JsonNull } })  // the column holds JSON null
+  ```
+
+  Both read back as `null`, so the difference is invisible from JavaScript and entirely visible to
+  anything else that reads the column — `psql`, a report, Prisma itself. Filtering takes the same
+  pair, and takes them as an explicit comparison:
+
+  ```ts
+  await User.findMany({ where: { metadata: { equals: Prisma.DbNull } } })   // the SQL NULLs
+  await User.findMany({ where: { metadata: { equals: Prisma.JsonNull } } }) // the JSON nulls
+  ```
+
+  A bare `where: { metadata: Prisma.DbNull }` raises `InvalidArgumentError` naming the explicit
+  form. Prisma rejects that spelling too, so this is parity rather than a gemi restriction.
 
 On Postgres the parameter is cast — `$1::text::jsonb` — and the value serialised to match, which
 is what lets a bare number or boolean through. Binding it raw makes the driver send an `integer` to
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1149,14 +1149,37 @@ everywhere, sort in the application or store an explicit rank column.
 ## `Json` columns
 
 A `Json` column round-trips whatever you give it: an object, an array, a string, a number, a
-boolean, `null`. The value is stored as JSON and comes back as the same JavaScript value, on both
+boolean. The value is stored as JSON and comes back as the same JavaScript value, on both
 dialects and whether it is read at the root or nested inside an `include`.
 
-One thing is worth knowing:
+Two things are worth knowing:
 
 - **A string is a string.** `metadata: "42"` stores the JSON string `"42"`, not the number, and
   `metadata: '{"a":1}'` stores that text as a string rather than as an object. If you want an
   object, pass an object.
+
+- **Empty is two values, and you have to say which.** `metadata: null` does not type-check —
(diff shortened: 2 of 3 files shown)


nstfkc:
**Combination checked.** #227, #229 and #231 all touch `docs-scope.test.ts`, so merge order could have mattered. It does not — merged sequentially onto `feat/orm` in that order, all three apply cleanly and the combined suite is green: **56 files, 1497 tests**.

Worth checking rather than assuming, since this repository has already had a semantic merge conflict that broke only in the combination of two independently-green branches (#159/#166). Each of these is green on its own *and* together.

nstfkc:
## Review — all four assertions are load-bearing, and the page is now the first place the plan-cache bug is met

**The guard does what it says.** Reverting `docs/orm.md` to `95e6212` and running it, all four fail — including the corrected wrap pattern:

```
× Prisma.DbNull is named
× Prisma.JsonNull is named
× it says which one is SQL NULL and which is the JSON value
× it does not offer a bare null as a way to write one
```

(`orm.md is embedded verbatim` fails alongside them, which is my method — reverting one file desyncs the copy — not a fifth assertion.)

Catching the one that could not fail is the right instinct and the right way round: running the new guard against the page it was written to reject, and noticing three failures where four were expected. Pinning by *name* rather than by the removed sentence is also what makes the other three immune to the same problem.

**The prose is accurate on every claim I have measured.** A bare `null` is a type error while both sentinels compile; `DbNull` stores SQL NULL and `JsonNull` the JSON value; both read back as `null` through Prisma, so "invisible from JavaScript and entirely visible to anything else" is exactly right; and the bare filter spelling raises, with Prisma rejecting it too.

### The two filter examples disagree with the page when run in the order it prints them

```
where: { metadata: { equals: Prisma.DbNull   } }  ->  sqlnull      // page says: the SQL NULLs
where: { metadata: { equals: Prisma.JsonNull } }  ->  sqlnull      // page says: the JSON nulls
```

One process, no `clearPlanCache()`, against a table holding one row of each kind. The second line returns the first line's rows.

This is the plan-key collision from #225's review — both sentinels canonicalise to `{}`, `DbNull` now compiles to `is null` with no parameter while `JsonNull` stays `= ?`, and one cache entry serves both. #225 merged without it and #224 has since been closed, so nothing tracks it; I put the reproduction on #45.

**Not this PR's defect, and I would not block on it.** But it is worth saying here rather than only there, because this is the change that makes it reachable by *following the documentation*. The two snippets sit four lines apart and are exactly what someone would paste in sequence — before this page existed, hitting the collision needed a reader who already knew both sentinels. The page is right about the compiler and wrong about the branch, and that is a worse position for a documentation change to leave things in than silence was.

If the fix lands first, this needs no edit at all. If it does not, the honest version of these four lines is smaller than the section — which is an argument for the fix, not for weakening the page.

### Smaller

`section.indexOf("\n## ", 1)` returns `-1` when the section is the last on the page, and `slice(0, -1)` would then trim one character instead of selecting the section — so the assertion would scan the whole remainder of the file. `## \`Json\` columns` has four sections after it today, so it does not bite; worth a `=== -1 ? undefined` the way the `errorsTable` slice a hundred lines above already does it, since the two are the same idiom and only one handles the case.

PR #233, opened 2026-07-29
The Relations section understated its own coverage by a third

### LINKED ISSUE #232: The Relations section says thirty relation shapes; the corpus holds 48
`docs/orm.md`'s Relations section closes with:

> the results are the same either way, and there are tests asserting exactly that against Prisma across **thirty** relation shapes.

`RELATION_CASES.length` is **48**.

Measured at runtime rather than counted by eye — the set is derived by filtering `CASES`, so no static count is authoritative. Two independent static parses said ~50; the runtime value is 48.

This is the same drift as the operation count in #182 and #195, in the other direction: a number in prose whose source grew and left it behind. Understating coverage is the harmless direction and it is the same defect — and the docs' own benchmarks section already states the principle, that "a number copied into prose drifts from its source the first time the source is re-run".

### Fix

Correcting thirty to forty-eight only resets the clock. The corpus should assert the sentence: read `docs/orm.md`, extract the number, compare it to `RELATION_CASES.length`. That needs the sentence worded so there is something to extract, and honest that the number is a snapshot.

### One trap when placing the test

`RELATION_CASES` is used inside `if (url) { … }`, so it is tempting to put the assertion there. That would be #205 exactly: the block only exists when `TEST_POSTGRES_URL` is set, and the Postgres job filters with `-t postgres`, which such a test's name would not match — so it would run in **neither** job while looking like coverage.

It needs no database. It belongs outside the gate, where every run sees it.

### Checked and found correct

The five declines the section lists — implicit many-to-many, self-relation, `_count`, relation ordering, and any node with one of those below it — are exactly the reasons `decline()` returns, plus the dialect condition the section states separately. No drift there.

nstfkc:
Closes #232.

`docs/orm.md` said the two strategies are checked against Prisma "across **thirty** relation shapes". The corpus holds **48** — measured at runtime, since `RELATION_CASES` is derived by filtering `CASES` and no static count is authoritative (two static parses said ~50; the runtime value is 48).

The same drift as the operation count in #182 and #195, in the other direction. Understating coverage is the harmless direction and it is the same defect — and the docs' own benchmarks section already names the principle: *"a number copied into prose drifts from its source the first time the source is re-run."*

So rather than correct thirty to forty-eight and wait for it to rot again, **the corpus now asserts the sentence**: the test reads `docs/orm.md`, extracts the number, and compares it to `RELATION_CASES.length`. The sentence is worded so there is something to extract, and says the number is a snapshot.

Verified by setting it back:

```
× the shape count in docs/orm.md matches the corpus
AssertionError: expected 30 to be 48
```

## The placement was nearly the bug I fixed in #206

`RELATION_CASES` is used inside `if (url) { … }`, so that is where the assertion first went. That would have been **#205 exactly**: the block only exists when `TEST_POSTGRES_URL` is set, and the Postgres job filters with `-t postgres`, which this test's name does not match — so it would have run in *neither* job while looking like coverage.

It needs no database, so it sits outside the gate where every run sees it. Confirmed by running the SQLite suite, where it now executes.

## Checked and found correct

The five declines the section lists — implicit many-to-many, self-relation, `_count`, relation ordering, and any node with one of those below it — are exactly the reasons `decline()` returns, plus the dialect condition stated separately. No drift there, recorded so it is not re-checked.

## Checks

- template suite — 17 passed / 3 skipped, **640 tests**
- `bun --bun vitest run orm database` — 55 files, **1459 passed**

Independent of #227, #229 and #231 — and I verified those three still merge cleanly in sequence with a green combined suite (56 files, 1497 tests) before starting this.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3057,8 +3057,9 @@ back to batching for that node alone. It declines:
 
 A declined node still runs under this strategy one level down, so a decline costs one statement, not
 the subtree. A mixed include tree therefore uses both, which is fine — the results are the same
-either way, and there are tests asserting exactly that against Prisma across thirty relation
-shapes.
+either way, and there are tests asserting exactly that against Prisma across every relation shape
+in the differential corpus — **48** of them today, and the suite fails if this number stops
+matching the corpus.
 
 Override per call when you need to:
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -362,8 +362,9 @@ back to batching for that node alone. It declines:
 
 A declined node still runs under this strategy one level down, so a decline costs one statement, not
 the subtree. A mixed include tree therefore uses both, which is fine — the results are the same
-either way, and there are tests asserting exactly that against Prisma across thirty relation
-shapes.
+either way, and there are tests asserting exactly that against Prisma across every relation shape
+in the differential corpus — **48** of them today, and the suite fails if this number stops
+matching the corpus.
 
 Override per call when you need to:
 
--- a/templates/saas-starter/app/models/differential.test.ts
+++ b/templates/saas-starter/app/models/differential.test.ts
@@ -1,3 +1,6 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
 import { Prisma as PrismaNS, type PrismaClient } from "@prisma/client";
 import { Model } from "gemi/orm";
 import { afterAll, beforeAll, describe, expect, test, vi } from "vitest";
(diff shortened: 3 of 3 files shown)


nstfkc:
## Review — the sentence is extractable by design, and the filter under it is the hand-written list now

**The wrap trap is avoided, and not by luck.** The extracted phrase sits entirely on line 366:

```
366: in the differential corpus — **48** of them today, and the suite fails if this number stops
```

so `/differential corpus — \*\*(\d+)\*\* of them today/` matches without spanning a break, returning `48`. That is the failure #231's first attempt had, and the difference is that the sentence was *worded* to keep the number beside its anchor rather than the pattern being written around whatever the prose happened to do. Worth naming as the general technique: make the thing extractable, then extract it.

`llms-full.txt` carries the identical sentence and #181's verbatim guard pins it, so the two copies cannot drift apart even though only one is read.

**Placement is right.** Outside `if (url)`, it runs in the SQLite job unfiltered and in the Postgres job through the describe label, which is what `-t postgres` matches on. The #205 hazard you describe is real and this avoids it.

One thing I could **not** confirm: that the corpus is 48. My Prisma client is generated for the schema before #219's `Organization.plan`, so this file's `beforeAll` fails here and the count test skips rather than runs. The static parses agreeing at ~50 and disagreeing with the runtime is exactly why I would not report a number I have not run — flagging that the 48 in this review is yours, not mine.

### The filter's `select` branch is a hand-written list of relation names

```ts
const RELATION_CASES = CASES.filter(([, , args]) => {
  …
  if (shape.include !== undefined) return true;                 // generic
  return select !== undefined && Object.keys(select).some(
    (key) => key === "accounts" || key === "organization" || key === "session");  // three names
});
```

`User` declares five relations — `organization`, `accounts`, `session`, `passwordResetToken`, `socialAccount`. The first three are accepted; the other two are not. Today nothing selects them, so the count is correct and there is no live defect.

What changed is the claim. The sentence went from *"across thirty relation shapes"* to *"across **every** relation shape in the differential corpus"*, and that stronger claim rests on the filter finding them all. A case added with `select: { socialAccount: true }` — a relation this harness already registers a model for — would not be counted, so the corpus length would not move, so the documented number would not move, so **the guard stays green while the word "every" becomes false**. The count is now pinned to the filter's output rather than to the corpus's relation shapes, and only the second is what the sentence promises.

The tell is inside the filter itself: `include` is handled generically and `select` by enumeration, so two halves of one predicate answer the same question by different rules and only one of them can go stale. That is the defect this PR is about, one level down from the number — and it is the level the strengthened wording moved the risk to.

Deriving the names rather than listing them closes it, and the schema is already loaded in this file. Cheap enough to be worth doing in the same change, since the sentence it protects is the one being written here.

Nothing above argues against merging. Correcting the number by making the corpus assert it is the right shape of fix, and the placement reasoning is sound.

PR #235, opened 2026-07-29
The docs still describe #98 as a known gap; it landed

### LINKED ISSUE #234: The docs describe #98 as a known gap; #98 landed
`docs/orm.md` carries a callout in the nested-writes section:

> **Known gap (#98).** A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — cannot use any relation operand that writes that key, including `connect`, unless it also declares an `onUpdate`.

**#98 landed.** `templates/saas-starter/app/models/nested-write-policies.test.ts:694` records it in as many words:

> **#98 has landed, so this flipped** — as the pin it replaces said it should, rather than being deleted. […] The guard now judges the caller's columns rather than the ORM's, so both operands go through.

The mechanism is in `policy.ts`: the ORM records which columns it wrote under a module-private `Symbol` (`markOrmAuthored` / `ormAuthoredFields`) and the scope-escape guard exempts those.

### Why this direction is the costly one

This is #145's failure with the sign reversed. There, the docs called an implemented operation *out of scope*, and a reader who believed it wrote raw SQL for a query that worked. Here they call a closed gap *open* — and a documented limitation is believed. Nobody tries `connect` to find out whether the note is stale; they restructure around it, or add an `onUpdate` that grants more than they meant to.

### Fix

Replace the callout with what the guard does now — including the half that has **not** changed: the exemption is provenance, not permission. A column *you* name is still a scope escape and still needs an `onUpdate`.

### Guarding it

Not by asking GitHub whether an issue is closed — a test that needs the network fails for reasons unrelated to the code. The contradiction is already inside the repository: a gap the docs name by number, and a test asserting that same number has landed, cannot both be right.

That guard needs one safeguard of its own: assert the "has landed" phrase is still in use somewhere, or an empty set on that side makes it pass by finding nothing — which is exactly what the state it catches looks like.

nstfkc:
Closes #234.

`docs/orm.md` carried a callout headed **Known gap (#98)**: a child scoped on its own foreign key "cannot use any relation operand that writes that key, including `connect`, unless it also declares an `onUpdate`".

**#98 has landed.** `nested-write-policies.test.ts:694` says so in as many words — *"**#98 has landed, so this flipped** — as the pin it replaces said it should"* — and pins both `connect` and `disconnect` succeeding. The mechanism is in `policy.ts`: the ORM records which columns it wrote under a module-private `Symbol`, and the scope-escape guard exempts those, judging the caller's columns rather than its own.

So the page told a reader to work around something the suite had already proved was fixed.

## The costlier direction

This is #145's failure with the sign reversed. There the docs called an implemented operation *out of scope*, and a reader who believed it wrote raw SQL for a query that worked. Here they call a closed gap *open* — and a documented limitation is **believed**. Nobody tries `connect` to find out whether the note is stale; they restructure around it, or add an `onUpdate` that grants more than they meant to.

Replaced with what the guard actually does, including the part that has **not** changed: the exemption is provenance, not permission. A column *you* name is still a scope escape and still needs an `onUpdate`.

## The guard does not ask GitHub

A test that needs the network fails for reasons that have nothing to do with the code. The contradiction is already inside the repository — a gap the docs name by number, and a test asserting that same number landed, cannot both be right.

It carries one safeguard of its own: an assertion that the "has landed" phrase is still in use, because an empty set on that side would make it pass by finding nothing, which is precisely what the state it catches looks like.

Verified against the page as it stood:

```
AssertionError: docs/orm.md calls #98 an open gap, but a test in this
repository says it has landed: expected [ '98' ] to deeply equal []
```

## Checks

- `bun --bun vitest run orm database` — 56 files, **1461 passed**

New file plus a docs edit, so it does not touch `docs-scope.test.ts` — deliberately, since #229 and #231 both do. I re-verified those two and #233 still merge cleanly onto the current `feat/orm` with a green combined suite (56 files, 1497 tests) before starting.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3134,11 +3134,15 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
-> **Known gap (#98).** A child whose policy *scopes on the foreign key itself* — `scope: () => ({
-> folderId: mine })` — cannot use any relation operand that writes that key, including `connect`,
-> unless it also declares an `onUpdate`. The scope-escape guard reads the payload and cannot tell a
-> column you supplied from one the nested step put there. Pre-existing rather than new to
-> `disconnect`, and tracked separately because the decision affects `connect` too.
+A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
+still use the relation operands that write that key. `connect`, `disconnect` and the rest go
+through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
+rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
+those from the check.
+
+The exemption is provenance, not permission. A column *you* name is refused exactly as before —
+writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
+`onUpdate` to say so deliberately.
 
 `disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
 Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -439,11 +439,15 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 `connectOrCreate` succeeded would together tell you a row with that key exists in someone else's
 tenant.
 
-> **Known gap (#98).** A child whose policy *scopes on the foreign key itself* — `scope: () => ({
-> folderId: mine })` — cannot use any relation operand that writes that key, including `connect`,
-> unless it also declares an `onUpdate`. The scope-escape guard reads the payload and cannot tell a
-> column you supplied from one the nested step put there. Pre-existing rather than new to
-> `disconnect`, and tracked separately because the decision affects `connect` too.
+A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
+still use the relation operands that write that key. `connect`, `disconnect` and the rest go
+through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
+rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
+those from the check.
+
+The exemption is provenance, not permission. A column *you* name is refused exactly as before —
+writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
+`onUpdate` to say so deliberately.
 
 `disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
 Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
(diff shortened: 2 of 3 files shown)


nstfkc:
## Review — the guard works on the case it was built for, and its one safeguard is satisfied by its own text

Confirmed the headline: reverting `docs/orm.md` and running it produces exactly the failure quoted, and the rewritten prose is accurate on both halves — `ScopeEscapeError` is still expected at `nested-write-policies.test.ts:408` and `:626` for a caller-named column, so *"provenance, not permission"* is pinned rather than asserted. Not asking GitHub is the right call for the reason given.

### The `landed` safeguard is satisfied by `docs-gaps.test.ts` itself

The comment states the hazard precisely:

> An empty `landed` set would make this pass by finding nothing — which is how the state it exists to catch would look.

But the scan reads every `*.test.ts` in those two directories, and the new file is one of them. Its own doc comment quotes the phrase — *"**#98 has landed, so this flipped**"* — and that quotation matches `/#(\d+) has landed/`:

```
landed set (full scan):
   #98 <- ['docs-gaps.test.ts', 'nested-write-policies.test.ts']

excluding the guard's own file:
   #98 <- ['nested-write-policies.test.ts']

if nested-write-policies.test.ts dropped the phrase, landed.size would be:  1
```

So the safeguard passes on the strength of the guard quoting itself. The one state it exists to detect — the phrase falling out of use — is the one state it cannot report, because removing it everywhere else still leaves this file. Skipping the file the check lives in is a one-line fix and restores the property the comment claims.

### After this change nothing matches `claimedGaps`, so the other side cannot fail either

The callout removed here was the only `Known gap (#N)` in either page, so `claimedGaps` is now empty and `contradicted` is `[].filter(…)`. That is a legitimate state rather than a broken read — no documented gaps is a true fact, not a parse failure — so it is not the same defect as above. But two things follow that are worth deciding deliberately rather than inheriting:

- The guard is entirely prospective, and it fires only on the exact phrasing `Known gap (#…)`. There are now **zero** surviving instances of that convention in the repository, so the next person writing a gap callout has no example to copy and no reason to guess this spelling. A guard pinned to a convention with no remaining exemplars is a guard pinned to nothing.
- The asymmetry is worth naming: `landed` got a non-emptiness assertion and `claimedGaps` did not, though both are extracted by pattern from prose. The reasoning that justified the first applies to the second the moment the convention is used again.

Widening the pattern — `known (gap|limitation|issue) \(#\d+\)`, or simply `#(\d+)` inside a blockquote — would give it something to bite on and remove the dependence on one exact phrase. Worth doing here, since this change is what emptied the set.

### Two smaller ones

`readdirSync` is not recursive, so `packages/gemi/orm/compile/` and `dialect/` are not scanned. Nothing in them records a landing today, so nothing is missed — but `landed` is silently narrower than "a test in this repository", which is what the failure message claims it means.

The message hardcodes `docs/orm.md` while `claimedGaps` reads `orm-rows-and-entities.md` as well, so a stale callout on the second page would be reported against the first. One interpolation.

None of this argues against merging — the documentation fix is right and overdue, and the direction it corrects is genuinely the costly one.

PR #237, opened 2026-07-29
The skew warning read the clock on every query

### LINKED ISSUE #236: The skew warning reads the clock on every query
#214 added a check to `execute` — the single point every statement passes through — with a comment claiming a correctly configured process "pays one boolean".

It does not. `warnedProtocolSkew` latches only when a warning **fires**, so in the case that matters — a correct deployment, where none ever does — the flag stays `false` forever and every query pays:

```ts
if (!warnedProtocolSkew) {
  const warning = protocolSkewWarning(…, new Date().getTimezoneOffset());
  …
}
```

Measured:

```
new Date().getTimezoneOffset()   60.0 ns
a boolean check                   1.58 ns
```

38× what the comment claims, on the hot path of a project whose stated first priority is performance. It is small beside a ~148 µs read — this is not a crisis — but it is a needless allocation paid forever by everyone who configured things correctly, for a diagnostic aimed at people who did not.

### Fix

Read the offset once at module load and reduce it to the question actually asked — *is this clock UTC*. A UTC process then never enters the check at all; a non-UTC one stops entering once it has warned. Either way the steady state is the boolean the comment promised.

Caching is safe across a DST boundary because the only thing read from it is whether the clock is UTC, and a process that is UTC stays UTC. Worth stating explicitly, because `context.ts` deliberately does *not* cache `NODE_ENV` — for reasons that do not apply here.

### Why the tests did not catch it

`protocol-skew.test.ts` passes `offsetMinutes` in as a parameter, which is right — CI runs under `TZ=UTC`, the configuration where the fault cannot occur, so a test reading the clock would be untestable in the one place it must run. But that also means the tests exercise the pure function and never the call site, and the call site is where the cost lives.

Any fix should be re-verified end to end rather than from the unit tests.

nstfkc:
Closes #236.

#214 put a check in `execute` — the single point every statement passes through — and its comment claimed a correctly configured process "pays one boolean".

**It did not.** `warnedProtocolSkew` latches only when a warning *fires*, so in the case that matters — a correct deployment, where none ever does — the flag stays `false` forever and every query paid a `Date` allocation and a `getTimezoneOffset()` call.

```
new Date().getTimezoneOffset()   60.0 ns
a boolean check                   1.58 ns
```

38× what the comment said it cost, on the hot path of a project whose stated first priority is performance. Small beside a ~148 µs read — not a crisis — and paid forever by everyone who configured things correctly, for a diagnostic aimed at people who did not.

The offset is now read once at module load and reduced to the question actually asked: *is this clock UTC*. A UTC process never enters the check; a non-UTC one stops entering it once it has warned. Either way the steady state is the boolean the comment promised.

Caching is safe across a DST boundary because the only thing read from it is whether the clock is UTC, and a process that is UTC stays UTC. Stated explicitly because `context.ts` deliberately does **not** cache `NODE_ENV` — for reasons that do not apply here.

## Why the tests could not catch it

`protocol-skew.test.ts` passes `offsetMinutes` in as a parameter. That was the right call and I would make it again: CI runs under `TZ=UTC`, the configuration where the fault cannot occur, so a test that read the clock would be untestable in the one place it must run.

But it also means the tests exercise the pure function and never the call site — and the call site is where the cost lived. So behaviour was re-verified end to end against Postgres 16 rather than inferred from them:

```
TZ=UTC               two findMany() calls -> 0 warnings
TZ=America/New_York  two findMany() calls -> 1 warning
```

## Checks

- `bun --bun vitest run orm database` — 56 files, **1492 passed**
- template suite — 17 passed / 3 skipped, **639 tests**
- scratch container removed

Touches `Model.ts` only; independent of #231, #233 and #235, which I re-verified still merge cleanly onto the current `feat/orm` with a green combined suite (57 files, 1499 tests).

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -957,6 +957,30 @@ async function runSteps(
   }
 }
 
+/**
+ * The process's UTC offset, read **once**.
+ *
+ * The skew warning below needs it, and reading it per call cost a `Date`
+ * allocation on the one path every query takes — 60ns against a boolean's
+ * 1.6ns, measured. That is small beside a query and it is paid by every
+ * correctly configured deployment forever, which is not what the comment on the
+ * call site claimed.
+ *
+ * Safe to cache even though a zone's offset moves across a DST boundary: the
+ * only thing read from it is whether the clock is UTC, and a process that is
+ * UTC stays UTC.
+ */
+const CLOCK_OFFSET_MINUTES = new Date().getTimezoneOffset();
+
+/**
+ * Whether the skew is possible in this process at all.
+ *
+ * `false` in the recommended configuration and in CI, which is what lets the
+ * check below actually cost one boolean rather than merely be described as
+ * costing one.
+ */
+const CLOCK_COULD_SKEW = CLOCK_OFFSET_MINUTES !== 0;
+
 /** Set the first time `protocolSkewWarning` fires; see the call site below. */
 let warnedProtocolSkew = false;
 
@@ -979,16 +1003,16 @@ async function execute(
   values: unknown[],
 ): Promise<unknown> {
   // Said once per process, and only for the configuration that silently returns
-  // the wrong instant — see `protocol-skew.ts`. Behind `warnedProtocolSkew` so
-  // the check disappears after it fires, and ordered so a correctly configured
-  // process pays one boolean.
-  if (!warnedProtocolSkew) {
+  // the wrong instant — see `protocol-skew.ts`. A UTC process cannot produce
+  // the skew, so it never enters this at all; a non-UTC one stops entering once
+  // it has said so. Either way the steady-state cost is the boolean above.
+  if (CLOCK_COULD_SKEW && !warnedProtocolSkew) {
     const warning = protocolSkewWarning(
       dialect.name,
       schema,
       text,
       values,
-      new Date().getTimezoneOffset(),
+      CLOCK_OFFSET_MINUTES,
     );
     if (warning) {
       warnedProtocolSkew = true;
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -20,7 +20,7 @@ import { resolveStrategy } from "./compile/strategy";
 import { matchUniqueKey } from "./compile/unique";
 import { upsertAbsentConflictKey } from "./compile/write";
 import { dialectFor, type SqlDialect } from "./dialect";
-import { protocolSkewWarning } from "./protocol-skew";
+import { clockCouldSkew, createProtocolSkewWarner } from "./protocol-skew";
 import {
   MissingModelSchemaError,
   RecordNotFoundError,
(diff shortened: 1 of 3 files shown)


nstfkc:
## Review — the measurement is right, but "a process that is UTC stays UTC" is not true of `Europe/London`

The cost is real and the diagnosis is correct: `warnedProtocolSkew` latches only on a firing, so the deployment that never warns paid the `Date` allocation forever. Hoisting it is the right instinct.

### The cached offset disables the warning for exactly the deployments it exists to catch

The justification for caching is:

> Safe to cache even though a zone's offset moves across a DST boundary: the only thing read from it is whether the clock is UTC, and a process that is UTC stays UTC.

A process whose offset is **0 right now** is not the same thing as a process that is UTC. `Europe/London` is `+00:00` in winter and `+01:00` in summer:

```
TZ=Europe/London   offset in January  0      offset in July  -60
TZ=Europe/Lisbon   offset in January  0      offset in July  -60
TZ=Africa/Accra    offset in January  0      offset in July    0     (genuinely fixed)
TZ=UTC             offset in January  0      offset in July    0
```

So a process started any time between late October and late March under `TZ=Europe/London` caches `0`, sets `CLOCK_COULD_SKEW` to `false`, and never enters the check again — through the whole of BST, when the skew is real. Driven through the actual `protocolSkewWarning`:

```
TZ=Europe/London
  offset cached at a winter module load = 0    -> silent
  real offset during a summer query     = -60  -> WARNS
```

Before this change the clock was read per query, so that process began warning the moment the clocks went forward. After it, the diagnostic is off for the life of the process — and the failure is seasonal and start-time-dependent, so it reproduces for one engineer and not another.

London is not an exotic case for this warning: the entire audience is deployments that did not set `TZ=UTC`.

**The fix keeps the optimisation.** What is invariant is the *zone*, not the offset — so ask the question of the zone rather than of today. Reading two instants at module load and setting the flag if *either* is non-zero separates the four rows above correctly, and the steady state is still one boolean:

```ts
const CLOCK_COULD_SKEW =
  new Date(Date.UTC(2001, 0, 1)).getTimezoneOffset() !== 0 ||
  new Date(Date.UTC(2001, 6, 1)).getTimezoneOffset() !== 0;
```

### The new gate has no test that can fail

Inverting it — `=== 0` instead of `!== 0` — leaves the suite completely green:

```
baseline        56 files, 1492 passed
gate inverted   56 files, 1492 passed
```

Which follows from the reason you give for the tests being shaped as they are: CI runs `TZ=UTC`, the one configuration where neither branch can produce a warning. So the gate is invisible in both directions, and the end-to-end verification in the description — the two `TZ=` runs — is the only thing that checked it and is not pinned anywhere.

That is worth closing while the file is open, because it is now the *second* untested property at this call site: `warnedProtocolSkew = true` has no test either, and removing that line also leaves the suite green. Both are testable without a database and without depending on the ambient clock, by exporting the predicate that decides whether to enter — the same move that made `protocolSkewWarning` itself testable under `TZ=UTC`, which the description rightly defends.

That refactor would also have caught the London case, since a predicate taking a zone can be asked about `Europe/London` from a UTC machine.

nstfkc:
## Addressed — both points fixed in fe97a27; you were right about London, and the cached offset was worse than one bug

### 1. "a process that is UTC stays UTC" is not true of `Europe/London` — **fixed** (fe97a27)

Confirmed exactly as described. The gate now asks the zone rather than today:

```ts
export function clockCouldSkew(timeZone?: string): boolean {
  const year = new Date().getUTCFullYear();
  for (let month = 0; month < 12; month++) {
    if (offsetMinutesAt(Date.UTC(year, month, 1), timeZone) !== 0) return true;
  }
  return false;
}
```

Twelve probes rather than your two, for one reason your table does not cover: northern-summer DST is not the only shape a transition comes in. `Africa/Casablanca` moves with Ramadan, which walks through the calendar, so a January/July pair is a bet on where the transitions fall rather than a question about the zone. It costs ~0.1 ms once at module load — less than a single query — so there was no reason to take the bet.

**There was a second half to this that the gate alone does not fix.** Opening the gate for London is not enough while `CLOCK_OFFSET_MINUTES` is still what gets passed *in*: a winter-started process clears the gate, hands `protocolSkewWarning` its cached `0`, and is silenced by that function's own `offsetMinutes === 0` test for the whole of BST. Same bug, one level down. So the offset goes back to being read per call — inside the gate, which means only a process that could skew ever reads the clock, and it stops as soon as it has warned. That is the deployment the cost is worth paying for, and it is what the code did before #237 anyway.

The steady state you were promised survives: **0.5–0.9 ns/op** for the call site as it now stands, against **~60 ns** for the `Date` allocation it replaces.

```
TZ=Europe/London  jan=0  jul=-60  old gate if started in January = false  new gate = true
TZ=Africa/Accra   jan=0  jul=0    old gate if started in January = false  new gate = false
```

### 2. The new gate has no test that can fail — **fixed** (fe97a27)

Reproduced: inverting `!== 0` left 1492/1492 green, and so did deleting `warnedProtocolSkew = true`. Both are now behind `createProtocolSkewWarner` in `protocol-skew.ts` — your suggested move, and it does what you predicted: a predicate that takes a zone can be asked about `Europe/London` from a UTC machine, which is what pins point 1 as a unit test rather than as a `TZ=` run in a description.

Five mutants, all green before this commit:

| mutant | result |
| --- | --- |
| gate inverted (`!== 0` → `=== 0`) | 6 failed |
| latch removed (`warned = true` deleted) | 1 failed |
| gate probes January only | 5 failed |
| offset captured once instead of read per call | 2 failed |
| gate dropped entirely | 1 failed |

`Model.ts` keeps the printing and gives up the deciding, since deciding is the part CI cannot see.

### Checks

- `bun --bun vitest run orm database` — 56 files, **1506 passed**
- the CI unit step (whole package minus the #163 six) — 64 files, **1561 passed**, 1 skipped
- the new tests re-run under `TZ=UTC`, `TZ=Europe/London` and `TZ=America/New_York` — identical, as they should be
- end to end against Postgres 16 through `Model.execute`, three `findMany()` each:

```
TZ=UTC               0 warnings
TZ=Europe/London     1 warning
TZ=America/New_York  1 warning
```

The London row is the one the previous version of this PR would have failed with a winter module load. Scratch container removed; `templates/` untouched.

Nothing deferred — both points were about code this PR introduced, and the second is what made the first testable.

PR #239, opened 2026-07-29
The Json null sentinel was forgeable, and could make a write throw

### LINKED ISSUE #238: The Json null sentinel is forgeable, and a hostile toString makes the write throw
`jsonNullKind` (added in #223) recognises Prisma's `Json` null sentinels by `toString` alone. Two faults follow, and the differential suite cannot find either, because no reasonable corpus contains the values that trigger them.

**1. The sentinel is forgeable — silently, with data loss.**

```ts
await User.create({ data: { …, metadata: { a: 1, toString: () => "Prisma.DbNull" } } })
```

is read as `Prisma.DbNull` and written as SQL NULL. The object is gone, and nothing raises — both outcomes are legal values for the column.

`policy.ts` makes exactly this point from the other side: its provenance marker is a module-private `Symbol` *"so an application cannot forge it"*. This had no such property.

**2. A value whose `toString` throws makes the bind throw.**

```ts
metadata: { a: 1, toString() { throw new Error("boom") } }
```

Prisma stores that fine — `JSON.stringify` never consults `toString`. gemi throws from the sentinel check.

### Fix

Recognise structurally rather than textually: **a sentinel carries no data.** Anything with a property in it is not one and never reaches `String`, which closes the forgery and the throw together.

Order matters for the hot path: the enumerable scan first, because it allocates nothing and exits on the first key — every real Json object. The own-property check is what keeps `[]` and `{}` out, an empty array being a legitimate value. Guard `String` too: a null-prototype object has no `toString` at all, and throwing on one would fail a write Prisma accepts.

### A trap when testing it

The ORM runtime may not import the Prisma client package, so the sentinels have to be modelled. Modelling them as `Object.create({ toString() {…} })` **does not work**: a method in an object literal is enumerable, so `for…in` walks it and the shape check rejects it — the model fails while the real sentinels pass.

Prisma builds them as classes, where the prototype's `toString` is non-enumerable. That difference is the whole reason the structural test is safe, so the model has to be a class too.

### Residual, worth stating

An empty class instance whose `toString` returns the exact string remains indistinguishable from the real sentinel without importing Prisma. It carries no data, so nothing is lost by treating it as one.

nstfkc:
Closes #238.

`jsonNullKind` (added in #223, mine) recognised Prisma's sentinels by `toString` alone. Two faults followed, and the differential suite could not have found either — no reasonable corpus contains the values that trigger them:

```
{ a: 1, toString: () => "Prisma.DbNull" }   read as the sentinel, stored as SQL
                                            NULL — the object lost, silently
{ a: 1, toString() { throw … } }            the bind threw, for a value Prisma
                                            stores without ever calling toString
```

The first is the serious one: a **forgeable sentinel whose failure is silent data loss**, both outcomes being legal values for the column. `policy.ts` makes the same point from the other side — its provenance marker is a module-private `Symbol` precisely so *"an application cannot forge it"* — and this had no such property.

## The fix is structural

**A sentinel carries no data.** Anything with a property in it is not one and never reaches `String`, which closes the forgery and the throw together.

Order matters, since this runs per bound Json value: the enumerable scan first — it allocates nothing and exits on the first key, which is every real Json object — then the own-property check, which is what keeps `[]` and `{}` out, an empty array being a legitimate value. `String` is guarded because a null-prototype object has no `toString` at all, and throwing on one would fail a write Prisma accepts.

Verified against the **genuine** `Prisma.DbNull` and `Prisma.JsonNull`, not only the model: both still recognised, `AnyNull` still deliberately not, `{}` / `[]` / null-prototype objects all treated as data.

## The test helper was wrong in the same way the code was

Worth reporting, because it nearly shipped as a passing test.

The first helper built the sentinel with `Object.create({ toString() {…} })`. A method in an object literal is **enumerable**, so `for…in` walks it and the shape check rejects it — the helper failed while the real sentinels passed. Prisma builds them as classes, where the prototype's `toString` is non-enumerable, and that difference is the entire reason the structural test is safe to apply. Modelled as a class now, with the reasoning written down beside it.

## Residual, stated rather than hidden

An empty class instance whose `toString` returns the exact string remains indistinguishable from the real sentinel without importing Prisma, which the runtime may not do. It carries no data, so nothing is lost by treating it as one.

## Checks

- `bun --bun vitest run orm database` — 57 files, **1508 passed**
- template SQLite — 17 passed / 3 skipped, 639 tests
- Postgres differential — 13 files, **668 passed**, unchanged

Touches `json-null.ts` plus a new test file; independent of #233, #235 and #237.

### DIFF
--- a/packages/gemi/orm/json-null.test.ts
+++ b/packages/gemi/orm/json-null.test.ts
@@ -1,0 +1,118 @@
+import { describe, expect, test } from "vitest";
+
+import { jsonNullKind } from "./json-null";
+
+/**
+ * What may and may not be mistaken for one of Prisma's `Json` null sentinels.
+ *
+ * The first version recognised them by `toString` alone, which had two faults
+ * and neither showed up in the differential suite, because no reasonable corpus
+ * contains the values that trigger them:
+ *
+ *   { a: 1, toString: () => "Prisma.DbNull" }   read as the sentinel -> stored
+ *                                               as SQL NULL, losing the object
+ *   { a: 1, toString() { throw … } }            the *bind* threw, for a value
+ *                                               Prisma stores without ever
+ *                                               calling `toString`
+ *
+ * The first is the serious one: a forgeable sentinel whose failure is silent
+ * data loss. `policy.ts` makes the same point in the other direction — its
+ * provenance marker is a module-private `Symbol` precisely so "an application
+ * cannot forge it".
+ *
+ * The fix is structural rather than textual: a sentinel carries **no data**, so
+ * anything with a property in it is not one and never reaches `String`. These
+ * cases are the boundary of that rule, written as a table because the
+ * interesting ones are the values that look like sentinels and are not.
+ *
+ * The real `Prisma.DbNull` and `Prisma.JsonNull` cannot be constructed here —
+ * the ORM runtime may not import the Prisma client package, which
+ * `runtime-isolation.test.ts` enforces against comments too — so they are
+ * modelled by their shape: no own properties, no enumerable keys, and a
+ * `toString` on the prototype. That is exactly how the real ones are built, and
+ * the template's suites exercise the genuine articles end to end.
+ */
+const sentinelShaped = (tag: string): object => {
+  // A **class** instance, not `Object.create({ toString() {…} })`. A method in
+  // an object literal is enumerable, so `for…in` walks it and the shape check
+  // rejects it — which is how the first version of this helper failed while the
+  // real sentinels passed. Prisma builds them as classes, where the prototype's
+  // `toString` is non-enumerable, and that difference is the whole reason the
+  // structural test is safe to apply.
+  class Sentinel {
+    toString() {
+      return tag;
+    }
+  }
+  return new Sentinel();
+};
+
+describe("jsonNullKind recognises a sentinel and nothing else", () => {
+  test.each([
+    ["Prisma.DbNull", "db"],
+    ["Prisma.JsonNull", "json"],
+  ] as const)("%s is recognised", (tag, kind) => {
+    expect(jsonNullKind(sentinelShaped(tag))).toBe(kind);
+  });
+
+  /**
+   * `AnyNull` is a filter-only value that Prisma rejects in a write, so mapping
+   * it would quietly accept something Prisma does not. Pinned as an omission
+   * rather than left to be inferred from its absence.
+   */
+  test("Prisma.AnyNull is deliberately not one", () => {
+    expect(jsonNullKind(sentinelShaped("Prisma.AnyNull"))).toBeNull();
+  });
+
+  test.each([
+    ["a plain object", { a: 1 }],
+    ["an array", [1, 2]],
+    ["an empty object", {}],
+    ["an empty array", []],
+    ["a string", "Prisma.DbNull"],
+    ["a number", 42],
+    ["a boolean", true],
+    ["null", null],
+    ["undefined", undefined],
+    ["a Date", new Date(0)],
+  ])("%s is not a sentinel", (_label, value) => {
+    expect(jsonNullKind(value)).toBeNull();
+  });
+
+  /**
+   * The forgery. An object that answers `Prisma.DbNull` was written as SQL NULL
+   * and the object was lost — silently, because both are legal values for the
+   * column.
+   */
+  test("an object carrying data cannot forge a sentinel", () => {
+    const forged = { a: 1, toString: () => "Prisma.DbNull" };
+
+    expect(jsonNullKind(forged)).toBeNull();
+  });
+
+  /** And the throw: `JSON.stringify` never calls `toString`, so nor may this. */
+  test("a value whose toString throws is data, not an error", () => {
+    const hostile = {
+      a: 1,
+      toString() {
+        throw new Error("toString should not have been called");
+      },
+    };
+
+    expect(() => jsonNullKind(hostile)).not.toThrow();
+    expect(jsonNullKind(hostile)).toBeNull();
+  });
+
+  /**
+   * A null-prototype object has no `toString` at all, and `String` throws on
+   * it. It is a legitimate Json value — `JSON.stringify` handles it — so the
+   * answer is "not a sentinel" rather than a failed write.
+   */
+  test("a null-prototype object does not fail the write", () => {
+    const bare = Object.assign(Object.create(null), { a: 1 });
+
+    expect(() => jsonNullKind(bare)).not.toThrow();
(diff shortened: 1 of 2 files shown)


nstfkc:
## Review — the structural check is right, and it leaves `AnyNull` pinned as a parity it does not deliver

Checked against the **genuine** sentinels rather than the model, since that is the half a model cannot vouch for:

```
Prisma.DbNull        for-in=[]  ownNames=[]          String="Prisma.DbNull"
Prisma.JsonNull      for-in=[]  ownNames=[]          String="Prisma.JsonNull"
[]                   for-in=[]  ownNames=["length"]  -> data, as the comment says
Object.create(null)  for-in=[]  ownNames=[]          String throws -> caught -> data
{a:1,toString:…}     for-in=["a","toString"]         -> data, forgery closed
```

And end-to-end through a write, which is where the fault actually lived:

```
forged literal          stored {"a":1}   the object survives
toString that throws    stored {"a":1}   no throw, matching Prisma
Prisma.DbNull           SQL NULL
Prisma.JsonNull         'null'
[]  /  {}               stored as themselves
```

All six correct. The ordering argument holds too — `for…in` exits on the first key for every real Json object, so `getOwnPropertyNames` only allocates for empty containers. And the residual is genuinely irreducible without importing Prisma; stating it beats pretending otherwise.

### `Prisma.AnyNull` is not refused — it is stored as `{}`

The new test pins *"Prisma.AnyNull is deliberately not one"*, and #223's reasoning for the omission was that mapping it *"would quietly accept something Prisma does not."* Omitting it does not refuse it either. It falls through to data. Measured on this branch, against the real `Prisma.AnyNull`:

```
              Prisma                        gemi
write         PrismaClientValidationError   accepted, stored {}
filter        sqlnull, jsonnull             obj
```

The write is the same silent shape this PR exists to close: an object carrying no data, stored as the jsonb object `{}`, with nothing raised and both outcomes legal for the column. The filter is worse than merely wrong — Prisma's `equals: AnyNull` means *either kind of null*, so it returns the two null rows, and gemi returns the row holding `{}`: the exact complement of the correct answer, silently.

Not introduced here — it has been true since #223. But this is the PR that hardens this function, and the assertion it adds reads as though the omission delivered the parity it was argued for. A test that pins a mechanism where the property is what was promised is the failure mode this repository keeps finding, and here it is one line below the fix.

Refusing `AnyNull` in a write the way #225 refuses a bare sentinel — naming it, and noting Prisma rejects it too — would make the pin true. The filter half is a separate decision: implement it as `is null or = 'null'`, or refuse it. Either beats returning the complement.

### Smaller: the real objects are not exercised where the guarantee is stated

`json-null.test.ts` builds a `class Sentinel`, and the verification against the genuine `Prisma.DbNull` described in the PR body is not pinned anywhere in this diff. That matters more after this change than before it, because the check is now **tighter**: a Prisma release that gave its sentinels any own property would stop them being recognised, and the failure would be a silent return to #222's `{}` mis-store.

The net exists — `writes.differential.test.ts` and `differential.test.ts` do use the real `Prisma.DbNull` — but it is in another package and needs a database, so the SQLite-only path would not see it. Worth a line in the comment pointing at where the real-object coverage actually lives, since this file reads as the place that owns it.

The trap you hit with `Object.create({ toString() {…} })` is worth having written down for the same reason: the model and the real object differ in exactly the property the fix depends on.

nstfkc:
## Addressed — the false claim is out of the diff in 6621b2c, the behaviour it implied is #259

Both points were about what this diff *says*, and one of them was also about what the ORM *does*. The first half is fixed here; the second half is a behaviour change I have kept out of a PR that closes #238.

### `Prisma.AnyNull` is not refused — **deferred to #259**, claim corrected in 6621b2c

You are right on the measurement, and it reproduces statically: `jsonNullKind` returns `null`, so `fieldParam` reaches `JSON.stringify(Prisma.AnyNull)` → `{}`, and in `compileFieldFilter` a bare `AnyNull` takes the `equals` shorthand while `{ equals: AnyNull }` compiles to `= $1::jsonb` bound with `{}` — the complement, silently, exactly as you have it.

What landed in 6621b2c is the part that is this PR's fault. The assertion said *"Prisma.AnyNull is deliberately not one"*, which reads as though the omission refused it. It now says what the function does and what that costs:

```
test("Prisma.AnyNull is not mapped to a kind (and is not refused either — #259)")
```

with the write-stores-`{}` and filter-returns-the-complement outcomes written beside it, and the same correction made to `jsonNullKind`'s doc comment, which carried #223's *"mapping it here would quietly accept something Prisma does not"* — true of mapping, silent about omitting. The assertion itself is unchanged; only what it claims about itself.

The refusal is **#259** rather than a commit here for three reasons, not one:

- It changes `jsonNullKind`'s contract — `AnyNull` has to be recognised, so either the return widens or a second predicate appears — and every call site in `cast.ts`, `where.ts` and `sqlite.ts` reads that return today.
- The throw cannot live in `fieldParam`, which has `field` and `dialect` and neither the schema nor the operation an `InvalidArgumentError` needs. It belongs in `compile/write.ts`, a file this PR does not touch.
- It turns a silent success into a raise for anything already passing `AnyNull` to a write. That is a behavioural surface worth its own review, and its coverage has to be differential against the real object in the template — which this package cannot construct.

The filter half is in #259 as the separate decision you called it: implement `is null or = 'null'`, or refuse. #259 records both, the measured table, and why it was not folded in here.

### The real objects are not exercised where the guarantee is stated — **fixed in 6621b2c**

Fixed as asked, and the header now disclaims ownership rather than implying it. It names the coverage:

```
templates/saas-starter/app/models/writes.differential.test.ts
  `create with Prisma.DbNull` / `create with Prisma.JsonNull`
templates/saas-starter/app/models/differential.test.ts
  `equals` / bare / `not` for both sentinels
```

with your reason for naming it written down — the check is now *tighter* than the `toString` one, so a Prisma release giving its sentinels an own property would stop them being recognised and the failure would be a silent return to #222's `{}` mis-store — plus the fact that those suites need a database, so a SQLite-only run does not see them.

Worth adding that they cannot be pinned here even by importing Prisma: the package has no generated client, and `import { Prisma } from "@prisma/client"` fails at `Cannot find module '.prisma/client/default'`. The comment says that too, so the next reader does not spend the attempt.

The `Object.create({ toString() {…} })` trap was already written down beside the helper in the original diff; I have left that text as it stands.

### Checks

`bun --bun vitest run orm database` — 57 files, 1508 passed. `bunx tsc --noEmit` exit 0. `bunx oxlint orm --deny-warnings` 0 warnings. The commit touches comments and one test title; `jsonNullKind`'s body is byte-identical to what you reviewed.

PR #241, opened 2026-07-29
Assert the compile-time guarantee ScopedPolicy exists for

### LINKED ISSUE #240: Nothing asserts the compile-time guarantee ScopedPolicy exists for
`docs/orm.md` gives exactly one reason to extend the generated `…ScopedPolicy` instead of writing an object literal:

> extending the generated `AccountScopedPolicy` makes `scope`, `onCreate` and `onUpdate` abstract members — so a policy that scopes reads but forgets the write halves is **a compile error rather than a runtime refusal**

`policy.ts` says the same from the other end: the runtime guards *"have to"* exist because "an object literal can always omit a key", while the class *"moves the same rule to compile time … a missing `onCreate` becomes `TS2515` at the class declaration rather than an error on the first `create` that reaches production"*.

**Nothing asserts it.** `ScopedPolicy` is not named in any test. The runtime halves are covered thoroughly by `policies.test.ts` — but the compile-time half is the part the class exists for, and it can only be checked by a type test.

The failure is silent in the worst way: the class keeps working as a base, the runtime guards still fire, and the only thing lost is that the mistake moves from the class declaration to the first `create` in production — which is precisely the trade the class was added to make.

Same shape as #220, where `select` narrowing was the headline guarantee with no type test.

### How it was found

By asking which of the ORM's public exports no test mentions. Worth noting that sweep needs a **recursive** walk: a first pass that read only the top level of each directory missed `orm/compile/*.test.ts` and reported 25 untested exports where the answer is 12.

### Two traps when writing the assertions

- **Directive placement.** With an *inferred* `scope()` return, a bad column is caught at the **method declaration**, not the `return` line. A `@ts-expect-error` on the return matches nothing — and "unused directive" is indistinguishable from "no error here", which is how a first attempt concluded there was no column checking at all.
- **Assert the base, not the subclass.** `new Subclass().scope(...)` has the subclass's *inferred* return type, not the abstract signature, so asserting it says nothing about the guarantee.

nstfkc:
Closes #240.

`docs/orm.md` gives one reason to extend the generated `…ScopedPolicy` rather than write an object literal:

> a policy that scopes reads but forgets the write halves is **a compile error rather than a runtime refusal**

and `policy.ts` says the same from the other end — the runtime guards *"have to"* exist because "an object literal can always omit a key", while the class *"moves the same rule to compile time … a missing `onCreate` becomes `TS2515` at the class declaration rather than an error on the first `create` that reaches production"*.

Nothing asserted it. `ScopedPolicy` was not named in any test: the runtime halves are covered thoroughly by `policies.test.ts`, but the compile-time half is the part the class exists for and can only be checked by a type test. Same shape as #220.

Six assertions: all three members required, each omission separately, and the columns checked in both spellings — which, as the review below establishes, are two *different* checks rather than one written twice. See the "Addressed" comment.

## Three of my own assertions were wrong before they were right

This is the part worth reading, because the last one is why the file is worth having.

1. **`scope()` returning the model's where type** — I asserted it against `new Subclass().scope(…)`, which has the *subclass's inferred* return type, not the abstract signature. Wrong question, so a passing or failing answer meant nothing.
2. **"an unknown column is not caught"** — concluded from an unused `@ts-expect-error` on the `return` line. The diagnostic was real; with an inferred return it lands on the **method declaration**, so the directive matched nothing. *"Unused directive"* and *"no error here"* look identical, and only the second is a finding.
3. **The correction to that** — "annotated catches it, inferred does not" — was wrong the same way in reverse. Both spellings catch it, in different places. That is now two tests with the placement written down. **And a fourth, from review:** "the same check in two places" was itself wrong. Mutating the binding shows the annotated test pins the *identity* of the bound `where` type while the inferred one pins that a binding exists at all — each passes under the mutation the other catches. Fixed in b345608.

## How it was found

By asking which public exports no test mentions. That sweep needs a **recursive** walk: a first pass reading only the top level of each directory missed `orm/compile/*.test.ts` and reported 25 untested exports where the answer is 12. I re-measured before drawing anything from it.

## Checks

- `bun run test:types` — **30 passed**, no type errors
- template runtime suite — 17 passed / 3 skipped, 640 tests
- `bun --bun vitest run orm database` — 57 files, **1499 passed**
- `bunx tsc --noEmit` exit 0 **in `packages/gemi`** — corrected from review: the bullet did not name a directory, and read as though it covered `templates/saas-starter`, where this file actually lives. It does not. The template has ~276 pre-existing type errors in its own app code (`ci.yml:137`), which is exactly why its gate is `bun run test:types` with `--typecheck.ignoreSourceErrors` rather than a bare `tsc`. Neither number moves with this PR.
- `bunx oxlint orm --deny-warnings` 0 warnings

Adds one file; independent of #237 and #239.

### DIFF
--- a/templates/saas-starter/app/models/scoped-policy.test-d.ts
+++ b/templates/saas-starter/app/models/scoped-policy.test-d.ts
@@ -47,9 +47,41 @@ describe("a generated ScopedPolicy demands all three members", () => {
       }
     }
 
-    expectTypeOf(new Complete()).toBeObject();
+    // Instantiating is one of the two things being asserted: a class with an
+    // unimplemented abstract member cannot be `new`ed, so this line is the
+    // positive case the three `@ts-expect-error` tests below are the negatives
+    // of. `expectTypeOf(new Complete()).toBeObject()` stood here and asserted
+    // nothing — every class instance is an object — while reading as though it
+    // checked something about the class.
+    void new Complete();
+
+    // The other thing: the members are bound to *this model's* Prisma types.
+    // That binding is what `emit.ts` contributes and the only part of the base
+    // a hand-written `ScopedPolicy<any, any, any>` would get wrong — the
+    // abstract-ness above would survive it intact.
+    expectTypeOf<AccountScopedPolicy["scope"]>().returns.toEqualTypeOf<
+      Prisma.AccountWhereInput | undefined
+    >();
+    expectTypeOf<AccountScopedPolicy["onCreate"]>()
+      .parameter(1)
+      .toEqualTypeOf<Prisma.AccountCreateInput>();
+    expectTypeOf<AccountScopedPolicy["onUpdate"]>()
+      .parameter(1)
+      .toEqualTypeOf<Partial<Prisma.AccountCreateInput>>();
   });
 
+  /**
+   * The three negatives below. Read them for what they assert and no more: a
+   * `@ts-expect-error` on a `class X extends Y {` line accepts **any**
+   * diagnostic on that line, not specifically the `TS2515` the comments name.
+   * A different error on the declaration — a renamed base, a base that stopped
+   * being a class — would satisfy the directive just as well.
+   *
+   * What pins them to the real thing is mutation rather than the directive:
+   * deleting each `abstract` member from `ScopedPolicy` in turn fails exactly
+   * its own test here, by name, and leaves the other five green. Re-run that if
+   * you change these — the directive alone will not tell you.
+   */
   test("omitting onCreate is a compile error", () => {
     // @ts-expect-error `onCreate` is abstract and unimplemented
     class ScopeOnly extends AccountScopedPolicy {
@@ -90,18 +122,42 @@ describe("a generated ScopedPolicy demands all three members", () => {
   });
 
   /**
-   * The columns are checked too — in both spellings, but the error lands in
-   * different places, which is the whole reason this is two tests.
+   * The columns are checked too. **These two tests are not one check written
+   * twice** — an earlier version of this comment said they were "the same
+   * assertion landing in different places", and mutating the binding in
+   * `generated/models.ts` disproves it. They assert different things, and each
+   * one passes under the mutation the other catches:
    *
-   * With the return type **annotated**, the object literal is checked directly
-   * and the error is on the literal. With it **inferred**, the method's own type
-   * stops being assignable to the abstract member and the error is on the
-   * *method declaration* — so a `@ts-expect-error` on the `return` line matches
-   * nothing and reads as though the check does not exist.
+   * - `ScopedPolicy<any, …>` on `UserScopedPolicy` — binding widened, model
+   *   forgotten. The **inferred** test fails (`Unused '@ts-expect-error'`); the
+   *   annotated test passes, because its excess-property error comes from the
+   *   author's own `: Prisma.UserWhereInput` annotation and would still be
+   *   there if the class extended nothing at all.
+   * - `ScopedPolicy<Prisma.AccountWhereInput, …>` on `UserScopedPolicy` — the
+   *   plausible `emit.ts` bug, right shape, wrong model. The **annotated** test
+   *   fails, on `Property 'scope' … is not assignable to the same property in
+   *   base type 'UserScopedPolicy'` at the method declaration; the inferred
+   *   test passes, because `{ nonexistentColumn: 1 }` is not on
+   *   `AccountWhereInput` either, so its directive is still used.
    *
-   * That is how the first version of this file concluded there was no column
-   * checking at all: the directive was in the wrong place, the diagnostic was
-   * real, and "unused directive" looks identical to "no error here".
+   * So: the **inferred** form pins that a binding exists at all — that the base
+   * is what constrains the return type. The **annotated** form pins the
+   * *identity* of the bound `where` type, by requiring the method to stay
+   * assignable to the abstract member.
+   *
+   * Placement follows from that, and is the other thing worth knowing here.
+   * Annotated, the literal is checked against the annotation and the error is
+   * on the `return`. Inferred, the method's own type stops being assignable and
+   * the error is on the *method declaration* — a `@ts-expect-error` on the
+   * `return` line then matches nothing and reads exactly like "there is no
+   * check here". That is how the first version of this file concluded there was
+   * no column checking at all: the directive was in the wrong place, the
+   * diagnostic was real, and "unused directive" looks identical to "no error".
+   *
+   * Note what each test does *not* prove. Neither failure above is an unused
+   * directive on the annotated test — under the wrong-model mutation it fails
+   * on an unexpected diagnostic instead, catching the bug but not via the error
+   * it names.
    */
   test("an annotated scope rejects a column the model does not have", () => {
     class Annotated extends UserScopedPolicy {
--- a/templates/saas-starter/app/models/scoped-policy.test-d.ts
+++ b/templates/saas-starter/app/models/scoped-policy.test-d.ts
@@ -1,0 +1,137 @@
+import { expectTypeOf, test, describe } from "vitest";
+
+import type { Prisma } from "@prisma/client";
+import type { PolicyContext } from "gemi/orm";
+
+import { AccountScopedPolicy, UserScopedPolicy } from "./generated";
+
+/**
+ * **A policy that scopes reads but forgets the write halves is a compile
+ * error**, which is the whole reason the generated `…ScopedPolicy` exists.
+ *
+ * `docs/orm.md` states it as the reason to extend the class rather than write an
+ * object literal:
+ *
+ * > extending the generated `AccountScopedPolicy` makes `scope`, `onCreate` and
+ * > `onUpdate` abstract members — so a policy that scopes reads but forgets the
+ * > write halves is a compile error rather than a runtime refusal
+ *
+ * and `policy.ts` says the same from the other end: `assertCreateCovered` and
+ * `assertNoScopeEscape` enforce it at runtime "and have to: an object literal
+ * can always omit a key", while extending the class "moves the same rule to
+ * compile time … a missing `onCreate` becomes `TS2515` at the class declaration
+ * rather than an error on the first `create` that reaches production".
+ *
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the claim holds and the tests are real, but one of the six asserts something other than what its comment says

I mutation-tested the file rather than reading it. Every claim below is from output I produced in a worktree at `f83c4ec` with `templates/saas-starter/node_modules/gemi` pointed at this branch's `packages/gemi`.

### The premise checks out

`ScopedPolicy` was genuinely unasserted on the base:

```
$ git grep -ln "ScopedPolicy" origin/feat/orm -- '*.test.ts' '*.test-d.ts'
(no output)
$ git grep -ln "ScopedPolicy" origin/feat/orm
docs/llms-full.txt  docs/orm.md  packages/gemi/bin/orm/emit.ts
packages/gemi/orm/index.ts  packages/gemi/orm/policy.ts
templates/saas-starter/app/models/generated/models.ts
```

Six mentions, none of them a test. And the guarantee is real in the code — `policy.ts:256-267` declares all three members `abstract`, and `generated/models.ts` binds each model's Prisma types into it (`UserScopedPolicy` at :1894).

`bun run test:types` in the template: **4 files, 30 passed, no type errors** — matches the body. CI already collects it: `.github/workflows/ci.yml:141` runs `bun run test:types` in `templates/saas-starter`, so this is gated, not decorative.

### The four abstract-member tests are load-bearing

I deleted each `abstract` member from `ScopedPolicy` in turn. Each mutation failed **exactly one** test, by name, with the right diagnostic:

| mutation in `packages/gemi/orm/policy.ts` | result |
|---|---|
| drop `abstract onCreate` | `× omitting onCreate is a compile error` — `Unused '@ts-expect-error' directive` at `:54:5` |
| drop `abstract scope` | `× omitting scope is a compile error too` — unused directive at `:80:5` |
| drop `abstract onUpdate` | `× omitting onUpdate is a compile error` — unused directive at `:67:5` |

5 passed / 1 failed in each case. That is the behaviour you want: no cross-talk, and the "all three compiles" test correctly stays green because it is a superset.

### Finding: the two column tests are not two spellings of one check

The doc comment at :92-105 says the annotated and inferred cases are the same assertion landing in different places, and that this is "the whole reason this is two tests". Two mutations to `UserScopedPolicy`'s type arguments say otherwise — they are two *different* checks, and each catches a bug the other misses:

**Binding widened** (`ScopedPolicy<any, Prisma.UserCreateInput, …>`):

```
× an inferred scope rejects it at the method instead
  Unused '@ts-expect-error' directive.  app/models/scoped-policy.test-d.ts:124:7
  Tests  1 failed | 5 passed (6)
```

The annotated test passed. It has to: `scope(): Prisma.UserWhereInput` supplies the type itself, so the excess-property error on `{ nonexistentColumn: 1 }` comes from the author's own annotation, not from anything `UserScopedPolicy` contributes. **That test would still pass if the class extended nothing at all** — as a check on the generated binding it is a tautology.

**Wrong model bound** (`ScopedPolicy<Prisma.AccountWhereInput, …>` on `UserScopedPolicy` — the plausible `emit.ts` bug):

```
× an annotated scope rejects a column the model does not have
  Property 'scope' in type 'Annotated' is not assignable to the same property
  in base type 'UserScopedPolicy'.   app/models/scoped-policy.test-d.ts:108:7
  Tests  1 failed | 5 passed (6)
```

Here the inferred test passed, and the annotated one failed on an *unexpected* diagnostic rather than an unused directive — it caught the bug, but not via the error it claims to be asserting.

So the pair earns its keep, just not for the stated reason. Given that this file's value proposition is that its comments teach directive placement, the comment being wrong about its own subject matter is worth a fix. Suggest replacing "the same check, different placement" with what the mutations show: the annotated form pins the *identity* of the bound `where` type, the inferred form pins that a binding exists at all.

### Smaller notes

- `expectTypeOf(new Complete()).toBeObject()` (:47) asserts nothing — every class instance is an object. The test still has value as a compile gate on the class body (it would catch abstract signatures becoming over-strict), but the assertion line is doing no work and reads as though it is.
- `@ts-expect-error` on a `class X extends Y {` line accepts *any* diagnostic on that line, not specifically TS2515. The mutations show it responding correctly to the real failures, so this is acceptable — but it is a looser assertion than the comments imply.
- The body's "`bunx tsc --noEmit` exit 0" is not true in `templates/saas-starter`, which is where this file lives — pre-existing `TS2786` errors in `app/views/*.tsx`, and `ci.yml:137` itself says the template has "276 type errors unrelated to the ORM". Pre-existing, not caused by this PR, but the claim should name the directory it was run in (`packages/gemi`, presumably, which is what CI typechecks).

### Nothing breaks

`.test-d.ts` does not match vitest's default `include`, confirmed by the runtime run collecting 20 files without it:

```
$ TZ=UTC bun --bun vitest run app/models
Test Files  1 failed | 16 passed | 3 skipped (20)
     Tests  1 failed | 639 passed | 32 skipped (672)
```

The single failure is `generated.test.ts > running prisma generate again produces a zero-line diff`, which fails on its own guard — "gemi-orm-generator is not linked" — an artifact of my synthetic `node_modules`, not of this diff. 640 non-skipped matches the body's count.

`bunx oxlint orm --deny-warnings`: 0 warnings, 0 errors over 96 files. `bunx oxlint` on the new file alone: clean.

### Verdict

Merge. The issue is real, the fix addresses it, and four of the six assertions are proven to fail when the guarantee is removed. Fix the :92-105 comment before or after — it is the one place where the file says something the file itself disproves.

nstfkc:
## Addressed — all four points fixed, nothing deferred

The finding is right and I reproduced both mutations before rewriting anything, in a worktree with `templates/saas-starter/node_modules/gemi` pointed at this branch's `packages/gemi`. Baseline first: `bun run test:types` → 4 files, 30 passed, no type errors.

**`ScopedPolicy<any, …>` on `UserScopedPolicy`** (binding widened):

```
× an inferred scope rejects it at the method instead
  Unused '@ts-expect-error' directive.   app/models/scoped-policy.test-d.ts:124:7
  Tests  1 failed | 29 passed (30)
```

**`ScopedPolicy<Prisma.AccountWhereInput, …>` on `UserScopedPolicy`** (right shape, wrong model):

```
× an annotated scope rejects a column the model does not have
  Property 'scope' in type 'Annotated' is not assignable to the same property
  in base type 'UserScopedPolicy'.       app/models/scoped-policy.test-d.ts:108:7
  Tests  1 failed | 29 passed (30)
```

Both exactly as reported, including which test stays green in each. Fixes in b345608.

### 1. The `:92-105` comment — **fixed** (b345608)

"The same check, different placement" is gone; the comment now leads with the two mutations and what each test is therefore for: the **inferred** form pins that a binding exists at all — that the base is what constrains the return type — and the **annotated** form pins the *identity* of the bound `where` type, by requiring the method to stay assignable to the abstract member. Your point that the annotated test "would still pass if the class extended nothing at all" is in there, as the reason its unused-directive assertion is not the part doing the work.

The placement paragraph is kept, demoted to what it actually is: the reason the two tests look like one, and the trap that produced the first version of this file. And the last paragraph writes down what neither test proves — that under the wrong-model mutation the annotated test fails on an *unexpected* diagnostic rather than an unused directive, catching the bug but not via the error it names.

### 2. `expectTypeOf(new Complete()).toBeObject()` — **fixed** (b345608)

Agreed it asserted nothing. Replaced with the two things the test is really for. The compile gate is now explicit as `void new Complete()` — a class with an unimplemented abstract member cannot be `new`ed, so it is the positive case the three negatives below it invert. Then the part that was missing entirely: that the members carry *this model's* Prisma types.

```ts
expectTypeOf<AccountScopedPolicy["scope"]>().returns.toEqualTypeOf<
  Prisma.AccountWhereInput | undefined
>();
expectTypeOf<AccountScopedPolicy["onCreate"]>()
  .parameter(1)
  .toEqualTypeOf<Prisma.AccountCreateInput>();
expectTypeOf<AccountScopedPolicy["onUpdate"]>()
  .parameter(1)
  .toEqualTypeOf<Partial<Prisma.AccountCreateInput>>();
```

Asserted against the base rather than an instance, which is the mistake noted in the PR body: `new Complete().scope()` has the subclass's inferred return type and answers a different question.

Mutation-tested like the rest. `AccountScopedPolicy extends ScopedPolicy<any, any, …>`:

```
× a policy implementing all three compiles
  Type 'AccountCreateInput' does not satisfy the constraint 'never'.        :67:22
  Type 'Partial<AccountCreateInput>' does not satisfy the constraint …      :70:22
  Tests  1 failed | 29 passed (30)
```

That mutation failed *nothing* before. It is also the mutation the four abstract-member tests cannot see — widening the binding leaves the abstract-ness intact — so this closes the one gap in the file that was not covered by anything.

### 3. `@ts-expect-error` accepts any diagnostic on the line — **fixed** as documentation (b345608)

Taking this as written: acceptable mechanism, comments implying more than it delivers. There is no tighter form available — vitest's typecheck has no assert-this-error-code — so the honest fix is to stop implying it. A block comment above the three negatives now says the directive accepts any diagnostic on a `class X extends Y {` line, not the `TS2515` the comments name, and that a renamed or no-longer-class base would satisfy it just as well. What ties those tests to the abstract members is your mutation run, not the directive, so the comment says to re-run it rather than trust the directive.

### 4. The `bunx tsc --noEmit` claim — **fixed** in the PR body

You are right that it named no directory and that the directory it implied is the one where it is false. The bullet now says exit 0 **in `packages/gemi`** — what `ci.yml:87` typechecks — and states outright that `templates/saas-starter` has ~276 pre-existing app type errors, which is why its gate is `bun run test:types` with `--typecheck.ignoreSourceErrors` rather than a bare `tsc`. Pre-existing and unmoved by this PR, but the bullet should not have needed a reviewer to establish that.

### Nothing deferred

No follow-up issues. All four are comments, one assertion, and a PR-body line inside this file's own concern — none of them widen #240.

### After

- `bun run test:types` in `templates/saas-starter` — 4 files, **30 passed**, no type errors. The count is unchanged because the new assertions live inside an existing test; what changed is that the test now fails when it should.
- `bunx oxfmt --check` on the file — correctly formatted; `bunx oxlint` — 0 warnings, 0 errors.
- Six tests, same six subjects. The diff is comments and assertions; no test's subject changed.

PR #243, opened 2026-07-29
The soft-delete "field" guarantee holds for one of three spellings

### LINKED ISSUE #242: The soft-delete "field" guarantee holds for one of three spellings
`docs/orm.md` claims it flatly:

> `field` is constrained to the model's own keys, so pointing it at a column that does not exist is a compile error rather than a `no such column` on the first read.

That is true of **one** of the three functions it appears to describe.

`SoftDeleteOptions<M>` constrains `field` to `keyof M & string`, and `M` defaults to `any`:

| call | `field` checked? |
| --- | --- |
| `softDeletes<User>({ field: "nope" })` | **yes** — the type parameter supplies `M` |
| `softDeletes({ field: "nope" })` | no — `M` is `any`, so `keyof any & string` is `string` |
| `softDelete(User, { field: "nope" })` | no — the model arrives as a *value* |
| `softDeleteMany(User, { field: "nope" })` | no |

Measured with `@ts-expect-error` on each, and confirmed by removing the directives and watching the calls compile clean.

The example in the docs sits two lines apart:

```ts
static $policies = [softDeletes<User>()]  // checked
static delete = softDelete(User)          // not checked
```

so a typo in the second produces exactly the `no such column` the sentence promises is impossible.

### `M = T` does not fix it

Worth recording, because it is the obvious first attempt. `SoftDeletable<T, Op>` makes `T` the row type, so defaulting `M = T` looks like it should recover the constraint. It does not: `T` is only inferred from the model's `update` signature, which is generic, so it infers too loosely to constrain anything and both calls stay legal.

### Fix

Narrow the sentence to what holds, and give the example the spelling that makes the unchecked call correct by construction — name the options once where they *are* checked, and pass the same object to the call that cannot check them.

Pin the boundary with type tests. The unchecked calls should be written **without** `@ts-expect-error`: if the inference is ever tightened they become errors, the file stops type-checking, and the note has to be revisited — which is the outcome worth having.

nstfkc:
Closes #242.

`docs/orm.md` claimed it flatly:

> `field` is constrained to the model's own keys, so pointing it at a column that does not exist is a compile error rather than a `no such column` on the first read.

That is true of **one** of the three functions it appears to describe.

| call | `field` checked? |
| --- | --- |
| `softDeletes<User>({ field: "nope" })` | **yes** — the type parameter supplies `M` |
| `softDeletes({ field: "nope" })` | no — `M` is `any`, so `keyof any & string` is `string` |
| `softDelete(User, { field: "nope" })` | no — the model arrives as a *value* |
| `softDeleteMany(User, { field: "nope" })` | no |

The docs' own example puts the two spellings two lines apart:

```ts
static $policies = [softDeletes<User>()]  // checked
static delete = softDelete(User)          // not checked
```

so a typo in the second produces exactly the `no such column` the sentence promises is impossible.

## `M = T` does not fix it

Worth recording, because it is the obvious first attempt and I made it. `SoftDeletable<T, Op>` makes `T` the row type, so defaulting `M = T` looks like it should recover the constraint. It does not — `T` is only inferred from the model's `update` signature, which is generic, so it infers too loosely to constrain anything and both calls stayed legal. Reverted, and recorded in the test file.

## The fix

The sentence is narrowed to what holds, and the example gains the spelling that makes the unchecked call correct by construction: name the options once where they *are* checked, and pass the same object to the call that cannot check them.

The type tests pin the boundary rather than a guarantee that does not exist. **The unchecked calls are written without `@ts-expect-error` on purpose** — if the inference is ever tightened they become errors, the file stops type-checking, and the note has to be revisited. That is the outcome worth having, and it is the opposite of what a `@ts-expect-error` would give.

## How it was found

By sweeping the docs for compile-time claims. There are exactly three: `wrap`'s complete-row requirement (covered by `wrap.test-d.ts`), `ScopedPolicy`'s abstract members (covered by #241), and this one — which turned out to be the only one that was not simply untested but partly untrue.

## Checks

- `bun run test:types` — **29 passed**, no type errors
- template runtime suite — 17 passed / 3 skipped, 640 tests
- `bun --bun vitest run orm database` — 57 files, **1499 passed**

Independent of #237, #239 and #241.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3649,8 +3649,22 @@ static $policies: UserPolicy[] = [softDeletes<User>(), new TenantPolicy()]
 the policy scopes on and something has to say that is intended. That permission is **per policy**:
 a tenant policy sitting beside it in the same list still has to answer for its own column.
 
-`field` is constrained to the model's own keys, so pointing it at a column that does not exist is a
-compile error rather than a `no such column` on the first read.
+`field` is constrained to the model's own keys **when the model type is given** —
+`softDeletes<User>({ field: … })`, as above — so a typo there is a compile error rather than a
+`no such column` on the first read.
+
+`softDelete(User)` and `softDeleteMany(User)` take the model as a *value*, and its row type is not
+recoverable from it, so their `field` is unchecked. Spell it once and share it if you override the
+default:
+
+```ts
+const archived = { field: "archivedAt" } as const
+
+export class User extends UserModel {
+  static $policies = [softDeletes<User>(archived)]   // checked here
+  static delete = softDelete(User, archived)          // and therefore correct here
+}
+```
 
 ## Errors
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -954,8 +954,22 @@ static $policies: UserPolicy[] = [softDeletes<User>(), new TenantPolicy()]
 the policy scopes on and something has to say that is intended. That permission is **per policy**:
 a tenant policy sitting beside it in the same list still has to answer for its own column.
 
-`field` is constrained to the model's own keys, so pointing it at a column that does not exist is a
-compile error rather than a `no such column` on the first read.
+`field` is constrained to the model's own keys **when the model type is given** —
+`softDeletes<User>({ field: … })`, as above — so a typo there is a compile error rather than a
+`no such column` on the first read.
+
+`softDelete(User)` and `softDeleteMany(User)` take the model as a *value*, and its row type is not
+recoverable from it, so their `field` is unchecked. Spell it once and share it if you override the
+default:
+
+```ts
+const archived = { field: "archivedAt" } as const
+
+export class User extends UserModel {
+  static $policies = [softDeletes<User>(archived)]   // checked here
+  static delete = softDelete(User, archived)          // and therefore correct here
+}
+```
 
 ## Errors
 
--- a/templates/saas-starter/app/models/soft-delete.test-d.ts
+++ b/templates/saas-starter/app/models/soft-delete.test-d.ts
@@ -1,0 +1,65 @@
+import { test, describe } from "vitest";
+
+import { softDelete, softDeleteMany, softDeletes } from "gemi/orm";
+
+import { UserModel } from "./generated";
+
+/**
+ * Where the `field` option is checked, and where it is not.
+ *
+ * `docs/orm.md` claimed it flatly — "`field` is constrained to the model's own
+ * keys, so pointing it at a column that does not exist is a compile error
+ * rather than a `no such column` on the first read" — and that is true of one of
+ * the three functions it appears to describe.
+ *
+ * `SoftDeleteOptions<M>` constrains `field` to `keyof M & string`, and `M`
+ * defaults to `any`. `softDeletes<User>()` supplies it, so the constraint bites.
+ * `softDelete(User)` and `softDeleteMany(User)` take the model as a *value*;
+ * `M` stays `any`, `keyof any & string` is `string`, and any typo is accepted —
+ * arriving as `no such column` on the first read, which is the failure the
+ * sentence promised was impossible.
+ *
+ * Defaulting `M = T` does not fix it. The row type is only recoverable from the
+ * model's `update` signature, which is generic, so `T` infers too loosely to
+ * constrain anything — tried, and it changed nothing. Recording that here so the
+ * next person does not spend the same half hour on it.
+ *
+ * So this pins the boundary rather than asserting a guarantee that does not
+ * hold. The unchecked calls are written without `@ts-expect-error` on purpose:
+ * if the inference is ever tightened, they become errors, this file stops
+ * type-checking, and the note above needs revisiting — which is the outcome
+ * worth having.
+ */
+type User = { id: number; email: string | null; deletedAt: Date | null };
+
+describe("softDeletes checks `field` when the model type is given", () => {
+  test("a real column is accepted", () => {
+    softDeletes<User>({ field: "deletedAt" });
+  });
+
+  test("a column the model does not have is a compile error", () => {
+    // @ts-expect-error `nope` is not a key of User
+    softDeletes<User>({ field: "nope" });
+  });
+
+  /**
+   * Without the type parameter there is nothing to check against: `M` is `any`,
+   * so `keyof M & string` is `string`. Pinned because the documented sentence
+   * reads as though it applied here too.
+   */
+  test("without the type parameter it is unchecked", () => {
+    softDeletes({ field: "nope" });
+  });
+});
+
+describe("softDelete and softDeleteMany do not check it", () => {
+  test("the model arrives as a value, so `field` is unconstrained", () => {
+    softDelete(UserModel, { field: "nope" });
+    softDeleteMany(UserModel, { field: "nope" });
+  });
+
+  test("the correct spelling compiles, which is all that is asserted", () => {
+    softDelete(UserModel, { field: "deletedAt" });
+    softDeleteMany(UserModel, { field: "deletedAt" });
+  });
+});


nstfkc:
## Review — the diagnosis is right, the conclusion drawn from it is not

The table in the body is correct; I reproduced every row. But the PR then documents around a limitation that a two-token signature change removes, and the sentence it rewrites still names a failure mode that cannot occur.

Environment: worktree at `3692c4a`, `bun install`, `prisma generate` (the checked-in `app/models/generated` restored afterwards — regenerating it without `build:bin` in `packages/gemi` degrades it and produces spurious `aggregate.test-d.ts` failures; that is a trap, not a finding).

### The claims that hold

Base `feat/orm` (`9f28c51`): `bun run test:types` → **24 passed, 3 files, no type errors**. This branch: **29 passed, 4 files, no type errors** — matches the body exactly. `bun --bun vitest run orm database` in `packages/gemi` → **57 files, 1499 passed**, also an exact match. `bunx tsc --noEmit -p packages/gemi/tsconfig.json` exit 0. `bunx oxlint` on the new file: 0 warnings.

I probed each row of the table with a throwaway `.test-d.ts`, using the *inverse* directive — an `@ts-expect-error` on a call the PR says is unchecked should report `Unused`:

| probe | result | verdict |
| --- | --- | --- |
| `softDeletes<RowType>({ field: "nope" })` | real error | checked, as claimed |
| `softDeletes({ field: "nope" })` | `Unused '@ts-expect-error' directive` | unchecked, as claimed |
| `softDelete(UserModel, { field: "nope" })` | `Unused` | unchecked, as claimed |
| `softDeleteMany(UserModel, { field: "nope" })` | `Unused` | unchecked, as claimed |

The `M = T` note is also correct, and worth the space it takes. I declared the `M = T` variant against the real generated models and both bogus-column directives came back `Unused` — it does not bite. Recording a failed attempt in the test file is the right call.

### 1. `no such column` is not what happens — in any of the four spellings

Both the old sentence and the new one turn on the same consequence: an unchecked `field` "arriv[es] as `no such column` on the first read". It does not. I ran it against a real sqlite database (template model, migrations applied):

```
READ WITH BOGUS SCOPE FIELD -> 'nope' is not a field on model User. Known fields: id, publicId,
  name, email, emailVerifiedAt, verificationToken, locale, globalRole, password, organizationId,
  createdAt, updatedAt, deletedAt, metadata, avatar.
SOFT DELETE WITH BOGUS FIELD -> 'nope' is not a field on model User. Known fields: ...
```

That is `UnknownFieldError` (`packages/gemi/orm/errors.ts:133`), raised by the compiler — *"A `where` key with no matching field is an error, never a passthrough"*. The query never reaches the database, so no dialect ever gets to say `no such column`. The unchecked path therefore fails at the first read with the model named and every column listed, which is a considerably better outcome than the docs promise for it.

This matters because the sentence's rhetorical force is entirely in that contrast — compile error *rather than* a cryptic database error. Once the alternative is a clear, named, framework-level refusal, the paragraph is arguing for less than it thinks. The same phrase is repeated in the new file's header comment (`soft-delete.test-d.ts:21`, "arriving as `no such column` on the first read, which is the failure the sentence promised was impossible"), so it needs fixing in three places, not one.

### 2. The row type *is* recoverable from the model value

> `softDelete(User)` and `softDeleteMany(User)` take the model as a *value*, and its row type is not recoverable from it, so their `field` is unchecked.

`M = T` fails, as the PR found. But the instance type is sitting on `prototype`. I applied this to the real source, both functions:

```ts
-  model: SoftDeletable<T, "update">,
+  model: SoftDeletable<T, "update"> & { prototype?: M },
```

and the only thing in the repo that stopped type-checking was this PR's own assertion that it cannot be checked:

```
FAIL app/models/soft-delete.test-d.ts > softDelete and softDeleteMany do not check it
  > the model arrives as a value, so `field` is unconstrained
TypeCheckError: Type '"nope"' is not assignable to type '"deletedAt" | "id" | "publicId" |
  "organizationId" | ... | "save" | undefined'.
  at soft-delete.test-d.ts:57:29  and  :58:33
```

Everything else held with the patch applied: `tsc --noEmit` on `packages/gemi` exit 0, `orm database` **1499 passed**, template `policies.test.ts` **56 passed / 1 skipped** — including the suites that pass structural fakes to `softDelete`, which still work because `prototype` is optional and `M` falls back to `any` when it is absent. The comment at `soft-deletes.ts:92` ("Structural rather than `typeof Model`, so a caller can pass anything that behaves like one") survives intact; I checked that separately with a `{ update, $schema }` object literal and a bogus `field`, and it compiles.

I am not asking you to take my patch as-is — `keyof` an instance type is a loose notion of "column" (see 3), and you may want the narrower version. But the docs change is premised on a limitation that costs two tokens to remove, and the PR should either take the fix or say why the fix is worse than the sentence. As written it does neither.

Note also that today a caller cannot even opt in: the parameters are `<T, M = any>` in that order, so `softDelete<User>(User, …)` binds `T`, not `M`. There is no spelling of the write half that checks `field`.

### 3. The surviving sentence is still wider than the guarantee

Look at the union in the error above — it contains `"save"`. The docs tell readers to write `softDeletes<User>()` where `User` is the model *class*, so `keyof M & string` is the keys of the instance type, methods included. I confirmed the consequence:

- `softDeletes<User>({ field: "save" })` where `class User extends UserModel` — compiles clean (`@ts-expect-error` reported `Unused`).
- At runtime: `READ WITH field:'save' -> 'save' is not a field on model User. Known fields: ...`

So even in the spelling the PR certifies as checked, a `field` that is not a column type-checks and then fails on the first read. "Constrained to the model's own keys" is literally true and is exactly why it is misleading — keys are not columns. This is the same species of overclaim the PR set out to correct, one level down, and the new test does not catch it because it asserts against a hand-written row type (`soft-delete.test-d.ts:31`, `type User = { id; email; deletedAt }`) rather than the generated model the docs point at. Swapping that local type for `UserModel` would both tighten the test and expose this.

### 4. The same flat claim is still in the source

`packages/gemi/orm/soft-deletes.ts:42-44` is untouched:

> Constrained to the model's own keys, so a typo or a model that has no such column is a compile error instead of a `no such column` from the database on the first read.

Word for word the sentence the PR is fixing, on the very declaration whose behaviour is at issue, and it is what shows on hover in an editor — closer to the point of use than `docs/orm.md`. Fixing the prose in two generated-adjacent files and leaving it in the source is backwards.

### 5. The tests

The design is right and I want to say so plainly: writing the unchecked calls *without* `@ts-expect-error` so that tightening the inference breaks the file is the correct instinct, and section 2 is that mechanism firing exactly as intended — the file caught my patch. That is a real test doing real work.

Two caveats. The `@ts-expect-error` case (`:38`) is the only genuine assertion here; the other four tests assert "this compiles", which is worth something as a tripwire but is not equivalent. And `type User` being local (see 3) means the certified-checked path is verified against a shape no caller has.

The runtime side is untested at any level: nothing in the repo asserts what actually happens when `field` names a non-column, which is why the `no such column` sentence has survived two passes. A three-line test asserting `UnknownFieldError` on a bogus `field` would pin the consequence the docs keep describing wrongly.

### 6. Minor: the new example does not compile against the template model

```ts
const archived = { field: "archivedAt" } as const
static $policies = [softDeletes<User>(archived)]   // checked here
```

`archivedAt` is not on the template's `User`, so against `templates/saas-starter` that line is an error — verified, the directive was consumed. The mechanism the example teaches is sound (`as const` is load-bearing; without it `field` widens to `string` and the checked call fails for the wrong reason), and the example is presumably about a hypothetical model. But every other snippet in this section runs against the template, and this one silently does not.

### Suggested shape

1. Take the `prototype` fix, or record why not — the sentence should not be the remedy for a limitation this cheap to remove.
2. Drop `no such column` from `docs/orm.md`, `docs/llms-full.txt` and `soft-delete.test-d.ts`; the real behaviour (`UnknownFieldError`, naming the model and listing the columns) is a better story than the one being told.
3. Update `soft-deletes.ts:42-44` to whatever the docs end up saying.
4. Point the checked test at `UserModel`, and add a runtime test for the bogus-`field` path.

### Environment note

`app/models/generated.test.ts` fails here with *"gemi-orm-generator is not linked. Run `bun install`, and `bun run build:bin` in packages/gemi"* — my environment, not this branch. Template runtime otherwise: **642 passed / 32 skipped**.

nstfkc:
## Addressed — the sentence was wrong about the consequence too, and the type fix is worth more than half of it

Four of the six points are fixed in `719bb4c`; two are deferred, one of them the central one, with reasons below.

Environment matched yours, including the trap: `bun install`, then `prisma generate --generator client` **only** — the `gemi` generator is what degrades `app/models/generated` without `build:bin`, and skipping it leaves the checked-in output alone. Baseline reproduced exactly: 24 → 29 on `test:types`, 1499 on `orm database`.

---

### 1. `no such column` never happens — **fixed** (`719bb4c`)

You are right, and it is worse than a wording problem: the whole paragraph was arguing for a contrast that was never on offer. Confirmed both halves against sqlite, and it is now asserted rather than described — `policies.test.ts`, `describe("a field that is not a column")`:

```
the read half refuses, naming the model and its columns
the write half refuses the same way
```

Both `expect(error).toBeInstanceOf(UnknownFieldError)` with `.field`, `.model` and the column list. That is the test you asked for in point 5, and it is the reason the sentence survived two passes: nothing in the repo had ever run the case.

Dropped from `docs/orm.md`, `docs/llms-full.txt` and the test file's header, and replaced with what actually happens.

### 2. The `prototype` fix — **deferred to #262**

I applied your patch and reproduced your result exactly, down to the error text. It works, structural fakes still pass, nothing else moved. I am not taking it here, and the reason is your own point 3.

Look at the union your patch produces:

```
Type '"nope"' is not assignable to type '"deletedAt" | … | "save" | undefined'.
```

`"save"` is in it. Taking `prototype` alone spreads a **keys-not-columns** check onto two more functions and licenses a docs sentence saying `field` is checked everywhere — which is the exact species of overclaim #242 exists to correct, one level down. Two tokens buy the *reach* of the check; they do not buy the check being about columns. I would rather do both at once, against `$schema.fields` — which the models already carry and `UnknownFieldError` already lists.

So: **#262**, with your patch quoted verbatim, my verification of it, and the argument for pairing it with the narrowing. Your "there is no spelling of the write half that checks `field`" is in there too — and it is sharper than stated: `softDelete<User>(User, …)` does not merely bind `T` instead of `M`, it is *itself* an error, because `update` returns the row without the instance's methods.

Recorded in `soft-delete.test-d.ts` beside the `M = T` note, as a mechanism that **works and was declined**, not one that failed. Your point that the file caught your patch is exactly why that distinction had to be written down.

### 3. The surviving sentence is still wider than the guarantee — **fixed** (`719bb4c`)

The docs now say it outright: the constraint is `keyof` the model's keys, methods included, `field: "save"` type-checks — *"It catches typos, not non-columns."*

And the test is pointed at the generated model as you asked — `class User extends UserModel {}`, which is what the docs tell readers to write, replacing the local `{ id; email; deletedAt }`. That change alone exposes it, so the case is pinned:

```ts
test("a method satisfies it too, which is why keys are not columns", () => {
  softDeletes<User>({ field: "save" });
});
```

No directive, deliberately — narrowing to schema fields (#262) turns this into an error and forces the note to be revisited. Same tripwire, now guarding the thing your review found.

### 4. The same flat claim in the source — **fixed** (`719bb4c`)

Agreed that fixing two generated-adjacent files and leaving the hover text is backwards. `SoftDeleteOptions.field` now carries both limits and the real consequence, and points at both tests.

### 5. The tests — **fixed** (`719bb4c`)

Runtime test added (point 1), checked test repointed at `UserModel` (point 3), and the "this compiles" tripwires kept as they were, since you agree they earn their place. Type tests 29 → 30, template runtime 642 → 644.

### 6. The example does not compile against the template — **fixed, and the column was not the reason** (`719bb4c` + **#263**)

`archivedAt` was invented, and you were right to flag it. But picking a real column does not fix the snippet. The *recipe* does not compile:

```
TypeCheckError: Class static side 'typeof User' incorrectly extends base class static side
  'typeof UserModel'. The types returned by 'delete(...)' are incompatible…
```

`static delete = softDelete(User)` is an illegal override against **any** generated model, regardless of the `field` value — the generated `delete` is generic in its arguments and returns a result narrowed by them. It is in `docs/orm.md`, `docs/llms-full.txt` and `soft-deletes.ts` alike, and the reason no test caught it is that `policies.test.ts:110` writes `static softDelete = softDelete(GuardedUser)` — a different name, so there is no override to be incompatible with. Every runtime soft-delete test passes because none of them uses the spelling the docs teach.

The new example is now `acceptedAt` on `OrganizationInvitation`, a real template column, with `static expire`, and I compiled it. The `as const` note is kept — you are right that it is load-bearing, and I verified the widened version fails.

The top-of-section recipe is **left as written** and filed as **#263**, because the fix may well be to make the override legal rather than to rename it, and choosing here would pre-empt that. Noted in the docs and on the declaration so it is visible rather than latent.

---

### Checks

- `bun run test:types` — **30 passed**, 4 files, no type errors
- `bun --bun vitest run orm database` in `packages/gemi` — 57 files, **1499 passed**
- template runtime — **644 passed / 32 skipped**; `generated.test.ts` fails on the `build:bin` link, same as yours
- `bunx tsc --noEmit -p packages/gemi/tsconfig.json` exit 0; `oxlint` on the three changed source files, 0 warnings
- `docs/llms-full.txt` mirrors `docs/orm.md` — enforced by `docs-scope.test.ts`, green

Follow-ups: **#262** (constrain `field` to the schema's columns, on all four spellings), **#263** (the documented `static delete` override).

PR #245, opened 2026-07-29
The section correcting a stale diagram had itself gone stale

### LINKED ISSUE #244: The section correcting a stale diagram has itself gone stale
`plans/orm/README.md` carries a section headed **"Where it actually is now"**, written for a stated reason:

> Recorded because the diagram above described the state on the day this plan was written and a reader following it would branch under work that has since landed on top.

It then asserts:

> **Every one of these is `MERGEABLE`/`CLEAN`** and each was verified at its own tip

of six PRs — #53, #54, #55, #56, #57, #58 — **all of which merged on 2026-07-28**.

So the correction went stale in exactly the way it was written to prevent, one section below the diagram it was correcting.

### Fix

Rewrite it as **history rather than status**: the six landed, on that date, in that order. The table is worth keeping — it is the only place recording what each PR added — but the claim about their *state* is not.

That framing is the fix, not a fresher number. Updating "MERGEABLE" to "merged" puts the next reader in the same position the day something else changes. A date cannot rot; a merge state can, and this document has no way to notice when it does.

### No guard, and why

The obvious local signal is whether each branch is an ancestor of `feat/orm`. It does not work here:

```
feat/orm-08-eloquent-doorway   exists, NOT merged into feat/orm
feat/orm-09-lateral-strategy   exists, NOT merged into feat/orm
fix/orm-bytes-container        exists, NOT merged into feat/orm
docs/orm                       exists, NOT merged into feat/orm
feat/orm-registration-audit    exists, merged into feat/orm
feat/orm-relation-filters      exists, merged into feat/orm
```

Four of the six merged into their own parents rather than into the trunk — which is the stacked-PR behaviour this same document explains two sections further down. A guard built on that signal would be wrong four times out of six.

A guard that asks GitHub is one that fails for reasons unrelated to the code, which #235 already settled. The structural fix — a claim that is a date rather than a state — needs no guard.

### Also

The diagram above annotates `feat/orm` as "iterations 1-6"; it carries iterations 1-9 and everything since. #30, #33 and #45 are genuinely still open, so the rest of that diagram is accurate.

nstfkc:
Closes #244.

`plans/orm/README.md` carries a section headed **"Where it actually is now"**, written for a stated reason:

> Recorded because the diagram above described the state on the day this plan was written and a reader following it would branch under work that has since landed on top.

It then says **"Every one of these is `MERGEABLE`/`CLEAN`"** of six PRs — #53 through #58 — all of which merged on 2026-07-28. The correction went stale in exactly the way it was written to prevent, one section below the diagram it was correcting.

Rewritten as **history rather than status**: the six landed, on that date, in that order. The table is worth keeping — it is the only place recording what each PR added — but the claim about their state is not.

That framing is the fix rather than a fresher number. Updating "MERGEABLE" to "merged" would put the next reader in the same position the day something else changes. A date cannot rot; a merge state can, and this document has no way to notice when it does.

## Stack order, named as such

The table's rows are `#53, #54, #56, #55, #57, #58`, which is the base chain, not the merge clock — `#55` (`docs/orm`) merged at `08:11:11`, seventeen seconds before its own base `#56` (`fix/orm-bytes-container`) at `08:11:28`. The original text called it merge order and this PR re-asserted that while re-checking only the date beside it; review caught it.

Relabelled rather than reordered, and the inversion is now named where it happens. Stack order is what a reader deciding where to branch wants, and that one child-before-parent landing is a concrete instance of the pooling the "had the Bytes fix / had the docs / had the audit" table two sections down records.

## No guard, deliberately

The naive local signal — whether each branch is an ancestor of `feat/orm` — does not work here:

```
feat/orm-08-eloquent-doorway   NOT merged into feat/orm
feat/orm-09-lateral-strategy   NOT merged into feat/orm
fix/orm-bytes-container        NOT merged into feat/orm
docs/orm                       NOT merged into feat/orm
feat/orm-registration-audit    merged into feat/orm
feat/orm-relation-filters      merged into feat/orm
```

Four of the six merged into their **own parents** rather than into the trunk — the stacked-PR behaviour this same document explains two sections down.

But that is a fact about *that* signal, not about local signals in general, and an earlier revision of this description overstated it into "I would rather ship no guard than one that is confidently wrong". Review pointed out the correction: `git rev-list --no-merges --count origin/feat/orm..origin/<branch>` returns `0` for all four, because every commit they hold over the trunk is a merge commit. It is right six times out of six — and it is the runbook this very file prescribes about forty lines below the edited section. A guard *could* be built locally.

The reason there is none is narrower: a claim that is a date rather than a state has nothing to guard. Building the check anyway is a change to what this document verifies about itself, which is not what #244 asks for, so it stays out of a docs-only PR. (A guard that asks GitHub fails for reasons unrelated to the code, which #235 already settled.)

## Also corrected

The diagram above annotates `feat/orm` as "iterations 1-6"; it carries iterations 1-9 and everything since. #30, #33 and #45 are genuinely still open, so the rest of that diagram is accurate and untouched.

`packages/gemi/orm/plan-record.test.ts` described this document as having a "Where it actually is now" section — the heading this PR renames. The string is wrapped across two lines, so nothing asserts on it and the suite stayed green with the reference wrong; a PR whose thesis is that this file contains claims which rot should not leave one behind. The neighbouring `line 496` citation went with it, for the same reason.

## Checks

- `bun --bun vitest run orm database` — 57 files, **1499 passed** (the plan is read by `plan-record.test.ts` from #198 and `operation-count.test.ts` from #196; neither claim is affected). Worth stating plainly, as review did: the suite is insensitive to this edit, so this is evidence of no breakage, not evidence the correction is correct. The correctness evidence is the `mergedAt` table.
- `bunx oxlint orm --deny-warnings` — 0 warnings

Documentation only; independent of #237, #239, #241 and #243.

### DIFF
--- a/packages/gemi/orm/plan-record.test.ts
+++ b/packages/gemi/orm/plan-record.test.ts
@@ -10,8 +10,8 @@ import * as orm from "./index";
  *
  * `plans/orm/README.md` is the ORM's design record — the document the invariants
  * come from, cited by name from `strategy.ts`, `lateral.ts`, `nested-writes.ts`
- * and `order-relation.ts`. It is a living document: it has a "Where it actually
- * is now" section, and line 496 reads "Both are now in all four".
+ * and `order-relation.ts`. It is a living document: it has a "How the stack
+ * landed" section, and a later line reads "Both are now in all four".
  *
  * Two of its invariants still described shipped API as forthcoming:
  *
--- a/plans/orm/README.md
+++ b/plans/orm/README.md
@@ -27,12 +27,20 @@ Nine levels, not three. Recorded because the diagram above described the state o
 the day this plan was written and a reader following it would branch under work
 that has since landed on top.
 
-**All six merged on 2026-07-28**, in the order below. This is written as history
-rather than as status on purpose: the section it replaces said "**Every one of
-these is `MERGEABLE`/`CLEAN`**", which was true when measured and false a day
-later — the same way the diagram above it went stale, in the section written to
-correct that diagram. A date cannot rot; a merge state can, and this document
-has no way to notice.
+**All six merged on 2026-07-28**, listed below in **stack order** — each row's
+branch is the base of the row under it, except #58, which targeted `feat/orm`
+directly. That is not quite merge order: #55 (`docs/orm`) merged at `08:11:11`,
+seventeen seconds *before* its own base #56 (`fix/orm-bytes-container`) at
+`08:11:28`. It is the one place here a child landed before its parent went
+upward, and it is a concrete instance of the pooling described two sections
+down. Stack order is the more useful ordering for a reader deciding where to
+branch, so the table keeps it and names it.
+
+This is written as history rather than as status on purpose: the section it
+replaces said "**Every one of these is `MERGEABLE`/`CLEAN`**", which was true
+when measured and false a day later — the same way the diagram above it went
+stale, in the section written to correct that diagram. A date cannot rot; a
+merge state can, and this document has no way to notice.
 
 | PR | Branch | What it adds |
 | --- | --- | --- |
--- a/plans/orm/README.md
+++ b/plans/orm/README.md
@@ -18,15 +18,21 @@ This work is a **third level in an open PR stack**. Branch from
 main
  └── refactor/laravel-container-architecture   PR #30, OPEN — container/providers
       └── feat/database-layer                  PR #33, OPEN — DatabaseManager, DB facade
-           └── feat/orm                        PR #45, OPEN — iterations 1-6
+           └── feat/orm                        PR #45, OPEN — iterations 1-9 and after
 ```
 
-### Where it actually is now
+### How the stack landed
 
 Nine levels, not three. Recorded because the diagram above described the state on
 the day this plan was written and a reader following it would branch under work
-that has since landed on top. **Every one of these is `MERGEABLE`/`CLEAN`** and
-each was verified at its own tip; they are listed in merge order.
+that has since landed on top.
+
+**All six merged on 2026-07-28**, in the order below. This is written as history
+rather than as status on purpose: the section it replaces said "**Every one of
+these is `MERGEABLE`/`CLEAN`**", which was true when measured and false a day
+later — the same way the diagram above it went stale, in the section written to
+correct that diagram. A date cannot rot; a merge state can, and this document
+has no way to notice.
 
 | PR | Branch | What it adds |
 | --- | --- | --- |


nstfkc:
## Review — the central claim checks out, but the sentence that replaced it still misstates the merge order

Docs-only, 10/4 on one file. I verified the claim against GitHub and git rather than against the description, and reproduced both stated checks.

### The fix does what it says

`"Every one of these is MERGEABLE/CLEAN"` was indeed false. All six are `MERGED`, and all six merged on 2026-07-28, so `**All six merged on 2026-07-28**` is accurate:

```
PR  mergedAt (UTC)        head                          base
53  2026-07-28T08:10:03Z  feat/orm-08-eloquent-doorway  feat/orm-07-performance
54  2026-07-28T08:10:54Z  feat/orm-09-lateral-strategy  feat/orm-08-eloquent-doorway
55  2026-07-28T08:11:11Z  docs/orm                      fix/orm-bytes-container
56  2026-07-28T08:11:28Z  fix/orm-bytes-container       feat/orm-09-lateral-strategy
57  2026-07-28T08:11:41Z  feat/orm-registration-audit   docs/orm
58  2026-07-28T09:19:33Z  feat/orm-relation-filters     feat/orm
```

`#30`, `#33`, `#45` are all still `OPEN`, so leaving the rest of the diagram alone is right. `"Five of those merges landed within 100 seconds"` two sections down survives the same check — #53 to #57 is 98 seconds. And `iterations 1-9 and after` is right: `plans/orm/` holds `01-` through `09-`, and `git diff origin/feat/orm origin/feat/orm-09-lateral-strategy` is 149 files / 32,794 deletions in the direction of `feat/orm`, i.e. the trunk carries far more than iteration 9.

### The one thing the rewrite got wrong

> **All six merged on 2026-07-28**, in the order below.

The order below is `#53, #54, #56, #55, #57, #58`. That is **stack order**, not merge order. Per the timestamps above, `#55` merged at `08:11:11` and `#56` at `08:11:28` — `#55` landed 17 seconds **first**. The table has them swapped.

This is inherited, not introduced (the old text said "they are listed in merge order"), but this PR is specifically a re-audit of that sentence and re-asserted the ordering claim while checking only the date beside it — from the same `gh pr view --json mergedAt` that produced the date. Either reorder the two rows or say "in stack order", which is what the table actually is and which is also the more useful ordering for a reader.

The swap is not a typo, incidentally — it is the one place a child merged into its parent before the parent merged upward, which is exactly the pooling the "had the Bytes fix / had the docs / had the audit" table below documents. Worth getting right for that reason.

### A cross-reference the rename breaks

`packages/gemi/orm/plan-record.test.ts:13`, present tense, on a test that reads this exact file:

```
 * and `order-relation.ts`. It is a living document: it has a "Where it actually
 * is now" section, and line 496 reads "Both are now in all four".
```

The PR renames that heading to `How the stack landed`. Nothing asserts on it — the string is wrapped across two lines, which is why a single-line grep for the heading comes back empty — so the suite stays green and the reference stays wrong. A PR whose thesis is that this document contains claims which rot should not leave a fresh one behind in a test's docblock. One-line fix.

(`operation-count.test.ts:19-20` cites `plans/orm/README.md:218` and `:491`; the +6 lines here shift those to 224 and 497. That block is explicitly historical narrative about #182, so I would leave it.)

### "No guard, deliberately" — the reasoning is half right

The naive signal is wrong exactly as claimed. `git merge-base --is-ancestor <branch> origin/feat/orm` reports four of six as unmerged:

```
feat/orm-08-eloquent-doorway   1 ahead of feat/orm
feat/orm-09-lateral-strategy   2 ahead
fix/orm-bytes-container        1 ahead
docs/orm                       1 ahead
feat/orm-registration-audit    0 ahead   (ancestor)
feat/orm-relation-filters      0 ahead   (ancestor)
```

But `--no-merges` is right six times out of six:

```
git rev-list --no-merges --count origin/feat/orm..origin/feat/orm-08-eloquent-doorway  -> 0
                                              ..origin/feat/orm-09-lateral-strategy    -> 0
                                              ..origin/fix/orm-bytes-container         -> 0
                                              ..origin/docs/orm                        -> 0
```

Every commit those four hold over the trunk is a merge commit (`c616fb9`, `dc0c4be` on lateral, for instance). Their content landed. And this is not an outside idea — it is the runbook in this same file, ~40 lines below the section being edited:

> `git rev-list --no-merges --count origin/<stack-tip>..origin/<each-branch>` … Zero everywhere means no branch holds unique work

So "a guard built on that signal would be wrong four times out of six, and I would rather ship no guard than one that is confidently wrong" is true of the signal the PR tried, and false of the one the document itself prescribes. I do not think a guard is required to land this — history framing genuinely does not need one, and I would not block on it — but the stated justification should not claim the local signal cannot work when the file next to it explains how.

### Checks

Both claimed checks reproduce exactly, from `packages/gemi`:

```
bun --bun vitest run orm database
  Test Files  57 passed (57)
  Tests       1499 passed (1499)

bunx oxlint orm --deny-warnings
  Found 0 warnings and 0 errors.   (96 files, 93 rules)
```

I also ran the suite with the **base** `plans/orm/README.md` restored: `57 passed / 1499 passed`, byte-identical result. That is the expected outcome for a docs change, and it is worth saying plainly — the suite is completely insensitive to this edit, so "1499 passed" is evidence of no breakage and not evidence the correction is correct. The correctness evidence is the `mergedAt` table at the top.

No tests are added, and none are warranted. `plan-record.test.ts` only sweeps for the `a future \`X\`` phrasing and is unaffected by the rename; no `#where-it-actually-is-now` anchor exists anywhere in `plans/`, `docs/` or `packages/`, so nothing links to the old heading.

### Verdict

Land it after the two small things: swap the `#55`/`#56` rows (or relabel the ordering), and update `plan-record.test.ts:13`. The date-not-status reframing is the right call and I would not water it down.

nstfkc:
## Addressed — all three fixed in `ebf487b`, plus the description claim you flagged

Thanks for reproducing the `mergedAt` table; the ordering catch is right, and I had re-asserted the ordering while checking only the date beside it, exactly as you describe.

**1. `#55`/`#56` — the table was in stack order, not merge order.** Fixed in `ebf487b`. Relabelled rather than reordered, taking the option you preferred: the rows are the base chain, and the sentence now says so, names the exception (`#58` targeted `feat/orm` directly), and records the inversion where it happens.

> That is not quite merge order: #55 (`docs/orm`) merged at `08:11:11`, seventeen seconds *before* its own base #56 (`fix/orm-bytes-container`) at `08:11:28`. It is the one place here a child landed before its parent went upward, and it is a concrete instance of the pooling described two sections down.

Your point that the swap is *not* a typo but the one child-before-parent landing is why it is called out in prose rather than quietly reordered — it is the same phenomenon the "had the Bytes fix / had the docs / had the audit" table documents, and it is more useful visible than smoothed over.

**2. `plan-record.test.ts:13` — cross-reference broken by the rename.** Fixed in `ebf487b`. It now reads `"How the stack landed"`. I took the neighbouring `line 496` with it: that citation is `510` after this PR and would move again on the next edit, and a line number in a docblock is the same species of claim this PR exists to delete. It cites the quoted phrase instead, which identifies the line without rotting.

```
 * and `order-relation.ts`. It is a living document: it has a "How the stack
 * landed" section, and a later line reads "Both are now in all four".
```

Left `operation-count.test.ts:19-20` alone as you suggested. Its `:218`/`:491` citations are now off by 14 rather than 6, but that block is describing a past state of the file, and re-pinning it to today's line numbers would be the wrong repair.

**3. "No guard, deliberately" — the justification, not the decision.** Fixed by rewriting that section of the PR description; no commit, because the claim lives only there and never entered the file. You are right and I reproduced it: `git rev-list --no-merges --count origin/feat/orm..origin/<branch>` returns `0` for all four branches the ancestor test misreports, so that signal is right six times out of six — and it is the runbook this same file prescribes forty lines below the section I edited. The description no longer claims a local guard cannot work. What it claims now is narrower: a claim that is a date has nothing to guard, and building the check anyway changes what this document verifies about itself, which is outside what #244 asks of a docs-only PR.

I did not open a follow-up issue for the guard, since you said you would not block on one and I do not think a date-framed table warrants one. Say the word and I will file it.

**Checks**, re-run from `packages/gemi` after the change:

```
bun --bun vitest run orm database
  Test Files  57 passed (57)
  Tests       1499 passed (1499)

bunx oxlint orm --deny-warnings
  Found 0 warnings and 0 errors.   (96 files, 93 rules)
```

Noted on your base-restored run, and I have folded the point into the description: 1499 passing is evidence of no breakage, not evidence the correction is correct. The `mergedAt` table is the evidence.

PR #247, opened 2026-07-29
The comment justifying -t postgres cited numbers that had drifted

### LINKED ISSUE #246: The comment justifying -t postgres cites numbers that have drifted
The Postgres job explains its `-t postgres` filter with a measurement:

> Run unfiltered, this job reports "877 passed" and still exits 1, because **three suites error at collection**.

Both halves are stale. Measured again against Postgres 16 on the current tree:

```
Test Files  2 failed | 18 passed (20)
Tests       911 passed | 395 skipped (1306)
```

**911 passed, and two suites fail** — not 877 and three.

### The reasoning is intact, and worth having checked

I ran the unfiltered job to find out whether the filter is needed *at all*, not to correct the numbers. `-t postgres` is what made three suites invisible in #205, and dropping it would have prevented that — so it is worth knowing whether it still earns its place.

It does. With `TEST_POSTGRES_URL` set, the harness makes the SQLite half of each differential suite **fail** rather than skip, deliberately, so a client generated for the wrong provider cannot look like a pass. An unfiltered run therefore exits 1 however many tests pass.

### Fix

State the mechanism, not the count. Neither number was ever the reason — they were evidence, gathered once, and evidence that goes stale in a comment reads as fact. Same fix as #244/#245 one line down: say the thing that cannot rot.

While there: the claim beside it — *"nothing is left uncovered between them"* — should point at the guard that makes it true. It was **false for three files** until #206, which is worth recording next to a sentence asserting it.

nstfkc:
Closes #246.

The Postgres job explains its filter with a measurement:

> Run unfiltered, this job reports "877 passed" and still exits 1, because **three suites error at collection**.

Both halves are stale. Measured again against Postgres 16 on the current tree: **911 passed, and two suites fail** — not 877 and three.

## The reasoning is intact, and that is what I set out to check

I ran the unfiltered job to find out whether the filter is needed **at all**, not to correct its numbers. `-t postgres` is what made three suites invisible in #205, and dropping it would have prevented that — so it is worth knowing whether it still earns its place.

It does. `@prisma/client` speaks one provider at a time, and the `Generate for Postgres` step above generates it for Postgres — so the SQLite half of each differential suite *fails* rather than skips, deliberately, so a client generated for the wrong provider cannot look like a pass. An unfiltered run exits 1 however many tests pass.

## The fix is the framing

Stated as the mechanism rather than the count, and **no numbers survive in the comment** — not the stale pair, and not the fresh pair either. `911` and `two` have exactly the shelf life `877` and `three` had; the next suite added to `app/models` moves them. Neither number was ever the reason, so the record of them belongs here, which is permanent, and not in the comment, which has to stay true.

The mechanism is attributed to the **generate step**, not to `TEST_POSTGRES_URL`. The env var only decides whether the Postgres half runs at all: `differential.ts:188` asks for the provider implied by the URL it was handed, and `assertPrismaSpeaks` throws when the generated client disagrees. Measured — with `TEST_POSTGRES_URL` set and the client generated for sqlite, the SQLite half passes and the *Postgres* half fails. Naming the step keeps the comment true if the variable ever moves.

The claim beside it — *"nothing is left uncovered between them"* — now points at the guard that makes it true. It was **false for three files** until #206, which is worth recording next to a sentence that asserts it.

## Checks

- unfiltered Postgres run, to establish the mechanism still holds — 2 suites fail, exits 1
- the same run with the client generated for sqlite instead — the Postgres half fails, the SQLite half passes; this is what fixed the attribution
- CI's unit step verbatim, the exclusion list rather than a path filter — 65 files, **1554 passed**, 1 skipped
- `bunx oxlint orm --deny-warnings` — 0 warnings
- workflow parses; `git diff feat/orm -- .github/workflows/ci.yml` is **0 non-comment lines**
- scratch container removed, template restored to SQLite

## On tests

None are added and none are possible: this is prose, and prose is what the workflow's guards deliberately cannot see. `ci-coverage.test.ts` strips comment lines before asserting — a guard that could not tell a command from a sentence about a command would be satisfied by deleting the explanation — and `postgres-suite-selection.test.ts` reads the `-t` filter off the command line. Both pass (15 tests), but neither covers this diff, and an earlier revision of this description wrongly listed them as if they did.

`ci.yml:17-20` still says Postgres is not in CI, 160 lines above the job that runs it. Separate claim, separate paragraph, filed as #261.

Comment-only change to `ci.yml`; independent of #237, #239, #241, #243 and #245.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,14 +240,19 @@ jobs:
 
       # `-t postgres` selects the dialect-labelled suites, and it is required
       # rather than tidy. With `TEST_POSTGRES_URL` set, the harness makes the
-      # *SQLite* half **fail** rather than skip — deliberately, so a client
-      # generated for the wrong provider cannot look like a pass. Run unfiltered,
-      # this job reports "877 passed" and still exits 1, because three suites
-      # error at collection.
+      # *SQLite* half of each differential suite **fail** rather than skip —
+      # deliberately, so a client generated for the wrong provider cannot look
+      # like a pass. So an unfiltered run here exits 1 no matter how many tests
+      # pass, and the failures are the ones the setup above guarantees.
+      #
+      # Stated as the mechanism rather than as a count. It read "877 passed …
+      # three suites error at collection", measured once and true then; the same
+      # run is now 911 and two, and neither number was ever the reason.
       #
       # So this job runs the Postgres-labelled suites and the `sqlite` job runs
       # the rest. The dialect-agnostic tests are that job's business; nothing is
-      # left uncovered between them.
+      # left uncovered between them — which was false for three files until
+      # #206, and is guarded now by `postgres-suite-selection.test.ts`.
       #
       # Serially: the harness truncates shared tables between cases, so parallel
       # files clobber each other's fixtures. Measured — the parallel run fails
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,15 +239,14 @@ jobs:
         working-directory: templates/saas-starter
 
       # `-t postgres` selects the dialect-labelled suites, and it is required
-      # rather than tidy. With `TEST_POSTGRES_URL` set, the harness makes the
-      # *SQLite* half of each differential suite **fail** rather than skip —
-      # deliberately, so a client generated for the wrong provider cannot look
-      # like a pass. So an unfiltered run here exits 1 no matter how many tests
-      # pass, and the failures are the ones the setup above guarantees.
-      #
-      # Stated as the mechanism rather than as a count. It read "877 passed …
-      # three suites error at collection", measured once and true then; the same
-      # run is now 911 and two, and neither number was ever the reason.
+      # rather than tidy. `@prisma/client` speaks one provider at a time, and
+      # `Generate for Postgres` above makes it Postgres — so the *SQLite* half
+      # of each differential suite **fails** rather than skips, deliberately,
+      # so a client generated for the wrong provider cannot look like a pass.
+      # `TEST_POSTGRES_URL` decides only whether the Postgres half runs at all;
+      # the generate step is what decides which half dies. So an unfiltered run
+      # here exits 1 no matter how many tests pass, and the failures are the
+      # ones the setup above guarantees.
       #
       # So this job runs the Postgres-labelled suites and the `sqlite` job runs
       # the rest. The dialect-agnostic tests are that job's business; nothing is


nstfkc:
## Review — the measurement is exact and the change is risk-free, but the fix is half-done: it adds two fresh numbers instead of removing the rotting ones, and the one mechanism sentence it rewrote still names the wrong cause

I reproduced the run rather than taking the body's word for it. Postgres 16 container, schema flipped to `postgresql`, `prisma generate`, `prisma db push`, then the job's command with `-t postgres` dropped:

```
 FAIL  app/models/writes.differential.test.ts > differential writes vs prisma — sqlite
 FAIL  app/models/differential.test.ts > differential vs prisma — sqlite
 Test Files  2 failed | 18 passed (20)
      Tests  911 passed | 395 skipped (1306)
EXIT=1
```

**911 and two, exactly as claimed.** Both failures are `assertPrismaSpeaks` at `differential.ts:494`, reached through `createDifferential` — so the mechanism the comment now states is the real one, and the old text was stale in both halves. It was also wrong in a third way the rewrite quietly fixes: those are `beforeAll` failures inside the harness, not "suites error at collection".

The job's actual command is still green:

```
bun --bun vitest run app/models --no-file-parallelism -t postgres
 Test Files  13 passed | 7 skipped (20)
      Tests  669 passed | 637 skipped (1306)      EXIT=0
```

Diff is **0 non-comment lines** (`git diff feat/orm...HEAD | grep -vE '^[+-]\s*#'` → 0), so nothing can break. Confirmed anyway: CI's unit step 65 files / 1554 passed, `bunx oxlint orm --deny-warnings` 0 warnings, `ci-coverage.test.ts` + `postgres-suite-selection.test.ts` 15 passed. `#206` checks out — it added `postgres-suite-selection.test.ts` and fixed the three files, so that new cross-reference is accurate and is the best line in the diff.

Four things.

**1. The stated fix is "state the mechanism, not the count", and the diff does not do the second half.** Before the change the comment held two numbers (877, three). After it holds four: it quotes the stale pair *and* records a fresh pair. `911` and `two` have exactly the shelf life `877` and `three` had — the next suite added to `app/models` moves them, and then the comment is misleading in the same way, one paragraph longer. The mechanism sentence is a genuine improvement and stands on its own; the paragraph under it ("It read \"877 passed …\" … the same run is now 911 and two") loses nothing by being deleted. #246 says it itself: *neither number was ever the reason*. Keeping both pairs is archaeology, and the place for it is this PR, which is permanent, rather than the comment, which has to stay true.

**2. The mechanism sentence names the wrong trigger — and it is the one sentence the diff rewrote.** "With `TEST_POSTGRES_URL` set, the harness makes the *SQLite* half of each differential suite **fail** rather than skip". The env var is not what does that. `differential.ts:188` picks the expected provider from the URL the harness was handed, and `assertPrismaSpeaks` throws only when the *generated client* disagrees:

```ts
await assertPrismaSpeaks(prisma, options.url ? "postgresql" : "sqlite");
```

Measured, with `TEST_POSTGRES_URL` set and the client regenerated for sqlite:

```
 FAIL  app/models/differential.test.ts > differential vs prisma — postgres
Error: This suite needs @prisma/client generated for the 'postgresql' provider, and it is not.
 Test Files  1 failed (1)
      Tests  252 passed | 317 skipped (569)
```

The SQLite half passes. It is the **Postgres** half that fails. What decides which half dies is the `Generate for Postgres` step above, not the env var — the env var only decides whether the Postgres half runs at all. The conclusion survives (either way an unfiltered run exits 1, which is all the filter's justification needs), and the trailing clause "the failures are the ones the setup above guarantees" gestures at the right cause. But the leading clause attributes it to the wrong one, in a comment being rewritten for accuracy. Attributing it to the generate step instead would also make the comment robust to the env var moving.

**3. Same defect, worse, 220 lines up, untouched.** `ci.yml:17-20`:

> Postgres is deliberately not here yet. […] Adding a service container is a sensible next step and a separate decision.

There is a `postgres:` job with a `postgres:16` service container at line 178 of the same file. That is not drift, it is false, and it is the first thing a reader of this workflow meets. If the series is walking this file for stale prose, this one outranks a count that was off by 34.

**4. The checks in the body are not the CI steps.** `bun --bun vitest run orm database` is the inclusive path filter CI abandoned, and `ci-coverage.test.ts` asserts against its return:

```ts
expect(commands, "the unit step is back to a path filter, which skips new files silently")
  .not.toContain("vitest run orm database");
```

The real step runs an exclusion list — 65 files / 1554 tests here, not the 57 / 1499 reported. Green either way, so nothing is hiding. But "57 files, 1499 passed" is one more number measured off a command nobody runs, in a PR about numbers measured once.

**On the tests.** No new ones, and none are possible — this is prose, and prose is what the guards deliberately cannot see. Both named in the body pass, and both are real tests of the workflow, but neither covers this diff: `ci-coverage.test.ts` strips comment lines before asserting (by design — "a guard that cannot tell a command from a sentence about a command would be satisfied by deleting the explanation"), and `postgres-suite-selection.test.ts` reads the `-t` filter off the command line. Worth not describing them as coverage for this change.

**Verdict.** Land it after deleting the numbers paragraph and fixing the attribution in (2) — both are edits to lines this diff already touches. (3) and (4) are separate.

nstfkc:
## Addressed — both comment edits landed in e55e600, the header paragraph is #261, and the body's checks are corrected

**1. The numbers paragraph — fixed** in `e55e600`. Deleted, both pairs. You are right that the diff did half its own stated job: it replaced one count with an account of two counts, and `911`/`two` rot on exactly the schedule `877`/`three` did. The comment is now the mechanism and nothing else; the archaeology is in #246 and in this PR's description, which do not have to stay true about `app/models`.

**2. The attribution — fixed** in `e55e600`. You are right, and the measurement is the proof: with `TEST_POSTGRES_URL` set and the client generated for sqlite, the SQLite half *passes*. `differential.ts:188` asks for the provider implied by the URL it was handed, and `assertPrismaSpeaks` throws when the generated client disagrees — so the env var only decides whether the Postgres half runs, and `Generate for Postgres` decides which half dies. The comment now names the step:

```
      # `-t postgres` selects the dialect-labelled suites, and it is required
      # rather than tidy. `@prisma/client` speaks one provider at a time, and
      # `Generate for Postgres` above makes it Postgres — so the *SQLite* half
      # of each differential suite **fails** rather than skips, deliberately,
      # so a client generated for the wrong provider cannot look like a pass.
      # `TEST_POSTGRES_URL` decides only whether the Postgres half runs at all;
      # the generate step is what decides which half dies. So an unfiltered run
      # here exits 1 no matter how many tests pass, and the failures are the
      # ones the setup above guarantees.
```

Conclusion unchanged: either way an unfiltered run exits 1, which is all the filter needs. And as you note, naming the step keeps the sentence true if the variable moves.

**3. `ci.yml:17-20` — deferred to #261.** Agreed on the substance, and agreed it outranks a count off by 34: it is false rather than stale, it is about the shape of the workflow rather than a number inside it, and it is the first thing a reader meets. Not folded in here because this PR closes #246, which is specifically the comment inside the `postgres:` job — a different paragraph, a different claim, 220 lines away, and you called it separate too. #261 has the replacement text worked out, including the half of the paragraph that is still true (the harness does skip loudly without `TEST_POSTGRES_URL`, which is what a local run does).

**4. The body's checks — fixed in the description**, not in the tree, since that is where the defect was. Re-measured with CI's actual unit step — the exclusion list, not the abandoned `vitest run orm database` path filter — and it is **65 files / 1554 passed / 1 skipped**, replacing the 57 / 1499 measured off a command `ci-coverage.test.ts` exists to assert nobody runs. Quoting a number from that command in a PR about numbers was the same mistake one level up. The description also no longer lists `ci-coverage.test.ts` and `postgres-suite-selection.test.ts` as coverage for this diff: both pass, both are real tests of this workflow, and neither can see a comment — the first strips comment lines by design, the second reads the `-t` filter off the command line. That is stated plainly now rather than implied by putting them under "Checks".

Still 0 non-comment lines against `feat/orm`. Guards green (15 passed), `bunx oxlint orm --deny-warnings` 0 warnings.

PR #249, opened 2026-07-29
Three branches of where.ts that nothing distinguished

### LINKED ISSUE #248: Mutation testing finds three unguarded branches in where.ts
Flipping every operator in `compile/where.ts` and running the unit suite kills **44 of 49** mutants. Five live. One is a floor; three are real.

### 1. The sentinel refusal does not name the sentinel it got

`where: { payload: Prisma.JsonNull }` is refused (#225), and the message names the sentinel **twice**, both through a ternary:

```ts
`Prisma.${sentinel === "db" ? "DbNull" : "JsonNull"} is not a filter on …`
`… { equals: Prisma.${sentinel === "db" ? "DbNull" : "JsonNull"} } }.`
```

Flipping either changes nothing. So a caller passing `JsonNull` can be told:

> `Prisma.DbNull` is not a filter on its own … Write it as `{ payload: { equals: Prisma.DbNull } }`

That is worse than no advice — following it writes SQL NULL where they meant a JSON null, and the two are different rows. The test that pinned the refusal never looked at *which* sentinel it named.

### 2. A relation filter given a scalar

`typeof value !== "object" || Array.isArray(value)` — the array half is covered, the scalar half is not. `where: { accounts: 42 }` reaching the relation planner as though it were a filter object is what the guard exists to stop.

### 3. An explicitly `undefined` operand

`operand === undefined || key === "mode"` — the `mode` half is covered. Nothing asserts Prisma's rule that `undefined` means "not supplied" wherever it appears, which is what lets a filter be built conditionally without deleting keys. Without it, `{ contains: undefined }` falls through to the unknown-operator check and becomes an error.

### The one that is not a defect

`!Array.isArray(fields) || fields.length === 0` sits under a comment saying *"these are its bugs rather than a caller's"* — a planner-internal assertion no caller can reach. A floor, in the sense #114 established, and left alone.

### Fix

Three tests, each written as the behaviour it protects rather than as "line N is uncovered" — a survivor is a question, and only three of these five turned out to be defects.

Worth re-running the mutants afterwards: a test that does not kill the mutant it was written for has not closed the gap.

nstfkc:
Closes #248.

Found by mutation rather than by reading: flipping every operator in `compile/where.ts` and running the unit suite killed **44 of 49**. Of the five that lived, one is a floor and three were genuine.

## The error message is the one that matters

`where: { payload: Prisma.JsonNull }` is refused, and the refusal names the sentinel **twice**, both times through a ternary. Flipping either changed nothing — so a caller passing `JsonNull` could be told:

> `Prisma.DbNull` is not a filter on its own … Write it as `{ payload: { equals: Prisma.DbNull } }`

Worse than no advice: following it writes SQL NULL where they meant a JSON null, and the two are different rows. That message is mine, from #225, and the test that pinned the refusal never looked at *which* sentinel it named.

## The other two

- **A relation filter given a scalar.** `where: { accounts: 42 }` was refused only by the half of the guard that catches arrays; the scalar half had no test.
- **An explicitly `undefined` operand.** Skipped by a condition whose other half (`key === "mode"`) was covered, so nothing asserted Prisma's rule that `undefined` means "not supplied" wherever it appears — the rule that lets a filter be built conditionally without deleting keys.

## The one that is not a defect

`!Array.isArray(fields) || fields.length === 0` sits under a comment saying *"these are its bugs rather than a caller's"* — planner-internal, unreachable from a caller. A floor in the sense #114 established, left alone and documented as such.

Each test is written as the behaviour it protects rather than as "line N is uncovered", because a survivor is a question and only three of five turned out to be defects.

## Verified by re-running the mutants

A test that does not kill the mutant it was written for has not closed the gap:

```
line  327 || -> &&   KILLED
line  693 === -> !==  KILLED
line  736 || -> &&   KILLED
```

## Checks

- `bun --bun vitest run orm database` — 58 files, **1503 passed**

One new test file; independent of #237, #239, #241, #243, #245 and #247.

### DIFF
--- a/packages/gemi/orm/compile/where-mutants.test.ts
+++ b/packages/gemi/orm/compile/where-mutants.test.ts
@@ -1,0 +1,103 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { SqliteDialect } from "../dialect/sqlite";
+import { InvalidArgumentError } from "../errors";
+import { account, mapped, user } from "../fixtures";
+import * as registry from "../registry";
+import { compileRead } from "./read";
+
+/**
+ * Three branches of `where.ts` that nothing distinguished, found by mutation.
+ *
+ * Flipping each operator in the file and running the unit suite killed 44 of 49
+ * mutants. Of the five that lived, one is a floor by the file's own account —
+ * `the planner built a 'compositeIn' with no fields` sits under a comment saying
+ * "these are its bugs rather than a caller's", so no caller can reach it — and
+ * the other three were genuine: a guard, a skip, and an error message, each
+ * doing something no assertion cared about.
+ *
+ * A survivor is a question rather than a defect, so each is written here as the
+ * behaviour it protects rather than as "line N is uncovered".
+ */
+const sqlite = new SqliteDialect();
+
+const text = (args: unknown) =>
+  compileRead(user, "findMany" as never, args as never, sqlite).text;
+
+/** `AuditLog.payload` is the fixtures' only `Json` column. */
+const json = (args: unknown) =>
+  compileRead(mapped, "findMany" as never, args as never, sqlite).text;
+
+class DbNull {
+  toString() {
+    return "Prisma.DbNull";
+  }
+}
+
+class JsonNull {
+  toString() {
+    return "Prisma.JsonNull";
+  }
+}
+
+describe("branches of where.ts that mutation found unguarded", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    registry.register("User", class { static $schema = user });
+    registry.register("Account", class { static $schema = account });
+    registry.register("AuditLog", class { static $schema = mapped });
+  });
+
+  /**
+   * `typeof value !== "object" || Array.isArray(value)` — the array half was
+   * covered and the scalar half was not, so `!==` could become `&&` and a
+   * number would reach the relation planner as if it were a filter object.
+   */
+  test("a relation filter given a scalar is refused, not planned", () => {
+    expect(() => text({ where: { accounts: 42 } })).toThrow(InvalidArgumentError);
+    expect(() => text({ where: { accounts: "some" } })).toThrow(InvalidArgumentError);
+  });
+
+  /**
+   * `operand === undefined || key === "mode"` — the `mode` half was covered.
+   * Without the other, an explicitly `undefined` operand stops being skipped
+   * and falls through to the unknown-operator check, turning a filter Prisma
+   * treats as absent into an error.
+   *
+   * Prisma's rule is that `undefined` means "not supplied" wherever it appears,
+   * which is what lets a caller build a filter conditionally without deleting
+   * keys.
+   */
+  test("an explicitly undefined operand is absent, not an error", () => {
+    expect(() => text({ where: { email: { contains: undefined } } })).not.toThrow();
+
+    // ...and contributes no SQL, which is the half that says it was skipped
+    // rather than compiled to something vacuous.
+    expect(text({ where: { email: { contains: undefined } } })).not.toContain("like");
+  });
+
+  /**
+   * The refusal for a bare `Json` sentinel names **which** sentinel it got, in
+   * both places the message spells one. Mutation flipped the two ternaries
+   * independently and nothing failed: a caller passing `JsonNull` could be told
+   * to write `equals: Prisma.DbNull`, which stores a different value — advice
+   * that is worse than no advice, since following it writes SQL NULL where the
+   * caller meant a JSON null.
+   */
+  test.each([
+    ["DbNull", new DbNull(), "JsonNull"],
+    ["JsonNull", new JsonNull(), "DbNull"],
+  ])("a bare %s is refused naming %s, not the other one", (name, value, other) => {
+    let message = "";
+    try {
+      json({ where: { payload: value } });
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the three tests do what they claim, and I reproduced every kill

Test-only diff (one new file, 103 lines). I re-ran the mutation experiment rather than take the table on trust, and ran each mutant twice: once against the branch, once with `where-mutants.test.ts` deleted, so "survives today, dies now" is measured rather than asserted.

### Mutants, re-run

`packages/gemi/orm/compile/where.ts`, `bun --bun vitest run orm database` (58 files):

| mutant | without the new file (1499 tests) | with it (1503 tests) |
| --- | --- | --- |
| L327 `typeof value !== "object" \|\| Array.isArray(value)` → `&&` | 1499 passed — **survives** | 1 failed — **killed** |
| L693 `sentinel === "db"` → `!==` | 1499 passed — **survives** | 2 failed — **killed** |
| L696 `sentinel === "db"` → `!==` | 1499 passed — **survives** | 2 failed — **killed** |
| L736 `operand === undefined \|\| key === "mode"` → `&&` | 1499 passed — **survives** | 1 failed — **killed** |
| L648 `!Array.isArray(fields) \|\| fields.length === 0` → `&&` (the floor) | survives | survives (1503 passed) |

So the claim holds in both directions: each branch really was undistinguished before, and each new test kills the mutant it was written for. The floor survives untouched, which is what the PR says it should.

The body's verification block lists three lines (327, 693, 736) and omits **696**, the second ternary — which the prose itself identifies as independently flippable. It is killed too (above), so the tests are ahead of the report, not behind it. Worth fixing the accounting while you're here: "five lived, one floor + three genuine" is four; the fifth is the second ternary of the same message defect.

### Checks

- `bun --bun vitest run orm database` → **58 files, 1503 passed**, matching the body exactly.
- New file alone: 4 passed (3 `test`s, one of them a 2-case `test.each`).
- `tsc --noEmit -p tsconfig.json` → exit 0, no output.
- `oxlint orm --deny-warnings` → 0 warnings, 0 errors, 97 files.
- Merge state `CLEAN` against `feat/orm`; branch is on top of base.
- Template suite: not verified — the worktree resolves `templates/saas-starter/node_modules/gemi` to a copy without the ORM exports. Environmental, not attributable to this diff.
- Note for anyone re-running: plain `vitest run orm database` fails 5 files on `import { SQL } from "bun"`. The `bun --bun` prefix in the body is load-bearing.

### The framing is stronger than the finding, and should be dialled back

There is no behaviour change here and none is needed. I compiled the real path:

```
where: { payload: <JsonNull sentinel> }
→ Invalid 'where.payload' (AuditLog.findMany). Prisma.JsonNull is not a filter on its own
  — Prisma rejects it here too. Write it as an explicit comparison:
  { payload: { equals: Prisma.JsonNull } }.
```

The message names the sentinel it got, correctly, both times. "A caller passing `JsonNull` **could** be told `DbNull`" is true only of the hypothetical mutant, and the body's opening — "the error message is the one that matters", "advice worse than no advice" — reads as a shipped defect. It is an untested branch, which is worth a test on its own terms; overselling it makes the next mutation PR harder to read at face value. Same for #248's headline "the sentinel refusal does not name the sentinel it got" — it does.

Also verified against the real detector: `jsonNullKind` recognises by `String(value)` behind a `constructor` own-`name` guard, so the file's hand-rolled `class DbNull { toString() { … } }` doubles do reach the sentinel branch — a plain object literal with a `toString` would too. The doubles are faithful today.

### Nits, none blocking

1. **`toHaveLength(2)` pins the sentence shape, not the behaviour.** `message.match(/Prisma\.JsonNull/g)` having exactly two hits is what makes both ternary mutants die, but it also means a legitimate rewording that names the sentinel once — or one that renders the second spelling from a variable — fails a test about *which* sentinel is named. A cheaper equivalent: assert the message contains `equals: Prisma.JsonNull` (the actionable half) and not `DbNull` anywhere. That kills L696 by meaning rather than by count.
2. **The scalar-relation test asserts only the error class.** `toThrow(InvalidArgumentError)` would be satisfied by any earlier `InvalidArgumentError` on that path. The mutant kill proves the intended line is the one reached today, but pinning the sentence (`Expected an object holding every, none, some.`) makes that durable rather than incidental.
3. **The sentinel doubles are local.** No shared fixture exists (I grepped — this file is the only `class DbNull` in the repo), so duplicating is fair. But they encode `json-null.ts`'s *detection strategy*, not Prisma's value: if recognition ever gains a second condition, these fakes keep passing while real `Prisma.DbNull` breaks. A `dbNull()` / `jsonNull()` export next to `jsonNullKind` — used by tests, defined beside the thing it mirrors — removes that drift.
4. **File named after the technique.** `where-mutants.test.ts` is "how these were found", and the file's own docblock argues against exactly that framing ("written as the behaviour it protects rather than as 'line N is uncovered'"). Two of the three are refusals and `compile/refusals.test.ts` is the existing home for "the refusals nothing reached"; the third is a skip. Merging them there, or naming this by behaviour, keeps the suite organised by what it protects.

### Verdict

Merge. The mutants are real, the kills reproduce, the floor is correctly left alone, and nothing regresses (1503/1503, tsc clean, lint clean). Fix the survivor accounting in the body (696 is missing), and consider softening the defect language — the code was right; only the tests were missing.

PR #251, opened 2026-07-29
A relation filter inside AND or OR had no scoping test

### LINKED ISSUE #250: A relation filter inside AND or OR has no scoping test
Mutation testing on `policy.ts`: 76 sites, **17 survive the unit suite**, **12 survive the template suite as well**. Most of those twelve are `x === undefined || x === null` guards whose halves cannot be told apart by any observable behaviour. One is different.

### The one that matters

`scopeRelationFilters` recurses into `AND` / `OR` / `NOT` arrays and shares structure on the way out — `changed ? out : where`, so an untouched tree comes back as the same object and the plan key does not move.

Inverting the line that sets `changed`:

```ts
if (scoped !== entry) changed = true;   //  !==  ->  ===
```

**survives both suites.** Under that mutant, an entry that *was* scoped leaves `changed` false, the function returns the original `where`, and the child's scope is dropped — a policy bypass for any relation filter written inside an `AND`, which is how anyone composes two conditions:

```ts
await User.findMany({
  where: { AND: [{ accounts: { some: {} } }, { email: { contains: "@" } }] },
})
```

Every existing test in `relation filters in where` puts the filter at the **top level**, where the array branch is never entered. The branch is reachable from ordinary code and has no assertion.

### Fix

Two tests — inside `AND`, and inside `OR` beside a plain condition it must leave alone. `OR` reaches the same branch by a different key, so covering only `AND` leaves half the combinator untested.

Re-run the mutant afterwards. A test that does not kill it has not closed the gap.

### The other eleven

Left alone deliberately. They are questions, not defects, and for most the honest answer is that flipping `||` to `&&` between `=== undefined` and `=== null` cannot change what any caller sees. Writing tests to kill those would be writing tests to satisfy a tool.

nstfkc:
Closes #250.

Mutation on `policy.ts`: 76 sites, **17 surviving the unit suite**, **12 surviving the template suite as well**. Most of those twelve are `x === undefined || x === null` guards whose halves cannot be told apart by any observable behaviour. One is not.

`scopeRelationFilters` recurses into `AND` / `OR` / `NOT` arrays and shares structure on the way out — `changed ? out : where`, so an untouched tree comes back as the same object and the plan key does not move. Inverting the line that sets `changed`:

```ts
if (scoped !== entry) changed = true;
```

survives both suites. Under that mutant an entry that **was** scoped leaves `changed` false, the function returns the original `where`, and the child's scope is dropped — a policy bypass for any relation filter written inside an `AND`, which is how anyone composes two conditions.

Every existing test in `relation filters in where` puts the filter at the **top level**, where the array branch is never entered. The branch is reachable from ordinary code and had no assertion.

## Two tests, and why two

Inside `AND`, and inside `OR` beside a plain condition it must leave alone. `OR` reaches the same branch by a different key, so covering only `AND` would leave half the combinator untested.

**Verified by re-running the mutant** — killed, by the first of them:

```
mutant 1345 !== -> ===: KILLED
    × a relation filter inside AND is scoped like one at the top level
```

## The other eleven survivors are left alone

Deliberately. They are questions, not defects, and for most the honest answer is that flipping `||` to `&&` between `=== undefined` and `=== null` cannot change what any caller sees. Writing tests to kill those would be writing tests to satisfy a tool — which is the opposite of what the exercise is for.

## Checks

- `bun --bun vitest run orm database` — 57 files, **1501 passed**

Adds two tests to an existing describe; independent of the other seven open PRs.

### DIFF
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -1054,6 +1054,49 @@ describe("relation filters in where", () => {
     });
   });
 
+  /**
+   * **Inside a boolean combinator**, which nothing reached.
+   *
+   * `scopeRelationFilters` recurses into `AND` / `OR` / `NOT` arrays and shares
+   * structure on the way out — `changed ? out : where`, so an untouched tree
+   * comes back as the same object and the plan key does not move. Every test
+   * above puts the relation filter at the top level, where that array branch is
+   * never entered.
+   *
+   * Mutation found it: inverting `if (scoped !== entry)` survives the unit suite
+   * *and* the template suite. Under that mutant an entry that **was** scoped
+   * leaves `changed` false, so the function returns the original `where` and the
+   * child's scope is dropped — a policy bypass for any relation filter written
+   * inside an `AND`, which is how anyone composes two conditions.
+   *
+   * Both spellings, because `OR` reaches the same branch by a different key and
+   * a guard that only covers `AND` would leave half the combinator untested.
+   */
+  test("a relation filter inside AND is scoped like one at the top level", () => {
+    expect(
+      scoped({ AND: [{ accounts: { some: { organizationRole: 0 } } }] }, {
+        Account: [tenant],
+      }),
+    ).toEqual({
+      AND: [
+        { accounts: { some: { organizationRole: 0, AND: [{ organizationId: 7 }] } } },
+      ],
+    });
+  });
+
+  test("and inside OR, beside a plain condition it must not disturb", () => {
+    expect(
+      scoped({ OR: [{ email: "a@b.c" }, { accounts: { some: {} } }] }, {
+        Account: [tenant],
+      }),
+    ).toEqual({
+      OR: [
+        { email: "a@b.c" },
+        { accounts: { some: { AND: [{ organizationId: 7 }] } } },
+      ],
+    });
+  });
+
   test("none is narrowed the same way", () => {
     expect(scoped({ accounts: { none: {} } }, { Account: [tenant] })).toEqual({
       accounts: { none: { AND: [{ organizationId: 7 }] } },


nstfkc:
## Review — one of the two tests is real and kills the mutant; the other is a reordered duplicate of a test the PR says does not exist

I reproduced the mutant and ran every suite the change touches. The `AND` test does exactly what it claims. The `OR` test does not, and the reason it does not is the same reason the gap existed in the first place.

### The gap is real, and the AND test closes it

Applied `packages/gemi/orm/policy.ts:1345` `!==` → `===` and ran the pre-PR unit suite:

```
# mutant + policy.test.ts at 9f28c51
bun --bun vitest run orm database
 Test Files  57 passed (57)
      Tests  1499 passed (1499)
```

Survives clean. Same mutant with this PR's test file:

```
 ❯ orm/policy.test.ts (107 tests | 1 failed)
   × a relation filter inside AND is scoped like one at the top level
 Test Files  1 failed | 56 passed (57)
      Tests  1 failed | 1500 passed (1501)
```

One test fails, and it is the `AND` one. That test earns its place.

### The `OR` test passes under the mutant

Note what the run above says: **1 failed**, not 2. `and inside OR, beside a plain condition it must not disturb` is green under the exact mutant this PR exists to kill.

The mechanism is worth stating because it is the whole point. Under `if (scoped === entry) changed = true`, an array sets `changed` from its *unchanged* entries. The new `OR` case is `[{ email: "a@b.c" }, { accounts: { some: {} } }]` — the plain condition comes back identical, sets `changed = true`, the rebuilt array with the scoped child is returned, and the assertion holds. Only an array in which *every* entry is rewritten leaves `changed` false. The `AND` case has exactly one entry, and it is rewritten; that is why it fires.

So the shipped `OR` shape is not merely weaker — it is the one arrangement structurally incapable of detecting this class of bug.

### The claim that no existing test entered the branch is false

Both the PR body and the new doc comment say:

> Every existing test in `relation filters in where` puts the filter at the **top level**, where the array branch is never entered.

`packages/gemi/orm/policy.test.ts:1160`, in that same describe, on the base commit:

```ts
test("combinators are walked", () => {
  expect(
    scoped(
      { OR: [{ accounts: { some: {} } }, { email: "a@b.c" }] },
      { Account: [tenant] },
    ),
  ).toEqual({
    OR: [
      { accounts: { some: { AND: [{ organizationId: 7 }] } } },
      { email: "a@b.c" },
    ],
  });
});
```

That is the new `OR` test with the two operands swapped. The array branch was reached; the entries were mixed, which is why it never caught the `changed` bug — and the new test reproduces the mixed shape rather than fixing it.

It also adds no kill of its own. Second mutant, dropping `|| key === "OR"` from `policy.ts:1360`:

```
   × and inside OR, beside a plain condition it must not disturb
   × combinators are walked
```

Both fail. The pre-existing test already covers that.

### Concrete fix for the second test

Make every entry of the `OR` scoped, so `changed` has nowhere to get a spurious `true`. I wrote this variant and ran it against the mutant:

```ts
scoped({ OR: [{ accounts: { some: {} } }, { organization: { name: "acme" } }] }, {
  Account: [tenant],
  Organization: [tenant],
})
```

```
   × a relation filter inside AND is scoped like one at the top level
   × TEMP variant: OR whose entries are all scoped
      Tests  2 failed | 106 passed (108)
```

Killed. That version guards the branch by the `OR` key instead of restating `combinators are walked`. The doc comment should be corrected along with it — it is the durable artifact, and right now it tells the next reader something about the existing suite that is not true.

### Nothing is broken

- Clean branch, full suite: `57 files, 1501 passed` — matches the body.
- `bunx tsc --noEmit -p tsconfig.json` → exit 0.
- `bunx oxlint orm --deny-warnings` → 0 warnings, 0 errors.
- Source is untouched; no behaviour change, no risk to the other open PRs.
- `oxfmt --check` flags `policy.test.ts`, but it flags the base version identically — pre-existing, not this PR.
- Template suite here: `2 failed | 17 passed | 3 skipped (22)`, `39 failed | 604 passed | 32 skipped (675)`. **Identical on base `9f28c51`**, so not attributable to this PR — the failures are `Model.aggregate` / `Model.groupBy` not existing on the model class in this branch. The body's "17 passed / 3 skipped, 640 tests" did not reproduce in my environment; worth a second look at where that number came from, since it is the evidence offered for the mutant surviving the template suite.

### Verdict

The `AND` test is a genuine, verified guard over a real policy-bypass path and I would take it as-is. The `OR` test should be reshaped as above or dropped, and the "every existing test puts the filter at the top level" line removed from the comment. Two tests were the right instinct; the second one just needs the shape that makes it a test rather than a restatement.

PR #253, opened 2026-07-29
Nothing asserted the planner picks the composite in

### LINKED ISSUE #252: Nothing asserts the planner picks the composite in on Postgres
Mutation testing on `plan-relations.ts`: 60 sites, **10 survive the unit suite**, **6 survive the template suite as well**. One of the six is a real gap.

The guard that decides whether a composite relation uses Postgres' `unnest`:

```ts
types.every((type) => type !== undefined) &&
dialect.canBindCompositeIn(types)
```

can be inverted — sending every composite include down the portable `OR` on both dialects — and **no suite notices**. The `OR` returns exactly the same rows. What changes is one plan cache entry and one prepared statement per distinct parent count, on the query the relation loader issues for *every* batched include. That is the whole of #97.

### The test that looks like it covers this

`plan.discrimination.test.ts` asserts the consequence:

> A composite key now binds one array per **column** through `unnest`, so the SQL text is fixed however many parents arrive and every length is one entry.

but it builds the marker **by hand** —

```ts
planKey(dialect, "LedgerEntry", "findMany", {
  where: { $compositeIn: { fields: […], values: […] } },
})
```

— and hands it to `planKey`. It says what happens once the marker exists, and nothing about whether the planner ever creates one. The mutant deletes the code that creates it, which that test cannot see by construction.

This is the shape worth guarding: a regression here is invisible to every correctness test, by construction, and shows up only as plan-cache churn.

### Fix

Assert the *choice*, on the arguments the loader hands the child model's `$exec` — that is where the marker travels, so it needs no database and does not depend on emitted SQL text. Both directions: Postgres gets it, SQLite does not. A test for only the positive case would pass with the dialect check removed entirely.

### Two things that will trip you up

- `load` is `(parents, args, executor)`, not `(parents, executor)`.
- `COMPOSITE_IN` is the string `"$compositeIn"`, not a symbol. Looking for it with `getOwnPropertySymbols` finds nothing — which is indistinguishable from the planner not setting it.

### Also from this run, and not defects

The template suite kills the implicit many-to-many `A`/`B` column assignment and the nested `omit` handling — both have real coverage. The remaining five survivors are `x === undefined || x === null` guards whose halves no caller can distinguish.

nstfkc:
Closes #252.

Mutation on `plan-relations.ts`: 60 sites, **10 surviving the unit suite**, **6 surviving the template suite as well**. Two of the four the template killed are worth naming — the implicit many-to-many `A`/`B` column assignment and the nested `omit` handling both have real coverage there. One of the six is a gap.

The guard that decides whether a composite relation uses Postgres' `unnest`:

```ts
types.every((type) => type !== undefined) &&
dialect.canBindCompositeIn(types)
```

can be inverted, sending every composite include down the portable `OR` on both dialects, and no suite notices. The `OR` returns exactly the same rows; what changes is one plan cache entry and one prepared statement per distinct parent count, on the query the relation loader issues for **every** batched include. That is the whole of #97.

## The test that looks like it covers this, and cannot

`plan.discrimination.test.ts` asserts the consequence — a `where` carrying `$compositeIn` is one plan key however many tuples it holds — but it builds that marker **by hand** and passes it to `planKey`. It says what happens once the marker exists, and nothing about whether the planner ever creates one. The mutant deletes the code that creates it.

That is the shape worth guarding: a regression here is invisible to every correctness test *by construction*, and surfaces only as plan-cache churn.

## What this asserts

The choice itself, on the arguments the loader hands the child model's `$exec` — where the marker travels, so no database and no dependence on emitted SQL. Both directions: Postgres gets it, SQLite does not, because a test for only the positive case would pass with the dialect check removed entirely.

**Verified by re-running the mutant: killed.**

## Two of my own assertions were wrong first

Worth recording, because both failures look exactly like the bug:

- `load` is `(parents, args, executor)`, not `(parents, executor)`.
- `COMPOSITE_IN` is the string `"$compositeIn"`, not a symbol — so the first version looked for it with `getOwnPropertySymbols` and found nothing, which is indistinguishable from the planner not setting it.

## Checks

- `bun --bun vitest run orm database` — 58 files, **1501 passed**

One new test file; independent of the other eight open PRs.

### DIFF
--- a/packages/gemi/orm/compile/composite-in-selection.test.ts
+++ b/packages/gemi/orm/compile/composite-in-selection.test.ts
@@ -1,0 +1,104 @@
+import { beforeEach, describe, expect, test } from "vitest";
+
+import { PostgresDialect } from "../dialect/postgres";
+import { SqliteDialect } from "../dialect/sqlite";
+import { ledger, ledgerEntry } from "../fixtures";
+import * as registry from "../registry";
+import { COMPOSITE_IN } from "./where";
+import { compileRead } from "./read";
+
+/**
+ * **The planner chooses the composite `in` on Postgres**, which is the half of
+ * #97 that nothing asserted.
+ *
+ * `plan.discrimination.test.ts` covers the consequence: a `where` carrying a
+ * `$compositeIn` marker produces one plan key however many tuples it holds,
+ * against SQLite's one per length. That test builds the marker **by hand** and
+ * hands it to `planKey`, so it says what happens once the marker exists and
+ * nothing about whether the planner ever creates one.
+ *
+ * Mutation found the difference. Inverting the guard that decides
+ *
+ *     types.every((type) => type !== undefined) &&
+ *     dialect.canBindCompositeIn(types)
+ *
+ * sends every composite include down the portable `OR` instead, on both
+ * dialects — and survives the unit suite *and* the template suite, because the
+ * `OR` returns exactly the same rows. Only the plan count changes, and the test
+ * that counts plans never asks the planner for one.
+ *
+ * That is the shape worth guarding: a regression here is invisible to every
+ * correctness test by construction. It costs one plan cache entry and one
+ * prepared statement per distinct parent count, on the query the relation
+ * loader issues for *every* batched include.
+ *
+ * Asserted on the arguments the loader passes to the child model's `$exec`,
+ * which is where the marker travels — no database, and no dependence on the
+ * emitted SQL text.
+ */
+const postgres = new PostgresDialect();
+const sqlite = new SqliteDialect();
+
+/** The `where` the loader hands the child model, for a composite relation. */
+async function childWhere(dialect: PostgresDialect | SqliteDialect) {
+  const plan = compileRead(
+    ledger,
+    "findMany" as never,
+    { include: { entries: true } } as never,
+    dialect as never,
+  );
+
+  const [entries] = (plan as { relations?: { load: unknown }[] }).relations ?? [];
+  expect(entries, "the include produced no relation node").toBeDefined();
+
+  let captured: Record<string, unknown> | undefined;
+  const executor = {
+    exec(_model: string, _op: string, args: unknown) {
+      captured = (args as { where?: Record<string, unknown> }).where;
+      return Promise.resolve([]);
+    },
+  };
+
+  // `load(parents, args, executor)` — the args are the caller's original tree,
+  // which for a bare `include: { entries: true }` is just that.
+  await (entries as unknown as {
+    load: (rows: unknown[], args: unknown, executor: unknown) => Promise<unknown>;
+  }).load(
+    [
+      { tenantId: 1, code: "a" },
+      { tenantId: 2, code: "b" },
+    ],
+    { include: { entries: true } },
+    executor,
+  );
+
+  return captured;
+}
+
+describe("the planner picks the composite `in` where the dialect has one", () => {
+  beforeEach(() => {
+    registry.clearRegistry();
+    registry.register("Ledger", class { static $schema = ledger });
+    registry.register("LedgerEntry", class { static $schema = ledgerEntry });
+  });
+
+  test("postgres gets the marker, so the text is fixed however many parents arrive", async () => {
+    const where = await childWhere(postgres);
+
+    expect(where, "the loader issued no child query").toBeDefined();
+    expect(Object.keys(where!)).toContain(COMPOSITE_IN);
+  });
+
+  /**
+   * SQLite has no `unnest`, so `canBindCompositeIn` says no and the portable
+   * `OR` stands. Asserted rather than assumed: the guard has two halves and a
+   * test for only the positive one would pass if the dialect check were dropped
+   * entirely.
+   */
+  test("sqlite keeps the portable OR", async () => {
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the central claim holds and the test is real; the body misdescribes its own mutant, and the assertion is one `toEqual` short of covering the line it guards

Reviewed at `ed0ea0c` against `feat/orm`, in a clean worktree. Every claim below is from output I produced.

### The checks reproduce

| check | result |
| --- | --- |
| `bun --bun vitest run orm database` | 58 files, **1501 passed** |
| `bunx tsc --noEmit` | exit 0 |
| new file alone | 2 passed |

The guard is where the body says it is, `packages/gemi/orm/compile/plan-relations.ts:1456-1461`, quoted accurately.

### The gap is real

Forcing the OR branch — `false && types.every(...) && dialect.canBindCompositeIn(...)`, i.e. no composite include ever gets the marker on either dialect — **survives the entire pre-existing suite**:

```
== mutant2: never use compositeIn ==   (new file excluded)
 Test Files  57 passed (57)
      Tests  1499 passed (1499)
```

and dies to the new file:

```
FAIL  the planner picks the composite `in` ... > postgres gets the marker
AssertionError: expected [ 'OR' ] to include '$compositeIn'
```

The diagnosis of why is correct too. `composite-relations.test.ts:372` does reach Postgres with a composite include, but only asserts `plan.relations` has length 1 — it never calls `load`, so nothing before this PR observed which filter the loader builds on Postgres. That is genuine new coverage of a regression that is invisible by construction, which is the argument the body makes and it is a good one.

### The body describes a mutant that does not survive

> The guard ... **can be inverted**, sending every composite include down the portable `OR` on both dialects, and no suite notices.

Inverting it does not do that — it swaps which dialect gets which filter, and the suite notices immediately. Wrapping the condition in `!(...)`:

```
FAIL  orm/compile/composite-relations.test.ts:177 > the batched strategy filters children by an OR of ANDs
AssertionError: expected { '$compositeIn': ... } to deeply equal { OR: [ ... ] }
 Test Files  1 failed | 56 passed (57)
```

Same for dropping only `dialect.canBindCompositeIn(...)` and keeping `types.every(...)` — `composite-relations.test.ts:177` fails, 1498/1499. So the stated justification for the SQLite case ("a test for only the positive case would pass if the dialect check were dropped entirely") is true of this file read alone but **not true of the suite**: that mutant was already killed. Keeping the SQLite case costs four lines and I would keep it, but it is redundant coverage described as load-bearing, and the surviving mutant is the force-false one, not an inversion. Worth correcting in the body since the number "6 surviving" is doing evidentiary work.

### The assertion stops one step short — and the step it skips is where the bug would be

The test asserts the key `$compositeIn` is *present*. It never looks underneath. So this mutant, on the same line the test exists to guard:

```ts
return { [COMPOSITE_IN]: plannerCompositeIn(childFields, keys.slice(0, 1)) };
```

— every parent but the first silently dropped from the batched child query, i.e. wrong rows, not a plan-cache cost — survives **including this PR**:

```
== mutant4: drop all but first parent tuple ==   (new file INCLUDED)
 Test Files  58 passed (58)
      Tests  1501 passed (1501)
```

The sibling test one file over already sets the pattern that catches this: `expect(seen[0].where).toEqual({ OR: [...] })` — whole shape, not key presence. The equivalent here is free:

```ts
expect(where).toEqual({
  [COMPOSITE_IN]: plannerCompositeIn(["tenantId", "ledgerCode"], [[1, "a"], [2, "b"]]),
});
```

That kills mutant 2 and mutant 4, and pins the field mapping `code → ledgerCode` besides. Given the file's own framing — mutation-driven, kill the survivor — leaving the strongest survivor on the guarded line alive is the one substantive thing to fix.

### The casts are unnecessary, and they cost exactly the checking the body says it wanted

```ts
compileRead(ledger, "findMany" as never, { include: { entries: true } } as never, dialect as never);
```

I replaced that with the uncast call and `const [entries] = plan.relations ?? []`: `bunx tsc --noEmit` exit 0, tests 2 passed. `composite-relations.test.ts` calls `compileRead` uncast throughout, so this is the house idiom, not a concession.

`as never` on the args argument means a future signature change to `compileRead` compiles silently in this file — the same class of error as the `load(parents, args, executor)` arity mistake the body records having hit. `plan.relations` is typed; the `as { relations?: { load: unknown }[] }` cast discards that and forces the second `as unknown as { load: ... }`, which is where the arity was guessed. Take the types.

### Smaller

- The `beforeEach` registry setup is load-bearing — removing it throws `UnregisteredRelationTargetError` — and matches the sibling. But the sibling also has `afterEach(() => registry.clearRegistry())` and this file does not, so it leaves the global registry populated. Harmless under per-file isolation today; add it to match the pattern.
- Not vacuous: `expect(where, "the loader issued no child query").toBeDefined()` fails loudly if `load` never calls `exec`, which is the failure mode that would otherwise make both assertions meaningless. Good.

### Verdict

Does what it says on the substance, breaks nothing (1501/1501, tsc, lint all clean), and the tests are real tests that kill a mutant 1499 existing tests miss. Land it after tightening the Postgres assertion to `toEqual` on the marker's contents and dropping the `as never`s; fix the mutant description in the body so the record is accurate.

PR #255, opened 2026-07-29
A disconnect through a join table reported itself as a connect

### LINKED ISSUE #254: A disconnect through a join table reports itself as a connect
`NestedWriteStep.operation` is documented as the thing that

> makes a plan legible from the outside: to a test, and to whatever logs queries later

because two steps on the same relation can produce byte-identical SQL and differ only in what the step does.

On the join-table path it is a two-way split:

```ts
operation: key === "create" ? "create" : "connect",
```

so every operand that is not `create` is labelled `connect`. Measured — three of the five that reach that push are wrong:

| operand | label |
| --- | --- |
| `create` | `create` ✓ |
| `connect` | `connect` ✓ |
| `connectOrCreate` | **`connect`** |
| `disconnect` | **`connect`** |
| `set` | **`connect`** |

`disconnect` reports itself as its opposite.

The union type already permits all five, so this is simply the label not being the operand.

### Why no test caught it

`plan.writes.discrimination.test.ts` **does** read `step.operation` — but folds it into a plan-identity string and asserts two plans *differ*:

```ts
after: plan.after?.map((step) => `${step.relation}:${step.operation}`) ?? [],
```

Swapping two labels swaps both identities and they stay distinct, so it passes either way. That is the difference between checking a value and checking that values are not equal, and it is why mutation found this where the suite could not.

### Fix

Label the step with the operand it implements, and assert all five on the join-table path.

### From the same run, not defects

Mutating `nested-writes.ts`: 72 sites, 13 re-checked against the template suite, **9 killed there** — the operand dispatch for `set`, `create`, `connectOrCreate`, `disconnect`, the empty-items guard, the filter validation and the required-field check all have real coverage. Of the four surviving both, three are `x === undefined || x === null` guards whose halves no caller can distinguish; they are left alone.

nstfkc:
Closes #254.

`NestedWriteStep.operation` is documented as the thing that "makes a plan legible from the outside: to a test, and to whatever logs queries later" — two steps on the same relation can produce byte-identical SQL and differ only in what the step does.

On the join-table path it was a two-way split:

```ts
operation: key === "create" ? "create" : "connect",
```

so every operand that is not `create` was labelled `connect`. Three of the five that reach that push are wrong:

| operand | was | now |
| --- | --- | --- |
| `create` | `create` | `create` |
| `connect` | `connect` | `connect` |
| `connectOrCreate` | **`connect`** | `connectOrCreate` |
| `disconnect` | **`connect`** | `disconnect` |
| `set` | **`connect`** | `set` |

`disconnect` reported itself as its opposite. The union already permitted all five, so this is simply the label not being the operand.

## The existing test that reads this field cannot see it

`plan.writes.discrimination.test.ts` **does** read `step.operation` — but folds it into a plan-identity string and asserts two plans *differ*. Swapping two labels swaps both identities and they stay distinct, so it passes either way.

That is the difference between checking a value and checking that values are not equal, and it is exactly why mutation found this where the suite could not.

**Verified by reverting the fix with the tests in place:** the three wrong labels fail, the two correct ones pass.

## From the same run, and not defects

Mutating `nested-writes.ts`: 72 sites, 13 re-checked against the template suite, **9 killed there** — the operand dispatch for `set`, `create`, `connectOrCreate` and `disconnect`, the empty-items guard, the filter validation and the required-field check all have real coverage. Of the four surviving both, three are `x === undefined || x === null` guards whose halves no caller can distinguish; left alone deliberately.

## Checks

- `bun --bun vitest run orm database` — 57 files, **1504 passed**

Independent of the other nine open PRs.

### DIFF
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -1578,7 +1578,12 @@ function planJoinTable(
 
   out.after.push({
     relation: relation.name,
-    operation: key === "create" ? "create" : "connect",
+    // The operand this step implements, not a two-way split of it. This read
+    // `key === "create" ? "create" : "connect"`, which labelled `disconnect`,
+    // `set` and `connectOrCreate` as `connect` — the first of those being the
+    // opposite operation. The field exists so a plan is legible from the
+    // outside, so a wrong label is the whole of its cost.
+    operation: key as NestedWriteStep["operation"],
     async run(args, _context, executor, rows) {
       const parent = rows[0];
       if (!parent) return;
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -1299,6 +1299,55 @@ describe("nested writes", () => {
       ).toThrow(/means something different through a join table/);
     });
 
+    /**
+     * **The step says which operand it is**, which is what the field is for.
+     *
+     * `NestedWriteStep.operation` is documented as the thing that "makes a plan
+     * legible from the outside: to a test, and to whatever logs queries later"
+     * — two steps on the same relation can produce byte-identical SQL and
+     * differ only in what the step does.
+     *
+     * Nothing on the join-table path asserted it. Mutation found the hole:
+     * flipping
+     *
+     *     operation: key === "create" ? "create" : "connect"
+     *
+     * survives the unit suite *and* the template suite, so a `create` step
+     * could report itself as a `connect` and the only reader would be a human
+     * debugging a plan.
+     *
+     * `plan.writes.discrimination.test.ts` does read the label, but folds it
+     * into a plan-identity string and asserts two plans *differ*. Swapping the
+     * two labels swaps both identities and they stay distinct, so that test
+     * passes either way — the difference between checking a value and checking
+     * that values are not equal.
+     */
+    test.each([
+      ["create", { create: { label: "new" } }],
+      ["connect", { connect: { id: 1 } }],
+      ["connectOrCreate", { connectOrCreate: { where: { id: 1 }, create: { label: "new" } } }],
+      ["disconnect", { disconnect: { id: 1 } }],
+      ["set", { set: [{ id: 1 }] }],
+    ] as [string, Record<string, unknown>][])(
+      "a %s through a join table is labelled as one",
+      (operand, payload) => {
+        const plan = compileWrite(
+          post,
+          "update",
+          { where: { id: 1 }, data: { tags: payload } } as never,
+          sqlite,
+        );
+
+        const steps = [...(plan.before ?? []), ...(plan.after ?? [])];
+        const labels = steps
+          .filter((step) => step.relation === "tags")
+          .map((step) => step.operation);
+
+        expect(labels, `no step was planned for ${operand}`).not.toHaveLength(0);
+        expect(labels).toContain(operand);
+      },
+    );
+
     test("the operands both paths implement are not refused there", () => {
       for (const operand of BOTH) {
         expect(() =>


nstfkc:
## Review — the claim holds up under a revert test; approve with two nits

Checked out `tests/orm-nested-step-labels` (6546075) in a clean worktree, `bun install`, and ran everything from source rather than reading the diff.

### The described defect is real, and the described blind spot is real

Reverted the one-line fix in place (`operation: key as NestedWriteStep["operation"]` → `operation: key === "create" ? "create" : "connect"`) and re-ran the two files that read the field:

```
× a connectOrCreate through a join table is labelled as one
× a disconnect through a join table is labelled as one
× a set through a join table is labelled as one
 Test Files  1 failed | 1 passed (2)
      Tests  3 failed | 182 passed (185)
```

Exactly the three the table in the body predicts, and `create` / `connect` stay green. The second half of the claim checks out too: `plan.writes.discrimination.test.ts` is the *passing* file in that run. It reads `step.operation` at lines 188–190, but only to fold it into a plan-identity string and assert two identities differ — so a swapped label swaps both sides and the assertion never notices. That is a genuine hole, not a rediscovery of covered ground.

Restored, everything green again (185/185).

### Blast radius is nil, and I checked rather than assumed

The label is diagnostic-only in the current tree. `runSteps` (`packages/gemi/orm/Model.ts:944`) iterates steps and calls `step.run(...)` — it never branches on `operation`. `planKey` (`packages/gemi/orm/plan.ts:439`) keys on dialect/strategy/model/op/canonicalised args and does not touch plan steps, so no cache entry changes shape. Grepping for readers of the field outside `context.operation` (a different thing) turns up only `write.test.ts` and `plan.writes.discrimination.test.ts`. So this cannot change what any query does — which is also why the differential suites were never going to catch it, and why a unit assertion is the right instrument.

### The cast is sound today, but it's the only thing holding

`key` is `string`, and `key as NestedWriteStep["operation"]` is unchecked. It is correct right now because the guard at `nested-writes.ts:1533` throws for anything outside `JOIN_TABLE_SUPPORTED`, and all five members (`connect`, `connectOrCreate`, `create`, `disconnect`, `set`) are members of the union. Worth noting what the cast buys and costs: typing the set as `new Set<NestedWriteStep["operation"]>([...])` and narrowing off `.has()` would make the same code compile without an assertion, and would turn a future typo'd or non-Prisma entry in that set into a type error instead of a silently bogus label — i.e. exactly the failure mode this PR is fixing, one level up. Non-blocking.

### Tests

Real tests, and mutation-proven above. Two small things:

- `expect(labels).toContain(operand)` is weaker than it needs to be. `planJoinTable` pushes exactly one step per key, so `expect(labels).toEqual([operand])` holds today and additionally catches a duplicate or an extra mislabelled step on the same relation. `toContain` passes as long as *something* is right.
- The vacuity guard (`expect(labels, ...).not.toHaveLength(0)`) is the right instinct and is why the `test.each` can't silently pass on a compile path that plans nothing. Good.

Observation, out of scope: the ordinary-relation paths hardcode `operation: "connect"` at `nested-writes.ts:685` and `:1378` rather than deriving it. Both are fallthroughs only `connect` reaches, so they're correct — but they're the same "label written by hand next to the operand" shape, and nothing asserts them either. Pre-existing; not asking for it here.

### Checks run

| check | result |
| --- | --- |
| `bun --bun vitest run orm database` | 57 files, **1504 passed** |
| `bun --bun vitest run` (templates/saas-starter) | **642 passed**, 32 skipped, 1 failed |
| `bunx tsc --noEmit` | exit 0 |

The single template failure is `generated.test.ts > running prisma generate again produces a zero-line diff`, which self-reports as environment: `gemi-orm-generator is not linked. Run 'bun install', and 'bun run build:bin' in packages/gemi`. Fresh worktree, generator never built — unrelated to this diff. (Confirmed by the shape of it: running `prisma generate` there with the unlinked generator rewrites `generated/models.ts` down by 260 lines and cascades ~40 differential failures; restoring the committed file returns the suite to 642 passed.) Matches the body's reported numbers.

Nothing here blocks. The fix is one line, the label now equals the operand it implements, and the test that proves it fails without it.

PR #257, opened 2026-07-29
An upsert on a Bytes key was only tested for the case it refuses

### LINKED ISSUE #256: An upsert on a Bytes key is only tested for the case it refuses
`sameEncoded` exists because `encode` is a per-dialect pass-through and what a driver binds natively is not always a primitive. Its own doc names both types:

> Postgres passes a `DateTime` through as the `Date` it was given, and **both dialects** pass `Bytes` through as a `Uint8Array` — so the same instant and the same bytes arrive here as two distinct objects, and identity refuses a correct upsert.

`DateTime` has a test in **each** direction:

- `a DateTime conflict key compares by value on sqlite/postgres` — equal but distinct, accepted
- `a DateTime conflict key that genuinely differs is still refused`

`Bytes` has only the refusal — `a mismatched Bytes of the same length is described distinguishably`. The branch that makes a **matching** pair work was never exercised.

Mutation found it. Inverting the length check inside that branch:

```ts
if (left.length !== right.length) return false;   //  !==  ->  ===
```

makes every equal pair compare unequal, and nothing fails. In effect a correct upsert on a `Bytes` unique key would be refused with a message printing two equal-looking values — the exact failure the function was written to prevent.

### Fix

Mirror the `DateTime` positive test for `Bytes`, on **both** dialects. Unlike the `Date` case, which is Postgres-only because SQLite encodes a `DateTime` to a number and never reaches this, SQLite passes `Bytes` through as an object exactly as Postgres does.

### One mutant to leave alive

`ArrayBuffer.isView(a) && ArrayBuffer.isView(b)` survives flipping to `||`, and should. The mixed case it then admits is unreachable — both values come from the same field through the same encoder, so they are either both views or neither. Equivalent, not uncovered, and the same is true of the `Date` pair above it.

nstfkc:
Closes #256.

`sameEncoded` exists because `encode` is a per-dialect pass-through and what a driver binds natively is not always a primitive. Its own doc names both types:

> Postgres passes a `DateTime` through as the `Date` it was given, and **both dialects** pass `Bytes` through as a `Uint8Array` — so the same instant and the same bytes arrive here as two distinct objects, and identity refuses a correct upsert.

`DateTime` has a test in each direction — equal-but-distinct accepted, genuinely different refused. `Bytes` had only the refusal, so the branch that makes a **matching** pair work was never exercised.

Mutation found it: inverting the length check inside that branch makes every equal pair compare unequal, and nothing failed. In effect a correct upsert on a `Bytes` unique key would be refused with a message printing two equal-looking values — the exact failure the function was written to prevent.

Both dialects, unlike the `Date` case which is Postgres-only: SQLite encodes a `DateTime` to a number and never reaches this, but it passes `Bytes` through as an object exactly as Postgres does.

**Verified by re-running the mutant:**

```
mutant 824 !== -> ===  KILLED
    × a Bytes conflict key compares by value on sqlite
    × a Bytes conflict key compares by value on postgres
```

## The `&&` mutant is killed too — it was not equivalent

An earlier version of this description recorded `ArrayBuffer.isView(a) && ArrayBuffer.isView(b)` → `||` as an equivalent survivor, on the grounds that both values come from the same field through the same encoder and are therefore either both views or neither. **That reasoning is wrong**, and review caught it: `a` is the `create` value and `b` is the `where` value, supplied independently, and neither is type-checked before it reaches `sameEncoded` — the neighbouring `1n`-against-`1` row is standing proof that a mixed pair arrives. Under `||` the non-view side is rebuilt as `new Uint8Array(undefined, undefined, undefined)`, length 0, so an empty `Bytes` key matches a wrong-typed `create` and a mismatched key is **silently accepted on a write** — the same class of divergence the function exists to stop.

So it is killed rather than excused, by one row on the existing "described distinguishably" case: an empty `Bytes` where clause against an `"abc"` create.

```
mutant 821 && -> ||  KILLED
    × a mismatched an empty Bytes against a string is described distinguishably
```

The length-guard mutant still fails both `Bytes` tests, so the new row does not displace the original coverage.

The `Date` pair a line above has the same shape and is presumably killable the same way; that is a separate line and not this PR's issue.

## Also from review

- The `Digest` fixture was copied verbatim into two tests that differ only in the key's type — now one `digestSchema(keyType)` factory at the top of the `upsert` describe.
- The new `Bytes` test binds and asserts the bytes reach the parameter list (`toContainEqual(new Uint8Array([1, 2, 3]))`) rather than resting on `not.toThrow()` alone.

## Checks

- `bun --bun vitest run orm database` — 57 files, **1502 passed**

Tests in an existing describe; independent of the other open PRs.

### DIFF
--- a/packages/gemi/orm/compile/write.test.ts
+++ b/packages/gemi/orm/compile/write.test.ts
@@ -469,6 +469,46 @@ describe("delete", () => {
 });
 
 describe("upsert", () => {
+  /**
+   * A one-field unique whose key type varies — the shape the conflict-key
+   * agreement tests below need. They differ only in that type, so this is a
+   * factory rather than a fixture copied per test.
+   */
+  const digestSchema = (keyType: string): any => ({
+    name: "Digest",
+    table: "Digest",
+    fields: {
+      id: {
+        name: "id",
+        column: "id",
+        type: "Int",
+        nullable: false,
+        isId: true,
+        isUpdatedAt: false,
+        default: { kind: "autoincrement" },
+      },
+      key: {
+        name: "key",
+        column: "key",
+        type: keyType,
+        nullable: false,
+        isId: false,
+        isUpdatedAt: false,
+      },
+      note: {
+        name: "note",
+        column: "note",
+        type: "String",
+        nullable: true,
+        isId: false,
+        isUpdatedAt: false,
+      },
+    },
+    primaryKey: ["id"],
+    uniques: [["key"]],
+    relations: {},
+  });
+
   test("one statement: insert with on-conflict do update", () => {
     expect(
       text("upsert", {
@@ -675,40 +715,7 @@ describe("upsert", () => {
     ["sqlite", sqlite],
     ["postgres", postgres],
   ])("a Bytes conflict key compares by value on %s", (_name, dialect) => {
-    const digest: any = {
-      name: "Digest",
-      table: "Digest",
-      fields: {
-        id: {
-          name: "id",
-          column: "id",
-          type: "Int",
-          nullable: false,
-          isId: true,
-          isUpdatedAt: false,
-          default: { kind: "autoincrement" },
-        },
-        key: {
-          name: "key",
-          column: "key",
-          type: "Bytes",
-          nullable: false,
-          isId: false,
-          isUpdatedAt: false,
-        },
-        note: {
-          name: "note",
-          column: "note",
-          type: "String",
-          nullable: true,
-          isId: false,
-          isUpdatedAt: false,
-        },
-      },
-      primaryKey: ["id"],
-      uniques: [["key"]],
-      relations: {},
-    };
+    const digest = digestSchema("Bytes");
 
     // Equal bytes, two distinct objects — which is how they arrive from a
     // caller who built the `where` and the `create` separately.
@@ -719,7 +726,12 @@ describe("upsert", () => {
     };
 
     const compiled = compileWrite(digest, "upsert", args, dialect);
-    expect(() => compiled.bind(args, createBindContext())).not.toThrow();
+
+    // Binding the values is the assertion: it does not throw, *and* the bytes
+    // survive the check into the parameter list rather than being dropped or
+    // replaced by whatever the agreement check compared.
+    const values = compiled.bind(args, createBindContext());
+    expect(values).toContainEqual(new Uint8Array([1, 2, 3]));
   });
 
   test("a DateTime conflict key that genuinely differs is still refused", () => {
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the claim holds; the two tests are real, and one paragraph of the body's reasoning is wrong

Reproduced the whole argument locally rather than reading it. Base `feat/orm` was at `60a31d1` while this branch sits on `9f28c51`; results below are on the branch head, then re-run after merging current base.

### The gap is real and the tests close it

Mutating the length guard inside the `ArrayBuffer.isView` branch of `sameEncoded` — `left.length !== right.length` → `===`, so every equal pair compares unequal — and running the **base** version of `write.test.ts` across the whole suite:

```
Test Files  57 passed (57)
     Tests  1499 passed (1499)
```

Mutant alive, exactly as claimed. Same mutant with this branch's test file:

```
× a Bytes conflict key compares by value on sqlite
× a Bytes conflict key compares by value on postgres
Tests  2 failed | 172 passed (174)
```

and the failure is the one the body predicts, verbatim from the run:

```
'key' is 3 bytes (0x010203) in the where clause but 3 bytes (0x010203) in 'create'.
```

A correct upsert refused, with a message printing two identical values. That is the failure `sameEncoded` exists to prevent, so this is a behavioural hole and not a coverage number.

The tests also kill the loop body (`left[i] !== right[i]` → `===` fails both new tests plus the existing same-length mismatch test), so the new coverage is the byte-compare path end to end, not just the guard. Both arguments are distinct `Uint8Array` instances, so `a === b` cannot short-circuit — the branch is genuinely entered.

Confirmed both encoders pass `Bytes` through untouched (`sqlite.encode` has no `Bytes` case and falls to `return value`; `postgres.encode` is a pass-through), so the SQLite half of the `test.each` is not padding — it exercises the same branch on the other dialect, unlike the Postgres-only `DateTime` case.

### Nothing breaks

Branch head, and again after merging current `feat/orm` (clean merge, no conflicts):

```
head:   57 files, 1501 passed          (base was 1499 — the two new tests)
merged: 58 files, 1531 passed
tsc --noEmit  exit 0
oxlint orm --deny-warnings  0 warnings, 0 errors
```

### One correction: the surviving mutant is not equivalent

The body says `ArrayBuffer.isView(a) && ArrayBuffer.isView(b)` → `||` should stay alive because "the mixed case it then admits is unreachable — both values come from the same field through the same encoder, so they are either both views or neither."

The encoder is the same, but the two *values* are not: `a` is the `create` value and `b` is the `where` value, independently supplied, and neither is type-checked before it gets here — the neighbouring test proves that by feeding `1n` against `1` on a `BigInt` field. A mixed pair reaches `sameEncoded`. Probed directly on a `Bytes` key:

```
where: Uint8Array([1,2,3]), create: "abc"   → UnsupportedQueryError  (unmutated and mutated — same)
where: Uint8Array([]),      create: "abc"   → unmutated: UnsupportedQueryError
                                            → with ||: ACCEPTED
where: Uint8Array([]),      create: 7       → unmutated: UnsupportedQueryError
                                            → with ||: ACCEPTED
```

With `||`, the non-view side becomes `new Uint8Array(undefined, undefined, undefined)` — length 0 — and matches an empty view, so a mismatched key is silently accepted. That is the same class of silent divergence on a write the function was written to stop, just behind an empty-`Bytes` key.

I would not block on it: an empty `Bytes` unique key against a wrong-typed `create` is a narrow shape, and killing it is a one-line `test.each` row on the existing "described distinguishably" case. But the body should not record it as equivalent — the stated reason ("either both views or neither") is the part that is untrue, and that reasoning is what a future reader will trust when they touch this line.

### Minor

- The `digest` fixture is copied verbatim into the new test from the `describe` block ~30 lines below; the only difference there is the computed `type`. Worth hoisting to a shared `const` in the `upsert` describe.
- `expect(...).not.toThrow()` is a weak assertion on its own; it earns its keep here because the mutant proof shows the throw is exactly what regresses, and it matches the `DateTime` test it parallels. Asserting the bound parameter carries the bytes would be strictly stronger, but is not required for this branch.

Verdict: correct, evidence-backed, safe to land. Fix the equivalence paragraph in the description (or add the row that kills the mutant) before merge.

PR #260, opened 2026-07-29
Eleven template suites loaded a Prisma client to read an env var

### LINKED ISSUE #217: relations.many-to-many.test.ts intermittently fails collection in the SQLite job
`templates/saas-starter/app/models/relations.many-to-many.test.ts` intermittently fails to collect when the template suite runs as a whole. Filing it characterised rather than diagnosed — I could not capture the error, and would rather record what is known than guess at a fix.

### What is known

```
 FAIL  app/models/relations.many-to-many.test.ts [ app/models/relations.many-to-many.test.ts ]
 Test Files  1 failed | 16 passed | 3 skipped (20)
      Tests  605 passed | 32 skipped (637)
```

- It is a **collection** failure, not a test failure: no individual test fails, and the file's 15 tests are simply absent — 605 instead of the usual 620.
- **It passes in isolation**, every time: `vitest run app/models/relations.many-to-many.test.ts` → 15 passed.
- Frequency is roughly **2 in 15** full-suite runs. Observed twice, on two different branches, days apart, so it is not specific to any change in flight.
- Always this file.

### What has been ruled out

- **Shared database file.** It uses `mkdtempSync` for its own workspace, like the other template suites, so two files cannot be writing the same SQLite database.
- **A change in flight.** First seen on `fix/orm-type-tests-never-run`, again on `feat/orm-json-scalars-on-postgres`; both times the surrounding runs were clean.

### What has not been established

The error itself. Eight consecutive full-suite runs after the failure were clean, so I have the shape of the failure but not its text. Whatever it is, it happens before any test in the file runs.

Worth noting the SQLite job runs the template suite **without** `--no-file-parallelism`, which the Postgres job uses precisely because those suites share state. That is a hypothesis, not a finding — the temp-directory isolation above argues against it, and I have not been able to confirm or refute it.

### Why it is worth tracking

It will redden CI at random on an unrelated PR, and a board that goes red for reasons nobody can act on stops being read — which is the same reasoning the workflow already records for #163 and #203. Whoever picks this up should start by capturing the error: run the full suite in a loop until it fails, with the verbose reporter and stderr kept.

nstfkc:
Closes #217.

## The failure shape is measurable, and it rules out the hook

#217 asks which of two shapes the failure is, and the thread settles it from vitest's output formatting — the bracketed `FAIL  <file> [ <file> ]` is the import-error form. That is right, and there is a second signal in the same paste that does not depend on reading vitest's printer: **the counts**.

Injecting each candidate into `relations.many-to-many.test.ts` and running `vitest run app/models` (clean baseline on this branch is `640 passed | 32 skipped (672)`):

```
injected                  reported as                totals
a hook that times out     FAIL <file> > <suite>      625 passed | 47 skipped (672)
a module that throws      FAIL <file> [ <file> ]     625 passed | 32 skipped (657)
```

A failed hook still contributes the file's 15 tests to the run, as **skipped** — the total is unchanged and `skipped` goes 32 → 47. A module that throws before `describe` runs leaves them **absent** — `skipped` stays 32 and the total drops by 15.

#217 reports `605 passed | 32 skipped (637)`: total short, skipped unchanged. So it is the second row, and **`beforeAll(start, 60_000)` cannot be the mechanism** — a fact rather than the judgement call the thread was left making. That line is the shortest hook timeout in the directory and the first thing anyone picking this up reads; it now carries the measurement in a comment, and is otherwise left exactly as it was.

## What that leaves, and the one part of it that was pure cost

The suspect surface is the import graph. `differential.ts` imports `@prisma/client` at module scope — it has to; it *is* the harness that compares against Prisma. It also exported

```ts
export const POSTGRES_URL = process.env.TEST_POSTGRES_URL;
```

and `applyMigrations`, which needs `bun`'s `SQL` and two node builtins.

Thirteen of the twenty suites in `app/models` imported that module. **Eleven of them wanted only those two things** — so eleven workers loaded a Prisma query engine, before a single `describe` ran, to read an environment variable. Both exports now live in `app/models/scratch.ts`, whose entire import list is `bun`, `node:fs`, `node:path`. The two suites that genuinely compare against Prisma (`differential.test.ts`, `writes.differential.test.ts`) still import the harness, because they use it.

**This does not claim to be the cause.** It has not been reproduced — not in the ~45 runs #217 records, and not in 50 more this round (30 plain, 15 under 2× CPU oversubscription, 20 after the change), which is why the issue's own recommendation was to stop guessing. It is the one suspect that could be removed for free, in the phase where a failure costs the whole file's tests rather than one of them, and it is worth doing whether or not it is the fault: eleven workers now do not open a query engine to read `process.env`.

Honest about the size: in isolation the removed `require` is ~19ms per worker, and the suite's own `import` total is too noisy for that to show (5.0–6.1s before, 4.8–6.3s after, ten runs each). The exact claim is countable rather than timed — **eleven of thirteen suites no longer reach `@prisma/client` at collection**.

## Why the rule is asserted rather than written down

The regression is invisible. Adding `POSTGRES_URL` back to an existing `./differential` import is one word in a diff, nothing fails, and collection gets a little slower — and it is the edit an author makes for a good local reason, because the symbol is in a module they already import. So `packages/gemi/orm/template-import-graph.test.ts` fails if a suite imports the harness without calling `createDifferential`, and if `scratch.ts` ever imports anything beyond those three.

Mutation-checked in the direction that matters, by putting the old import back on the one file #217 is about:

```
FAIL  orm/template-import-graph.test.ts > the template's model suites >
      relations.many-to-many.test.ts imports the differential harness only to use it

AssertionError: relations.many-to-many.test.ts imports ./differential but never
calls createDifferential, so it pays for @prisma/client at collection time and
gets nothing for it. POSTGRES_URL and applyMigrations live in ./scratch.
```

It also asserts its own premises — that suites were found at all, and that `differential.ts` still imports `@prisma/client` — so a rename cannot quietly turn it into a tautology.

## If it recurs

The counts above are the diagnostic. The file-level line names the file; the **`Tests` line says which half of the run to look in**, and a job that greps for failing tests will see neither. The error text is in the log immediately under the `FAIL` line — that is what is still missing, and nothing here substitutes for it.

## Checks

- `bun --bun vitest run app/models` (template, `TZ=UTC`) — **17 passed | 3 skipped (20)**, `640 passed | 32 skipped (672)`; 20 consecutive runs clean
- `bun --bun vitest run` with CI's six exclusions (`packages/gemi`) — **66 files, 1579 passed | 1 skipped**
- `bun run test:types` (template) — 24 passed, no type errors
- `bun run typecheck` (`packages/gemi`) exit 0; `oxlint` clean on both new files (the 6 warnings in `app/models` are pre-existing, in `bench/run.ts` and `select.test-d.ts`)

Postgres was not run — no scratch database here. The change is import-graph only and the gating expressions are untouched, so `postgres-suite-selection.test.ts` still matches all three gated files.

### DIFF
--- a/packages/gemi/orm/template-import-graph.test.ts
+++ b/packages/gemi/orm/template-import-graph.test.ts
@@ -1,0 +1,103 @@
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * A template suite loads a Prisma client at collection time only if it uses one.
+ *
+ * **The cost this exists to keep off the collection phase.** `differential.ts`
+ * imports `@prisma/client` at module scope — it has to; it *is* the harness that
+ * compares against Prisma. It also used to export `POSTGRES_URL`, which is
+ *
+ *     export const POSTGRES_URL = process.env.TEST_POSTGRES_URL;
+ *
+ * and `applyMigrations`, which needs `bun`'s `SQL` and two node builtins. Eleven
+ * of the thirteen suites that imported the module wanted only those two things,
+ * so eleven workers loaded a query engine, before any `describe` ran, to read an
+ * environment variable. Both now live in `app/models/scratch.ts`, whose import
+ * list is `bun`, `node:fs`, `node:path`.
+ *
+ * **Why it is asserted rather than left to reviewers.** The regression is
+ * invisible: adding `POSTGRES_URL` to an existing `./differential` import is one
+ * word in a diff, nothing fails, and the suite is a little slower to collect. It
+ * is also the kind of edit an author makes for a good local reason — the symbol
+ * is right there in a module they already import.
+ *
+ * **The failure it removes a suspect for is #217**, where
+ * `relations.many-to-many.test.ts` intermittently fails *collection* in the
+ * SQLite job. That the failure is at import rather than in `beforeAll` is
+ * measurable from the run's own counts rather than inferred from vitest's
+ * formatting — injecting each shape into that file and running the directory:
+ *
+ *     a hook that times out    625 passed | 47 skipped (672)
+ *     a module that throws     625 passed | 32 skipped (657)
+ *
+ * A failed hook still contributes the file's 15 tests, as *skipped*. A module
+ * that throws leaves them absent and the total short. #217 reports
+ * `605 passed | 32 skipped (637)`, so it is the second, and the 60s hook timeout
+ * that reads as the obvious suspect is not the mechanism.
+ *
+ * This does not claim to be that bug's cause; it has never been reproduced, and
+ * a check that passes today would have passed the day it happened. It makes the
+ * import surface where such a failure can live smaller, and keeps it that way.
+ */
+const MODELS = join(
+  import.meta.dirname,
+  "../../../templates/saas-starter/app/models",
+);
+
+/** The module whose import pulls in `@prisma/client`, and what earns importing it. */
+const HARNESS = "./differential";
+const HARNESS_API = "createDifferential";
+
+const suites = readdirSync(MODELS)
+  .filter((file) => file.endsWith(".test.ts"))
+  .map((file) => ({ file, source: readFileSync(join(MODELS, file), "utf8") }));
+
+describe("the template's model suites", () => {
+  /**
+   * Both halves of the comparison have to be readable. A rename that made the
+   * regex match nothing would turn every assertion below into a tautology, which
+   * is precisely how a check that has stopped checking looks from the outside.
+   */
+  test("the suites were found", () => {
+    expect(suites.length).toBeGreaterThan(10);
+    expect(
+      suites.filter(({ source }) => source.includes(`from "${HARNESS}"`)).length,
+      `no suite imports ${HARNESS} — has it been renamed?`,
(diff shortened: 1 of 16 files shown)


nstfkc:
Content landed on `feat/orm` in 1c2741f, so this is closed rather than merged.

This PR targeted `tests/orm-bytes-conflict-key` (#260) / `tests/orm-many-to-many-collection` (#264) — branches that had already merged into `feat/orm` and had no open PR of their own. Merging it would have put the work one more level from the trunk while reporting `MERGED`: the pooling failure `plans/orm/README.md` documents and the runbook there exists to catch.

So the stack tip was merged into `feat/orm` directly, which lands both commits:

```
156e642  Eleven template suites loaded a Prisma client to read an env var   (#260)
b3a36ed  Every cuid invariant except length and prefix was unguarded        (#264)
```

No content was dropped — `git rev-list --no-merges origin/feat/orm..` is empty for both branches now. CI is green on the result (`sqlite` and `postgres`, on the proposal and the merge result).

Filed #267 for the branch hygiene this is an instance of.

nstfkc:
Correction to the comment above: the branch-hygiene issue is **#268**, not #267. (#267 is a separate find from the same audit — JSON path filters built on a branch and never merged.)

PR #264, opened 2026-07-29
Every cuid invariant except length and prefix was unguarded

### LINKED ISSUE #258: Every cuid invariant except length and prefix is unguarded
`defaults.ts` has four exports and no test names any of them. That would be a
thin observation on its own — plenty of code is covered indirectly, and
`createCuid` is: making it throw fails 10 tests in `compile/write.test.ts`. So
it runs. The question is what runs *on it*, and the answer is one regex.

## What constrains a cuid today

A single assertion, in `plan.writes.discrimination.test.ts`, incidental to a
plan-cache test that is really about something else:

```
expect(String(first[0])).toMatch(/^c[a-z0-9]{24}$/);
```

Prefix, charset, length. Nothing about the structure those 24 characters have,
which is the part `defaults.ts` documents at length:

```
c  ms3j09fa  0000     a045         2l77tbma
^  ^         ^        ^            ^
|  |         |        fingerprint  random
|  |         counter
|  timestamp, base 36
literal 'c'
```

## Measured

Five mutations, each run against the unit suite and the template suite:

| mutation | |
|---|---|
| drop the `c` prefix | killed |
| halve the random suffix | killed |
| **counter frozen** — `counter++` becomes `0` | **survives** |
| **fingerprint per-call** rather than per-process | **survives** |
| **bias resample removed** — `byte >= 252` becomes `byte >= 256` | **survives** |

The two that die are the two the regex can see: it pins the first character and
the total length. Every mutation that stays 25 characters of lowercase
alphanumeric survives both suites.

## Why each survivor is worth pinning

**The counter.** Its own comment states the invariant and nothing checks it:

> is what keeps two cuids minted in the same millisecond distinct, so it must
> not be reset or made per-call.

A documented hazard with no test is the shape that `docs`-guard work in this
repo keeps finding — the note tells a reader it matters and tells CI nothing.
Worth being precise about severity: with eight characters of randomness in the
id, freezing the counter does not by itself produce a collision. It removes a
layer that is there *because* the other layers are probabilistic, and it removes
it silently.

**The fingerprint.** Also not a collision risk — per-call randomness carries the
same entropy. What it removes is the process identity, which the module chose
deliberately over cuid v1's pid+hostname and explains at length. It is the field
someone diagnosing a duplicate would read first, and it would be lying.

**The bias resample.** 256 is not a multiple of 36, so without the resample the
first four symbols of the alphabet arrive about 14% more often than the rest.
Invisible to any assertion on a single id — and the exact kind of loop that gets
"simplified" by someone who reads `while (byte >= 252)` as arbitrary.

## Proposed

`defaults.test.ts`, covering the three survivors plus the module's other three
exports, which have no direct test either:

- the block widths are fixed, so the counter and fingerprint can be sliced out
  by position — assert the counter advances between consecutive ids and the
  fingerprint does not
- the bias as a frequency check over the random blocks of many ids. Sized so it
  is not a flaky test: the biased distribution sits tens of standard deviations
  from the uniform one at 200k symbols, so the threshold can be generous
- `isClientSideDefault`, `hasClientSideValue`, `clientSideValue` — including
  that `@updatedAt` beats a `@default`, and that `now` is the *same* instant
  across fields, which is the behaviour the header verified against Prisma
  6.19.2 and left unguarded

Not proposing a change to `defaults.ts`. The module looks right — the resample
bound is correct, the counter is module-scoped, the fingerprint is per-process.
This is about noticing if that stops being true.

nstfkc:
Closes #258.

`defaults.ts` had four exports and no test naming any of them. `createCuid` was covered *indirectly* — making it throw fails ten tests in `compile/write.test.ts` — so it ran. What ran **on** it was one assertion, incidental to a plan-cache test that is really about something else:

```
expect(String(first[0])).toMatch(/^c[a-z0-9]{24}$/);
```

Prefix, charset, length. Nothing about the structure those 24 characters have, which is the part the module documents at length.

## Reproduced before writing anything

Five mutations, each run against the unit suite and the template suite:

| mutation | unit | template |
|---|---|---|
| drop the `c` prefix | killed | — |
| halve the random suffix | killed | — |
| **`counter++` → `0`** | **1579 passed** | **640 passed** |
| **fingerprint per-call, not per-process** | **1579 passed** | **640 passed** |
| **`byte >= 252` → `byte >= 256`** | **1579 passed** | **640 passed** |

The two that die are the two the regex can see. Every mutation that stays 25 characters of lowercase alphanumeric survives both suites untouched — same counts as a clean run.

Worth being precise about severity, because none of the three is a collision on its own:

- **The counter** is the one field in the id that is *not* probabilistic, and its own comment states the invariant — *"must not be reset or made per-call"*. Freezing it does not by itself collide; eight characters of randomness follow. It removes a layer that is there *because* the other layers are probabilistic, silently.
- **The fingerprint** carries the same entropy per-call as per-process. What per-call destroys is process identity — the field someone diagnosing a duplicate reads first, chosen deliberately over cuid v1's pid+hostname and explained at length. Made per-call it would still be four right-looking characters, and it would be lying.
- **The bias resample.** 256 is 7×36+4, so without it the first four symbols of the alphabet arrive 8 times in 256 against the others' 7 — 12.5% above uniform. Invisible to any assertion about a single id, and exactly the loop that gets "simplified" by someone reading `while (byte >= 252)` as arbitrary.

## What the new file pins

`defaults.test.ts`, 30 tests, no change to `defaults.ts`. The module is right — the resample bound is correct, the counter is module-scoped, the fingerprint is per-process. This is about noticing if that stops being true.

The block widths are fixed, so the id is taken apart by position — **from the end**, not the front. The timestamp is the one block whose width is not a constant (`Date.now().toString(36)` is eight characters until 2059 and nine after), and a test that starts failing on a date is not a test of anything here.

- **counter** — advances by exactly one between consecutive ids; and with `Date.now` held still, two ids minted in the same millisecond share a timestamp block and differ in the counter, which is the failure the field exists for. Compared on the counter alone rather than through the whole id, which the random blocks would keep distinct regardless.
- **fingerprint** — identical across 100 ids in this process, *and* different in a fresh one. Constant is not enough: a hardcoded literal satisfies the first half forever. Five `vi.resetModules()` instances stand in for five processes, asserting only that they are not all equal — a hardcoded value yields exactly one distinct fingerprint, and the flake probability is (1/36⁴)⁴.
- **bias** — a frequency check over 200,000 random symbols (25,000 ids × 8; the fingerprint is excluded, being one draw repeated). Sized not to be flaky rather than sized to just catch the mutant: uniform puts the share of `0123` at 4/36 = 0.1111 with sd 0.00031, the unresampled distribution sits at 0.125 — **44 sd away** — so the window is ±0.004, 12.7 sd of headroom each way. Over 40 sampling runs the observed share ranged **0.1095–0.1125**; the mutant measured **0.1242**. Plus: all 36 symbols still appear, so the resample cannot truncate the alphabet.
- **timestamp** — decodes to base-36 milliseconds bracketed by `Date.now()` either side of the call, which is what pins where the other three blocks start.
- **`isClientSideDefault`** — an exhaustive `Record<DefaultKind, boolean>`, so a new kind added without deciding which side of the split it falls on is a type error rather than an untested branch.
- **`hasClientSideValue`** — client-side default, database-side default, no default, and `@updatedAt`, which is the one field recognised by its attribute rather than by its default.
- **`clientSideValue`** — that **`@updatedAt` beats a `@default`** on the same field (the code says so in a comment; nothing checked it), and that **`now` is the same instant** across every field of an operation — identity, not equality, since reading the clock per field would compare equal most of the time and differ under load. That is the behaviour the module header records as verified against Prisma 6.19.2, and it was the reason with no test behind it.

## Verification

Each mutation re-applied with the new file in place:

| mutation | `defaults.test.ts` |
|---|---|
| `counter++` → `0` | 2 failed |
| fingerprint per-call | 1 failed |
| fingerprint hardcoded | 1 failed |
| resample removed | 1 failed — `expected 0.124205 to be less than 0.1151` |
| timestamp not the clock | 1 failed |
| `dbgenerated` dropped from the split | 1 failed |
| `@updatedAt` early return removed | 2 failed |

The file was also run 15 times consecutively to check the frequency test is not flaky: 30/30 every time.

## Checks

- `bun --bun vitest run` (CI's exclusion set) — 67 files, **1609 passed** / 1 skipped
- template SQLite, `TZ=UTC` — 17 passed / 3 skipped, **640 tests**

One new file. Touches nothing else, so it is independent of anything else in flight.

https://claude.ai/code/session_01EvWT1rfjgjbowF7Axua4fd

### DIFF
--- a/packages/gemi/orm/defaults.test.ts
+++ b/packages/gemi/orm/defaults.test.ts
@@ -1,0 +1,365 @@
+import { beforeAll, describe, expect, test, vi } from "vitest";
+
+import {
+  clientSideValue,
+  createCuid,
+  hasClientSideValue,
+  isClientSideDefault,
+} from "./defaults";
+import type { DefaultKind, FieldSchema } from "./schema";
+
+/**
+ * `defaults.ts` had four exports and no test naming any of them (#258).
+ *
+ * `createCuid` ran — making it throw failed ten tests in `compile/write.test.ts`
+ * — but exactly one assertion constrained what it produced, incidental to a
+ * plan-cache test about something else:
+ *
+ * ```
+ * expect(String(first[0])).toMatch(/^c[a-z0-9]{24}$/);
+ * ```
+ *
+ * Prefix, charset, length. Five mutations run against the unit suite and the
+ * template suite; the two the regex can see died, and three lived:
+ *
+ * | mutation                                         |          |
+ * |--------------------------------------------------|----------|
+ * | drop the `c` prefix                              | killed   |
+ * | halve the random suffix                          | killed   |
+ * | `counter++` becomes `0`                          | survived |
+ * | fingerprint per-call rather than per-process     | survived |
+ * | `byte >= 252` becomes `byte >= 256`              | survived |
+ *
+ * Every mutation that stayed 25 characters of lowercase alphanumeric survived
+ * both suites. None of the three is a collision risk on its own — that is the
+ * point. They are the layers that are there *because* the other layers are
+ * probabilistic, and each could be removed without a single test changing
+ * colour.
+ *
+ * This file is not a proposal to change `defaults.ts`. The module looks right:
+ * the resample bound is correct, the counter is module-scoped, the fingerprint
+ * is per-process. This is about noticing if that stops being true.
+ */
+
+// The block widths are fixed, so a cuid can be taken apart by position. Sliced
+// from the *end* rather than the front, because the timestamp is the one block
+// whose width is not a constant — `Date.now().toString(36)` is eight characters
+// until 2059 and nine after it, and a test that would start failing on a date is
+// not a test of anything here.
+const RANDOM = -8;
+const FINGERPRINT = -12;
+const COUNTER = -16;
+
+const fingerprintOf = (cuid: string) => cuid.slice(FINGERPRINT, RANDOM);
+const randomOf = (cuid: string) => cuid.slice(RANDOM);
+const counterOf = (cuid: string) =>
+  Number.parseInt(cuid.slice(COUNTER, FINGERPRINT), 36);
+
+describe("createCuid emits the shape Prisma 6.19.2 emits", () => {
+  test("prefix, charset and length", () => {
+    expect(createCuid()).toMatch(/^c[0-9a-z]{24}$/);
+  });
+
+  /**
+   * The timestamp is the only block with a meaning outside the id, and the only
+   * one that pins where the other three start. A cuid minted now decodes to now.
+   */
+  test("the timestamp block is base-36 milliseconds", () => {
+    const before = Date.now();
+    const cuid = createCuid();
+    const after = Date.now();
+
+    const timestamp = Number.parseInt(cuid.slice(1, COUNTER), 36);
+
+    expect(timestamp).toBeGreaterThanOrEqual(before);
+    expect(timestamp).toBeLessThanOrEqual(after);
+  });
+});
+
+/**
+ * The counter's own comment states the invariant:
+ *
+ * > is what keeps two cuids minted in the same millisecond distinct, so it must
+ * > not be reset or made per-call.
+ *
+ * Nothing checked it. Freezing it at `0` does not by itself produce a collision
+ * — eight characters of randomness follow it — which is exactly why no existing
+ * assertion could see the difference. What it removes is the one field in the id
+ * that is *not* probabilistic, silently.
+ */
+describe("the counter is monotonic within the process", () => {
+  test("it advances by one between consecutive ids", () => {
+    const counters = Array.from({ length: 5 }, () => counterOf(createCuid()));
+
+    for (let i = 1; i < counters.length; i++) {
+      expect(counters[i]).toBe(counters[i - 1] + 1);
+    }
+  });
+
+  /**
+   * The failure being guarded is two ids minted in the same millisecond, so mint
+   * them in the same millisecond: with the clock held still the counter is the
+   * only field that has to keep them apart, and it is compared on its own rather
+   * than through the whole id, which the random blocks would carry regardless.
+   */
+  test("ids minted in one millisecond differ in the counter", () => {
+    vi.spyOn(Date, "now").mockReturnValue(1_700_000_000_000);
+    try {
+      const first = createCuid();
+      const second = createCuid();
+
+      expect(first.slice(1, COUNTER)).toBe(second.slice(1, COUNTER));
+      expect(counterOf(second)).toBe(counterOf(first) + 1);
+    } finally {
+      vi.restoreAllMocks();
+    }
+  });
+
+  /** Rolls over at one block's worth of values, so it always pads to four. */
+  test("the counter block is four characters", () => {
+    expect(createCuid().slice(COUNTER, FINGERPRINT)).toHaveLength(4);
+  });
+});
+
+/**
+ * Per-call randomness in the fingerprint's position carries the same entropy, so
+ * this is not a collision risk either. What it destroys is process identity —
+ * the field someone diagnosing a duplicate reads first, and the one the module
+ * chose deliberately over cuid v1's pid+hostname and explains at length. Made
+ * per-call it would still be four characters of the right alphabet, and it would
+ * be lying.
+ */
+describe("the fingerprint identifies the process", () => {
+  test("it is the same across every id in this process", () => {
+    const fingerprints = new Set(
+      Array.from({ length: 100 }, () => fingerprintOf(createCuid())),
+    );
+
+    expect(fingerprints.size).toBe(1);
+  });
+
+  /**
+   * The other half: constant is not enough, it has to be *this process's*. A
+   * literal would satisfy the test above forever.
+   *
+   * Fresh module instances stand in for fresh processes. Five of them, asserting
+   * only that they are not all equal — two would be a one-in-1.68-million flake,
+   * five is (1/36^4)^4, and a hardcoded fingerprint still yields exactly one
+   * distinct value.
+   */
+  test("a fresh process gets a different one", async () => {
+    const fingerprints = new Set<string>();
+    for (let i = 0; i < 5; i++) {
+      vi.resetModules();
+      const fresh = await import("./defaults");
+      fingerprints.add(fingerprintOf(fresh.createCuid()));
+    }
+
+    expect(fingerprints.size).toBeGreaterThan(1);
+  });
+});
+
+/**
+ * 256 is not a multiple of 36 — it is 7*36 + 4 — so `byte % 36` without the
+ * resample delivers the first four symbols of the alphabet 8 times in 256 and
+ * the other thirty-two 7 times in 256. That is 12.5% above uniform for `0123`,
+ * and it is invisible to any assertion about a single id.
+ *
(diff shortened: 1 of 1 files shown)


nstfkc:
## Review — the tests are real and the mutation claims hold; two of the file's own justifying claims are wrong

Reproduced the whole table independently rather than reading it. Every headline claim about *behaviour* checks out. The two problems are both in the reasoning the file writes down to justify itself, not in what it tests.

### The mutation claims reproduce exactly

Applied each mutation to `packages/gemi/orm/defaults.ts` and ran the new file:

| mutation | `defaults.test.ts` | body claims |
|---|---|---|
| `counter++` → `0` | 2 failed \| 28 passed | 2 failed |
| fingerprint per-call | 1 failed \| 29 passed | 1 failed |
| fingerprint hardcoded `"a045"` | 1 failed \| 29 passed | 1 failed |
| `byte >= 252` → `byte >= 256` | 1 failed \| 29 passed | 1 failed |
| timestamp not the clock | 1 failed \| 29 passed | 1 failed |
| `dbgenerated` dropped from split | 1 failed \| 29 passed | 1 failed |
| `@updatedAt` early return removed | 2 failed \| 28 passed | 2 failed |
| drop the `c` prefix | 3 failed \| 27 passed | — |
| halve the random suffix | 8 failed \| 22 passed | — |
| clean | 30 passed | 30 passed |

And the other half — that three of these were invisible before. Same CI exclusion set, plus `--exclude '**/orm/defaults.test.ts'`:

```
clean                   Test Files  66 passed (66)   Tests  1579 passed | 1 skipped (1580)
counter-frozen          Test Files  66 passed (66)   Tests  1579 passed | 1 skipped (1580)
fingerprint-per-call    Test Files  66 passed (66)   Tests  1579 passed | 1 skipped (1580)
no-resample             Test Files  66 passed (66)   Tests  1579 passed | 1 skipped (1580)
```

Identical counts to a clean run in all three cases. The gap was real and is now closed.

### Checks

- `bun --bun vitest run` (CI's exclusion set) — **67 files, 1609 passed / 1 skipped**
- template `app/models`, `TZ=UTC` — **17 passed / 3 skipped, 640 tests**
- new file alone: 30 passed in 139 ms; **25 consecutive runs, 30/30 every time** — the frequency test is not flaky

Diff is one added file, no production code touched. Restored `defaults.ts` and `schema.ts` after every mutation; tree verified clean.

---

### 1. The bias test's stated standard deviation is wrong by 2.3x — the test is still correctly sized

Lines 196–199 (and the same paragraph in the PR body) say the share of `0123` over 200,000 symbols has "a standard deviation of 0.00031", making 0.125 "44 standard deviations away" and the ±0.004 window "12.7 sd of headroom".

`sqrt(p(1-p)/n)` with `p = 4/36`, `n = 200_000` is **0.000703**. Measured it over 60 independent sampling runs of the real generator:

```
symbols per run   : 200000
observed mean     : 0.111147
observed min      : 0.109230
observed max      : 0.112570
observed sd       : 0.000736
theoretical sd    : 0.000703 = sqrt(p(1-p)/n)
comment claims sd : 0.000310

window +-0.004 = 5.7 sd   (comment says 12.7)
0.125 is 19.8 sd from uniform  (comment says 44)
mutant clears upper bound by 14.1 sd
```

0.00031 is the sd you get at `n ≈ 1_000_000`, so this looks like `SYMBOLS` being taken as 1e6 rather than `25_000 × 8`.

The conclusion survives: ±0.004 is 5.7 sd, a two-sided flake rate around 1.2e-8, and the unresampled mutant clears the upper bound by 14 sd — which is why it dies reliably. The tolerance is fine. **Only the arithmetic needs fixing**, in the comment and in the body. It matters because that paragraph is the stated basis for the number `0.004`, and the next person to widen or narrow the window will do it from these figures.

### 2. The exhaustive `Record<DefaultKind, boolean>` is not a safety net in this repo

Lines 214–217 claim that adding a `DefaultKind` without deciding which side of the split it falls on "is a type error rather than an untested branch". It isn't, and nothing in CI makes it one.

`packages/gemi/tsconfig.json` has `"exclude": [..., "**/*.test.ts", ...]`, and `typecheck` is `tsc --noEmit -p tsconfig.json` — so `defaults.test.ts` is never typechecked. Vitest transpiles without checking types. The `Template type tests` CI step runs `test:types` in `templates/saas-starter` against `.test-d.ts` files, which this is not.

Demonstrated by adding a seventh kind to `orm/schema.ts`:

```
export type DefaultKind =
  | "autoincrement" | "cuid" | "uuid" | "now" | "dbgenerated" | "value"
  | "sequence";
```

Result with the record left untouched:

```
Test Files  1 passed (1)
      Tests  30 passed (30)
$ tsc --noEmit -p tsconfig.json
typecheck exit=0
```

Green on both. `Object.entries(expected)` simply doesn't yield the new kind, so it generates no case and says nothing — precisely the "untested branch" outcome the comment says it prevents.

This is a pre-existing config limitation rather than something the PR breaks, and the pattern is the right one. But the file asserts a guarantee it does not currently have, which is the same class of problem the PR was opened to fix. Either make it real — move the assertion into a `.test-d.ts` covered by a typecheck step, or add a runtime exhaustiveness check that fails on an unlisted kind — or soften the comment to describe intent rather than an enforced property.

---

### Smaller notes

- `hasClientSideValue` covers `autoincrement` but not `dbgenerated`; the `dbgenerated` side of the split reaches it only transitively through `isClientSideDefault`. The `dbgenerated-dropped` mutant dies on the `isClientSideDefault` table, so it is covered — a direct case would be one line.
- Slicing the id from the end rather than the front is the right call and the 2059 reasoning is correct. Worth keeping.
- The fresh-process fingerprint test via `vi.resetModules()` is genuinely stronger than a constancy check, and the flake bound is right: `Set.size > 1` fails only if all five draws collide, `(1/36^4)^4`.
- Test isolation is clean — `vi.resetModules()` in the fingerprint describe does not disturb the statically bound `createCuid` used by the later frequency test, confirmed by 25 green runs.

### Verdict

The behaviour is pinned, the mutants die, nothing else moves, and the suite and template counts match the body exactly. Good tests. Fix the standard deviation arithmetic, and either implement or downgrade the exhaustiveness claim; neither requires touching what the tests actually assert.

nstfkc:
Content landed on `feat/orm` in 1c2741f, so this is closed rather than merged.

This PR targeted `tests/orm-bytes-conflict-key` (#260) / `tests/orm-many-to-many-collection` (#264) — branches that had already merged into `feat/orm` and had no open PR of their own. Merging it would have put the work one more level from the trunk while reporting `MERGED`: the pooling failure `plans/orm/README.md` documents and the runbook there exists to catch.

So the stack tip was merged into `feat/orm` directly, which lands both commits:

```
156e642  Eleven template suites loaded a Prisma client to read an env var   (#260)
b3a36ed  Every cuid invariant except length and prefix was unguarded        (#264)
```

No content was dropped — `git rev-list --no-merges origin/feat/orm..` is empty for both branches now. CI is green on the result (`sqlite` and `postgres`, on the proposal and the merge result).

Filed #267 for the branch hygiene this is an instance of.

nstfkc:
Correction to the comment above: the branch-hygiene issue is **#268**, not #267. (#267 is a separate find from the same audit — JSON path filters built on a branch and never merged.)

PR #282, opened 2026-07-30
Prefetch routes ahead of a navigation

nstfkc:
Adds `usePrefetch`, and a `prefetch` prop on `Link` that warms the target route's page data, stylesheets and component chunks before it is clicked. A navigation that lands on a warmed route commits from the cache instead of waiting on a request.

Strategies are `hover`, `intent`, `viewport` and `render`, and can be combined — `["viewport", "intent"]` warms the link on the way past and warms it again on approach once the cached payload has gone stale. Prefetching is off unless asked for, stands down on metered or very slow connections, and skips links pointing at the current path, whose navigation is shallow and never fetches.

The payload is requested without the `x-gemi-from` header: the route a link is warmed from is not necessarily the one it is clicked from, and a partial response computed against the wrong base has nothing sound to merge onto. A full payload is always safe to commit.

Claude-Session: https://claude.ai/code/session_01RbZVhufKvXQ4TSL5FcdzoT


packages/gemi/client/PrefetchCache.ts
```diff
@@ -0,0 +1,71 @@
...
+      .then((payload) => {
+        // A failed prefetch is dropped rather than remembered: the navigation
+        // falls back to its own request and the next hover gets to retry.
+        if (payload == null && this.entries.get(url) === entry) {
+          this.entries.delete(url);
+        }
+        return payload;
+      });
+
+    this.entries.set(url, entry);
```
nstfkc, line 80: **Entries are never evicted — unbounded growth.**

An entry leaves the map only when it's consumed by `take`, or when its load fails. A payload that is prefetched and never clicked is retained for the lifetime of the session, and each one is a **full page payload**.

That's precisely the `viewport`/`render` case — a long list where most links are never clicked. On an infinite feed this grows without bound.

Suggest sweeping expired entries on `prime` (cheap, it already walks in), and/or capping the map with LRU eviction.

packages/gemi/client/PrefetchCache.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * How long a prefetched payload stays usable. Long enough to cover the gap
+ * between hovering a link and clicking it, short enough that a link left on
+ * screen doesn't hand a navigation minutes-old data.
+ */
+export const PREFETCH_TTL = 30_000;
```
nstfkc: **Nothing invalidates the cache.**

A payload cached at hover time can be committed up to 30s later — after a mutation has changed the underlying data. That's worse than a stale render, because `ClientRouter` also calls `hydrate(prefetchedData)` on commit, so the stale snapshot is seeded into the *query* cache and `useQuery` consumers inherit it.

Suggest a `clear()` on the cache, called from the mutation path and on auth/locale change. Failing that, a shorter TTL — 30s is a long time to hold authorized page data on the assumption nothing changed. At minimum the window should be documented for consumers.

packages/gemi/client/PrefetchCache.ts
```diff
@@ -0,0 +1,71 @@
...
+ * consumes one becomes the live route data, so keeping a copy around would only
+ * let a later visit render from a stale snapshot.
+ */
+export class PrefetchCache {
+  private entries = new Map<string, Entry>();
+  private ttl: number;
+
+  constructor(ttl: number = PREFETCH_TTL) {
+    this.ttl = ttl;
+  }
```
nstfkc: Nit: the `ttl` parameter has no caller — `ClientRouterContext` constructs with `new PrefetchCache()` and the tests reach for fake timers instead. Either use it in the tests (simpler than `vi.useFakeTimers`) or drop it.

packages/gemi/client/ClientRouterContext.tsx
```diff
@@ -266,10 +280,48 @@ export const ClientRouterProvider = (
...
+  /**
+   * Warms everything a navigation to `target` would need: the route's page
+   * data, its stylesheets and its component chunks.
+   *
+   * The payload is requested *without* the partial-render header, because the
+   * route on screen when the link is prefetched is not necessarily the one it
+   * will be clicked from — a partial response computed against the wrong base
+   * has nothing sound to merge onto. A full payload is always safe to commit.
+   */
+  const prefetchRoute = async (target: PrefetchTarget) => {
```
nstfkc, line 297: **No `Purpose: prefetch` header on these requests.**

A prefetch runs the server's route loaders for real. With `render` or `viewport` on a list page, that multiplies backend load by the number of links — and it's indistinguishable from genuine traffic in logs, analytics and rate limiters. Any loader with a side effect (a view counter is the obvious one) now fires for pages nobody opened.

Sending `Purpose: prefetch` (or `Sec-Purpose`) on the prefetch `fetch` gives applications a way to shed the load and keeps those side effects honest. Cheap to add here, hard to retrofit once people have shipped `viewport` prefetching.

packages/gemi/client/ClientRouterContext.tsx
```diff
@@ -266,10 +280,48 @@ export const ClientRouterProvider = (
...
+    const routePath = getRoutePathnameFromHref(pathname);
+    if (!routePath) {
+      return;
+    }
+
+    const url = routeDataUrl({ pathname, search, localeSegment });
+
+    await prefetchCache.prime(url, async () => {
+      const [response] = await Promise.all([
+        fetch(url),
```
nstfkc: **One failing asset warm discards a good payload.**

`load()` resolves a `Promise.all` over the JSON fetch *and* `fetchRouteCSS`. If `fetchRouteCSS` rejects — a CSS chunk 404s, a network blip — the whole `load()` rejects, `prime` catches it and deletes the entry, and the page data that was fetched successfully is thrown away. The click then pays full price for a request that had already completed.

Await the JSON, and fire the asset warming alongside it rather than as a precondition for it (`allSettled`, or just don't join them).

packages/gemi/client/ClientRouterContext.tsx
```diff
@@ -266,10 +280,48 @@ export const ClientRouterProvider = (
...
+        fetchRouteCSS(routePath),
+        ...(routeManifest[routePath] ?? []).map((view) => {
+          window?.loaders?.[view]?.();
+        }),
```
nstfkc: The arrow has a block body with no `return`, so `Promise.all` receives `undefined` for each loader — the chunks are fired but not awaited, and `prime` resolves before any of them land. (Same shape as the pre-existing code in `ClientRouter`, so this is inherited rather than introduced.)

Harmless in practice for a prefetch, but the `Promise.all` claims a guarantee the code isn't making. Either `return` the loader result or leave them out of the join.

packages/gemi/client/ClientRouter.tsx
```diff
@@ -226,49 +228,67 @@ const Routes = (props: { componentTree: ComponentTree }) => {
...
       const from = renderedRouteRef.current;
       setIsFetching(true);
-      let res = { ok: false, json: async () => ({}) } as Response;
-      try {
-        const result = await Promise.all([
-          fetch(url, { headers: { [PARTIAL_RENDER_HEADER]: from } }),
+
+      const warmRouteAssets = () =>
+        Promise.all([
           fetchRouteCSS(pathname),
```
nstfkc: `prefetchRoute` resolves the route *pattern* first (`getRoutePathnameFromHref`) before calling `fetchRouteCSS`; this call still passes the concrete pathname. `fetchRouteCSS` does `routeManifest[routePath]` and returns early on `undefined`, so for any dynamic route (`/posts/123`) this is a no-op.

Pre-existing, but the PR now puts the correct and incorrect calls side by side, and the net effect is that a *prefetched* navigation gets its CSS while a plain one doesn't. `routerState.routePath` is already in scope — one-word fix.

packages/gemi/client/Link.tsx
```diff
@@ -84,8 +126,120 @@ export const Link = memo(<T extends keyof Views>(props: LinkProps<T>) => {
...
+  // a plain prop, and the spread below would otherwise hand the element to one
+  // of us and null to the other.
+  const setAnchorRef = (node: HTMLAnchorElement | null) => {
+    anchorRef.current = node;
+    if (typeof ref === "function") {
+      ref(node);
+    } else if (ref) {
+      ref.current = node;
+    }
+  };
```
nstfkc: **Ref identity churn detaches the consumer's ref on every render.**

`setAnchorRef` is a new function each render, so React detaches and reattaches. `Link` re-renders on every navigation via `useRouteTransition` (`isTransitioning`), which means a caller passing a ref *callback* gets `null` and then the node again on every navigation in the app — and `anchorRef.current` briefly nulls with it.

`useCallback(..., [ref])` fixes it.

packages/gemi/client/Link.tsx
```diff
@@ -18,13 +26,39 @@ type Views = {
...
+
 type LinkBaseProps<T extends keyof Views> = Omit<
   ComponentProps<"a">,
   "href"
 > & {
   active?: boolean;
   href: T;
   hash?: string;
+  /** Off unless set. `false` is accepted so it can be driven by a variable. */
+  prefetch?: PrefetchStrategy | PrefetchStrategy[] | false;
```
nstfkc: Nit: the array form is the one the doc comment recommends (`["viewport", "intent"]`), but an inline array literal is a fresh identity every render and defeats the `memo` on `Link`. Worth a note in the docs to hoist it to a module constant — or normalise the prop to `strategyKey` before the comparison.

nstfkc (commented):
Reviewed on the branch: 16/16 new tests pass, `oxlint` clean, `tsc -p tsconfig.browser.json` clean.

## What's right

- **Skipping `x-gemi-from` on prefetch is the correct call.** The committed payload has no `partial`, so `mergeCarriedSegments` with `carriedViews: []` does a clean full replace, and `renderedRouteRef` is still updated afterwards.
- **Search serialization deliberately mirrors `useNavigate`** (raw `URLSearchParams`) rather than `Link`'s `normalizeSearch` — the right choice, since the key has to match what the click produces, even though it inherits `useNavigate`'s `a=undefined` quirk.
- **The locale round-trip holds.** `usePrefetch` mirrors `useNavigate`'s defaultLocale-stripping, and the navigation URL is re-parsed from what `useNavigate` pushed, so the two agree in both the prefixed and unprefixed cases. Fragile by construction — the shared `routeDataUrl` helper plus its tests are the right mitigation.
- `prime` correctly dedupes in-flight and fresh entries, and the `entries.get(url) === entry` guard avoids clobbering a newer entry on failure.
- `if (payload)` replacing `if (res.ok)` also turns a `null` JSON body into a no-op instead of a destructure crash.

Inline comments below. The three I'd want resolved before this goes near a `viewport`-heavy page are cache eviction, invalidation, and the `Purpose: prefetch` header — see those threads.

## Two things that don't anchor to a line

**Test coverage stops at the two pure units.** `PrefetchCache` and `routeDataUrl` are well covered, but the risky logic isn't: strategy selection, the intent timer, the observer, the connection gate, and the take-then-commit path in `ClientRouter`. The minimum worth adding is the two branches the restructure introduced — a prefetched payload commits without issuing a `fetch`, and a null (failed) prefetch falls through to the normal request.

**`docs/navigation.md` isn't updated.** It enumerates `Link`'s props (`params`, `search`, `hash`, `active`) and the navigation hooks; `prefetch` and `usePrefetch` are public API and should land with their docs.

## Drive-by, pre-existing

`onClick?.(e)` is called twice in `Link`'s click handler (once before the `hash === ""` check, once after). Outside this diff so I can't anchor it, but this PR is careful to call each new user handler exactly once — natural moment to fix the odd one out.

packages/gemi/client/PrefetchCache.ts
nstfkc: Fixed in 8df3228. `prime` now sweeps expired entries and evicts the oldest beyond a 12-entry ceiling (`PREFETCH_MAX_ENTRIES`), so an infinite feed retains at most 12 payloads. Covered by three tests: stays under the ceiling, drops oldest first, and sweeps expired entries rather than counting them toward the cap.

packages/gemi/client/PrefetchCache.ts
nstfkc: Fixed in 8df3228, both ways you suggested. `clear()` added and called from `useMutation` on every successful write (which covers `Form`, `usePost`/`usePut`/`usePatch`/`useDelete` and `useUpload`), and the TTL is down from 30s to 10s — still covers hover-to-click and scroll-to-click, but bounds the window on data changed elsewhere. Locale needs no call: the locale segment is part of the cache key, so a locale change misses rather than serving the wrong one. Both the window and the `hydrate` interaction are now documented under "Staleness" in docs/navigation.md.

packages/gemi/client/PrefetchCache.ts
nstfkc: Dropped — `PREFETCH_TTL` is a module constant now and the tests keep fake timers, which they need anyway for the sweep and expiry cases.

packages/gemi/client/ClientRouterContext.tsx
nstfkc: Added in 8df3228: the prefetch `fetch` sends `Purpose: prefetch`. Not `Sec-Purpose` — that prefix is forbidden to scripts, so setting it would silently no-op. Verified over CDP: the prefetch request carries `Purpose: prefetch` and no `x-gemi-from`, the real navigation the reverse. Documented with a view-counter example under "What it costs".

packages/gemi/client/ClientRouterContext.tsx
nstfkc: Fixed in 8df3228. `fetchRouteCSS` and the chunk loaders now run alongside the payload rather than inside its `Promise.all`, so a 404 on a stylesheet cannot discard page data that arrived fine.

packages/gemi/client/ClientRouterContext.tsx
nstfkc: Gone with the fix above — the loaders are fired in a plain loop now, so nothing claims to await them.

packages/gemi/client/ClientRouter.tsx
nstfkc: Fixed in 8df3228: `fetchRouteCSS(routerState.routePath)`. Dynamic routes now get their CSS on a plain navigation, not only on a prefetched one.

packages/gemi/client/Link.tsx
nstfkc: Fixed in 8df3228 — wrapped in `useCallback(..., [ref])`.

packages/gemi/client/Link.tsx
nstfkc: Documented rather than normalised: the array example in docs/navigation.md is hoisted to a module constant with a comment saying why. The prop also accepts `readonly PrefetchStrategy[]` so the natural `as const` spelling typechecks.

nstfkc:
Pushed 8df3228 — all nine inline threads plus the three general points are addressed, replies on each thread.

**The three you flagged as blocking**

- **Eviction** — `prime` sweeps expired entries and evicts the oldest beyond a 12-entry ceiling.
- **Invalidation** — `clear()` on the cache, called from `useMutation` on every successful write; TTL 30s → 10s. Locale is safe by construction (it is part of the key).
- **`Purpose: prefetch`** — added. Deliberately not `Sec-Purpose`: that prefix is forbidden to scripts, so it would have silently no-opped.

**The two that did not anchor to a line**

- **Test coverage** — extracted `loadRoutePayload` out of the `ClientRouter` effect so the restructured branches are testable without a DOM harness. Nine new tests, including the two you named: a warmed payload commits with `fetch` never called, and a failed prefetch (`take` resolving `null`) falls through to the normal request. Suite is 65 tests, all passing.
- **Docs** — `docs/navigation.md` has a Prefetching section: the strategy table, combining, the hoisting note from your last comment, cost and side effects with a view-counter example, staleness and what clears the cache, and `usePrefetch`.

**Drive-by** — duplicate `onClick?.(e)` removed.

Re-verified end to end in headless Chrome after the refactor: `render` warms on mount, `hover` warms on mouseover, a click on a warmed link issues no request and commits, an unprefetched link still fetches normally, and the headers are right on both paths.

PR #285, opened 2026-07-31
Make useQuery suspense-ready, with streaming SSR

nstfkc:
## Summary

Two stacked features that together make data fetching suspense-first end to end:

### 1. Suspense-ready `useQuery` (default on)

- A query with no cached data **suspends its route segment**; `data` is non-nullable by default. Opt out per query with `{ suspense: false }` (`lazy: true` implies it).
- Every route segment is wrapped in `Suspense` + an error boundary, keyed by tree slot so a suspending navigation keeps the previous page on screen. Views can export `Loading` and `Error` for the segment's fallback/error UI.
- The promise is **thrown**, not passed to `use()` — React 19 never retries `use()` on client-created promises; the thenable-throw protocol retries reliably (pinned by tests).
- HTTP failures throw a `QueryError` (`status`, `body`, `path`) into the segment's boundary; resetting clears stored errors.
- Prefetched data (server `Query.prefetch`, `<Link prefetch>` payloads) lands in the same cache and wakes suspended readers via `hydrate()` → `settle()` — no duplicate `/api` round trips. Stale data never suspends (background revalidation unchanged).

### 2. Streaming SSR (no server waterfalls)

- Document responses stream: the shell — layout chrome, each segment's `Loading` export — is sent as soon as the route handlers finish, and query data streams in behind it. Measured on the demo: **TTFB 860ms → 6ms** on the prefetched page, blank-until-1.2s → shell-at-4ms on the un-prefetched one.
- **Fetch start is decoupled from render**: `Query.prefetch` fires the moment it's called (live store — also fixes prefetches after a handler's first `await` being silently dropped), and every query the render's first pass reaches starts immediately, in parallel. Total wait is `max(query times)`, never a sum.
- `Query.prefetch` is now a **performance optimization, not a requirement**. The remaining waterfall cases (query nested under another suspending query, conditionally rendered on fetched data) log a dev warning with the delay paid and the exact prefetch call that removes it.
- `Query.instant` is the explicit "block the shell, data in first paint" opt-in.
- Payload scripts (`__GEMI_STREAM__.push`) are injected into the stream strictly **before** their segment's reveal chunk, so hydration can never see content without its data. Crawler UAs get the fully settled document (`stream.allReady`); a 10s deadline aborts hung queries into client rendering.
- A failing api handler rejects the entry — the segment falls back to client rendering and the browser's fetch surfaces the error into the view's `Error` boundary. (Previously a failing prefetch stored the error `Response` object as data.)
- `.json` navigation payloads keep the settle-everything contract.

## Breaking changes

- `useQuery` suspends by default — existing call sites relying on `loading` need `{ suspense: false }`.
- A rejected `Query.prefetch` no longer fails the whole document request; the affected segment error-bounds instead.
- `Query.prefetch` no longer blocks the document response; use `Query.instant` where first-paint data is required.

## Test plan

- 125 unit/integration tests green (store dedupe + settle ordering, stream injector ordering incl. payload-before-reveal, 4 `renderToReadableStream` integration tests, suspense/transition/SSR suites).
- Live-verified on saas-starter: streaming timings, bot behavior, error paths, `.json` payloads, prefetch variants (`/about` locale search), partial rendering.
- Remaining manual step: real-browser pass over `/suspense/*` for hydration warnings under StrictMode.


docs/data-fetching.md
```diff
@@ -40,6 +37,68 @@ export default function Todos() {
...
+  return <ul>…</ul>;
+}
+
+// Shown while the segment's queries (or its chunk) load.
+export function Loading() {
+  return <TodosSkeleton />;
+}
+
+// Shown when a query throws — receives { error, resetErrorBoundary }.
+export function ErrorFallback({ error, resetErrorBoundary }) {
```
nstfkc: rename this to just `Error`

packages/gemi/client/QueryResource.ts
nstfkc, line 1: This QueryResource might need to be rewritten, not sure if useExternalSyncStore works correctly with transitions. Research this.

docs/data-fetching.md
nstfkc: Renamed to `Error` in 708d3d0 — the router now reads `mod.Error`, docs and comments updated.

packages/gemi/client/QueryResource.ts
nstfkc: Researched and pinned as tests in `useQuery.transition.test.tsx` (708d3d0) — no rewrite needed.

The uSES caveat is real but narrow: store updates are always **urgent**, so a write landing while a transition renders makes React restart that transition. That costs a wasted render; it never produces wrong UI. This design keeps the exposure small in two ways:

1. Render-phase reads never write the store synchronously — `read()` starts *silent* fetches, so rendering (including the attempts React discards) can't invalidate its own snapshot.
2. A suspended component was never mounted, so it has **no uSES subscription** — it's woken by its thrown promise resolving, not by the store. The transition's suspense path doesn't go through uSES at all.

Verified against the router shape on React 19.2.3:
- a transition into a suspending query keeps the previous page on screen until it resolves;
- a store write during a pending navigation urgently re-renders the *visible* page and the pending navigation still lands;
- hydrating the suspended query mid-transition (Link prefetch payload) commits the navigation without waiting on its own fetch.

nstfkc (commented):
## Review

Reviewed at `b70cd8a`. Findings ordered by severity; #1 and #2 are reproduced against the real classes.

## Overview

Two stacked changes:

1. **Suspense-first `useQuery`** — a query with no cached data throws a promise instead of returning `loading: true`. `ClientRouter.Route` wraps every route segment in `Suspense` + `ErrorBoundary`, with fallbacks from the view module's optional `Loading`/`Error` exports. Boundaries are keyed by tree *slot* so a suspending navigation keeps the previous page on screen.
2. **Streaming SSR** — `ctx.prefetchedResources`/`prefetchPromiseQueue` are replaced by a live per-request `ServerQueryStore`. Fetches start at call time rather than render time; `renderToReadableStream` sends the shell immediately and `injectQueryPayloads` interleaves `__GEMI_STREAM__.push(...)` scripts with React's chunks, queuing each payload *before* settling the promise that unblocks its segment.

The architecture is sound and the write-up is unusually honest about its own invariants. Three things worth singling out as genuinely well done: `QueryResource.peek` vs `getVariant` (separating the render-safe read from the fetching read is exactly the right cut, and it's the fix for React discarding suspended renders); the `inflight` set as the dedupe mechanism rather than the `loading` flag, with the reasoning documented at the field; and the decision to *throw* the promise rather than use `use()` — correct for React 19, and pinned by tests. `serverQueryFetcher` converting a `RequestBreakerError` `Response` into a `QueryError` fixes a real prior bug where the error object was stored as data.

**Test suite:** the PR's new tests all pass. The 6 failures + 3 import errors in the package suite (`createRouteManifest`, `createComponentTree`, `createFlatViewRoutes`, `App.test.ts`) reproduce identically on `origin/main` — pre-existing, not from this PR.

---

## Findings

### 1. Falsy data causes an unbounded refetch loop under suspense — High

`QueryResource.read()` uses truthiness to decide whether data exists (`QueryResource.ts:143`). An endpoint that legitimately returns `null`, `0`, `false`, or `""` never satisfies that check, so each render attempt allocates a fresh deferred and starts another fetch. Because `settle()` wakes the reader on every landing, React re-renders → reads → fetches, forever.

Confirmed against the real class — 6 suspend/retry cycles produced **6 fetches**, and the cached state stayed `{data: null}`:

```
fetch calls for null data: 6   (expected 1)
cached state: { loading: false, data: null, error: null, version: … }
```

`getVariant` compounds it: the mount effect hits the same `if (!variant.data) resolveVariant(...)` branch at `QueryResource.ts:185`. Pre-suspense this was throttled by the effect cycle; under suspense it's a tight loop that will hammer the endpoint.

The `useUser` patch in this PR (`suspense: false`, "an anonymous visitor has no `/auth/me` data and never will") reads like this class of problem being handled at one call site rather than at the root.

**Fix:** track presence explicitly rather than inferring it from truthiness — e.g. a `hasData` flag on `State`, or `store.has(variantKey) && state.error == null` as the "settled with a value" test. `read`, `getVariant`, and `hydrate` all need the same predicate.

### 2. `hydrate()` drops falsy streamed payloads, stranding the suspended reader — High

`QueryResource.ts:55` — `if (!data) continue;`. A streamed `__GEMI_STREAM__` payload whose data is `null`/`false`/`0` is skipped, so `settle(variantKey)` never runs and the promise the segment threw never resolves. Confirmed:

```
settled after hydrate(false): false   (expected true)
peek: undefined
```

Same root cause as #1, and it's *deliberate* today — `QueryResource.test.ts` has "ignores empty, nullish and falsy payloads without notifying". That was harmless when nothing suspended on hydration; it isn't now. Worth updating that test alongside the fix, since it currently locks in the broken behaviour.

### 3. The document body now varies by User-Agent, with no `Vary: User-Agent` — Medium

`ViewRouterServiceContainer.ts:312` — `isBotUserAgent(userAgent) || noStream` changes both `progressiveChunkSize` and whether the response awaits `allReady`, producing a materially different body for the same URL. The `.json` path correctly appends `Vary: PARTIAL_RENDER_HEADER` (line 591), but the document path adds no `Vary`.

With `cache:public` (`public, max-age=864000` per `docs/middleware.md`) a shared cache can store a browser-shell document and serve it to a crawler — the crawler then indexes `Loading` skeletons and `<div hidden>` reveal blocks, which is precisely what the bot branch exists to prevent. The `no-stream` doc example in this PR is `middlewares = ["cache:public", "no-stream"]`, so cache + streaming routes are expected to co-occur.

**Fix:** append `Vary: User-Agent` when the streamed/settled decision is UA-derived (not needed for `no-stream`, where both audiences get the same body). Note this largely defeats CDN caching for those routes — which may itself argue for keying off something coarser than raw UA.

### 4. Streamed payloads are adopted in an effect, after the first hydration render — Medium *(unverified in a browser)*

`__GEMI_STREAM__` is consumed only in `QueryManagerProvider`'s `useEffect` (`QueryManagerContext.tsx:80`) — there is no other consumer. Effects run after the render phase, but suspension happens *during* it. So on the first hydration pass a streamed segment's `useQuery` reads an empty cache, suspends, and starts a duplicate `/api` fetch for data already inlined in the document.

The stream-ordering invariant documented in `streamQueryInjection.ts` is real, but it guarantees *script-before-reveal in the byte stream* — not *cache-populated-before-hydration-render*, which is what matters here.

In practice this may be masked: `viewImportMap` entries are `lazy(() => loadViewModule(...))`, so the segment probably suspends on its chunk first, letting the provider commit and drain the buffer before the query renders. That's timing-dependent rather than guaranteed. This is the one finding not confirmable without a real browser — and it lines up with the open item in the description ("real-browser pass over `/suspense/*` for hydration warnings under StrictMode"), so it's worth checking specifically for a duplicate `/api` request in the network tab, not just for warnings.

**Cheap fix:** drain the buffer synchronously when the provider module initialises (or during first render), rather than in an effect.

### 5. Smaller notes

- **`docs/data-fetching.md`** says `error` is "only populated with `suspense: false`". Not quite: `useQuery.ts:440` only throws when `state.error && !state.data`, so a refetch that fails while stale data is on screen returns a populated `error` under suspense too. That's good behaviour — the doc just under-describes it.
- **Blank segment when a view has no `Loading` export** — `ClientRouter.tsx:143` falls back to `null`. On a hard load that's an empty region with no indication anything is happening. Worth a line in the docs that `Loading` is effectively required for streamed routes.
- **`httpDev.ts` port change** (`process.env.PORT || 5173`) is a reasonable fix but unrelated to this PR's scope; it'll be invisible in the changelog under this title.
- **`clearErrors()` is app-wide** — resets errors on every resource, not just the failing one. Checked `react-error-boundary@6.1.1`: `onReset` is guarded by `didCatch`, so this does *not* fire on ordinary navigation. Low impact; noting only because the blast radius is wider than the boundary that reset.

---

## Verdict

The design is right and the test suite is substantial — `ServerQueryStore.test.ts` and `QueryResource.test.ts` cover the ordering and dedupe invariants carefully, including the `htmlSafeJson` script-breakout case. Escaping in the injected payloads is correct — escaping `<` to `<` is sufficient to prevent script-context breakout, and React escapes `bootstrapScriptContent` itself.

**#1 and #2 should block merge** — they're a single root cause (truthiness as a proxy for presence), reproducible against the real classes, and since this PR makes suspense the *default*, any endpoint returning a falsy body will loop against the server on first render. The fix is small and localised to `QueryResource`; the accompanying test change is to `QueryResource.test.ts`'s falsy-payload case.

**#3 and #4 are worth resolving before this ships as a non-RC**, but neither is a correctness failure in the happy path.

nstfkc:
All findings addressed in 5d96dae:

**#1 + #2 (falsy data / falsy hydrate — the blockers):** fixed at the root you identified — presence is now an explicit `hasData` flag on cached state, and every presence check (`read`, `getVariant`, `hydrate`, `mutate`, and all of `useQuery`'s) goes through it; only `undefined` means "no value". Your repro is pinned as a regression test: a resolved `null` body yields exactly **1 fetch across 6 read attempts** with `hasData: true`, and `hydrate({ "": false })` settles a suspended reader. The old "ignores empty, nullish and falsy payloads" test locked the broken behavior in, as you noted — it's reworked to "ignores *absent* payloads, adopts falsy bodies". (Agreed on the `useUser` observation: `suspense: false` there was masking this class at one call site; it still stands on its own merits — anonymous `/auth/me` wants the `{user: null, loading}` shape — but it's no longer load-bearing.)

**#3 (Vary):** document responses on streaming routes now append `Vary: User-Agent`; `no-stream` routes serve one body to everyone and stay Vary-free, exactly per your scoping. The CDN-hostility point is real — noted it in the code comment; keying on something coarser than raw UA is a good follow-up if it bites.

**#4 (effect-timed drain):** the `__GEMI_STREAM__` buffer is now drained synchronously during `QueryManagerProvider`'s first render (ref-guarded, idempotent under StrictMode's double-invoke, and nothing is subscribed to the store yet at that point so no render invalidates mid-pass). You were right that the documented invariant covered bytes-on-the-wire, not cache-before-hydration-render — the drain timing was the missing half.

**#5:** docs updated — the `error` row now describes the stale-data-refetch-failure case (populated under suspense, only throws with nothing to show), and `Loading` is called out as effectively required for streamed routes. `clearErrors`' app-wide blast radius and the `httpDev` port change: acknowledged, leaving both as-is (the former guarded by `didCatch` as you verified; the latter will get a line in the release notes).

129 tests green, tsc clean. Say the word for `rc.5` — folio is on rc.4, which has the falsy-data loop.

PR #295, opened 2026-08-01
keepPreviousData under suspense: defer the variant key so variant changes never flash the fallback

### LINKED ISSUE #291: keepPreviousData should work under suspense: defer the variant key internally instead of requiring startTransition at every call site
Follow-up to #285. Under suspense, changing a query's `search`/`params` to an uncached variant suspends the component — the segment's fallback flashes in place of the previous rows unless the caller remembered to wrap the state update in `startTransition`. In the folio migration this forced a blanket `{ suspense: false }` on all 28 `keepPreviousData: true` call sites (pagination, search, filters) — the single largest remaining category of opt-outs, and the least interesting one: every site wants the same behavior ("keep showing what I have while the next page loads").

`useQuery` can provide that behavior itself. When `keepPreviousData` is on and the *current* variant has no cached data but a *previous* one does, read through `useDeferredValue(variantKey)`:

- the component keeps rendering the deferred (previous) variant's data — no suspension, no fallback flash;
- React re-renders in the background with the new variant, which suspends *that* attempt only, and commits when the data lands;
- callers get a pending signal for pager UI without owning a transition — either expose the internal `deferredVariantKey !== variantKey` as the existing `loading` flag under suspense (today it is always `false`), or as a new `isRefreshing`.

First-mount behavior is unchanged (no previous variant → suspend into the segment fallback as today), as is navigation (the router's transition already covers it).

## Notes

- `useDeferredValue` is already in the dependency budget (React 19); no new machinery.
- The escape hatch stays: `keepPreviousData: false` + suspense = current suspend-on-variant-change semantics, for callers who *want* the fallback (e.g. tab-like switches where stale rows would mislead).
- Docs: the migration guidance "wrap page changes in startTransition or opt out" collapses to "keepPreviousData just works"; the `useTransition` recipe remains for coordinating multiple queries.

## Acceptance

- A paginated list under suspense: clicking next page keeps the current rows rendered, exposes a pending flag, commits the new rows when resolved — with no `startTransition` in app code.
- Jsdom test pinning: variant change with cached previous variant does not unmount/flash the fallback; first mount still suspends.

nstfkc:
Closes #291

Stacked on `main`.

## Approach

`useQuery` now reads the store through `useDeferredValue` when `keepPreviousData` (the default) is combined with suspense (also the default). Both the resource path and the variant key are deferred together, so a `search` *or* `params` change:

- keeps rendering the previous variant's data in the urgent pass — the committed tree never re-suspends, so the segment fallback never flashes and no `startTransition` is needed at the call site;
- suspends only React's background re-render with the new keys (the new variant's fetch starts immediately via the render-phase `read`), and commits when the data lands;
- exposes the pending window through the existing `loading` flag (`deferred !== current`), which under suspense previously never reflected a variant change — pager UI gets its signal without owning a transition.

First mount is unchanged: deferred and current keys are equal, so an uncached query suspends into the segment fallback as before. The escape hatch also stays: `keepPreviousData: false` + suspense restores suspend-on-variant-change for callers who want the fallback (e.g. tab-like switches). Both `useDeferredValue` hooks run unconditionally to keep hook order stable; the deferred values are only used when deferring applies. No new dependencies — `useDeferredValue` is already in the React 19 budget.

## Tests

New `keepPreviousData under suspense` describe block in `packages/gemi/client/useQuery.suspense.test.tsx` (jsdom):

- a paginated list: first mount suspends into the fallback; clicking next page (plain `setState`, no transition) keeps the previous rows committed with `loading: true`, never shows the fallback, fires the `page=2` request immediately, commits the new rows on resolve, and never unmounts the component along the way;
- `keepPreviousData: false` restores suspend-on-variant-change (fallback shown in place).

`bun run test -- run client` — 13 files, 123 tests pass; `test:types` and `tsc --noEmit` clean; oxlint 0 errors. The 6 failing tests in `app/`/`services/`/`utils/` fail identically on `main` (verified via stash) and are unrelated.

### DIFF
--- a/packages/gemi/client/useQuery.suspense.test.tsx
+++ b/packages/gemi/client/useQuery.suspense.test.tsx
@@ -333,6 +333,95 @@ describe("keepPreviousData under suspense", () => {
     expect(mounts).toBe(1);
   });
 
+  test("a params change swaps the resource but keeps the same contract — previous rows, pending flag, no remount", async () => {
+    // Unlike a `search` change (one resource, new variant key), a `params`
+    // change swaps the entire `QueryResource`: new store, new subscription,
+    // new read promise. The committed tree stays subscribed to the *old*
+    // resource for the whole pending window and the background render is
+    // woken by the new resource's thrown promise — the invariants must hold
+    // across that swap too.
+    function ListTodos() {
+      const [listId, setListId] = useState(1);
+      const { data, loading } = useQuery("/lists/:listId/todos" as any, {
+        params: { listId } as any,
+      });
+      useEffect(() => {
+        mounts += 1;
+      }, []);
+      return (
+        <button type="button" onClick={() => setListId((id) => id + 1)}>
+          {`ids:${(data as any[]).map((t: any) => t.id).join(",")} loading:${loading}`}
+        </button>
+      );
+    }
+
+    const screen = render(
+      <Providers>
+        <ListTodos />
+      </Providers>,
+    );
+
+    // First mount is unchanged: suspend into the fallback.
+    expect(screen.queryByText("suspense-fallback")).not.toBeNull();
+    expect(net.fetchMock).toHaveBeenLastCalledWith("/api/lists/1/todos", {
+      cache: "default",
+    });
+    await net.resolve([{ id: 1 }]);
+    expect(screen.queryByText("ids:1 loading:false")).not.toBeNull();
+
+    // Switch lists — a plain urgent state update.
+    await act(async () => {
+      screen.getByRole("button").click();
+    });
+
+    // The old list's rows stay committed, the pending window is flagged, and
+    // the new resource's request went on the wire immediately.
+    expect(screen.queryByText("suspense-fallback")).toBeNull();
+    expect(screen.queryByText("ids:1 loading:true")).not.toBeNull();
+    expect(net.fetchMock).toHaveBeenLastCalledWith("/api/lists/2/todos", {
+      cache: "default",
+    });
+
+    await net.resolve([{ id: 2 }]);
+
+    expect(screen.queryByText("ids:2 loading:false")).not.toBeNull();
+    expect(mounts).toBe(1);
+  });
+
+  test("a fetch failure during the pending window throws into the error boundary", async () => {
+    // Pins the error path of the deferral: the background render suspends on
+    // the new variant, the fetch fails, and the retry render throws the
+    // stored error — the boundary replaces the previous rows. The previous
+    // variant's data does NOT mask the failure: under suspense the caller
+    // writes no error branch, so silently returning `error` while keeping
+    // the old page on screen would leave a failed page change with no
+    // visible signal at all. The boundary is the suspense contract's error
+    // channel, exactly as it was for a suspended variant change before the
+    // deferral existed.
+    vi.spyOn(console, "error").mockImplementation(() => {});
+    const screen = render(
+      <Providers>
+        <Boundary>
+          <PagedList />
+        </Boundary>
+      </Providers>,
+    );
+
+    await net.resolve([{ id: 1 }]);
+    expect(screen.queryByText("ids:1 loading:false")).not.toBeNull();
+
+    await act(async () => {
+      screen.getByRole("button").click();
+    });
+    // Pending window: previous rows visible, page 2 in flight.
+    expect(screen.queryByText("ids:1 loading:true")).not.toBeNull();
+
+    await net.resolve({ message: "boom" }, false);
+
+    expect(screen.queryByText("caught:QueryError")).not.toBeNull();
+    expect(screen.queryByText("ids:1 loading:true")).toBeNull();
+  });
+
   test("keepPreviousData: false restores suspend-on-variant-change", async () => {
     const screen = render(
       <Providers>
--- a/packages/gemi/client/useQuery.suspense.test.tsx
+++ b/packages/gemi/client/useQuery.suspense.test.tsx
@@ -1,6 +1,13 @@
 /** @vitest-environment jsdom */
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
-import { Component, Suspense, act, useContext } from "react";
+import {
+  Component,
+  Suspense,
+  act,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
 import type { PropsWithChildren, ReactNode } from "react";
 import { cleanup, render } from "@testing-library/react";
 
@@ -28,7 +35,11 @@ function createFetch() {
   const fetchMock = vi.fn((_url: string, _init?: RequestInit) => {
     return new Promise((resolve) => {
       pending.push(({ ok, body }) =>
-        resolve({ ok, status: ok ? 200 : 500, json: async () => body } as Response),
+        resolve({
+          ok,
+          status: ok ? 200 : 500,
+          json: async () => body,
+        } as Response),
       );
     });
   });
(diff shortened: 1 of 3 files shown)


packages/gemi/client/useQuery.ts
```diff
@@ -457,7 +485,7 @@ export function useQuery<T extends keyof GetRPC>(
 
   return {
     data: state?.data as NestedPrettify<Data<T>>,
-    loading: state?.loading ?? !lazy,
+    loading: isDeferred || (state?.loading ?? !lazy),
```
nstfkc, line 495: **Docs are now stale.** `docs/data-fetching.md` (not in this diff) says of `loading`: "Only meaningful with `suspense: false` — a suspense query doesn't render until data exists" (line ~140), and describes `keepPreviousData` only as "keep old data visible while refetching" (line ~157). Both are contradicted by this change: under suspense `loading` is now the pending signal for a variant change, and that is exactly the flag pager UI is expected to consume. Issue #291's Notes explicitly include the docs update ("the migration guidance ... collapses to 'keepPreviousData just works'").

Suggested fix: in this PR, update the `loading` row to describe the deferred-variant pending window under suspense, and extend the `keepPreviousData` row (or the Suspense section) with the new behavior plus the `keepPreviousData: false` escape hatch. Otherwise the docs actively steer users toward `{ suspense: false }` for pagination — the pattern this PR exists to eliminate.

packages/gemi/client/useQuery.suspense.test.tsx
```diff
@@ -265,6 +276,84 @@ describe("useQuery with suspense (the default)", () => {
...
+      mounts += 1;
+    }, []);
+    return (
+      <button type="button" onClick={() => setPage((p) => p + 1)}>
+        {`ids:${(data as any[]).map((t: any) => t.id).join(",")} loading:${loading}`}
+      </button>
+    );
+  }
+
+  test("a variant change keeps the previous rows and flags pending — no startTransition in app code", async () => {
```
nstfkc, line 302: **Missing coverage: a `params` change is a meaningfully different code path from a `search` change, and only the latter is tested.** The PR body claims "a `search` *or* `params` change" is handled, and the implementation defers `currentPath` for that reason — but a path change swaps the entire `QueryResource` (new store, new `subscribe` callback, new read promise), whereas this test stays on one resource and only moves the variant key. The resource swap is where a regression would hide: the committed tree stays subscribed to the *old* resource for the whole pending window, and the suspended background render is woken by the new resource's thrown promise rather than any subscription.

Suggested fix: add a sibling test that drives a `params` change (e.g. `/todos/:listId` with local state) and asserts the same invariants — previous rows stay committed, `loading: true`, no fallback, new fetch fired, commit on resolve, no remount.

packages/gemi/client/useQuery.suspense.test.tsx
```diff
@@ -265,6 +276,84 @@ describe("useQuery with suspense (the default)", () => {
   });
 });
 
+describe("keepPreviousData under suspense", () => {
```
nstfkc, line 279: **Unpinned edge case: the new variant's fetch failing during the pending window.** Trace: background render suspends on `read(newKey)`; the fetch fails; the store writes `{ error, hasData: false, loading: false }` and settles the promise; React retries the background render; the `keepPreviousData` substitution in `useQuery` is skipped (it requires `snapshot?.loading`), so the suspense block throws `state.error` — and the error boundary replaces the previously-good rows the user was looking at, after a `loading: true` window that suggested the old rows were safe.

That may well be the intended semantics (it matches pre-PR suspend-then-error on variant change, and the docs' "throws when there is no data to show"), but it sits in tension with this feature's promise of "keep showing what I have while the next page loads", and nothing pins it. A future refactor of the error path could silently change it either direction.

Suggested fix: add a test that resolves the page-2 request with `ok: false` and asserts whichever behavior is intended — either `caught:QueryError` replaces the rows (current behavior), or, if you decide the previous rows should survive a failed page fetch, that `error` is returned while the rows stay committed.

packages/gemi/client/useQuery.ts
```diff
@@ -132,10 +142,28 @@ export function useQuery<T extends keyof GetRPC>(
...
+  // commits when the data lands — no `startTransition` needed at the call
+  // site. First mount is unchanged: deferred and current keys are equal, so
+  // an uncached query suspends into the segment fallback as today. Both hooks
+  // run unconditionally (hook order); the values are only *used* when
+  // deferring applies.
+  const deferredPath = useDeferredValue(currentPath);
+  const deferredVariantKey = useDeferredValue(currentVariantKey);
+  const deferVariant = suspense && config.keepPreviousData;
+  const normalPath = deferVariant ? deferredPath : currentPath;
+  const variantKey = deferVariant ? deferredVariantKey : currentVariantKey;
```
nstfkc, line 169: Non-blocking question: during the pending window, the *committed* tree's `mutate`/`refetch`/`trigger`/`prefetch` (and the variant-revalidation effect) all close over the deferred — i.e. previous — `variantKey` and `resource`. So a `refetch()` fired while `loading: true` after a page change refetches the old page, and a `mutate()` writes into the old variant. That's defensible (those callbacks act on the data that's visible), and the window is short, but it's a behavior choice worth making deliberately — e.g. a pager with a refresh button can now silently target the wrong variant for a frame-to-seconds window on slow networks. If it's intentional, a one-line addition to the big deferral comment ("imperative callbacks target the deferred/visible variant during the window") would keep the next reader from treating it as an oversight.

nstfkc (commented):
The core mechanism is correct and well-reasoned: deferring both the resource path and the variant key keeps the urgent pass on cached data, the background render's `resource.read` dedupes across discarded attempts (verified against `QueryResource.read`), first mount is unchanged (deferred === current), navigation isn't double-deferred (`useDeferredValue` returns the new value immediately inside the router's transition), and the `keepPreviousData: false` escape hatch is pinned. The tests cover the issue's acceptance criteria for the search-variant case.

Three actionable items, none blocking the approach itself:

1. `docs/data-fetching.md` is now wrong about `loading` under suspense and silent about the new `keepPreviousData` semantics — the issue's Notes explicitly scope in the docs update.
2. The `params`-change path (`deferredPath`, which swaps the *resource* mid-deferral) is claimed by the PR body but untested.
3. The error-during-pending-window behavior (background render throws into the boundary, tearing down the previous rows) is subtle, arguably surprising given keepPreviousData's promise, and unpinned.

Plus one non-blocking question about imperative callbacks closing over the deferred variant key during the pending window.

nstfkc:
Review addressed in 3886ca7 — all four points, taking each in turn:

**1. Stale docs (`docs/data-fetching.md`)** — fixed in 3886ca7. The `loading` row no longer says the flag is only meaningful with `suspense: false`; it now describes the variant-change pending window as the flag pager UI consumes. The `error` row clarifies that "no data to show" is judged against the *requested* variant — a failed variant change throws even with the previous rows on screen. Added a "Variant changes don't re-suspend" bullet to the suspense section (deferral behavior, `loading: true`, the `keepPreviousData: false` escape hatch, error-path semantics), a `keepPreviousData` paragraph under the config block, and rewrote the opt-out guidance that was steering pagination toward `{ suspense: false }` — it now points paging/search/filters at the default. (`docs/llms-full.txt` left alone: it's a snapshot that predates the suspense work entirely and isn't maintained in lockstep by doc commits.)

**2. Untested `params` path (resource swap mid-deferral)** — fixed in 3886ca7. New test drives `/lists/:listId/todos` across a `listId` change and asserts the same invariants as the search test: previous rows stay committed, `loading: true`, no fallback, the new resource's fetch on the wire immediately, commit on resolve, no remount. Verified it exercises the swap: without the `deferredPath` half of the deferral, the urgent render would read the fresh resource, suspend the committed tree, and the fallback assertion fails.

**3. Unpinned error-during-pending-window behavior** — pinned in 3886ca7, keeping the current semantics: the failed fetch throws into the segment's error boundary, replacing the previous rows. Deliberate, for the reason the test comment now records: under suspense the caller writes no error branch, so returning `error` while keeping the old variant's rows would leave a failed page change with no visible signal at all — pager clicked, `loading` flips off, rows silently stay on the old page. The boundary is the suspense contract's error channel, exactly as it was for a suspended variant change pre-deferral, and resetting it retries cleanly. Documented in the docs bullet and `error` row as well.

**4. (Non-blocking) callbacks closing over the deferred variant** — intentional, now stated in the deferral comment in `useQuery.ts`: during the pending window `mutate`/`refetch`/`trigger`/`prefetch` and the revalidation effects target the deferred — i.e. visible — variant and resource, acting on the data the user is looking at; they re-bind when the deferred render commits.

Tests: all 16 in `useQuery.suspense.test.tsx` and all 125 client tests pass; `tsc --noEmit` and oxlint clean. The 7 failing router/app test files pre-exist on the branch tip (verified via stash) and are untouched by this PR.

PR #296, opened 2026-08-01
Provider-level queryConfig: app-wide useQuery defaults, one line instead of ~120 edits

### LINKED ISSUE #292: Provider-level query defaults: flipping suspense per app should be one line, not ~120 edits
Follow-up to #285. `useQuery`'s config surface is per-call only, so an existing app cannot land the 0.49 upgrade behind a behavior freeze — the folio migration's Phase 0 was ~120 mechanical `{ suspense: false }` edits (every site destructuring `loading`/`error`, every `keepPreviousData` site), each of which must later be found and removed again to actually adopt suspense. The edits carry no information; they encode a single app-wide decision at 120 places.

## Proposal

A `queryConfig` accepted at the provider level (wherever `QueryManagerProvider` is configured from — plausibly a prop threaded through `createRoot`, or a `gemi.config.ts` client entry):

```ts
queryConfig: {
  suspense: false,        // app-wide default; per-call config still wins
  staleTime: 30_000,
  keepPreviousData: true,
  retryIntervalOnError: 10_000,
}
```

Resolution order: per-call config → provider defaults → framework defaults. Only keys with app-wide meaning belong here (`suspense`, `staleTime`, `keepPreviousData`, retry/refresh intervals) — `lazy`, `fallbackData`, `refetchUntil` stay call-site-only.

## Why it matters beyond convenience

- **Migration**: upgrade day is `queryConfig: { suspense: false }` + commit; adoption is then deleting that line route-group by route-group (or keeping per-call `suspense: true` opt-*ins* during the transition). "Zero behavior change" stops requiring a codemod.
- **Typing**: the non-nullable-`data` overload currently keys off the per-call literal. With a provider default in play, the honest cheap contract is: the overloads reflect the *call site* only, and an app that sets `suspense: false` globally reads `data` as nullable via a module-augmentation switch (`interface GemiQueryDefaults { suspense: false }`) the same way RPC types flow today. Worth deciding in review rather than silently lying in either direction.

## Acceptance

- An app with zero per-call configs and `queryConfig: { suspense: false }` behaves byte-for-byte like 0.48 useQuery (the existing non-suspense tests re-run under the flag).
- Per-call config always overrides the provider default; framework internals (`useUser`'s `suspense: false`, `useSignIn`'s `lazy`) are unaffected by app config.

nstfkc:
Closes #292

> Stacked on #295 (branch `feat/291-suspense-keep-previous-data`).

## Approach

`useQuery` defaults can now be set once, app-wide, where the roots are created:

```ts
// server entry (ViewRouterServiceProvider)
root = createRoot(RootLayout, { queryConfig: { suspense: false } });
// client entry (app/client.tsx)
init(RootLayout, { queryConfig: { suspense: false } });
```

- **Resolution order**: per-call config → provider `queryConfig` → framework defaults. Implemented as a single spread merge: `useQuery` becomes a thin wrapper that reads a new `QueryConfigContext` (provided by `QueryManagerProvider` via a `queryConfig` prop, threaded from `createRoot`/`init` through `ClientRouter`) and delegates to `useFrameworkQuery`, which holds the previous implementation unchanged.
- **Only app-wide keys are accepted**: `suspense`, `staleTime`, `keepPreviousData`, `retryIntervalOnError`, `refreshInterval`. `lazy`, `fallbackData`, and `refetchUntil` stay call-site-only (the `QueryConfig` type simply doesn't have them).
- **Framework internals are insulated**: `useUser` and `useSignIn` now call `useFrameworkQuery` directly (not exported from the package root), so their pinned semantics (`suspense: false`, `lazy`) can never be changed by an app's `queryConfig` — including keys they don't set themselves, like `refreshInterval`.
- **Typing**: the overloads keep reflecting the call site only. An app that sets `queryConfig: { suspense: false }` flips the default `data` nullability with the same module-augmentation switch RPC types use — `declare module "gemi/client" { interface GemiQueryDefaults { suspense: false } }` — and a new explicit-literal overload keeps a per-call `suspense: true` non-nullable either way. With no augmentation, all existing call sites type exactly as before.

## Tests

`packages/gemi/client/useQuery.queryConfig.test.tsx` (8 tests):
- The acceptance freeze: `queryConfig: { suspense: false }` with **zero per-call configs** re-runs the non-suspense assertions from `useQuery.suspense.test.tsx` — loading-flag sequence, errors returned not thrown, and the 0.48 `keepPreviousData` path on variant changes.
- Per-call config wins in both directions (`suspense: true` over app-wide `false` and vice versa; `keepPreviousData: true` over app-wide `false`).
- A numeric provider key flows through (`retryIntervalOnError` drives the retry timer).
- `useUser` under an aggressive `queryConfig` (`refreshInterval: 5, staleTime: 0`): still resolves `user: null` semantics without suspending and never refetches — one fetch total.

`bun run vitest run client/` → 14 files, 131 tests pass. `tsc --noEmit` on both `tsconfig.json` and `tsconfig.browser.json` clean; `test:types` clean; oxlint clean. (7 pre-existing failures in `app/`/`services/`/`utils/` reproduce identically on the base branch without this change.)

Note on type tests: the package's own tsconfig has `strict: false`, under which `| undefined` unions collapse and nullability assertions are vacuous — the `GemiQueryDefaults` contract was verified manually under `--strict`, which is what consuming apps use.

### DIFF
--- a/packages/gemi/client/QueryManagerContext.tsx
+++ b/packages/gemi/client/QueryManagerContext.tsx
@@ -10,6 +10,31 @@ import { QueryResource } from "./QueryResource";
 /** One streamed query payload: `[path, variantKey, data]`. */
 type StreamedQueryPayload = [string, string, any];
 
+/**
+ * App-wide `useQuery` defaults, threaded from `createRoot(RootLayout, {
+ * queryConfig })` (and `init` on the client). Resolution order is per-call
+ * config → these defaults → framework defaults, so a call site always wins.
+ * Only keys with app-wide meaning are accepted — `lazy`, `fallbackData` and
+ * `refetchUntil` stay call-site-only. Framework-internal hooks (`useUser`,
+ * `useSignIn`) never see these.
+ */
+export interface QueryConfig {
+  /**
+   * App-wide suspense switch. `suspense: false` makes every unconfigured
+   * `useQuery` behave like pre-0.49: `loading` flags instead of suspension,
+   * errors returned instead of thrown. Pair it with the `GemiQueryDefaults`
+   * module augmentation so `data`'s nullability matches.
+   */
+  suspense?: boolean;
+  /** How long cached data stays fresh before a read revalidates it, in ms. */
+  staleTime?: number;
+  keepPreviousData?: boolean;
+  retryIntervalOnError?: number;
+  refreshInterval?: number;
+}
+
+export const QueryConfigContext = createContext<QueryConfig | null>(null);
+
 export type PrefetchedData = Record<string, Record<string, any>>;
 
 export interface QueryManagerContextValue {
@@ -29,7 +54,10 @@ export const QueryManagerContext = createContext<QueryManagerContextValue>({
   clearErrors: () => {},
 });
 
-export const QueryManagerProvider = ({ children }: PropsWithChildren<{}>) => {
+export const QueryManagerProvider = ({
+  children,
+  queryConfig = null,
+}: PropsWithChildren<{ queryConfig?: QueryConfig | null }>) => {
   const resourcesRef = useRef<Map<string, QueryResource>>(new Map());
 
   const getResource = useCallback(
@@ -107,7 +135,9 @@ export const QueryManagerProvider = ({ children }: PropsWithChildren<{}>) => {
 
   return (
     <QueryManagerContext.Provider value={value}>
-      {children}
+      <QueryConfigContext.Provider value={queryConfig}>
+        {children}
+      </QueryConfigContext.Provider>
     </QueryManagerContext.Provider>
   );
 };
--- a/packages/gemi/client/QueryManagerContext.tsx
+++ b/packages/gemi/client/QueryManagerContext.tsx
@@ -35,6 +35,36 @@ export interface QueryConfig {
 
 export const QueryConfigContext = createContext<QueryConfig | null>(null);
 
+/**
+ * The runtime mirror of the `QueryConfig` type: TS excess-property checking
+ * only fires on fresh object literals, so a dynamically built (or plain-JS)
+ * config could smuggle call-site-only keys (`lazy`, `fallbackData`,
+ * `refetchUntil`) into every query. The provider picks these keys — and only
+ * these — before the value enters the context. Keys whose value is
+ * `undefined` are dropped too, so an absent value (e.g. an unset env-derived
+ * `staleTime`) falls through to the framework default instead of shadowing it.
+ */
+const APP_WIDE_QUERY_CONFIG_KEYS = [
+  "suspense",
+  "staleTime",
+  "keepPreviousData",
+  "retryIntervalOnError",
+  "refreshInterval",
+] as const satisfies ReadonlyArray<keyof QueryConfig>;
+
+export function pickAppWideQueryConfig(
+  queryConfig: QueryConfig | null | undefined,
+): QueryConfig | null {
+  if (!queryConfig) return null;
+  const picked: QueryConfig = {};
+  for (const key of APP_WIDE_QUERY_CONFIG_KEYS) {
+    if (queryConfig[key] !== undefined) {
+      (picked as Record<string, unknown>)[key] = queryConfig[key];
+    }
+  }
+  return picked;
+}
+
 export type PrefetchedData = Record<string, Record<string, any>>;
 
 export interface QueryManagerContextValue {
@@ -59,6 +89,13 @@ export const QueryManagerProvider = ({
   queryConfig = null,
 }: PropsWithChildren<{ queryConfig?: QueryConfig | null }>) => {
   const resourcesRef = useRef<Map<string, QueryResource>>(new Map());
+
+  // Sanitized once at the choke point every query reads through, so the
+  // "only app-wide keys" contract holds at runtime, not just in the types.
+  const appQueryConfig = useMemo(
+    () => pickAppWideQueryConfig(queryConfig),
+    [queryConfig],
+  );
 
   const getResource = useCallback(
     (key: string, initialState?: Record<string, any>) => {
@@ -135,7 +172,7 @@ export const QueryManagerProvider = ({
 
   return (
     <QueryManagerContext.Provider value={value}>
-      <QueryConfigContext.Provider value={queryConfig}>
+      <QueryConfigContext.Provider value={appQueryConfig}>
         {children}
       </QueryConfigContext.Provider>
     </QueryManagerContext.Provider>
--- a/packages/gemi/client/init.tsx
+++ b/packages/gemi/client/init.tsx
@@ -2,7 +2,17 @@ import { useEffect, type ComponentType } from "react";
 import { hydrateRoot, createRoot } from "react-dom/client";
 import { ServerDataProvider } from "./ServerDataProvider";
 import { ClientRouter } from "./ClientRouter";
+import type { QueryConfig } from "./QueryManagerContext";
 import { ErrorBoundary } from "react-error-boundary";
+
+export interface InitOptions {
+  /**
+   * App-wide `useQuery` defaults (per-call config always wins). Mirror the
+   * value passed to `createRoot` in the view router's service provider so the
+   * server render and hydration agree.
+   */
+  queryConfig?: QueryConfig;
+}
 
 const StackTrace = () => {
   useEffect(() => {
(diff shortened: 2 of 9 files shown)


packages/gemi/client/useQuery.ts
```diff
@@ -132,6 +171,24 @@ export function useQuery<T extends keyof GetRPC>(
   url: T,
   ...args: [options?: Options<T>, config?: Config<Data<T>>]
 ) {
+  const [options, config] = args;
+  // Resolution order: per-call config → provider `queryConfig` → framework
+  // defaults. The first two merge here; `useFrameworkQuery` applies the last.
+  const queryConfig = useContext(QueryConfigContext);
+  return useFrameworkQuery(url, options, { ...queryConfig, ...config });
```
nstfkc: **Bug: a per-call key explicitly set to `undefined` clobbers the provider default instead of falling through to it.**

`{ ...queryConfig, ...config }` copies `undefined`-valued keys from `config`, so the documented resolution order (per-call → provider → framework) is violated for them. This is a very common call-site shape — conditional config or forwarding an optional prop:

```tsx
useQuery("/todos", {}, { suspense: cond ? true : undefined });
useQuery("/todos", {}, { staleTime: props.staleTime }); // props.staleTime?: number
```

Verified repro on this branch: with `queryConfig={{ suspense: false }}`, a call site passing `{ suspense: undefined }` **suspends** (falls through to the framework default `suspense: true`) instead of returning `loading` flags — so the migration freeze the issue is about silently doesn't apply to those call sites. Before this PR the same shape was harmless because the only thing an `undefined` could clobber was the framework default it would have resolved to anyway; the provider layer makes it observable.

Suggested fix — drop `undefined` entries from the per-call config before merging:

```ts
const defined = Object.fromEntries(
  Object.entries(config ?? {}).filter(([, v]) => v !== undefined),
);
return useFrameworkQuery(url, options, { ...queryConfig, ...defined });
```

(Applying the same filter to `queryConfig` costs nothing and also protects the provider→framework step, e.g. `queryConfig: { staleTime: env.STALE_TIME }` where the env value is absent.) Worth a test with an `undefined`-valued per-call key next to the existing "per-call config wins" cases.

packages/gemi/client/QueryManagerContext.tsx
```diff
@@ -29,7 +54,10 @@ export const QueryManagerContext = createContext<QueryManagerContextValue>({
   clearErrors: () => {},
 });
 
-export const QueryManagerProvider = ({ children }: PropsWithChildren<{}>) => {
+export const QueryManagerProvider = ({
+  children,
+  queryConfig = null,
+}: PropsWithChildren<{ queryConfig?: QueryConfig | null }>) => {
```
nstfkc, line 90: **The "only app-wide keys are accepted" contract is enforced only at the type level, and only for object literals.**

TS excess-property checking doesn't fire when the value isn't a fresh literal, so any app that builds its config dynamically (or is plain JS) can pass extra keys and they flow straight through `{ ...queryConfig, ...config }` into **every** `useQuery` in the app:

```ts
const cfg = { suspense: false, lazy: true, fallbackData: [] };
init(RootLayout, { queryConfig: cfg }); // typechecks; lazy now applies app-wide
```

`lazy: true` app-wide would stop every unconfigured query from ever fetching — exactly the class of footgun the type carve-out was meant to prevent. Since the provider is the single choke point, a one-time runtime pick keeps the contract honest:

```ts
const ALLOWED = ["suspense", "staleTime", "keepPreviousData", "retryIntervalOnError", "refreshInterval"] as const;
const picked = queryConfig && Object.fromEntries(
  ALLOWED.filter((k) => k in queryConfig).map((k) => [k, queryConfig[k]]),
);
```

(Or do the pick once in `useQuery` next to the merge — either way it's a few lines and makes the JSDoc claim true at runtime.)

packages/gemi/client/init.tsx
```diff
@@ -2,8 +2,18 @@ import { useEffect, type ComponentType } from "react";
...
+import type { QueryConfig } from "./QueryManagerContext";
 import { ErrorBoundary } from "react-error-boundary";
 
+export interface InitOptions {
+  /**
+   * App-wide `useQuery` defaults (per-call config always wins). Mirror the
+   * value passed to `createRoot` in the view router's service provider so the
+   * server render and hydration agree.
+   */
+  queryConfig?: QueryConfig;
```
nstfkc, line 14: **Non-blocking design note: the app-wide default now lives in two places that can silently disagree.**

The JSDoc here and in `CreateRootOptions` says to mirror the value, but nothing checks it. If the server entry says `suspense: false` and the client entry forgets (or vice versa), the SSR HTML renders one behavior (loading-flag markup) and hydration renders the other (suspension), producing a hydration mismatch that will look like an unrelated bug and won't point anyone at `queryConfig`.

Since `create()` already stamps the server's route payload into `window.__GEMI_DATA__`, the server could include the `queryConfig` it rendered with, and `init` could `console.warn` in dev when the client-passed value differs. Alternatively, documenting the pattern of a shared `queryConfig.ts` module imported by both entries gets most of the way there with zero code. Fine to defer — but worth deciding before the API is public, since the issue itself floated a single-source `gemi.config.ts` shape.

nstfkc (commented):
Reviewed against base `feat/291-suspense-keep-previous-data` and issue #292. The shape is right: a thin `useQuery` wrapper merging `QueryConfigContext` over `useFrameworkQuery`, framework internals (`useUser`, `useSignIn`) routed around the app config, and the acceptance freeze (`queryConfig: { suspense: false }` with zero per-call configs reproducing the 0.48 loading/error/keepPreviousData behavior) is genuinely tested — I ran the new suite and the full `client/` suite on the head commit (131 passing).

Two findings that warrant code changes:

1. **Explicitly-`undefined` per-call keys clobber provider defaults** (`useQuery.ts`) — the documented resolution order breaks for a very common call-site shape; repro included inline.
2. **The "only app-wide keys" contract is type-only** (`QueryManagerContext.tsx`) — a non-literal `queryConfig` can smuggle `lazy`/`fallbackData`/`refetchUntil` into every query at runtime.

Plus one non-blocking design note on the dual `createRoot`/`init` configuration point.

The overload story (call-site-only typing + `GemiQueryDefaults` augmentation) matches what the issue asked to be decided explicitly, and the `ExplicitSuspenseConfig` overload keeps per-call `suspense: true` non-nullable under the augmentation — that part looks right.

nstfkc:
Addressed the review in e740f63 — all three points below. `client/` suite: 134 passing (131 + 3 new); `tsc --noEmit` clean on both `tsconfig.json` and `tsconfig.browser.json`; `test:types` clean; oxlint clean.

**1. Explicitly-`undefined` per-call keys clobber provider defaults (`useQuery.ts`) — fixed in e740f63.**
`useQuery` no longer spreads the raw per-call config over the provider value; it merges only keys whose value is defined, so `{ suspense: cond ? true : undefined }` and forwarded optional props fall through to the provider default as documented. Regression tests cover both directions: a per-call `suspense: undefined` under `queryConfig: { suspense: false }` now returns loading flags (previously suspended), and a provider-level `suspense: undefined` falls through to the framework default. The suggested symmetric filter on the provider value landed too, but at the provider (see next point) so it runs once instead of per hook call.

**2. "Only app-wide keys" contract is type-only (`QueryManagerContext.tsx`) — fixed in e740f63.**
`QueryManagerProvider` now runs `pickAppWideQueryConfig` (memoized) before the value enters `QueryConfigContext`: only `suspense`, `staleTime`, `keepPreviousData`, `retryIntervalOnError`, `refreshInterval` pass, and `undefined`-valued entries are dropped (covering the `staleTime: env.STALE_TIME`-absent case from point 1). The allowlist is `satisfies ReadonlyArray<keyof QueryConfig>`, so it can't drift from the type silently. New test smuggles `lazy: true` + `fallbackData` through a non-literal config and asserts the query still fetches on mount, isn't seeded, and that the legitimate `suspense: false` key still applies.

**3. Dual `createRoot`/`init` config point can silently disagree (`init.tsx`) — deferred, as the comment allows.**
Not changed in this PR: stamping the server's `queryConfig` into `__GEMI_DATA__` touches the server document-render path, which this PR otherwise doesn't, and the right shape depends on whether a single-source `gemi.config.ts` (floated in #292) lands — a dev-mode warning built on `__GEMI_DATA__` would be throwaway if it does. The runtime pick from point 2 slightly narrows the blast radius (both sides are sanitized identically), and the JSDoc on both `CreateRootOptions.queryConfig` and `InitOptions.queryConfig` already instructs mirroring the value. Agreed it should be decided before the API is public — worth a follow-up issue tying it to the config-file decision.

Issue #293, opened 2026-08-01
Stream lifecycle hooks: spans end at time-to-shell and streamed query failures bypass onRequestFail


nstfkc:
Follow-up to #285. Streaming broke two observability assumptions that every instrumented app will hit; both surfaced immediately in a real app's migration:

1. **Response timing is now wrong by default.** The app wraps requests in a manual Sentry `http.server` span with `span.end()` in `finally` — which fires when the handler *returns*, i.e. at time-to-shell, while the body streams for another second. Durations collapse from ~1.2 s to ~6 ms without the page getting faster, and `setHttpStatus` runs against a response whose body is still open. Every APM integration that wraps `app.fetch` has this bug now.

2. **Streamed query failures are invisible to error reporting.** A query that rejects during the streamed render is logged via the React stream's `onError` (dev-only console today) and the segment client-renders — `onRequestFail` in the router providers never fires, so Sentry/error tracking sees nothing from the server side even though the server did the failing work.

## Proposal

Give the view router an explicit stream lifecycle instead of making apps hand-wrap response bodies:

```ts
// ViewRouterServiceProvider (naming flexible)
onStreamComplete(req: HttpRequest, summary: {
  shellAt: number;          // ms from request start to first byte
  settledAt: number;        // ms to last chunk (allReady or deadline)
  aborted: boolean;         // deadline hit
  queries: Array<{ path: string; variantKey: string; startedAt: number; settledAt: number;
                   status: "resolved" | "rejected"; source: "prefetch" | "render" }>;
}): void
```

- Fire it when the response body actually closes (hook the injector's `flush`/`cancel` — it already sits at the end of the pipe), not when the handler returns. `ServerQueryStore` already records per-entry timings; the summary is a serialization of what exists.
- Route streamed query rejections through the existing `onRequestFail(req, error)` (or a dedicated `onQueryFail`) so error tracking wired today keeps working under streaming. The error is a `QueryError` with path/variant/status attached.
- Docs: an "instrumenting streamed responses" section showing the Sentry pattern — start the span in instrumentation, end it in `onStreamComplete`.

## Acceptance

- A test asserting `onStreamComplete` fires after the last injected chunk with `settledAt` ≈ the slowest query, and `shellAt` ≪ `settledAt` for a suspending page.
- A rejected streamed query invokes the fail hook exactly once with the `QueryError`.
- Non-streamed paths (`.json`, `no-stream`, bots) report `shellAt === settledAt`.

[...]

PR #297, opened 2026-08-01
Stream lifecycle hooks: onStreamComplete on body close, streamed query failures through onRequestFail

nstfkc:
-

### DIFF
--- a/docs/project-structure.md
+++ b/docs/project-structure.md
@@ -205,6 +205,7 @@ The router providers expose request lifecycle hooks you can override for cross-c
 - **`onRequestStart(req)`** — before a matched request is handled (API path).
 - **`onRequestEnd(req)`** — after a request is handled successfully.
 - **`onRequestFail(req, error)`** — when handling throws.
+- **`onStreamComplete(req, summary)`** — view provider only; when a response body actually closes, after the last streamed chunk (see [Instrumenting streamed responses](#instrumenting-streamed-responses)).
 
 A real-world example from a production app reports failures to Sentry:
 
@@ -229,6 +230,40 @@ export default class extends ApiRouterServiceProvider {
 Other providers expose their own hooks — for example the logging provider has `onLogCreated(logEntry)` and `onLogFileClosed(file)`, the i18n provider has `onLocaleChange(locale)`, and the authentication provider has `onSignUp`, `onForgotPassword`, and `onMagicLinkCreated`.
 
 > **Gotcha:** On the view path, gemi does not call `onRequestStart`. If you need per-request setup for views, use `onRequestEnd` / `onRequestFail`, which both fire on the view path.
+
+### Instrumenting streamed responses
+
+View responses stream: the handler returns at time-to-shell while query payloads keep streaming for as long as the slowest query takes. A span that ends when the handler returns (or in a `finally` around `app.fetch`) therefore records ~milliseconds for a request whose body stayed open for a second — the duration of the shell, not the response.
+
+The view router provider's `onStreamComplete(req, summary)` fires when the response body actually closes — after the last streamed chunk, or when the stream deadline aborts rendering. The `summary` reports:
+
+- **`shellAt`** — ms from request start to the response's first byte.
+- **`settledAt`** — ms to the last chunk, when the body closed.
+- **`aborted`** — whether the stream deadline cut rendering short.
+- **`queries`** — per-query `path`, `variantKey`, `startedAt`, `settledAt`, `status`, and `source` (`"prefetch"` or `"render"`).
+
+Non-streamed responses (`.json` navigation payloads, `"no-stream"` routes, bot requests) report `shellAt === settledAt`.
+
+Start the span where you already do, and end it here instead:
+
+```typescript
+import { ViewRouterServiceProvider } from "gemi/services";
+import type { StreamSummary } from "gemi/services";
+import type { HttpRequest } from "gemi/http";
+
+export default class extends ViewRouterServiceProvider {
+  // ...
+
+  onStreamComplete(req: HttpRequest, summary: StreamSummary) {
+    // e.g. end the Sentry span opened for this request:
+    // span.setAttribute("gemi.shell_ms", summary.shellAt);
+    // span.setAttribute("gemi.aborted", summary.aborted);
+    // span.end(); // now measures the full streamed response
+  }
+}
+```
+
+Query failures during the streamed render are routed through `onRequestFail(req, error)` exactly once per rejected query, with a `QueryError` carrying the query's path, variant, and status — so error tracking wired to `onRequestFail` keeps seeing server-side failures that React quietly hands over to client rendering. (`Query.instant` rethrows into the handler; that rejection is still reported only once.)
 
 ## App bootstrap
 
--- a/docs/project-structure.md
+++ b/docs/project-structure.md
@@ -242,7 +242,9 @@ The view router provider's `onStreamComplete(req, summary)` fires when the respo
 - **`aborted`** — whether the stream deadline cut rendering short.
 - **`queries`** — per-query `path`, `variantKey`, `startedAt`, `settledAt`, `status`, and `source` (`"prefetch"` or `"render"`).
 
-Non-streamed responses (`.json` navigation payloads, `"no-stream"` routes, bot requests) report `shellAt === settledAt`.
+Non-streamed responses (`.json` navigation payloads, `"no-stream"` routes, bot requests) report `shellAt === settledAt`. The hook also fires when a shell render crashes (or the deadline aborts it before the shell resolved) — the fallback error document's body still closes, so the span still ends.
+
+> **Scope:** `onStreamComplete` fires for document and `.json` navigation bodies — the responses that stream. Responses without a query lifecycle (OG images, `FILE` routes, redirects and error responses produced by `RequestBreakerError`) do not fire it, so a span opened unconditionally in instrumentation needs a fallback end (e.g. also ending it in `onRequestEnd` / `onRequestFail` when no stream ever started).
 
 Start the span where you already do, and end it here instead:
 
@@ -263,7 +265,9 @@ export default class extends ViewRouterServiceProvider {
 }
 ```
 
-Query failures during the streamed render are routed through `onRequestFail(req, error)` exactly once per rejected query, with a `QueryError` carrying the query's path, variant, and status — so error tracking wired to `onRequestFail` keeps seeing server-side failures that React quietly hands over to client rendering. (`Query.instant` rethrows into the handler; that rejection is still reported only once.)
+Query failures during the streamed render are routed through `onRequestFail(req, error)` exactly once per rejected query, with a `QueryError` carrying the query's path, variant, and status — so error tracking wired to `onRequestFail` keeps seeing server-side failures that React quietly hands over to client rendering. (`Query.instant` rethrows into the handler; that rejection is still reported only once.) An api handler that fails with a `RequestBreakerError` (or an error `Response`) yields a `QueryError` with that response's status; a raw `throw` inside the handler yields a status-500 `QueryError` with the original error on `cause`.
+
+> **Gotcha:** a raw `throw` inside an api handler also fires the **api** provider's `onRequestFail` — the query runs the api handler in-process, and the api router reports its own failures. If both providers report to the same error tracker, dedupe there (the view provider's error is the `QueryError` wrapper, the api provider's is the original error on its `cause`).
 
 ## App bootstrap
 
--- a/packages/gemi/services/router/ServerQueryStore.test.ts
+++ b/packages/gemi/services/router/ServerQueryStore.test.ts
@@ -1,11 +1,16 @@
 import { afterEach, describe, expect, test, vi } from "vitest";
-import { ServerQueryStore, type ServerQueryFetcher } from "./ServerQueryStore";
+import {
+  ServerQueryStore,
+  type ServerQueryFetcher,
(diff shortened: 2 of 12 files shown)


packages/gemi/services/router/ViewRouterServiceContainer.ts
```diff
@@ -145,7 +145,26 @@ export class ViewRouterServiceContainer extends ServiceContainer {
...
   }
 
+  /**
+   * Fires the provider's `onStreamComplete` from a stream callback — an
+   * observability hook must never break the response body it observes, so
+   * sync throws and rejections are logged and swallowed.
+   */
+  private completeStream(req: HttpRequest, summary: StreamSummary) {
+    const logError = (err: any) => {
+      Log.error(err?.message ?? 'Error in "onStreamComplete" event handler', {
```
nstfkc, line 162: **Correctness: this hook (and its error guard) runs outside the kernel `AsyncLocalStorage` scope on the document path, so facades break inside it — and this `Log.error` fallback itself throws.**

`app.fetch` wraps request handling in `kernel.run(...)`, which ends when the `Response` is returned. The injector's `pull`/`onClose` is driven afterwards by the HTTP server consuming the body — outside that scope. This is the exact phenomenon `createServerQueryFetcher` documents ("the http server drives [the stream] *outside* the kernel scope — so capture the store now and re-enter it around every fetch") and works around with `kernelContext.run(kernelStore, ...)`.

Consequences:
- Any user `onStreamComplete` that touches a facade (`Log`, or anything resolved via `ServiceContainer.use()`) hits `use()` returning `undefined` → TypeError. For an observability hook whose documented purpose is instrumentation, that's a landmine.
- That TypeError is then handed to `logError`, whose `Log.error` hits the same `undefined.log` TypeError — so the guard built to "never break the response body" itself throws: an unhandled rejection from the `.catch(logError)` branch, or a throw propagating into the stream machinery from the sync branch.
- Same problem for the `onQueryFail` → `onRequestFail` routing below: for *render-discovered* queries the settle chain was registered during React's render (outside the kernel scope), so a facade-using `onRequestFail` silently no-ops for streamed failures (the empty `catch` swallows the TypeError) while working fine for handler-phase failures.

Suggested fix: capture `const kernelStore = kernelContext.getStore()` in `handleViewRequest` (where it is guaranteed live) and invoke both `completeStream` and the `onQueryFail` listener body inside `kernelContext.run(kernelStore, ...)` — mirroring `serverQueryFetcher`. Worth deciding (and documenting) the same for `RequestContext`: `req.ctx()` inside these hooks will also fail today.

Note the `.json` path mostly escapes this by accident — `createRoutePayloadStream`'s deadline timer and `allSettled().then(close)` continuations are registered inside `start()` under the kernel scope, so ALS restores it there. Only the document path (the headline feature) is affected, plus the `cancel()` paths of both streams. That asymmetry is exactly the kind of thing that survives manual testing.

packages/gemi/services/router/ViewRouterServiceContainer.ts
```diff
@@ -360,10 +380,24 @@ export class ViewRouterServiceContainer extends ServiceContainer {
...
+            onClose: () => {
+              this.completeStream(req, serverQueries.summarize(deadline.signal.aborted));
+            },
+          }),
+          {
+            status: !currentPathName ? 404 : 200,
+            headers,
+          },
+        );
       } catch (err) {
```
nstfkc, line 417: **Spec gap: the render-error fallback response never fires `onStreamComplete`, so the documented span pattern leaks spans precisely on failed requests.**

This catch (shell render crash, or deadline abort before the shell resolved) builds a fallback stream that bypasses `injectQueryPayloads` — no `onClose`, no `completeStream`. An app following the docs ("start the span in instrumentation, end it in `onStreamComplete`") gets a span that never ends for exactly the requests it most wants to see; issue #293 pins the hook to "when the response body actually closes", which this body also does.

Suggested fix: either wrap the fallback stream in `injectQueryPayloads(stream, serverQueries, { onClose: () => this.completeStream(req, serverQueries.summarize(deadline.signal.aborted)) })`, or call `completeStream` directly in this catch. Same consideration applies (less urgently) to the other view-path responses that close a body without the hook — OG, FILE, breaker-error responses — worth at least a docs note that `onStreamComplete` fires only for document/`.json` bodies, so span-ending code needs a fallback (e.g. also ending in `onRequestFail`).

docs/project-structure.md
```diff
@@ -230,6 +231,40 @@ Other providers expose their own hooks — for example the logging provider has
...
+  onStreamComplete(req: HttpRequest, summary: StreamSummary) {
+    // e.g. end the Sentry span opened for this request:
+    // span.setAttribute("gemi.shell_ms", summary.shellAt);
+    // span.setAttribute("gemi.aborted", summary.aborted);
+    // span.end(); // now measures the full streamed response
+  }
+}
+```
+
+Query failures during the streamed render are routed through `onRequestFail(req, error)` exactly once per rejected query, with a `QueryError` carrying the query's path, variant, and status — so error tracking wired to `onRequestFail` keeps seeing server-side failures that React quietly hands over to client rendering. (`Query.instant` rethrows into the handler; that rejection is still reported only once.)
```
nstfkc: **Docs overpromise: the error is only a `QueryError` when the api handler failed via a `RequestBreakerError`/error `Response`.**

`createServerQueryFetcher` wraps only the `data instanceof Response && !data.ok` case in `QueryError`. A plain `throw` inside the api handler propagates out of `getRouteData` unwrapped, so `onRequestFail` receives the raw error with no path/variant/status attached — despite this line (and issue #293's acceptance) promising a `QueryError`.

Two options:
- Wrap non-`QueryError` rejections at the fetcher boundary: `catch (err) { throw err instanceof QueryError ? err : new QueryError(path, variantKey, 500, { message: err?.message }); }` (keep the original on `cause`). This also keeps the `reportedQueryErrors` identity dedupe working, since `Query.instant` rethrows `entry.error` which would then be the wrapped instance.
- Or soften this sentence to say the error is a `QueryError` for HTTP-shaped failures and the raw error otherwise.

Related behavior worth a docs note either way: for raw throws, `getRouteData` already calls the **Api** provider's `onRequestFail` before rethrowing, so with this PR the same underlying failure now reaches both providers' fail hooks — an app that wires Sentry into both (the docs present the hooks as one family) will double-report every raw-throw streamed query.

packages/gemi/services/router/ServerQueryStore.test.ts
```diff
@@ -249,6 +254,55 @@ describe("ServerQueryStore", () => {
     await muted.promise;
     expect(warn).not.toHaveBeenCalled();
   });
+
+  test("a rejected query fires the fail listener exactly once, with the error", async () => {
```
nstfkc, line 258: **Test gap: nothing exercises the container wiring the acceptance criteria actually name.**

Issue #293's acceptance is phrased at the provider level — "`onStreamComplete` fires after the last injected chunk", "a rejected streamed query invokes the fail hook exactly once" — but every new test stops at the store/stream primitives (`onQueryFail`, `onClose`, `summarize`). The genuinely tricky new logic lives in `ViewRouterServiceContainer` and is untested:

- `completeStream` actually invoking `service.onStreamComplete` (and swallowing a throwing hook without corrupting the body),
- `onQueryFail` → `service.onRequestFail` routing,
- the `reportedQueryErrors` identity dedupe for `Query.instant` rethrows ("reported only once" is asserted in the docs but nowhere in tests).

`packages/gemi/app/App.test.ts` already drives full requests through `app.fetch`, so an integration test is feasible: a view route with a rejecting prefetch + a spy provider, asserting `onRequestFail` fired exactly once and `onStreamComplete` fired after the body was fully read. That test would also have surfaced the kernel-context issue flagged in `completeStream`, since reading the body after `app.fetch` resolves happens outside `kernel.run` — the same shape as production.

packages/gemi/services/router/streamQueryInjection.ts
```diff
@@ -74,18 +96,21 @@ export function injectQueryPayloads(
         // last chunk — they still belong in the client cache.
         flush(controller);
         controller.close();
+        closeOnce();
```
nstfkc, line 110: **Robustness: a rejecting `reader.read()` skips `closeOnce`, breaking the "fires exactly once" contract — it fires zero times.**

If the React source stream errors mid-flight, `pull` rejects, the outbound stream errors, and neither `done` nor `cancel` runs — so `onClose` (and therefore `onStreamComplete`) never fires for that request, leaking the span the same way the render-catch path does. Rare (post-shell fatal errors), but the fix is a one-liner:

```ts
async pull(controller) {
  let result: ReadableStreamReadResult<Uint8Array>;
  try {
    result = await reader.read();
  } catch (err) {
    closeOnce();
    throw err;
  }
  // ...
}
```

That makes the close guarantee hold on all three exits (done, cancel, source error), matching the JSDoc on `StreamLifecycleHooks.onClose`.

nstfkc (commented):
Reviewed against issue #293 (diff vs `feat/292-provider-query-defaults` only). The core design is sound: the injector is the right place to observe first-byte/close, `markShell`/`summarize` on the store keeps the summary a serialization of existing state, and the `reportedQueryErrors` identity-set dedupe for `Query.instant` rethrows is correct (the fail listener fires before `settle()` resolves the entry promise, so the set is always populated before the catch sees the rethrow). Unit tests for the store and both stream primitives are good.

Five actionable points, most-severe first:

1. `onStreamComplete` (document path) and `onQueryFail`-routed `onRequestFail` fire outside the kernel AsyncLocalStorage scope — facades break inside the hooks, and `completeStream`'s own `Log.error` fallback throws a TypeError there (details inline).
2. The render-error fallback path returns a body with no lifecycle hooks, so `onStreamComplete` never fires for those requests — the documented span pattern leaks spans exactly when something went wrong.
3. The docs' "a `QueryError` carrying path/variant/status" claim only holds for breaker/error-Response failures; raw api-handler throws arrive unwrapped and are additionally double-reported via the Api provider's `onRequestFail`.
4. No test exercises the container wiring — the acceptance criteria name provider-level behavior (`onStreamComplete` fires, fail hook exactly once) but tests stop at the store/stream primitives; the trickiest logic (instant-rethrow dedupe, kernel-context question above) is untested.
5. A source-stream error in the injector's `pull` skips `closeOnce`, breaking the exactly-once close guarantee.

None of these require a design change — the fixes are local.

nstfkc:
All five review points addressed:

**1. Hooks outside the kernel AsyncLocalStorage scope — fixed in d837a3a.** `handleViewRequest` now captures the kernel store *and* the request context while they are live and re-enters both (`runInRequestScope`, via a new `RequestContext.runWith`) around every hook invocation: the document path's `onClose`, the `.json` path's `onClose` (covering its `cancel()` path too), and the `onQueryFail` → `onRequestFail` routing — with the rejection handler attached inside the scope so facade-using hooks work for render-phase failures exactly as for handler-phase ones. This mirrors `createServerQueryFetcher`, as suggested. The `RequestContext` question is decided in favor of re-entering: `req.ctx()` works inside the hooks, and the provider JSDoc now says so. The integration test asserts both scopes are live inside the hooks (verified it fails against the unfixed code).

**2. Render-error fallback never fired `onStreamComplete` — fixed in d837a3a.** The fallback stream now goes through `injectQueryPayloads` with the same `onClose`, so the hook fires when that body closes (no shell mark — it's one settled error document). Covered by an integration test that crashes the shell render. The OG/FILE/breaker gap got the docs note you suggested (in 131e886): the hook fires for document and `.json` bodies; spans opened unconditionally need a fallback end in `onRequestEnd`/`onRequestFail`.

**3. Raw throws arrived unwrapped — fixed in e1bbd85.** Non-`QueryError` rejections are wrapped at the fetcher boundary as a status-500 `QueryError` with the original error on `cause` (your first option), which also keeps the `reportedQueryErrors` identity dedupe working for `Query.instant` rethrows. The api-provider double-report is documented rather than changed (131e886): the api router reporting its own handler failures is defensible in isolation, and suppressing it for in-process query subrequests would be a behavior change beyond this PR — the docs now tell double-wired trackers to dedupe via `cause`.

**4. No container-level test — fixed in 4521b3e.** New `streamLifecycle.test.ts` drives full requests through `app.fetch` and consumes the body only after the handler returned — outside `kernel.run`, the exact production shape you predicted would surface point 1 (it does: 2 of 5 tests fail without the d837a3a fix). It covers: `onStreamComplete` exactly once after body close with `shellAt < settledAt` for a suspending page, `.json` reporting `shellAt === settledAt`, a rejected streamed prefetch reaching `onRequestFail` exactly once as a wrapped `QueryError`, the `Query.instant` rethrow dedupe, and the crashed-shell fallback. (Not added to `App.test.ts` — that suite currently fails at kernel boot because its `App({ root, apiRouter, viewRouter, ... })` params predate the current `App` constructor, a pre-existing issue this PR doesn't touch; the new suite builds its own kernel with real providers instead.) One find along the way: a Suspense boundary at the very root makes React defer the shell flush until the boundary settles, so the test root wraps it in a host element like any real layout — noted in the test.

**5. Rejecting `reader.read()` skipped `closeOnce` — fixed in bdb9fe1.** Your one-liner, plus a unit test erroring the source mid-stream and asserting `onClose` fires exactly once. The close guarantee now holds on all three exits (done, cancel, source error).

Validation: `vitest` for `services/router/` passes (72 tests, the one failure is the pre-existing `createRouteManifest > DeeplyNestedRouter` snapshot drift, also failing on the base branch); full-suite failure set is identical to the base branch's 7 pre-existing failures; `tsc --noEmit` clean; oxlint warning count unchanged (89).

Issue #289, opened 2026-08-01
Pages over progressiveChunkSize render blank without JS: the rc.2 chunking opt-out is reachable only by bot UAs


nstfkc:
#286 fixed the crawler path by disabling progressive chunking for bot user-agents. Human visitors with JS disabled still get a blank page, because the same mechanism applies to them and nothing lets a route opt out.

Any view whose HTML exceeds React's `progressiveChunkSize` (~12.8 kB) is emitted as `<div hidden id="S:0">…</div><script>$RC("B:0","S:0")</script>`. The content is in the response, but moving it into place requires executing the reveal script. No JS, no page.

## Measured

`0.49.0-rc.3`, production build, ordinary Chrome UA, JS disabled (text extracted from the shell only, i.e. what the parser renders before any script runs):

| route | renders without JS | only reachable via JS |
|---|---|---|
| `/` (landing) | **0 chars** | 3,666 |
| `/terms-and-conditions` | **0 chars** | 7,417 |
| `/about-us` | 4,150 | 0 |
| `/pricing` | 890 | 0 |

The split is purely size. `/about-us` and `/pricing` work only because they happen to sit under the threshold — which makes this a silent trap: adding a section to a page that currently works will blank it, with nothing in CI or typecheck to catch the transition.

## It is not about suspending

Worth stating explicitly, because it rules out the obvious workarounds. A boundary is deferred on size alone, even when nothing inside it ever suspends — no promise is thrown anywhere in this repro:

```jsx
const app = <html><body><h1>shell</h1><Suspense fallback={null}>{big(rows)}</Suspense></body></html>;
const s = await renderToReadableStream(app);
await s.allReady;
```

```
no-suspension rows=50    bytes=4025     deferred($RC)=false
no-suspension rows=400   bytes=32978    deferred($RC)=true
no-suspension rows=2000  bytes=161978   deferred($RC)=true
```

Since gemi wraps every route segment in `<Suspense>`, this reaches every sufficiently large page. An app cannot fix it by removing queries, avoiding suspense, or restructuring the view — the page is simply big. The only app-level lever is keeping every page under ~12.8 kB of HTML, which is not a real constraint to design against.

## The switch already exists

The rc.2 fix is a one-line conditional, gated solely on the UA regex:

```js
...T ? { progressiveChunkSize: Number.MAX_SAFE_INTEGER } : {}   // T === isBotUserAgent(...)
```

Nothing else can reach it. That is the whole request: let a route ask for the same treatment.

[...]

PR #298, opened 2026-08-01
Dev-mode hint when a streamed route ships near-zero shell-readable text

### LINKED ISSUE #294: Guard against the silent no-JS blanking threshold: report shell-readable content per route in dev/build
Follow-up to #289. The sharpest part of that report was not the bug but the trap: React defers any boundary over `progressiveChunkSize` (~12.8 kB) on size alone, so a public page that renders fine without JS today goes blank the day one more section pushes it over the budget — and nothing in typecheck, tests, or CI notices the transition. `"no-stream"` fixes it, but only for routes someone already knows are affected. Discovery currently happens via bug report (in folio's case: `/` and `/terms-and-conditions` at 0 readable chars, found by hand-measuring).

## Proposal

Make the framework report what a no-JS reader gets, at the moment it changes:

- **Dev-mode**: after streaming a document response, gemi already has the full byte stream in hand (the injector sits at the end of the pipe). Measure shell-readable text — visible text outside `<div hidden>` reveal segments and `<template>` fallback markers — versus deferred text. When a route ships ≥ some floor of deferred text (say 1 kB) while its shell readable text is near zero, log once per route:
  `[gemi] GET / streams 3.7 kB of its content behind reveal scripts — a no-JS visitor sees ~0 chars. If this is a public/content route, add "no-stream" to its middleware.`
  Suppress for routes already marked `no-stream` and for authenticated-app routes is not distinguishable automatically — once per route per boot keeps it low-noise.
- **Build-time (stretch)**: a `gemi build --report-shell` (or part of the default build summary) that renders each public GET view route shell-only and prints the table from #289 (`route / shell chars / deferred chars`), giving CI a diffable artifact. Needs handler execution, so realistically it renders routes whose handlers succeed without auth; skipping the rest with a note is fine — the routes that matter for no-JS audiences are exactly the public ones.

## Why dev-first

The measurement is nearly free where the bytes already flow, and the failure mode is a *transition* (worked yesterday, blank today) — a persistent dev-console line at the moment the threshold is crossed is worth more than a CI step someone has to adopt.

## Acceptance

- A route over the deferral budget without `"no-stream"` logs the hint once; adding `"no-stream"` (or shrinking the page) clears it.
- The measurement counts the #289 repro correctly: `/` at 0 shell chars / 3,666 deferred flags; `/about-us` at 4,150 / 0 stays silent.

nstfkc:
Closes #294

Stacked on #297 (branch `feat/293-stream-lifecycle-hooks`) — review only the top commit here until that merges.

## Approach

React defers any boundary over `progressiveChunkSize` (~12.8 kB) on size alone, so a page that reads fine without JS goes blank the day one more section pushes it over the budget — and nothing in typecheck, tests, or CI notices the transition (#289 was found by hand-measuring `/` at 0 readable chars). This adds the dev-mode guard proposed in the issue:

- **`onChunk` lifecycle hook** on the stream injector (`streamQueryInjection.ts`) — the injector is the last thing to see the response body, so an observer hung there holds the document exactly as the visitor received it, injected payload scripts included.
- **`shellContentReport.ts`** — a small HTML scanner that splits the streamed document into shell-readable text (visible text outside `<div hidden>` reveal segments, scripts, styles, templates, and titles; comments and attributes never count) versus deferred text parked behind `$RC()` reveal scripts, plus the hint formatter and a streaming-safe byte accumulator.
- **`ViewRouterServiceContainer`** wires it up dev-only for progressive document responses: when a route ships ≥ 1 kB of deferred text while its shell carries < 64 readable chars, it logs once per route per boot:

  `[gemi] GET / streams 3.7 kB of its content behind reveal scripts — a no-JS visitor sees ~0 chars. If this is a public/content route, add "no-stream" to its middleware.`

Settled documents (bot UAs, `"no-stream"` routes) skip collection entirely — their content is inline by construction — so adding `"no-stream"` (or shrinking the page under the floor) clears the hint, per the acceptance criteria.

## Tests

`services/router/shellContentReport.test.ts` (10 tests, all passing):

- Scanner: readable text vs markup/scripts/styles/titles/comments; nested divs inside `<div hidden>` segments; `hidden` in an attribute value not treated as a segment; whitespace never counts.
- The #289 repro numbers: 0 shell / 3,666 deferred flags (message says `3.7 kB` / `~0 chars`); `/about-us` at 4,150 shell / 0 deferred stays silent; both threshold edges.
- Observer: multi-byte characters split across chunks decode correctly.
- Integration: a real suspended `renderToReadableStream` render piped through `injectQueryPayloads` with `onChunk` wired measures the fallback-only shell and the deferred segment, and trips the hint.

`bunx vitest run` and `tsc --noEmit` are clean apart from failures that pre-exist on the base branch (verified by stashing).

### DIFF
--- a/packages/gemi/services/router/ViewRouterServiceContainer.ts
+++ b/packages/gemi/services/router/ViewRouterServiceContainer.ts
@@ -34,11 +34,7 @@ import { KernelIdServiceContainer } from "../kernel-id/KernelIdServiceContainer"
 import { ServerQueryStore, type StreamSummary } from "./ServerQueryStore";
 import { createServerQueryFetcher } from "./serverQueryFetcher";
 import { injectQueryPayloads, isBotUserAgent } from "./streamQueryInjection";
-import {
-  createShellContentObserver,
-  shellContentHint,
-  type ShellContentMeasurement,
-} from "./shellContentReport";
+import { createShellContentObserver, createShellContentReporter } from "./shellContentReport";
 import { createRoutePayloadStream } from "./routePayloadStream";
 
 /**
@@ -118,10 +114,14 @@ export class ViewRouterServiceContainer extends ServiceContainer {
   flatComponentTree: string[] = [];
   root: any = null;
   /**
-   * Routes already hinted for near-blank shell text — the hint fires once per
-   * route per boot (#294) so a hot route doesn't spam the dev console.
+   * The shell-content hint (#294): React defers any boundary over
+   * `progressiveChunkSize` on size alone, so a page that reads fine without
+   * JS can go blank the day one more section pushes it over the budget — and
+   * nothing else notices the transition (#289 was found by hand-measuring).
+   * The reporter owns the once-per-route-per-boot dedup so a hot route
+   * doesn't spam the dev console.
    */
-  private shellHintedRoutes = new Set<string>();
+  private shellReporter = createShellContentReporter();
 
   constructor(public service: ViewRouterServiceProvider) {
     super();
@@ -171,22 +171,6 @@ export class ViewRouterServiceContainer extends ServiceContainer {
     } catch (err) {
       logError(err);
     }
-  }
-
-  /**
-   * The shell-content hint (#294): React defers any boundary over
-   * `progressiveChunkSize` on size alone, so a page that reads fine without
-   * JS can go blank the day one more section pushes it over the budget — and
-   * nothing else notices the transition (#289 was found by hand-measuring).
-   * The injector's `onChunk` collected the exact bytes the visitor got;
-   * measured here once the body closed, hinted once per route per boot.
-   */
-  private maybeReportShellContent(routePath: string, measurement: ShellContentMeasurement) {
-    if (this.shellHintedRoutes.has(routePath)) return;
-    const hint = shellContentHint(routePath, measurement);
-    if (!hint) return;
-    this.shellHintedRoutes.add(routePath);
-    console.warn(hint);
   }
 
   private async render(props: {
@@ -366,7 +350,7 @@ export class ViewRouterServiceContainer extends ServiceContainer {
         process.env.NODE_ENV !== "production" &&
         !settled &&
         currentPathName &&
-        !this.shellHintedRoutes.has(currentPathName)
+        this.shellReporter.shouldObserve(currentPathName)
           ? createShellContentObserver()
           : null;
 
@@ -431,7 +415,9 @@ export class ViewRouterServiceContainer extends ServiceContainer {
             onClose: () => {
               this.completeStream(req, serverQueries.summarize(deadline.signal.aborted));
               if (shellObserver) {
-                this.maybeReportShellContent(currentPathName, shellObserver.measure());
+                // The injector's `onChunk` collected the exact bytes the
+                // visitor got; measured once the body closed (#294).
+                this.shellReporter.report(currentPathName, shellObserver.measure());
               }
             },
           }),
--- a/packages/gemi/services/router/ViewRouterServiceContainer.ts
+++ b/packages/gemi/services/router/ViewRouterServiceContainer.ts
@@ -34,6 +34,11 @@ import { KernelIdServiceContainer } from "../kernel-id/KernelIdServiceContainer"
 import { ServerQueryStore, type StreamSummary } from "./ServerQueryStore";
 import { createServerQueryFetcher } from "./serverQueryFetcher";
 import { injectQueryPayloads, isBotUserAgent } from "./streamQueryInjection";
+import {
+  createShellContentObserver,
+  shellContentHint,
+  type ShellContentMeasurement,
+} from "./shellContentReport";
 import { createRoutePayloadStream } from "./routePayloadStream";
 
 /**
@@ -112,6 +117,11 @@ export class ViewRouterServiceContainer extends ServiceContainer {
   componentTree: ComponentTree = [];
   flatComponentTree: string[] = [];
   root: any = null;
+  /**
+   * Routes already hinted for near-blank shell text — the hint fires once per
+   * route per boot (#294) so a hot route doesn't spam the dev console.
+   */
+  private shellHintedRoutes = new Set<string>();
 
   constructor(public service: ViewRouterServiceProvider) {
     super();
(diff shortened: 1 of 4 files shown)


packages/gemi/services/router/shellContentReport.ts
```diff
@@ -0,0 +1,153 @@
...
+ * `<template id="B:...">` fallback markers are empty anyway; skipping the tag
+ * also covers hand-written templates.
+ */
+const RAW_TEXT_TAGS = new Set(["script", "style", "template", "title"]);
+
+/**
+ * `hidden` as a standalone attribute (`<div hidden id="S:0">`, `hidden=""`),
+ * not a substring of an attribute value like `class="hidden-md"`.
+ */
+const HIDDEN_ATTR = /\shidden([\s=/]|$)/i;
```
nstfkc, line 41: **Confirmed bug: `HIDDEN_ATTR` matches inside quoted attribute values, producing false hints on readable pages.**

The regex is tested against the whole tag string, so any attribute *value* containing the space-delimited word `hidden` opens a deferred segment. Verified against this file as-is:

```ts
measureShellContent(`<body><div title="show hidden files">${'word '.repeat(400)}</div></body>`)
// => { shellChars: 0, deferredChars: 1600 }
```

which trips `shellContentHint` on a page whose entire content is shell-visible:

```
[gemi] GET /docs streams 1.6 kB of its content behind reveal scripts — a no-JS visitor sees ~0 chars. ...
```

Realistic triggers: `title="the hidden cost"`, `alt`/`aria-label`/`data-*` prose, and Tailwind class lists like `class="mt-2 hidden md:block"` (mid-list `hidden` matches; only first/last position happens to escape). The failure direction is the worst one for this feature — a false warning instructs the developer to add `"no-stream"` and give up streaming on a route that didn't need it, and the existing test only covers `class="hidden"` where the word touches the quote so no `\s` precedes it.

Since React escapes `"` in attribute values, stripping quoted spans before the test is safe and one line:

```ts
} else if (HIDDEN_ATTR.test(tag.replace(/"[^"]*"|'[^']*'/g, '""'))) {
  hiddenDepth = 1;
}
```

I verified this keeps `div hidden id="S:0"`, bare `hidden`, and `hidden=""` matching while rejecting all the quoted-value cases above. Worth adding `title="the hidden cost"` and `class="a hidden b"` cases next to the existing `class="hidden"` test.

(Related, lower stakes, fine to punt: a hand-written `<section hidden>`/`<p hidden>` is invisible to every reader but counts as shell text because only `div` is tracked — that errs in the false-negative direction, so it can just be a comment in the code.)

packages/gemi/services/router/ViewRouterServiceContainer.ts
```diff
@@ -163,6 +173,22 @@ export class ViewRouterServiceContainer extends ServiceContainer {
...
+   * JS can go blank the day one more section pushes it over the budget — and
+   * nothing else notices the transition (#289 was found by hand-measuring).
+   * The injector's `onChunk` collected the exact bytes the visitor got;
+   * measured here once the body closed, hinted once per route per boot.
+   */
+  private maybeReportShellContent(routePath: string, measurement: ShellContentMeasurement) {
+    if (this.shellHintedRoutes.has(routePath)) return;
+    const hint = shellContentHint(routePath, measurement);
+    if (!hint) return;
+    this.shellHintedRoutes.add(routePath);
```
nstfkc: **Test coverage: the "logs the hint once" acceptance criterion is untested.**

Issue #294's first acceptance bullet is "logs the hint once" per route, and the PR's stated behavior is once per route per boot — but the dedup lives entirely here (`shellHintedRoutes` + the observer-creation guard) and `shellContentReport.test.ts` never exercises it; every test calls the pure functions directly. A regression here (e.g. adding to the set before the null-hint check, or dropping the guard at observer creation) would ship silently while the whole suite stays green.

A small unit test is enough — call `maybeReportShellContent` twice with a failing measurement and assert `console.warn` fired once, and once with a healthy measurement asserting the route is *not* added to the set (so a route that later crosses the threshold still gets its hint).

nstfkc (commented):
Reviewed against the base branch `feat/293-stream-lifecycle-hooks` only. The design is sound: hanging the observer on the injector's `onChunk` measures exactly the bytes the visitor got, settled (bot/`no-stream`) responses correctly skip collection so adding `"no-stream"` clears the hint, dedup keys on `match.routePath` (the route pattern, not the URL) so dynamic routes can't spam or grow the set unboundedly, and truncated streams (client cancel) can only under-count deferred text — never fabricate a hint from truncation, since the shell always precedes the segments. The #289 repro numbers and both threshold edges are pinned by tests.

One confirmed correctness bug: `HIDDEN_ATTR` matches the word `hidden` inside quoted attribute values, which flips the scanner into deferred mode on a fully readable page and emits a false hint telling the developer to disable streaming (details + verified repro inline). One smaller note on test coverage for the once-per-route acceptance criterion.

nstfkc:
Addressed both review points in b8b8f6e:

**1. `HIDDEN_ATTR` matching inside quoted attribute values — fixed.** Took the suggested quote-stripping approach: quoted spans (double- and single-quoted) are blanked to `""` before the `HIDDEN_ATTR` test, with a comment noting why the regex is safe (React escapes quotes inside attribute values, so a span never contains its own delimiter). Re-ran your repro against the fixed module: `title="show hidden files"` + 1600 chars now measures `{ shellChars: 1600, deferredChars: 0 }` — no false hint. Added the suggested test cases and more: `title="show hidden files"`, `title="the hidden cost"`, `class="a hidden b"`, `class="mt-2 hidden md:block"`, a single-quoted variant, plus a companion test pinning that every standalone spelling (`<div hidden id="S:0">`, bare `hidden`, `hidden=""`, attribute-last position) still opens a deferred segment. Also took the "fine to punt" note as a comment: the `name !== "div"` check now documents that a hand-written `<section hidden>` counts as shell text — a false negative (stays silent), never a false hint.

**2. Once-per-route dedup untested — fixed, via a small extraction.** Testing `maybeReportShellContent` directly would have meant constructing the container (satori/sharp/auth imports), so the dedup moved into `shellContentReport.ts` as `createShellContentReporter(warn = console.warn)` — a factory owning the hinted-routes set, exposing `report()` and `shouldObserve()` (the observer-creation guard). The container holds one instance; behavior is unchanged. New tests pin exactly the regressions you named: two failing reports warn once (adding to the set before the null-hint check would fail the healthy-measurement test), a second route still gets its own hint, `shouldObserve` flips only after an emitted hint (dropping the guard's semantics fails it), and a healthy measurement does not consume the route's one hint so a route that later crosses the threshold still warns.

All 14 tests in `shellContentReport.test.ts` pass and `tsc --noEmit` is clean. The two other failures in `services/router/` (`assertNoReservedRoutePaths.test.ts`, `createRouteManifest.test.ts`) pre-exist on the branch HEAD and are unrelated (vitest can't resolve the `bun` package in this environment / an unrelated manifest expectation).

Issue #271, opened 2026-07-29
Multi-field relations in a nested write are refused, and the refusal is pending rather than declined


nstfkc:
The one **pending** entry in `docs/orm.md`'s *Not in scope* list — the others there (`distinct`, `cursor`) are decisions, and this one is unfinished work that reads like a decision if you skim.

## What works and what does not

`@relation(fields: [a, b], references: [c, d])` works everywhere a relation is **read**: `include` under either strategy, a relation filter, `_count`, and an `orderBy` through a relation all correlate on every field. That is #67, and #97 added the Postgres `unnest` form so a batched composite `include` stops churning the plan cache.

Writing through one is refused. `planNestedWrites` narrows the link to a single field on purpose:

```
// **Narrowed to one field, deliberately.** Reading across a composite
// relation works (#67); writing through one would have to contribute that
// many foreign-key columns to this insert, which is a different piece of
// work — so it is refused here by name rather than silently writing the
// first field.
```

The narrowing is a function call rather than an index, so there is no single-field property on `Link` to reach for by accident — the refusal is structural rather than a check someone has to remember.

The error names the fields on both sides and the operation it came from. So the failure is loud, specific, and cannot write plausible wrong rows, which is the property that makes shipping without this defensible.

## Why it deserves an issue rather than a docs line

`docs/orm.md` marks it *"pending rather than declined"*, which is the honest word and also the one nothing tracks. The two entries beside it throw `UnsupportedByDesignError` — a distinct class, so a caller can tell *"this is a decision"* from *"not yet"* programmatically. This one throws the plain `UnsupportedQueryError`, correctly, and there is no issue number attached to the *yet*.

## Scope, so it is not mistaken for small

The work is contributing N foreign-key columns to an insert instead of one, in every operand that can write a key — `connect`, `create`, `connectOrCreate`, `set`, `disconnect` — and on both sides of the relation. `resolveLink` already computes the full multi-field link; `singleFieldLink` is the narrowing that would come off. What makes it more than plumbing is the ordering: a nested `create` writes the child before the parent's key exists in one direction and after it in the other, and with N columns the *partial* failure mode appears — some keys written, some not — which is the thing the current refusal makes impossible.

Also needs the template to grow a composite relation the differential harness can reach. `templates/saas-starter`'s schema has none, so `composite-relations.test.ts` asserts against Prisma's documented shape rather than a second generated client — the same limitation the plan records for implicit many-to-many.

[...]

Issue #270, opened 2026-07-29
The template still points app/config/auth.ts at the Prisma adapter, not the ORM one


nstfkc:
The last of `plans/orm/README.md`'s **Open decisions**, and the one the plan itself flags as *"a behaviour change to a working application"* that wants its own call.

## Where it stands

`OrmAuthenticationAdapter` ships. It implements all twenty-two methods of `IAuthenticationAdapter` on this ORM, with no changes to the ORM to make them fit — which the plan records as *the first evidence the query surface is sufficient rather than merely tested*, since they were written against a real application's needs rather than against the compiler's known capabilities.

Both adapters satisfy the same interface, so an application selects one and nothing else in `auth/` knows which. `packages/gemi/auth/adapters/prisma.ts` and the template's `app/database/prisma.ts` are untouched.

The template is still on the Prisma one:

```ts
// templates/saas-starter/app/config/auth.ts
import { PrismaAuthenticationAdapter } from "gemi/kernel"
…
userProvider: new PrismaAuthenticationAdapter(prisma),
```

## The question

Should the starter template ship pointed at `OrmAuthenticationAdapter`?

**For:** it is the thing that finally lets `auth/adapters/prisma.ts` be deprecated, which is #33's stated goal for the whole database-layer effort. A template that ships on the old adapter is a template that teaches the old adapter, and every application generated from it inherits that choice silently. It also puts the ORM on the hot path of the one flow every application runs, which is the strongest test the query surface could get.

**Against:** it is a behaviour change to a working application, made at the same moment as a large merge. If something is wrong with the twenty-two methods under real load, the failure lands on sign-in — the least recoverable place in a starter — and the switch would be the obvious suspect in a merge that changes a great deal else at once. Doing it as its own change, after the stack lands, isolates it.

## Recommendation

Switch it, but **not in the merge that lands the stack** — as a separate change afterwards, so a regression has one plausible cause instead of four hundred commits. `docs/orm.md`'s **See also** already describes the adapter and says *"the starter template still points at the Prisma adapter; switching it is a per-application call"*, which stays true for applications either way and needs one word changed if the template moves.

nstfkc:
## Closing: the question dissolved rather than being answered

This asked whether the starter template should be repointed from
`PrismaAuthenticationAdapter` to `OrmAuthenticationAdapter`, and recommended
doing it *after* the stack landed so a regression would have one plausible cause.

[...]

Issue #268, opened 2026-07-29
Land the ORM stack: PR #45 needs a new base, and 117 of 120 ORM branches hold nothing unique


nstfkc:
The plumbing between `feat/orm` and `main`, separated from the ORM work so it can be done by someone not holding the compiler in their head.

## Where it is stuck

```
main
 └── #30  refactor/laravel-container-architecture   OPEN  MERGEABLE/CLEAN
      └── #33  feat/database-layer                  OPEN  CONFLICTING/DIRTY   <- here
           └── #45  feat/orm                        OPEN  MERGEABLE/CLEAN, CI green both dialects
```

**#45 is ready.** `sqlite` and `postgres` both pass, on the proposal and on the merge result. Nothing about the ORM is blocking. #33 is `CONFLICTING` and is the only red link in the chain.

## The thing that makes this odd, and worth reading before starting

**#30 and #33 have already landed** — on `release/v0.50.0-rc.2`, not on `main`, and neither PR was closed:

```
d51f8a4 Release 0.50.0-rc.2
994ebed Merge PR #33: database layer on Bun's SQL client
0cd3384 Merge PR #30: Laravel container/provider architecture
```

`git merge-base --is-ancestor` confirms both branches are contained in `rc.2` and in neither case in `main`. So #33 reports a conflict against a base it arguably no longer needs to target, and its conflict resolution **already exists** on `rc.2` — the container refactor meeting `main`'s range-request / Azure / S3 / rate-limiter work, resolved once already.

That is why the merge cost depends entirely on the target:

| target | conflicts |
| --- | --- |
| `origin/release/v0.50.0-rc.2` | **2** — `bun.lock`, `packages/gemi/package.json` |
| `origin/main` | **10** — 8 of them the already-resolved container/storage conflicts, plus the two above |

Measured with `git merge-tree --write-tree` against `feat/orm` at `b6aa637`.

The decision to keep the three-deep stack has been made, so this issue is about landing it as a stack. But whoever resolves #33 should look at `rc.2` first rather than resolving those eight files from scratch — the answer is already written down.

Version skew to expect in `packages/gemi/package.json`: `0.43.0` on `feat/orm`, `0.45.0` on `main`, `0.50.0-rc.2` on the release branch. `plans/orm/README.md` predicted this one and calls it small and mechanical.

## Branch hygiene, measured

[...]

Issue #265, opened 2026-07-29
feat/orm: what is left before it can merge, and where it should merge to


nstfkc:
Measured on `feat/orm` at `f228df3` and `main` at `f2c68ef`, 2026-07-29.

## Verdict

**The ORM is done. What stands between it and `main` is stack plumbing, not ORM work.**

PR #45 (`feat/database-layer` ← `feat/orm`) is `MERGEABLE` / `CLEAN` with all four checks green — `sqlite` and `postgres`, on the proposal and on the merge result. All nine iterations in `plans/orm/` have shipped; the plan's own iteration map says so and the branch backs it up:

| | |
| --- | --- |
| ahead of `main` | 387 commits |
| ORM diff | 142 files, +57,031 |
| test files | 57 in `packages/gemi/orm/`, 25 in `templates/saas-starter/app/models/` |
| suite | 67 files / 1,609 passing (package, CI's exclusion set); 640 passing (template, SQLite) |
| open ORM issues that are actually *ORM* bugs | **one** (#259) |

The remaining work is four small issues, two in-flight PRs pointed at a dead branch, and three decisions.

---

## 1. The stack is the blocker, and it is not the stack the plan describes

`plans/orm/README.md` still routes the reader through three open PRs to `main`:

```
main
 └── #30  refactor/laravel-container-architecture   OPEN, MERGEABLE/CLEAN
      └── #33  feat/database-layer                  OPEN, CONFLICTING/DIRTY
           └── #45  feat/orm                        OPEN, MERGEABLE/CLEAN
```

**Both of the lower two have already landed** — on `release/v0.50.0-rc.2`, not on `main`, and their PRs were never closed:

```
d51f8a4 Release 0.50.0-rc.2
994ebed Merge PR #33: database layer on Bun's SQL client
0cd3384 Merge PR #30: Laravel container/provider architecture
```

`git merge-base --is-ancestor` confirms both branches are contained in `rc.2` and in neither case in `main`. So #33 reports `CONFLICTING/DIRTY` against a base it no longer needs to target, and #30 proposes to `main` a change that shipped in an RC three days ago. **Two open PRs are describing history.**

This changes the answer to "how far is it", because the conflict cost depends entirely on which branch you aim at:

| merge target | conflicts |
| --- | --- |
[...]

PR #299, opened 2026-08-02
Ship the open ORM issues for 0.51.0

### LINKED ISSUE #262: The soft-delete `field` constraint is keys, not columns — and only on one of four spellings
Split out of the review on #243, which narrowed the *prose* around this. The prose is now accurate; the type is still the wrong shape, in two independent ways.

### 1. `field` is unchecked on three of the four spellings

`SoftDeleteOptions<M>` constrains `field` to `keyof M & string` and `M` defaults to `any`. Only `softDeletes<User>()` supplies it.

| call | `field` checked? |
| --- | --- |
| `softDeletes<User>({ field: "nope" })` | yes |
| `softDeletes({ field: "nope" })` | no |
| `softDelete(User, { field: "nope" })` | no |
| `softDeleteMany(User, { field: "nope" })` | no |

A caller cannot even opt in on the write half: the parameters are `<T, M = any>` in that order, so `softDelete<User>(User, …)` binds `T`, not `M` — and binding `T` to the class instance type is itself an error, because `update` returns the row without the instance's methods.

There **is** a mechanism that recovers the row type from the model value, found in review:

```ts
-  model: SoftDeletable<T, "update">,
+  model: SoftDeletable<T, "update"> & { prototype?: M },
```

Verified against this repo: it makes both unchecked calls errors, `tsc --noEmit` on `packages/gemi` stays at exit 0, `bun --bun vitest run orm database` stays at 1499 passed, and the template's `policies.test.ts` stays green — including the suites that pass structural fakes to `softDelete`, which still work because `prototype` is optional and `M` falls back to `any` when absent. The `SoftDeletable` comment ("structural rather than `typeof Model`, so a caller can pass anything that behaves like one") survives intact.

It was deliberately **not** taken in #243, for the reason below.

### 2. `keyof` an instance type is not a column check

The error text the `prototype` patch produces is the argument against merging it on its own:

```
Type '"nope"' is not assignable to type '"deletedAt" | "id" | "publicId" |
  "organizationId" | ... | "save" | undefined'.
```

`"save"` is in that union. The docs tell readers to write `softDeletes<User>()` where `User` is the model class, so `keyof M & string` is the keys of the *instance* type, methods included. Confirmed both ways:

- `softDeletes<User>({ field: "save" })` compiles clean.
- At runtime: `'save' is not a field on model User. Known fields: id, publicId, name, …`

So the check catches typos and not non-columns. Taking the `prototype` patch alone would spread that same half-check onto two more functions and license a docs sentence saying `field` is checked everywhere — which is the species of overclaim #242 existed to correct, one level down. That is why #243 recorded the mechanism in `soft-delete.test-d.ts` and left the signature alone rather than half-fixing it.

### What this issue wants

Both at once: constrain `field` to the **schema's field names** (`$schema.fields`, which the generated models already carry and which `UnknownFieldError` already lists), and make that constraint reachable from the model value so all four spellings get it. `prototype` is the plausible route to the second half; the first half is the part that makes it worth doing.

Then the sentence in `docs/orm.md`, the `SoftDeleteOptions.field` doc comment in `packages/gemi/orm/soft-deletes.ts`, and the boundary pinned in `templates/saas-starter/app/models/soft-delete.test-d.ts` all get to say something stronger — and three tests in that file are written *without* `@ts-expect-error` on purpose, so tightening the inference breaks the file and forces the note to be revisited. That is the tripwire; this is the change meant to trip it.

### LINKED ISSUE #263: The documented soft-delete recipe does not type-check: `static delete = softDelete(User)` is an illegal override
Found while addressing review point 6 on #243 — the reviewer's complaint was that the new snippet used `archivedAt`, which the template's `User` does not have. Picking a real column does not make the snippet compile. The recipe itself does not.

`docs/orm.md`, `docs/llms-full.txt` and the `softDelete` doc comment in `packages/gemi/orm/soft-deletes.ts:110-114` all teach this:

```ts
export class User extends UserModel {
  static $policies = [softDeletes<User>()]
  static delete = softDelete(User)
  static deleteMany = softDeleteMany(User)
}
```

Against the template's generated `UserModel`, that class does not type-check:

```
TypeCheckError: Class static side 'typeof User' incorrectly extends base class static side
  'typeof UserModel'.
  The types returned by 'delete(...)' are incompatible between these types.
    Type 'Promise<{ name: string | null; id: number; … }>' is not assignable to
      type 'Promise<GetFindResult<$UserPayload<DefaultArgs>, T, {}>>'.
```

`softDelete` returns `(args: any) => Promise<T>`, and `T` is the plain row; the generated `delete` is generic in its arguments and returns a `GetFindResult` narrowed by them. The assignment is not compatible in either direction, so *any* `static delete =` overriding a generated model's `delete` is an error regardless of the `field` value. Same for `deleteMany`.

Reproduced with a `.test-d.ts` in `templates/saas-starter/app/models` containing only the snippet above; it is the sole error, and it survives replacing `softDelete(User)` with any expression of that return type.

The repo's own soft-delete tests sidestep it without saying so: `templates/saas-starter/app/models/policies.test.ts:110` is

```ts
static softDelete = softDelete(GuardedUser);
```

— a *different name*, so there is no override to be incompatible with. Every runtime soft-delete test passes because none of them uses the spelling the docs teach.

Two ways out, and they are not equivalent:

1. **Change the docs** to a non-overriding name (`static remove = softDelete(User)`), which is what the tests already do and what #243's replacement snippet now uses (`static expire`). Cheap, and it leaves `User.delete()` meaning a hard delete — arguably a worse recipe, since the whole appeal is that `delete` transparently becomes an update.
2. **Make the override assignable** — widen `softDelete`'s return to match the generated signature, or have the generator emit a `delete` a soft-delete wrapper can legally replace. This is the version that keeps the recipe, and it is a real change to the generated surface.

Whichever is chosen, `soft-deletes.ts:110-114`, `docs/orm.md` and `docs/llms-full.txt` should end up agreeing with a snippet that has been compiled at least once. Worth a `.test-d.ts` that holds the documented recipe verbatim, so this cannot rot again.

Out of scope for #243, which is about the `field` option specifically; this predates it and affects every soft-delete snippet in the docs.

### LINKED ISSUE #267: JSON path filters were built and never merged, and #70 was closed covering them
Found by auditing the 120 ORM branches on `origin` for work `feat/orm` does not have. Three hold unique non-merge commits; this is the substantial one.

## The work exists

`origin/feat/orm-json-filters` (and `origin/feat/orm-json-coverage`, same commit) carries `1fae1ea`, *"JSON path filters, and the dialect split they inherit from Prisma"*, dated 2026-07-28. Its own message says it **"closes the second half of #70"**.

It is not a sketch. The commit message records the feature having been measured on both dialects through a generated client:

- The **path grammar differs by dialect** — Prisma takes `path: ["a","b"]` on Postgres and `path: "$.a.b"` on SQLite, and refuses the other form on each. It sits behind the `SqlDialect` seam as `jsonPathSyntax`, refusing with the form *this* database wants rather than letting the driver fail.
- **Which filters apply differs too:**
  ```
  equals, not, string_contains/starts_with/ends_with   both
  array_contains, lt, lte, gt, gte                     postgres only
  ```
  Prisma refuses the second row on SQLite with "Unknown argument", so `jsonFilters` refuses them as well — the stated reason being that implementing them there would make gemi answer a query the differential harness has no oracle to check.
- **The path is bound, never interpolated**, with the reasoning written out: it is the one place a caller value decides part of an expression's meaning, and neither dialect needs it in the text — Postgres's `#>` takes a `text[]` parameter and SQLite's `json_extract` a string.

## What is on `feat/orm` instead

Nothing. `OPERATORS` in `compile/where.ts` is still the twelve scalar operators:

```
equals, not, in, notIn, lt, lte, gt, gte, contains, startsWith, endsWith, mode
```

`path` is absent, and `grep -rn '"path"' packages/gemi/orm/` returns nothing. So `where: { metadata: { path: …, equals: … } }` is still refused.

## Why nobody noticed

#70 was **two** findings in one issue — *"no `Json` column is exercised against a real database, **and** JSON path filters are unsupported"*. The first half landed as `9c8f3ef` (*"Json against a real database, and the two bugs that were hiding each other"*). The second half is `1fae1ea`, which never reached `feat/orm`.

The issue was closed. Closing on the first half is the ordinary outcome of an issue that names two things and gets one PR — and it is why an audit of merged issues would not have found this, while an audit of *branches* did.

## What to decide

Two questions, and they are not the same:

1. **Does the work still apply?** It is dated 2026-07-28 and `where.ts` has moved since — #248's mutation work, #250's scoping, and the `AnyNull` handling in b6aa637 all touch `compileFieldFilter`. Rebasing it is the first cost.
2. **Is it in scope for this release?** `docs/orm.md` currently documents no `path` support at all, so shipping without it is coherent — but then #70's second half should be reopened as a known gap rather than left looking closed, and `docs/orm.md`'s **Not in scope** section should say so, the way it does for `distinct` and `cursor`.

Either answer is defensible. What is not defensible is the current state: a complete feature on a branch nobody will look at, and an issue saying it is done.

Filed as its own issue rather than by reopening #70, because half of #70 genuinely did land and the record should say which half.

### LINKED ISSUE #269: Decision: ship the ORM with MySQL/MariaDB refused rather than supported
One of the two remaining entries under **Open decisions** in `plans/orm/README.md`. It is a decision rather than work — nothing is half-built.

## Where it stands

`DatabaseManager` infers all four dialects from `DATABASE_URL`. Only SQLite and Postgres have compilers. The other two are refused rather than left to fail:

- `resolveCompiler` throws `UnsupportedDialectError` (`orm/dialect/index.ts`), which the docs table describes as *"a model operation on a dialect with no compiler — MySQL and MariaDB"*.
- `ReturningUnsupportedError` is the same gap seen from the write path, so a write on a dialect without `RETURNING` names itself rather than emitting broken SQL.
- The error is careful to say the *connection* is fine and the *ORM* is what does not speak that dialect.

So #71 — *"support it, or refuse it explicitly"* — was answered by refusing, and that shipped. What is left is a release call, not a build.

## The question

Is refusing acceptable for the release this stack lands in?

**If yes:** nothing to build. `docs/orm.md`'s **Dialects** section already says which two are supported, and the errors table already names the refusal — this issue closes as confirmation, and `plans/orm/README.md` should be edited to say *settled* rather than *deferred*, the way the ambient-transaction and implicit-m2m entries were.

**If no:** it is a real iteration. The strategy seam keeps the door open by design, but the differential harness is the constraint rather than the compiler — every correctness claim in this ORM is *"the same query through Prisma and through gemi against one database, compared on rows and table contents"*, and that needs a MySQL Prisma client, a MySQL service container in CI, and a decision about `RETURNING`, which MySQL does not have. The write path's row counts come from `RETURNING` deliberately (see the header comment in `compile/write.ts`), so MySQL is not a dialect file — it is a second answer to how writes report what they did.

## Why it is worth an explicit answer rather than silence

A deferral nobody records reads as an oversight to the first person who points `DATABASE_URL` at MySQL. The refusal is good — it names the dialect and says the connection is fine — but the *documentation* of it is one row in an errors table, and the plan still lists this as undecided. Whichever way it goes, the plan and the docs should agree with each other and with the code.

nstfkc:
Closes #262, closes #263, closes #267, closes #269.

The `orm`-labelled issues that were work rather than decisions, for the release
being cut from `next`. Three commits, verified on both dialects.

## What was actually open

The two tracker issues turned out to describe a world that had moved. #265 and
#268 route the reader through landing `feat/orm` as a three-deep PR stack onto
`rc.2` or `main` — but `feat/orm` is already an ancestor of `origin/next`, PRs
#30/#33/#45 are all closed, and `next` carries `0.51.0-rc.0` with #273 open to
`main`. So the "stack plumbing" that both issues call the blocker is done, and
what was left was four pieces of real work.

## `cc22399` — JSON path filters (#267, and the second half of #70)

`1fae1ea` was complete, green, and never merged: #70 named two findings, the
first landed, the issue was closed on it, and the second sat on a branch. Rebased
rather than replayed — `compile/where.ts` moved underneath it:

- The `path` branch is placed **after** the `AnyNull`/sentinel refusal from
  `b6aa637`, so `Prisma.DbNull` is still refused by name rather than read as an
  operator map that happens to carry a `path`.
- `where.ts` binds through `fieldParam` now, which adds the `::text::jsonb` cast
  a Json *column* needs (#209). A path comparison binds an extracted scalar, not
  a column, so it uses `param` directly.
- The dialect members landed beside #97's composite-`in` work, which had claimed
  the same insertion point in all three files.

`docs/llms-full.txt` is regenerated — `docs-scope.test.ts` postdates the original
commit and caught the omission.

## `5f959ff` — soft deletes (#262, #263)

One commit because they retype the same two functions.

**#262** had two defects and the second is the one that mattered. The constraint
reached one of four spellings, and where it applied it was `keyof` an *instance*
type — so `field: "save"` compiled clean and raised `UnknownFieldError` at
runtime. A check that disagrees with the runtime error about what a field is
catches typos and nothing else.

The fix starts in the generator, which is why #243 could not get here:

```diff
-export const User: ModelSchema = { … };
+export const User = { … } satisfies ModelSchema;
```

The annotation was widening `fields` to `Record<string, FieldSchema>`, leaving no
column names in the type system to constrain against. The emitted value is
byte-identical and the regeneration test passes unchanged.

**#263** — `static delete = softDelete(User)` cannot be made to type-check. The
repair that keeps the name is circular: the member's type would be read off the
class the member is being defined on (`TS7022`). Annotating does not break the
cycle, and passing the generated base breaks the feature, since policies resolve
per registered class and the base carries none. So the recipe changed to
`static expire`, and the consequence is now stated rather than implied:
**`User.delete()` on a soft-deleting model is still a hard delete.**

## `2b087ec` — decisions and stale prose (#269, plus #268/#270 sections)

#269 settled as **refused**: nothing was half-built, and the constraint is the
differential harness rather than the compiler — MySQL has no `RETURNING`, which
is where the write path's row counts come from deliberately.

Two plan sections retired. `Stack position` would have sent a reader to branch
under a stack that no longer exists; the `Coexistence` entry held open a question
that `922c45d` dissolved by retiring `IAuthenticationAdapter` entirely (#270).

Also carries the runtime pin from `719bb4c`, the commit stranded on
`fix/orm-soft-delete-field-constraint`. Its write-half case is now a *compile*
error, so it bypasses the check with a cast rather than being dropped — the
backstop still has to hold for callers arriving from untyped code.

## Verification

| | |
| --- | --- |
| `packages/gemi` | 100 files / 2082 tests, CI's exclusion set; `tsc` clean |
| template, sqlite | 654 tests + 36 type tests |
| template, postgres | 928 tests, real server via `app/models/postgres.sh` |

The 5 files failing outside CI's exclusion set are #163's `app/` copies,
untouched here.

## Not in this PR

- **#271** — multi-field relations in a nested write. Still refused, deliberately;
  the issue says no urgency and the work is N foreign-key columns across five
  operands plus a partial-failure mode the current refusal makes impossible.
- **#268's branch deletion** — 117 of 120 ORM branches now hold zero unique
  non-merge commits against `next`, re-measured. The three that did are all
  resolved by this PR, so all 120 are deletable; the list is on the issue rather
  than run from here.
- **`corpus.ts` is not extended for `path`.** It says it is the thing to extend
  when a new argument is implemented, and the invariant suites walk it against
  `userWithProfile`, which has no Json column — so covering it means adding one
  to a shared fixture and re-baselining every suite that reads all columns. Too
  broad for a release. The correctness direction still holds by construction:
  filter names are entries in the plan key, so two JSON shapes compiling to
  different SQL cannot share one.

## Breaking changes

**`SqlDialect` gains five required members** — `jsonPathSyntax`, `jsonFilters`,
`jsonComparesAsText`, `jsonExtract`, `jsonArrayContains`. The interface is
exported, so any out-of-tree implementor of it stops compiling until they are
added. Only the two in-tree dialects exist today, so this is a release-note line
rather than a migration.

Nothing else in the public surface changes. `SoftDeleteOptions` and
`softDelete` / `softDeleteMany` keep their existing type parameters — an earlier
revision narrowed both and that was reverted in `453c4ad`, so
`SoftDeleteOptions<User>` and `softDelete<Row>(model)` compile as before.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3694,11 +3694,23 @@ Ships as a policy, which is the proof the hook is expressive enough to be worth 
 import { softDelete, softDeleteMany, softDeletes } from "gemi/orm"
 
 export class User extends UserModel {
-  static $policies = [softDeletes<User>()]  // reads skip rows with a deletedAt
-  static delete = softDelete(User)          // delete becomes an update
-  static deleteMany = softDeleteMany(User)
+  static $policies = [softDeletes<typeof User>()]  // reads skip rows with a deletedAt
+  static expire = softDelete(User)                 // an update that sets deletedAt
+  static expireMany = softDeleteMany(User)
 }
 ```
+
+> **`expire`, not `delete`, and that is a limitation rather than a style.** Naming the wrapper
+> `delete` would be the better recipe — `User.delete()` would transparently become an update — and
+> it cannot be written: the member's type would have to be read off the class the member is being
+> defined on, which TypeScript refuses with *"'delete' implicitly has type 'any' because it does
+> not have a type annotation and is referenced directly or indirectly in its own initializer"*.
+> Annotating it does not help, and passing the generated base instead of your subclass breaks the
+> feature, because policies are resolved per registered class and the base carries none.
+>
+> **So `User.delete()` remains a hard delete.** On a model that soft-deletes, that is a real edge:
+> reads are scoped by the policy, but a `delete` still removes the row. Call `User.expire()`, and
+> if a hard delete should be unreachable, keep the model's `delete` behind a policy of your own.
 
 `deletedAt` must be a nullable `DateTime` on the model (`field` overrides the column name). The
 read half applies to nested reads for the same reason every policy does — a deleted `Account` does
@@ -3708,27 +3720,34 @@ the include site, which is the half ORM-level soft deletes usually get wrong.
 Composing it with your own is one list:
 
 ```ts
-static $policies: UserPolicy[] = [softDeletes<User>(), new TenantPolicy()]
+static $policies: UserPolicy[] = [softDeletes<typeof User>(), new TenantPolicy()]
 ```
 
 `softDeletes()` carries an `onUpdate` pass-through, because `softDelete()` writes the very column
 the policy scopes on and something has to say that is intended. That permission is **per policy**:
 a tenant policy sitting beside it in the same list still has to answer for its own column.
 
-`field` is constrained to the model's own keys **when the model type is given** —
-`softDeletes<User>({ field: … })`, as above — so a typo there is a compile error rather than a
-`no such column` on the first read.
-
-`softDelete(User)` and `softDeleteMany(User)` take the model as a *value*, and its row type is not
-recoverable from it, so their `field` is unchecked. Spell it once and share it if you override the
-default:
+`field` is constrained to the model's **schema columns** — the same list `UnknownFieldError` prints
+— on every spelling that names a model:
 
 ```ts
-const archived = { field: "archivedAt" } as const
+softDeletes<typeof User>({ field: "archivedAt" })   // checked
+softDelete(User, { field: "archivedAt" })           // checked
+softDeleteMany(User, { field: "archivedAt" })       // checked
 
+softDeletes({ field: "archivedAt" })                // unchecked — no model named
+```
+
+Columns rather than keys is the part that matters: the constraint used to be `keyof` the *instance*
+type, so `field: "save"` — a method — compiled clean and failed at runtime. Pass the model as
+`typeof User` and it is a compile error.
+
+Nothing needs spelling twice. If you override the default, each call checks its own argument:
+
+```ts
 export class User extends UserModel {
-  static $policies = [softDeletes<User>(archived)]   // checked here
-  static delete = softDelete(User, archived)          // and therefore correct here
+  static $policies = [softDeletes<typeof User>({ field: "archivedAt" })]
+  static expire = softDelete(User, { field: "archivedAt" })
 }
 ```
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3059,6 +3059,23 @@ implementing them there would answer a query the differential harness has no ora
 expression's meaning, and both dialects take it natively — Postgres's `#>` accepts a `text[]`,
 SQLite's `json_extract` a string — so the feature fits inside invariant 2 rather than bending it.
 
+Three things a path filter refuses, each because answering would be silently wrong rather than
+merely unsupported:
+
+- **The null sentinels.** `{ path: […], equals: Prisma.DbNull }` is refused. An extracted value
+  cannot tell an absent key from a JSON `null` — `#>>` yields SQL NULL for both — so the
+  distinction the sentinels exist to make is already gone by the time the comparison happens.
+  Filter the column itself: `{ metadata: { equals: Prisma.DbNull } }`.
+- **A non-string operand to `string_contains` / `string_starts_with` / `string_ends_with`**, for
+  the same reason the scalar `contains` refuses one: the pattern would become `%null%`, which runs
+  and returns the wrong rows.
+- **An object or array under `equals` / `not` / the numeric comparisons.** Prisma accepts one;
+  gemi does not yet. It would bind as `"[object Object]"` on Postgres and match nothing. Use
+  `array_contains` for containment, or narrow the path until it names a scalar.
+
+An empty path — `[]` or `""` — is refused on both dialects. On Postgres it would extract the whole
+document, which is a filter on the column rather than on a path.
+
 ### What a refusal tells you
 
 Three classes, and which one you get says what to do next:
@@ -3541,7 +3558,7 @@ initializer from an inherited static declaration, so without it `ctx` and `data`
 Policies compose by listing them, in the order they apply:
 
 ```ts
-static $policies: AccountPolicy[] = [softDeletes<Account>(), new TenantPolicy()]
+static $policies: AccountPolicy[] = [softDeletes<typeof Account>(), new TenantPolicy()]
 ```
 
 Entries may be objects or classes. A class is instantiated once, and extending the generated
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3029,6 +3029,36 @@ On SQLite the error names the dialect and says so. If a batch is too large for o
 split inside a transaction, and `skipDuplicates` survives the split: the counts sum, and a conflict
 in a later chunk does not roll back an earlier one, because `do nothing` is not an error.
 
+### JSON path filters
+
+```ts
+// postgres
+await User.findMany({ where: { metadata: { path: ["plan"], equals: "pro" } } })
+// sqlite
+await User.findMany({ where: { metadata: { path: "$.plan", equals: "pro" } } })
+```
+
+**The path grammar differs by dialect, and that is Prisma's split rather than this ORM's.** Its
+generated client takes an array of keys on Postgres and a JSONPath string on SQLite, and refuses the
+other form on each — so the argument is dialect-specific before it reaches gemi. The refusal here
+says which form the database you are on wants.
+
+Which filters apply also differs, again following Prisma:
+
+| | SQLite | Postgres |
+| --- | --- | --- |
+| `equals`, `not` | yes | yes |
+| `string_contains`, `string_starts_with`, `string_ends_with` | yes | yes |
+| `array_contains` | no | yes |
+| `lt`, `lte`, `gt`, `gte` | no | yes |
+
+Prisma refuses the bottom two rows on SQLite with *"Unknown argument"*, so gemi refuses them too:
+implementing them there would answer a query the differential harness has no oracle for.
+
+**The path is always a bound parameter.** It is the one place a caller's value decides part of an
+expression's meaning, and both dialects take it natively — Postgres's `#>` accepts a `text[]`,
+SQLite's `json_extract` a string — so the feature fits inside invariant 2 rather than bending it.
+
 ### What a refusal tells you
 
 Three classes, and which one you get says what to do next:
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2787,6 +2787,27 @@ Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 **The output is committed on purpose.** Diffs stay reviewable and CI needs no codegen step.
 
+### Columns the generator refuses
+
+Two kinds of column stop generation with an `UnsupportedSchemaError` naming the model and field,
+rather than being skipped:
+
+| Column | Why |
+| --- | --- |
+| A scalar list — `tags String[]` | Postgres-only, and needs array decoding this ORM does not have. |
+| `Decimal` | Prisma types it as `Prisma.Decimal`; SQLite stores a REAL and the driver returns a JS number, so the value would be a float pretending to be an arbitrary-precision decimal. |
+
+**Refused rather than omitted, and the whole generation fails rather than the one field.** Skipping
+the column would generate cleanly and then hand back a row shape that silently disagrees with the
+type Prisma gave you — which is the failure this ORM is built to make impossible. One unsupported
+column means no `schema.ts`, no `models.ts` and no `index.ts` for any model, so the problem is
+visible at the moment you introduce it.
+
+A `Unsupported(...)` column is different and is *not* refused: Prisma's own client omits those from
+its result types, so omitting them is what keeps the shapes identical.
+
+Scalar lists are tracked in [#300](https://github.com/nstfkc/gemi/issues/300).
+
 ### Your model class
 
 The generated base is not the class you write code on. Subclass it, and **re-register it**:
@@ -4078,6 +4099,10 @@ if (!ormSupports(DB.dialect)) throw new Error("this deployment needs Postgres")
 
 Stated so you can plan around them rather than discover them:
 
+- **No scalar lists.** `tags String[]` is refused at *generation* time, so this one stops your build
+  rather than a query — see [Columns the generator refuses](#columns-the-generator-refuses). Prisma
+  itself refuses them on SQLite, so half the supported matrix could never have had them.
+  Pending rather than declined ([#300](https://github.com/nstfkc/gemi/issues/300)).
 - **No identity map and no unit of work.** Two queries for the same row give you two objects. Half
   an identity map is worse than none.
 - **No lazy loading.** A relation you did not `include` is absent, not a proxy that queries when
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -298,6 +298,36 @@ On SQLite the error names the dialect and says so. If a batch is too large for o
 split inside a transaction, and `skipDuplicates` survives the split: the counts sum, and a conflict
 in a later chunk does not roll back an earlier one, because `do nothing` is not an error.
 
+### JSON path filters
+
+```ts
+// postgres
+await User.findMany({ where: { metadata: { path: ["plan"], equals: "pro" } } })
+// sqlite
+await User.findMany({ where: { metadata: { path: "$.plan", equals: "pro" } } })
+```
+
+**The path grammar differs by dialect, and that is Prisma's split rather than this ORM's.** Its
+generated client takes an array of keys on Postgres and a JSONPath string on SQLite, and refuses the
+other form on each — so the argument is dialect-specific before it reaches gemi. The refusal here
+says which form the database you are on wants.
+
+Which filters apply also differs, again following Prisma:
+
+| | SQLite | Postgres |
+| --- | --- | --- |
+| `equals`, `not` | yes | yes |
+| `string_contains`, `string_starts_with`, `string_ends_with` | yes | yes |
+| `array_contains` | no | yes |
+| `lt`, `lte`, `gt`, `gte` | no | yes |
+
+Prisma refuses the bottom two rows on SQLite with *"Unknown argument"*, so gemi refuses them too:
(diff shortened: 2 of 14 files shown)


nstfkc:
## Review

Checked the branch out in a worktree, ran `orm/compile` (695 tests green), and probed the new code
paths and type signatures directly — the findings below are confirmed by execution rather than by
reading.

**What holds up.** The path is genuinely bound on both dialects — `#>> $1` and
`json_extract(…, ?)`, nothing of the caller's path in the SQL text, and the plan is correctly reused
across different path values. The differential coverage is real: seven cases per dialect against a
fixture that already carries `{}`, `[]`, `"42"` and a null, so the NULL-propagation and shape edges
are actually exercised rather than assumed. The `satisfies` change is type-level only and the
emitted value is byte-identical.

Items 1–3 are the same shape of defect the surrounding code is unusually careful about, all in
newly added lines, and all silent rather than loud.

### 1. Null sentinels bypass both refusals inside a `path` filter

`compileFieldFilter` refuses a bare sentinel at the top, and refuses `AnyNull` under a
non-`equals`/`not` operator inside the key loop (`where.ts:970`). The new `path` branch sits at
`where.ts:922` — **between** them — so a sentinel *inside* a JSON filter reaches the binder, which
is exactly the failure class #259 existed to close. Measured:

```
{ metadata: { path: ["a"], equals: Prisma.DbNull } }   // postgres
  → ("metadata" #>> $1) = $2            bind: ["{\"a\"}", "Prisma.DbNull"]

{ metadata: { path: "$.a", equals: Prisma.AnyNull } }  // sqlite
  → (json_extract("metadata", ?)) = ?   bind: ["$.a", {}]
```

Postgres silently compares against the literal string `Prisma.DbNull` — zero rows, no error. SQLite
binds the sentinel object itself. `jsonNullKind` / `carriesAnyNull` on each operand at the top of
`compileJsonFilter`'s loop, raising `InvalidArgumentError` there, would close it in the place that
has the field and operation in scope to name them.

### 2. The JSON string filters skip the non-string guard their scalar siblings have

`where.ts:1020` refuses a non-string `contains` / `startsWith` / `endsWith`, with a comment
explaining that `String(null)` makes `%null%`, "a query that runs and returns the wrong rows".
`jsonComparison` reintroduces exactly that shape without the guard:

```
{ path: "$.a", string_contains: null }
  → (json_extract("metadata", ?)) like ?   bind: ["$.a", null]
```

`like NULL` matches nothing, silently; `string_contains: 5` becomes `%5%`. Prisma raises here.

### 3. `equals` with a non-scalar binds `"[object Object]"`

Prisma accepts `equals: { … }` and `equals: [ … ]` on a path. On Postgres `jsonComparesAsText`
sends `String(raw)`:

```
{ path: ["a"], equals: { b: 1 } }
  → ("metadata" #>> $1) = $2   bind: ["{\"a\"}", "[object Object]"]
```

Zero rows, no error. By the file's own standard — refuse rather than answer a question you cannot
answer — this wants either a refusal naming the limitation or the `#>` + `::jsonb` form
`array_contains` already uses.

### 4. Two unannounced breaking changes to exported types

Both confirmed with `vitest --typecheck`:

- `SoftDeleteOptions` is exported from `orm/index.ts:88`. Its parameter changed from the model type
  to a string union, so an existing `SoftDeleteOptions<User>` annotation now fails with *"Type
  'User' does not satisfy the constraint 'string'"*.
- `softDelete<T, M = any>` → `softDelete<T, M>`, same for `softDeleteMany`. `softDelete<Row>(model)`
  now fails with *"Expected 2 type arguments, but got 1."*

Restoring `M = any` costs nothing — `SchemaFields<any>` already degrades to `string` — and keeps the
one-explicit-argument call compiling. `SoftDeleteOptions` is harder to preserve, and at minimum
belongs in the release notes, since the PR body describes the migration as additive.

### 5. `soft-deletes.ts:25` still asserts the thing this PR disproves

```
 *     await User.delete({ where })   // sets deletedAt, does not remove the row
```

That is the exact claim #263 establishes is false, sitting 150 lines above the note that says so
(`:172`, `:195`). Two more in the same file:

- `:194` says "see the note **below** on #263" — the note is above it, `:146`–`:176`.
- `:200` still says `delete({ where })` where it now means `expire({ where })`.

### 6. One `softDeletes<Instance>()` spelling missed in the docs

`docs/orm.md:813`, and `docs/llms-full.txt:3544`, still read
`[softDeletes<Account>(), new TenantPolicy()]`. Every other occurrence moved to `typeof`; this one
is now the spelling the PR documents as checking nothing.

### 7. `dialect/index.ts` doc contradicts the code

The `jsonExtract` docstring says "the JSON form is what `equals` and `array_contains` need", but
`jsonComparison` calls `jsonExtract(column, path, true)` unconditionally — only `array_contains`
uses the JSON form.

### Smaller notes

- `assertPathShape` accepts `path: []` on Postgres — `[].every` is vacuously true — and `path: ""`
  on SQLite. The first extracts the whole document, the second raises inside SQLite.
- Adding five required members to the exported `SqlDialect` interface breaks any external
  implementor. Only the two in-tree dialects exist, so this is a release-note line rather than a
  change.
- The #263 block comment at `soft-deletes.ts:146`–`176` is an orphan JSDoc directly preceding
  `softDelete`'s own JSDoc; editors will surface only the second.
- `%` / `_` are unescaped in the new string filters, which matches `likePattern` and Prisma. No
  change wanted.
- `corpus.ts` not being extended for `path` is called out in the description with a defensible
  reason. Agreed for a release cut — worth an issue so it does not get lost.

nstfkc:
## Addressed in `453c4ad`

All seven, plus the smaller notes. Items 1–3 reproduced exactly as described
before fixing — thank you for the compiled SQL and binds, they made each one a
five-minute confirmation instead of a hunt.

### 1. Sentinels bypassing both refusals

Fixed, and the diagnosis was exact: the `path` branch sits between the
bare-sentinel refusal above and the `AnyNull` check in the key loop below, so it
reached neither. `jsonNullKind` / `carriesAnyNull` now run per operand at the top
of `compileJsonFilter`'s loop, raising `InvalidArgumentError` where the field and
operation are in scope — the placement you suggested.

**Refused rather than mapped**, which is the one place I went further than the
suggestion, and the reasoning is now in the code: the scalar path can answer
these because it compiles against the column, where SQL NULL and the JSON value
`null` are distinguishable. An extracted value has already lost that — `#>>`
yields NULL for an absent key and a JSON null alike — so mapping would mean
picking one meaning and being silently wrong about the other. That seemed worse
than refusing, given the whole point of the sentinels is to make a caller choose.

A test asserts the string `Prisma.DbNull` never reaches the SQL, since that was
the actual defect rather than the missing error.

### 2 and 3. Missing operand guards

Both fixed in one `assertJsonOperand`, which mirrors the scalar guard at
`where.ts:1020` including its reasoning.

For 3 I took the **refusal** rather than the `#>` + `::jsonb` form. Implementing
it properly is Postgres-only, so SQLite would get either a second wrong answer or
a silent divergence — the thing this file refuses everywhere else. The message
names the limitation, says Prisma accepts it, and points at `array_contains` or a
narrower path. `array_contains` stays exempt: containment is the one operator
whose right-hand side is a document.

### 4. The two breaking changes

Reproduced both, then fixed both. `M = any` is restored on `softDelete` /
`softDeleteMany`.

**`SoftDeleteOptions` turned out to be preservable, so it is preserved rather
than release-noted.** It takes the model type again and resolves the union
internally — a row type has no `$schema`, so it degrades to `string` exactly as
before, and `SoftDeleteOptions<typeof User>` is the spelling that gains the
check. So `SoftDeleteOptions<User>` compiles again and the PR body's "additive"
claim is now true rather than needing a caveat.

That leaves the **five new `SqlDialect` members** as the only breaking change.
Added to the PR description under *Breaking changes* — there is no changelog in
the repo, so that is the place; say the word if you would rather it go somewhere
else.

### 5–7. Stale comments

All fixed, and 5 was the worst of them — `soft-deletes.ts:25` promising
`User.delete({ where })` sets `deletedAt` is the exact claim this PR exists to
disprove, sitting above the note that disproves it. Also the "below" that pointed
up, the `delete({ where })` that means `expire({ where })`, the orphan JSDoc
(merged into `softDelete`'s own so editors surface it), the missed
`softDeletes<Account>()` in both docs files, and the `jsonExtract` docstring —
which now describes the single caller the JSON form actually has.

### Smaller notes

- **Empty paths** now refused on both dialects, with their own sentence rather
  than the grammar message — that would have told the caller to use the form
  they already used.
- **`%` / `_`** left unescaped, as agreed.
- **`corpus.ts`** filed as #301. It records the one concrete consequence I could
  find: on Postgres a path array's *length* reaches the plan key, so `["a"]` and
  `["a","b"]` mint separate entries holding identical SQL — the same shape
  `plan.ts` already handles for `in` via `LIST_KEYS`, and not currently in it.

### Verification

| | |
| --- | --- |
| `orm/compile` | **708** tests (695 before; +13 for the new refusals) |
| `packages/gemi` | 100 files / **2095** tests, CI's exclusion set; `tsc` clean |
| template, sqlite | 654 tests + 36 type tests |
| template, postgres | **928** tests, real server |

Lint unchanged at 83 warnings / 0 errors.

PR #304, opened 2026-08-02
Scalar lists on Postgres, and the second schema that gives them an oracle

### LINKED ISSUE #300: Scalar lists are refused at generation time, and the deferral points at an iteration that shipped
`tags String[]` cannot be used with the gemi ORM at all. The refusal is correct
and deliberate; what is missing is that nothing tracks it, nothing documents it,
and the comment recording the deferral points at an iteration that closed seven
iterations ago.

## What happens today, measured on both dialects

Added `tags String[]` to `User` in `templates/saas-starter/prisma/schema.prisma`
and ran `prisma generate`.

**Postgres** — the field is legal in Prisma, so gemi's own refusal fires and the
whole generation fails:

```
Error:
User.tags is a scalar list, which the gemi ORM does not support yet.
```

**SQLite** — Prisma refuses first and gemi never sees it:

```
error: Field "tags" in model "User" can't be a list.
The current connector does not support lists of primitive types.
Validation Error Count: 1
```

Note the failure is **generation-wide**, not per-field: `buildModelSchemas`
throws, so one `String[]` anywhere means no `schema.ts`, no `models.ts`, no
`index.ts` for any model.

## Why it is refused, which is worth keeping

`packages/gemi/bin/orm/emit.ts:257`:

```ts
// A list of scalars is Postgres-only and needs array decoding; iteration 2
// handles it. Skipping silently would change the result shape, so refuse.
if (field.isList) {
  throw new UnsupportedSchemaError(
    `${model.name}.${field.name} is a scalar list, which the gemi ORM ` +
      `does not support yet.`,
  );
}
```

The reasoning is the same one `UNSUPPORTED_SCALARS` gives for `Decimal` twenty
lines up, and it is right: skipping the column would generate cleanly and then
return a row shape that silently disagrees with the type Prisma handed the
caller. Failing loudly at generation is the correct call for an unsupported
column. Pinned by `emit.test.ts`, *"refuses a scalar list"*.

**The stale part is `iteration 2 handles it`.** Iterations 1-9 have all shipped
and none of them did. `grep -rn 'scalar list\|isList\|array decoding'
plans/orm/*.md` returns nothing — no iteration ever scoped this. So the deferral
lives in one code comment pointing at a closed iteration, which is the same shape
as #267 and #271: work recorded somewhere nobody re-reads.

### The type surface does not exist yet

`FieldSchema` (`packages/gemi/orm/schema.ts:53`) has `type`, `nullable`, `isId`,
`isUpdatedAt`, `enum`, `default` — and no list flag. `ScalarType` is nine scalars
with no list notion. So this starts by deciding whether a list is a *flag on the
field* or a *ScalarType constructor*, and that choice reaches `decode`,
`castParameter` and every `switch` over `ScalarType`.

### The query surface Prisma exposes

| | |
| --- | --- |
| filters | `equals`, `has`, `hasEvery`, `hasSome`, `isEmpty` |
| writes | `set`, `push` |

None exist: `grep -rn '"has"\|hasEvery\|hasSome\|isEmpty' packages/gemi/orm/`
returns nothing, and `OPERATORS` in `compile/where.ts:485` is still the twelve
scalar operators. These are a **different operator set on a different left-hand
side**, which is structurally the shape #267's JSON path filters just landed —
that is the closest precedent for how to add one.

### Decoding is the part with teeth

`dialect.decode(value, field)` (`dialect/index.ts:141`) is where a driver value
becomes a typed row value, and it raises `DecodeError` when it cannot. A Postgres
array has to round-trip *element-wise* — a `DateTime[]` inherits the zoneless
timestamp question `COMPOSITE_IN_TYPES` already declines to answer for a single
`DateTime`, and a `Bytes[]` or `Decimal[]` compounds a representation problem
that is open for the scalar. Worth deciding up front **which element types are in
scope**, rather than supporting `String[]`/`Int[]` and discovering the rest.

### Prior art already in the repo

Binding an array as one parameter is solved: `arrayLiteral` plus an explicit
`::type[]` cast, used by `inList`, `compositeIn` and now `jsonExtract`. And
`COMPOSITE_IN_TYPES` is the existing precedent for *a deliberately partial type
table that falls back rather than guessing* — the same discipline applies here.

## The hard part: the differential harness cannot reach this

Every correctness claim in this ORM is *the same query through Prisma and through
gemi against one database, compared on rows and table contents*. That is not
available here, and the reason is structural rather than an oversight.

`templates/saas-starter` has **one** `schema.prisma`, and
`app/models/postgres.sh` switches dialects by `sed`-flipping `provider` in it:

```sh
sed -i '' 's/provider = "sqlite"/provider = "postgresql"/' prisma/schema.prisma
```

So a scalar list added to that schema makes the **SQLite** suites fail to
generate at all — demonstrated above, `Validation Error Count: 1`. One schema
cannot carry a column that is legal on one dialect and a validation error on the
other.

Three ways out, and they are not equivalent:

1. **A dialect-specific schema or a generated fragment.** Buys a real differential
   oracle on Postgres, and is a change to the harness every other suite depends
   on. Most work, most value.
2. **Assert against Prisma's documented shape rather than a second generated
   client** — which is exactly what `relations.many-to-many.test.ts` already does
   for implicit m-n, and what `plans/orm/README.md:600` records as a known
   limitation. Cheapest, and weakest: it checks gemi against a reading of the
   docs, not against Prisma.
3. **Postgres-only test file, skipped without `TEST_POSTGRES_URL`.** Middle
   ground; the CI `postgres` job would cover it and the `sqlite` job would not.

**This decision should be made before the compiler work, not after** — it decides
what "done" can even be evidenced by, and #267's own commit message makes the
argument for why: implementing something the harness has no oracle for is how a
dialect gets an answer nothing can check.

## Docs debt that exists regardless of whether this is built

Independent of implementing it, three things are wrong today and worth fixing
even if this issue sits:

- `docs/orm.md`'s **Not in scope** lists the identity map, lazy loading,
  multi-field nested writes, migrations, `distinct` and `cursor` — **not scalar
  lists.** The one refusal that stops a build outright is the one not written
  down.
- `UnsupportedSchemaError` appears nowhere in `docs/`. It is generator-only and
  so correctly absent from the ORM errors table, which is a runtime table — but
  it is the error a reader actually hits, and nothing tells them what it means.
- The `iteration 2` comment should point here instead.

## What to decide

1. Build it, or record it as scope and close? Refusing is defensible — Prisma
   itself refuses on SQLite, so half the supported matrix could never have it.
2. If built: which element types, and which of the three harness options.
3. Either way, the docs debt above should land, because it is true now.

nstfkc:
Closes #300. Stacked on #299 — based on `feat/orm-release-issues` at `b5db21e`, whose docs commit this reconciles.

`tags String[]` could not be used with the ORM at all. The refusal was correct; what was missing is that nothing tracked it, nothing documented it, and the comment recording the deferral pointed at "iteration 2", which shipped seven iterations ago without doing it.

## The three decisions the issue asked for

**1. Build it.**

**2. The harness, decided before the compiler** — because it decides what "done" can be evidenced by. `templates/saas-starter` has one `schema.prisma` and `postgres.sh` covers both dialects by `sed`-flipping its provider. A scalar list is legal on Postgres and a *validation error* on SQLite, so a `String[]` added there does not merely go untested — it stops every SQLite suite from generating.

Of the issue's three options this takes the first: `prisma/postgres-only.prisma`, with its own client and its own database. That keeps the property every other correctness claim in this ORM rests on — the same query through Prisma and through gemi against one database, compared on rows *and* table contents. The cheaper options are weaker in a way worth naming: asserting against Prisma's documented shape checks gemi against a reading of the docs, and a hand-built table gives a real round trip with no oracle.

**3. Element types: everything Prisma allows on Postgres** — `String`, `Int`, `Float`, `Boolean`, `BigInt`, `DateTime`, `Bytes`, `Json` and enums. `Decimal[]` stays refused because `Decimal` is; a list of a scalar no dialect can round-trip is not more supportable for being a list.

## What the oracle caught

Three defects survived design and unit tests and died on contact with Prisma:

- **`create` refused calls Prisma answers.** A `String[]` with no `@default` is not "missing" — Prisma writes `[]`. No reading of the docs produces this: they describe the *input type*, where the field is optional, and say nothing about what is written when it is absent.
- **A bare array is a value but not a filter.** `{ tags: ["a"] }` compiled as `equals`; Prisma rejects it with *"Expected StringNullableListFilter"*. Accepting it made gemi a silent superset that returned plausible rows.
- **`$1::jsonb = any(col)` answers `false` where `$1::text::jsonb` answers `true`**, with no error on either. That is #209 one level in, and it is why the dialect's list methods take a `Fragment` rather than a `Binder` — `fieldParam` stays the single authority on cast-plus-serialisation, instead of each dialect restating a rule whose wrong version returns no rows.

## The refusal moved rather than disappeared

`emit.ts` was the wrong place for it and would have been even once arrays decoded: the artifact is dialect-agnostic on purpose, since `DATABASE_URL` can name a different database than `prisma generate` saw. Refusing there refused the column for Postgres too — and refused the *whole artifact*, since one `String[]` anywhere threw before any model was emitted. `SqliteDialect.listFilters` is empty now, and the compiler refuses from there naming the dialect and saying Prisma rejects the column on SQLite as well.

## Measured rather than reasoned about

Every operator resolves against an untyped array literal on `text[]`, `jsonb[]`, `bytea[]`, `timestamp(3)[]` and an `enum[]` — so no casts, no enum type name in the artifact, and `SCHEMA_ARTIFACT_VERSION` does not move.

The teeth were in decoding, where the driver returns three different containers:

```
text[] float8[] bool[] timestamp[] bytea[] jsonb[] bigint[]  -> Array
int[]                                                        -> Int32Array
enum[]  domain[]                                             -> "{a,b}"
```

`int[]`'s container depends on the *protocol*, so it cannot be decided once; an enum array arrives unparsed, which is why there is a literal parser at all. `BigInt[]` needed the array form of the `::text` cast the scalar case already documents — without it `9007199254740993` came back as `...992`, element-wise, through `json_agg`.

`isList` is a flag beside `type` rather than a `ScalarType` constructor, so `type` keeps meaning the *element* type and every existing `switch` over it stays exhaustive instead of quietly acquiring a shape it had never seen.

## Reconciling `b5db21e`

That commit documented this refusal as permanent — a **Columns the generator refuses** table listing scalar lists beside `Decimal`, and a *No scalar lists* entry under **Not in scope**. Implementing them makes both wrong, so both are fixed here rather than left to contradict the feature two sections away. The table keeps `Decimal`, loses the scalar-list row, and says the row *moved* rather than vanished. A section this branch had written independently is dropped and its one unique fact folded into `b5db21e`'s instead of duplicating it.

`453c4ad`'s changes to `compile/where.ts` and `dialect/index.ts` merged cleanly; the conflicts were `docs/orm.md` and its `llms-full.txt` embed, both by adjacency. `docs-links.test.ts` caught the one dangling anchor the reconciliation left behind.

## Verification

Both dialects, separately, as `postgres.sh` requires:

| | |
| --- | --- |
| `packages/gemi` | 61 files / 1674 ORM tests, `tsc` clean, `oxlint orm --deny-warnings` at 0 |
| template, sqlite | 652 tests + 36 type tests |
| template, postgres | 1027 tests on a real server (928 before — the +99 is the new suite) |
| template, CI's `-t postgres` | 783 tests, 0 failures |

Two things stated rather than left to be discovered. The six `app/`/`services/router/` failures in a full package run are #163 and predate this — confirmed identical at `2b087ec` in a scratch worktree rather than assumed. And `postgres.sh` unfiltered still fails the two SQLite-only differential suites, which is the documented one-dialect-at-a-time property, also identical at baseline; CI's filter excludes them.

**Known gap, stated rather than left to be found.** The differential cannot reach the literal parser's *escaping*: a Prisma enum member is `[A-Za-z][A-Za-z0-9_]*`, and Postgres quotes an array element only when it contains a delimiter, a quote or a backslash — so no schema Prisma accepts can produce a quoted element there. Those cases are unit tested against the parser directly in `orm/scalar-list.test.ts`, and the split is written down in both files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,6 +205,10 @@ jobs:
       # parameter, which reads as a connection failure rather than a bad URL.
       TEST_POSTGRES_URL: postgres://postgres:postgres@localhost:5432/gemi_test
       DATABASE_URL: postgres://postgres:postgres@localhost:5432/gemi_test
+      # The scalar-list schema's own database — see the two steps that create
+      # and push it. A second database rather than a second set of tables,
+      # because `prisma db push` reconciles a whole database to one schema.
+      TEST_POSTGRES_LISTS_URL: postgres://postgres:postgres@localhost:5432/gemi_lists
       TZ: UTC
 
     steps:
@@ -244,6 +248,32 @@ jobs:
       - name: Create the schema
         run: bunx prisma db push --skip-generate
         working-directory: templates/saas-starter
+
+      # `prisma/postgres-only.prisma` — the scalar lists (#300), which cannot
+      # live in the schema above because a `String[]` is a *validation error* on
+      # SQLite, and that schema has its provider flipped between dialects. So
+      # they get their own schema, their own client and their own database.
+      #
+      # A separate database because `prisma db push` reconciles a whole database
+      # to one schema: pushed at `gemi_test`, it would drop every table the main
+      # schema owns and the rest of this job would fail with missing relations.
+      #
+      # Skipping this step does not fail anything — `scalar-list.test.ts` gates
+      # on `TEST_POSTGRES_LISTS_URL` — so it would read as a pass. That is
+      # exactly the silence `postgres-suite-selection.test.ts` exists to catch,
+      # and the suite prints a loud warning when it does not run.
+      - name: Create the scalar-list database
+        run: >-
+          PGPASSWORD=postgres psql -h localhost -U postgres -d gemi_test
+          -c 'CREATE DATABASE gemi_lists'
+
+      - name: Generate and push the scalar-list schema
+        run: |
+          bunx prisma generate --schema prisma/postgres-only.prisma
+          bunx prisma db push --schema prisma/postgres-only.prisma --skip-generate
+        working-directory: templates/saas-starter
+        env:
+          LISTS_DATABASE_URL: postgres://postgres:postgres@localhost:5432/gemi_lists
 
       # `-t postgres` selects the dialect-labelled suites, and it is required
       # rather than tidy. `@prisma/client` speaks one provider at a time, and
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -262,10 +262,14 @@ jobs:
       # on `TEST_POSTGRES_LISTS_URL` — so it would read as a pass. That is
       # exactly the silence `postgres-suite-selection.test.ts` exists to catch,
       # and the suite prints a loud warning when it does not run.
+      # `|| true` for the same reason `app/models/postgres.sh` has one: a
+      # re-created database is an error rather than a no-op, so a rerun against
+      # a persisted service container would fail a step that is not the point of
+      # the job. A fresh container makes it succeed the first time either way.
       - name: Create the scalar-list database
         run: >-
           PGPASSWORD=postgres psql -h localhost -U postgres -d gemi_test
-          -c 'CREATE DATABASE gemi_lists'
+          -c 'CREATE DATABASE gemi_lists' || true
 
       - name: Generate and push the scalar-list schema
         run: |
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3113,6 +3113,7 @@ merely unsupported:
 
 An empty path — `[]` or `""` — is refused on both dialects. On Postgres it would extract the whole
 document, which is a filter on the column rather than on a path.
+
 ### Scalar lists (Postgres only)
 
 ```ts
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2789,12 +2789,11 @@ Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 ### Columns the generator refuses
 
-Two kinds of column stop generation with an `UnsupportedSchemaError` naming the model and field,
+One kind of column stops generation with an `UnsupportedSchemaError` naming the model and field,
 rather than being skipped:
 
 | Column | Why |
 | --- | --- |
-| A scalar list — `tags String[]` | Postgres-only, and needs array decoding this ORM does not have. |
 | `Decimal` | Prisma types it as `Prisma.Decimal`; SQLite stores a REAL and the driver returns a JS number, so the value would be a float pretending to be an arbitrary-precision decimal. |
 
 **Refused rather than omitted, and the whole generation fails rather than the one field.** Skipping
@@ -2806,7 +2805,25 @@ visible at the moment you introduce it.
 A `Unsupported(...)` column is different and is *not* refused: Prisma's own client omits those from
 its result types, so omitting them is what keeps the shapes identical.
 
-Scalar lists are tracked in [#300](https://github.com/nstfkc/gemi/issues/300).
+```
+UnsupportedSchemaError: User.price is a Decimal, which the gemi ORM does not support yet:
+Prisma types a Decimal field as `Prisma.Decimal`, but SQLite stores it as a REAL and the
+driver returns a JS number — so the value would be a float pretending to be an
+arbitrary-precision decimal, losing exactly the precision the type exists to keep.
+Keep using the Prisma client for this model, or change the field's type.
+```
+
+A scalar type the generator cannot map at all — a future Prisma addition — raises the same error.
+Both are *generator* errors rather than runtime ones, so neither is in the
+[errors table](#errors) below, which is about queries. If you see one, the fix is in
+`schema.prisma`.
+
+**A scalar list used to be in this table and no longer is.** `tags String[]` is implemented on
+Postgres — see [Scalar lists](#scalar-lists-postgres-only). The refusal did not disappear so much as
+move: this artifact is dialect-agnostic on purpose, because `DATABASE_URL` can name a different
+database than `prisma generate` saw, so refusing here refused the column for Postgres too. SQLite
+now declines it at query time instead, with a message naming the dialect. `Decimal[]` is still
+refused, by the row above rather than by a rule about lists.
 
 ### Your model class
 
@@ -3096,6 +3113,59 @@ merely unsupported:
 
 An empty path — `[]` or `""` — is refused on both dialects. On Postgres it would extract the whole
 document, which is a filter on the column rather than on a path.
+### Scalar lists (Postgres only)
+
+```ts
+// model Post { tags String[] }
+await Post.findMany({ where: { tags: { has: "urgent" } } })
+await Post.findMany({ where: { tags: { hasEvery: ["urgent", "draft"] } } })
+await Post.update({ where: { id }, data: { tags: { push: "urgent" } } })
+```
+
+**Postgres only, and that is Prisma's line rather than this ORM's.** SQLite has no array type, and
+`prisma generate` refuses the column outright there — *"Field `tags` in model `Post` can't be a
+list. The current connector does not support lists of primitive types."* So there is no SQLite
+behaviour to match.
+
+The filters, which are a **different set from the scalar operators** on a different kind of column:
+
+| | |
+| --- | --- |
+| `has` | the list contains this element |
+| `hasEvery` | the list contains all of these |
+| `hasSome` | the list and this one share an element |
+| `isEmpty` | `true` or `false` |
+| `equals` | the whole list, in order |
+
+`contains`, `startsWith`, `in` and the comparisons are **not** among them: `contains` on a `String`
+asks about a substring, and on a `String[]` there is no such question — the one Prisma spells is
+`has`. Writing one gets an error naming both sets.
+
+One asymmetry worth knowing before it surprises you, because it is Prisma's and gemi reproduces it:
+
+```ts
+await Post.findMany({ where: { tags: ["a", "b"] } })      // ✗ refused
+await Post.findMany({ where: { tags: { equals: ["a", "b"] } } })  // ✓
+await Post.update({ where: { id }, data: { tags: ["a", "b"] } })  // ✓ a bare array is a value
+```
+
+A bare array is a **value** but not a **filter**. Prisma rejects the first line with *"Expected
+StringNullableListFilter"*, so gemi does too rather than guessing that `equals` was meant.
+
+Writes take `set` and `push`; `push` accepts one element or a list. `increment` and its siblings
+apply to a number, not to a list of them, and are refused by name.
+
+**An omitted list is written as the empty list, not refused.** A `String[]` with no `@default` is
+still not "missing" on `create` — Prisma writes `[]`, and so does gemi. That was measured against a
+generated client rather than read off the input type, which says only that the field is optional.
+
+Every element type Prisma allows is supported — `String`, `Int`, `Float`, `Boolean`, `BigInt`,
+`DateTime`, `Bytes`, `Json` and enums. `Decimal[]` is refused, because `Decimal` itself is (see
+below); a list of a scalar no dialect can round-trip is not more supportable for being a list.
+
+An enum list is worth one note: the driver hands those back as an unparsed Postgres array literal
+rather than as an array, so gemi parses them. You will not see the difference — that is the point —
+but it is why a list column costs a decode where a `String` does not.
 
 ### What a refusal tells you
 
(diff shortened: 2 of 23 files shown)


nstfkc:
## Review

Checked the branch out, ran the suites — 61 files / 1674 tests green, matching the description — and
probed the write paths and the plan cache directly. Findings below are measured rather than read.

**What holds up under checking.** The design calls are right and well-evidenced. `isList` as a flag
beside `type` keeps every `switch (field.type)` exhaustive, which is the choice the whole diff rests
on. `castParameter` testing `isList` *before* `Json` is the correct order and is pinned by a test.
The four dialect list methods taking a `Fragment` rather than a `Binder` is the right fix for the
`$1::jsonb` vs `$1::text::jsonb` silent-false trap, and `array_cat` over `||` is correct for exactly
the reason given. I traced `parseArrayLiteral` by hand over quoted delimiters, escaped quotes and
backslashes, embedded braces, `NULL` vs `"NULL"`, empty elements and nested literals — it is correct
and it terminates on malformed input. And the plan cache **does** distinguish `data: { tags: [...] }`
from `{ set: [...] }` from `{ push: … }`, so deciding `setForm` at compile time is safe.

Items 1–3 are all in the write path, and all in the seam between `create` and `createMany` — which
is where the new differential coverage is thinnest.

### 1. `createMany` writes `[]` over a list's declared default, and disagrees with `create`

`createManyColumns` picks the empty-list binder ahead of the default binder:

```js
value: field.isList ? emptyList(field, dialect) : defaultBinder(field, dialect)
```

So for `defaulted String[] @default(["seed"])`, measured on the branch:

```
create     ("id","defaulted") values ($1,$2)          bind: [1, '{"seed"}']         ✓
createMany ("id","defaulted") values ($1,$2),($3,$4)  bind: [1, '{}', 2, '{"x"}']   ✗
```

Row 1 omitted the field in both cases. `create` writes the default, `createMany` writes `[]`; Prisma
writes the default in both. The `insertColumns` ordering is right — `if (field.default ||
field.nullable) continue;` sits above the `isList` branch — and `createManyColumns` simply does not
mirror it. `suppliesEmptyList` already keeps defaulted lists out of the *union*, so the fix is the
same test at the per-row site.

The differential cannot see this today, and it is one line from being able to:
`postgres-only.prisma` has exactly the right column, but the `createMany` case only supplies
`strings`, so `defaulted` never enters the union. One `createMany` row supplying it beside one
omitting it would have caught this.

### 2. `insertColumns` skips the dialect check on the path that invents the value

`assertListDialect` is called only inside `if (supplied.has(field.name))`. The new "a scalar list is
never missing" branch below it constructs a value without one, so on SQLite:

```
create({ data: { id: 1 } })
  → insert into "Tagged" ("id","tags","defaulted") values (?,?,?)   bind: [1, [], []]
```

It compiles, and binds raw JS arrays for `bun:sqlite` to reject with a driver-level type error —
instead of the carefully written *"sqlite has no array type … It works on postgres"* that the
supplied path produces (verified: `UnsupportedQueryError` there). `createManyColumns` gets this right
with its `for (const field of union)` loop; `insertColumns` wants the same check before
`columns.push({ field, value: emptyList(...) })`.

### 3. `createMany` has no `assertListOperand`, so a bad operand escapes to bind time

`assertListOperand`'s own docstring makes the case: *"Compile is a pure function of the argument
shape and this is a shape question, so it belongs here."* `insertColumns` and `assignment` both call
it; `createManyColumns` does not. Measured:

```
createMany({ data: [{ id: 1, tags: { push: ["a"] } }] })
  → compiles, then at bind:
    "…so it takes an array — or { set: [...] } / { push: … }. Received object."
```

The refusal is late, and the message offers `{ push: … }` as the remedy on the one operation where
`push` is never legal — which is the "next step vs puzzle" distinction the `create` test asserts with
`/scalar list, so it takes set —/`.

### 4. Every list operand mints one plan-cache entry per list length

`canonicalShape` records arrays element-wise and `collapseLists` fires only for the `in` / `notIn`
keys, so:

```
hasEvery ["a"]          → …{tags:{hasEvery:[string]}}
hasEvery ["a","b","c"]  → …{tags:{hasEvery:[string,string,string]}}
compiled SQL text:      identical
```

That is the churn `collapsedList` exists to prevent — *"it mints one entry per distinct list length,
each holding SQL identical to its neighbours'"* — reintroduced for `hasEvery`, `hasSome`, list
`equals`, list `set` / `push`, and bare-array `data` values. This PR adds a test asserting the text
does not vary with length; the key should follow it. The batched-relation urgency does not apply
here, but a caller passing a user-sized `hasEvery` will churn the LRU. `LIST_KEYS` cannot simply
absorb the names — `equals` and `set` are scalar operators elsewhere — so this may want a
schema-aware collapse, or an explicit note that it is accepted.

### Smaller notes

- `prisma/postgres-only.prisma:22` says the suite is *"gated on `TEST_POSTGRES_URL`"*. It is gated on
  `TEST_POSTGRES_LISTS_URL` — and `postgres-suite-selection.test.ts` was widened in this same PR
  precisely because those are two different names.
- `createDifferential`'s `client` docstring says it is *"required to come with `url` and `tables`"*,
  but only `url` is enforced. Omitting `tables` silently truncates the main schema's table list
  against the lists database.
- `docs/orm.md:385` — `### Scalar lists (Postgres only)` has no blank line before it. CommonMark
  still renders it; every other heading in the file has one.
- The generator's new `client` option is type-checked but not checked for emptiness; `client = ""`
  emits `import type { Prisma } from ""`. One `!client.trim()` in the same guard.
- CI's `CREATE DATABASE gemi_lists` step has no idempotence guard, unlike `postgres.sh`'s `|| true`.
  Fine on a fresh service container; a rerun against a persisted one fails a step that is not the
  point of the job.
- `assertKeyable` refusing list join keys, and `elementOf` dropping the flag for `has`, are both
  exactly right — worth saying, since they are the two places this could have gone quietly wrong.

nstfkc:
All four findings fixed in `832646c`. Each was reproduced first, and each fix mutation-checked by reintroducing the defect and confirming a test fails — findings 1–3 had no coverage until now, and finding 3's mutation survived my first attempt at a test, which is why there are two.

### 1 — `createMany` writing `[]` over a declared default

Correct, and worse than the report: the same `field.isList` check appears twice, and the second one was wrong in the other direction. The validation loop's unconditional `if (field.isList) continue;` also skipped the *"some rows set it and others leave it to the database default"* refusal, so a list with a **database-side** default would have bound NULL into a NOT NULL column. Both sites now read `suppliesEmptyList`, which already encodes the distinction the code needed — "does this list have anything else to fall back on" rather than "is this a list".

You were right that the differential was one line from seeing it, and it needed two changes rather than one: the fixture's `@default([])` made the correct value and the wrong one the same array, so the default is now `["seed"]`, and three `createMany` cases cover the seam. With the fix reverted they fail against Prisma:

```
× createMany mixing the bare and { set } spellings
× createMany honours a list default on the rows that omit it
× createMany with every list omitted on every row
```

### 2 and 3 — the two missing checks

Both as described. On 2, the point that made it click: the dialect check belongs to the *column*, not to the caller having mentioned it. On 3, `["set"]` rather than the full operator list, so the refusal keeps naming what `create` actually takes.

### 4 — the plan key, where checking `set` was the whole exercise

`hasEvery`, `hasSome` and `push` join `LIST_KEYS`. **`equals` and `set` do not, and `set` would have been a bug rather than a missed optimisation** — it is also a *relation* operator, and `data: { tags: { set: [{ id: 1 }, { id: 2 }] } }` rewrites a join table with a statement whose text grows with the list. Collapsing by the name alone would have handed that plan the wrong number of placeholders, which is exactly what the note above `collapsedList` warns about for `OR`. There is now a test asserting `set` keeps its length, so the omission cannot later read as an oversight.

One case is left uncollapsed and recorded rather than guessed at: a bare array as a *write* value is keyed by the field name, so recognising it needs to know the field is a list, and `canonicalShape` is deliberately schema-free. The cost is bounded by the LRU and the SQL is identical either way, where a wrong collapse is a wrong statement. The filter side — where a user-sized list actually arrives — is covered.

### Smaller notes

All five landed. On `createDifferential`, I enforced `tables` rather than softening the docstring: the default list names the *main* schema's tables, so omitting it would `TRUNCATE` tables absent from the database it was pointed at — a quiet failure, not a loud one.

One incidental: a comment I wrote contained the phrase *"Two operations"*, and `operation-count.test.ts` read it as a claim about how many operations the ORM implements. It was right to fire; the comment is reworded.

### Verification

| | |
| --- | --- |
| `packages/gemi` | 61 files / 1689 ORM tests, `tsc` clean, `oxlint orm --deny-warnings` at 0 |
| template, sqlite | 652 + 36 type tests |
| template, postgres | 1029 on a real server |
| scalar-list differential | 101 |

PR #306, opened 2026-08-03
Export AuthManager, fix gemi migrate, and cut 0.51.0-rc.2

nstfkc:
Found while checking a 0.51.0-rc.1 migration report against the source. Each fix was mutation-checked by reverting it and confirming a test fails.

## The documented auth seam did not exist

`docs/authentication.md` has described the only supported way to install a custom `UserProvider` since 922c45d — subclass it, rebind `AuthManager` in a service provider — and neither import in that recipe resolved.

`AuthManager` was never exported. The `// Auth` section of `services/index.ts` stopped at `AuthServiceProvider`, while every other manager in the file sits beside its provider. `ServiceProvider` had moved to `gemi/support` — a move the codemod's own `MODULE_MOVES` table already knew about.

`AuthConfig` has no provider field, so rebinding `AuthManager` is the *whole* of the seam. An app had a documented path, a container that would have accepted the binding, and no way to name the token — which reads as a missing capability rather than a missing line. It shipped in both 0.51 release candidates.

## `gemi migrate` was destructive on an already-migrated app

The command advertises itself as a no-op without `app/kernel/providers/`, and was not one. Steps 2 and 3 ran unconditionally, so on a current app it rewrote `Kernel.ts` and carried the existing `config` member over commented out — `buildKernel` writes that member itself, so anything already there looked like a leftover.

`config` is part of the current Kernel shape. Commenting it out leaves a file that compiles and an app that boots with every config slice silently unmerged. No error, no crash, just defaults everywhere.

Also fixed: `memberRenames: { adapter: "userProvider" }` rewrote the 0.42 member into a key `AuthConfig` no longer has, so the codemod's own output did not type-check. And a new `RETIRED_CONFIG_FIELDS` pass annotates `userProvider` in `app/config/auth.ts` — the case that applies to apps which migrated when 0.43 was current, and the one the command previously did least for.

## Tests

- `docs-imports.test.ts` — checks every `gemi/*` import in `docs/` against what the entrypoint actually exports. `tsc` does not read fenced code, so the one artifact stating the intended design had no check on it. It immediately found a second live bug in `docs/file-storage.md`.
- `migrate.test.ts` — the codemod had no tests. The Kernel bug is the kind tests catch and review does not: the broken output still parses.

## Release

Bumps to `0.51.0-rc.2`. Verified `bun run build:publish` stages cleanly, and the published `dist/services/index.d.ts` carries the new export.

Full suite green on this base: 2203 passed, 1 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,13 +87,13 @@ export default function SignIn() {
 
 ### Type-safe network layer
 
-Consuming api endpoints using `useGet` hook.
+Consuming api endpoints using the `useQuery` hook.
 
 ``` tsx
-import { useGet } from 'gemi/client'
+import { useQuery } from 'gemi/client'
 
 export default function OrderDetail() {
-  const { data: orders } = useGet('/orders/:orderId')
+  const { data: orders } = useQuery('/orders/:orderId')
 
   return (
     <div>...</div>
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -81,10 +81,19 @@ export default defineRouteConfig({
 and `route.view.rootRouter` have no defaults. Everything else can be omitted
 entirely.
 
-### One property was renamed
+### One property was retired
 
-`AuthenticationServiceProvider.adapter` is now `auth.userProvider`. The codemod
-renames it for you and says so in the summary.
+`AuthenticationServiceProvider.adapter` briefly became `auth.userProvider`, and
+then the seam it selected between was removed altogether: auth persistence is
+now the ORM-backed `UserProvider`, and `AuthConfig` has no field for it.
+
+The codemod comments the member out and leaves a TODO carrying the replacement —
+subclass `UserProvider` from `gemi/kernel`, override the methods the adapter
+implemented, and install it by rebinding `AuthManager` (from `gemi/services`) in
+a ServiceProvider, which takes the provider as its second constructor argument.
+The same TODO is written over a `userProvider` field left in an
+`app/config/auth.ts` by an earlier migration. See
+[docs/authentication.md](docs/authentication.md) for the worked example.
 
 ---
 
--- a/bun.lock
+++ b/bun.lock
@@ -30,7 +30,7 @@
     },
     "packages/gemi": {
       "name": "gemi",
-      "version": "0.51.0-rc.1",
+      "version": "0.51.0-rc.2",
       "bin": {
         "gemi": "./dist/bin/gemi.js",
       },
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -189,9 +189,11 @@ argument:
 ```typescript
 // app/providers/AppServiceProvider.ts
 import { AuthManager } from "gemi/services";
-import { ServiceProvider } from "gemi/services";
+import { ServiceProvider } from "gemi/support";
 
-export class AppServiceProvider extends ServiceProvider {
+import { OrgProvisioningUserProvider } from "@/app/auth/OrgProvisioningUserProvider";
+
+export default class AppServiceProvider extends ServiceProvider {
   register() {
     this.app.singleton(
       AuthManager,
@@ -204,6 +206,10 @@ export class AppServiceProvider extends ServiceProvider {
   }
 }
 ```
+
+The provider has to be listed in your `Kernel`'s `providers` array to run — see
+[Project Structure](./project-structure.md#service-providers). App providers register *after*
+the framework's, so this binding replaces the default `AuthManager` rather than racing it.
 
 > **Note:** gemi runs `createUser` and then fires `onSignUp` *separately*. Provisioning
 > inside the `onSignUp` callback is therefore **not** atomic with user creation — a failure
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -61,7 +61,7 @@ It launches `dist/server/server.mjs` in a fresh Bun process with `NODE_ENV=produ
 
 ## `gemi migrate`
 
-Upgrades an app from the 0.42 service-provider layout to the 0.43 config + container layout.
+Upgrades an app from the 0.42 service-provider layout to the config + container layout introduced in 0.43, and flags the APIs retired since.
 
 ```bash
 gemi migrate --dry-run   # print the plan, write nothing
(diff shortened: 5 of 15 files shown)


nstfkc (commented):
## Verification

| Check | Result |
|---|---|
| Full package suite on the branch | **2203 passed, 1 skipped, 106 files** — matches the PR claim |
| `tsc --noEmit` | clean |
| `oxlint` | 83 warnings / 0 errors — identical to base, the new files add none |
| Mutation: un-gate step 2 (`if (true)`) | `its Kernel is left exactly as it was` **fails** |
| Mutation: drop the `AuthManager` export | `every named import exists on the entrypoint it comes from` **fails** (2 missing) |

Both new tests are load-bearing, not decoration.

I also hand-checked the claims the diff makes about framework behaviour:

- `kernel/Kernel.ts:43` — `registerMany([...frameworkProviders, ...this.providers])`, and `Application` runs every `register()` synchronously in order before any `boot()`. The new docs sentence ("App providers register *after* the framework's") is accurate.
- Stronger than the docs say, in fact: `Container.singleton()` also calls `instances.delete(token)`, and `Facade.getFacadeRoot()` resolves per call rather than capturing — so the rebind wins even if something resolved `AuthManager` first.
- The docs snippet is now a verbatim match for `AuthServiceProvider.register()`, and `gemi/support` does export `ServiceProvider`.
- Ran the codemod against synthetic 0.42, already-migrated, and re-run fixtures: the generated `app/config/auth.ts` type-checks (the dead `adapter`/`prisma` imports are correctly dropped), the annotation is idempotent across runs, and the already-migrated Kernel is untouched.

No new export weight from the `AuthManager` line — `AuthServiceProvider` already pulled it into the same module graph.

## Findings

### 1. `UPGRADE.md` now contradicts the codemod, and this PR is what made it wrong — *medium*

`UPGRADE.md:85-87`:

> ### One property was renamed
> `AuthenticationServiceProvider.adapter` is now `auth.userProvider`. The codemod renames it for you and says so in the summary.

That is exactly the rename this PR deletes. Worse, the codemod's report footer prints `See UPGRADE.md for the 0.42 -> 0.43 notes.` — so a user reads a TODO saying `userProvider` no longer exists, follows the pointer, and lands on a page saying it is the destination. And `tables.ts:1-3` opens by claiming data tables are what keep "`UPGRADE.md` and the codemod ... from drifting apart".

The heading is stale too: nothing is renamed anymore. It wants to say the member is retired, carrying the same sentence `ADAPTER_RETIRED` does.

### 2. `annotateRetiredFields` stops at the first match, so a nested key can swallow the annotation — *low-medium*

`bin/migrate/index.ts:355-367` runs `property.exec(out)` once per field. The doc comment budgets for the false positive ("a nested `userProvider` inside some unrelated option would be a false positive the author can see and ignore") but not the false *negative* that follows when the nested one comes first:

```ts
export default defineAuthConfig({
  social: {
    // TODO(gemi-migrate): The authentication adapter seam was removed; …
    userProvider: "unrelated",
  },
  userProvider: new OrgAdapter(prisma),   // ← the real retired field, unmarked
});
```

Add `g` and loop over every match — each is still individually guarded by the existing `endsWith(message)` idempotency check. Then the tolerated false positive stays tolerable and the real field is never missed. `AuthConfig` has no nested `userProvider` today, so this is forward-looking rather than live, but the loop is cheaper than the reasoning about why it is safe.

### 3. A half-migrated app's hand-written config is silently clobbered, and the report says it is not — *medium, pre-existing but adjacent*

An app with **both** `app/kernel/providers/` and a hand-written `app/config/auth.ts`:

```
  update  app/config/auth.ts
  …
Migration complete. Nothing needs manual attention.
```

`handcrafted: true` and the app's own `userProvider` were both gone, replaced by the provider-derived file. Two things make this adjacent rather than a drive-by: this PR takes "what does this command do to an app that is not cleanly on 0.42" as its subject, and the surrounding code *already has the guard* — `index.ts:130-136` refuses to overwrite an existing extracted file and notes "already exists — left untouched, review it by hand". The config file itself is one `existsSync` + note away from the same treatment.

### 4. `ADAPTER_RETIRED` prints twice, verbatim, for one file — *low*

For a provider with an `adapter` member importing `PrismaAuthenticationAdapter`, the summary emits the same ~330-character paragraph twice — once from `DELETED_EXPORTS` (unprefixed) and once from `memberRemovals`:

```
  app/kernel/providers/AuthenticationServiceProvider.ts
    - The authentication adapter seam was removed; auth persistence is now the ORM-backed…
    - `adapter` — The authentication adapter seam was removed; auth persistence is now the ORM-backed…
```

The report is this command's entire output surface, and reusing one sentence across two tables is what causes it. Dedup on `(file, message)` in `report()` covers it generically.

### 5. `exportsOf` will `ENOENT` instead of failing cleanly — *low*

`docs-imports.test.ts:105` does an unguarded `readFileSync(join(ROOT, EXPORTS[entry]))`. Not hypothetical: `"./runtime": "./client/runtime.ts"` points at a **file that does not exist** — `scripts/build-publish.ts:20` already documents it as a legacy dead export. Nothing in `docs/` imports `gemi/runtime` today, so the test passes; the first snippet that does gets a stack trace instead of a useful assertion. An `existsSync` skip closes it — or better, drop the dead `./runtime` entry from `exports`.

### 6. Minor

- **`toBeGreaterThan(50)` is very loose** (`docs-imports.test.ts:132`). The docs currently produce **496** named imports across 14 entrypoints. If the regex silently degraded to matching 15% of statements the guard still passes. `> 400`, or asserting the set of modules seen, would be a real canary.
- **`UPGRADE.md` and `README.md` carry `gemi/*` imports the new test does not see.** `README.md` is straightforwardly checkable; `UPGRADE.md` deliberately shows pre-0.43 code (`import { Singleton } from "gemi/services"` at :209), so it would need a before/after fence convention first. Follow-up, not this PR.
- `index.ts:182` runs to 90 chars against the file's otherwise-wrapped style, and passes both `file` and its `rel()` label; `annotateRetiredFields` could derive the label itself.

## Verdict

The two substantive fixes are real, correctly diagnosed, and now covered by tests I confirmed fail without them. Nothing here blocks merge on correctness. I would fix **#1** before cutting rc.2 — the release ships a codemod pointing users at a document that contradicts it, and it is a two-sentence edit. **#2** is a cheap loop worth taking while the file is open; **#3** and the rest are fine as follow-ups.

PR #310, opened 2026-08-03
Remove the dead gemi/runtime export

nstfkc:
Follow-up to the review on #306, which flagged `exportsOf` throwing `ENOENT` on this entry. Fixing the entry rather than guarding against it.

## The entry has resolved to nothing since 2024

`"./runtime": "./client/runtime.ts"` outlived its target by over a year — `client/runtime.ts` was deleted in June 2024 (`18057af`, "vite doesn't build client") and the entry stayed.

Dead in every direction, each checked rather than assumed:

- **No source.** `client/runtime.ts` does not exist. The similarly-named `runtime/index.ts` is a different file: not a build entrypoint, imported by nothing, and it sets a `window.views` that nothing reads.
- **No built target either.** `bun run build` never emits `dist/runtime/index.js` — only `tsc` visits the directory, producing a declaration for a module with no exports. Verified against the real artifact: `npm pack gemi@0.51.0-rc.1` ships `dist/runtime/index.d.ts` and `.d.ts.map` with no `.js` beside them. Types for a module that cannot be loaded.
- **No consumer.** Nothing in the framework, templates or docs imports `gemi/runtime`.

**This cannot break a working consumer, because there is no way to have been one.** That is the whole safety argument, and it rests on the published tarball rather than on the repo.

## What kept it alive

`build-publish.ts` held an override mapping `./runtime` to a built path, added to avoid "changing the published surface". That path was as absent as the source — so the override did not preserve a working export, it made a broken one look deliberate, and put a comment where a check belonged.

The override table is removed along with its only entry. Also gone: `runtime/index.ts` and its `tsconfig` include, which only put a stray `.d.ts` in the tarball, and `gemi/runtime` from `GEMI_EXTERNAL_SPECIFIERS` — a list documented as tracking the view-facing `exports`.

## The check

`exports-map.test.ts` asserts every source target exists, and that every non-`dist` target is a `.ts`/`.tsx` file the publish rewrite can actually compile. The second is this bug's specific shape: an entry whose published form was never built. Confirmed to fail when the dangling entry is restored.

## Verification

| Check | Result |
|---|---|
| Full suite | 2197 passed, 1 skipped, 105 files |
| `tsc --noEmit` | clean |
| `oxlint` | 83 warnings / 0 errors — unchanged |
| `bun run build:publish` | stages 19 exports; **all 19 targets resolve** inside `.publish/` |
| `dist/runtime/` after build | no longer produced |

## Interaction with #306

Independent — based on `origin/next`, no shared files, mergeable in either order. Once both land, the `existsSync` guard in #306's `docs-imports.test.ts` becomes belt-and-braces: this PR removes the only entry that could have tripped it, and `exports-map.test.ts` now enforces that invariant directly.


nstfkc (commented):
Follow-up to finding #5 on #306. It takes the better of the two options offered there — fix the entry rather than guard against it — and the archaeology holds up.

## Verification

| Check | Result |
|---|---|
| Full suite on the branch | **2197 passed, 1 skipped, 105 files** — matches the claim |
| `tsc --noEmit` | clean |
| `oxlint` | 83 warnings / 0 errors — unchanged |
| Mutation: restore `"./runtime": "./client/runtime.ts"` | `every source target exists` **fails** |

**The central claim is confirmed against a real artifact, not just the source.** The `.publish/package.json` already staged in my working tree (rc.1, pre-PR) has 20 exports; checking every target against `.publish/dist/`:

```
MISSING ./runtime -> ./dist/runtime/index.js
total 20 missing 1
```

Exactly one dangling entry, exactly the one this PR deletes. Everything else resolves. `dist/runtime/` holds `index.d.ts` + `.d.ts.map` and no `.js`, as described.

Also checked independently:

- **No consumer.** Only two `gemi/runtime` references exist in `packages/gemi`, both removed here. `window.views` / `window.manifest` appear *only* in the deleted file. Nothing imports `runtime/index`. Template views import `gemi/client` and `gemi/email`, nothing else.
- **The safety argument holds under classic resolution too** — the one way it could have failed. `files: ["dist/**/*"]` means the tarball has no package-root `runtime/`, so a `moduleResolution: node10` consumer bypassing `exports` never found it either.
- **`GEMI_EXTERNAL_SPECIFIERS`** is consumed only by `httpDev.ts`'s exact-match `ssr.external`, and its header documents it as tracking view-facing `exports` — so removing the entry restores that stated invariant rather than weakening anything.

## Findings

### 1. The test does not check the half of the invariant the PR says it does — and that half is the one that caught this bug — *medium*

Both assertions in `exports-map.test.ts` start with `.filter(([, target]) => !target.startsWith("./dist/"))`. But "an entry whose published form was never built" is `./dist/…`-shaped **by construction** — that was `./runtime`'s override target. The shape is excluded from both checks.

Demonstrated: restoring the exact dangling entry leaves `every published target is a built path` **green**. Only `every source target exists` fires — 1 failed, 2 passed. So the docblock's "the second is the shape of this bug specifically, an entry whose published form was never built" (and the matching sentence in the PR body) describes an assertion that does not do that. What it actually asserts is that the *source* target has a `.ts` extension.

This matters beyond wording, because the gap is live. `./client -> ./dist/client/index.js` and `./vite -> ./dist/vite/index.mjs` are produced by two **separate** vite builds that `scripts/build.ts`'s entrypoint list does not cover. Drop `client/index.ts` from `vite.client.config.mts`'s input and the new test stays green while the tarball ships a dangling export again — this bug, recurring, undetected.

The check that does work is the one I ran above, and it is three lines at the end of `build-publish.ts`, after staging:

```ts
const dangling = Object.entries(publishPkg.exports)
  .filter(([, v]) => typeof v === "string" && !existsSync(join(STAGING, v as string)));
if (dangling.length) throw new Error(`Published exports with no built target: ${...}`);
```

That is precisely the manual verification the PR body reports doing ("stages 19 exports; all 19 targets resolve"). It runs on every release, needs no test infrastructure, and lands in the file whose new comment says an exception "should carry a test rather than a comment".

### 2. `GEMI_EXTERNAL_SPECIFIERS` has an invariant held up only by a comment — *low*

Its header says "Keep in sync with the view-facing `exports`", and names the intentional omissions: `vite`, `server`, `bun/*`, `config`. That accounts for 14 of the 19 subpaths. Five are absent and unexplained: **`orm`, `foundation`, `container`, `support`, `database`**.

The documented failure mode is silent ("or dev SSR will silently bundle a copy"), and `gemi/orm` carries a module-level `registry` — a duplicated copy is exactly the class of bug the comment describes. Nothing is broken today: template views import only `gemi/client` and `gemi/email`. But this PR is editing that list to restore the sync, and `exports-map.test.ts` is the natural home for `exports − documented omissions ⊆ GEMI_EXTERNAL_SPECIFIERS`, which would make the omissions a deliberate list instead of an absence.

### 3. Nits

- **The list's first entry, `"gemi"`, names a specifier the map has no `"."` key for.** Harmless as an `ssr.external` matcher, but it is the same species as the entry just deleted, one line above it.
- **`tsconfig.json:9` still names `runtime`** in the "Six directories, two of them shipped" narrative. It is a historical statement about pre-iteration-5 state, so it is defensible — but it is now the only mention of a directory that does not exist, and unverifiable from the tree.
- **`entries.length > 15`** against 19 actual. Same note as `> 50` vs 496 on #306, milder here.
- **`const ROOT = join(import.meta.dirname)`** — single-argument `join` is a no-op, carried over from `docs-imports.test.ts`.

## Verdict

Correct, well-evidenced, and genuinely safe — the "no way to have been a working consumer" argument survives both ways I tried to break it (`exports`-aware and classic resolution). Merge it.

Finding **#1** is worth acting on in this PR specifically, because it is the thing the PR claims to have already done: the release-time invariant that actually catches this bug class is still a manual step, three lines from being enforced. **#2** and the nits are fine as follow-ups.

PR #314, opened 2026-08-04
Own the ORM's types, so an app never installs @prisma/client

nstfkc:
Stacked on #313 — base it at `next` once that merges.

## The problem

The generated model bases did `import type { Prisma } from "@prisma/client"` and built every signature out of `Prisma.<M>FindManyArgs` and `Prisma.<M>GetPayload<T>`. One type-only line, erased at build, never in a bundle — and still required to **resolve** at typecheck, so every gemi app carried a 74MB package, and generated a further 23MB of client, for types it never ran. (Measured against 6.19.2. The query engine is *not* in `@prisma/client` — it ships with the `prisma` CLI, which an app keeps, and is copied into whatever client `prisma generate` writes. So what goes away is ~97MB of client, not the engine.)

**Nothing outside gemi ever asked for it.** Measured, not assumed:

| | |
|---|---|
| `prisma` CLI 6.19.2 deps | `@prisma/config`, `@prisma/engines` — no client, not even optional or peer |
| `migrate dev` / `deploy` / `db push` / `migrate diff` / `validate` / `format` | all exit 0 against a schema with **zero** generator blocks and `@prisma/client` unresolvable |
| gemi's generator binary | requires only `child_process`, `stream`, `util`; zero import sites for `@prisma/client` |

The requirement was gemi's own, enforced by `assertPrismaClientGenerator`.

It was also describing the wrong library. Prisma's argument types describe Prisma's query engine; gemi compiles and runs its own SQL. `Prisma.<M>FindManyArgs` admits `cursor` and `distinct` — both of which gemi [refuses permanently and by design](https://github.com/nstfkc/gemi/blob/next/packages/gemi/orm/compile/read.ts) — so the emitted types type-checked code that threw at runtime.

## The split

`orm/types.ts` holds the **combinators**: `select`/`include` narrowing, filter grammars, argument shapes, aggregate payloads. The generator emits the **facts** only it knows — each column's concrete TypeScript type, where each relation points, which key combinations are unique, which columns `create` may omit. None of those are derivable from the runtime metadata, which deliberately records a column's *storage* classification rather than its type.

```ts
export type UserScalars = {
  id: number;
  email: string;
  plan: "free" | "pro";      // enum members, which the metadata does not carry
  metadata: JsonValue | null;
};
export type UserRelations = { posts: { kind: "many"; nullable: false; target: PostTypes } };
export type UserUnique = { id: number } | { email: string };
```

Each grammar is taken from the compiler it feeds — read keys from `READ_ARGS`, writes from `WRITE_ARGS`, scalar operators from `OPERATORS`, list filters from `LIST_FILTERS`, nested writes from `SUPPORTED`, and `_sum`/`_avg` restricted to numeric columns as `compile/aggregate.ts` restricts them.

Recursion is bounded by the caller's own literal — a relation is descended into only when the argument object names it — so a cyclic relation graph cannot produce an unbounded instantiation.

## What gets better

- `cursor` and `distinct` are **compile errors**. The runtime refusal stays for untyped callers.
- `_sum` over a `String` is a compile error, at the same place the runtime refuses it.
- Foreign-key columns are accepted in `create` — gemi's write compiler takes them, and Prisma's `CreateInput` hid them behind `connect`.

## Consequences, not extras

- **`Prisma.DbNull` / `Prisma.JsonNull`** were the one piece of ordinary app code needing a *value* import of `@prisma/client`. `json-null.ts` has always recognised the sentinels structurally, so `gemi/orm` exports its own; Prisma's keep working.
- **`app/database/prisma.ts`** — a `PrismaClient` singleton shipped to every scaffolded app, imported by one demo cron job. Deleted.
- **The generator's `client` option** and `assertPrismaClientGenerator`. Deleted.

## Packaging

`prisma/schema.prisma` has no `generator client` block — that is what an app author starts from. gemi's differential harness still needs a Prisma to compare against, so it generates one from `differential.prisma`: that file plus a client block, **derived rather than duplicated**, with a guard that fails when the two drift. `@prisma/client` moves to the workspace root, where it is what it actually is — a dependency of gemi's test suite, not of the app the template describes.

## Verification

End to end in a scratch app with only `prisma` installed and `@prisma/client` unresolvable (`MODULE_NOT_FOUND`): `prisma migrate dev` migrates *and* runs the gemi generator; the emitted models typecheck with 0 errors; and nine invalid queries each fail with a precise message — unselected column, `cursor`, `distinct`, `_sum` over a string, unknown aggregate field, unknown `where` column, non-member enum value, `create` missing a required column, partial row to `wrap`.

- 2510 gemi tests pass (108 files)
- 655 template tests, including **404 differential comparisons** against a real Prisma client
- 36 type tests, no type errors
- template's pre-existing typecheck errors: 278 → 277, verified per-file against a baseline worktree at HEAD
- `oxlint orm --deny-warnings`: 0 warnings

New guard `orm/app-dependencies.test.ts` asserts the four things that must hold together: the template's manifest does not name `@prisma/client`, its schema has no client generator, its generated artifacts import no Prisma package, and the differential schema stays in sync.


packages/gemi/orm/types.ts
```diff
@@ -0,0 +1,678 @@
...
+  notIn?: NonNullable<V>[];
+};
+
+type OrderingFilter<V> = NonNullable<V> extends Comparable
+  ? {
+      lt?: NonNullable<V>;
+      lte?: NonNullable<V>;
+      gt?: NonNullable<V>;
+      gte?: NonNullable<V>;
+    }
```
nstfkc, line 311: **`Record<string, never>` is not an empty object — it is `{ [k: string]: never }`, and intersecting with it poisons every sibling property.**

`NestedFilter<V> = BaseFilter<V> & OrderingFilter<V> & StringFilter<V>`. For any non-`String` column at least one of these two conditionals takes the false branch, so `BaseFilter`'s `equals`/`not`/`in`/`notIn` each have to satisfy `never` as well:

```
Type '{ gt: number; }' is not assignable to type 'NestedFilter<number>'.
  Type '{ gt: number; }' is not assignable to type 'Record<string, never>'.
    Property 'gt' is incompatible with index signature.
      Type 'number' is not assignable to type 'never'.
```

Verified with `tsc --strict` against the committed `app/models/generated/models.ts` on this branch:

```ts
User.findMany({ where: { id: { gt: 5 } } })                   // TS2322 — Int
User.findMany({ where: { createdAt: { gte: new Date() } } })  // TS2322 — DateTime
User.findMany({ where: { id: { in: [1, 2] } } })              // TS2322 — Int
Account.findMany({ where: { token: { equals: null } } })      // TS2322 — Bytes
User.findMany({ where: { email: { contains: "x" } } })        // ok — String only
```

So on every `Int`, `Float`, `BigInt`, `DateTime`, `Boolean` and `Bytes` column, the operator-object form is gone entirely and only the bare-value shorthand survives. The where compiler accepts all of these, so this is the type-checked-but-diverges failure inverted: the types now refuse what the runtime implements.

The fix is `{}` in both false branches (here and line 276) — an intersection with `{}` is the identity. I confirmed the five queries above compile with that change, and that the narrowing still bites in the intended direction:

```ts
// @ts-expect-error `contains` is not a filter on a number column
const a: WhereInput<UserTypes> = { id: { contains: "x" } };
// @ts-expect-error `gt` is not a filter on a boolean column
const b: WhereInput<UserTypes> = { active: { gt: true } };
```

Both still error, so `{}` costs none of the precision the split was for.

If `{}` reads too loose to leave uncommented, `Record<never, never>` works too and says "no keys" rather than "any key, valued `never`" — which is the thing that was actually meant here.

packages/gemi/bin/orm/emit.ts
```diff
@@ -359,18 +359,267 @@ export function emitSchemaFile(schemas: ModelSchema[]): string {
...
+function relationsType(schema: ModelSchema): string {
+  const relations = Object.values(schema.relations).map(
+    (relation) =>
+      `  ${propertyName(relation.name)}: { kind: "${relation.kind}"; ` +
+      `nullable: ${relation.nullable}; target: ${relation.model}Types };`,
+  );
+
+  if (relations.length === 0) {
+    return `
+export type ${schema.name}Relations = Record<string, never>;
```
nstfkc: **Same `Record<string, never>` trap, and here it takes out a whole model.**

`WhereInput<M>` spreads `{ [K in keyof Relations<M>]?: RelationFilter<Relations<M>[K]> }`. When `Relations<M>` is `Record<string, never>`, `keyof` is `string`, so the model's `WhereInput` acquires an index signature `[key: string]: RelationFilter<never>` — and every scalar key then has to satisfy it too.

`Membership` in the template is exactly this shape (four columns, no relations). Verified against the committed generated models:

```ts
Membership.findMany({ where: { organizationId: 1 } })
// TS2322: Property 'organizationId' is incompatible with index signature.
//   Type 'number' has no properties in common with type
//   '{ some?: …; every?: …; none?: … }'

Membership.findUnique({ where: { organizationId_userId: { organizationId: 1, userId: 2 } } })
// TS2322 — `WhereUniqueInput` is `M["unique"] & Partial<WhereInput<M>>`,
//          so it inherits the same poisoning
```

That is `findUnique`, `findUniqueOrThrow`, `update`, `delete`, `upsert`, `deleteMany`, `updateMany` and every `where` — a relation-less model cannot be addressed by its own primary key. `select` and `orderBy` happen to survive, which makes it look partly working rather than broken.

`{}` here as well:

```ts
export type ${schema.name}Relations = {};
```

`{}` still satisfies `ModelTypeInfo["relations"]`'s `Record<string, RelationInfo>` constraint (confirmed under `--strict`), and `keyof {}` is `never`, so the mapped type contributes nothing — which is what was intended.

Worth a case in `emit.test.ts` too: the existing descriptor test asserts `UserRelations` exists, and `User` has relations, so nothing currently emits or typechecks this branch.

packages/gemi/orm/types.ts
```diff
@@ -0,0 +1,678 @@
...
+  | { [key: string]: JsonValue };
+
+/**
+ * What a `Json` column accepts on a write.
+ *
+ * The two sentinels are the column's two empty states — SQL NULL and the JSON
+ * value `null` — and a bare `null` is not among them: the choice is the whole
+ * reason they exist, and guessing between them is what silently stores the
+ * wrong one of two legal answers.
+ */
```
nstfkc, line 96: **`JsonInput` accepts a bare `null`, which is the one thing it exists to refuse.**

`JsonValue` includes `null` — correctly, since a `Json` column can hold the JSON value `null` — and `JsonInput` is `JsonValue | DbNullValue | JsonNullValue`, so `null` is assignable. Verified against the committed generated models:

```ts
User.create({ data: { metadata: null } })  // compiles
```

`docs/orm.md`, in this PR, says the opposite two ways over:

> **Empty is two values, and you have to say which.** `metadata: null` does not type-check, because a `Json` column has two distinct empty states…

and the doc comment right above this line says guessing between them "is what silently stores the wrong one of two legal answers". Prisma's `InputJsonValue` excluded top-level `null` for exactly this reason, so this is a regression the migration guide actively tells people not to expect.

```ts
export type JsonInput =
  | Exclude<JsonValue, null>
  | DbNullValue
  | JsonNullValue;
```

Confirmed: `metadata: null` becomes an error, `metadata: DbNull` and `metadata: { a: [1, null] }` both still compile — nested nulls are unaffected, which is the distinction Prisma draws too.

One related note on `JsonFilter` (line 305): `equals: null` stays accepted even after that change, because `Exclude<JsonValue, null>` still contains `{ [key: string]: JsonValue }` and `{ equals: null }` is itself a valid JSON object. Prisma has the same ambiguity, so I would leave it — but the comment above `JsonFilter` currently states flatly that "a bare `equals: null` is not accepted, matching the compiler", and the runtime does raise `InvalidArgumentError` there. Worth softening the comment to say the type cannot express it rather than that it does.

packages/gemi/orm/types.ts
```diff
@@ -0,0 +1,678 @@
...
+interface NestedUpsert<M extends ModelTypeInfo> {
+  where: WhereUniqueInput<M>;
+  create: CreateInput<M>;
+  update: UpdateInput<M>;
+}
+
+/**
+ * `create`'s `data`: the generator's scalar split, with the relation half laid
+ * over it.
+ */
```
nstfkc, line 577: **A required FK stays required even when the row is created through nested `connect`.**

`CreateInput<M>` is `M["create"] & { relations }`, and `optionalOnCreate()` marks a column optional only when it is nullable, defaulted, or `@updatedAt`. A required FK is none of those, so it is required in the intersection — regardless of whether the relation half supplies it. Verified against the committed generated models:

```ts
SocialAccount.create({
  data: {
    provider: "github", providerId: "1",
    accessToken: "a", refreshToken: "b", expiresAt: new Date(),
    user: { connect: { id: 1 } },
  },
})
// TS2322: Property 'userId' is missing … but required in SocialAccountCreateScalars

LedgerEntry.create({
  data: { amount: 1, ledger: { connect: { tenantId_code: { tenantId: 1, code: "x" } } } },
})
// TS2322: 'tenantId' and 'ledgerCode' missing
```

The runtime supports `connect` — it is in `SUPPORTED` and `NestedCreate` models it here — so this is the types refusing something the write compiler does, which is the failure mode the PR body sets out to eliminate in the other direction.

The PR body frames exposing FK columns as a strict improvement over Prisma hiding them behind `connect`, and it is. But Prisma offers *both*: `CreateInput` takes `connect`, `UncheckedCreateInput` takes the raw column, and the operation accepts either. Here the two halves are intersected rather than unioned, so `connect` is offered and simultaneously not sufficient — the one combination neither library supports.

The smallest fix that keeps the FK columns visible is to make a column optional in `create` when it backs a relation, and let the runtime keep enforcing that exactly one of the two spellings arrives (which it must do anyway for untyped callers). `buildModelSchemas` already knows each relation's `fields`, so `optionalOnCreate` can take the set of relation-backing columns and add a fourth reason.

That trades a compile error for a runtime one on the genuinely-missing case. If that is the wrong trade for this project, the alternative is a union of two exact shapes per relation, which is more type machinery than the rest of this file uses — worth deciding deliberately either way, but the current shape makes the canonical Prisma spelling uncompilable, which I do not think is intended.

templates/saas-starter/prisma/differential.prisma
```diff
@@ -0,0 +1,267 @@
...
+  previewFeatures = ["driverAdapters"]
+}
+
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+// Runtime model metadata and the typed model bases the gemi ORM executes
+// queries through. The output is committed.
+//
+// **There is no `generator client` block, and that is the point.** Prisma owns
```
nstfkc: The comment carried over from `schema.prisma` now contradicts the file it lives in — this is twelve lines below a `generator client` block:

> **There is no `generator client` block, and that is the point.**

Same in `postgres-only-differential.prisma`, where the inherited paragraph reads "No `generator client` here, for the same reason `schema.prisma` has none".

Both files are `Do not edit.`, so the only place to fix it is `derive()`. Since the banner already explains what the file is, stripping the source's leading comment block — or replacing the specific paragraph — would keep the derived artifact from asserting the opposite of its own contents. Cosmetic, but these are committed files that a reader will hit while working out how the split works, which is precisely when a contradictory comment costs the most.

templates/saas-starter/prisma/differential-schema.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * So the block lives here instead, appended to a copy. Derived rather than
+ * maintained as a second file because the two must describe the *same* models
+ * or the comparison is worthless — and a hand-kept duplicate drifts silently,
+ * with the failure showing up as a differential test that passes because it
+ * compared the wrong thing.
+ *
+ * The output is committed, like the ORM artifacts and for the same reason: CI
+ * needs no codegen step before it can read the schema, and the diff stays
+ * reviewable. `differential-schema.test.ts` fails if it falls out of date.
```
nstfkc: `differential-schema.test.ts` does not exist. The guard is `packages/gemi/orm/app-dependencies.test.ts`, which imports `SCHEMAS` and `derive` from this file — worth naming, since the path is not one a reader would guess from here (it lives in the other package).

packages/gemi/orm/template-import-graph.test.ts
```diff
@@ -68,9 +80,11 @@ describe("the template's model suites", () => {
     ).toBeGreaterThan(0);
   });
 
-  test(`${HARNESS} still imports @prisma/client, which is why this test exists`, () => {
+  test(`${HARNESS} still imports a Prisma client, which is why this test exists`, () => {
     const source = readFileSync(join(MODELS, "differential.ts"), "utf8");
-    expect(source).toMatch(/^import .*from "@prisma\/client";$/m);
+    expect(source).toMatch(
+      new RegExp(`^import .*from "${CLIENT.replace(".", "\\.")}";$`, "m"),
```
nstfkc: `String.prototype.replace` with a string pattern replaces only the first occurrence, so this escapes the leading `.` of `"./prisma-client"` and would silently leave any later one unescaped. Correct for the current constant, and quietly wrong if `CLIENT` ever gains a second dot (`./prisma-client/index`, say) — at which point the `.` matches any character and the test still passes.

`CLIENT.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")` — or just `new RegExp(...)` built from a split/join — keeps it honest.

nstfkc (commented):
## Review

The premise is right and well-evidenced: the `prisma` CLI genuinely does not need `@prisma/client`, the coupling really was one type-only import, and deriving `differential.prisma` rather than duplicating it is the correct way to keep the harness's oracle. The three headline improvements are real — I verified against the committed `app/models/generated/models.ts` that `cursor`, `distinct` and `_sum` over a `String` are now compile errors.

**But the new filter grammar has a defect that makes the most common query form in the ORM uncompilable, and the template's typecheck cannot see it.** Verified with `tsc --strict` against the committed generated models on this branch:

```ts
User.findMany({ where: { id: { gt: 5 } } })                  // TS2322
User.findMany({ where: { createdAt: { gte: new Date() } } }) // TS2322
Account.findMany({ where: { token: { equals: null } } })     // TS2322
Membership.findMany({ where: { organizationId: 1 } })        // TS2322
Membership.findUnique({ where: { organizationId_userId: {…} } }) // TS2322
```

Two independent root causes, both `Record<string, never>` used as an "empty" object — it is not empty, it is an index signature whose every value is `never`, and intersecting with it poisons every sibling property. Details inline on `types.ts:266` and `emit.ts:536`.

CI is green because **no typechecked code in the template exercises an operator-object filter, a relation-less model's `where`, or a nested `connect`.** The model suites go through `$exec` with hand-built schemas and `any`; the `.test-d.ts` files cover `wrap`, `select` and policies. That gap is the finding behind the finding — this PR moves the entire argument grammar in-house, and the grammar has no type tests.

### Blocking

1. **`Record<string, never>` in the `OrderingFilter`/`StringFilter` false branches** breaks `equals`/`not`/`in`/`notIn`/`gt`/`gte`/… on every non-`String` column. (`types.ts:266`, `types.ts:276`)
2. **`relationsType()` emits `Record<string, never>` for a relation-less model**, giving its `WhereInput` a `[key: string]: RelationFilter<never>` index signature. `Membership` in the template cannot be filtered, `findUnique`d, updated or deleted. (`emit.ts:536`)
3. **`JsonInput` accepts a bare `null`**, because `JsonValue` includes `null`. `docs/orm.md` in this very PR says `metadata: null` "does not type-check". It does. (`types.ts:75`)
4. **A required FK is mandatory even when the row is created by nested `connect`** — Prisma's canonical spelling is a compile error. (`types.ts:526`)

I confirmed all four fixes typecheck: `{}` for both `Record<string, never>` sites, `Exclude<JsonValue, null>` in write position. Fix 4 needs a design call.

### Suggested test coverage

A `filters.test-d.ts` next to `wrap.test-d.ts`, covering one column of each storage class (`Int`, `DateTime`, `String`, `Boolean`, `Bytes`, `Json`, `String[]`) against each operator the compiler accepts, plus one relation-less model and one nested `connect`. Every failure above would have been a red test.

### Non-blocking

- The derived `.prisma` files declare `generator client` and then, ten lines later, carry the inherited sentence "**There is no `generator client` block, and that is the point.**" Inline on `differential.prisma:23`.
- `differential-schema.ts:19` names a `differential-schema.test.ts` that does not exist.
- `template-import-graph.test.ts:86` escapes only the first `.`.
- The template still declares `"prisma": "^6.5.0"` while the root pins `6.19.2`. Hoisting makes them agree today; the floating range is what a scaffolded app inherits.
- `RUNTIME_IMPORTS` is emitted unconditionally — a zero-model schema emits 28 unused imports; only the two `Json` ones are gated.
- The header claims recursion "is bounded by the caller's own literal". True of `Payload`; `WhereInput`, `OrderByInput`, `SelectInput` and `IncludeInput` recurse unconditionally through the relation graph and typecheck because TS defers object-type resolution, not because the literal bounds them. Worth restating so the next person does not rely on a bound that is not there.
- `enumMap` uses `value.dbName ?? value.name`. For a `@map`ped enum member Prisma's client speaks the TS name and gemi (as far as I can tell) speaks the DB name — so this is right for gemi and deliberately divergent from Prisma. No schema in the repo has a value-level `@map`, so the differential harness never compares it. Worth a test or a note either way.

### What is good

- Deriving `differential.prisma` with a drift guard, rather than keeping a second copy, is the right call — and `derive()` refusing a source that already has a client block closes the failure mode properly.
- Rewriting `runtime-isolation.test.ts` to assert on *module specifiers* rather than on file text, with the reasoning recorded, is exactly right: the old grep would have taught authors to delete the explanations.
- Moving `DbNull`/`JsonNull` into `gemi/orm` while keeping the `Prisma.…` tag so both spellings interoperate is a clean migration path.
- The CI comments explaining *why* the differential generate runs from the repository root will save someone a long afternoon.

PR #317, opened 2026-08-04
Register models from the modules, and let the kernel's importer exit

### LINKED ISSUE #316: 0.51.0 release readiness: an unregistered subclass silently serves unpolicied rows, and the reference migration has never executed a query
Feedback from migrating Folio (`apps/web`, 79-model schema, Prisma 7 with a custom client output) from 0.48.1 to `0.51.0-rc.3`. The four issues that blocked the earlier RCs are all closed — this is what I'd want resolved before cutting `0.51.0`.

### Verified on rc.3

| Check | Result |
|---|---|
| `bun run type-check` (root, all packages) | 3/3, 0 errors |
| `bunx vitest run` (apps/web) | 1470/1470, 191 files |
| `bun run build` (root, all apps) | 3/3 |
| `bunx oxlint` on new/changed files | clean |
| `gemi migrate --dry-run` | accurate plan, nothing left to codemod |
| `gemi-orm-generator` on the real 79-model schema | succeeds, all 79 emitted |
| registry override, all 79 names | every name resolves to the app subclass, none to a base |

rc.3's type-layer rewrite cost this repo **nothing but a regenerate** — no source file changed. It also removed the rc.2 workaround: the generator block no longer needs `client = "@folio/db"`, because nothing it emits resolves to a Prisma package. That option was load-bearing for any repo whose client output isn't `@prisma/client`, and now nothing depends on where the client lives.

---

- [ ] **1. An unregistered subclass silently serves unpolicied rows through `include`**

The generator registers `UserModel` under the name `"User"`. The natural place to hold policies is a hand-written `class User extends UserModel {}` in app code. If that module isn't imported, or the `register("User", User)` call is missing, a nested `include` resolves the relation **by name**, finds the generated base, and returns rows with no policy applied. No error, no warning.

This is the one I'd gate the release on. It fails in the wrong direction — the mistake is invisible and the failure mode is an authorization hole — and the generator having already registered the base is exactly what makes it silent. The docs are honest about it, but that's doc-only mitigation for a security-shaped default.

`assertPoliciesRegistered` is already exported from `gemi/orm`, so the machinery looks like it exists and just isn't on by default. Any of these would close it: call it during boot; have relation resolution refuse a name whose registered class isn't the policy-carrying one; or have the generator emit bases *unregistered* so registration is the app's explicit act. Post-1.0 all three are breaking changes. Right now it's a patch.

- [ ] **2. No runtime evidence from the reference migration**

Everything in the table above is static — compile, unit tests, build, generator. The ORM has not executed a single query against Postgres in this migration, and the unit suites covering the ported paths mock the model module outright, so they prove call shapes and nothing about the SQL.

What integration coverage does gemi have of its own against a real database for the combinator-typed generator output? If it's solid, this item is just me owing you a run. I'll put the Folio auth port through production-mode QA — signup across all three auth paths, a deliberately failed provisioning to prove the user rolls back rather than orphaning, and a concurrent-provision race against the real driver's unique violation — and report back either way.

- [ ] **3. `await import("gemi/kernel")` never lets the process exit**

Resolves in ~110ms, then the process hangs forever: no open sockets, no timers, empty kqueue. Reproduces identically on 0.48.1, so it predates the 0.51 line and is not a release regression. Practical cost: any one-off script importing `gemi/kernel` needs an explicit `process.exit()` or it hangs CI.

- [ ] **4. Installing a custom `UserProvider` forces full ORM adoption, and that cost isn't documented**

Removing the auth adapter seam means every `UserProvider` method resolves its model through the ORM registry. An app with zero ORM adoption must therefore generate and commit the models (764 KB, 3 files, 79 models here) and import them at boot before auth works *at all*. The error message is genuinely excellent —

```
ModelNotRegisteredError: No model is registered under the name 'User'.
Nothing is registered — is app/models/generated/index.ts imported?
```

— but you meet it at runtime, after the port is already written. Worth a line in the auth and migration docs saying the ORM comes with the provider, not optionally alongside it.

- [ ] **5. Transaction-crossing code loses its unit-test seam**

Before, a test injected a fake client. Now `Model.transaction` resolves its connection ambiently, so a test that crosses a transaction must either mock the model module or boot a Kernel. Both affected Folio suites took the mock route (`vi.mock("@/app/models", …)` with `transaction: (cb) => cb()`), and one moved from a fake client to `vi.spyOn` on the model statics. Same seam, one level down — but it's a real change in how DB-touching code gets unit-tested, and there's currently no documented story for it.

- [ ] **6. Two connection pools, no sizing guidance**

`DatabaseConfig.url` defaults to `DATABASE_URL`, so an app mid-migration opens gemi's Bun `SQL` pool alongside its existing Prisma pool against the same database. Not a bug, and unavoidable for an incremental migration — but pool sizing should be a decision the migration docs prompt rather than a default nobody notices.

---

Items 4–6 are docs and ergonomics. Item 2 is evidence rather than code. Item 1 is the only one I'd actually hold the release for.

nstfkc:
Closes #316 items 1 and 3, and documents 4–6. Item 2 is answered below rather than coded.

## 1. An unregistered subclass silently serves unpolicied rows — fixed

The gate. Two things were wrong, and the second is why the first had no working mitigation.

**The hole.** `register("User", User)` is a correct instruction and a bad mechanism: forgetting it is invisible. A relation read resolves its target by name, so an unregistered policied subclass leaves the generated base owning the name and every nested `include` comes back unscoped. `$exec`'s guard begins `registered !== this`, so a model only ever read through an include never trips it.

**`registerModels(...modules)` derives the registration from the module** instead of asking an author to repeat the name. The name is `$schema.name` — the generator wrote it, the subclass inherits it. `Kernel.models` is that call at boot:

```ts
// app/kernel/Kernel.ts
import * as generated from "../models/generated"
import * as models from "../models"

export default class extends Kernel {
  models = [generated, models]
}
```

Later modules win, so each subclass takes the name its base was holding, and the audit runs over the whole set before the container is touched. For a 79-model schema that is one line and one barrel, not 79 `register` calls.

Election is per module, and the two shapes that put more than one candidate in a namespace want **opposite** answers:

| the module | wants | why |
|---|---|---|
| `export * from "./generated"` + `class User extends UserModel` | the subclass | electing the base *is* the leak |
| `class User` + `class AdminUser extends User {}` | `User` | the view inherits the policies either way, and would apply its own narrowing to every nested read |

An own `$schema` separates them — the generator declares it on the base, a subclass inherits — so application classes beat bases, and among application classes the least derived wins. Two unrelated claims raise `AmbiguousModelRegistrationError` rather than getting picked between.

**`register` still works** and is still what you write for a class in a module the Kernel doesn't own. The template keeps its line in `User.ts` because that module is also imported alone, by tests and by `preload.ts`.

### The second defect: the recommended audit threw on the correct arrangement

`docs/orm.md` told you to run `assertPoliciesRegistered(generated, models)`. That call walked the generated namespace, found `UserModel` diverging from the policied `User` that had replaced it, and threw — so **the only documented mitigation for the hole above failed the moment an app wrote its first policy**, and passed only while there was nothing to protect. On a 79-model schema you'd have hit this on the first policy.

Reachability through a base is not the bug; *querying* through one is, and that's a property of a call site rather than of a module's export list. `$exec` keeps that direction (`UserModel.findMany()` still raises with `carries: "registered"`); the audit stops reporting ancestors. Siblings — two classes extending the same base, only one registered — still diverge and are still refused.

### Residual, stated plainly

It sees the modules it's handed. A policied class in a file no barrel re-exports is still invisible, and closing *that* needs the framework enumerating the filesystem at boot. The barrel is the discipline: one file to add a model to, where a missing line is a thing you can notice.

Not taken: emitting bases unregistered (every app then needs explicit registration for models it never subclassed, and a one-call helper puts you straight back where you started), and refusing at relation resolution (unimplementable — nothing at runtime can see a class that was never registered and never queried).

## 3. `await import("gemi/kernel")` never let the process exit — fixed

Reduced to one module-scope import. On React 19.2.3 under Bun:

```
import 'react-dom/server.browser'   → hangs, getActiveResourcesInfo() === []
import 'react-dom/server.node'      → exits
import 'satori'; import 'sharp'     → exits
```

`ViewRouteDispatcher` had `react-dom/server.browser` at the top, and that module is in the graph of everything reaching `RouteServiceProvider` — so the hang propagated to `gemi/kernel` and every script that touched it. Chain: `kernel → RouteServiceProvider → ViewRouteDispatcher → react-dom/server.browser`.

Loaded on the first render now, wrapped so the three call sites read unchanged. `react-dom` is external in the framework build, so it stays a runtime resolve out of the app's `node_modules` exactly as the static import was. A server pays it once at first request; a script that never renders no longer pays it at all.

Tested by running a script and watching it end (`kernel/exit.test.ts`) rather than by asserting the import list — a static check would pass while the same hang came back through a different module, and would fail on a lazy import that is fine. Verified it fails when the static import is restored.

`satori` and `sharp` are still eager. Neither hangs, and making them lazy would move a missing-peer-dependency failure from boot to first OG request. Happy to do it as a follow-up if you want the cold start.

## 2. Runtime evidence — what exists today

Not just static. The ORM has a real-database suite that runs on every CI push:

- **21 model suites in the template, 14 of them database-backed**, on the real generated artifact for the template's 13-model schema — the same generator output path your 79 models take.
- **Two CI jobs**: `sqlite` runs the dialect-agnostic half; `postgres` brings up a `postgres:16` service container, repoints `schema.prisma`, `db push`es, regenerates, and runs the Postgres-labelled suites against it. 133 tests in the template suite are Postgres-gated and skip locally without `TEST_POSTGRES_URL`.
- **Three of those suites are differential**: they execute the same query through gemi and through a real generated Prisma client against the same Postgres and compare results, which is what pins the combinator-typed output to Prisma's actual semantics rather than to our reading of them.

What that does *not* cover is your auth port specifically — provisioning rollback, the three sign-up paths, a concurrent-provision race against the driver's unique violation. Those are app-shaped and I'd want your run. The generic paths underneath them (nested writes rolling back the parent row, `UniqueConstraintError` identification, savepoint semantics) are covered against real Postgres.

## 4, 5, 6 — docs

- **4.** `docs/authentication.md` now says the ORM comes *with* `UserProvider` rather than optionally alongside it: no adapter seam, every method resolves through the registry, so an app with zero other ORM adoption still generates, commits and registers the models before auth works at all — ~750 KB across three files on a 79-model schema. Stated where you plan the port, not where you hit it at runtime.
- **5.** `docs/orm.md` documents the unit-test seam for code crossing `Model.transaction`: `transaction: (cb) => cb()` in a module mock, or `vi.spyOn` on the model statics — and says explicitly that rollback, savepoints and the Postgres abort need a database and belong in an integration test.
- **6.** A migration section prompts the pool-size decision: `DatabaseConfig.url` defaults to `DATABASE_URL`, so registering your first gemi model opens a second pool against the same database with two defaults each picked as though it were the only one. Shows splitting the budget via `options: { max }` and lowering Prisma's `connection_limit` by the same amount.

Also fixed in passing: the Json filter examples still wrote `Prisma.AnyNull` while the surrounding examples had moved to the bare `gemi/orm` spelling.

## Verification

```
packages/gemi     110 files, 2540 passed, 1 skipped
template           20 files, 655 passed, 133 skipped (Postgres-gated)
tsc --noEmit      clean (packages/gemi)
test:types        61 passed, no type errors
oxlint            clean on changed files (1 pre-existing warning)
```

Postgres jobs run in CI. The `exit.test.ts` regression check was confirmed to fail with the static import restored and pass with it lazy.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -78,6 +78,39 @@ structurally rather than by identity.
 
 </details>
 
+<details>
+<summary>Migrating model by model, with a Prisma client still in the app</summary>
+
+Nothing requires the switch to happen all at once. A `PrismaClient` and the gemi ORM coexist
+fine — but they are **two connection pools against the same database**, and by default nobody
+chooses their sizes.
+
+`DatabaseConfig.url` defaults to `DATABASE_URL`, which is normally the same URL the Prisma client
+already has. So the moment you register your first gemi model, the process holds Prisma's pool and
+gemi's Bun `SQL` pool at once, and their two defaults were each picked as though they were the only
+one. Against a managed Postgres with a connection cap — or a PgBouncer with a fixed pool — the
+symptom is connection exhaustion under a load that used to be fine, and it names neither library.
+
+Decide the split before you start rather than after the first incident:
+
+```typescript
+export default defineDatabaseConfig({
+  url: process.env.DATABASE_URL,
+
+  // Passed straight through to Bun's `SQL` client.
+  options: { max: 5 },   // gemi's share of the budget
+});
+```
+
+and lower Prisma's `connection_limit` by the same amount, so the two together stay inside the
+budget one of them used to have. Rebalance as models move across; when the last one has, drop the
+Prisma client and give gemi the whole budget.
+
+Pointing gemi at a *different* `url` is the other honest answer — a separate pool with its own cap,
+at the cost of two things to configure per environment.
+
+</details>
+
 Then `bunx prisma generate`. You get three files under `app/models/generated/`:
 
 - `schema.ts` — runtime metadata: tables, columns, types, defaults, relations.
@@ -126,45 +159,79 @@ refused, by the row above rather than by a rule about lists.
 
 ### Your model class
 
-The generated base is not the class you write code on. Subclass it, and **re-register it**:
+The generated base is not the class you write code on. Subclass it, put the subclass in a barrel,
+and list that barrel on your Kernel:
 
 ```ts
 // app/models/User.ts
-import { register } from "gemi/orm"
 import { UserModel } from "./generated"
 
 export class User extends UserModel {}
-
-register("User", User)
 ```
 
-That `register` line is not bookkeeping. A relation read resolves its target through the registry
-by name, so whatever is registered under `"User"` is the class that runs inside every nested
-`include`. If that is the generated base while your policy lives on your subclass, the policy
-applies to root queries and is skipped inside includes — scoped one way, unscoped the other, with
-nothing to notice it. `Model.$exec` raises `UnregisteredPolicyClassError` when a class carrying
-policies is queried while a *different* class owns its name — but that guard is narrower than it
-sounds, and the gap runs the wrong way. **A model you only ever read through an `include` never
-trips it:** the include resolves the name to the unpolicied generated base, nothing diverges from
-nothing, and the rows come back unscoped with no error. That is the shape a membership or pivot
-model usually has, and it is exactly the kind that carries a tenant scope. Write the line next to
-every subclass regardless — the guard is a backstop for the cases it can see, not a substitute.
+```ts
+// app/models/index.ts
+export { User } from "./User"
+```
 
-For the case it cannot see, there is an audit you can run once instead of trusting thirteen
-subclasses:
+```ts
+// app/kernel/Kernel.ts
+import { Kernel } from "gemi/kernel"
+import * as generated from "../models/generated"
+import * as models from "../models"
+
+export default class extends Kernel {
+  models = [generated, models]
+}
+```
+
+`boot()` registers every model class those modules export under the name its schema carries, later
+modules winning — so each subclass takes the name its generated base was holding — and then checks
+that no policied class lost its name to something else.
+
+**Why the framework does this rather than you.** A relation read resolves its target through the
+registry by name, so whatever is registered under `"User"` is the class that runs inside every
+nested `include`. If that is the generated base while your policy lives on your subclass, the
+policy applies to root queries and is skipped inside includes — scoped one way, unscoped the other.
+The registration used to be a `register("User", User)` line next to each subclass, and forgetting
+it is invisible: `Model.$exec` raises `UnregisteredPolicyClassError` when a class carrying policies
+is *queried* while a different class owns its name, but that guard is narrower than it sounds and
+the gap runs the wrong way. **A model you only ever read through an `include` never trips it** — the
+include resolves the name to the unpolicied base, nothing diverges from nothing, and the rows come
+back unscoped with no error. That is the shape a membership or pivot model usually has, and exactly
+the kind that carries a tenant scope.
+
+Deriving the registration from the module removes the mistake instead of reporting it. The name is
+not a string you retype; it is `$schema.name`, which the generator wrote and your subclass inherits.
+
+`register` still exists and still works, for a class you want registered from a module you are not
+handing to the Kernel. And you can call the same thing directly — `registerModels(generated,
+models)` from `gemi/orm` — in a script or a test that boots no Kernel.
+
+Two classes in **one** module claiming the same model raise `AmbiguousModelRegistrationError`,
+unless one extends the other. A generated base and the subclass over it are not ambiguous (the
+subclass wins), nor are a class and a typed view over it (the class wins, and the view inherits its
+policies either way). Two siblings both written for the same model are, and `register` is how you
+say which.
+
+#### What this still cannot see
+
+Modules you hand it. A policied class in a file that no barrel re-exports is invisible to the
+Kernel exactly as it was to a forgotten `register` line — which is the reason for the barrel: one
+file to add a model to, and a missing line in it is a thing you can notice.
+
+`assertPoliciesRegistered` is the same audit without the registration, for running over modules the
+Kernel does not own:
 
 ```ts
 import { assertPoliciesRegistered } from "gemi/orm"
 import * as generated from "@/app/models/generated"
-import * as models from "@/app/models/User"
+import * as models from "@/app/models"
 
 assertPoliciesRegistered(generated, models)
 ```
 
-Same rule, triggered differently: it reads the classes out of the module namespace, so a policied
-class that nothing queries is still visible. In a test it closes the hole for CI at no runtime cost;
-at boot it turns a deploy of the mistake into a failure to start rather than a quiet cross-tenant
-read. It can only see modules you hand it, so it does not replace the `register` line either.
+In a test it closes the hole for CI at no runtime cost.
 
 ## Querying
 
@@ -921,6 +988,34 @@ migration whose transactions are legitimately long.
 The warning is development-only and never fires in production, whatever the config says. It is a
 diagnostic, not a limit: nothing cancels a long transaction.
 
+### Unit-testing code that opens one
+
+`Model.transaction` resolves its connection ambiently, out of the `AsyncLocalStorage` scope — which
+is exactly what lets `audit(user)` above join the transaction without being handed anything. The
+same property means there is no client to inject: a unit test of a function that crosses a
+transaction has to either boot a Kernel or replace the model module.
+
+Replacing the module is usually what you want, because the point of the test is the branching
+around the write rather than the SQL:
+
+```ts
+import { vi } from "vitest"
+
+vi.mock("@/app/models", () => ({
+  User: { create: vi.fn(), findFirst: vi.fn() },
+  transaction: (cb: () => unknown) => cb(),   // run the callback, open nothing
+}))
+```
+
+`transaction: (cb) => cb()` is the whole trick: the callback runs inline, the assertions are about
+what it called, and no connection is involved. `vi.spyOn` on the model statics does the same job
+when you want the real module for everything else.
+
+What this does not cover is the behaviour that only a real transaction has — the rollback, the
+savepoint, the Postgres abort described above. Those need a database, and they belong in an
+integration test rather than in a mock. Keep the seam for the control flow and test the transaction
+itself against Postgres.
+
 ## Policies
 
 A model carries a **list** of policies. Every member of a policy is optional; a model with none
@@ -1172,6 +1267,7 @@ Every failure is a typed error from `gemi/orm`, not a driver string.
 | `UnknownFieldError` / `UnknownRelationError` | A name that is not on the model. |
 | `UnsupportedQueryError` | A query shape the compiler does not implement — with what and why. |
 | `ModelNotRegisteredError` / `UnregisteredPolicyClassError` | Registry problems (see [Setup](#your-model-class)). |
+| `AmbiguousModelRegistrationError` | `registerModels` found two unrelated classes in one module claiming the same model, and refused to pick (see [Setup](#your-model-class)). |
 | `RelationDepthExceededError` | An include tree past `MAX_RELATION_DEPTH`. |
 | `ParameterLimitError` | A statement exceeding the dialect's parameter ceiling. |
 | `MissingRequiredValueError` | A write leaving a required column with no value and no default. |
@@ -1396,11 +1492,13 @@ Two things are worth knowing:
   There is a **third** sentinel, and it belongs only in a filter:
 
   ```ts
-  await User.findMany({ where: { metadata: { equals: Prisma.AnyNull } } })  // both kinds at once
-  await User.findMany({ where: { metadata: { not: Prisma.AnyNull } } })     // neither kind
+  import { AnyNull } from "gemi/orm"
+
+  await User.findMany({ where: { metadata: { equals: AnyNull } } })  // both kinds at once
+  await User.findMany({ where: { metadata: { not: AnyNull } } })     // neither kind
   ```
 
-  `Prisma.AnyNull` asks for *either* empty state, which is usually what you want when you do not
+  `AnyNull` asks for *either* empty state, which is usually what you want when you do not
   care how the row got that way — it compiles to `is null or = 'null'`, one predicate rather than
   an `OR` you assemble yourself. It is a question about rows rather than a value, so writing it
   raises `InvalidArgumentError`, and so does using it under any operator other than `equals` and
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -189,6 +189,11 @@ export default class extends Kernel {
 modules winning — so each subclass takes the name its generated base was holding — and then checks
 that no policied class lost its name to something else.
 
+**Upgrading an app that already has models?** `models` defaults to `[]`, so until you write that
+line your app behaves exactly as it did before — which is the arrangement the rest of this section
+is about. A Kernel with an empty `models` and a populated registry warns at boot in development.
+[UPGRADE.md](https://github.com/nstfkc/gemi/blob/main/UPGRADE.md) has the short version.
+
 **Why the framework does this rather than you.** A relation read resolves its target through the
 registry by name, so whatever is registered under `"User"` is the class that runs inside every
 nested `include`. If that is the generated base while your policy lives on your subclass, the
@@ -209,10 +214,33 @@ handing to the Kernel. And you can call the same thing directly — `registerMod
 models)` from `gemi/orm` — in a script or a test that boots no Kernel.
 
 Two classes in **one** module claiming the same model raise `AmbiguousModelRegistrationError`,
-unless one extends the other. A generated base and the subclass over it are not ambiguous (the
-subclass wins), nor are a class and a typed view over it (the class wins, and the view inherits its
-policies either way). Two siblings both written for the same model are, and `register` is how you
-say which.
+unless one extends the other. A generated base and the subclass over it are not ambiguous — the
+subclass wins. Nor are a class and a typed view over it, *as long as the view adds no policies of
+its own*: the class wins and the view inherits its policies, so which one the registry holds changes
+nothing. Two siblings both written for the same model are ambiguous, and `register` is how you say
+which.
+
+Across **different** modules there is no election at all — the later one simply wins. What catches a
+bad outcome there is the audit that runs at the end, so a conflict between two policied classes is
+still refused; two unpolicied ones resolve by import order, which changes no query's scope.
+
+**A view that carries its own policies is refused rather than chosen between.** The registry holds
+one class per name, so registering `AdminUser` would apply its narrowing to every nested read of
+`User`, and leaving `User` registered would skip `AdminUser`'s policies entirely — both silent. Keep
+such a class out of the modules you hand to `Kernel.models` and query it directly where you want the
+narrowing:
+
+```ts
+// app/models/views/AdminUser.ts — not exported from app/models/index.ts
+export class AdminUser extends User {
+  static $policies: UserPolicy[] = [{ scope: () => ({ archived: false }) }]
+}
+
+const rows = await AdminUser.findMany({ … })   // narrowed, at the call site
+```
+
+If it is not a view but the model itself, move the policies onto the class the name resolves to and
+every subclass inherits them everywhere.
 
 #### What this still cannot see
 
--- a/packages/gemi/kernel/Kernel.ts
+++ b/packages/gemi/kernel/Kernel.ts
@@ -2,6 +2,7 @@ import { Application } from "../foundation/Application";
 import type { ServiceProviderConstructor } from "../foundation/Application";
 import type { ServiceToken } from "../container/Container";
 import type { ConfigItems } from "../support/Repository";
+import { registerModels } from "../orm/registration";
 import { Scheduler } from "../services/cron/Scheduler";
 import { BroadcastManager } from "../services/pubsub/BroadcastManager";
 import { QueueManager } from "../services/queue/QueueManager";
@@ -29,6 +30,36 @@ export class Kernel {
    */
   protected providers: ServiceProviderConstructor[] = [];
 
+  /**
+   * The modules holding the app's model classes — normally
+   * `[generated, models]`, the generated namespace and the barrel of subclasses
+   * written over it. Every model class they export is registered under the name
+   * its schema carries, later modules winning, and the whole set is audited
+   * before anything else boots.
+   *
+   * ### Why this is a Kernel field and not a line in a file somewhere
+   *
+   * Registration used to be `register("User", User)` next to each subclass, and
+   * the failure mode of forgetting it is an authorization hole that nothing
+   * reports: a relation read resolves its target by name, so an unregistered
+   * policied subclass leaves the generated base owning the name and every
+   * nested `include` of that model comes back unscoped. No error at the root
+   * either, because `$exec`'s divergence guard begins `registered !== this` and
+   * the base it is running on *is* the registered one. A model only ever read
+   * through an include — a membership, a pivot, the kind that carries a tenant
+   * scope — never trips anything.
+   *
+   * Listing the modules here says it once, in the same place the app already
+   * declares its config slices and providers, and turns the mistake into a
+   * failure to boot rather than a quiet cross-tenant read.
+   *
+   * It sees the modules it is given, so a model class in a file that no barrel
+   * re-exports is still invisible. That is the residual, and it is a smaller
+   * one: a barrel is a place to notice an omission, and thirteen `register`
+   * lines scattered across thirteen files is not.
+   */
+  protected models: Array<Record<string, unknown>> = [];
+
   readonly app = new Application();
 
   /**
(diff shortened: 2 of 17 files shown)


packages/gemi/kernel/Kernel.ts
```diff
@@ -38,6 +69,11 @@ export class Kernel {
    * Phase two is `waitForBoot()`.
    */
   boot() {
+    // Before the container, deliberately. This throws on a model whose policies
+    // would be skipped, and the useful moment for that is the one where nothing
+    // has been constructed and no request can be in flight.
+    if (this.models.length > 0) registerModels(...this.models);
```
nstfkc, line 76: **The fix defaults to off, so every app that upgrades keeps the hole.**

`models` is `[]` and this line no-ops on empty, which means an app on 0.48.1 that upgrades to rc.4 is in exactly the state #316 describes until somebody edits their Kernel by hand. Nothing in the repo tells them to: `docs/orm.md` documents it under **Setup**, which is the page an existing app read once and will not read again, and `UPGRADE.md` is still the 0.42→0.43 guide and is untouched here. There is no 0.50→0.51 note anywhere.

#316 was filed by an app mid-migration with a 79-model schema. That reader gets the new mechanism, the barrel, the audit — and none of it runs.

The cheap version that actually reaches somebody is a dev-mode warning right here:

```ts
if (this.models.length > 0) registerModels(...this.models)
else if (process.env.NODE_ENV !== "production" && registry.registeredNames().length > 0) {
  console.warn(
    `[gemi] ${registry.registeredNames().length} models are registered but Kernel.models is empty. ` +
    `Policies on a subclass are skipped inside nested includes unless the subclass owns its name — ` +
    `set models = [generated, models]. See docs/orm.md#your-model-class.`,
  )
}
```

It fires exactly for the app that has models and hasn't adopted this, and stays silent for a Kernel with no ORM at all. An upgrade note would help too, but a note only reaches the people who go looking.

packages/gemi/orm/registration.ts
```diff
@@ -87,6 +122,159 @@ export function assertPoliciesRegistered(
...
+      if (existing === undefined) candidates.set(name, [model]);
+      else existing.push(model);
+    }
+
+    for (const [name, classes] of candidates) {
+      registry.register(name, elect(name, classes));
+    }
+  }
+
+  assertPoliciesRegistered(...modules);
```
nstfkc: **The registry is already mutated by the time this throws**, so a refused registration leaves the arrangement it refused in place.

Verified against this branch:

```
after a failed registerModels: UnregisteredPolicyClassError
  registry names     = ["User"]
  "User" resolves to = Theirs
```

`Theirs` is the class the audit just refused, and it is now what every nested `include` of `User` resolves to. `Kernel.test.ts` makes a point of this exact principle — *"a registration that would leak should stop where nothing has been constructed"* — and the registry is constructed, process-wide, and survives the throw. `boot()` propagating is enough for a server that dies, but not for a dev server that catches and keeps serving, or any harness that continues past the error.

Elect into a local map, audit that, and commit only on success:

```ts
export function registerModels(...modules: Array<Record<string, unknown>>): void {
  const staged = new Map<string, ModelClass>()
  for (const module of modules) {
    // ...elect as today, into `staged` rather than `registry`
  }
  const restore = [...staged.keys()]
    .filter((name) => registry.has(name))
    .map((name) => [name, registry.get(name)] as const)

  for (const [name, model] of staged) registry.register(name, model)
  try {
    assertPoliciesRegistered(...modules)
  } catch (error) {
    for (const [name, model] of restore) registry.register(name, model)
    throw error
  }
}
```

Restoring is a bit awkward because `registry` has no delete, but the property is worth it: after a throw, the registry says what it said before the call.

packages/gemi/orm/registration.ts
```diff
@@ -87,6 +122,159 @@ export function assertPoliciesRegistered(
...
+ *     export * from "./generated"      // UserModel, and
+ *     export class User extends UserModel { static $policies = [...] }
+ *
+ * wants the subclass — electing the base here *is* the leak, in the one file an
+ * author most plausibly writes it in. While
+ *
+ *     export class User extends UserModel { static $policies = [...] }
+ *     export class AdminUser extends User {}
+ *
+ * wants `User`. `AdminUser` is a typed view over the same rows; it inherits the
```
nstfkc, line 274: **This case is reasoned about here and then rejected sixty lines later.**

The docblock contemplates a view that adds policies of its own — *"silently applies its own narrowing to every nested read when it does"* — and elects `User` anyway, which is the right call. But `registerModels` runs `assertPoliciesRegistered` immediately after, `policiesFor(AdminUser)` accumulates its own entries on top of `User`'s, the lengths differ, and the boot dies. Verified both spellings on this branch:

```
THROW view adds its own policies (one module)       -> UnregisteredPolicyClassError
THROW view adds its own policies (separate modules) -> UnregisteredPolicyClassError
PASS  view adds nothing (the tested shape)
```

The message is the part I'd fix regardless of what you decide about the shape:

> AdminUser carries policies but the registry resolves 'User' to User. […] **Register the class that carries the policy**

Following that advice — `register("User", AdminUser)` — produces precisely the every-include narrowing this function refused to produce. The error is correct that something is wrong and wrong about the way out.

Two things separate this from the sibling case cleanly, and `descendsFrom` already computes both:

- `descendsFrom(model, registered)` → the exported class is a **descendant** of the registered one. Either the leak #316 is about (an unregistered policied subclass of a base) or a policied view. Same shape, opposite advice.
- neither direction → siblings. Today's message is right for these.

So either give the descendant case its own message ("`AdminUser` extends the registered `User` and adds policies of its own. A class that narrows an existing model belongs in a module you do not hand to `Kernel.models` — registering it would apply its narrowing to every nested read of `User`."), or, if a policied view is simply unsupported, say that here and in `docs/orm.md` instead of describing the shape as handled.

packages/gemi/orm/registration.test.ts
```diff
@@ -200,3 +245,184 @@ describe("more than one module", () => {
...
+    class User extends UserBase {
+      static $policies = [scope];
+    }
+
+    registerModels({ UserBase, User });
+
+    expect(registry.get("User")).toBe(User);
+  });
+
+  test("a base, its subclass and a typed view in one module elects the subclass", () => {
```
nstfkc, line 298: This is the untested half of the shape. `class AdminUser extends User {}` with no policies of its own is the case that works; a view that adds one does not boot at all (see my note on `elect`). Whichever way that resolves, it wants a test — either

```ts
test("a view that adds policies of its own is refused", () => { ... })
```

pinning the refusal and the message, or a passing case if you decide to support it.

Separately, everything in this describe block runs against `class UserBase { static $schema = user }`, a hand-written stand-in. `elect` depends on the **generator** declaring `$schema` as an own property on the base and on subclasses inheriting rather than redeclaring it, and nothing here or in the template checks that against real generator output. `templates/saas-starter/app/models/registration.test.ts` is where that belongs — it exists for exactly this reason ("whether it holds for a schema somebody actually generated") and this PR didn't touch it.

packages/gemi/orm/registration.ts
```diff
@@ -87,6 +122,159 @@ export function assertPoliciesRegistered(
...
+ * Whether the generator wrote this class, rather than an application extending
+ * one it wrote.
+ *
+ * `$schema` is declared on the emitted base and inherited by everything below
+ * it, so owning the property is the difference. Nothing has to be added to the
+ * generated output for this to hold — it is already how `models.ts` is written,
+ * and `Model.$exec` already reads `$schema` off the prototype chain for exactly
+ * that reason.
+ */
+function isGeneratedBase(candidate: ModelClass): boolean {
```
nstfkc, line 334: Worth a line about what happens when this guesses wrong. A subclass that redeclares `static $schema` — to point at a modified schema object, or by copy-paste — reads as a generated base, `written` comes back empty, `eligible` falls through to every candidate, and the least-derived one wins: the base takes the name over the policied subclass. Verified:

```
THROW subclass redeclares $schema, in one module with its base
  -> UnregisteredPolicyClassError: User carries policies but the registry resolves 'User' to UserBase
```

So it fails loudly rather than leaking, which is the important part and worth saying out loud in this docblock — the reader's question at `Object.hasOwn` is exactly "what if that's wrong". The error is misleading in the same way as the case above (it asks for a `register` line that is not the problem), but the safety property holds.

packages/gemi/orm/errors.ts
```diff
@@ -396,6 +396,41 @@ export class UnregisteredPolicyClassError extends Error {
...
+ * a typed view over it — and a chain has a least element to elect. Two classes
+ * that extend the same base without extending each other do not: both were
+ * written for this model, and only the author knows which one nested reads
+ * should run.
+ *
+ * Refused rather than guessed at, because both guesses are silent. Picking
+ * either would scope every `include` of the model by one class's policies while
+ * the other's code keeps reading as though its own applied.
+ */
+export class AmbiguousModelRegistrationError extends Error {
```
nstfkc, line 446: "two unrelated classes **in one module**" is the accurate scope, and the asymmetry behind it is worth one sentence here: across modules there is no election, the later import simply wins, and what catches it is the audit at the end of `registerModels` rather than this error. That is fine when the two diverge in policies — `registration.test.ts` pins it — but two unrelated *unpolicied* claims split across modules resolve silently by import order and nothing reports it.

Nothing to fix in the behaviour; it just currently reads as though `registerModels` refuses every unresolvable conflict, and it refuses the subset it can see at once.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -38,6 +36,44 @@ import { injectQueryPayloads, isBotUserAgent } from "./streamQueryInjection";
...
+ * Wrapped rather than awaited at each call site so the three of them read
+ * unchanged.
+ */
+let serverRenderer:
+  | Promise<{ renderToReadableStream: (...args: any[]) => Promise<any> }>
+  | undefined;
+
+const renderToReadableStream = async (...args: any[]): Promise<any> => {
+  // @ts-ignore
+  serverRenderer ??= import("react-dom/server.browser");
```
nstfkc: `??=` caches the rejection as well as the resolution: if the first import fails, every later render awaits the same rejected promise, and the second failure reads as whatever React's loader threw the first time with no stack from the call that actually failed. A missing `react-dom` is fatal either way, so this is about the diagnostic rather than about recovery — clearing the slot on failure costs one line:

```ts
const renderToReadableStream = async (...args: any[]): Promise<any> => {
  // @ts-ignore
  serverRenderer ??= import("react-dom/server.browser").catch((error) => {
    serverRenderer = undefined
    throw error
  })
  return (await serverRenderer).renderToReadableStream(...args)
}
```

The deferral itself checks out — `dist/chunk-yxgmazze.js` already emits a literal `from"react-dom/server.browser"`, so it is external in the built package today and the dynamic form stays a runtime resolve exactly as you describe.

packages/gemi/kernel/exit.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const result = spawnSync(
+      "bun",
+      ["-e", `await import(${JSON.stringify(entry)}); console.log("loaded")`],
+      { encoding: "utf8", timeout: 10_000 },
+    );
+
+    // `signal` is what a timeout kill shows up as, and it is the failure this
+    // test is for — distinguished from an ordinary non-zero exit, which would
+    // mean the import itself broke and is a different bug.
+    expect(result.signal, result.stderr).toBeNull();
```
nstfkc, line 53: `result.error` is the third outcome and it isn't checked. If `bun` isn't on `PATH`, `spawnSync` returns `status: null`, `signal: null` and sets `error` to an ENOENT — so this assertion passes, and the failure surfaces on the next line as `expected undefined to contain "loaded"`, which names neither the cause nor the fix.

```ts
expect(result.error, "could not spawn bun").toBeUndefined()
```

ahead of the signal check turns the environment problem into its own message. Cheap, and this test is one somebody will meet in a context where bun isn't the runner.

The test works, by the way — I restored the static `react-dom/server.browser` import and it fails on the 10s timeout with `signal` set, then passes again once reverted.

docs/orm.md
```diff
@@ -126,45 +159,79 @@ refused, by the row above rather than by a rule about lists.
...
+Deriving the registration from the module removes the mistake instead of reporting it. The name is
+not a string you retype; it is `$schema.name`, which the generator wrote and your subclass inherits.
+
+`register` still exists and still works, for a class you want registered from a module you are not
+handing to the Kernel. And you can call the same thing directly — `registerModels(generated,
+models)` from `gemi/orm` — in a script or a test that boots no Kernel.
+
+Two classes in **one** module claiming the same model raise `AmbiguousModelRegistrationError`,
+unless one extends the other. A generated base and the subclass over it are not ambiguous (the
+subclass wins), nor are a class and a typed view over it (the class wins, and the view inherits its
```
nstfkc: "the view inherits its policies either way" is true only for a view that adds none. A view that carries its own policies fails the boot — `registerModels` elects the class, then its own audit refuses the view — so as written this sentence describes an arrangement an author cannot run.

Suggest making the condition explicit:

> nor are a class and a typed view over it, as long as the view adds no policies of its own (the class wins, and the view inherits its policies). A view that *does* carry its own policies belongs in a module you don't hand to the Kernel — registering it would apply its narrowing to every nested read of the model, and leaving it registered under the base would skip its policies entirely, so `registerModels` refuses rather than choosing.

nstfkc (commented):
## What this does

Closes #316's gating item by replacing `register("User", User)` with `Kernel.models = [generated, models]`, deriving each registration from `$schema.name` so the forgettable line is gone; fixes the kernel import hang by deferring `react-dom/server.browser` to first render; and documents items 4–6.

Both fixes are real, and I checked them rather than taking the body's word for it.

**Item 3 is verified end to end.** I restored the static `react-dom/server.browser` import in `ViewRouteDispatcher.ts` and `exit.test.ts` fails — `spawnSync` kills it at the 10s timeout, `signal` set. Reverted, it passes in ~200ms. So the regression check earns its keep, and testing the *property* (the process ends) instead of the import list is the right call. The build claim holds too: `dist/chunk-yxgmazze.js` already carries a literal `from"react-dom/server.browser"`, so it is external today and the dynamic form stays a runtime resolve out of the app's `node_modules`.

**The second defect you found is the better half of item 1.** `assertPoliciesRegistered(generated, models)` throwing on the correct arrangement — so the only documented mitigation failed the moment an app wrote its first policy, and passed only while there was nothing to protect — is a sharper diagnosis than the issue asked for. The ancestor skip is also sound in a way worth stating: `policiesFor` accumulates up the prototype chain unconditionally, so an ancestor can never carry a policy its registered descendant lacks. The skip cannot swallow a real divergence, and the sibling test pins the case it must not.

32 tests across `exit.test.ts`, `Kernel.test.ts` and `registration.test.ts` pass under `bun --bun vitest`.

## What I'd change

Two I'd want addressed, both verified by running them:

1. **The fix is opt-in and defaults to off.** `Kernel.models` is `[]` and `boot()` no-ops on empty, so every app upgrading to rc.4 is exactly as exposed as it was before this PR until somebody edits their Kernel — and the only place that says so is the Setup section of `docs/orm.md`, which an existing app has already read. `UPGRADE.md` is scoped to 0.42→0.43 and untouched here; there is no 0.50→0.51 note in the repo. #316 came from an app migrating a 79-model schema. That is precisely the reader who will not see this.

2. **`registerModels` mutates the process-wide registry before it audits**, so a refused registration leaves the leaking mapping in place. After `registerModels({Ours}, {Theirs})` throws, `registry.get("User")` is `Theirs` — the class the audit just refused. `Kernel.test.ts` asserts the throw lands "where nothing has been constructed", but the registry is constructed and global.

And one design/docs inconsistency: a typed view that carries its own policies — the shape `elect`'s docblock explicitly reasons about — cannot boot at all, and the error tells you to do the thing `elect` refused to do. Details inline.

## Coverage gap

`elect` turns entirely on `Object.hasOwn(candidate, "$schema")` being true for a generated base and false for a subclass. That is a contract with the **generator**, and nothing in this PR checks it against generator output — the unit tests all use a hand-written `class UserBase { static $schema = user }` stand-in.

The file whose stated purpose is to close that gap, `templates/saas-starter/app/models/registration.test.ts` ("what they cannot cover is whether it holds for a schema somebody actually generated"), was not updated: it still imports `* as models from "./User"` and still exercises only `assertPoliciesRegistered`. So `registerModels` and `Kernel.models` never touch a real generated class anywhere in this PR. Two lines in that file would fix it:

```ts
expect(Object.hasOwn(generated.UserModel, "$schema")).toBe(true)
expect(Object.hasOwn(models.User, "$schema")).toBe(false)
```

plus pointing its import at the new `./index` barrel, so the barrel is load-bearing in a test rather than only in the template's Kernel.

## On item 2

The answer is fair, and the differential suites against real Postgres are the right thing to point at. Worth noting the boundary explicitly in the reply: those pin *query semantics*, and none of the registration machinery this PR adds runs against a database at all — it is class identity and prototype chains, which is fine, but it means "we have runtime evidence" and "item 1's fix has runtime evidence" are different claims.

nstfkc:
All nine addressed in `ad167c3`. I reproduced each one first — three of them changed what I thought the design was, and one turned out to be broader than reported.

## The two you'd hold for

**1. Defaulting to off.** You're right that this is the whole difference between shipping a fix and shipping a mechanism. `boot()` now warns in development when `models` is empty *and* the registry is populated — that combination is the signature of the exposed app specifically, so an app with no ORM stays silent. Added a `0.50 → 0.51` section to `UPGRADE.md`, which had nothing past 0.42→0.43, covering both this and the `@prisma/client` removal.

Warning rather than throwing: an app in this state is running today, and turning a working deploy into a failure to start over an arrangement that might be fine isn't a patch-level move.

**2. Mutating before auditing.** Confirmed exactly as you described — after the throw, `"User"` resolved to `Theirs`. Now elected into a local map and committed only after the audit passes, with a rollback to the prior classes on failure. One residue I called out in the comment: `register` has no inverse, so a name that was *absent* before the call stays pointing at what this call elected. That's the harmless direction — the alternative is `ModelNotRegisteredError`, and the class elected is the one that module set would have used anyway.

## The policied view — and a case underneath it

You were right that the shape is reasoned about and then rejected, and right that the message was the worst part of it. Digging in, my ancestor rule was keyed on the wrong question. `isGeneratedBase` doesn't separate "view" from "model" — a plain unpolicied intermediate isn't a view either, and your `Base`-two-levels-up test was passing only by accident of it owning `$schema`.

The rule is now: **does the ancestor carry policies at all?** `policiesFor` accumulates upwards, so an unpolicied ancestor means the descendant is the first class to define policies and therefore *is* the model (register it — #316's advice, correct); a policied ancestor means the descendant is narrowing something already established (a view — keep it out). Three answers instead of two, via `carries: "narrowing"`:

> `AdminUser` extends `User`, which owns the model 'User', and carries policies of its own on top of it. […] Keep the view out of the modules you hand to `Kernel.models`, and query it directly where you want it: `AdminUser.findMany(…)`

Dropping `isGeneratedBase` from the audit is a bonus — the generator contract is now confined to `elect`.

**The case that rule also closes, which I don't think the review reached.** A policied view in a *separate* module never met competition, so `elect` never ran on it, it won the name outright, and the ancestor skip then read `User` as a base its subclass had replaced and waved it through:

```
view-separate: PASS -> AdminUser      # before
view-separate: THROW narrowing        # after
```

Silently, every nested read of the model acquired the view's narrowing. Both spellings are tested now.

## The rest

- **`isGeneratedBase` guessing wrong** — documented, with your finding: the base takes the name, the audit refuses it a moment later, loud-and-wrong beats silent-and-wrong. Test added (`carries: "queried"`).
- **`AmbiguousModelRegistrationError` scope** — added the asymmetry: election is per module, cross-module conflicts are caught by the audit instead, and two unrelated *unpolicied* claims resolve by import order and report nothing. Tolerable for the same reason it's invisible — with no policies on either, the choice changes no query's scope.
- **Cached rejection** — took your `.catch` verbatim.
- **`result.error`** — added ahead of the signal check, with your message.
- **docs typed-view sentence** — rewritten with the condition explicit, plus a worked example of where a policied view should live.

## Coverage gap

Fixed as you specified, and a bit past it. `templates/saas-starter/app/models/registration.test.ts` now imports the new `./index` barrel, asserts both halves of the `Object.hasOwn(_, "$schema")` contract against real generator output, and — new describe block — runs `registerModels(generated, models)` on the actual thirteen generated classes, checking that every name registers and that `models.User` takes the name from `UserModel` with no `register` call involved. So the mechanism, not just the audit, touches real generated classes.

## On item 2

Taking the correction. Those differential suites pin *query semantics* against a real Prisma client on real Postgres; the registration machinery this PR adds is class identity and prototype chains and never touches a database. "The ORM has runtime evidence" and "item 1's fix has runtime evidence" are different claims, and only the first is true. What backs the second is unit tests plus the template's real generated classes — which is appropriate for what it is, but it isn't runtime evidence and I shouldn't have let the two sit next to each other.

## Verification

```
packages/gemi     110 files, 2552 passed, 1 skipped   (was 2540)
template           20 files,  659 passed, 133 skipped (was 655)
tsc --noEmit      clean (packages/gemi)
test:types        61 passed, no type errors
oxlint orm --deny-warnings   0 warnings, 0 errors
```

Reproduced against the branch after the change:

```
rollback:          UnregisteredPolicyClassError; User -> Established  (restored)
view-one-module:   THROW narrowing
view-separate:     THROW narrowing
redeclare-$schema: THROW queried
intermediate:      PASS -> User          (your Base-two-levels-up shape)
```

PR #319, opened 2026-08-04
Check the files, and let the generator say which classes it wrote

### LINKED ISSUE #318: Model registration's two remaining silent cases: a class no barrel exports, and a subclass that redeclares $schema
Follow-up to #316 / #317. **Neither of these blocks 0.51.0** — #317 states both in its own comments and I agree with the calls it made. Filing so they're written down somewhere other than a merged PR body.

Both verified against the #317 branch (`fix/316-registration-and-kernel-exit`), four assertions, all passing as described below.

---

## A. A policied class no barrel re-exports is still invisible

`registerModels` sees the modules it is handed, so the #316 hole survives one level up: instead of forgetting `register("Membership", Membership)`, you forget the `export { Membership } from "./Membership"` line in `app/models/index.ts`. The generated base keeps the name, every nested `include` of that model comes back unscoped, and nothing raises — the audit can't report a class it was never given, and `$exec`'s guard needs a root query that an include-only model never gets.

```ts
class User extends UserBase { static $policies = [scope] }   // written
const models = {}                                            // and not exported

registerModels({ UserModel: UserBase }, models)              // does not throw
registry.get("User")                                         // -> UserBase
assertPoliciesRegistered({ UserModel: UserBase }, models)    // does not throw
```

**Why it isn't a blocker.** It's a strict improvement on what shipped: one file where a missing line is noticeable, against 79 `register` calls scattered across 79 files. And #317 is right that closing it *at boot* means the framework enumerating the filesystem, which buys a real cost — bundler coupling, dev/prod divergence, a boot-time walk every process pays.

**What might close it cheaply:** enumerate at **check** time instead of boot time. The filesystem walk is fine when it's a CLI step nobody's request is waiting on — `gemi check models` (or a dev-server-start scan) reading `app/models/*.ts`, finding exported classes with a `$schema`, and reporting any the barrel doesn't re-export. Runs in CI, costs production nothing, and needs no bundler cooperation. The check is nearly the one `registerModels` already does; only the source of the class list changes.

## B. `isGeneratedBase` is an inference, and its silent direction is the benign one

`Object.hasOwn(candidate, "$schema")` reads "the generator wrote this." A subclass that redeclares `static $schema` — copy-paste, or pointing at a modified schema object — therefore reads as a base, every candidate looks generated, and `elect` falls through to least-derived: the base takes the name over the subclass.

#317 says this fails safe, and it does, in the direction that matters:

```ts
class User extends UserBase { static $schema = user; static $policies = [scope] }
registerModels({ UserModel: UserBase, User })   // throws UnregisteredPolicyClassError ✓
```

The audit runs immediately after election, sees a policied class that didn't get its name, and refuses — loud and wrong beats silent and wrong.

The direction that stays silent is the one with nothing to protect:

```ts
class User extends UserBase { static $schema = user; static displayName = () => "…" }
registerModels({ UserModel: UserBase, User })   // no throw
registry.get("User")                            // -> UserBase, so displayName never runs
```

No policy, so no scope changes and it isn't an authorization bug — but statics, row methods and hooks on that class quietly don't run inside nested reads, and the same class without the `$schema` line wins the name normally. Confusing to debug, not dangerous.

**Why it isn't a blocker.** The dangerous direction throws, and `templates/saas-starter/app/models/registration.test.ts` pins the contract against real generator output, so a generator change can't invalidate the assumption without a red test.

**What would remove the inference:** have the generator mark its own output — `static $generated = true` on the emitted base — and read that in `isGeneratedBase`, falling back to `Object.hasOwn(candidate, "$schema")` for artifacts emitted by older generators. Then "the generator wrote this class" is a fact the generator stated rather than a property of how it happens to declare `$schema` today, and the redeclaration case stops being special at all.

---

Both are small. A is the one I'd actually pick up, because it is the last silent version of #316 and a CLI check is a contained piece of work.

nstfkc:
Closes #318. **Stacked on #317** — the base branch is `fix/316-registration-and-kernel-exit`, because both items are edits to code that PR introduced. Retarget to `main` once #317 lands.

Neither item blocks 0.51, and the issue says so. Both are the last silent cases left in model registration.

---

## A. A policied class no barrel re-exports

`Kernel.models` audits the modules it is handed, so #316's mistake survives one level up: instead of forgetting `register("Membership", Membership)`, you forget

```ts
export { Membership } from "./Membership"
```

The generated base keeps the name, every nested `include` of that model comes back unscoped, and nothing raises — the audit cannot report a class it was never given, and `$exec`'s guard needs a root query an include-only model never gets.

**`gemi check models`** closes it where the issue suggested: at check time, not boot time.

```
$ gemi check models
Checked 7 model files against 13 registered models. Every policied class owns its name.
```

It walks `app/models`, imports each file, and applies the rule `registerModels` already applies — only the source of the class list changes. Exit `1` on a finding, printed under its file with the export that fixes it. A step in the sqlite CI job runs it against the template.

`assertPoliciesRegistered` is now a two-line wrapper over a new **`auditModelRegistrations`**, which returns the errors rather than throwing the first. The checker needs all of them, and needs `carries` to sort a leak from a view.

### Three decisions worth your attention

**It does not credit a `register` call a file made while being imported.** A class that owns its name only because something loaded its module loses it the day nothing does — which is the failure being checked for, so crediting it would make the check agree with the bug. The registry is restored to what the declared modules made it before each file is audited, and the baseline is taken *before* the first import for the same reason.

**It does not report unpolicied classes.** One missing from the declared modules is a real if smaller problem, but it is indistinguishable from a typed view somebody meant to keep local.

**It does not report a policied view.** `registerModels` refuses one precisely so its narrowing does not reach every nested read, and that error tells the author to keep it out of `Kernel.models`. A checker demanding they export it would be telling them to undo it.

### It imports what it walks

Unavoidable — you cannot find a class without evaluating the module that declares it. Not hypothetical either: the first build, pointed at the template, imported `app/models/bench/run.ts`, ran the benchmark, and rewrote `plans/orm/benchmarks.json`. Tests, type tests, benchmarks, `.d.ts` and directories with their own `package.json` are skipped; `--ignore` takes the rest and the command prints what it skipped, so a narrowed run does not read like a complete one.

---

## B. `isGeneratedBase` was an inference

`Object.hasOwn(candidate, "$schema")` reads "the generator wrote this", and a subclass that redeclares `static $schema` defeats it. The generator now marks its own output:

```ts
export class UserModel extends Model {
  static $schema = schema.User;
  static $generated = true;
}
```

Own-vs-inherited is the whole test, since a subclass inherits the mark and does not own it. `"$generated" in candidate` selects between two worlds rather than two classes: an app whose `app/models/generated` predates 0.51 keeps the old inference — including its sharp edge, which is loud rather than silent, and which regenerating removes. Asking `Object.hasOwn` first and falling back on `false` would have sent a subclass of a *marked* base back into the inference, which is the case the mark exists to fix.

`Model` declares the property with `declare` so nothing is emitted, and a test pins `"\$generated" in Model === false`: drop that keyword and the property exists valued `undefined` on every model class in every app, which answers the world-selecting question `true` everywhere and retires the fallback for exactly the apps that have nothing else to read.

The template's `registration.test.ts` asserts the mark against real generator output, which is what would fail if `emit.ts` ever dropped the line — the fallback cannot tell a regression from an old artifact.

---

## Verification

- `packages/gemi`: 111 files, 2581 passed, 1 skipped. `tsc --noEmit` clean. `oxlint orm --deny-warnings` at zero.
- Template: 18 files, 657 passed, 133 skipped (Postgres-gated). `test:types` 61 passed, no type errors.
- `prisma generate` re-run for both schemas; the zero-diff test passes.
- End-to-end: `bunx gemi check models --ignore bench` exits 0 on the template, and exits 1 naming the file when a policied `Membership` subclass is dropped into `app/models` without a barrel export.

## Docs

`docs/cli.md` gains the command. `docs/orm.md`'s *What this still cannot see* is rewritten around it, including the three decisions above. `UPGRADE.md` gains a "run the check once" step and a "regenerate, so registration stops guessing" section. `docs/llms-full.txt` resynced for both pages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,19 @@ jobs:
         run: bun run typecheck
         working-directory: packages/gemi
 
+      # The generator's *output*, compiled against the runtime it is generated
+      # for. The step above covers `bin/orm` — the code that writes the artifact
+      # — and not the artifact, and the two can disagree: #319 shipped a
+      # `static $generated = true` that widened to `boolean` against a base
+      # declaring `?: true`, so every emitted class was a TS2417 and every check
+      # here was green. It was found by an app regenerating a 79-model schema.
+      #
+      # After `prisma generate` above, so it reads the artifact this run
+      # produced rather than the one that happens to be committed.
+      - name: Typecheck the generated artifact
+        run: bunx tsc --noEmit -p tsconfig.generated.json
+        working-directory: packages/gemi
+
       # Everything except a named red set, rather than a path filter.
       #
       # This step used to run `vitest run orm database`, chosen so
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,6 +161,27 @@ jobs:
         env:
           TZ: UTC
 
+      # `gemi check models`, against the one real project in this repository.
+      #
+      # The command's unit tests hand it module objects; nothing in them
+      # exercises the parts that only exist outside a test — resolving the app's
+      # `gemi/orm` rather than the copy bundled into `dist/bin/gemi.js`, reading
+      # `Kernel.models` off a constructed Kernel, and importing thirteen real
+      # model files. Each of those is a way for the command to be wrong while
+      # every unit test passes, and the failure it would hide is a check that
+      # reports nothing because it is looking at an empty registry.
+      #
+      # `--ignore bench` because `app/models/bench/` is a benchmark harness that
+      # runs on import — which the command found by running it, and rewriting its
+      # report, the first time it was pointed at this directory. That is the
+      # hazard the flag exists for, so exercising the flag here is not
+      # incidental.
+      - name: Check the template's model registrations
+        run: bunx gemi check models --ignore bench
+        working-directory: templates/saas-starter
+        env:
+          TZ: UTC
+
       # `.test-d.ts` files do not match vitest's default `include`, so the step
       # above collects none of them: 15 type assertions in `wrap.test-d.ts` and
       # `aggregate.test-d.ts` ran nowhere, locally or here. They are the only
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,13 +163,16 @@ jobs:
 
       # `gemi check models`, against the one real project in this repository.
       #
-      # The command's unit tests hand it module objects; nothing in them
-      # exercises the parts that only exist outside a test — resolving the app's
-      # `gemi/orm` rather than the copy bundled into `dist/bin/gemi.js`, reading
-      # `Kernel.models` off a constructed Kernel, and importing thirteen real
-      # model files. Each of those is a way for the command to be wrong while
-      # every unit test passes, and the failure it would hide is a check that
-      # reports nothing because it is looking at an empty registry.
+      # `check-models.test.ts` covers the command over a fixture project, so
+      # what is left for this step is the one thing a fixture cannot do:
+      # resolve the app's *installed* `gemi/orm` — the same instance the model
+      # files register into — rather than the copy bundled into
+      # `dist/bin/gemi.js`. Get that wrong and the command reads an empty
+      # registry, finds nothing to report, and exits 0.
+      #
+      # Which is why the count is asserted rather than the exit code. A green
+      # run and a run that checked nothing print the same *shape* of line, and
+      # differ only in the number in it.
       #
       # `--ignore bench` because `app/models/bench/` is a benchmark harness that
       # runs on import — which the command found by running it, and rewriting its
@@ -177,7 +180,10 @@ jobs:
       # hazard the flag exists for, so exercising the flag here is not
       # incidental.
       - name: Check the template's model registrations
-        run: bunx gemi check models --ignore bench
+        run: |
+          output=$(bunx gemi check models --ignore bench)
+          echo "$output"
+          echo "$output" | grep -qE 'against [1-9][0-9]* registered models'
         working-directory: templates/saas-starter
         env:
           TZ: UTC
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -49,6 +49,38 @@ than never.
 See [docs/orm.md](./docs/orm.md#your-model-class) for the full rules, including
 what happens with a typed view that carries its own policies.
 
+### And then run the check once
+
+`Kernel.models` can only audit the modules it is handed, so the mistake it
+removes has a smaller version one level up: a policied class in a file the
+barrel does not re-export. Nothing raises for that either.
+
+```sh
+bunx gemi check models
+```
+
+It walks `app/models`, imports every file, and reports any policied class the
+declared modules do not register — with the `export` line that fixes it. Exit
+code `1` on a finding, so it is worth a step in CI. It imports what it walks,
+which matters if a file under `app/models` does work on import; `--ignore` takes
+a comma-separated list, and the command prints what it skipped.
+
+## Regenerate, so registration stops guessing
+
+Nothing breaks if you skip this, and it is one command:
+
+```sh
+bunx prisma generate
+```
+
+The generator now marks each base it emits with `static $generated = true`, and
+`Kernel.models` reads that mark to decide which of several classes claiming one
+name is the generated one and which is yours. Artifacts generated before 0.51
+carry no mark, so registration falls back to the older signal — whether a class
+declares `$schema` itself — which a subclass that redeclares `static $schema`
+defeats, handing the name to the base. That case fails loudly rather than
+silently: boot refuses it and names both classes. Regenerating removes it.
+
 ## `@prisma/client` is gone
 
 0.51 removed the type-only `@prisma/client` import from the generated model
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -94,11 +94,11 @@ This command walks `app/models`, imports every file, and asks the same question 
 ```
 
 - `--dir <path>` — walk somewhere other than `app/models`.
-- `--ignore <paths>` — comma-separated paths under `--dir` to skip. What was skipped is printed.
+- `--ignore <paths>` — paths under `--dir` to skip. Comma-separated, and repeatable. What was skipped is printed.
 
 > **Gotcha:** finding a class means evaluating the module that declares it, so every file walked is imported and a file that *does* something on import does it here. Tests, type tests, benchmarks, `.d.ts` files and directories with their own `package.json` are skipped already; `--ignore` is for the rest.
 
-A typed view carrying its own policies is deliberately *not* reported — that class is supposed to be absent from the declared modules. See [ORM → Your model class](./orm.md#your-model-class).
+It reports one thing: a class carrying policies the registered class does not. A typed view carrying its own narrowing, and an unpolicied class written against a model's schema, are both deliberately *not* reported — each is supposed to be absent from the declared modules, and exporting either would turn a working boot into `AmbiguousModelRegistrationError`. See [ORM → Your model class](./orm.md#your-model-class).
 
 ## `gemi ide:generate-api-manifest`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -95,6 +95,15 @@ This command walks `app/models`, imports every file, and asks the same question 
 
 - `--dir <path>` — walk somewhere other than `app/models`.
 - `--ignore <paths>` — paths under `--dir` to skip. Comma-separated, and repeatable. What was skipped is printed.
+- `--models <paths>` — register from these modules instead of reading `Kernel.models`. Comma-separated, and repeatable.
+
+> **When you need `--models`:** the command reads the module list off your Kernel, which means importing it — and a Kernel's import graph does not have to survive a bare runtime import. `?raw` imports, virtual modules and asset imports are all ordinary in a gemi app, and a bundler resolves them where `await import()` does not. If loading the Kernel fails, the command says which file and exits `1` rather than passing green; naming the modules yourself is the way through:
+>
+> ```bash
+> gemi check models --models app/models/generated,app/models
+> ```
+>
+> This is not the tool guessing your model layout from a filename convention — that would let the check pass on an app whose Kernel declares nothing, which is the very state it exists to find. It's you stating the list, the same act as writing `models = [...]`, and the report prints how many models it registered so a wrong list is visible.
 
 > **Gotcha:** finding a class means evaluating the module that declares it, so every file walked is imported and a file that *does* something on import does it here. Tests, type tests, benchmarks, `.d.ts` files and directories with their own `package.json` are skipped already; `--ignore` is for the rest.
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -12,7 +12,7 @@ bun run build
 bun run start
 ```
 
-> **Note:** Apart from `gemi migrate --dry-run`, the commands take no flags or options — each is a bare subcommand. gemi discovers your project from the current working directory (it expects `app/` and, for tooling commands, `app/kernel/Kernel.ts`).
+> **Note:** Apart from `gemi migrate --dry-run` and `gemi check models`, the commands take no flags or options — each is a bare subcommand. gemi discovers your project from the current working directory (it expects `app/` and, for tooling commands, `app/kernel/Kernel.ts`).
 
 ## `gemi dev`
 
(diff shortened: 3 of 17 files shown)


nstfkc (commented):
## Review

Read the whole diff against `origin/fix/318-model-registration-check`, plus `orm/errors.ts`, `orm/registry.ts`, the template's `Kernel.ts` and both barrels, and the sqlite job's log for the new step.

**Part B is done.** `"$generated" in candidate` selecting the *world* and `Object.hasOwn` selecting the *class* within it is the right decomposition, and the argument for why asking `Object.hasOwn` first would be wrong — it sends a subclass of a marked base back into the inference, the exact case the mark exists to fix — is correct. `declare` plus the `"$generated" in Model === false` test is the guard that keeps the fallback reachable, and asserting the contract at both ends (`emit.test.ts` for the line, the template's `registration.test.ts` for real output) is what makes it a contract rather than a coincidence.

Part A is the right shape too. Five things below, one of which I'd fix before merge.

---

### 1. The printed `export` line names the wrong module for any model in a subdirectory

`packages/gemi/bin/check-models.ts:420`

```ts
`    export { ${finding.className} } from "./${basename(finding.file).replace(/\.tsx?$/, "")}"\n`,
```

`finding.file` is relative to the project root, so for `app/models/billing/Invoice.ts` this prints

```
export { Invoice } from "./Invoice"
```

from a barrel at `app/models/index.ts`, where it resolves to nothing. `basename` drops the directory, and `modelFiles` has a test named *"walks nested directories, because apps group models into folders"* — so nesting is the case this was built for, not a corner of it. Someone pastes the line, gets a module-not-found, and now has two problems.

`Finding` only carries the root-relative label, and `printReport` doesn't know `modelsDir`, so the fix is a second field: carry the path relative to the models directory alongside `file`, and spell the specifier from that (`./billing/Invoice`). `auditModules` already receives `path` per file and never reads it — that's where it can come from.

The existing test uses a flat `app/models/ScopedMembership.ts`, which is why it passes.

### 2. `auditModules` filters `narrowing` but not `registered`, and the advice it then prints is actively wrong

`packages/gemi/bin/check-models.ts:232`

```ts
if (problem.carries === "narrowing") continue;
```

Three values, two of which mean something here. `carries: "registered"` is reachable from the unrelated-classes branch of `auditModelRegistrations` (`ours.length === 0 && theirs.length > 0`), and `asModelClass` only requires a `$schema` with a name — nothing has to extend `Model`. So:

```ts
// app/models/reporting/UserSnapshot.ts — not in the barrel
export class UserSnapshot extends Model {
  static $schema = schema.User;   // no policies
}
```

is not on `User`'s prototype chain, diverges from it, and comes back as a finding. Two problems with that:

- It contradicts the stated decision. *"It does not report an unpolicied class"* — this one is unpolicied and it is reported.
- The message it carries is written for a call site (*"This query goes through `UserSnapshot`… Query `User` instead"*), and this command appends *"Or export it from a module `Kernel.models` declares"* underneath it. Following that lands two unrelated classes claiming `User` in one namespace, which is `AmbiguousModelRegistrationError` and a boot failure. And `AmbiguousModelRegistrationError`'s own text is what tells authors to write this class and keep it out of the barrel — so the checker would be undoing framework advice, which is the thing the `narrowing` skip exists to avoid.

Same discriminator, other direction: `if (problem.carries !== "queried") continue;`. That's exactly "a class that carries policies the registered one doesn't", which is what this command is for.

### 3. Nothing automated asserts the command ever reports a finding

`checkModels()` itself has no test — the unit tests all enter at `auditModules`/`modelFiles`/`printReport` with module objects — and the CI step asserts the green path only. Between them, the uncovered case is precisely the one the CI comment names: *"the failure it would hide is a check that reports nothing."* A regression that made `auditModules` return `[]`, or that resolved the wrong `gemi/orm`, prints `Every policied class owns its name`, exits 0, and CI is green.

The PR body says the red path was verified by hand — dropping a policied `Membership` into the template without a barrel export and watching it exit 1. That's the test. A `mkdtempSync` fixture with an `app/kernel/Kernel.ts`, a generated-ish base and two model files (one exported, one not), asserting `findings` on the real entry point, would also cover `declaredModules`, the `Bun.resolveSync` branch and the ordering below — none of which anything currently executes outside CI's happy path.

Cheaper alternative if the fixture is more than it's worth: a second CI step that writes a leaky model into the template, expects exit 1, and deletes it. Uglier, but it pins the direction that matters.

### 4. `snapshotRegistry`'s two claims are both slightly off — the property holds for another reason

`packages/gemi/bin/check-models.ts:191`

> Taken *before* a single model file is imported

It isn't. `declaredModules()` imports `app/kernel/Kernel.ts`, which does `import * as generated from "../models/generated"` and `import * as models from "../models"` — so by the time `snapshotRegistry` runs at line 354, the generated `index.ts` has already executed thirteen `register()` calls and every file the barrel reaches has run its top-level side effects, including any legacy `register("User", User)`.

The guarantee survives, but `registerModels(...declared)` is what delivers it: it runs after that import and re-asserts the declared arrangement over whatever the imports left, so a name the generator emits is put back to the elected class regardless. Worth saying that in the comment instead, since the ordering it currently credits isn't what's load-bearing — and this file is otherwise unusually careful about that distinction.

> `register` has no inverse

`clearRegistry` is exported — `orm/index.ts:226` re-exports the module whole, and `registry.clearRegistry` is on the surface this file already injects. So `restore()` can be exact:

```ts
const restore = () => {
  orm.registry.clearRegistry();
  for (const [name, model] of baseline) orm.registry.register(name, model);
};
```

which closes the residual the comment goes on to concede (a name the baseline doesn't hold, from a hand-written `$schema` the generator never emitted, staying registered from its own import). One line, and it removes a paragraph of caveat rather than adding one.

### 5. Smaller things

- **CI comment miscounts.** *"importing thirteen real model files"* — the run walks **7** (`Checked 7 model files against 13 registered models`); thirteen is the registered-model count. The PR body's sample output says 10, which matches neither. Cheap to correct, and this repo's comments are usually load-bearing enough that a wrong number is worth fixing.
- **`--ignore` doesn't accumulate.** `gemi.ts:1224`'s coercion ignores commander's `previous`, so `--ignore a --ignore b` keeps only `b`. Documented as comma-separated, so it's defensible — but `(value, previous) => [...previous, ...value.split(",")…]` is the same line and stops the silent-drop reading.
- **`$generated = false` would read as a base.** `Object.hasOwn` treats presence as the answer, so a hand-written `static $generated = false` — someone "opting out" — elects that class as the generated one. The type says `?: true` so TypeScript blocks it, and the comment says presence is the whole question deliberately. Fine to leave; noting it in case `&& candidate.$generated === true` reads cheaper than the sentence defending its absence.

---

**Summary.** #1 is the one I'd fix before merge — it's wrong advice printed by the feature's main output path, in the layout the walker is explicitly built for. #2 is narrow but the fix is one character of the same expression. #3 is the coverage gap that matters, since this command's whole failure mode is silence. #4 and #5 are accuracy, not behaviour.

nstfkc:
All five addressed in `85435e8`. Notes on the two where I did something other than the obvious.

**1. The `export` line.** Fixed as suggested — `Finding` carries a `specifier` computed relative to `modelsDir`, and the unused `path` on the file record is gone rather than repurposed, since the specifier needs the models directory and `auditModules` does not know it. The report test now uses `app/models/billing/ScopedMembership.ts`, so the flat case can no longer hide it. Verified against the template: a policied `Membership` dropped at `app/models/billing/` prints `export { Membership } from "./billing/ScopedMembership"`.

**2. `!== "queried"`.** Taken verbatim. Test added with your exact shape — a `UserSnapshot` carrying `static $schema = schema.User` and no policies, outside `User`'s prototype chain — which was reported before and is not now.

**3. Coverage.** Went with the fixture rather than the CI variant. It writes a project to a temp directory — a marked base, a subclass the barrel exports, and a policied subclass in a subdirectory it does not — and calls `checkModels` itself, so `declaredModules`, the walk, the ordering and the specifier are all executed. Six tests, including the empty-`Kernel.models` and no-Kernel refusals.

The fixture imports nothing from `gemi`: the audit reads `$schema.name` and `$policies`, and `declaredModules` wants a default-exported constructor with a `models` field, so plain classes are enough. That is what lets it live in `tmpdir()` with no `node_modules` above it — and it is why `orm` is injectable, which is the one seam I added for a test. The command never passes it.

That leaves `Bun.resolveSync` as the only line a fixture cannot reach, and it is the one whose failure is silent: resolve a *different* gemi and the registry it registers into is the registry it audits against, so the run is self-consistent and reports nothing. The exit code cannot tell those apart, but the count can, so the CI step now asserts `against [1-9][0-9]* registered models` rather than trusting exit 0.

**4.** Both taken. The `snapshotRegistry` docblock now says the ordering is *not* what makes it the declared arrangement — `registerModels(...declared)` re-asserting the election after the Kernel's imports is — and `restore()` is `clearRegistry()` plus a replay, which deletes the caveat paragraph rather than defending it.

**5.** The CI comment no longer names a file count: it is 7 in CI and 10 locally, because `generated-lists/` is gitignored and only exists after the postgres job's second `prisma generate`. A number that depends on which job you are reading is worse than no number. PR body corrected too. `--ignore` accumulates. And I took the `$generated === true` suggestion — `= false` plainly means "not generator output", and honouring it is more correct than a sentence explaining why presence is enough.

Full run after the changes: package 111 files / 2588 passed / 1 skipped, `tsc --noEmit` clean, `oxlint bin orm --deny-warnings` at zero; template 18 files / 657 passed / 133 skipped, `test:types` 61 passed.

One thing the review surfaced that I have not changed: `packages/gemi/tsconfig.json` does not include `*.test.ts`, so `bun run typecheck` did not catch the four `Finding` literals I left missing `specifier` while making this change — I found them by running `tsc` over the file directly. Pre-existing and wider than this PR, so I have left it alone rather than widen the config here.

nstfkc:
Review from the app that filed #316 and #318 — Folio, 79 models. The design is right and does what the issue asked. One defect blocks merge; one limitation is worth knowing before this lands.

## Blocking: the marker does not type-check

`emit.ts` emits

```ts
static $generated = true;
```

which TypeScript widens to `boolean`, while `Model` declares `declare static $generated?: true`. Every generated model class is therefore an error:

```
error TS2417: Class static side 'typeof AccountModel' incorrectly extends base class static side 'typeof Model'.
  Types of property '$generated' are incompatible.
    Type 'boolean' is not assignable to type 'true'.
```

**79 errors in Folio** after regenerating with this branch's generator, and **15 in this repo's own template** — `app/models/generated` and `app/models/generated-lists` both.

Why it got through: the stated gate is `tsc --noEmit` in `packages/gemi`, which does not cover template generated output. The template's own `tsc --noEmit` does cover it, but it also reaches into `packages/gemi` source and reports unrelated pre-existing errors, so it is not currently usable as a gate — the 15 new errors are indistinguishable from the noise already there.

Two one-line fixes, both verified clean in isolation:

```ts
static $generated = true as const;   // or
static readonly $generated = true;   // readonly literals do not widen
```

Either keeps the base declaration at `?: true`, so `static $generated = false` remains the type error it was meant to be, and `isGeneratedBase`'s `=== true` value read still means what it says.

Worth considering alongside it: a type-check over the template's generated output that is green today, otherwise the next generator change lands the same way.

## Not blocking: `check models` cannot load a Kernel that needs build-time transforms

`declaredModules` does a plain `await import("app/kernel/Kernel.ts")`, so the whole Kernel import graph has to be resolvable by a bare runtime import. Folio's is not — it transitively reaches `import … from "./PhotographyCraftSkillsRouting.md?raw"`:

```
Could not load app/kernel/Kernel.ts:
    Cannot find module 'folio-raw:/…/WebFolioAiRuntime.ts'
```

Exit code 1, so it fails loudly rather than passing green on a project it never checked — the right failure mode, and I would not hold the PR for this. But the command cannot run on the app that reported both issues, and `?raw` imports, virtual modules and asset imports are ordinary in a gemi app.

The check only needs the *model modules*. It reads them off the Kernel because that is where they are declared, and the docstring is right that inferring the list from a filename convention would let the check pass on an app whose Kernel declares nothing. A flag does not have that problem, because it is the author stating the list rather than the tool guessing it:

```
gemi check models --models app/models,app/models/generated
```

## Verified working, at 79 models

Substituting a minimal Kernel to get past the loader issue above:

| | result |
|---|---|
| clean run | `Checked 83 model files against 79 registered models. Every policied class owns its name.` — exit 0 |
| policied class outside the barrel | exit 1, names `app/models/billing/ScopedAccount.ts` |
| the printed fix | `export { ScopedAccount } from "./billing/ScopedAccount"` — subdirectory specifier correct |
| unmarked pre-0.51 artifacts | the fallback world works; all 79 subclasses still elected |
| regenerated with this branch | 79 `$generated = true` markers emitted, check still passes |

Suite on the branch: 111 files, 2588 passed, 1 skipped.

## Two calls worth singling out

**Injecting the ORM resolved from the project** rather than importing it statically. A second registry copy bundled into `dist/bin` would have made the check read an empty registry and report every model in the project as broken — a failure that looks like a finding, which is the worst kind.

**Not crediting an import-time `register` call.** Crediting it would make the check agree with the bug it exists to find, and the reasoning — correct only for as long as something imports the file, which is the property that just went missing — is the same argument the whole issue rests on.

nstfkc:
Both fixed in `81130ab`. Thank you for regenerating against the branch — neither of these was reachable from anything I was running.

## The marker

`static readonly $generated = true`. Reproduced first: 15 `TS2417`s in the template, 292 total errors before and 277 after, so exactly those 15 and nothing else moved.

I took `readonly` over `as const` because it also stops a runtime reassignment of a flag that decides who owns a name, and the asymmetry with the neighbouring `static $schema` is fine — that one is not a boolean sentinel. The base declaration stays `?: true`, so `= false` is still the type error `isGeneratedBase`'s `=== true` read treats it as.

`emit.test.ts` gained an assertion that the *unwidened* spelling is absent, next to the one that checks the line is present, so the keyword cannot be tidied away as noise.

## The gate

Took the suggestion, and put it in `packages/gemi/tsconfig.generated.json` rather than the template — the options have to be ones something is already green under. gemi's source is green under the package's and is not under the template's, and the artifact imports `gemi/orm`, so checking it from the template side drags every gemi source file in under options it was never checked against: I tried it, and got your noise problem back one level down (`AuthController.ts`, `HttpRequest.ts`, and so on).

Verified in both directions, which is the part that matters for a gate:

| | |
|---|---|
| as it stands | 0 errors |
| widening restored, regenerated | 15 errors |

CI runs it after `prisma generate`, so it reads the artifact that run produced rather than the one that happens to be committed. `generated-lists` is in the include and is gitignored, so it is covered in the job that creates it.

## `--models`

Implemented — I did not want to leave the command unusable on the app that filed both issues.

```
gemi check models --models app/models/generated,app/models
```

Comma-separated and repeatable, and it replaces the Kernel load rather than supplementing it. The empty-`Kernel.models` error now mentions it too, since "your Kernel declares nothing" and "your Kernel will not import" arrive at the same place.

Your framing is the one I put in the docstring: this is not the filename convention `declaredModules` rejects, because that guess would let the check pass on an app whose Kernel declares nothing — the tool would have supplied the list the app failed to. A flag is the author stating it, and a wrong list is visible in the registered count.

The test uses your shape: a fixture Kernel importing `./guide.md?raw`, asserting first that the default path fails naming `app/kernel/Kernel.ts`, then that `--models` gets the same finding without going near it.

## Verification

Package 111 files / 2592 passed / 1 skipped. `tsc --noEmit` clean on both configs. `oxlint bin orm --deny-warnings` at zero. Template 18 files / 657 passed / 133 skipped, `test:types` 61 passed, and its `tsc` down from 292 errors to 277. Both CI jobs green.

One thing I will not claim was clean: the first template run after these changes had a single failure that did not reproduce across four subsequent runs and that I did not capture the name of. Given `template-import-graph.test.ts` documents an intermittent collection failure in that suite (#217), and nothing here touches the query path, I am recording it rather than attributing it — but I would rather say so than report five-for-five.

Issue #324, opened 2026-08-05
Give CronJob a supported gate, so an app does not have to shadow callback


nstfkc:
This app cannot use `CronJob` directly. Every scheduled job it runs reports outward — email, Sentry — and none of that may fire outside a real prod deployment, so there is a gate. Because gemi calls `onTick`/`onComplete` **outside** `callback`, the gate cannot live in a lifecycle hook: anything placed there runs on a tick that `callback` was supposed to have stopped. The only place that holds is inside `callback` itself, which means every job has to remember to open with the same guard.

So the app inserts a base class whose whole purpose is that nobody can forget:

```ts
export abstract class AlertingCronJob extends CronJob {
  abstract run(): Promise<void>;

  async callback() {
    if (!alertingMode().enabled) return;
    await this.run();
  }
}
```

Four jobs extend it, and a test asserts that every job in the schedule does — because a job that extended `CronJob` directly would look identical and quietly mail from a developer's laptop.

**The ask:** a supported way to wrap or gate a job's execution, so an app does not have to shadow `callback` and rename the real body to `run`. Any of these would remove the class:

- a `before()`/`shouldRun()` hook that runs *in the same path as* `callback` and skips it on `false`;
- config-level middleware over scheduled jobs, the way HTTP middleware wraps a handler;
- or simply documenting `callback` as the wrap point and giving `onTick`/`onComplete` the same short-circuit, so a gate written once holds for the whole lifecycle.

The second-order cost is worth naming: the indirection is invisible at the call site. A job reads `async run()` where the framework's own examples read `async callback()`, so anyone following the docs writes a job that silently bypasses the gate — which is exactly the failure the base class exists to prevent, reintroduced one level up.

*From the Folio 0.51 migration (4 scheduled jobs).*

Issue #323, opened 2026-08-05
Discover cron jobs from app/cron, so the schedule does not restate the directory


nstfkc:
Same shape as #322, for the scheduler. `app/config/schedule.ts` restates what `app/cron` already contains:

```ts
import { DailyOpsDigest } from "@/app/cron/DailyOpsDigest";
import { RcAccessLockAudit } from "@/app/cron/RcAccessLockAudit";
import { RcSubscriptionAudit } from "@/app/cron/RcSubscriptionAudit";
import { RcWebhookHeartbeat } from "@/app/cron/RcWebhookHeartbeat";
import { defineScheduleConfig } from "gemi/services";

export const jobs = [DailyOpsDigest, RcSubscriptionAudit, RcAccessLockAudit, RcWebhookHeartbeat];

export default defineScheduleConfig({ jobs });
```

A cron that is written but not listed never fires, and nothing says so — worse than the queue case, because a scheduled job has no caller to notice the silence. The one in this app that reports RevenueCat webhook silence would itself go silent.

**Ask:** discover `CronJob` subclasses from `app/cron`, with the explicit list staying available as an override.

One wrinkle worth naming: this app exports `jobs` from the config module so a test can walk the same list the scheduler runs (it asserts every scheduled job inherits an alerting gate — see #324). Under discovery there is no array to import, so a discovered set needs to be readable from somewhere for that kind of test to keep working.

*From the Folio 0.51 migration.*

PR #325, opened 2026-08-05
Read app/jobs and app/cron, and give CronJob a gate that holds

### LINKED ISSUE #322: Discover queue jobs from app/jobs, so the config does not restate the directory
Every app that uses the queue writes the same list twice: once as the files under `app/jobs`, and once as an import block in `app/config/queue.ts`.

```ts
import { defineQueueConfig } from "gemi/services";
import { EnhanceProductImageJob } from "@/app/jobs/EnhanceProductImageJob";
import { GenerateProductKeywordsJob } from "@/app/jobs/GenerateProductKeywordsJob";
import { ImageGenerationJob } from "@/app/jobs/ImageGenerationJob";
import { Studio360VideoJob } from "@/app/jobs/Studio360VideoJob";
import { TranslationJob } from "@/app/jobs/TranslationJob";

export default defineQueueConfig({
  concurrency: 20,
  jobs: [
    EnhanceProductImageJob,
    TranslationJob,
    GenerateProductKeywordsJob,
    ImageGenerationJob,
    Studio360VideoJob,
  ],
});
```

Ten lines that say nothing the directory does not already say, and the failure mode when they disagree is a job that is dispatched and never runs — silent, because `Job.dispatch` has no way to know the class was never registered.

**Ask:** discover `Job` subclasses from `app/jobs` and let `jobs` be optional, the way `Kernel.models` could not be but this can — a job list is not an authorization surface, so the argument in #316 (a convention-based scan can pass on an app that declares nothing) does not carry the same weight here.

Keeping the explicit list working matters: an app with jobs somewhere else, or one that wants a subset registered, still needs it. Discovery as the default with the array as the override is the shape.

Worth deciding alongside #323 (the same request for `app/cron`), since whatever mechanism reads a directory of classes at boot should be one mechanism.

*From the Folio 0.51 migration (5 jobs, 82 models).*

nstfkc:
Closes #322, #323 and #324. The first two are one mechanism, which is what #322 asked for, so they share a walk.

## What changed

**Discovery.** `app/jobs` and `app/cron` are read at boot. A `Job` or `CronJob` subclass is registered by existing, and there is no second list to keep in step.

The override rule is **declared wins, and `[]` is declared**. A `jobs` that is present is used verbatim and no directory is read; leaving the key out — or leaving the slice out entirely — is what asks for discovery. `jobsDir` moves the directory without giving up the walk.

That distinction is load-bearing and easy to lose: `withDefaults` deliberately treats `undefined` as absent, so both providers read the raw slice before defaults are applied. The template previously shipped `jobs: []` in `app/config/queue.ts`; left alone, that would have silently disabled discovery in every scaffolded app. Both template config files now omit the key and say why.

**The resolved set stayed readable.** #323 flagged that an app exports `jobs` from its config module so a test can walk what the scheduler runs. `app(Scheduler).jobs` and `app(QueueManager).registeredJobs` are where that test goes now, plus `discoverCronJobs()` / `discoverJobs()` for asking without an application around.

**The gate.** `CronJob.shouldRun()` is evaluated once per tick, before anything else, and `false` skips `onTick`, `callback` and `onComplete` together — which is the whole point, since a guard inside `callback` leaves an outward-reporting job still announcing that it started and finished. A gate that **throws skips the tick** and logs: cron work recurs, so a missed tick is recovered by the next one, and an email sent from the wrong machine is not.

## Two failures found while reviewing this, fixed here

Both were reachable before this branch, and both are the kind of silence these issues are about:

- **Two cron jobs claiming one name collapsed silently.** The second registration hit `registry.get(name).stop()` and killed the *first* job's handle on its way past, while `jobs` still listed both — so a report stopped arriving and a test walking the schedule saw two healthy entries. The first now keeps the name; the second is refused with a message naming both classes.
- **`QueueManager.next()` recursed forever on an unresolvable name.** The head was only deleted on the branch that resolved it, so an unknown name left `queue.size` unchanged and `next()` re-entered on the same entry until the stack gave out. It is now dropped and logged. Four passages in this branch had asserted the benign reading; they were wrong and are corrected.

## Decisions worth disagreeing with

**A shared base class must live outside the walked directory.** `abstract` is erased before the scheduler runs, so a base extending `CronJob` in `app/cron` is found, constructed, and rejected for having no `name` — on every boot, and it lands in `scheduler.jobs`. This is exactly the shape #324 makes people write, so it matters.

I considered having discovery drop any class that is another discovered class's prototype. I did not, because a *concrete* job that is also extended would then stop being scheduled with nothing said — a new silent failure, which is the thing being removed. Instead the docs put the base outside `app/cron` and explain why, and the scheduler's rejection now names the class and says what it probably is. Open to the other call if you'd rather the pattern work in place.

**Discovery evaluates every file it walks.** There is no way to find a class nobody imported without running its module, so a file that does work on import does it at boot. Tests, type tests, benchmarks, `.d.ts` and directories with their own `package.json` are skipped; nothing else is guessed at. Documented on both pages.

**`GEMI_PROJECT_DIR`, not `ROOT_DIR`.** Provider `boot()` runs inside `waitForBoot()`, which finishes before `httpProd`/`httpDev` set `ROOT_DIR`, so discovery computes the root the way `httpProd` does. `httpProd` now delegates to the same function.

## Verification

`bun --bun vitest run` in `packages/gemi`: 115 files, 2652 passed, 1 skipped. Template: 20 files, 660 passed. `bun run typecheck` clean; `oxlint` 0 errors (79 warnings, all pre-existing and in untouched files). Test files were typechecked separately, since `packages/gemi/tsconfig.json` excludes `**/*.test.ts`.

I mutation-checked the three assertions this stands on rather than trusting green: reverting the duplicate-name refusal, the unconditional queue delete, and the `projectRoot()` join each fails a test (the second reproduces the stack overflow). The last one mattered — two tests booted `{}` and asserted `[]`, which passed whether discovery ran or refused, and `resolveDir`'s relative branch had no coverage at all. Both now assert against a fixture reached through the default relative path.

`docs/llms-full.txt` was verified line-for-line against the `cron.md` and `jobs-and-queues.md` regions, not assumed.

## Known, not fixed

`worker = true` jobs are broken independently of this: `QueueManager.createWorker()` imports `app/bootstrap.ts`, which was deleted in a4df8fc, and nothing emits `dist/server/bootstrap.mjs`. `runInWorker` also sets no `onerror`, so the promise never settles. Out of scope here, but if that path is ever revived, note that discovered jobs resolve in `boot()` and a worker that skips phase two would not see them.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,7 +1,9 @@
 # Upgrading from 0.50 to 0.51
 
-Two changes need a hand. There is no codemod for either — `bunx gemi migrate` is
-the 0.42→0.43 tool and does not touch any of this.
+Three changes need a hand, and the third is a look rather than an edit — it only
+becomes work if you were using an unlisted job as an off switch. There is no
+codemod for any of them — `bunx gemi migrate` is the 0.42→0.43 tool and does not
+touch any of this.
 
 ## Declare your model modules on the Kernel
 
@@ -96,6 +98,37 @@ passed `Prisma.DbNull`, `Prisma.JsonNull` or `Prisma.AnyNull`, import them from
 
 The full detail, including the `Prisma.*` type mapping, is under
 **Setup** in [docs/orm.md](./docs/orm.md#setup).
+
+## Check `app/cron` and `app/jobs` before you drop the explicit list
+
+0.51 discovers jobs from the filesystem. Every `Job` subclass under `app/jobs` is
+registered and every `CronJob` under `app/cron` is scheduled — unless the config
+slice declares `jobs` itself, which still wins and still reads no directory.
+
+**Nothing to do if your `app/config/queue.ts` and `app/config/schedule.ts`
+already declare `jobs`.** That includes `jobs: []`, which every app scaffolded
+before 0.51 has in `app/config/queue.ts`: an empty array is an application saying
+it has no jobs, it is honoured as such, and nothing starts running under you.
+Delete the key when you want the directory read instead.
+
+Two things to look at before you do:
+
+- **A job you switched off by unlisting it.** Deleting a class from the array and
+  leaving the file in place used to disable it. Discovery finds the file, so it
+  starts running on the next boot. Delete the file, or keep the explicit list.
+- **Anything in those directories that is not a declaration.** Finding the
+  classes means importing the files — a class does not exist until its module has
+  run — so a helper sitting in `app/cron` that opens a connection or seeds a
+  cache at the top level now does that at boot, on every start. Move it out, or
+  keep the explicit list and skip the walk.
+
+The walk itself skips `.d.ts` files, tests and benchmarks by their filename
+suffix, dot-directories, `node_modules`, and anything under a directory with its
+own `package.json`. Nothing else is guessed at, so a file it cannot import fails
+the boot naming itself rather than being quietly left out.
+
+Both directories are covered in [docs/cron.md](./docs/cron.md) and
+[docs/jobs-and-queues.md](./docs/jobs-and-queues.md).
 
 ---
 
--- a/docs/cron.md
+++ b/docs/cron.md
@@ -203,6 +203,12 @@ export class SubscriptionAudit extends CronJob {
 
 Both are scheduled. Neither is mentioned anywhere else.
 
+### Two jobs, one name
+
+`name` is the scheduler's key, so only one job can hold it. If two claim the same one, the first is scheduled and the second is refused with both class names on stderr — a directory walk makes this easy to write by accident, since `billing/DailyReport.ts` and `analytics/DailyReport.ts` never have to appear side by side the way two entries in a list would.
+
+Both still appear in `scheduler.jobs`, which reports what the scheduler was handed rather than what it accepted, so a test walking the schedule sees the clash instead of a tidied-up set.
+
 ### What the walk costs
 
 A class does not exist until its module has run, so there is no way to read a directory of classes without importing it. **Every `.ts`/`.tsx` file under `app/cron` is imported at boot** — in development and in production, on every start — and a file that *does something* when it is imported does that thing at boot. A module that opens a connection, seeds a cache, or registers a listener at the top level is doing it before the first request, from a directory nobody thought of as an entry point.
--- a/docs/cron.md
+++ b/docs/cron.md
@@ -1,6 +1,6 @@
 # Cron
 
-Cron jobs run code on a recurring schedule — nightly reports, hourly cleanups, periodic audits — without any incoming request. You define each one as a class extending `CronJob` (from `gemi/services`) and list it in `app/config/schedule.ts`. gemi schedules them in-process using `Bun.cron`.
+Cron jobs run code on a recurring schedule — nightly reports, hourly cleanups, periodic audits — without any incoming request. You define each one as a class extending `CronJob` (from `gemi/services`) in a file under `app/cron/`, and gemi schedules it in-process using `Bun.cron`. Nothing else is needed — the directory is read at boot.
 
 ## Defining a cron job
 
@@ -22,15 +22,7 @@ export class DailyDigest extends CronJob {
 }
 ```
 
-```typescript
-// app/config/schedule.ts
-import { defineScheduleConfig } from "gemi/services";
-import { DailyDigest } from "@/app/cron/DailyDigest";
-
-export default defineScheduleConfig({
-  jobs: [DailyDigest],
-});
-```
+That file is the whole registration — there is no list to add it to. **Registering jobs**, below, covers how the directory is read and how to take registration over yourself.
 
 Fields and hooks on `CronJob`:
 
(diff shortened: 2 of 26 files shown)


nstfkc (commented):
Reviewed at `855a249`. Verified independently in a clean worktree: `vitest run` → 115 files, 2652 passed, 1 skipped; `tsc --noEmit` clean. Both match the claims in the description.

The core design holds. "Declared wins, and `[]` is declared" read from the raw config slice *before* `withDefaults` collapses `undefined` into `[]` is the only place that distinction survives, and both providers do it that way. I checked the boot ordering it depends on: `Application.boot()` awaits each provider's `boot()` sequentially in registration order, `QueueServiceProvider` before `ScheduleServiceProvider`, so the newly-`async` boots genuinely complete before `Scheduler.start()` runs. Provider-level coverage is thorough — absent key, `undefined`, `[]`, no slice at all, and missing directory are each pinned for both subsystems.

## One real defect

**Two `Job` subclasses with the same class name collide silently.** `QueueManager.useJobs` builds the registry as `Object.fromEntries(jobs.map((job) => [job.name, job]))` — last discovered wins, the other vanishes, nothing is logged.

Reproduced against this branch with two `SendEmail` classes under `app/jobs/auth/` and `app/jobs/billing/`:

```
discovered: [ "SendEmail", "SendEmail" ]   count: 2
registry entries: 1 for 2 discovered classes
dispatching the AUTH class; watch which one runs:
BILLING SendEmail ran
```

Dispatching one class runs the other's code — worse than the dropped dispatch this PR sets out to fix, because the dispatch appears to succeed and the wrong work happens. Discovery is what makes it reachable: with a hand-written list, two same-named classes force an import alias in one visible file; with a directory walk, `auth/SendEmail.ts` and `billing/SendEmail.ts` are an entirely ordinary thing to write.

The asymmetry is the tell — the cron side gains exactly this guard in this same diff (`claimed` in `Scheduler.start`: the first keeps the name, the second is refused with both class names in the message). Mirroring it in `useJobs` — first wins, second refused out loud — is the fix. Worth noting it reverses the current precedence (today the last one wins), so it is a behaviour change rather than just an added warning. `docs/jobs-and-queues.md` says nothing about collisions either; the cron page's equivalent paragraph would port over.

## Minor

- `Scheduler.jobs` and `QueueManager.registeredJobs` hand back the live internal arrays behind a `ReadonlyArray` type. These are advertised as the thing a test walks, so a `[...]` copy is cheap insurance.
- Curiosity rather than a defect: a file *inside* a walked directory that imports the discovery module deadlocks on the circular top-level await. Hit while writing the repro above; no real app would.

## Checked and clean

The `next()` delete-before-dispatch fix is correct and its test does fail when reverted. `warnIfSourceIsMissing`'s parent-directory test distinguishes "no jobs yet" from "no source shipped" properly. `check-models` still stats its directory before calling the now-non-throwing `sourceFiles`. `projectRoot()` reproduces what `httpProd` computed inline. The docs match the code on the details I spot-checked — `onTick` running sequentially before `callback`, `onComplete` running whether `callback` threw or not, and `shouldRun` throwing being the one caught error that also skips the tick.

nstfkc:
Both taken, in `edd0707`. The defect was real — I reproduced it against the branch before changing anything, and it also makes me distrust my own review pass: three independent verifier agents had refuted exactly this claim, and all three were wrong.

**The collision.** `useJobs` now builds the registry one entry at a time: first claim wins, second is refused with both the name and what happened. You're right that it reverses precedence rather than only adding a warning, and the commit message says so. Reverting the guard fails the new test, which asserts the *behaviour* rather than the log line — `dispatchJob("SendEmail")` returns `"auth"`, and returned `"billing"` before.

Both classes still appear in `registeredJobs`, matching what `Scheduler.jobs` does with a duplicate cron name, so a test walking either sees the clash instead of a tidied set. Neither page documented collisions, so both got a section — not just the queue one.

**Chasing it turned up a worse one, and it is mine.** A job that does not declare `static name` is registered under its class name, and a production build renames it:

```
BUNDLED name: D          # Bun.build({minify: true}) of a file importing @/app/jobs/TestJob
SOURCE  name: TestJob    # await import() of the same file, which is what discovery does
```

`gemi build` minifies the server entry and bundles the app code reachable from it, so a controller dispatching `TestJob` enqueues under `D` while discovery registered `TestJob`. The lookup misses and the job is dropped — in production and nowhere else.

**Discovery introduced this.** With an explicit `jobs` array the config module was bundled too, so both halves were mangled identically and agreed; that is why it has never bitten anyone. Splitting registration (source, at runtime) from dispatch (bundle) is what breaks the symmetry. `docs/jobs-and-queues.md` has always called `static name` required, but nothing enforced it and the template's scaffolded job — the example every new app copies — omitted it.

So: `discoverJobs` warns, naming the class and the exact line to add; the template declares it; the docs note explains why it fails only in production. The check is exact rather than a heuristic — a `static name = "..."` class field defines a *writable* own property while the implicit class name does not, which is the only thing separating two classes that report the same string. Cron is unaffected: `CronJob.name` is an instance field holding a string literal, not the class binding.

I'd flag this as the thing most worth a second opinion. The alternative was to leave it undetected and rely on the docs, which is what the state of things was.

**Minor.** Both accessors return `[...]` copies now.

**The deadlock.** Not chasing it — no top-level `await` in either discovery module, so I could not reproduce it from the description, and I agree an app that imports the discovery module from inside a walked directory is not a real shape. Happy to look if you have the repro.

Gate after the changes: 115 files / 2659 passed, 1 skipped; template 660 passed; `tsc --noEmit` clean; `oxlint` 0 errors. `llms-full.txt` re-verified line-for-line against both edited pages.

PR #329, opened 2026-08-05
Support a second named connection, so a slow analytics pool cannot starve the hot path

### LINKED ISSUE #327: Support a second named connection, so a slow analytics pool cannot starve the hot path
`DatabaseManager` holds one connection — `config.url`, one `SQL`. An app that needs a second pool has nowhere to put it.

The case that hit this: a hot path (auth, requests) and an analytics path (dashboard aggregations, digest cron, audits) sharing a database. The analytics queries run 5–8 seconds; the hot path must never wait behind them. After a pool-starvation incident where dashboard load blocked sign-in, the app split them into two pools with different sizes and different timeouts:

| | hot path | analytics |
|---|---|---|
| pool max | 12 | 3 |
| statement timeout | 8s | 45s |
| idle-in-tx timeout | 10s | 60s |

Those are opposite workloads: a value that protects one is the wrong value for the other, which is the whole reason they cannot be one pool with one setting.

**Ask:** named connections — `DB.connection("analytics")`, and a way for a model read to run on one, so `config/database.ts` can define more than one `url`/`options` pair. Prisma has no first-class answer here either (the app runs two `PrismaClient` instances), so anything that makes the second pool declarable rather than instantiated by hand is an improvement on the status quo.

Two things worth deciding alongside it:

- **Transactions cannot span connections**, and should say so loudly rather than silently committing half. `Model.transaction` resolving one handle while a raw statement resolves another is the failure mode to close by construction.
- **Which connection a model uses** is a per-query property, not a per-model one: the same `Subscription` model is read on the hot path during sign-in and swept by the nightly audit.

Concretely, this is what keeps ~30 call sites of an in-progress ORM port on Prisma — not because the ORM cannot express the queries, but because moving them would collapse the two pools into one and undo the isolation the incident put in.

*From the Folio ORM port, on 0.51.0-rc.4.*

nstfkc:
Closes #327.

`DatabaseManager` held one `SQL` built from one `url`. This adds named connections, a per-query way for a model to use one, and a refusal for the case that cannot work.

## The shape

```typescript
// app/config/database.ts — the top level describes the connection called "default"
export default defineDatabaseConfig({
  url: process.env.DATABASE_URL,
  options: { max: 12 },

  connections: {
    analytics: {
      url: process.env.DATABASE_URL,          // the same database, deliberately
      options: { max: 3, idleTimeout: 60 },   // a pool of its own
      slowTransactionThreshold: 60_000,       // and a threshold that suits it
    },
  },
})
```

```ts
const rows = await Subscription.on("analytics").findMany({ where })
const raw  = await DB.connection("analytics").query(sql`select …`)
await DB.connection("analytics").transaction(async () => { … })
```

## The two decisions the issue asked to make alongside it

**Which connection a model uses is per query, not per model.** The same `Subscription` is read on the hot path during sign-in and swept by the nightly audit, so a `connection` on the class would force one of the two to be wrong. `on(name)` hands back the model class with the name bound to it, so all fifteen operations, `transaction` and `save` are there and `select` / `include` narrow exactly as they do on the class — asserted in `select.test-d.ts`, since a later implementation reaching for `any` to bind the name would lose precisely that and nothing else would notice.

The name travels in the ORM's **ambient scope**, not as an argument. A relation read recurses through the *target* model's `$exec` by way of the registry, so an argument would have to be carried by every intermediate that has no interest in it — and the one that forgot would read a nested `include` off the other pool, with no error and usually a table of the same shape to land in. `connections.test.ts` puts a user and an account in *both* files and asserts the include comes back with the right one, so a leak is a wrong row rather than a missing one.

**A transaction cannot span two connections, and it says so loudly.** The single-connection note in `orm/context.ts` predicted this exactly — a `DB.transaction` on B inside a `Model.transaction` on A would savepoint on **A** and hand the callback A's handle, so statements landed in the wrong database with no error. That note is now a check: both doors call `assertConnectionUsable`, and a statement naming another connection raises `CrossConnectionTransactionError` rather than running outside the transaction, where it would stay committed through the rollback. Catching it leaves the transaction usable, so the fix is local. An *unqualified* query inside a named transaction inherits the name and joins it — otherwise `DB.connection("analytics").transaction(() => Digest.create(…))` would refuse its own contents.

## Notable in the implementation

- **The manager is the default connection.** `connection("default")` returns `this`, so `DB.sql`, `db.dialect`, `db.config` and every wrapper of them keep working unchanged. `Connection` carries what is true of one pool at a time — including SQLite's `pragma foreign_keys`, which is per client and so was per connection all along; a second SQLite connection now gets its own.
- **An unknown name raises and lists the configured ones**, never falling back. `on("analitycs")` quietly running on the hot path is the incident the second pool was configured to prevent, arriving weeks later with nothing pointing at the typo.
- `url` and `options` are not inherited from the top level (a connection that borrowed the default's URL by omission would be a second pool created by a typo); `slowTransactionThreshold` is, since it is a diagnostic about holding any pooled connection.
- A config error part-way through construction closes the clients already built, instead of leaving them to surface as an unhandled rejection with no configuration error nearby.
- The dev-only "the ORM does not implement this dialect" warning now runs per connection.
- `Model.$exec` and `Model.transaction` are `async`, so a refusal *rejects* rather than throwing synchronously — the reasoning `DB.query` already records about `.catch()` missing errors.

## Tests

- `packages/gemi/database/connections.test.ts` — the manager: identity, listing, the reserved name, the errors, the per-connection pragma against two real SQLite files, close-all.
- `packages/gemi/orm/context.test.ts` — the scope: inheriting, merging rather than replacing, the savepoint-vs-refusal split, and that the default connection enters no scope at all.
- `templates/saas-starter/app/models/connections.test.ts` (new, 16 cases) — end to end against **two SQLite files**, so a statement on the wrong connection is a row in the wrong file. In production both connections usually hold the same URL, which is exactly what a test cannot observe.
- `select.test-d.ts` — `on(name)` keeps the typed surface.

Three counting Proxies in the template harnesses now bind their delegated methods to the proxy rather than the target: `connection()` answering the default with `this` would otherwise hand back the unwrapped manager, and the round trips would go uncounted. That failure showed up as a count of zero in `relations.many-to-many.test.ts` and is the reason those tests exist.

## Verification

`packages/gemi`: 2624 passed, 1 skipped. Template models: 268 passed, 32 skipped. Type tests: 64 passed. `oxlint` clean on everything touched.

Three template suites could not run here — `differential`, `writes.differential` and `scalar-list` all fail at import with `Cannot find module './prisma-client'`, which needs a generated Prisma client this environment could not produce (`prisma generate` shells out to `npm i @prisma/client`). They were failing that way before this branch, and CI is where they get their say.

https://claude.ai/code/session_01DP1faTmyRut7mBC5tZ95VD

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -43,7 +43,7 @@ See **[Getting Started](./getting-started.md)** for requirements and a first rou
 - **[Navigation](./navigation.md)** — `Link`, `useNavigate`, `useParams`, `useSearchParams`, and the `Redirect` component vs. facade.
 
 ### Data
-- **[ORM](./orm.md)** — the Prisma-typed, gemi-executed query layer: models, relations, ambient transactions, policies, soft deletes.
+- **[ORM](./orm.md)** — the Prisma-typed, gemi-executed query layer: models, relations, ambient transactions, named connections, policies, soft deletes.
 - **[Rows & Entities](./orm-rows-and-entities.md)** — plain objects by default, `track` + `save`, and `wrap`.
 
 ### Auth
--- a/docs/index.html
+++ b/docs/index.html
@@ -109,7 +109,7 @@
 
       <h2>Data</h2>
       <div class="grid">
-        <a class="card" href="orm.md"><span class="t">ORM</span><span class="d">Models, relations, ambient transactions, policies, soft deletes.</span></a>
+        <a class="card" href="orm.md"><span class="t">ORM</span><span class="d">Models, relations, ambient transactions, named connections, policies, soft deletes.</span></a>
         <a class="card" href="orm-rows-and-entities.md"><span class="t">Rows &amp; Entities</span><span class="d">POJOs by default, track + save, wrap.</span></a>
       </div>
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -2871,7 +2871,8 @@ budget one of them used to have. Rebalance as models move across; when the last 
 Prisma client and give gemi the whole budget.
 
 Pointing gemi at a *different* `url` is the other honest answer — a separate pool with its own cap,
-at the cost of two things to configure per environment.
+at the cost of two things to configure per environment. The same arithmetic applies to gemi's own
+second pool: see [Connections](#connections).
 
 </details>
 
@@ -3856,6 +3857,102 @@ savepoint, the Postgres abort described above. Those need a database, and they b
 integration test rather than in a mock. Keep the seam for the control flow and test the transaction
 itself against Postgres.
 
+## Connections
+
+One connection is the default and needs no configuration. A second one is for the case where two
+workloads share a database and want opposite settings from it — a hot path that must never block,
+and an analytics path whose queries legitimately run for seconds:
+
+| | hot path | analytics |
+| --- | --- | --- |
+| pool max | 12 | 3 |
+| statement timeout | 8s | 45s |
+
+A value that protects one of those is the wrong value for the other, which is the whole reason they
+cannot be one pool with one setting. Without a second connection, a dashboard aggregate can drain
+the pool and block sign-in.
+
+The top-level `url` and `options` describe the connection called `default`. Everything else goes
+under `connections`:
+
+```typescript
+// app/config/database.ts
+export default defineDatabaseConfig({
+  url: process.env.DATABASE_URL,
+  options: { max: 12 },
+
+  connections: {
+    analytics: {
+      url: process.env.DATABASE_URL,          // the same database, deliberately
+      options: { max: 3, idleTimeout: 60 },   // a pool of its own
+      slowTransactionThreshold: 60_000,       // and a threshold that suits it
+    },
+  },
+})
+```
+
+The same URL twice is normal and is not a mistake: what differs is the pool, not the database.
+`url` and `options` are never inherited from the top level — a connection that borrowed the
+default's URL by omitting one would be a second pool created by a typo — but
+`slowTransactionThreshold` is, since it is a diagnostic that applies to any pool.
+
+### Naming one, per query
+
+```ts
+const rows = await Subscription.on("analytics").findMany({ where })
+const raw = await DB.connection("analytics").query(sql`select …`)
+```
+
+**Which connection a model uses is a per-query property, not a per-model one.** The same
+`Subscription` is read on the hot path during sign-in and swept by the nightly audit, so a
+`connection` declared on the model class would force one of those two to be wrong. `on` returns the
+model class with the connection bound to it, so the whole typed surface is there and narrows exactly
+as it does on the class itself.
+
+The connection reaches everything that query fans out into — the nested reads of an `include`, the
+extra statements of a nested write, the relation reads a `_count` compiles — because it travels in
+the same ambient scope the transaction does. There is no argument to forget to pass on.
+
+An unknown name raises `UnknownConnectionError` and lists the ones that are configured. It never
+falls back to the default: `Subscription.on("analitycs")` running on the hot path is the exact
+incident the second pool was configured to prevent, and it would arrive weeks later with nothing
+pointing at the typo.
+
+### A transaction cannot span two connections
+
+This is the constraint to design around, and it is enforced rather than documented:
+
+```ts
+await Model.transaction(async () => {
+  await Invoice.create({ data })
+  await Subscription.on("analytics").findMany({})   // CrossConnectionTransactionError
+})
+```
+
+A transaction lives on one reserved connection of one pool. A statement on another pool cannot join
+it and cannot be rolled back with it, so the alternatives to raising are both worse: run it outside
+the transaction and it stays committed when the transaction rolls back; run it on the open handle
+and it goes to the wrong database, which is usually a table of the same shape and no error at all.
+
+Catching the error leaves the transaction usable, so the fix is local — move that query outside the
+transaction, or do the whole unit on one connection.
+
+Inside a transaction, a query that does **not** name a connection joins it, whichever connection it
+was opened on:
+
+```ts
+await DB.connection("analytics").transaction(async () => {
+  await Digest.create({ data })    // on "analytics", inside the transaction
+})
+```
+
+### What a second connection costs
+
+Every connection is a real pool, counted separately against whatever limit the server or the pooler
+enforces. Two pools of 12 is 24 connections, not 12 — so split the budget rather than adding to it,
+the same arithmetic as [running a Prisma client alongside gemi](#setup). `DB.close()` closes all of
+them.
+
 ## Policies
 
 A model carries a **list** of policies. Every member of a policy is optional; a model with none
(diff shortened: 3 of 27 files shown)


nstfkc:
## Review — named connections

Fetched the branch into a scratch worktree and ran it. **`packages/gemi`: 112 files, 2625 passed / 1 skipped. Template `app/models`: 267 passed, 1 failed** — `generated.test.ts` needs a real `prisma generate`, matching the environmental caveat in the description. `oxlint` reports only the two pre-existing warnings in the untouched `facades/Auth.ts`.

The design decisions are the right ones and the reasoning behind them is recorded where the next person will look. Three worth calling out: `connection("default")` returning `this` — which is what keeps `DB.sql`, `db.dialect` and every counting Proxy pointed at the object that actually executes, and the `bind(receiver)` fix in the three harnesses that follows from it is a genuinely subtle catch; the ambient scope rather than a threaded argument, tested against two SQLite files so a leak is a *wrong row* and not a missing one; and the constructor closing already-built clients when a later one's config throws.

---

### 1. `Model.save` ignores the connection a tracked row came from — silent write to the wrong database

`packages/gemi/orm/provenance.ts` (`track`) / `packages/gemi/orm/Model.ts:329` (`save`)

Provenance records `model`, and `save` checks `record.model !== schema.name` loudly. It records no connection, and nothing checks one. Confirmed against two SQLite files:

```ts
const [row] = await User.on("analytics").findMany({}, { track: true });
row.name = "written-by-save";
await User.save(row);
// → analytics: "cold-original"   hot: "written-by-save"
```

No error. The row was read from one database and the update landed in the other — in production, where both connections usually point at the same database with the same id space, that is a wrong row updated rather than a missed one. This is precisely the failure shape the PR refuses everywhere else; it is the one path that carries a database identity in *data* instead of in scope, so the ambient mechanism cannot reach it.

`User.on("analytics").save(row)` works correctly, so the fix is small: stamp `currentConnectionName()` into `Provenance` at `track` time and compare in `save` — either raise, alongside the model-mismatch `UnsupportedQueryError` immediately above it, or route through `this.on(record.connection)`. `Model.wrap` shares the root cause.

This is the one I would fix before merging: a silent cross-database write, reachable through the documented `track` + `save` path, with no test covering it.

---

### 2. `ConnectionQueries.sql` / `.dialect` throw synchronously

`packages/gemi/facades/DB.ts`

`query`, `execute` and `transaction` were all made `async` so a `.catch()` cannot miss a refusal. The two getters cannot follow that rule — `DB.connection("analytics").sql` inside a transaction on `default` throws `CrossConnectionTransactionError` synchronously, and `DB.connection("typo").sql` throws `UnknownConnectionError` synchronously. Unavoidable for a getter, but worth one line in the doc comment so it is not discovered.

Related, and worth a line in `docs/orm.md`: `DB.connection("analytics").sql\`…\`` inside that connection's own transaction returns the *pool*, not the transaction handle, so it runs outside the transaction. Pre-existing `DB.sql` semantics, but a named handle makes it easier to reach for.

### 3. `DB.connection()` caches a handle before validating the name

`packages/gemi/facades/DB.ts:238`

`handles.set(name, …)` runs unconditionally, so `DB.connection(someRequestDerivedString)` grows a module-global `Map` without bound. A handle is a one-field object that resolves everything per call — either skip the cache or validate first.

### 4. Every `on()` query takes the policy-divergence guard's slow path

`packages/gemi/orm/Model.ts:644`

The bound subclass is never what `registry.get(schema.name)` returns, so `registered !== this` always holds and `policiesFor(registered)` plus the array compare run on every query through `on()`. The outcome is correct — the chains are identical, so nothing diverges — but the guard's own comment advertises "one `Map.get` and one reference compare" on the common path, and that is no longer true for named-connection queries. `Object.getPrototypeOf(this) === registered` short-circuits it.

### 5. Connection names are unchecked at compile time

`Model.on(name: string)`, `DB.connection(name: string)`

The PR argues at length that `on("analitycs")` is the incident the second pool exists to prevent — and then catches it only when that query runs, which for a nightly sweep means 3am. A declaration-merged `ConnectionNames` interface, in the style the framework already uses for generated surfaces, would make it a type error. Follow-up issue rather than a blocker.

### 6. Nit — prettier drift

`DatabaseManager.ts`, `DB.ts`, `orm/context.test.ts` and both `connections.test.ts` are not prettier-clean on this branch; on `main` every one of those except `orm/context.ts` was. `prettier --write` touches two over-width `.set(...)` calls, one `if (…) throw`, and the `}, { connection: … })` trailers in the tests. There is also a stray blank line added to `docs/llms-full.txt` around line 157.

---

Findings 2–6 are polish. Finding 1 is the blocker.

nstfkc:
All six addressed in `aa86f36`.

### 1. `save` wrote to the wrong database — fixed, and it was worse than reported

Confirmed and fixed as suggested: `Provenance` carries `connection`, `track` stamps it, `save` writes back to it. The default is `currentConnectionName()`, which is right for `$exec` — it tracks inside the scope it entered — and wrong for `wrap`, which runs after that scope has closed, so `wrap` passes the argument's own connection explicitly.

Naming a different connection is a contradiction rather than an instruction and raises, alongside the model-mismatch check above it, pointing at the `update` that would express a cross-connection copy deliberately.

**Writing the test for it found a second bug, in the mechanism this PR added.** The routing test — save a row read on `analytics` from inside a transaction on `default` — did not raise. `runOnConnection` merged the new name into a scope that still held the *old* connection's transaction handle, producing a store that said "a transaction on analytics" while holding the default connection's handle. Every check downstream then agreed with itself and the statement ran on the wrong connection's transaction: the exact failure this feature exists to prevent, reintroduced by the fix for it.

The refusal now lives in `runOnConnection`, *before* the scope is entered, because that is the only moment at which the handle and the name it belongs to are both still visible. `$exec` and `Model.transaction` no longer check separately. `save` is the one caller that switches connections with a transaction already open, which is why nothing else had found it.

Five cases in the template suite (`describe("save")`), plus the mechanism itself in `context.test.ts`.

### 2. Getters throw where the methods reject — documented

`sql` and `dialect` now say so, and say why: an accessor has nowhere to put a rejection, which is exactly why `query` / `execute` / `transaction` were made `async`. The pool-not-transaction point is documented on `sql` and in `docs/orm.md` — you are right that a named handle makes it easier to reach for, even though the semantics are `DB.sql`'s own.

### 3. Unbounded handle cache — removed

`DB.connection(name)` builds a handle per call. A handle is one string and methods that resolve everything per call, so the `Map` bought an allocation and cost a slow leak on any request-derived name — which an unknown name cannot even short-circuit, since it is not rejected until the handle is used.

### 4. Divergence guard's slow path — short-circuited

`inheritsPoliciesFrom(queried, registered)`: a direct subclass declaring no `$policies` of its own has, by construction, the chain its parent has, because `policiesFor` walks the prototype chain taking each level's own list. Structural rather than a flag on the classes `on` mints, so `class AdminUser extends User {}` — the typed view the guard's comment already argues should be allowed — gets it too instead of resolving the chain twice.

### 5. Typed connection names — #331

Filed with the three questions I think decide it: where the names come from (the config is application code, not a generated artifact), keeping a computed name expressible, and `keyof` on an empty interface degrading to `string` rather than `never`.

### 6. Prettier — fixed, and one correction

`prettier --write` on the files this branch made unclean: `DatabaseManager.ts`, `Model.ts`, `context.test.ts` and both `connections.test.ts`. `DB.ts` was already clean by then.

Left alone: `context.ts`, `dialect.ts`, `provenance.ts` and `DatabaseServiceProvider.ts` are unclean on the base commit too — I checked each against `c374c02` rather than reformatting them, since that would put unrelated churn in this diff.

The `llms-full.txt` blank line was real and is gone: the first mirror pass joined the page body to the `---` separator with an extra newline. Re-mirrored both ORM pages, and the diff against the base is now 135 insertions and 4 deletions, the deletions being two lines this branch rewrote.

---

`packages/gemi`: 2628 passed, 1 skipped. Template `app/models`: 273 passed, 32 skipped, with the three `prisma-client` suites still unrunnable here. Type tests: 64 passed. `oxlint` adds one warning, on the `// @ts-expect-error user.name;` line in `select.test-d.ts` — the same `no-unused-expressions` the five existing `@ts-expect-error` assertions in that file already raise.

PR #330, opened 2026-08-05
Hoist the app CSS into <head> so the first paint is not unstyled

### LINKED ISSUE #328: FOUC on first paint: app CSS injected into <body>, not render-blocking in <head>
## Summary

On the first paint of every SSR page, there is a flash of unstyled content (FOUC): the layout renders unstyled for a frame (blocks collapse to natural/zero height, images to 0), so page content is very short and elements that normally sit at the bottom (e.g. a footer) briefly appear near the top, then snap into place once styles apply.

Root cause: the app's CSS `<style>` tags are rendered into `<body>` (as the first children, before the app root), not into `<head>`, so the browser has **no render-blocking stylesheet** and paints before the CSS is applied.

Version: gemi 0.36.2.

## Where it happens (source)

`packages/gemi/services/router/ViewRouteDispatcher.ts` (render root, ~L399 + L436-460):

```ts
const styles = await getStyles(currentViews);
...
const stream = await renderToReadableStream(
  createElement(Fragment, {
    children: [
      ...styles,                                  // <style> elements
      createElement("script", { key: "theme-script", ... }),
      createElement(Root, { ..., key: "root" }),  // renders <html><Head/><body>
    ],
  }),
  { bootstrapScriptContent, bootstrapModules, ... },
);
```

The `styles` are spread as siblings **before** `Root` inside a top-level `Fragment`. `Root` is what renders `<html>` + `<Head>` + `<body>`, and `Head` renders only meta/title/OpenGraph (no app CSS).

`packages/gemi/server/styles.ts` builds those `<style>` elements without `href`/`precedence`:

```ts
// createStyles (prod)
createElement("style", { key: i, id: style?.id, type: "text/css",
  dangerouslySetInnerHTML: { __html: style.content } });

// createDevStyles (dev) — same shape, adds data-vite-dev-id
createElement("style", { key: i, type: "text/css",
  "data-vite-dev-id": style.id, dangerouslySetInnerHTML: { __html: style.content } });
```

React 19 only hoists a `<style>` (or `<link rel="stylesheet">`) into `<head>` and treats it as render-blocking when it is given a **`precedence`** prop (with `href`/`id` as the dedupe key). Without `precedence`, React leaves the tag where it is rendered; since these tags sit before `<html>` in the Fragment, React normalizes them into `<body>`.

## Evidence (served HTML)

- `<head>` contains meta / preconnect / image preloads only — zero `<link rel="stylesheet">` and zero `<style>`.
- `<body>` begins (right after `<body>`) with a single large inline `<style>` (the whole app CSS, ~175KB in our app).
- Because nothing in `<head>` is render-blocking, the first paint is unstyled → FOUC.

## Related second FOUC path (client navigation)

`packages/gemi/client/ClientRouterContext.tsx` fetches the next route's CSS and appends it to `document.head` asynchronously:

```ts
const result = await Promise.all(cssFiles.map((file) => fetchCSS(file)));
for (const { content, id } of result) {
  const style = document.createElement("style");
  style.id = id;
  style.textContent = await content;
  document.head.appendChild(style);
}
```

Here the CSS does go to `<head>`, but it is appended after the new route renders, so a client-side transition can also paint before its CSS is present.

## Impact

- Affects every gemi-rendered page's initial load (whole-page FOUC, most visible as bottom-of-page content flashing to the top and settling).
- Worse on pages that additionally load web fonts at render time (a font `<link rel="stylesheet">` without `precedence` is also left in `<body>`, adding a font-swap reflow on top).

## Proposed fix

Make the app CSS render-blocking in `<head>`. Options, smallest first:

1. Give the `<style>` elements a `precedence` (and a stable `href`/key) in `server/styles.ts` so React 19 hoists them into `<head>` and coordinates loading, e.g. `createElement("style", { href: style.id, precedence: "default", ... })`. Same idea for `createDevStyles`.
2. Render `getStyles()` output inside the document `<head>` (via `Head`/`Root`) instead of as body-preceding `Fragment` siblings.
3. In prod, emit a `<link rel="stylesheet" href="/assets/<hash>.css" precedence="default">` in `<head>` for the built CSS file (it already exists on disk and is served by the static handler) rather than inlining a `<style>` in `<body>`.

For the client-navigation path, appending the `<link>`/`<style>` to `<head>` and awaiting its load before revealing the new route (or letting React manage it via `precedence`) would remove the transition FOUC too.

nstfkc:
Forward-port of the fix shipped in [v0.49.1](https://github.com/nstfkc/gemi/releases/tag/v0.49.1), which was cut from the `v0.49.0` tag. Without this, 0.51.0 ships the flash back.

## The bug

Every SSR page painted unstyled for a frame before snapping into place — blocks collapsing to their natural height, images to zero, footers flashing near the top.

The app's `<style>` tags are rendered as siblings *before* `Root` (which renders `<html>`), and without a `precedence`. React 19 only lifts a style into `<head>` — and only treats it as render-blocking — when it carries both `precedence` and `href`. Without them React leaves the tag where it was rendered, which normalizes into `<body>`: nothing in `<head>` blocks the paint.

## The change

- `server/styles.ts` — both the dev and prod builders now emit `href` + `precedence`, so React hoists them. One shared precedence keeps them in a single tag in insertion order: app bundle first, then the route's CSS layered over it.
- `server/httpProd.ts` — the app bundle stops being one concatenated blob. Each file needs its own stable id to be hoisted and deduped by.
- `client/ClientRouterContext.tsx` — React strips the `id` off anything it hoists and records the merged files in a space-separated `data-href`. The "do I already have this?" check reads that shape too; without it every navigation re-fetches and re-appends the CSS the page was already served with.
- `client/ClientRouter.tsx` — the second flash: a route's CSS was fetched but never awaited, so a new surface could mount ahead of its stylesheet. The fetch still starts alongside the payload request; the transition now waits for it.

## Verification

Two new test files:

- `server/styles.test.ts` renders through the real document shape — styles as siblings before `<html>` — and asserts the CSS lands in `<head>` and not `<body>`, that order is preserved, and that `data-href` lists every merged file. Dropping `href` again fails all three.
- `client/hoistedStyles.hydration.test.tsx` hydrates that document with the empty-slot shape `init` uses, and asserts no recoverable errors and that nothing evicts the stylesheet the client never renders.

Full suite green (117 files, 2663 tests).

## Note for anyone reading served markup

In dev, `data-vite-dev-id` no longer survives to the HTML — React strips it along with every other attribute on a hoisted style. Vite injects its own tag for HMR rather than adopting the server-rendered one, so hot updates still apply; dev just carries the stylesheet twice.

Closes #328

### DIFF
--- a/packages/gemi/client/ClientRouter.tsx
+++ b/packages/gemi/client/ClientRouter.tsx
@@ -292,7 +292,13 @@ const Routes = (props: { componentTree: ComponentTree }) => {
 
       // `fetchRouteCSS` keys off the route manifest, so it needs the pattern
       // rather than the concrete path — `/posts/:id`, not `/posts/123`.
-      fetchRouteCSS(routerState.routePath).catch((e) => console.error(e));
+      // Started here so it runs alongside the payload fetch, but awaited
+      // before the transition commits: a surface that mounts ahead of its own
+      // stylesheet paints unstyled first (#328). Swallowing the rejection is
+      // what keeps a CSS failure from stranding the navigation.
+      const cssReady = fetchRouteCSS(routerState.routePath).catch((e) =>
+        console.error(e),
+      );
       // Through `loadViewModule` so the module registry — and with it each
       // view's `Loading`/`Error` exports — is populated before the
       // transition commits the new surface.
@@ -312,6 +318,8 @@ const Routes = (props: { componentTree: ComponentTree }) => {
           hydrate({ [path]: { [variantKey]: data } });
         },
       });
+
+      await cssReady;
 
       if (payload) {
         const {
--- a/packages/gemi/client/ClientRouterContext.tsx
+++ b/packages/gemi/client/ClientRouterContext.tsx
@@ -77,6 +77,25 @@ interface ClientRouterProviderProps {
   breadcrumbs: Record<string, Breadcrumb>;
 }
 
+/**
+ * Whether `file`'s CSS is already on the page, in either of the two shapes it
+ * can take. A client navigation appends `<style id={file}>`; the styles the
+ * document was served with are hoisted into `<head>` by React, which strips
+ * their `id` and records the files it merged into one space-separated
+ * `data-href` list instead (#328). Missing the second shape means re-fetching
+ * and re-appending the CSS the page already has on every navigation.
+ */
+function isStyleOnPage(file: string) {
+  if (document.getElementById(file)) {
+    return true;
+  }
+  // `Array.from` rather than `for...of`: the browser build compiles without
+  // `DOM.Iterable`, so a `NodeList` is only an ArrayLike there.
+  return Array.from(document.querySelectorAll("style[data-href]")).some(
+    (style) => style.getAttribute("data-href").split(" ").includes(file),
+  );
+}
+
 export const ClientRouterProvider = (
   props: PropsWithChildren<ClientRouterProviderProps>,
 ) => {
@@ -260,7 +279,7 @@ export const ClientRouterProvider = (
         return cssManifest?.[view];
       })
       .filter(Boolean)
-      .filter((file) => !document.getElementById(file));
+      .filter((file) => !isStyleOnPage(file));
 
     if (cssFiles.length === 0) {
       return;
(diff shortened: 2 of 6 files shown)


nstfkc:
Not merging — 0.51 will handle this differently. The fix is released in 0.49.1; the branch `fix/328-fouc-main` is kept if it's wanted later.

PR #332, opened 2026-08-05
Keep a relation's nested relations when it carries arguments

### LINKED ISSUE #326: A relation carrying orderBy/where loses its nested relations in the payload type
A to-many relation that carries `orderBy` (or `where` / `take` / `skip`) stops resolving its own nested relations in the payload type. The row comes back as the scalar type, so reading the nested relation is a compile error even though the query returns it at runtime.

Reproduced in `templates/saas-starter` on `main`, `bunx tsc --noEmit`:

```ts
import { UserModel as User } from "./models/generated";

// depth 2, no args on the outer relation — infers correctly
async function ok() {
  const r = await User.findFirst({
    select: { accounts: { select: { organization: { select: { id: true } } } } },
  });
  return r!.accounts[0].organization.id;   // fine
}

// the same shape, with an orderBy on the outer relation
async function broken() {
  const r = await User.findFirst({
    select: {
      accounts: {
        orderBy: { id: "asc" },
        select: { organization: { select: { id: true } } },
      },
    },
  });
  return r!.accounts[0].organization.id;
}
```

```
error TS2551: Property 'organization' does not exist on type 'AccountScalars'.
  Did you mean 'organizationId'?
```

## What narrows it

- **The args are accepted.** `RelationArgs` admits `where`/`orderBy`/`take`/`skip` on a to-many, and the object type-checks — only the *payload* resolution falls back.
- **One level is fine.** `accounts: { orderBy, select: { id: true } }` — scalars only in the inner select — infers.
- **Two levels without args is fine.** The first function above.
- **`include` behaves identically**, so there is no spelling of this that types.

So the trigger is precisely: relation args + a nested relation underneath them.

## Why it matters more than it looks

This is the ordinary shape of a list read. In the app I hit it in — 82 models, porting ~780 Prisma call sites onto the ORM — it accounts for **56 of the type errors**, all of them variants of "the products of this list, ordered by position, each with its product row" or "the memberships of this org, excluding removed ones, each with its user".

The workaround is a cast, which is the wrong shape of fix: it discards the payload type for exactly the reads whose nesting made the type worth having. So the port is parked on this rather than carrying 56 `as` assertions.

*From the Folio ORM port, on 0.51.0-rc.4.*

nstfkc:
Closes #326.

## What was wrong

A to-many carrying `orderBy` / `where` / `take` / `skip` stopped resolving its own nested relations. The relation came back as the target's bare scalars, so reading anything underneath was a compile error even though the query returned it at runtime — and `include` spelled it identically, so no filtered, ordered list read typed at all.

Reproduced on `main` before touching anything, both the template shape from the issue body and the `List.products` shape from your comment.

## The cause

`Payload` asked `A extends { select: infer S }`.

Arguments reach it through `Subset`, the excess-property guard the generated signatures wrap them in. Inferring `T` through that mapped type leaves a relation's arguments in a shape where these two disagree:

```ts
declare const acc: AccWith;
const ok: { select: unknown } = acc;                        // assignable — no error
type Q = AccWith extends { select: unknown } ? "y" : "n";   // "n"
```

So the conditional took its else branch and returned the default selection. Only relations *carrying arguments* went through that path, which is why one level, or two levels without arguments, always looked fine.

The fix asks whether the key is there and then indexes — `"select" extends keyof A` plus `A["select"]`, which resolves in exactly the cases the `extends` did not. Same for `include`, `omit` and `_count`. The `[…] extends [undefined]` guards were already present for an explicitly-`undefined` key and stay.

## Two alternatives, measured rather than assumed

Both fix the payload; both cost more than this does.

| | #326 fixed | rejects `nonExistentColumn` | rejects `orderBey` in relation args |
|---|---|---|---|
| drop `Subset` | ✅ | ✅ | ❌ |
| `T & Subset<T, U>` | ✅ | ✅ | ❌ |
| **this** | ✅ | ✅ | ✅ |

`Subset` is the only thing that catches a typo'd key *inside* a relation's arguments, so both alternatives trade one silent failure for another.

## Compile time on a cyclic schema

`types.ts`'s header warns that a conditional resolving a whole shape "would turn a cyclic schema into an instantiation-depth error rather than a slow compile", and this moves four conditionals from inference to indexing — so that is the first thing to ask of it.

Measured by @nstfkc against Folio (82 models, cycles throughout), two runs each:

| | run 1 | run 2 |
|---|---|---|
| rc.5 | 25.6s | 25.4s |
| rc.5 + this | 24.0s | 23.8s |

No instantiation-depth error, and slightly faster if anything. Full monorepo type-check across three apps and `turbo run test` both pass under the patch.

## Verification

- **Five new type tests** in `select.test-d.ts`, covering select + `orderBy`, include + `orderBy`, `where`/`take`/`skip`, arguments three levels deep, and that the narrowing is still a narrowing. All five fail without the fix (`Type Errors 5 failed`) and pass with it.
- Template `tsc --noEmit`: **277 errors before, 277 after** — the fix introduces none and the remaining ones are the pre-existing backlog.
- `packages/gemi`: 116 files / 2695 tests, `tsc --noEmit` clean.
- Template: 681 tests, type tests 69 passing, no type errors.

## Note for the port

This does not touch runtime behaviour — the rows were always correct, only the type fell back — so nothing needs regenerating and no query changes.

### DIFF
--- a/packages/gemi/orm/types.ts
+++ b/packages/gemi/orm/types.ts
@@ -58,11 +58,7 @@
 // in a conditional that has to resolve the whole shape — would turn a cyclic
 // schema into an instantiation-depth error rather than a slow compile.
 
-import type {
-  AnyNullValue,
-  DbNullValue,
-  JsonNullValue,
-} from "./json-null";
+import type { AnyNullValue, DbNullValue, JsonNullValue } from "./json-null";
 
 /**
  * Any value a `Json` column can hold, on either dialect.
@@ -94,10 +90,7 @@ export type JsonValue =
  * `metadata: null` compiled while `docs/orm.md` said in this same change that
  * it does not.
  */
-export type JsonInput =
-  | Exclude<JsonValue, null>
-  | DbNullValue
-  | JsonNullValue;
+export type JsonInput = Exclude<JsonValue, null> | DbNullValue | JsonNullValue;
 
 /**
  * Is this column a `Json` one?
@@ -202,10 +195,8 @@ type RelationPayload<R extends RelationInfo, A> = R["kind"] extends "many"
  * `_count` inside a `select` or an `include`: the number of related rows, per
  * relation named.
  */
-type RelationCountPayload<M extends ModelTypeInfo, A> = A extends {
-  select: infer S;
-}
-  ? { [K in Requested<S> & keyof Relations<M>]: number }
+type RelationCountPayload<M extends ModelTypeInfo, A> = "select" extends keyof A
+  ? { [K in Requested<NonNullable<A["select"]>> & keyof Relations<M>]: number }
   : { [K in keyof Relations<M>]: number };
 
 type SelectPayload<M extends ModelTypeInfo, S> = {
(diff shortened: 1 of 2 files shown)


nstfkc (commented):
Verified against Folio's own generated models — 82 models, heavily cyclic — rather than re-running the template. **The fix lands the case it closes, and I'd take it.**

## #326 is fixed, on the real schema

Patched the four sites into `node_modules/gemi/dist/orm/types.d.ts` on rc.5 and ran the shape from my comment on the issue, plus a level of nesting it did not have:

```ts
const bySelect = await List.findUnique({
  where: { publicId: "x" },
  select: {
    products: {
      where: { product: { deletedAt: null } },
      orderBy: { position: "asc" },
      select: { product: { select: { publicId: true, name: true, sku: true } } },
    },
    StoreOnLists: {
      select: {
        store: {
          select: {
            orgsSharedWith: { where: { sharedWithId: "o" }, select: { id: true } },
          },
        },
      },
    },
  },
});

const a: string = bySelect!.products[0].product.name;                        // ok
const b: number = bySelect!.StoreOnLists[0].store.orgsSharedWith[0].id;      // ok

// @ts-expect-error `visibility` was not selected on the nested product
bySelect!.products[0].product.visibility;
// @ts-expect-error `position` was not selected on the join row itself
bySelect!.products[0].position;
```

`tsc --noEmit` clean, and clean *including* both `@ts-expect-error`s — so the narrowing is still a narrowing three levels down, not widened away to make the error go. The `include` spelling passes identically.

## The data point the PR body is missing

`types.ts`'s own header warns that a conditional resolving a whole shape "would turn a cyclic schema into an instantiation-depth error rather than a slow compile", and the fix moves four conditionals from inference to indexing. Folio is the stress case for that — 82 models, cycles everywhere — so I measured it, two runs each on `apps/web`:

| | run 1 | run 2 |
|---|---|---|
| rc.5 | 25.6s | 25.4s |
| rc.5 + this | 24.0s | 23.8s |

No instantiation-depth error, and if anything slightly faster. Full monorepo `type-check` (3 apps) and `turbo run test` (2622 web + 50 store) both pass under the patch. Worth putting in the PR body — it is the risk a reader of this diff will ask about first.

## What it actually unblocks

`CustomerListsV2Controller` — the file behind my repro — ports onto the ORM **completely** under this fix. That was the file I gave up on. One exception, below.

## Adjacent gap found while verifying: a filtered `_count` cannot be spelled

Not this PR's to fix, and not #326, but it is the last thing standing between that controller and the port, so it is worth knowing about now:

```
error TS2322: Type '{ where: { product: { deletedAt: null; }; }; }'
              is not assignable to type 'boolean | undefined'.
```

from

```ts
_count: { select: { products: { where: { product: { deletedAt: null } } } } },
```

`types.ts:439` and `:445` declare it as

```ts
_count?: boolean | { select?: { [K in keyof Relations<M>]?: boolean } };
```

**but the runtime already supports it.** `compile/relation-count.ts:116-131` goes looking for exactly that key:

```ts
const locate = (a: any) => a?.[container]?.[COUNT_KEY]?.select?.[relation.name]?.where;
…
if (filter) parts.push(sql(" and "), filter);
```

So the correlated subquery would AND the filter in correctly; the type is what refuses to let anyone write it. Same class of bug as #326 — the type not describing what the compiler does — which is presumably why this PR's fourth site, `RelationCountPayload`, needed the same treatment as the other three. The payload half is now right and the input half is a `boolean`.

For Folio this one is not cosmetic: that `where` is what stops soft-deleted products being counted on a list card, so without it the number rendered to the user is wrong rather than merely untyped. Happy to file it with the repro if you want it separate.

## On the diff itself

The reasoning is right and the comment earns its length — the assignability-vs-`extends` disagreement under `Subset` is genuinely surprising, and `Q = AccWith extends { select: unknown } ? "y" : "n"` being `"n"` while the assignment on the line above type-checks is the thing a future reader will not reconstruct on their own. Keeping it next to the fix is correct.

The alternatives table is the part I'd have gotten wrong. Dropping `Subset` looks like the smaller change and reads as strictly simpler; the third column is the reason it isn't, and measuring it beats arguing about it.

One nit: roughly two-thirds of the 52 deleted lines are formatter reflow unrelated to the fix — `JsonInput`, `OrderingFilter`, `StringFilter`, `RelationFilter`'s comment indent, `NestedCreate`, `NestedUpdate`, `GroupByArgs`, and the `findMany` call in `select.test-d.ts`. It makes the four real hunks harder to find in a diff whose whole point is four hunks. Not worth another push if this is going in as-is, but a separate formatting commit would have made this a ~15-line review.

nstfkc:
All three addressed.

**Compile time is now in the PR body**, as its own section above Verification, with your table and the note about `types.ts`'s own header warning. You're right that it is the first question a reader of this diff has, and it was missing.

**The reflow nit is fixed rather than waved through** — `876ce0d`. You said it wasn't worth another push, but the measurement made it cheap: I restored `types.ts` from `origin/main` and reapplied the four hunks by hand instead of running prettier over a file that was not prettier-clean to begin with. The diff is now two hunks in `types.ts` and a pure append in the type test; **deletions are down from 52 to 15, and all fifteen are the type definitions actually being replaced**. Verified the only difference from the reviewed revision is whitespace, and re-ran the whole gate: 2695 tests, `tsc --noEmit` clean, type tests green, template still 277.

**The `_count` gap is filed and fixed** — #333, PR #334. I checked the runtime claim rather than taking it from the code read, and it is stronger than you put it: `relation-count.test.ts` already asserts both the emitted SQL *and* that the filter's value is a parameter rather than text. So the compiler was tested for a feature the input type refused to let anyone write. Type-only change; both sites now share one `CountSelection` instead of the two copies that let them drift.

I kept it out of this PR and out of the rc — separate bug, and this PR's review asked for tighter diffs, so widening it would have been the wrong lesson to draw.

Merging this and cutting the rc now.

PR #334, opened 2026-08-05
Let a `_count` carry the filter the compiler already applies

### LINKED ISSUE #333: A filtered _count cannot be spelled, though the compiler already emits it
`_count` accepts `boolean` per relation, so a filtered count does not type:

```ts
_count: { select: { products: { where: { product: { deletedAt: null } } } } },
```

```
error TS2322: Type '{ where: { product: { deletedAt: null; }; }; }'
              is not assignable to type 'boolean | undefined'.
```

`types.ts` declares it in two places — `SelectInput` and `IncludeInput`:

```ts
_count?: boolean | { select?: { [K in keyof Relations<M>]?: boolean } };
```

**The runtime already supports it.** `compile/relation-count.ts` goes looking for exactly that key and ANDs it into the correlated subquery:

```ts
const locate = (a: any) =>
  a?.[container]?.[COUNT_KEY]?.select?.[relation.name]?.where;
…
if (filter) parts.push(sql(" and "), filter);
```

So the SQL is right and the type is what refuses to let anyone write it — the same class of divergence as #326, the type not describing what the compiler does, and the reason `RelationCountPayload` needed the same treatment as the other three sites in #332.

## Why it is not cosmetic

Found while verifying #332 against the Folio port. The `where` is what stops soft-deleted products being counted on a list card, so without it the number rendered to a user is *wrong*, not merely untyped — and it is the last thing between `CustomerListsV2Controller` and a complete port onto the ORM.

The payload half is already correct: `Requested` treats `{ where: … }` as requested, so a relation counted with a filter still resolves to `number`.

*Reported by @nstfkc on #332, from the Folio ORM port on 0.51.0-rc.5.*

nstfkc:
Closes #333 and #335. Found by @nstfkc while verifying #332 against the Folio port.

**Not in the rc cut from #332** — separate on purpose, since #332's review asked for tighter diffs and this is a different bug.

## The gap

```ts
_count: { select: { products: { where: { product: { deletedAt: null } } } } },
// error TS2322: Type '{ where: … }' is not assignable to type 'boolean | undefined'.
```

## The SQL was always right

`compileRelationCount` reaches for `_count.select.<relation>.where` and ANDs it into the correlated subquery — and `relation-count.test.ts` has asserted the emitted text *and* its parameterisation for as long as the feature has existed:

```
, (select count(*) from "Account" as "_c0"
   where "_c0"."userId" = "User"."id" and "_c0"."organizationRole" = ?) as "_count.accounts"
```

So this is a type-only change. The input type was the sole thing refusing the argument the compiler was already looking for — the same divergence as #326: the type not describing what the compiler does.

## Why it isn't cosmetic

The wrong answer is silent in a way a missing column is not. A count that drops its filter renders a number that is *merely wrong* — the case that found it counts a list's products excluding soft-deleted ones, and without the filter the figure shown to a user includes them.

## The change

Both declaration sites now share one `CountSelection` instead of the two copies that let `SelectInput` and `IncludeInput` drift apart unnoticed.

Four type tests: filtered, the `include` spelling, `true` mixing with a filtered sibling, and that the filter is checked against the *counted* model rather than its parent.

## Verification

`packages/gemi` 2695 tests and `tsc --noEmit` clean; template type tests green; template `tsc` error count unchanged at 277.

### DIFF
--- a/packages/gemi/orm/compile/relation-count.test.ts
+++ b/packages/gemi/orm/compile/relation-count.test.ts
@@ -71,6 +71,35 @@ describe("the projected subquery", () => {
     );
   });
 
+  /**
+   * The filter goes through `compileWhere`, so it is a whole `where` and not a
+   * column list — a relation filter inside it nests an `exists` correlated to
+   * the count's own alias rather than to the outer table.
+   *
+   * Worth its own test because it is the shape #335 reported and the one every
+   * other test here misses: they all filter a scalar, which never exercises the
+   * qualifier the correlated alias has to supply. Get that wrong and the inner
+   * `exists` correlates to `"User"` instead of `"_c0"` — a count that silently
+   * answers a different question rather than failing.
+   */
+  test("the filter can reach through a relation of the counted model", () => {
+    expect(
+      projection({
+        include: {
+          _count: {
+            select: { accounts: { where: { organization: { name: "acme" } } } },
+          },
+        },
+      }),
+    ).toBe(
+      `, (select count(*) from "Account" as "_c0" ` +
+        `where "_c0"."userId" = "User"."id" and exists ` +
+        `(select 1 from "Organization" as "_r0" ` +
+        `where "_r0"."id" = "_c0"."organizationId" and "_r0"."name" = ?)) ` +
+        `as "_count.accounts"`,
+    );
+  });
+
   test("the filter's value is a parameter, not text", () => {
     const compiled = plan({
       include: {
--- a/packages/gemi/orm/types.ts
+++ b/packages/gemi/orm/types.ts
@@ -431,18 +431,37 @@ export type OrderByInput<M extends ModelTypeInfo> = {
  * is mutually recursive with the operation args — bounded, as ever, by what the
  * caller wrote.
  */
+/**
+ * What `_count` accepts per relation: `true`, or a filter over the rows counted.
+ *
+ * The filter half was emitted long before it could be written. `compileRelationCount`
+ * reaches for `_count.select.<relation>.where` and ANDs it into the correlated
+ * subquery, so the SQL for a filtered count has always been correct; this type
+ * said `boolean` and was the only thing refusing it (#333). The gap mattered
+ * because the wrong answer is silent — a card counting soft-deleted rows renders
+ * a number that is simply wrong, not one that fails.
+ *
+ * Shared by `SelectInput` and `IncludeInput` rather than written twice, because
+ * the two spellings of the same argument disagreeing is how it went unnoticed.
+ */
+type CountSelection<M extends ModelTypeInfo> = {
+  [K in keyof Relations<M>]?:
+    | boolean
+    | { where?: WhereInput<Relations<M>[K]["target"]> };
+};
+
 export type SelectInput<M extends ModelTypeInfo> = {
   [K in keyof Scalars<M>]?: boolean;
 } & {
   [K in keyof Relations<M>]?: boolean | RelationArgs<Relations<M>[K]>;
 } & {
-  _count?: boolean | { select?: { [K in keyof Relations<M>]?: boolean } };
+  _count?: boolean | { select?: CountSelection<M> };
 };
 
 export type IncludeInput<M extends ModelTypeInfo> = {
   [K in keyof Relations<M>]?: boolean | RelationArgs<Relations<M>[K]>;
 } & {
-  _count?: boolean | { select?: { [K in keyof Relations<M>]?: boolean } };
+  _count?: boolean | { select?: CountSelection<M> };
 };
 
 export type OmitInput<M extends ModelTypeInfo> = {
--- a/packages/gemi/orm/types.ts
+++ b/packages/gemi/orm/types.ts
@@ -454,11 +454,6 @@ export type OrderByInput<M extends ModelTypeInfo> = {
 };
 
 /**
- * `select`, one level. Relations carry their own nested arguments, so the type
- * is mutually recursive with the operation args — bounded, as ever, by what the
- * caller wrote.
- */
-/**
  * What `_count` accepts per relation: `true`, or a filter over the rows counted.
  *
  * The filter half was emitted long before it could be written. `compileRelationCount`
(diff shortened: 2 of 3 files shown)


nstfkc:
Also closes #335, which is this bug again — filed independently while verifying #332 against the Folio port.

Worth more than a duplicate-close, though, because #335's repro is a shape none of the tests here covered. Every filtered count asserted so far filters a **scalar** (`where: { organizationRole: 1 }`). #335 filters through a **relation** — `where: { product: { deletedAt: null } }`, a condition on what a join row points at.

That is a different code path. The filter goes through `compileWhere`, so a relation inside it nests an `exists` that has to correlate to the count's *own* alias:

```sql
, (select count(*) from "Account" as "_c0"
   where "_c0"."userId" = "User"."id" and exists
     (select 1 from "Organization" as "_r0"
      where "_r0"."id" = "_c0"."organizationId" and "_r0"."name" = ?)) as "_count.accounts"
```

A scalar filter never exercises the qualifier that supplies `"_c0"` there. Had it been wrong, the inner `exists` would correlate to `"User"` and the count would answer a different question without failing — the same silent-wrong-number failure this PR is about.

It is correct, and now asserted: one compiler test for the nested `exists`, and two type tests (the filter typechecks, and a bad column one level down is still rejected). `packages/gemi` 2696 tests, template type tests 75, template `tsc` unchanged at 277.

nstfkc (commented):
## Overview

A type-only fix: `_count.select.<relation>` now accepts `{ where }` alongside `boolean`, factored into one `CountSelection<M>` shared by `SelectInput` and `IncludeInput`. Plus a new SQL test for a filter that reaches through a relation of the counted model, six type tests in the template, and a comment correction on `ArgsOf` made stale by #326. No runtime code changes.

The framing checks out: `compileRelationCount` has always read `_count.select.<rel>.where` (`packages/gemi/orm/compile/relation-count.ts:113`), and `docs/orm.md:885` already documented the filtered form. The type was the only thing refusing it.

## What I verified

- **`relation-count.test.ts` at PR head: 20/20 pass**, including the new nested-`exists` assertion — the inner subquery does correlate to `"_c0"`, not `"User"`.
- **Template `select.test-d.ts`: 23/23, no type errors**, against the PR's `packages/gemi`. Worth noting for anyone reproducing this: the template's `node_modules/gemi` symlink resolves to the *main* checkout, so a naive run inside a PR worktree silently typechecks against the old types and reports 13 failures. Repoint it before concluding anything.
- **The negative tests are real, not vacuous.** Raw `tsc` confirms `nonExistentColumn` → TS2353 against `WhereInput<AccountTypes>`, and an unknown key such as `take` inside a count node → TS2353 against `{ where?: … }`.
- **Plan cache is unaffected.** `where` is in `VALUE_KEYS` (`packages/gemi/orm/plan.ts:403`), so it terminates the literal `_count` subtree — one entry per shape, values rebound through `locate`. No churn from user-supplied filter values, and `plan.discrimination.test.ts:832-878` already asserts it.
- **Policy scoping survives a caller's filter.** `scopeCounts` → `applyPolicies` → `withScope` ANDs the scope into an existing `where` rather than replacing it (`packages/gemi/orm/policy.ts:576`), so a now-writable count filter cannot drop a tenant scope. Covered by `policy.test.ts:818`.

## Findings

**1. The same bug one type over, left unfixed (medium).** `orderBy: { accounts: { _count: "desc", where: { organizationRole: 1 } } }` is supported by the compiler and exercised by `plan.discrimination.test.ts:127`, `:693` and `:843` — but `OrderByInput`'s relation branch is `{ _count?: SortOrder } | OrderByInput<target>`, and neither arm holds both keys. Verified it fails with TS2322. Given this PR's own thesis, and that the sibling relation-ordering case sits three lines above `CountSelection` in the same file, I would either fix it here or open the issue now while the context is loaded.

**2. `CountSelection` maps to-one relations the compiler refuses (low).** `countPlan` throws `UnsupportedQueryError` for `kind !== "many"`, but the type accepts `_count: { select: { organization: true } }` and `RelationCountPayload` types it as `number`. `RelationInfo` carries `kind`, so it is a key remap away:

```ts
type CountSelection<M extends ModelTypeInfo> = {
  [K in keyof Relations<M> as Relations<M>[K]["kind"] extends "many"
    ? K
    : never]?: boolean | { where?: WhereInput<Relations<M>[K]["target"]> };
};
```

Pre-existing rather than a regression — but it is the same "type does not describe the compiler" gap, in the direction that accepts too much.

**3. Orphaned doc comment (nit).** `CountSelection` and its docblock were inserted *between* `SelectInput`'s existing "`select`, one level. Relations carry their own nested arguments…" comment and `SelectInput` itself. That comment now sits above the wrong symbol and `SelectInput` has none. Moving the new type and its doc above it restores both.

**4. Filter value types are not checked (observation, out of scope).** `where: { organizationId: "x" }` on an `Int` column typechecks inside `_count` — but it does at the root `where` too, so this is `WhereInput`'s looseness rather than anything this PR introduces. Raising it only because the new tests could give the impression the filter is fully checked; they verify key names, not value types.

## Quality and risk

The change is the right shape — one shared type instead of the two copies that let `select` and `include` drift, which is the actual root cause of the class. Widening a union cannot break existing callers, and no internal consumer reads `_count.select[k]` as `boolean` in typed code (`relation-count.ts` and `policy.ts` both go through `any` / `Record<string, unknown>`). Test coverage is proportionate: SQL text and parameterisation on the runtime side, positive and negative cases on the type side. Comment density matches the file's prevailing style.

Looks good to merge once the doc-comment move is in. Finding 1 is the one worth tracking before this closes, since #333 and #335 both arrived through that same gap.

nstfkc:
Addressed. Thanks — the verification section saved me real time, particularly the plan-cache and policy-scoping checks.

**Finding 3 (doc comment)** — fixed. `CountSelection` and its docblock moved above `SelectInput`'s comment, which is back on `SelectInput`.

**Finding 2 (to-one relations)** — fixed, but **not by the suggested key remap**, because it has a hole. A model whose relations are *all* to-one maps to `{}`, and every object literal is assignable to `{}` — so on exactly the models where counting is most obviously wrong, nothing is checked. `Account` in the template is one:

```ts
// key remap: compiles clean, throws at runtime
AccountModel.findFirst({ select: { _count: { select: { organization: true } } } })
```

Keeping the key and making its value uninhabitable holds for both shapes:

```ts
type CountSelection<M extends ModelTypeInfo> = {
  [K in keyof Relations<M>]?: Relations<M>[K]["kind"] extends "many"
    ? boolean | { where?: WhereInput<Relations<M>[K]["target"]> }
    : ToOneRelationsCannotBeCounted;
};
```

The named alias earns its place: a bare `never` under an optional key prints as *"Type 'true' is not assignable to type 'undefined'"*, which sends the reader looking for a missing value. Now it reads *"not assignable to type 'ToOneRelationsCannotBeCounted'"*. Two type tests, one per model shape — the second is the one the remap would have passed.

**Finding 1 (`orderBy`)** — reproduces, and I've filed it as #338 rather than fixing it here, because I don't think it is the same bug.

For `_count.select.<rel>.where` the `where` is documented caller grammar (`docs/orm.md:885`) and the type was the only thing refusing it. In `orderBy` it is an internal channel, and the codebase says so in both places:

> Prisma's `orderBy` grammar has no slot for a `where`, so the policy walk adds one and the compiler reads it.
> — `order-relation.test.ts:194`, echoed at `policy.ts:1182`

The three `plan.discrimination.test.ts` sites you cite are asserting that the **policy-written** node discriminates the plan cache — `:127`'s own comment reads *"the key is not in Prisma's grammar and only that walk produces it."* No caller writes it in any of them.

So a type refusing a slot only the policy walk writes is defensible, and widening it is a feature — a grammar gemi has and Prisma doesn't — not a divergence fix. I did check the thing that would have made it a non-starter: a caller-supplied `where` is **not** clobbered, since `scopeRelationOrderings` routes through `applyPolicies` → `withScope`, which ANDs (`policy.ts:576`). #338 records that, so whichever way it goes is a decision rather than an accident.

**Finding 4** — agreed and untouched; `WhereInput`'s value looseness is the root `where`'s too.

Gate after the changes: `packages/gemi` 2696 pass / `tsc` clean, template 681 pass, type tests 77, template `tsc` unchanged at 277.

One correction to the reproduction note: in this worktree `templates/saas-starter/node_modules/gemi` symlinks to `../../../packages/gemi`, which resolves *inside* the worktree — I confirmed the resolved `types.ts` carries `CountSelection` before trusting any of the type runs above. Worth knowing that it's per-checkout rather than always pointing at main.

Issue #335, opened 2026-08-05
A filtered `_count` is implemented and tested, but the type forbids writing it


nstfkc:
Found while verifying #332 against Folio's models. It is the one line standing between `CustomerListsV2Controller` — the file behind #326's repro — and a complete port onto the ORM.

## The error

```ts
List.findMany({
  select: {
    publicId: true,
    _count: { select: { products: { where: { product: { deletedAt: null } } } } },
  },
});
```

```
error TS2322: Type '{ where: { product: { deletedAt: null; }; }; }'
              is not assignable to type 'boolean | undefined'.
```

`SelectInput` and `IncludeInput` both declare it as a boolean per relation (`orm/types.ts:462` and `:468` on `main`):

```ts
_count?: boolean | { select?: { [K in keyof Relations<M>]?: boolean } };
```

## Why this is a type bug and not a missing feature

The compiler already does it. `orm/compile/relation-count.ts` goes looking for exactly that key and ANDs it into the correlated subquery:

```ts
const where = node === true ? undefined : (node as any)?.where;
const container = args?.include !== undefined ? "include" : "select";
const locate = (a: any) => a?.[container]?.[COUNT_KEY]?.select?.[relation.name]?.where;
…
if (filter) parts.push(sql(" and "), filter);
```

And there is a test for it — `orm/compile/relation-count.test.ts:60`:

```ts
test("a filtered count carries its own where", () => {
  … _count: { select: { accounts: { where: { organizationRole: 1 } } } } …
  `where "_c0"."userId" = "User"."id" and "_c0"."organizationRole" = ?) `
});
```

The policy layer knows about it too — the same file asserts a scope merging into `include._count.select.accounts.where` at `:255`, and `locate` exists precisely so the policy walk and the compiler agree about where the caller's own `where` lives.

So three subsystems handle a filtered count and the input type is the only thing that does not. Same shape as #326: the type not describing what the runtime does. `RelationCountPayload` needs nothing — it yields `number` per relation named either way, which is presumably why the payload half came along in #332 without this surfacing.

## Suggested shape

Mirroring what the compiler reads, and Prisma's spelling:

```ts
_count?:
  | boolean
  | {
      select?: {
        [K in keyof Relations<M>]?:
          | boolean
          | { where?: WhereInput<Relations<M>[K]["target"]> };
      };
    };
```

## Why it matters here rather than being a nice-to-have

[...]

PR #339, opened 2026-08-05
Let `orderBy` say where nulls go, and name what replaced `distinct`

### LINKED ISSUE #337: `orderBy` cannot say where nulls go, and `distinct` has no replacement to point at
Two smaller things from the same migration as #336, both about `findMany` args. Filing together because the answer to the second may well be "by design", and that answer is worth writing down somewhere a reader will find it.

## 1. `orderBy` has no null placement

```ts
export type SortOrder = "asc" | "desc";
export type OrderByInput<M> = { [K in keyof Scalars<M>]?: SortOrder } | …;
```

Prisma also accepts `{ sort, nulls }`:

```ts
orderBy: [
  { conversation_id: "desc" },
  { platform: { sort: "asc", nulls: "last" } },
],
```

The call site is a `groupBy` whose consumer dedupes by `conversation_id` and wants a real platform badge to win over a legacy `NULL`-platform row — so the null placement is the thing being relied on, not incidental.

It happens to port: Postgres sorts nulls as larger than any non-null, so `ASC` already means `NULLS LAST` and plain `{ platform: "asc" }` emits equivalent SQL. But the port is then resting on a dialect default rather than saying what it means, and it would silently mean the opposite under `desc`. `nulls` on `SortOrder` is a small addition and the intent survives it.

## 2. `distinct` is gone with nothing to point at

`0.51`'s upgrade notes say `cursor` and `distinct` "now fail to compile instead of throwing at runtime", which is an improvement over a runtime throw — but it leaves a caller who had one nowhere to go, and the compile error does not suggest anything.

The honest version of this is that Folio's own code already contains the argument for removing it:

```ts
// Prisma's `distinct` is applied IN MEMORY by the query engine (pg adapter): it
// does not emit SQL DISTINCT/LIMIT, so a `take`/`skip` on a distinct findMany
// neither reduces the rows pulled from Postgres nor paginates by distinct group.
```

That controller was rewritten to raw `DISTINCT ON` for exactly that reason. Its sibling still uses `distinct: ["userId"]` and has the same latent problem, unfixed.

So I am not asking for `distinct` back — it was a footgun and dropping it is defensible. What is missing is the sentence saying so. A caller hitting the compile error learns that the key is unknown, not that the operator was doing something in memory that they probably did not want, nor that `DISTINCT ON` via `DB.query(sql\`…\`)` is where they should go.

Suggestion: name it in `UPGRADE.md` next to the removal, and if the ORM can afford a nicer error, make an unknown `distinct`/`cursor` key say what replaced it rather than "not assignable".

Found on `gemi@0.51.0-rc.6`.

nstfkc:
Closes #337.

## 1. The nulls placement was type-only

The compiler has always done it:

- `parseOrderBy` accepts the `{ sort, nulls }` long form
- `compileOrderBy` emits the standard `nulls first` / `nulls last` on **both** dialects — Prisma's `CASE WHEN … IS NULL` workaround for SQLite predates 3.30, and Bun bundles newer
- `reverse()` flips the placement along with the direction, so a negative `take` keeps the intent

And all three were already tested — `read.test.ts:186` (scalar), `order-relation.test.ts:55` (through a relation), `refusals.test.ts:85` (a `nulls` that is neither).

`SortOrder` was `"asc" | "desc"`, and was the only thing refusing the argument. **The same divergence as #333 and #335**, which is now the third time on this line: the type not describing what the compiler does.

```ts
orderBy: { email: { sort: "asc", nulls: "last" } }
orderBy: { accounts: { _count: { sort: "desc", nulls: "last" } } }   // through a relation too
```

One change reaches every caller: all seven `orderBy` sites route through `OrderByInput`, so `findMany`, relation args, `aggregate` and `groupBy` come along together.

### Why it matters even where the SQL is identical

Postgres sorts nulls above every non-null, so `asc` already means `nulls last` — the reporting `groupBy` in #337 ports today and emits equivalent SQL. The problem is that it is then **correct by accident**: the same default makes `desc` put nulls *first*, so flipping a direction silently inverts the behaviour someone was relying on. Saying it makes the intent survive the edit.

## 2. The `distinct` half is documentation

`docs/orm.md` already explains both refusals under **Not in scope**, and the runtime messages in `PERMANENTLY_REFUSED` are long and specific. But `UPGRADE.md` said only "which gemi refuses by design" and pointed at `#setup`, which is not where the reasoning lives.

So a caller hitting the compile error learned the key was unknown — not that Prisma applied `distinct` **in memory**, emitting no `DISTINCT` at all, so their `take` never reduced the rows pulled from the database nor paginated by group. That is a correctness and performance problem they probably want to know about, not a stylistic one. `UPGRADE.md` now says it, with the `DISTINCT ON` to write instead, and links to the right anchor.

**What I did not do:** #337 also floats making the *compile error itself* name the replacement. That is possible — an uninhabitable branded key, the technique #334 used for to-one `_count` — but `types.ts:561` carries an explicit, reasoned decision that `distinct` and `cursor` are "absent on purpose and permanently", and adding the keys would put them back in autocomplete. Reversing a documented decision seemed like it should be a decision rather than a side effect of a docs fix. Happy to do it if you want it.

## Verification

`packages/gemi` 2696 pass / `tsc` clean; template 681 pass, 81 type tests (4 new), `tsc` unchanged at 277; `docs-scope.test.ts` green after re-embedding `llms-full.txt`.

No new runtime tests, deliberately — the runtime was already covered, and adding assertions that restate `read.test.ts:186` would suggest this changed behaviour when it changed a type.


nstfkc (commented):
## Overview

Two unrelated halves. The type half widens `SortOrder` to a new `SortOrderInput = SortOrder | { sort; nulls? }` at both `OrderByInput` arms, exports `NullsOrder` / `SortOrderInput`, and adds four type tests. The docs half rewrites the `distinct` / `cursor` paragraph in `UPGRADE.md` to say *what those arguments were actually doing* and what to write instead. No runtime change.

The framing holds up. I read the three runtime claims rather than taking them: `parseOrderBy` accepts the long form (`orderBy.ts:142-182`), `compileOrderBy` appends `nulls first` / `nulls last` on both dialects with no SQLite special case (`:207-224`), `reverse()` flips the placement with the direction (`:189-192`), and the relation path has its own copy in `readDirection` (`order-relation.ts:198-234`). `SortOrder` was the only thing refusing the argument.

## What I verified

- **Bun's SQLite is 3.51.0**, and `order by a asc nulls last` executes there — so the code comment's version claim is exact, not approximate.
- **Template type tests: 6 files, 81 pass, no type errors**, against the PR's `packages/gemi`. Same reproduction caveat as #334: the template's `node_modules/gemi` symlink resolves to the main checkout, so a naive run inside a PR worktree typechecks the *old* types.
- **The negative tests are load-bearing.** Raw `tsc`: an excess key in the long form is TS2353 against `{ sort: SortOrder; nulls?: NullsOrder }`, and `nulls` without `sort` is refused.
- **Plan cache is unaffected.** `orderBy` is in `LITERAL_KEYS` (`plan.ts:180`), so `nulls` is recorded verbatim and the two placements cannot share a plan. It is a closed two-value set, so the key space grows by a constant factor per term, not with user input.
- **Docs tests: 151 pass** across `docs-scope`, `docs-links`, `docs-gaps`, `docs-summary`, `docs-imports` and `generated-surface` — the `llms-full.txt` re-embed and the new `#not-in-scope` anchor both check out.

## Findings

**1. `groupBy` now typechecks a long form it throws on (blocking).** The PR body says the widening reaches `findMany`, relation args, `aggregate` **and `groupBy`** together. Type-side it does — `GroupByArgs.orderBy` is `OrderByInput<M>` (`types.ts:804`). Runtime-side `groupBy` does not go through `parseOrderBy` at all: `compileGroupOrdering` has its own `direction()` (`group-by.ts:651`) that accepts the two strings and nothing else. Compiled at PR head:

```
UnsupportedQueryError: gemi ORM does not support 'orderBy.globalRole' yet (User.groupBy).
Expected 'asc' or 'desc', got {"sort":"asc","nulls":"last"}.
```

That is this PR's own thesis inverted — a type that accepts what the compiler refuses — and it lands on `groupBy`, which is the operation #337's motivating example is written in. `aggregate` is fine (it calls `parseOrderBy` at `aggregate.ts:172`); `groupBy` is the one caller that reimplemented the parse. Teaching `direction()` the long form is a few lines and the SQL is valid in a grouped query. If it is out of scope for a type fix, then `GroupByArgs` should `Omit` its way back to `SortOrder`, so the refusal stays at compile time where it is today.

**2. `OrderByInput` still counts to-one relations (medium).** `{ _count?: SortOrderInput }` applies to every relation, but `projection` throws `UnsupportedQueryError` for `kind !== "many"` (`order-relation.ts:156-163`). #334 closed the identical hole for `_count` *selection* with `ToOneRelationsCannotBeCounted` (`types.ts:509-513`) — sixty lines above the line this PR edits, and the branded-key technique the PR body cites. The same remap applies here, and the runtime's error text is already written to be lifted into it.

**3. The `distinct` replacement is Postgres-only (medium, docs).** `UPGRADE.md`'s snippet is `select distinct on ("userId") …`, offered without a dialect note, in a project whose SQLite dialect is first-class. On Bun's SQLite that is `SQLiteError: near "on": syntax error` — I ran it. Either scope the snippet or give the portable form beside it (`group by` with `max`, or `row_number()`, which both dialects have). Worth noting nothing would have caught this: `docs-imports.test.ts:100` excludes root `UPGRADE.md` from compilation on purpose.

**4. SQLite's default is the *opposite* of Postgres's, and that is the better argument (docs).** The new section says the long form is worth writing "even when the default already agrees", and argues from intent surviving an edit. True, but it undersells it. Postgres sorts nulls above every non-null; SQLite sorts them below — `select a from t order by a asc` returns `[null, 1, 2]` there, which I checked rather than recalled. So the same gemi query sorts differently on the two dialects unless it says `nulls`, and a `findMany` ported from one to the other changes its answer silently. For a dual-dialect ORM that is the headline, and the current text never mentions SQLite's default at all.

**5. `groupBy`'s aggregate ordering cannot be spelled either (observation, out of scope).** `orderBy: { _count: { role: "desc" } }` is documented (`group-by.ts:548`), implemented, and tested — and `OrderByInput<M>` has no `_count` at the top level, so it is TS2353 from typed code in both the short and long forms. Not introduced here, but it is the same divergence family as #337/#333/#335 and it is in the operation finding 1 is about; `groupBy`'s `orderBy` is the loosest-fitting type left in the file. Probably an issue rather than a change here.

## Quality and risk

The type change is the right shape and cannot break an existing caller — a union widened, both arms edited together, and the doc comment carries the reasoning at the density the file already uses. The four type tests are well chosen: they pin the required `sort`, the closed `nulls` set, and the relation path, which is exactly where the union could have gone wrong. The `UPGRADE.md` rewrite is a real improvement — "the key is unknown" to "here is what it was doing to you and here is the replacement" is the difference between a refusal and an explanation.

Skipping new runtime tests is the right call and the PR is right to say so; `corpus.ts:57-58` already exercises both placements against a live Prisma client, which is stronger evidence than a restated assertion.

One unrelated observation from running subsets: `read.test.ts` "a relation in a select needs the registry" fails when run in the same process as other compile test files and passes alone (140/140) — a cross-file registry dependency that predates this PR and is invisible in a full-suite run.

Finding 1 is the one I would fix before merging: the PR closes a type/runtime gap on `findMany` and opens a smaller one on `groupBy`, and the next person to hit it will read the release note that says `groupBy` came along.

nstfkc:
All four addressed. Finding 1 was a real defect and mine — thank you for catching it.

**1 — `groupBy` (blocking).** Confirmed by compiling it, not by reading: `UnsupportedQueryError … Expected 'asc' or 'desc', got {"sort":"asc","nulls":"last"}`. I asserted in the PR body that the widening reached `groupBy` because every `orderBy` routes through `OrderByInput`, which is true of the *type* and told me nothing about the compiler. `groupBy` is the one caller that reimplements the parse, because its ordering may name an aggregate.

Fixed by teaching `direction()` the long form rather than `Omit`-ing back to `SortOrder`, since the SQL is valid on a grouped ordering on both dialects and narrowing the type would have left `groupBy` the only operation that cannot say where nulls go — in the operation where it matters most, given a grouped column is exactly the kind that is often null. Works on a column and after an aggregate:

```
order by "globalRole" asc nulls last
order by count("globalRole") desc nulls first
```

**2 — to-one `_count` ordering.** Fixed, reusing `ToOneRelationsCannotBeCounted` rather than a second brand, so the select and ordering paths refuse it in the same words.

**3 — `distinct on` is Postgres-only.** Ran it: `near "on": syntax error`. Now labelled as Postgres, with a `row_number()` form beside it — which I also ran on SQLite against real rows to check it dedupes rather than merely parses.

**4 — the defaults are opposite.** You're right that this is the better argument and the docs never made it. Verified on Bun's SQLite 3.51.0 rather than recalled:

```
order by a asc   →  [null, 1, 2]      (postgres: [1, 2, null])
order by a desc  →  [2, 1, null]
```

So the same gemi query returns a different order on the two databases unless it says `nulls`. That now leads the section, with the intent-survives-an-edit point second where it belongs.

**5 — `groupBy`'s aggregate ordering.** Filed as #340. It doesn't fit here: fixing it by widening `OrderByInput` would put an aggregate ordering on `findMany`, where it means nothing and would throw. It wants a `GroupByOrderByInput`, which could also express the grouped-column restriction the runtime already enforces — the shared type can't say that.

**Runtime tests this time**, since finding 1 turned this into a runtime change: long form on a column and on an aggregate, short form unchanged, and three refusals.

On the unrelated observation — `read.test.ts`'s registry test failing only in a shared process — I saw the same thing and it reproduces on `main`, so it predates this branch. Not touching it here; worth its own issue if you want one.

Gate: `packages/gemi` 2702 pass / `tsc` clean; template 681 pass, 84 type tests, `tsc` unchanged at 277.

PR #342, opened 2026-08-06
Give `groupBy` its own ordering type, and stop calling it unimplemented

### LINKED ISSUE #340: `groupBy`'s aggregate ordering has no type, so `orderBy: { _count: … }` is a compile error
Raised as finding 5 of the review on #339, filed rather than fixed there because it is a different shape of gap from the one that PR closes.

## The gap

```ts
User.groupBy({ by: ["globalRole"], _count: true, orderBy: { _count: { globalRole: "desc" } } });
// error TS2353: Object literal may only specify known properties,
//               and '_count' does not exist in type 'OrderByInput<UserTypes>'
```

Documented (`compile/group-by.ts:548`), implemented (`aggregateOrdering`), and tested — `group-by.test.ts` asserts both `order by count("globalRole") desc` and `order by count(*) asc`, and after #339 the long form on an aggregate too. `GroupByArgs.orderBy` is `OrderByInput<M>`, which has `_count` only under a *relation* key, never at the top level. So the top-N-by-count query — the one `groupBy` mostly exists for — cannot be written from typed code in any spelling.

## Why it is its own issue

#339 closed a divergence by widening a type both operations share. This one cannot be fixed that way: `groupBy`'s ordering grammar is genuinely *different* from `findMany`'s, not a superset of it.

- `findMany` orders by columns and relations.
- `groupBy` orders by **grouped columns** — a strict subset, enforced at runtime with Prisma's own rule, *"every field used for orderBy must be included in the by-arguments"* — and by **aggregates**, which `findMany` has no notion of.

Widening `OrderByInput` to carry `_count` at the top level would put an aggregate ordering on `findMany`, where it means nothing and would throw. The fix is a `GroupByOrderByInput` that `GroupByArgs` uses instead — which also lets it express the grouped-column restriction the runtime already enforces, something the shared type cannot say.

So this is the loosest-fitting type left in the file, in both directions at once: it accepts orderings `groupBy` refuses, and refuses orderings `groupBy` accepts.

## Related

Same family as #333, #335 and #337 — the type not describing what the compiler does. The three of those were type-only; this one wants a new type rather than a wider one, which is why it is worth its own change.

### LINKED ISSUE #341: `sql.ts` names `groupBy` as unimplemented, and it ships in the published types
`orm/sql.ts`'s header comment lists `groupBy` among the shapes the ORM declines, as the motivation for composable fragments existing at all:

```
 * The ORM deliberately does not implement every SQL shape — `distinct`, cursor
 * pagination, `groupBy`, window functions, recursive CTEs. That is the right
 * call, but it is only half a decision: the shapes it declines have to have
 * somewhere to go […]
```

`groupBy` is implemented: `compile/group-by.ts`, `group-by.test.ts`, a `static groupBy` on every generated model, `GroupByArgs`/`GroupByPayload` in `types.ts`, and a documented section in `docs/orm.md` ("`groupBy`", including the note that `orderBy` is optional). #339 extended `orderBy` nulls to it by name.

The other four in the list are genuinely declined, so the sentence reads as authoritative — which is why it's worth fixing rather than leaving. It ships in the published types (`dist/orm/sql.d.ts:7`), so it's the first thing a caller reads when they go looking for the escape hatch, and it points them at raw SQL for something they can write directly.

That's how it caught me: I was scoping which of an app's analytics reads could move onto the ORM, and wrote off a `groupBy` call site on the strength of this comment before finding the compiler module.

Dropping the one word is probably the whole fix. `distinct` and cursor pagination are the load-bearing examples anyway — both are refused *permanently*, which is exactly the case the paragraph is arguing about, whereas window functions and recursive CTEs are simply not on the roadmap.

Found on `gemi@0.51.0-rc.8`.

nstfkc:
Closes #340. Closes #341.

## #340 — `groupBy`'s ordering had no type of its own

`GroupByArgs.orderBy` was `OrderByInput<M>`, the type the reads use. The two grammars are *different*, not one wider than the other, so borrowing it was loose in both directions at once:

| | `findMany` | `groupBy` |
|---|---|---|
| column | yes | yes, and only if it is in `by` |
| relation | yes | **no** — the reader looks keys up in `schema.fields` |
| aggregate | no | **yes** — `orderBy: { _count: { role: "desc" } }` |

So the top-N-by-count query was a compile error, including the one `docs/orm.md:431` leads the `groupBy` example with, and a relation ordering typechecked and threw.

`GroupByOrderByInput<M>` is columns plus the five aggregate kinds. Which fields each kind takes reuses `AggregateArgs`' own `NumericKeys` / `OrderableKeys`, so it cannot drift from `assertAggregable`: `_sum` over a String and `_max` over `Json` stay compile errors, `_count` applies to anything, and `_all` — which compiles to `count(*)` and names no column — stays `_count`'s alone.

**What it deliberately still does not say** is that a column has to be one of the `by` ones. That depends on the `by` literal rather than on `M`, and it would mean threading `by` through the generated signature as its own type parameter. `bin/orm/emit.ts:890` already recorded the reason it stays a runtime refusal, before this PR: the type-level form reports a mapped type where `compile/group-by.ts` names the field and explains why a group has one value for a grouped column and many for everything else. A worse-worded compile error is not an improvement. I kept that decision and pointed the docblock at the new type so the two do not drift.

**One narrowing.** `GroupByArgs.orderBy` no longer accepts an `OrderByInput`, so a shared `const ordering: OrderByInput<T>` passed to both `findMany` and `groupBy` now fails at the `groupBy` call. Nothing that worked breaks — a relation ordering there has always thrown `UnknownFieldError` — but it is a compile error where there was none.

## #341 — `sql.ts` named `groupBy` among the shapes the ORM declines

One word, in the header that motivates composable fragments existing at all. It ships in the published types (`dist/orm/sql.d.ts`), so it was the first thing a caller read on the way to the escape hatch, and it sent them to raw SQL for something they can write directly. The reporter hit exactly that: wrote off a `groupBy` analytics call site on the comment's authority before finding `compile/group-by.ts`.

The other four — `distinct`, cursor pagination, window functions, recursive CTEs — are genuinely declined, which is why the sentence read as authoritative. Only `groupBy` is gone; the paragraph's load-bearing examples are the two permanent refusals, which is the case it is arguing about.

## Verification

- **Runtime, unchanged.** `compile/group-by.ts` already emitted all of this — this PR adds no compiler behaviour. Added one test pinning the claim the new type rests on: a relation key in a `groupBy` `orderBy` throws `UnknownFieldError`, where `findMany` orders by it fine.
- **Types.** 11 new type tests, including the four refusals (`_sum` over String, `_max` over Json, `_all` under `_sum`, a relation) each with `@ts-expect-error` — an unused directive would fail the run, so each is a live refusal, not a decoration. The existing "on a column and on an aggregate" test was only testing the column half, because the aggregate half did not compile; it does now.
- gemi `2703 passed | 1 skipped`, `tsc` 0 errors.
- template `681 passed`, `91` type tests (was 80), `tsc` 277 errors — the unchanged pre-existing baseline.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3254,6 +3254,9 @@ shape. Four rules are worth knowing because they are not what the arguments sugg
   in `by` and many for everything else, so ordering by anything else is a question with no answer.
   SQLite answers it anyway, by picking an arbitrary row's value, so this is refused rather than
   passed through. `orderBy: { _count: { field: "desc" } }` is the top-N-by-count query and is fine.
+  It compiles to `count(field)`, which counts the rows where `field` is not null — for the size of
+  the group, which is what `_count: true` returns, order by `_count: { _all: "desc" }` instead. On a
+  non-nullable column the two agree; on a nullable one they disagree exactly where nobody checks.
 - **`orderBy` is optional.** `groupBy` with none is a legal query.
 - **`having` filters groups, `where` filters rows.** `having: { role: { gt: 0 } }` needs `role` to be
   in `by`; `having: { email: { _count: { gt: 1 } } }` does not, because a count has one value per
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -440,6 +440,9 @@ shape. Four rules are worth knowing because they are not what the arguments sugg
   in `by` and many for everything else, so ordering by anything else is a question with no answer.
   SQLite answers it anyway, by picking an arbitrary row's value, so this is refused rather than
   passed through. `orderBy: { _count: { field: "desc" } }` is the top-N-by-count query and is fine.
+  It compiles to `count(field)`, which counts the rows where `field` is not null — for the size of
+  the group, which is what `_count: true` returns, order by `_count: { _all: "desc" }` instead. On a
+  non-nullable column the two agree; on a nullable one they disagree exactly where nobody checks.
 - **`orderBy` is optional.** `groupBy` with none is a legal query.
 - **`having` filters groups, `where` filters rows.** `having: { role: { gt: 0 } }` needs `role` to be
   in `by`; `having: { email: { _count: { gt: 1 } } }` does not, because a count has one value per
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -887,7 +887,10 @@ function countOperation(model: string): string {
  * `compile/aggregate.ts` — so a `_sum` over a `String` is a compile error at the
  * same place the runtime would refuse it.
  *
- * `groupBy` is beside it, and one rule of its is still a runtime check.
+ * `groupBy` is beside it, and takes `GroupByOrderByInput` rather than the
+ * `OrderByInput` the other reads take, because its ordering may name an
+ * aggregate and theirs may name a relation — different grammars, not a wider
+ * one (#340). One rule of its is still a runtime check.
  *
  * `orderBy` may only name a column that is in `by`. Expressing that in the type
  * would mean making `orderBy`'s key set depend on the `by` literal, which is
--- a/packages/gemi/orm/compile/group-by.test.ts
+++ b/packages/gemi/orm/compile/group-by.test.ts
@@ -3,7 +3,7 @@ import { afterEach, beforeEach, describe, expect, test } from "vitest";
 import { PostgresDialect } from "../dialect/postgres";
 import { SqliteDialect } from "../dialect/sqlite";
 import { UnknownFieldError, UnsupportedQueryError } from "../errors";
-import { account, organization, user } from "../fixtures";
+import { account, mapped, organization, user } from "../fixtures";
 import * as registry from "../registry";
 import { compileGroupBy } from "./group-by";
 
@@ -208,6 +208,43 @@ describe("having", () => {
     ).toThrow(/having.globalRole.contains/);
   });
 
+  /**
+   * The other half of what borrowing `WhereInput` got wrong: a `having` key is
+   * resolved against `schema.fields`, so a relation is refused here even though
+   * `where` filters by one. `GroupByHavingInput` has no relation arm for the
+   * same reason `GroupByOrderByInput` has none (#340).
+   *
+   * `resolveField` names it as a relation rather than reporting an unknown
+   * field. `compileGroupOrdering` does not — it reads `schema.fields` directly,
+   * so the same mistake in an `orderBy` gets the generic message. Left as it is:
+   * both are now compile errors for typed callers, and changing which error an
+   * untyped one gets is a behaviour change this PR did not set out to make.
+   */
+  test("a relation is refused here, unlike in where", () => {
+    expect(() =>
+      text({ by: ["globalRole"], _count: true, having: { accounts: { some: {} } } }),
+    ).toThrow(/'accounts' is a relation/);
+  });
+
+  /**
+   * `assertAggregable` applies to a `having` aggregate, not only to a selected
+   * one. It matters most here because this is the one case `GroupByHavingInput`
+   * cannot refuse: a `Json` column's filter is a union with `JsonValue`, which
+   * contains `{ [key: string]: JsonValue }`, so an object literal spelled like
+   * an aggregate filter is also a legal *value* for it. The type documents that
+   * hole; this is what closes it.
+   */
+  test("an aggregate the column cannot answer is refused, Json included", () => {
+    expect(() =>
+      compileGroupBy(
+        mapped,
+        "groupBy",
+        { by: ["size"], _count: true, having: { payload: { _max: { gt: 1 } } } },
+        sqlite,
+      ),
+    ).toThrow(/has no ordering/);
+  });
+
   test("no value reaches the SQL text", () => {
     const args = {
       by: ["globalRole"],
@@ -327,6 +364,19 @@ describe("orderBy", () => {
     expect(() =>
       text({ by: ["globalRole"], _count: true, orderBy: { globalRole: "asc; drop table User" } }),
     ).toThrow(UnsupportedQueryError);
+  });
+
+  /**
+   * An aggregate with no fields under it is no ordering at all, and the type
+   * admits it. Left that way rather than made loud: `orderBy: {}` and
+   * `where: {}` behave the same, an empty object is what a conditionally-built
+   * argument produces, and refusing it would refuse the shape that spelling is
+   * for. Pinned so it stays a decision.
+   */
+  test("an empty aggregate is no ordering, not an error", () => {
+    expect(text({ by: ["globalRole"], _count: true, orderBy: { _count: {} } })).not.toContain(
+      "order by",
+    );
   });
 
   test("an unknown column is an unknown field, not a group error", () => {
--- a/packages/gemi/orm/compile/group-by.test.ts
+++ b/packages/gemi/orm/compile/group-by.test.ts
@@ -334,6 +334,19 @@ describe("orderBy", () => {
       text({ by: ["globalRole"], _count: true, orderBy: { nope: "asc" } }),
     ).toThrow(UnknownFieldError);
   });
+
+  /**
+   * A relation reaches the same refusal, because this reader looks a key up in
+   * `schema.fields` and relations are not among them. `findMany` orders by a
+   * relation happily; `groupBy` never could, and `GroupByOrderByInput` is the
+   * type that finally says so — it has no relation arm, where the shared
+   * `OrderByInput` it used to borrow did (#340).
+   */
+  test("a relation is an unknown field here, unlike in findMany", () => {
+    expect(() =>
+      text({ by: ["globalRole"], _count: true, orderBy: { accounts: { _count: "desc" } } }),
+    ).toThrow(UnknownFieldError);
+  });
 });
 
 describe("arguments", () => {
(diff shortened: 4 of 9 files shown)


nstfkc (commented):
## Overview

Two independent surface fixes, no runtime change. `GroupByArgs.orderBy` stops borrowing `OrderByInput` and gets `GroupByOrderByInput` — columns plus the five aggregate kinds, no relation arm — and `sql.ts` stops listing `groupBy` among the shapes the ORM declines.

The framing is right, and the table in the body is the thing I'd have had to draw myself: the two grammars are different, not nested, so #340 genuinely could not be closed the way #337 was. Borrowing the reader's type was loose in both directions at once, and only one of them was visible as a bug.

## What I verified

I compiled every shape the new type admits against `compileGroupBy` at PR head rather than reading the types across:

| written | emitted |
|---|---|
| `_count: { globalRole: "desc" }` | `order by count("globalRole") desc` |
| `_count: { globalRole: { sort: "desc", nulls: "first" } }` | `… count("globalRole") desc nulls first` |
| `_count: { _all: "asc" }` | `order by count(*) asc` |
| `_sum: { id: "desc" }` | `order by sum("id") desc` |
| `_min: { createdAt: "asc" }` | `order by min("createdAt") asc` |
| column + aggregate, and the array form | both, in written order |

And every shape it refuses is refused by the compiler too, with an error that says the same thing: `_sum` over a `String` → `'email' is a String, and _sum needs a number`; `_all` under `_sum` and a relation key → `UnknownFieldError`. `NumericKeys` / `OrderableKeys` are `assertAggregable`'s own sets (`NUMERIC`, `UNORDERABLE`), so the drift the docblock promises can't happen is structurally prevented, not just currently true.

- **The four negative tests are load-bearing.** I flipped `_sum: { email: … }` to a numeric column: `TypeCheckError: Unused '@ts-expect-error' directive`. Each refusal is live.
- **gemi: 116 files, 2703 passed | 1 skipped. `tsc -p` 0 errors. Template: 6 files, 91 type tests, no type errors** — the last against *this PR's* `packages/gemi`, via a shadow `node_modules` (the template's `gemi` symlink otherwise resolves to the main checkout and typechecks the old types).
- **The narrowing is safe as claimed.** No `.groupBy(` call site outside the tests exists in the repo, so nothing in-tree passes a shared `OrderByInput`.
- **#341 is complete.** `sql.ts:20` was the only place left claiming `groupBy` is declined — nothing in `docs/`, `UPGRADE.md` or `README.md` still says it.
- Also confirmed my #339 finding is closed: `direction()` learned `{ sort, nulls }` in 7872165, so the aggregate ordering this PR types accepts the long form all the way down.

## Findings

**1. `having` is the same divergence, on the line above (medium).** `having?: WhereInput<M>` (`types.ts:857`) sits directly above the line this PR changes and has exactly #340's shape: the aggregate filter is implemented, tested (`group-by.test.ts:126`, `:186`, `:200`), documented (`docs/orm.md:445`) — and a compile error from typed code.

```
having: { email: { _count: { gt: 1 } } }
  → TS2353: '_count' does not exist in type 'NestedFilter<string | null>'
having: { id: { _sum: { gt: 0 } } }
  → TS2353: '_sum' does not exist in type 'NestedFilter<number>'
```

Both compile and run: `having count("email") > ?`, `having sum("id") > ?`.

The sharp version: the `groupBy` example at `docs/orm.md:425` is the one the PR cites as the motivating case, and its `having` and `orderBy` lines are adjacent — 430 and 431. This PR makes 431 compile and leaves 430 as TS2353. The lead example still doesn't typecheck as written, just on a different key now. `having` needs its own input type for the same reason `orderBy` did — a per-field filter *or* a filter on an aggregate over that field. Fold it in or file it as #340's sibling, but it shouldn't sit unlisted after this PR ships the argument for fixing it.

**2. Nothing measures the shape this unlocks against Prisma (medium, tests).** `GROUPS` (`corpus.ts:108-118`) has one ordering entry — `{ by: ["name"], _count: true, orderBy: { name: "asc" } }`, a column. There is no aggregate ordering anywhere in the differential corpus. That harness is the project's strongest evidence and the reason PRs here can skip restating assertions; the top-N-by-count query has never been run against a live client, and this PR is what invites everyone to write it. Two entries (`{ _count: { _all: "desc" } }`, `{ _count: { name: "desc" } }`) would cost nothing and would settle finding 3 by measurement.

**3. `_count: { field }` is `count(field)`, not `count(*)` (docs).** From the probe above: the aggregate arm emits `count("globalRole")`, where `_all` emits `count(*)`. On a nullable grouped column those differ — the null group's ordering key is 0, not its size — so the top-N ordering and the `_count: true` value being displayed disagree in the one case a reader would never check. This matches Prisma, so it's a note rather than a defect; but `docs/orm.md:442` currently says `orderBy: { _count: { field: "desc" } }` "is the top-N-by-count query and is fine", and one clause pointing at `_all` for group size would earn its space. It's also what the corpus entries in finding 2 would pin.

**4. `orderBy: { _count: {} }` is silently no ordering (minor).** The type admits an empty aggregate object and `aggregateOrdering` returns no parts, so the query compiles with no `order by` at all. Same class as an empty `where`, and probably fine to leave — noting it because the other refusals here are all loud.

## Quality and risk

The type is the right shape: mirroring `AggregateArgs`' member layout means `_count` taking any column while `_sum` needs arithmetic isn't a rule restated in a second place, it's the same expression. `_all` under `_count` alone matches `CountSelect`'s `{ _all?: true } & …` exactly, which is what makes the two readable side by side.

Not typing the `by` restriction is the right call, and pointing the docblock at `emit.ts`'s existing record of *why* is better than repeating the reasoning — a mapped-type error naming twelve columns really is worse than the sentence `compile/group-by.ts` throws.

Risk is low. The narrowing is a compile error where there was none, but only for code that was already guaranteed to throw at runtime, and nothing in-tree does it. Merging this and handling finding 1 separately is fine — with the caveat that the docs example the PR is written around still won't compile until that one lands.

nstfkc:
All four addressed in eaf69ae. Findings 1–3 folded in; finding 4 pinned as a decision rather than changed.

## 1 — `having`, folded in

You were right that it shouldn't sit unlisted after this PR ships the argument for fixing it, and it turned out to be a stronger case than the finding states: `having` was wrong in **three** directions, not two.

Borrowing `WhereInput` refused the aggregate filter, as you showed. It also **offered operators the compiler does not have** — `compileHaving` walks its own operand reader with six comparisons (`OPERATORS`, `group-by.ts:479`), so `having: { name: { contains: "a" } }` typechecked and threw *"A 'having' filter takes equals, gt, gte, lt, lte, not"*. And it offered a relation arm, where a `having` key resolves against `schema.fields`.

`GroupByHavingInput<M>` is the union your quoted rule is — a comparison on a grouped column, *or* a filter on an aggregate over any column, which is why it can't be a per-field choice. `_count` and `_avg` compare against a number whatever the column was; `_sum`/`_min`/`_max` keep the column's type, which is `AggregatePayload`'s rule read from the other side.

The two refusals are **named types rather than absent keys** — `AggregateNeedsArithmetic`, `AggregateNeedsAnOrdering` — because a `never` under an optional key reports "not assignable to type 'undefined'", which says nothing about arithmetic. Same reason `ToOneRelationsCannotBeCounted` exists. `orderBy` keeps the key-set form: there the *field* is the key, so restricting the key set is the natural spelling; in a `having` the *kind* is the key.

One shape it still admits and the compiler still refuses: `{ role: { gt: 0, _count: { gt: 1 } } }`. TypeScript relaxes excess-property checking against a union, so both keys are known to some member. `assertPureAggregate` refuses it by name and gives the `AND` spelling. Documented in the docblock rather than papered over — the exclusion trick that would close it costs the error message, and the runtime refusal here is loud.

`docs/orm.md:425`'s example now compiles on both lines.

## 2 — corpus

Four entries, not two. Your `{ _count: { _all: "desc" } }` and `{ _count: { name: "desc" } }`, plus the long form on an aggregate, plus an aggregate `having` on an ungrouped column — the shape finding 1 unlocked, which had no entry for the same reason.

## 3 — `count(field)` vs `count(*)`

Added to `docs/orm.md:442`, and it earns its space exactly where you said: `_count: true` is `count(*)`, the ordering is `count(field)`, and on a nullable column the number displayed and the number ordered by disagree. Re-embedded in `llms-full.txt`.

## 4 — the empty aggregate

Left as no-ordering, and pinned with a test so it is a decision rather than an accident. `orderBy: {}` and `where: {}` already behave this way, and an empty object is what a conditionally-built argument produces — refusing it would refuse the shape that spelling exists for.

## Two corrections to my own last commit

Both caught by running the claims rather than reading them.

**A relation in a `having` is not an unknown field.** `resolveField` gives relations their own message — *"'accounts' is a relation. Aggregate its own model instead"*. My docblock said unknown field, and the test I wrote asserting `UnknownFieldError` failed. Fixed both. Note the two paths differ: `compileGroupOrdering` reads `schema.fields` directly, so the same mistake in an `orderBy` still gets the generic message. Left alone — both are compile errors for typed callers now, and changing which error an untyped one gets is a behaviour change this PR didn't set out to make. Worth its own issue if you want them consistent.

**The `_max`-needs-an-ordering refusal cannot be expressed on a `Json` column.** My first draft used `metadata` and the directive came back unused. A `Json` column's filter is a union with `JsonValue`, which contains `{ [key: string]: JsonValue }`, so an object literal spelled like an aggregate filter is also a legal *value* — the hole `FieldFilter` already documents. The type test now uses `avatar` (`Bytes`) for the refusal and keeps the `Json` case deliberately **without** a directive, saying why. A runtime test covers what the type can't: `assertAggregable` refuses `_max` over `Json` in a `having`.

## Gate

- gemi **2706 passed | 1 skipped**, 116 files, `tsc` 0.
- template **681 passed**, **102** type tests (80 before this PR), `tsc` 277 — the unchanged baseline.
- Confirms your shadow-`node_modules` caveat is unnecessary here, incidentally: the `Json` type test failed with *"Unused '@ts-expect-error' directive"* before I fixed it, which only the new types could produce — the template's `gemi` symlink resolves inside this worktree.

PR #347, opened 2026-08-06
Add `onUserCreated`, and stop handing the OAuth path a dead token

### LINKED ISSUE #344: Add an `onUserCreated` hook that runs inside the user-creation transaction
An application that has to create rows **alongside** a user — a workspace, an organization, a default settings row — has no hook that can do it atomically, and the hook it will reach for first is wrong in two ways at once.

## Why `onSignUp` cannot carry it

**It runs after the user is committed.** `AuthController.signUp`:

```js
L = await Y.createUser({ email: z, name: w, password: H, verificationToken: B, locale: A });
return await K.onSignUp(L, B, Z.search.toJSON()), L;
```

Two separate awaits, no transaction spanning them. Provisioning in the hook means the user row is already committed when the second insert runs, so a failure leaves a user with no workspace — and this is the failure mode you only see in production, because the second insert works fine until it doesn't.

**It never fires for OAuth.** `oauthCallback` creates the user and goes straight to the session:

```js
U = "signup";
$ = await Z.createUser({ email: W, name: w, locale: z, emailVerifiedAt: new Date });
await Z.createSocialAccount({ provider: Y, userId: $.id, ... });
H = await K.createOrUpdateSessionV2({ email: $.email, id: $.id });
```

`signUp` is `onSignUp`'s only call site. So an app that provisions there works for email/password signups and silently produces org-less users for every Google one — no error, no log, a user that lands in the product with nothing to open.

The workaround is subclassing `UserProvider` and overriding `createUser`, which is what Folio does. It works, but it requires knowing that `createUser` is the *only* seam where this is possible, and nothing points there: the app is reaching past the documented config surface into a class method to get atomicity that the config surface implies it already has.

## The proposal

A `onUserCreated` hook on `defineAuthConfig`, called inside a transaction that wraps the user insert.

`createUser` is one line today:

```js
async createUser(s) {
  let i = await this.run(() => this.models.User.create({ data: s }));
  return eC(i);
}
```

and `Model.transaction`'s contract does the rest — "every ORM query in it, at any call depth, through any number of services, joins it automatically". So the hook's own writes join with nothing threaded through them, which is the property that makes this worth having as a hook rather than a documented subclassing recipe. Nesting is a savepoint, so an app that already wrapped its own transaction around the signup composes instead of breaking.

**All three signup paths go through `createUser`** — `signUp` twice (invited and not) and `oauthCallback` once, which is every `createUser({` call site in the bundle. One hook covers them all, including the one `onSignUp` misses.

## Four things worth deciding rather than discovering

**1. Which scope the hook runs in.** `this.run()` is `asSystem`, because auth runs before there is a user and a deny-by-default policy would otherwise turn a failed sign-in into a server error. If the hook fires *inside* that, application writes silently run unpoliced. If it fires inside the transaction but *outside* `run()`, they run under normal policy resolution — which during a signup has no authenticated user either, so an app whose provisioning writes are policied has to think about it. Folio's provisioning runs on normal scope today and works, so outside-`run()`-inside-the-transaction is the behaviour I'd argue for, but it is a decision either way and the two differ silently.

**2. The hook needs to know why the user is being created.** This is the sharpest one. `signUp`'s invited branch creates the user and then attaches them to the inviting org:

```js
if (V) L = await Y.createUser({...}),
       await Y.createAccount({ organizationId: V.organizationId, userId: L.id, organizationRole: V.role });
```

A hook that unconditionally provisions an own workspace gives that user two. Folio's override documents this as an *assumption* — that no gemi-native invitation signup is wired, because neither of its forms sends `invitationId` — which is a load-bearing assumption living in a docblock, one form field away from being false. Passing context (`{ user, invitation?, provider? }`, or whatever shape fits) turns it into something an app can branch on instead of assume.

**3. The user object should be the password-stripped one.** Same reasoning `createUser`'s own docblock gives for `onSignUp`: a hook that logs its argument would otherwise log a credential hash.

**4. Failure semantics.** A throw rolling back the user is the whole point, but it changes what `signUp` returns. Worth stating whether a `ValidationError` from the hook reaches the client as a 400 or becomes a 500.

Two constraints worth putting in the hook's own docblock rather than leaving to be found: raw Prisma inside it will not join the transaction, and `Promise.all` over ORM calls is not safe there, since a transaction holds one reserved connection.

## Naming

`onUserCreated` reads as a notification, and it sits next to `onSignUp`, `onSignIn` and `onSignOut`, which all are. This one is different in kind — it participates in the write and can fail it. Whatever it ends up called, the docblock should lead with that, because an app that mistakes it for a notification and swallows its own errors gets exactly the orphan this exists to prevent.

Found on `gemi@0.51.0-rc.9`. Context: Folio provisions an organization and its owning Account alongside every user, currently via a `UserProvider` subclass.

### LINKED ISSUE #345: `onSignUp`'s token argument is a dead string on the OAuth signup path
`config.onSignUp(user, verificationToken, search)` fires on both signup paths, but its second argument means something different on each — and on the OAuth path it is a string that resolves to nothing.

## The two paths

`AuthController.signUp` generates an **email verification token**, persists it on the user row, and hands the same value to the hook:

```js
B = await K.generateEmailVerificationToken(z);
L = await Y.createUser({ email: z, name: w, password: H, verificationToken: B, locale: A });
return await K.onSignUp(L, B, Z.search.toJSON()), L;
```

`AuthController.oauthCallback` generates a **magic link token** and hands that over instead:

```js
if (U === "signup") {
  let V = await X.generateMagicLinkToken(W);
  await X.onSignUp($, V, J.search.toJSON());
} else await X.onSignIn($, J.search.toJSON());
```

## Why the second one is dead

`config.generateMagicLinkToken` is a pure hash — the default is `sha256(email + Date.now())` — and `oauthCallback` never persists what it returns. The method that stores one is `userProvider.createMagicLinkToken`, which writes a `MagicLinkToken` row, and it is not called here.

So the token the hook receives on an OAuth signup is in neither place a token can be looked up:

- `findUserByVerificationToken` reads `User.verificationToken`, which for this user was never set — `createUser` is called without one.
- `findUserMagicLinkToken` reads the `MagicLinkToken` table, where no row was written.

An application that does the obvious thing with the hook's second argument — mails "confirm your address" with a link carrying it — sends every Google signup a link that cannot resolve. It works in testing if you only test email/password signup, which is the path where the argument means what its name says.

## Smaller, same area

When `verifyEmail` is false, `generateEmailVerificationToken` returns `""` by default, so the email path hands the hook an empty string while the OAuth path hands it 64 characters of unusable hex. Two paths, one parameter, three different meanings depending on config.

## Options, in the order I'd rank them

1. **Pass `""` on the OAuth path.** A Google signup arrives with `emailVerifiedAt` already set — there is nothing to verify — so the honest value is "no token", which is what `verifyEmail: false` already produces on the other path. Smallest change, and it makes the argument mean one thing.
2. **Persist it**, if the intent was to hand new OAuth users a magic link. That needs the `createMagicLinkToken` call this path is missing, and it is a behaviour decision rather than a fix.
3. **Tell the hook which path it is on**, so an app can branch instead of guessing from the token's shape.

Found on `gemi@0.51.0-rc.9`.

nstfkc:
Closes #344. Closes #345.

## `onUserCreated`

An application that has to create rows *alongside* a user — an organization, a workspace, a default settings row — had nowhere to do it atomically. `onSignUp` fires after the user is committed on both paths, so provisioning there leaves an orphaned user when the second insert fails, and that is a failure that only turns up in production. The workaround was subclassing `UserProvider` to override `createUser`, which the docs recommended and which requires knowing that `createUser` is the only seam where atomicity is reachable.

`config.onUserCreated(user)` runs *inside* the transaction that writes the user, after the row and before the commit. A throw rolls the user back. It is the one hook here that participates in the write rather than reporting on it, which is what its docblock leads with — the past-tense name reads like its neighbours and is not one of them.

`UserProvider.transaction(fn)` is the seam it runs in. `AuthController` has no model to reach for, and a bare `Model.transaction` would open on the ambient connection rather than the one the provider's models are bound to. Deliberately **not** wrapped in `run`: `asSystem` stays applied per query, so an application's provisioning writes run under the same policy resolution as the rest of its code instead of silently unpoliced.

All three creation paths now write inside one transaction:

| Path | In the transaction |
| --- | --- |
| sign-up | `createUser` → `onUserCreated` |
| invited sign-up | `deleteInvitationById` → `createUser` → `createAccount` → `onUserCreated` |
| OAuth first login | `createUser` → `createSocialAccount` → `onUserCreated` |

Two writes moved inside beyond the hook itself, because the rollback is otherwise incoherent. `createSocialAccount` has a foreign key onto the user. `deleteInvitationById` used to run *before* creation, so a rollback would burn the invitation with no user to show for it and leave the invitee unable to retry.

## The OAuth token argument

`oauthCallback` handed `onSignUp` a `generateMagicLinkToken` result. That function is a pure hash and nothing persisted what it returned — `createMagicLinkToken`, which writes the `MagicLinkToken` row, was never called — so the value resolved in neither place a token can be looked up: not `User.verificationToken`, which `createUser` does not set here, and not the `MagicLinkToken` table. An app mailing a confirmation link carrying it sent every OAuth signup a link that could not resolve, and it tested clean against email/password signup, where the same parameter is the persisted verification token.

It now passes `""`. An OAuth user arrives with `emailVerifiedAt` already set, so there is nothing to verify, and `""` is what the email path already passes under `verifyEmail: false`. An app that does want to hand new OAuth users a magic link should mint a persisted one with `Auth.createMagicLink` from the hook.

## Tests

Three tests in the template's `user-provider.test.ts` cover the property the design rests on: a throw after the insert leaves no user, it also rolls back what the callback wrote through *other* models, and the commit case keeps the password strip. `AuthController` needs an HTTP request to reach, so the hook's own invocation is not covered there.

- `tsc --noEmit` clean, `oxlint` 0 errors
- `vitest run` in `packages/gemi`: 116 files, 2706 passed
- `vitest run app/models/user-provider.test.ts`: 24 passed

## Docs

`docs/authentication.md` gains an `onUserCreated` section and loses the note telling applications to provision inside a `createUser` override. The `Changing a query` example was swapped for one that is actually a query change and that exercises the `this.models` / `this.run` paragraph below it. `llms-full.txt` carries the page verbatim and was updated to match; the two were diffed afterwards.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -173,25 +173,31 @@ The twenty-two methods, all overridable:
 | `createSocialAccount(args)` | Persist an OAuth-linked social account. |
 | `findInvitation(id, email)` / `deleteInvitationById(id)` / `createAccount(args)` | Invitation-based sign-up. |
 
+One more method is not a query: `transaction(fn)` runs `fn` inside a transaction on the
+connection the `User` model lives on. `AuthController` uses it to write a user and everything
+that has to exist beside it — the invited `Account`, the OAuth `SocialAccount`, and whatever
+[`onUserCreated`](#onusercreated) writes — atomically.
+
 ### Changing a query
 
-Subclass `UserProvider` and override the method(s) you need — for example to make sign-up
-atomic by provisioning an organization in the same transaction that creates the user:
+Subclass `UserProvider` and override the method(s) you need — for example so that a
+soft-deleted user cannot sign back in:
 
 ```typescript
 import { UserProvider } from "gemi/kernel";
-import type { CreateUserArgs, User } from "gemi/kernel";
+import type { User } from "gemi/kernel";
 
-import { Organization } from "@/app/models/Organization";
-import { User as UserModel } from "@/app/models/User";
-
-export class OrgProvisioningUserProvider extends UserProvider {
-  async createUser(args: CreateUserArgs): Promise<User> {
-    return UserModel.transaction(async () => {
-      const user = await super.createUser(args);
-      await Organization.create({ data: { name: `${user.name}'s org`, ownerId: user.id } });
-      return user;
-    });
+export class SoftDeleteUserProvider extends UserProvider {
+  async findUserByEmailAddress(email: string, verifyEmail: boolean): Promise<User> {
+    return this.run(() =>
+      this.models.User.findFirst({
+        where: {
+          email,
+          deletedAt: null,
+          ...(verifyEmail ? { emailVerifiedAt: { not: null } } : {}),
+        },
+      }),
+    );
   }
 }
 ```
@@ -209,7 +215,7 @@ argument:
 import { AuthManager } from "gemi/services";
 import { ServiceProvider } from "gemi/support";
 
-import { OrgProvisioningUserProvider } from "@/app/auth/OrgProvisioningUserProvider";
+import { SoftDeleteUserProvider } from "@/app/auth/SoftDeleteUserProvider";
 
 export default class AppServiceProvider extends ServiceProvider {
   register() {
@@ -218,7 +224,7 @@ export default class AppServiceProvider extends ServiceProvider {
       () =>
         new AuthManager(
           this.app.config.get("auth", {}),
-          new OrgProvisioningUserProvider(),
+          new SoftDeleteUserProvider(),
         ),
     );
   }
@@ -229,10 +235,11 @@ The provider has to be listed in your `Kernel`'s `providers` array to run — se
 [Project Structure](./project-structure.md#service-providers). App providers register *after*
 the framework's, so this binding replaces the default `AuthManager` rather than racing it.
 
-> **Note:** gemi runs `createUser` and then fires `onSignUp` *separately*. Provisioning
-> inside the `onSignUp` callback is therefore **not** atomic with user creation — a failure
-> there leaves an orphaned user. Do transactional provisioning inside a `createUser`
-> override, as above.
+> **Note:** provisioning does not need a subclass any more. `onSignUp` fires *after* the user
+> is committed, so provisioning there is not atomic — a failure leaves an orphaned user — but
+> [`onUserCreated`](#onusercreated) runs inside the transaction that writes the user and rolls it
+> back on a throw. Reach for a `createUser` override when you need to change *the query*; use
+> the hook to write rows alongside it.
 
 ### Upgrading from the adapter config
 
@@ -263,7 +270,8 @@ may be sync or async.
 
 | Hook | Signature | Fires when |
 | --- | --- | --- |
-| `onSignUp` | `(user, verificationToken, search)` | A new account is created (email/password, or first OAuth sign-in). `verificationToken` is empty when `verifyEmail` is off. |
+| `onUserCreated` | `(user) => Promise<void>` | A user row is written — **inside the transaction that writes it**, before it commits. A throw rolls the user back. See [onUserCreated](#onusercreated). |
+| `onSignUp` | `(user, verificationToken, search)` | A new account is created (email/password, or first OAuth sign-in), after it is committed. `verificationToken` is empty when `verifyEmail` is off, and on the OAuth path, where there is nothing to verify. |
 | `onSignIn` | `(session, search)` | A user authenticates (password, magic-link/PIN, or returning OAuth). |
 | `onSignOut` | `(session)` | The `/auth/sign-out` endpoint runs. |
 | `onForgotPassword` | `(user, token)` | A password-reset is requested — send the reset email with `token`. |
@@ -273,6 +281,55 @@ may be sync or async.
 
 `search` is the request's query string as a plain object (useful for attribution / redirect
 params).
+
+### onUserCreated
+
+Every other hook in that table is a notification: it fires after its work is committed and
+can only report. `onUserCreated` is not. It runs **inside the transaction that creates the
+user**, after the row is written and before it commits, and if it throws, the user is rolled
+back and the sign-up fails.
+
+That is what it is for. An application that has to create rows *alongside* every user — an
+organization, a workspace, a default settings row — has nowhere else to do it atomically.
+Provisioning in `onSignUp` runs after the commit, so a failed second insert leaves a user
+with no organization: no error the user sees, nothing to retry, and a failure that only
+shows up in production.
+
+```typescript
+import { defineAuthConfig } from "gemi/services";
+
+import { Account } from "@/app/models/Account";
+import { Organization } from "@/app/models/Organization";
+
+export default defineAuthConfig({
+  async onUserCreated(user) {
+    // Both of these join the open transaction automatically. If either throws,
+    // the user is rolled back with them.
+    const org = await Organization.create({ data: { name: `${user.name}'s org` } });
+    await Account.create({
+      data: { userId: user.id, organizationId: org.id, organizationRole: 0 },
+    });
+  },
+});
+```
+
+It fires on all three paths that create a user — email/password sign-up, invited sign-up, and
+first OAuth sign-in — and receives the same password-stripped user the endpoint returns.
+
+Four things to know:
+
+- **Errors reach the client.** A [`ValidationError`](./forms.md) thrown here is a 400 on
+  `POST /sign-up`; anything else is a 500. Either way no user is created.
+- **The invited path already has an `Account`.** When a sign-up carries an `invitationId`, the
+  inviting organization's `Account` row is written before this runs, so a hook that
+  unconditionally provisions an own workspace gives that user two. Check before adding one.
+- **Only ORM queries join the transaction.** They do at any call depth, through any number of
+  services, with nothing threaded through — that is `Model.transaction`'s contract. A raw
+  Prisma or `DB` statement runs outside it and survives the rollback.
+- **One statement at a time.** The transaction holds a single reserved connection, so
+  `Promise.all` over ORM calls is not safe here — await them in sequence. Keep network and
+  filesystem I/O out of the hook entirely; the connection is held for as long as it runs. Send
+  the welcome email from `onSignUp`, which fires after the commit.
 
 ### Example: magic-link PIN emails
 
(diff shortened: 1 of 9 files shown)


nstfkc (commented):
Checked out the branch, ran the template suite (24 passed, 1 skipped — the Postgres suite, no DB available), and probed the ORM behaviour the design rests on.

## The design holds up

I verified the load-bearing part: `transaction` is deliberately *not* wrapped in `run`, and per-query `asSystem` still works inside it — `runAsSystem` spreads the current store (`packages/gemi/orm/context.ts:106`), so `tx` survives and the provider's own queries join the transaction. Moving `deleteInvitationById` inside is a real bug fix, and `""` on the OAuth path is correct.

## Findings

### 1. The documented `onUserCreated` example fails on any policied model

I ran it. A `create` on a model carrying a `scope` policy, inside `userProvider.transaction` with no user in scope, throws:

> `PolicyDeniedError: Organization.create is governed by a policy and there is no user in scope. If this is a cron tick, a queue worker or a script, say so at the call site: Model.asSystem(() => ...). Policies are never skipped just because a user failed to turn up.`

Sign-up has no authenticated user by definition, so a tenant-scoped `Organization` — the exact model in the example at `docs/authentication.md:305-311` and in the `config.ts` docblock — 500s on every sign-up, and the rollback takes the user with it.

`UserProvider.transaction`'s docblock argues the other way:

> an application's provisioning writes should run under the same policy resolution as the rest of its code

The rest of an app's code runs with a user. This doesn't.

The design choice is right — hoisting `asSystem` into `transaction` would be worse. It's the docs that mislead. Suggest a fifth bullet under "Four things to know", and showing `Model.asSystem(() => ...)` (or `asUser(user, ...)`) in the example itself. As written, the first person to follow it hits a 500 they cannot sign up past.

### 2. `UserProvider.transaction`'s stated rationale is not true today

`packages/gemi/auth/UserProvider.ts:175`. The docblock says a bare `Model.transaction`

> would open on the ambient connection rather than the one this provider's models are bound to

But `Model.transaction` is `transact(app(DatabaseManager), () => fn())` (`packages/gemi/orm/Model.ts:418`) — it never touches `this`, and `withTransaction` pins the SINGLE-CONNECTION ASSUMPTION in its own comment. So `this.models.User.transaction(fn)` and `Model.transaction(fn)` are the same call, and the failure described cannot happen.

The method still earns its place — the controller shouldn't import a model, and this is the right seam when a second connection becomes configurable. Just state it as forward-looking rather than as current fact.

### 3. The invited path still passes `undefined` where OAuth now passes `""`

`packages/gemi/auth/AuthController.ts:330`. `let verificationToken: string` is assigned only in the `else` branch, so the invited path reaches `config.onSignUp(newUser, undefined, …)` against a signature typed `verificationToken: string`.

This PR's own argument applies verbatim: the invited path sets `emailVerifiedAt` and never writes a token, so there is nothing to verify and `""` is the honest value. Initialising to `""` is one line, and it makes the docs table — "empty when `verifyEmail` is off, and on the OAuth path" — true of all three paths rather than two.

### 4. The hook's invocation is untested on every path

Fair as stated: there are no `AuthController` tests in `packages/gemi` at all, so there's no harness to extend, and the `transaction` tests pin the right property. What's left unpinned is ordering — hook after `createAccount` on the invited path, after `createSocialAccount` on OAuth — which the docs state as contract and a refactor would break silently. Worth a follow-up issue rather than a blocker here.

### 5. Minor

- The error bullet covers `POST /sign-up` (400 / 500), but the hook also runs in `oauthCallback`, where a throw is a 500 page mid-redirect rather than a form error. One clause would cover it.
- Sign-up now opens a BEGIN/COMMIT around a single INSERT for apps with no hook configured. Fine, just noting it.

---

Good to merge with 1 and 3 addressed; 2 is a comment reword.

nstfkc:
Thanks — 1 and 3 were both real. Pushed in 06c8c0b.

## 1. The policy trap — fixed, and it was worse than a doc bug

I reproduced it rather than taking it on faith, and got your error verbatim:

```
PolicyDeniedError: Organization.create is governed by a policy and there is no
user in scope. If this is a cron tick, a queue worker or a script, say so at the
call site: Model.asSystem(() => ...).
```

You're right that the design choice is correct and the docs were the problem. The example now wraps its writes in `Model.asSystem`, and the bullet list gained a fifth entry that leads with the failure, explains why the hook is deliberately *not* inside `UserProvider.run`, and notes that `asUser(user, …)` does **not** rescue the common case — a tenant scope reading `organizationId` can't be satisfied by the user whose organization is the row being created.

Three tests pin it, in a new file because `register` is global and a policied `Organization` can't share a file with the tests that query the unpolicied one:

- the denial fires and the user is rolled back
- `Model.asSystem` is the documented fix and works
- `asSystem` does **not** escape the transaction — which is the one that actually matters, because if it did, everything the hook wrote would survive a rollback and the hook would be pointless

Writing that turned up something adjacent: the ORM refuses `create` outright on a model that scopes reads and declares no `onCreate` ("as written the policy would confine reads to the caller and let an insert name any value it likes"). So a scope-only policy can't even reach the denial — the test needed a policy with both members, which is the realistic shape anyway.

## 3. `undefined` on the invited path — fixed

Confirmed: `let verificationToken: string` assigned only in the `else`, passed at what was line 381. Now `let verificationToken = ""`, with the reasoning at the declaration. Your framing is right that it's this PR's own argument applied to the third path, and the docs table is now true of all three.

## 5. Fixed

The error bullet now covers `oauthCallback`, where a throw is a 500 page mid-redirect rather than a form error. The BEGIN/COMMIT around a lone INSERT I've left as-is, as you suggested.

## 4. Filed as #348

Agreed it's a follow-up. I wrote it up with the specific hazard: move the `onUserCreated` call one line up on the invited path and the documented "the `Account` row already exists" contract becomes false, every test still passes, and invited users silently get two organizations.

## 2. I think this one is mistaken — leaving the docblock as written

`Model.ts:418` is inside `save`. `transaction` is at 573:

```ts
static async transaction<T>(fn: () => Promise<T>): Promise<T> {
  const name = this.$connection ?? currentConnectionName();
  return transact(app(DatabaseManager).connection(name), () =>
    runOnConnection(name, () => fn()),
  );
}
```

It does read `this`, against `static $connection?: string` declared at `Model.ts:212` and set by `on()` at 256 — and the comment at 221 documents `static $connection = "analytics"` on a class as a supported way to bind one. So `this.models.User.transaction(fn)` and `Model.transaction(fn)` diverge whenever the application's `User` carries `$connection` or came from `.on(…)`: the former opens on the model's connection, the latter on whatever is ambient.

Happy to be wrong if you're reading a different line — but on 573 the docblock describes today, not a forward-looking claim, so rewording it as aspirational would make a true sentence false.

---

Suite after the changes: 116 files / 2707 passed in `packages/gemi`, 278 passed in the template (the 4 red files there are the `prisma generate` + `build:bin` ones CI sets up first, red on `main` too). `tsc` clean, `oxlint` 0 errors.

PR #351, opened 2026-08-07
Mint the nanoid gemi was leaving to a database that has no default

### LINKED ISSUE #350: `@default(nanoid())` is classified `dbgenerated`, so an ORM insert omits a column Prisma fills client-side
`@default(nanoid())` is a Prisma **client-side** default: the generated Prisma client fills it, and the column Prisma's own migration emits carries no `DEFAULT`.

```sql
CREATE TABLE "Organization" (
    "id" SERIAL NOT NULL,
    "publicId" TEXT NOT NULL,   -- @default(nanoid(10)) — no DEFAULT
    ...
```

The ORM generator records it as `{ "kind": "dbgenerated" }`, and `isClientSideDefault` excludes `dbgenerated` from the values gemi supplies:

```ts
// orm/defaults.ts
export function isClientSideDefault(spec: DefaultSpec): boolean {
  return spec.kind !== "autoincrement" && spec.kind !== "dbgenerated";
}
```

So `Model.create` omits the column and Postgres rejects the row on NOT NULL. `cuid`, `uuid`, `now` and literal values are all handled; `nanoid` is the one Prisma client-side default that is not.

This is the same class of bug `defaults.ts`'s own docblock is about — it opens by explaining that `publicId` and `updatedAt` have to be client-side *because Prisma's DDL gives them no database default*. `nanoid` is that exact case, just routed to the other branch.

## How it turns up

Silently at first, then on every write. Nothing fails at boot or at generate time — the field looks handled, because the generator did classify it. The first `create` on the model raises a NOT NULL violation from the driver, naming the column but not the reason.

Hit while moving an app onto 0.51: `onUserCreated` is the only seam where provisioning can join the sign-up transaction, and only ORM queries join it, so the one `Organization.create` that *must* go through the ORM is also the one that cannot. Worked around by minting the id in application code from the alphabet in `@prisma/client`'s bundled nanoid:

```ts
const URL_ALPHABET = "useandom-26T198340PX75pxJACKVERYMINDBUSHWOLF_GQZbfghjklqvwyzrict";
// crypto.getRandomValues, byte & 63, `size` chars
publicId: prismaNanoid(10),
```

That works, but it puts a copy of a schema default in application code with nothing keeping the two in agreement.

## Suggested fix

Add `nanoid` alongside `cuid`/`uuid` in `clientSideValue`, and have the generator emit `{ kind: "nanoid", length }` rather than `dbgenerated` — the length argument is part of the default and has to survive into the artifact (`nanoid()` is 21, `nanoid(10)` is 10). Prisma's implementation is `size` characters of its url alphabet masked with `& 63`, so matching it byte-for-byte is a few lines, the same way `createCuid` already matches Prisma's cuid v1 output.

Worth deciding at the same time: whether an unrecognised `dbgenerated` on a NOT NULL column with no database default should be *detectable* at all, rather than only at first write. `bunx gemi check models` already walks the schema, so it has somewhere to say so.

## Impact on existing apps

Additive — no app currently relies on gemi omitting a `nanoid` column, because doing so fails. Regenerating changes the artifact for schemas that use it.

Filed from the folio 0.51 upgrade; the app-side note is Quantus-Labs/folioai.com#727.

nstfkc:
Closes #350.

## The bug

`@default(nanoid())` is a Prisma **client-side** default. The generated client fills it, and the column Prisma's own migration emits carries no `DEFAULT`:

```sql
CREATE TABLE "Organization" (
    "id"       SERIAL NOT NULL,
    "publicId" TEXT   NOT NULL,   -- @default(nanoid(10)) — no DEFAULT
```

The gemi generator recorded it as `{ kind: "dbgenerated" }`, which `isClientSideDefault` excludes from the values gemi supplies. So `Model.create` omitted the column and the driver rejected the row on NOT NULL — on the first write and every write after. `cuid`, `uuid`, `now` and literals were all handled; `nanoid` was the one Prisma client-side default routed to the other branch.

Nothing failed at generation time or at boot, because the field *looked* handled: the generator had classified it.

## The fix

`nanoid` becomes its own `DefaultKind`, and `defaults.ts` mints the id the way Prisma does — its url alphabet, `& 63`, `size` characters. The alphabet is 64 symbols, so unlike `createCuid`'s `% 36` the mask needs no resample loop: 256 is exactly four whole alphabets and every byte maps to one symbol.

**The length travels with the kind.** `nanoid()` is 21 and `nanoid(10)` is 10, and the migration writes a plain `TEXT` for both — so `DefaultSpec.length` is the only place the width survives. A spec that reached the runtime without it would still produce an id, 21 characters of the right alphabet, and the column would just be silently the wrong size.

`SCHEMA_ARTIFACT_VERSION` does **not** move, for the reason already written beside it. The hazard a bump guards against is a new artifact read by an old runtime — which here would fall to `clientSideValue`'s `default:` branch and reproduce this exact bug. That pairing cannot occur: generator and runtime ship in the same package at the same version, so a runtime that cannot read `kind: "nanoid"` came with a generator that does not emit it. The reverse pairing is the one apps hit on upgrade, and it is fine — an older artifact still says `dbgenerated` and is read as it always was, until `prisma generate` runs again.

## The silence, which is the other half

The issue asked whether an unrecognised default should be detectable at all, rather than only at first write. It should, and the generator is the place — not `gemi check models`, which walks model *classes* and the registry and never reads the schema.

`prisma generate` is the last moment anything knows the function's **name**. By the time the driver rejects the row, all that is left is a NOT NULL violation on a column that ought to have had a default. So a function default with no case for it, on a **required** column, now warns and names model, field and function:

```
gemi ORM: Organization.publicId is required and defaults to auto(), which this generator
does not recognise. It is recorded as a database-side default, so the ORM will omit the
column on insert — and if auto() is a default Prisma fills in its own client, the column
has none in the database and every write to Organization will fail on NOT NULL. Make the
field optional, give it a default the ORM knows, or upgrade gemi.
```

A warning rather than a refusal: most `dbgenerated(...)` columns really are the database's, and breaking a schema that generates and runs fine today is a worse trade than a noisy one. A nullable column says nothing at all — the guess costs nothing there, and noise is how a warning stops being read.

## Tests

Five new tests, one per layer the value passes through:

| Where | What |
| --- | --- |
| `emit.test.ts` | `nanoid` classifies as its own kind, and the length is recorded — including the `nanoid()` → 21 case |
| `defaults.test.ts` | `isClientSideDefault` and `hasClientSideValue` both answer `true`; `clientSideValue` returns an id of the schema's width |
| `compile/write.test.ts` | the column reaches the compiled `insert`, bound at the right length |

Plus a `createNanoid` block covering length, alphabet, per-call freshness and that the `& 63` mask is unbiased across all 64 symbols, and the warning's own block — nullable stays quiet, literals stay quiet, and every name in `RECOGNISED_DEFAULTS` stays quiet. That test reads the exported set rather than listing the six names a third time.

`isClientSideDefault`'s exhaustive `Record<DefaultKind, boolean>` gains `nanoid: true`, so the type makes a future kind a compile error rather than an untested branch. It could not have caught this one: `nanoid` was not a member at all.

**Mutation-checked rather than assumed.** Reintroducing the bug in both places — `isClientSideDefault` excluding `nanoid`, and the generator emitting `dbgenerated` — fails exactly those five and nothing else.

- `vitest run` in `packages/gemi`: 116 files, 2732 passed, 1 skipped
- `tsc --noEmit` clean; `oxlint` 0 warnings, 0 errors over `orm/` and `bin/`

## Docs

`docs/orm.md` had no defaults section at all, so the split this issue turned on was written down only in a source comment — a reader had no way to learn that `cuid()` is gemi's while `autoincrement()` is the database's, or why `now()` is gemi's despite the column having a real `DEFAULT`.

A new **Column defaults** section: the DDL that forces the split, a table of who fills what, and the two consequences that surprise people — `createdAt` and `updatedAt` come back from one `create` as the same instant, and a nullable column with no default is omitted rather than bound as `null`. The warning gets a subsection so its message can be looked up.

`llms-full.txt` carries the page verbatim and was regenerated from it; `docs-scope.test.ts` compares the two exactly.

### DIFF
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -124,6 +124,24 @@ function isFunctionDefault(
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
 
+/** Prisma's `nanoid()` with no argument. */
+const NANOID_DEFAULT_LENGTH = 21;
+
+/**
+ * The length `@default(nanoid(n))` was written with.
+ *
+ * DMMF types an argument as `string | number`, so a non-number is not
+ * unreachable by the types even though the Prisma parser rejects one — and the
+ * failure it would cause is the quiet kind: `undefined` reaches
+ * `clientSideValue`, which falls back to 21, and every id is silently the wrong
+ * width. Falling back here instead keeps the wrong width from also being
+ * unrecorded, and there is one place that decides it.
+ */
+function nanoidLength(args: readonly (string | number)[] | undefined): number {
+  const [size] = args ?? [];
+  return typeof size === "number" ? size : NANOID_DEFAULT_LENGTH;
+}
+
 function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
   if (!field.hasDefaultValue || field.default === undefined) return undefined;
 
@@ -136,15 +154,91 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
       case "now":
       case "dbgenerated":
         return { kind: value.name };
+      // `nanoid()` is a *client-side* default, like `cuid()` and `uuid()` and
+      // unlike everything else in this switch: the column Prisma's own
+      // migration emits carries no `DEFAULT`, so a row inserted without the
+      // value violates NOT NULL (#350). It is the only one of the three that
+      // takes an argument, and the argument has to come with it — see
+      // `DefaultSpec.length`.
+      case "nanoid":
+        return { kind: "nanoid", length: nanoidLength(value.args) };
       default:
-        // An unrecognised function default (`auto()`, `nanoid()`, a future
-        // addition) still tells the runtime "the database supplies this", which
-        // is all iteration 4 needs to know when it omits the column on insert.
+        // An unrecognised function default (`auto()`, a future addition) still
+        // tells the runtime "the database supplies this", which is all
+        // iteration 4 needs to know when it omits the column on insert.
+        //
+        // `warnOnUnrecognisedDefault` is the other half: for a *required* column
+        // that guess is a row the driver will reject, and this is the last
+        // moment anything can say so before the first write does.
         return { kind: "dbgenerated" };
     }
   }
 
   return { kind: "value", value: value as unknown };
+}
+
+/**
+ * Every function default `defaultSpec` has a `case` for.
+ *
+ * A second list of the same names, and the duplication is worth naming rather
+ * than hiding: a `switch` is not enumerable, so there is nothing to derive this
+ * from. What drift costs is asymmetric, which is why the copy is tolerable. A
+ * name added to the switch and not to this set produces a warning about a
+ * default that is in fact handled — noise, and noise is how a warning stops
+ * being read. A name added here and not to the switch silences a warning that
+ * should have fired, which is the state before #350.
+ *
+ * Exported so the test that asserts each of these stays quiet reads the set
+ * rather than becoming a third copy of it.
+ */
+export const RECOGNISED_DEFAULTS = new Set([
+  "autoincrement",
+  "cuid",
+  "uuid",
+  "nanoid",
+  "now",
+  "dbgenerated",
+]);
+
+/**
+ * A required column whose default this generator did not recognise.
+ *
+ * The `default:` branch above assumes the database fills the column. When that
+ * assumption is wrong — the default is client-side, as `nanoid()` was — nothing
+ * fails at generation time and nothing fails at boot: the field *looks* handled,
+ * because it was classified. The first `create` on the model is what fails, with
+ * a NOT NULL violation from the driver naming the column but not the reason,
+ * and it fails on every write after that.
+ *
+ * A warning rather than a refusal, and rather than nothing:
+ *
+ * - A refusal would break a schema that generates and runs fine today. Most
+ *   `dbgenerated(...)` columns really are the database's, and a nullable one is
+ *   nobody's problem either way.
+ * - Silence is what made #350 cost an afternoon. The generator is the only place
+ *   that sees the function's *name*; by the time the driver rejects the row, the
+ *   name is gone and all that is left is a column that "should have a default".
+ *
+ * Prisma's own client-side defaults are all handled above, so in practice this
+ * fires for a provider-specific default (`auto()` on MongoDB) or one Prisma adds
+ * after this version of gemi — which is exactly the set worth naming out loud.
+ */
+function warnOnUnrecognisedDefault(model: string, field: DMMF.Field): void {
+  if (!field.isRequired) return;
+
+  const value = field.default;
+  if (!isFunctionDefault(value)) return;
+  if (!value.name || RECOGNISED_DEFAULTS.has(value.name)) return;
+
+  console.warn(
+    `gemi ORM: ${model}.${field.name} is required and defaults to ` +
+      `${value.name}(), which this generator does not recognise. It is ` +
+      `recorded as a database-side default, so the ORM will omit the column on ` +
+      `insert — and if ${value.name}() is a default Prisma fills in its own ` +
+      `client, the column has none in the database and every write to ` +
+      `${model} will fail on NOT NULL. Make the field optional, give it a ` +
+      `default the ORM knows, or upgrade gemi.`,
+  );
 }
 
 function fieldSchema(model: string, field: DMMF.Field): FieldSchema {
@@ -176,6 +270,8 @@ function fieldSchema(model: string, field: DMMF.Field): FieldSchema {
 
   const def = defaultSpec(field);
   if (def) schema.default = def;
+
+  warnOnUnrecognisedDefault(model, field);
 
   return schema;
 }
--- a/packages/gemi/bin/orm/emit.ts
+++ b/packages/gemi/bin/orm/emit.ts
@@ -1,6 +1,7 @@
 import type { DMMF } from "@prisma/generator-helper";
 
 import {
+  NANOID_DEFAULT_LENGTH,
   SCHEMA_ARTIFACT_VERSION,
   type DefaultSpec,
   type FieldSchema,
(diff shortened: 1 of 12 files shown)


packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
...
         return { kind: value.name };
+      // `nanoid()` is a *client-side* default, like `cuid()` and `uuid()` and
+      // unlike everything else in this switch: the column Prisma's own
+      // migration emits carries no `DEFAULT`, so a row inserted without the
+      // value violates NOT NULL (#350). It is the only one of the three that
+      // takes an argument, and the argument has to come with it — see
+      // `DefaultSpec.length`.
+      case "nanoid":
+        return { kind: "nanoid", length: nanoidLength(value.args) };
       default:
```
nstfkc: **`@default(ulid())` is still the #350 bug**
_correctness · CONFIRMED_

`ulid` never reaches a `case`, so it falls through here to `{ kind: "dbgenerated" }` — the exact classification this PR is removing for `nanoid`.

Prisma 6.19.2's parser accepts **seven** default functions; the string table in `prisma_schema_build_bg.wasm` reads:

```
autoincrement ulid cuid dbgenerated nanoid now uuid
```

and `@prisma/client/runtime/client.js` registers client-side generators for four of them:

```
register("cuid")  register("nanoid")  register("ulid")  register("uuid")
```

The switch handles six of the seven. `publicId String @default(ulid())` therefore gets a plain `TEXT NOT NULL` column with no DDL default, `hasClientSideValue` answers `false`, the column is omitted from the INSERT, and every `create` fails on NOT NULL — #350, on a fourth id function.

Two comments added in this PR assert the completeness that does not hold:
- `orm/defaults.ts:23` — "`cuid()`, `uuid()` and `nanoid()` are Prisma's three client-side id functions and all three belong here"
- `emit.ts:222` — "Prisma's own client-side defaults are all handled above"

There are four, and a reader who trusts either line will not re-check.

packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
       case "now":
       case "dbgenerated":
         return { kind: value.name };
+      // `nanoid()` is a *client-side* default, like `cuid()` and `uuid()` and
+      // unlike everything else in this switch: the column Prisma's own
+      // migration emits carries no `DEFAULT`, so a row inserted without the
+      // value violates NOT NULL (#350). It is the only one of the three that
```
nstfkc: **`uuid(7)` and `cuid(2)` still drop their argument**
_correctness · CONFIRMED_

> It is the only one of the three that takes an argument

It isn't. Prisma 6.19.2 accepts a version argument on `uuid` and `cuid` too — both error strings are in the shipped client runtime:

```
Invalid UUID generator arguments
Invalid CUID generator arguments
```

`case "uuid"` and `case "cuid"` return `{ kind: value.name }` and never read `value.args`, and `clientSideValue` answers `uuid` with an unconditional `crypto.randomUUID()`, which is **v4**.

So `id String @id @default(uuid(7))` — declared precisely to get time-ordered, index-local ids — gets random v4 on every gemi-written row, while any row the Prisma client writes to the same table gets a real v7. Ordering by id is then wrong, and the B-tree insert locality that is the whole point of v7 is gone. `cuid(2)` is the same shape: cuid2 requested, cuid v1 minted.

This is the identical argument-dropping defect the PR fixes for `nanoid`, left standing on the two neighbouring arms of the switch it edits.

packages/gemi/orm/schema.ts
```diff
@@ -27,6 +27,16 @@
...
+ * reason and with a sharper consequence if the reasoning were wrong: an old
+ * runtime reading `{ kind: "nanoid" }` falls to `clientSideValue`'s `default:`
+ * branch, returns `undefined`, and omits the column — which is #350 exactly.
+ * The pairing is still impossible, because a runtime that does not know the
+ * kind shipped with a generator that does not emit it. The reverse pairing is
+ * the one apps actually hit on upgrade, and it is fine: an artifact generated
+ * before #350 says `dbgenerated`, which the new runtime reads as it always did,
+ * until `prisma generate` runs again.
  */
 export const SCHEMA_ARTIFACT_VERSION = 1;
```
nstfkc: **The non-bump argument doesn't hold, and the predicted failure is wrong**
_correctness · CONFIRMED_

Two problems with the reasoning above.

**1. The pairing is not impossible.** The generated artifact is committed to git; the gemi version lives in the lockfile. They move independently. A teammate or a CI/deploy box still on the previous gemi reading a freshly committed artifact containing `{ kind: "nanoid", length: 21 }` is an ordinary Monday, not a contrived case. `ARTIFACT_VERSION` is `1` on both sides, so `assertSchemaArtifactVersion` passes and nothing detects the skew.

**2. The stated consequence is wrong.** The comment says an old runtime "returns `undefined`, and omits the column — which is #350 exactly." It does not omit the column. `isClientSideDefault` is a **denylist**:

```ts
return spec.kind !== "autoincrement" && spec.kind !== "dbgenerated";
```

An unknown `"nanoid"` is neither, so it answers **`true`**. `hasClientSideValue` is then true, `insertColumns` pushes the column with `defaultBinder`, and `clientSideValue` falls to its `default:` arm returning `undefined` — which the dialect encodes as `null`.

The INSERT binds NULL. On a required column that is a NOT NULL violation; on a nullable one it silently writes NULL into every row instead of an id, which is strictly harder to diagnose than #350.

Worth either bumping the version, or rewriting the comment around what actually happens.

packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
...
+ * - Silence is what made #350 cost an afternoon. The generator is the only place
+ *   that sees the function's *name*; by the time the driver rejects the row, the
+ *   name is gone and all that is left is a column that "should have a default".
+ *
+ * Prisma's own client-side defaults are all handled above, so in practice this
+ * fires for a provider-specific default (`auto()` on MongoDB) or one Prisma adds
+ * after this version of gemi — which is exactly the set worth naming out loud.
+ */
+function warnOnUnrecognisedDefault(model: string, field: DMMF.Field): void {
+  if (!field.isRequired) return;
```
nstfkc: **The nullable early-return hides the silent half of this bug**
_correctness · CONFIRMED_

The doc-comment justifies this with "a nullable one is nobody's problem either way", and `emit.test.ts` pins that wording. It holds for genuinely database-side defaults, and not for the case the warning exists to catch.

An unrecognised **client-side** default on a nullable column — `ulid()` today — is classified `dbgenerated`, so gemi omits it from the INSERT, the database has no `DEFAULT` to supply, and the row lands with `publicId = NULL` where Prisma would have written a ulid. Nothing fails at generate time, insert time, or read time. It surfaces later, when a lookup by that id finds nothing or a backfill trips over a column of NULLs.

The warning text compounds it by offering **"Make the field optional"** as the first remedy — which converts a loud NOT NULL failure into exactly this silent one.

packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
...
+
+  const value = field.default;
+  if (!isFunctionDefault(value)) return;
+  if (!value.name || RECOGNISED_DEFAULTS.has(value.name)) return;
```
nstfkc: **`!value.name` skips the case with the least evidence behind the guess**
_correctness · PLAUSIBLE_

`isFunctionDefault` deliberately admits an object default carrying no `name` — its doc comment argues at length that the nameless path must keep falling through `switch (undefined)` to `{ kind: "dbgenerated" }`.

A required column on that path is the *strongest* instance of what this guard was written for: the generator classified it "the database supplies this" with no name to base the guess on at all. `!value.name` short-circuits the warning there, so the one case with the least justification reports nothing.

Low-confidence on reachability — DMMF always carries a name today — but the guard and `isFunctionDefault` disagree about whether that path matters.

packages/gemi/bin/orm/emit.ts
```diff
@@ -124,6 +124,24 @@ function isFunctionDefault(
...
+ * DMMF types an argument as `string | number`, so a non-number is not
+ * unreachable by the types even though the Prisma parser rejects one — and the
+ * failure it would cause is the quiet kind: `undefined` reaches
+ * `clientSideValue`, which falls back to 21, and every id is silently the wrong
+ * width. Falling back here instead keeps the wrong width from also being
+ * unrecorded, and there is one place that decides it.
+ */
+function nanoidLength(args: readonly (string | number)[] | undefined): number {
+  const [size] = args ?? [];
+  return typeof size === "number" ? size : NANOID_DEFAULT_LENGTH;
```
nstfkc: **The non-numeric fallback is silent, in the file that just grew a warning channel**
_correctness · PLAUSIBLE_

The comment above concedes a string is not excluded by the DMMF types. If some Prisma version hands back `args: ["10"]` for `@default(nanoid(10))`, this returns `21` and the artifact records `length: 21` — authoritative-looking and wrong.

On Postgres with `@db.VarChar(10)` every insert then overflows the column; on an unconstrained `TEXT` the ids are simply the wrong width forever, with nothing anywhere recording that a fallback fired.

Given this PR adds a `console.warn` channel for precisely the "generator knows something the runtime cannot" class, this arm seems like a natural second caller.

packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
...
+ * from. What drift costs is asymmetric, which is why the copy is tolerable. A
+ * name added to the switch and not to this set produces a warning about a
+ * default that is in fact handled — noise, and noise is how a warning stops
+ * being read. A name added here and not to the switch silences a warning that
+ * should have fired, which is the state before #350.
+ *
+ * Exported so the test that asserts each of these stays quiet reads the set
+ * rather than becoming a third copy of it.
+ */
+export const RECOGNISED_DEFAULTS = new Set([
```
nstfkc: **The set can be deleted — warn from the switch's own `default:` arm**
_cleanup · CONFIRMED_

The twelve lines above carefully justify a hand-maintained copy of the switch's case labels. The copy isn't needed.

`defaultSpec`'s `default:` branch *is* the "unrecognised" branch, already holds `value.name`, and is called from `fieldSchema(model, field)`, which holds both the model name and `field.isRequired`. Threading `model` into `defaultSpec` and warning inside that arm removes:

- the six-name `RECOGNISED_DEFAULTS` set
- its justification comment
- the second `field.default` + `isFunctionDefault` pass in `warnOnUnrecognisedDefault`

and makes the drift the comment worries about structurally impossible rather than merely documented. (The export exists for a test; the test can assert through `defaultSpec` instead.)

packages/gemi/orm/defaults.ts
```diff
@@ -143,3 +153,45 @@ export function createCuid(): string {
...
+ * instead — same alphabet, same mask, same length, so a gemi-written column and
+ * a Prisma-written one are indistinguishable.
+ *
+ * Unlike a cuid there is no structure to match: a nanoid is `size` independent
+ * symbols and nothing more. No timestamp, no counter, no fingerprint — which is
+ * also why none of the collision reasoning around `createCuid` applies here.
+ * Uniformity is the only property, and `& 63` over a 64-symbol alphabet is what
+ * gives it.
+ */
+export function createNanoid(size: number = NANOID_DEFAULT_LENGTH): string {
```
nstfkc, line 250: **This is `randomBlock`'s loop again, thirty lines below `randomBlock`**
_cleanup · CONFIRMED_

Both functions do the same three things — allocate a `Uint8Array`, `crypto.getRandomValues` it, map each byte into an alphabet — differing only in whether the modulus needs a rejection step.

One helper covers both:

```ts
function randomChars(alphabet: string, size: number): string {
  const limit = 256 - (256 % alphabet.length); // 256 for 64 symbols; 252 for 36
  const bytes = new Uint8Array(size);
  crypto.getRandomValues(bytes);
  let out = "";
  for (let i = 0; i < size; i++) {
    let byte = bytes[i];
    while (byte >= limit) { /* resample */ }
    out += alphabet[byte % alphabet.length];
  }
  return out;
}
```

For nanoid's 64 symbols `limit` is 256, so the loop never rejects and the result is identical to `& 63`; for cuid's 36 it reproduces the existing 252 threshold. It also puts the uniformity argument in one place, instead of two comment blocks that each explain half of it.

packages/gemi/bin/orm/emit.ts
```diff
@@ -136,17 +154,93 @@ function defaultSpec(field: DMMF.Field): DefaultSpec | undefined {
...
+ * after this version of gemi — which is exactly the set worth naming out loud.
+ */
+function warnOnUnrecognisedDefault(model: string, field: DMMF.Field): void {
+  if (!field.isRequired) return;
+
+  const value = field.default;
+  if (!isFunctionDefault(value)) return;
+  if (!value.name || RECOGNISED_DEFAULTS.has(value.name)) return;
+
+  console.warn(
```
nstfkc: **stderr side effect in a file whose header declares it pure**
_cleanup · PLAUSIBLE_

`emit.ts` opens with:

> Everything here is a pure function of the DMMF so the emitters can be unit tested with no Prisma CLI and no database

and the repo already has a home for generator diagnostics: `bin/orm-generator.ts` holds `warnOnPrismaVersion` and `warnOnDatasource`, both `gemi ORM:`-prefixed `console.warn`s at the process boundary.

Putting the third inside `fieldSchema` means `buildModelSchemas` now writes to stderr, so every unit test of it has to stub or tolerate console output, and the stated invariant no longer holds. Returning the diagnostics and letting `orm-generator.ts` print them would keep both.

packages/gemi/bin/orm/emit.ts
```diff
@@ -124,6 +124,24 @@ function isFunctionDefault(
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
 
+/** Prisma's `nanoid()` with no argument. */
+const NANOID_DEFAULT_LENGTH = 21;
```
nstfkc: **`NANOID_DEFAULT_LENGTH` is declared twice**
_cleanup · PLAUSIBLE_

This and `orm/defaults.ts:175` each define `21` independently, and `emit.ts` already imports from `../../orm/schema` — so one exported constant would serve both.

With two copies the failure is quiet: change the runtime's value (say Prisma redefines what a bare `nanoid()` means) and the generator keeps stamping `length: 21` into every artifact, which the runtime then prefers over its own constant because `length` is present. The disagreement never surfaces.

nstfkc (commented):
## Code review — high effort

The `nanoid` fix itself is correct: I checked the alphabet against the one `nanoid` ships (64 unique characters, so `& 63` is genuinely unbiased and needs no resample loop), and `length` is threaded through the artifact properly. Tests and typecheck are green — 282 passing across the three touched files.

What the review turned up is mostly **the same class of defect left standing on the neighbouring cases** of the switch this PR edits:

- `ulid()` is unhandled and still becomes `dbgenerated` — #350 verbatim, on a fourth client-side id function
- `uuid(7)` / `cuid(2)` have their version arguments discarded, so gemi mints v4 and cuid v1 where Prisma mints v7 and cuid2
- the new `RECOGNISED_DEFAULTS` guard silences the new warning for both of them by name

Plus two independent correctness issues: the artifact-version comment's "impossible pairing" argument doesn't hold for a committed artifact read by an older installed runtime (and the failure it predicts is a NULL bind, not the column omission it claims), and the warning's nullable early-return hides the silent half of the very bug it's built to catch.

Details inline. 4 confirmed correctness · 2 plausible correctness · 2 confirmed cleanup · 2 plausible cleanup.

PR #353, opened 2026-08-07
Preload the route's view chunks so hydration stops flashing Loading

### LINKED ISSUE #352: Route view chunks aren't preloaded → hydration waterfall causes layout shift (CLS) on real networks
## Summary

Route view modules are code-split into per-view chunks and loaded lazily on the client via `window.loaders[name]()`, but those chunks are never preloaded. On a real (non-localhost) network they load in a **hydration waterfall**, and while each chunk is in flight React shows the boundary's `Loading` fallback in place of the server-rendered content. On a tall page this collapses the layout to the fallback height and then restores it when the chunk arrives, producing a large **Cumulative Layout Shift (CLS)**.

The bug is invisible on localhost (chunks resolve in single-digit ms, so hydration is effectively synchronous and CLS = 0) and only reproduces with realistic network latency, which makes it easy to misdiagnose as a font/FOUT or a scroll issue.

## Root cause

- Each route segment is registered as `lazy(() => loadViewModule(viewName))`, and `loadViewModule` calls `window.loaders[name]()`, i.e. a dynamic `import()` of that view's chunk.
- The SSR shell is produced by `renderToReadableStream(..., { bootstrapModules })`. React only emits `<link rel="modulepreload">` for `bootstrapModules` (the client entry). The route's view chunks referenced by `window.loaders` get **no** preload hint.
- Result: the browser cannot discover the view chunks until the entry runs and starts importing them, and it discovers each nested chunk only after its parent parses. For a `layout → view → child components` tree this is a serial chain of round trips.

## Evidence (from a production build, real page with a layout + view + a few sub-components)

Navigation `responseEnd` ≈ 118ms, `domContentLoaded` ≈ 170ms. The view chunks then load in a waterfall well after first paint:

```
Layout chunk     start 230ms  end 264ms
 ├ Nav/CTA/Footer start 264ms  end 266ms   (discovered only after the layout chunk parses)
 └ View chunk     start 268ms  end 270ms
    └ child chunk start 273ms  end 274ms
```

Head of the document contains exactly one preload for JS:

```html
<link rel="modulepreload" fetchPriority="low" href="/assets/client-<hash>.js" />
```

…and nothing for any of the view chunks above. (Also note `fetchPriority="low"` on the one entry that *is* preloaded — it deprioritizes the very module that unblocks the whole chain.)

On localhost the same page measures CLS = 0 because those 230–274ms chunks arrive within a few ms of each other; add latency and the fallback-collapse window widens to hundreds of ms per waterfall layer.

## Suggested fix

gemi already knows, at shell-render time, both the current route's view chain (from `componentTree`) and the built asset for each `"/app/views/<Name>.tsx"` (the same mapping the prod build uses to rewrite the `loaders` dynamic imports into hashed chunks). Emitting a `<link rel="modulepreload">` for each of the current route's view chunks in the shell `<head>` would let the browser fetch them **in parallel with the entry**, so they're resident by the time hydration reaches each boundary — no fallback shown, no shift.

Two smaller wins alongside it:
- Drop `fetchPriority="low"` on the entry's `modulepreload` (or raise it) — it's on the critical path to hydration.
- Preloading the whole route chain (not just the top view) also removes the nested per-child round trips.

Happy to help test a fix against a real deployment if useful.

nstfkc:
Fixes #352.

## What was happening

Route segments hydrate through `window.loaders[name]()` — an `import()` that nothing in the document points at. The browser cannot start those fetches until the client entry has run, and it only learns about a nested chunk once its parent has arrived and parsed, so a `layout → view → components` tree costs one round trip per level. While each one is in flight React shows that boundary's `Loading` export in place of content the server already rendered: the page collapses to the fallback height and springs back when the chunk lands.

The head carried exactly one JS hint, for the entry, and React hardcodes `fetchPriority="low"` on it (Fizz has no option for it) — Chrome holds low-priority requests back while the parser is blocked, so the module everything else waits on was also the last to arrive.

## What this does

**1. Preload the current route's chunks** (`43f066d`). `httpProd` already walks the client manifest to build `window.loaders`; it now also records, per view, every chunk an `import()` of that view pulls in (`imports`, transitively — `dynamicImports` stay out, those are lazy on purpose). The dispatcher unions the current route's chain and renders `<link rel="modulepreload">` for each, which React hoists into `<head>` ahead of the inlined stylesheets.

**2. Take the entry off the low-priority preload** (`7ce817e`). Prod hands the entry to the shell instead of to React: a plain `modulepreload` in the head, and an `import()` appended to the bootstrap script React already emits at the end of the shell. Same request, same execution point — `window.__GEMI_DATA__` is assigned in that very script immediately before, so the entry still cannot observe a document without it — but discovered on the shell's first bytes at full priority. Its own static imports ride along in the list.

Dev is untouched: Vite serves unbundled source, there is no chunk graph to announce, and `/refresh.js` has to stay in front of `/app/client.tsx`.

This second commit is separable — drop it and the first still fixes the waterfall, it just leaves the entry at `fetchPriority="low"` while everything downstream of it is at high.

## Before / after

`templates/saas-starter`, production build, `/partial/:orgId/settings/general` (layout → nested layout → view):

```html
<!-- before: one hint, for the entry, at low priority -->
<link rel="modulepreload" fetchPriority="low" href="/assets/client-<hash>.js"/>

<!-- after -->
<link rel="modulepreload" href="/assets/client-4uHAbY12.js"/>
<link rel="modulepreload" href="/assets/jsx-runtime-DGeXAQPT.js"/>
<link rel="modulepreload" href="/assets/client-DilZqjuL.js"/>
<link rel="modulepreload" href="/assets/react-CBMZq9VP.js"/>
<link rel="modulepreload" href="/assets/Layout-_1U5QNCl.js"/>
<link rel="modulepreload" href="/assets/Panel-Be_S7yJV.js"/>       <!-- Layout's own import -->
<link rel="modulepreload" href="/assets/SettingsLayout-BcvPt2DQ.js"/>
<link rel="modulepreload" href="/assets/SettingsGeneral-mVeu_ny_.js"/>
```

A route with no match (404) and any view without a built chunk simply contribute nothing.

## Verification

- `bun --bun vitest run` in `packages/gemi`: 118 files, 2774 passing. New: `server/modulePreloads.test.ts` (transitive walk, dedup, cycles, dynamic imports excluded) and `services/router/modulePreload.test.ts` (full requests through `app.fetch`, asserting on the bytes: which chunks a route announces, that another route's view is absent, that a shared chunk appears once, that the hints precede the bootstrap script, and that a server passing `bootstrapModules` still boots the old way).
- Real production build of `templates/saas-starter`, driven through headless Chrome over CDP: document hydrates (`__reactContainer` on `document`), no console errors or exceptions, 8 `modulepreload` hints, 0 `<script type="module">`, and client-side navigation through `window.loaders` still works.
- Same template in dev: unchanged output (3 React bootstrap module scripts, no extra hints), hydrates clean.

@ugurcan-aytar — the offer to test against a real deployment would be very welcome here; the numbers in the issue are the ones this is aimed at, and localhost can't show the difference.

### DIFF
--- a/packages/gemi/server/httpProd.ts
+++ b/packages/gemi/server/httpProd.ts
@@ -4,7 +4,7 @@ import { generateETag } from "./generateEtag";
 import { URLPattern } from "urlpattern-polyfill";
 import { exists } from "node:fs/promises";
 import { createStyles } from "./styles";
-import { collectModulePreloads } from "./modulePreloads";
+import { CLIENT_ENTRY_KEY, collectModulePreloads, createClientEntry } from "./modulePreloads";
 import type { App } from "../app";
 import { Instrumentation } from "./types";
 import { printStartupBanner } from "./banner";
@@ -77,10 +77,14 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
   // Booted from the bootstrap script rather than through React's
   // `bootstrapModules`, which would preload it at `fetchPriority="low"` — see
   // `clientEntry` in `ViewRouteDispatcher`.
-  const clientEntry = {
-    module: `/${manifest["app/client.tsx"].file}`,
-    preload: collectModulePreloads(manifest, "app/client.tsx"),
-  };
+  const clientEntry = createClientEntry(manifest);
+  if (!clientEntry) {
+    // Loudly, but without dying: this is boot, and an incomplete `dist/` must
+    // not cost the API routes and static assets too.
+    console.error(
+      `Client manifest has no "${CLIENT_ENTRY_KEY}" entry — dist/client is incomplete, so documents will render but never hydrate.`,
+    );
+  }
 
   // Vite's manifest `css` field is a `string[]` — a client entry can emit more
   // than one CSS chunk. Read and concatenate them all instead of interpolating
--- a/packages/gemi/server/httpProd.ts
+++ b/packages/gemi/server/httpProd.ts
@@ -74,6 +74,14 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
 
   const loaders = `{${templates.join(",")}}`;
 
+  // Booted from the bootstrap script rather than through React's
+  // `bootstrapModules`, which would preload it at `fetchPriority="low"` — see
+  // `clientEntry` in `ViewRouteDispatcher`.
+  const clientEntry = {
+    module: `/${manifest["app/client.tsx"].file}`,
+    preload: collectModulePreloads(manifest, "app/client.tsx"),
+  };
+
   // Vite's manifest `css` field is a `string[]` — a client entry can emit more
   // than one CSS chunk. Read and concatenate them all instead of interpolating
   // the array into a single path (which only works for a one-element array).
@@ -187,7 +195,7 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
 
         return await result({
           getStyles,
-          bootstrapModules: [`/${manifest["app/client.tsx"].file}`],
+          clientEntry,
           modulePreloadManifest,
           loaders,
           viewImportMap,
--- a/packages/gemi/server/httpProd.ts
+++ b/packages/gemi/server/httpProd.ts
@@ -4,6 +4,7 @@ import { generateETag } from "./generateEtag";
 import { URLPattern } from "urlpattern-polyfill";
 import { exists } from "node:fs/promises";
 import { createStyles } from "./styles";
+import { collectModulePreloads } from "./modulePreloads";
 import type { App } from "../app";
 import { Instrumentation } from "./types";
 import { printStartupBanner } from "./banner";
@@ -35,6 +36,9 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
   const ogMap = {};
   const viewModules = {};
   const cssManifest = {};
+  // Which chunks a route's views pull in, so the shell can preload the chain
+  // the client would otherwise discover one round trip at a time (#352).
+  const modulePreloadManifest: Record<string, string[]> = {};
   const template = (viewName: string, path: string) => `"${viewName}": () => import("${path}")`;
   const templates = [];
 
@@ -61,6 +65,10 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
     }
     if (clientFile) {
       templates.push(template(fileName, `/${clientFile?.file}`));
+      modulePreloadManifest[fileName] = collectModulePreloads(
+        manifest,
+        `app/views/${fileName}.tsx`,
+      );
     }
   }
 
@@ -180,6 +188,7 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
         return await result({
           getStyles,
           bootstrapModules: [`/${manifest["app/client.tsx"].file}`],
+          modulePreloadManifest,
           loaders,
           viewImportMap,
           viewModules,
--- a/packages/gemi/services/router/ViewRouteDispatcher.ts
+++ b/packages/gemi/services/router/ViewRouteDispatcher.ts
@@ -339,6 +339,11 @@ export class ViewRouteDispatcher {
        * carries real fallbacks, and they must match what the client hydrates.
        */
       viewModules?: Record<string, any>;
+      /**
+       * Built chunk URLs per view name, like `cssManifest`. Only the current
+       * route's chain is preloaded.
+       */
+      modulePreloadManifest?: Record<string, string[]>;
     }) => {
       const {
         bootstrapModules,
@@ -348,7 +353,23 @@ export class ViewRouteDispatcher {
         cssManifest,
         ogMap,
         viewModules,
+        modulePreloadManifest,
       } = params;
+
+      // Hydration reaches each route segment through `window.loaders`, i.e. an
+      // `import()` the browser cannot see until the entry has run — and it
+      // sees a nested chunk only once its parent has parsed. On a real
+      // network that serial chain shows every boundary's `Loading` fallback
+      // where server-rendered content already is, collapsing the page height
+      // and restoring it a round trip later (#352). Preloaded in the shell
+      // head, the whole chain is fetched in parallel with the entry and is
+      // resident before hydration asks for it.
+      const preloadHrefs = new Set<string>();
+      for (const view of currentViews ?? []) {
+        for (const href of modulePreloadManifest?.[view] ?? []) {
+          preloadHrefs.add(href);
+        }
+      }
 
       if (isOgRequest) {
         let ogHandler = null;
@@ -436,6 +457,16 @@ export class ViewRouteDispatcher {
         const stream = await renderToReadableStream(
           createElement(Fragment, {
             children: [
+              // React hoists these into `<head>`, ahead of the inlined
+              // stylesheets, so the browser starts the fetches on the shell's
+              // first bytes.
+              ...[...preloadHrefs].map((href) =>
+                createElement("link", {
+                  key: `modulepreload:${href}`,
+                  rel: "modulepreload",
+                  href,
+                }),
+              ),
               ...styles,
               createElement("script", {
                 key: "theme-script",
--- a/packages/gemi/services/router/ViewRouteDispatcher.ts
+++ b/packages/gemi/services/router/ViewRouteDispatcher.ts
@@ -329,7 +329,7 @@ export class ViewRouteDispatcher {
     return async (params: {
       getStyles: (p: string[]) => Promise<any[]>;
       viewImportMap: any;
-      bootstrapModules: string[];
+      bootstrapModules?: string[];
       loaders: string;
       cssManifest: Record<string, string[]>;
       ogMap: Record<string, any>;
@@ -340,19 +340,33 @@ export class ViewRouteDispatcher {
        */
       viewModules?: Record<string, any>;
       /**
+       * The client entry, when the server wants it booted from the bootstrap
+       * script instead of handed to React as a `bootstrapModule`: `module` is
+       * imported, `preload` is every chunk that import pulls in.
+       *
+       * React preloads its own `bootstrapModules` with `fetchPriority="low"`
+       * hardcoded (Fizz has no option for it), which parks the one module
+       * that unblocks hydration behind everything else the page is fetching.
+       * Emitting the `modulepreload` ourselves gets it the browser's default
+       * script priority, and it is the same request either way — the
+       * `import()` below consumes the preloaded module (#352).
+       */
+      clientEntry?: { module: string; preload: string[] };
+      /**
        * Built chunk URLs per view name, like `cssManifest`. Only the current
        * route's chain is preloaded.
        */
       modulePreloadManifest?: Record<string, string[]>;
     }) => {
       const {
-        bootstrapModules,
+        bootstrapModules = [],
         loaders,
         getStyles,
         viewImportMap,
         cssManifest,
         ogMap,
         viewModules,
+        clientEntry,
         modulePreloadManifest,
       } = params;
 
@@ -364,12 +378,21 @@ export class ViewRouteDispatcher {
       // and restoring it a round trip later (#352). Preloaded in the shell
       // head, the whole chain is fetched in parallel with the entry and is
       // resident before hydration asks for it.
-      const preloadHrefs = new Set<string>();
+      const preloadHrefs = new Set<string>(clientEntry?.preload ?? []);
       for (const view of currentViews ?? []) {
         for (const href of modulePreloadManifest?.[view] ?? []) {
           preloadHrefs.add(href);
         }
       }
+
+      const bootstrapScriptContent = (data: string) =>
+        [
+          data,
+          `window.loaders=${loaders}`,
+          // Nothing imports the entry when the server routed it through
+          // `clientEntry` — the bootstrap script is where it starts.
+          ...(clientEntry ? [`import(${JSON.stringify(clientEntry.module)})`] : []),
+        ].join(";");
 
       if (isOgRequest) {
         let ogHandler = null;
@@ -484,7 +507,9 @@ export class ViewRouteDispatcher {
             ],
           }),
           {
-            bootstrapScriptContent: `window.__GEMI_DATA__ = ${JSON.stringify(result.data)}; window.loaders=${loaders}`,
+            bootstrapScriptContent: bootstrapScriptContent(
+              `window.__GEMI_DATA__ = ${JSON.stringify(result.data)}`,
+            ),
             bootstrapModules,
             signal: deadline.signal,
             ...(settled ? { progressiveChunkSize: Number.MAX_SAFE_INTEGER } : {}),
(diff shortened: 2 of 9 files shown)


packages/gemi/server/httpProd.ts
```diff
@@ -61,11 +65,23 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
...
 
   const loaders = `{${templates.join(",")}}`;
 
+  // Booted from the bootstrap script rather than through React's
+  // `bootstrapModules`, which would preload it at `fetchPriority="low"` — see
+  // `clientEntry` in `ViewRouteDispatcher`.
+  const clientEntry = {
+    module: `/${manifest["app/client.tsx"].file}`,
+    preload: collectModulePreloads(manifest, "app/client.tsx"),
+  };
```
nstfkc: **`manifest["app/client.tsx"].file` is now dereferenced unguarded at boot.**

This expression used to live inside `handleWithApp`'s `try`/`catch`, so a manifest missing that key cost one document request a 500. Hoisted here it runs before `Bun.serve`, so a partial or stale `dist/` (a build interrupted between the server and client passes) throws `TypeError: undefined is not an object` and the process exits — no API routes, no static assets, no health check.

The inconsistency is visible in the same block: `collectModulePreloads` tolerates a missing entry (returns `[]`) and line 88 right below reads the identical key as `manifest["app/client.tsx"]?.css ?? []`. `ViewRouteDispatcher` already treats `clientEntry` as optional (`clientEntry?.preload`, `clientEntry ? [...] : []`), so degrading is cheap:

```suggestion
  const clientEntryFile = manifest["app/client.tsx"]?.file;
  const clientEntry = clientEntryFile
    ? {
        module: `/${clientEntryFile}`,
        preload: collectModulePreloads(manifest, "app/client.tsx"),
      }
    : undefined;
```

If silently not hydrating is worse than dying, the other defensible option is an explicit `throw new Error("client manifest is missing app/client.tsx — dist/ is incomplete")` here, which at least fails loudly instead of with a bare TypeError. Either beats the current unguarded deref.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -339,17 +339,61 @@ export class ViewRouteDispatcher {
...
+          data,
+          `window.loaders=${loaders}`,
+          // Nothing imports the entry when the server routed it through
+          // `clientEntry` — the bootstrap script is where it starts.
+          ...(clientEntry ? [`import(${JSON.stringify(clientEntry.module)})`] : []),
+        ].join(";");
+
       if (isOgRequest) {
         let ogHandler = null;
         let ogComponent = null;
```
nstfkc, line 458: **The OG branch iterates `currentViews` unguarded, two lines below code that establishes it can be `undefined`.**

Line 400 (just past this hunk) is `for (const view of currentViews) {`. On an unmatched route `currentPathName` stays `null`, so `currentViews = this.routeManifest[null]` is `undefined` — which is exactly why the new preload loop above correctly writes `currentViews ?? []`.

Reproduced: `GET http://gemi.dev/nope.og` through `app.fetch` rejects with `TypeError: undefined is not an object (evaluating 'currentViews')`. `httpProd.handleWithApp` catches it and returns a 500 whose body is the raw stack trace, so a crawler or link unfurler asking for the OG image of a dead URL gets a stack leak instead of the `new Response("Not found", { status: 404 })` sitting right below the loop.

Pre-existing, not introduced here — but the diff puts the correct guard directly above the missing one, so it is cheap to close now:

```js
for (const view of currentViews ?? []) {
```

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -339,17 +339,61 @@ export class ViewRouteDispatcher {
...
+      // Hydration reaches each route segment through `window.loaders`, i.e. an
+      // `import()` the browser cannot see until the entry has run — and it
+      // sees a nested chunk only once its parent has parsed. On a real
+      // network that serial chain shows every boundary's `Loading` fallback
+      // where server-rendered content already is, collapsing the page height
+      // and restoring it a round trip later (#352). Preloaded in the shell
+      // head, the whole chain is fetched in parallel with the entry and is
+      // resident before hydration asks for it.
+      const preloadHrefs = new Set<string>(clientEntry?.preload ?? []);
+      for (const view of currentViews ?? []) {
```
nstfkc: **404 documents get no view modulepreloads, so the `Loading` flash #352 describes survives on every not-found page.**

On an unmatched URL `currentPathName` is `null`, `currentViews` is `undefined`, and this loop contributes nothing — the head carries only the two client-entry hints. But the document *is* server-rendered, from the `404` view, and `httpProd` already collected that view's chunks into `modulePreloadManifest["404"]` at boot. So hydration still walks that chain one `import()` round trip per level and flashes `Loading` over already-rendered 404 content.

That is the one route class most likely to be hit cold and uncached. Falling back to the 404 view's chain when nothing matched — mirroring whatever `currentViews` the not-found render actually mounts — would make the fix complete rather than match-only.

packages/gemi/services/router/modulePreload.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    );
+  });
+
+  test("boots the entry from the bootstrap script, not a React bootstrap module", async () => {
+    const html = await fetchDocument("/app");
+
+    // React would preload its own `bootstrapModules` at `fetchPriority="low"`,
+    // which is the whole reason the entry is imported here instead.
+    expect(html).toContain(`import("/assets/client-entry.js")`);
+    expect(html).not.toContain(`fetchPriority="low"`);
```
nstfkc, line 148: **These two assertions are vacuous — this test never passes `bootstrapModules`, so React could not emit either string regardless of what `httpProd` does.**

`fetchDocument("/app")` uses `prodParams` (lines 58-68), which deliberately omits `bootstrapModules`. With none passed, React emits no bootstrap `<script type="module">` and no `<link rel="modulepreload" fetchPriority="low">` no matter what — so `not.toContain('fetchPriority="low"')` and `not.toContain('<script type="module"')` are true by construction. (The attribute spelling is right; it is the fixture that makes them toothless.)

Concretely: revert httpProd.ts:195-199 to `bootstrapModules: [\`/${manifest["app/client.tsx"].file}\`]` and drop `clientEntry` — the low-priority-preload regression from #352 is fully reintroduced in production and this suite, the only coverage for it, stays green. Nothing here exercises the `clientEntry`-vs-`bootstrapModules` decision, because that decision lives in `httpProd` and `httpProd` is never imported.

The assertion at line 123 (`toContain('import("/assets/client-entry.js")')`) does carry weight. To make the *negative* half real, the test has to reach the code that chooses: either exercise `httpProd`'s param construction, or assert that both cannot be set at once.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -339,17 +339,61 @@ export class ViewRouteDispatcher {
...
+        }
+      }
+
+      const bootstrapScriptContent = (data: string) =>
+        [
+          data,
+          `window.loaders=${loaders}`,
+          // Nothing imports the entry when the server routed it through
+          // `clientEntry` — the bootstrap script is where it starts.
+          ...(clientEntry ? [`import(${JSON.stringify(clientEntry.module)})`] : []),
```
nstfkc: **A bare `import()` with no `.catch` turns a failed boot into a silent unhandled rejection.**

Previously React emitted `<script type="module" src=... async>`, and a chunk that 404s after a deploy (cached HTML pointing at a hash that no longer exists) or an entry that throws at evaluation surfaced as a window `error`/`onerror` event — which page-level error reporters and the dev overlay can observe. As a bare expression statement inside the inline classic script, the same failure produces only an `unhandledrejection`, which most `window.onerror`-based reporters do not capture. The page then sits permanently unhydrated — no interactivity, no navigation — with nothing reported.

Rethrowing out of the microtask restores the old signal:

```suggestion
          ...(clientEntry
            ? [
                `import(${JSON.stringify(clientEntry.module)}).catch(function(e){setTimeout(function(){throw e})})`,
              ]
            : []),
```

packages/gemi/server/httpProd.ts
```diff
@@ -179,7 +195,8 @@ export async function httpProd(app: App, instrumentation: Instrumentation) {
 
         return await result({
           getStyles,
-          bootstrapModules: [`/${manifest["app/client.tsx"].file}`],
+          clientEntry,
+          modulePreloadManifest,
```
nstfkc, line 203: **`modulePreloadManifest` stops at the server, so client-side navigation keeps the serial chunk chain this PR removes from the first load.**

`cssManifest` is handed to the client through `result.data` (ViewRouteDispatcher.ts:442) precisely so navigation can resolve a route's CSS (`fetchRouteCSS`, ClientRouter.tsx:295). The module list has no such path. On navigation `ClientRouter.tsx:300` calls `loadViewModule(component)` per view, and each chunk still learns its own static imports only after it has arrived and parsed — so a `layout → view → components` route costs one round trip per level on *every* in-app navigation, holding `Link[data-pending]` and the transition open for exactly the interval the fix removed from the hard load.

The per-view lists are already computed at boot and already passed here. Shipping them alongside `cssManifest` and warming them (link injection, or `loadViewModule` on the prefetch path) would fix navigation with the same data instead of only the first paint. Reasonable to leave for a follow-up, but worth naming — the PR body frames this as *the* waterfall fix.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -339,17 +339,61 @@ export class ViewRouteDispatcher {
...
 
+      // Hydration reaches each route segment through `window.loaders`, i.e. an
+      // `import()` the browser cannot see until the entry has run — and it
+      // sees a nested chunk only once its parent has parsed. On a real
+      // network that serial chain shows every boundary's `Loading` fallback
+      // where server-rendered content already is, collapsing the page height
+      // and restoring it a round trip later (#352). Preloaded in the shell
+      // head, the whole chain is fetched in parallel with the entry and is
+      // resident before hydration asks for it.
+      const preloadHrefs = new Set<string>(clientEntry?.preload ?? []);
```
nstfkc: **The preload set and its `<link>` elements are rebuilt per request from data fixed at boot.**

`preloadHrefs` unions `clientEntry.preload` with `modulePreloadManifest[view]` across `currentViews` on every document request, and line 486 then builds one React element per href — all derived from boot-time values plus the route pattern. This class already precomputes exactly this shape of per-route data once in the constructor (`routeManifest`, `clientRouteManifest`, `componentTree`, `flatComponentTree`); a `Map<currentPathName, string[]>` filled lazily on first hit would drop a Set allocation, an O(views × chunks) union, and N `createElement` calls out of the per-request path.

It also sits above the `isOgRequest` early return at line 397, so `.og` requests — which render through satori and never emit a shell — build a set they discard.

packages/gemi/services/router/modulePreload.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      loaders: "{}",
+      cssManifest: {},
+      ogMap: {},
+    });
+
+    // Vite serves unbundled source in dev, so there is no chunk graph to
+    // announce — React's own bootstrap handling is left alone.
+    expect(html).toContain(`<script type="module" src="/app/client.tsx"`);
+    expect(html).not.toContain("/assets/");
+    expect(html).not.toContain("import(");
```
nstfkc: **This assertion passes only because the fixture supplies `loaders: "{}"`.**

Real dev shells always contain `import(`: `httpDev.ts:242-243` builds each loader as `` "Home": () => import("/app/views/Home.tsx") `` and that string is inlined verbatim into `bootstrapScriptContent`. So this does not establish what the comment above it claims (that dev never boots the *entry* through `import()`), and the obvious next improvement — giving the dev case a realistic `loaders` fixture, in a test file about module loading — turns it red with no defect present.

Asserting the actual claim instead:

```suggestion
    expect(html).not.toContain(`import("/app/client.tsx")`);
```

packages/gemi/server/httpProd.ts
nstfkc: Fixed in 64dbc5f. Went with degrade-and-log rather than an explicit throw — the deref cost a great deal more than the document it serves, and the dispatcher already renders fine without it. The construction moved into `createClientEntry` (`server/modulePreloads.ts`) so it is one tested unit rather than a shape rebuilt inline, and `httpProd` logs which key is missing when it comes back `undefined`.

packages/gemi/services/router/ViewRouteDispatcher.ts
nstfkc: Fixed in 64dbc5f — `for (const view of currentViews ?? [])`. Covered by `renders the .og 404 instead of throwing on an unmatched route`, and confirmed against the built template: `GET /definitely-not-a-route.og` now returns 404 with body `Not found`, where before it was a 500 carrying the stack.

packages/gemi/services/router/ViewRouteDispatcher.ts
nstfkc: Fixed in 64dbc5f. The preload views are `currentViews ?? ["404"]`, mirroring the `["404"]` `ClientRouterProvider` mounts. On the built template the 404 document now carries `/assets/404-zdDK4kQH.js` alongside the entry chunks, hydrates clean, and Chrome logs no "was preloaded but not used" — the hint is consumed.

packages/gemi/services/router/ViewRouteDispatcher.ts
nstfkc: Fixed in 64dbc5f, taking the suggestion as written. Asserted in `reports a failed entry import where window.onerror can see it`.

packages/gemi/services/router/modulePreload.test.ts
nstfkc: Fixed in 64dbc5f. Took the second option — the dispatcher now decides, so there is something for the test to exercise: `clientEntry` and `bootstrapModules` are explicitly exclusive (`const reactBootstrapModules = clientEntry ? [] : bootstrapModules`), which honouring both would have needed anyway since it boots the entry twice.

`ignores bootstrapModules when the server also offers a client entry` passes both in the fixture, so React would emit `fetchPriority="low"` and a module script unless the dispatcher deliberately withholds them — the negative assertions have teeth now. Confirmed by mutation: reverting that line turns it red.

The `httpProd` half is a call to `createClientEntry`, which `server/modulePreloads.test.ts` covers directly. `httpProd` itself still cannot be imported without a built `dist/`, so nothing asserts it calls that helper — instead I ran the real production build of `templates/saas-starter`: 0 occurrences of `fetchPriority="low"`, 0 of `<script type="module"`, and the entry booting through the `import().catch(...)`.

packages/gemi/services/router/modulePreload.test.ts
nstfkc: Fixed in 64dbc5f. Took the suggestion, and gave the dev case a realistic `loaders` fixture in the same change — `{"AppLayout": () => import("/app/views/AppLayout.tsx")}` — so the test states the claim it meant to and would not have been tripped by the improvement.

packages/gemi/services/router/ViewRouteDispatcher.ts
nstfkc: Fixed in 64dbc5f. Both halves: it moved below the `isOgRequest` early return, so a `.og` request no longer builds a set it discards, and it is built once per route in `modulePreloadLinks` — the React elements are cached too, since they are immutable and everything they derive from is boot-fixed.

Not a plain field, though: the cache is a `WeakMap` keyed on the `modulePreloadManifest` object. A process that ever renders against two of them would otherwise serve one route's chunks out of the other's cache — the test suite does exactly that, dev params and prod params through one dispatcher, which is how I noticed.

packages/gemi/server/httpProd.ts
nstfkc: Done rather than deferred — 0db4847. You are right that the PR frames this as *the* waterfall fix, and shipping it half-done would have left the same round-trip-per-level on every navigation for the sake of the first paint.

`modulePreloadManifest` now rides in the payload beside `cssManifest`, and `preloadRouteModules` announces the target route's chunks at the two points that already warm a route: the navigation subscriber (before `loadViewModule` starts the imports) and `prefetchRoute` (alongside `fetchRouteCSS`). `modulepreload` rather than `loadViewModule` on the prefetch path, since a speculative link should fill the HTTP cache without evaluating anything.

Verified over CDP against the built template: the landing page carries 7 hints, navigating to `/about` appends exactly one — `About-<hash>.js`, everything it shares with the landing route was already resident — and `history.back()` appends none (the dedup set is seeded from the shell's own hints, so a return trip does not accumulate dead `<link>` elements).

Issue #10, opened 2026-07-17
0.40.0 is unusable at runtime: server2 runtime ships types-only + exports map missing ./server, ./bun/preload, ./bun/plugin, ./config


nstfkc:
## Summary

The published **`gemi@0.40.0`** tarball is missing the Bun-native server runtime and the `exports` entries the CLI + app entry rely on. As a result an app cannot `dev`/`start` on it, even though `gemi build` succeeds.

Reproduced against both `npm pack gemi@0.40.0` and a fresh `bun install` of `0.40.0`.

## Symptoms

```
$ gemi start
Starting server...
error: preload not found "gemi/bun/preload"
```

If you get past that, the built `dist/server/server.mjs` next fails on `import { Server } from "gemi/server"` → `Cannot find module 'gemi/server'`.

Direct checks:

```
$ bun --preload gemi/bun/preload -e 'console.log("ok")'
error: preload not found "gemi/bun/preload"

$ bun -e 'await import("gemi/server")'
error: Cannot find module 'gemi/server'
```

## Root cause

**1. `dist/server2/` ships types-only — no compiled JS.**

```
$ ls node_modules/gemi/dist/server2
banner.d.ts  generateEtag.d.ts  httpDev.d.ts  httpProd.d.ts  index.d.ts
renderErrorPage.d.ts  Server.d.ts  styles.d.ts  types.d.ts  watchEnv.d.ts
# (only .d.ts — no .js anywhere)

$ grep -rl "httpProd\|httpDev" node_modules/gemi/dist --include='*.js'
# (nothing)
```

`Server`, `httpProd`, `httpDev`, `renderErrorPage`, `watchEnv`, etc. have no implementation in the package, so `new Server({ kernel, instrumentation }).start()` has nothing to run. `dist/server/index.js` exists but only exports `startProdServer` / `startDevServer` / `dev` — **not** `Server`.

**2. The `exports` map is missing subpaths the bin + app entry use.**

Published `package.json#exports` only has the 10 "dev" subpaths (`./http`, `./client`, `./app`, `./facades`, `./email`, `./vite`, `./runtime`, `./kernel`, `./services`, `./i18n`). Missing:

- `./bun/preload` — the bin does `bun --preload gemi/bun/preload ...` in `dev` and `start`
- `./bun/plugin`
- `./server` — `app/server.ts` does `import { Server } from "gemi/server"`
- `./config`

[...]

PR #358, opened 2026-08-08
Close the four defects the first Prisma port reported: #354, #355, #356, #357

### LINKED ISSUE #354: A to-one whose foreign key is on the child has no supported spelling for `update` / `delete` / `upsert`
#116 closed by making `planForeignSide` refuse `update` / `delete` / `upsert` **by name** on a to-one whose foreign key lives on the child, instead of letting them fail for looking like the wrong spelling. That refusal is correct and its message is good. It is also the whole of what exists: the three operands are still unimplemented, nothing tracks them, and `docs/orm.md`'s *Not in scope* list does not mention the shape — so a caller meets it at run time.

The first real Prisma application ported onto the ORM hit it in the first admin controller that touched the relation.

## The shape

```prisma
model Media { thumbnail Image? }

model Image {
  thumbnailForMediaId Int?   @unique
  thumbnailForMedia   Media? @relation(fields: [thumbnailForMediaId], references: [id], onDelete: Cascade)
}
```

Prisma accepts the nested upsert on the parent:

```ts
prisma.media.update({
  where: { id },
  data: { thumbnail: { upsert: { create: { src, width, height }, update: { src, width, height } } } },
  include: { thumbnail: true },
})
```

`Media.update` with the identical tree throws from `compile/nested-writes.ts:760`:

> 'thumbnail' is a to-one whose foreign key lives on Image, and 'upsert' is not implemented for that shape yet. On a to-one Prisma takes the data directly rather than the { where, data } form this side expects. Reach the row through the Image model directly, or use 'connect' and 'disconnect'.

It is thrown during compilation, so on **every** call — not only on a missing row, and not only under some argument shape. The port's diff changed the client and left the data tree alone, which means the endpoint went from working to unconditionally 500 with no type error.

## Measured in the port

Two live call sites, one admin-facing and one customer-facing, both `setThumbnail`-style handlers reached from real UI. The failure is worse than a plain 500 because of the ordering: the upload runs first, so `Storage.put(file)` **succeeds** and writes the blob, then the compile throws before a single statement is issued. The request fails, the thumbnail is never linked, and the blob is orphaned.

## The suggested workaround is real but not a substitute

"Reach the row through the Image model directly" works, and is what the port did:

```ts
await Image.upsert({
  where:  { thumbnailForMediaId: id },
  create: { src, width, height, thumbnailForMediaId: id },
  update: { src, width, height },
})
const media = await Media.findFirst({ where: { id }, include: { thumbnail: true } })
```

What it costs, relative to the nested write:

- **Two round trips instead of one**, and the parent's `include` has to be re-issued as a second read purely to rebuild the response shape.
- **Atomicity is now the caller's job.** Prisma's nested upsert was one unit; this is two statements that are only whole if the call site remembers `Model.transaction`.
- **The foreign key is written by hand at the call site** (`thumbnailForMediaId: id`), which is the thing nested writes exist to stop callers doing.

`connect` / `disconnect` do not cover it either: `upsert` here means *create the child row, or update the one that is already there*, and `connect` needs a child that already exists. There is no spelling of this relation's write that the ORM accepts.

## Why it deserves an issue rather than the docs line it does not have

*Not in scope* lists multi-field relations in a nested write as **"pending rather than declined"** and #271 exists behind it. This shape is in exactly that category — unimplemented, not decided against — and has neither the entry nor the issue. It throws the plain `UnsupportedQueryError`, correctly, and the *yet* has no number attached.

It is also the more likely of the two to be met: a to-one with the key on the child is Prisma's ordinary one-to-one, not a composite-key relation.

## Acceptance

- `update`, `delete` and `upsert` implemented on a foreign-side to-one, in Prisma's spelling as #116 already measured it off the generated client: `update` takes the data directly rather than `{ where, data }`, `delete` takes a boolean or a `WhereInput`, `upsert` takes `{ create, update }`.
- The write stays one unit, the way the existing nested path is one unit — a caller should not have to reach for `Model.transaction` to get back what the nested spelling gave them.
- The walk in #116's both-sides table extended to cover the three operands once they compile, so the two sides of a to-one cannot drift apart again.
- Until it lands: an entry in *Not in scope* pointing here, so the shape is discoverable while writing the code rather than while reading a 500.

### LINKED ISSUE #355: `update` with an empty `data` is refused, where Prisma reads the row back — and the read path it needs is already built
`compileUpdate` refuses an `update` whose `data` compiles to no assignments and no nested steps (`compile/write.ts:531-540`):

```ts
if (assignments.length === 0 && !relationOnly) {
  throw new UnsupportedQueryError("data", schema.name, op, "At least one field must be updated.");
}
```

Prisma accepts it: `update` with `data: {}` issues a read and returns the row unchanged — no `UPDATE`, no error. So a call that returned 200 with a row returns a 500 after the port.

## Why an empty `data` is not a silly input

It is what the ordinary partial-update pattern produces on its quiet branch:

```ts
const data = {
  ...(typeof alt === "string" ? { alt } : {}),
  ...(typeof fileName === "string" && fileName.trim() !== "" ? { fileName } : {}),
}
await Media.update({ where: { id }, data, include: { thumbnail: true } })
```

Those two guards exist *precisely* so a payload carrying neither field is tolerated. Under Prisma that branch was a no-op returning the row; under the ORM it is the only branch that throws. The defensive code became the hazard.

Reached by a `PUT` with body `{}` or `{ "fileName": "   " }` — a scripted edit, an API consumer, a form that later submits one field instead of two. The in-app form happened to always submit both, which is the only reason this was low severity in the port rather than an outage.

## The read path it needs is already built

The branch immediately above it does the work:

> An `update` whose `data` is *only* nested relation writes has no column to set — `Post.update({ where, data: { tags: { connect: … } } })` changes the join table and nothing on `Post` itself. Prisma accepts it, and there is no `UPDATE … SET` that expresses it.
>
> So the row is *selected* instead. […] Same plan shape, same steps, one statement that happens to read rather than write.

`relationOnly` already turns a no-assignment update into a select that honours the caller's `select` / `include`. The empty case differs from it by one clause. The comment that closes the branch is explicit that this was decided, not overlooked:

> Only reachable when there are steps to run: with neither an assignment nor a nested write there is genuinely nothing to do, and that stays an error.

"Genuinely nothing to do" is true of the *write*. It is not true of the *call*: a caller who passed `include` asked a question with a defined answer, and the select the branch above already emits answers it. The difference between the two cases is whether there are nested steps to run — which matters to the plan, but not to what the caller gets back.

## Two ways to close it

1. **Match Prisma.** Empty `data` takes the same select the relation-only path emits. Removes the divergence, reuses the mechanism, costs the one clause.
2. **Keep refusing, but say so properly.** `UnsupportedByDesignError` plus an entry in *Not in scope*, so it reads as a decision rather than a gap. Honest — but it puts an `Object.keys(data).length === 0` guard at every partial-update call site in every ported app, which is a lot of surface to defend a refusal that has no wrong answer behind it.

(1) is the better trade. Unlike the fractional-`take` divergence, there is no dialect disagreement or silently-wrong result being prevented here — the refusal buys nothing but strictness.

## Acceptance

- `Model.update({ where, data: {}, include })` returns the row unchanged: one `SELECT`, no `UPDATE`, the caller's `select` / `include` honoured — the plan shape the relation-only branch already produces.
- The no-`include` spelling pinned to whatever Prisma actually does, measured through the differential harness rather than reasoned about, so the two cases cannot disagree.
- A test that an empty `data` and a relation-only `data` produce the same plan shape, since that is the invariant the fix rests on.
- If (2) is chosen instead: `UnsupportedByDesignError`, an entry in *Not in scope*, and a line in the Prisma-migration notes — a refusal that ported code will meet on its defensive branch should not be discoverable only at run time.

### LINKED ISSUE #356: Refusing a fractional `take` is right, but a port has no way to find the query-string call sites it breaks
**This is not a request to change the decision.** `assertPageArgument` (`compile/paginate.ts:64`) refuses a non-integer `take` / `skip`, and the note above it earns that: SQLite hands back an opaque driver error and Postgres rounds `limit 1.5` to 2 rows, so matching Prisma's truncation would mean disagreeing with one dialect silently and also reworking #72's relation-node rule. One rule at both levels, failing loudly, is right. `docs/orm.md:456` records it as a divergence rather than leaving it to be found.

The gap is that the divergence is stated from the *implementer's* side, and there is nothing on the *porter's* side — no way to find the call sites before production does.

## Measured

In the first Prisma app ported onto the ORM, **9 controllers** derive pagination straight from the query string, all in the same two lines:

```ts
const limit = Number(search.limit) || 25
const page  = Number(search.page)  || 1
…
Model.findMany({ …, take: limit, skip: limit * (page - 1) })
```

`?limit=2.5` returned 2 rows under Prisma. After the port it throws `InvalidArgumentError` out of the query and the endpoint 500s. `?page=1.5` is worse to spot, because `page` never reaches the ORM — it reaches it multiplied, as a fractional `skip`.

Nothing in the toolchain catches this:

- **tsc is happy** — `Number(x)` is `number`, and `number` is what `take` wants. The type system cannot express "integer".
- **Unit tests are happy** — every test passes integer literals, because that is what a developer types.
- **The failing input is external** — a hand-edited URL, a shared link, a client that computed a page size, a typo. Nothing the app itself generates, so nothing in the app exercises it.

Two of the nine sites already truncated, for unrelated reasons. The other seven did not, and the two that were caught were caught by a human reading the port diff — after the port had been reviewed and was heading for merge.

## Why it is worth an issue and not just a truncation at each call site

The truncation *is* the fix, and it is one `Math.trunc` per site. The problem is finding the sites. A divergence whose only symptom is a 500 on an input the app never produces itself is exactly the kind that ships: it survives review, type-check and the whole unit suite, and surfaces as a rare production error long after the port, at which point nobody connects it to the ORM migration.

Compare the neighbouring divergences. `distinct` and `cursor` are refused at the call site the moment anyone writes them — you cannot get a `distinct` into production without meeting the refusal in development, because the argument is a literal in the source. A fractional `take` is a *value*, not an argument, so the refusal is invisible until a real request carries one.

## What would close it, cheapest first

- **A line in the divergence note naming the source.** It currently explains the SQLite/Postgres reasoning, which answers "why refuse" but not "where will this bite me". Something as short as *"`take` and `skip` derived from a query string must be truncated at the boundary — `Number(req.query.limit)` is the shape that breaks"* turns the note into something a porter can act on.
- **A Prisma-migration note** (`UPGRADE.md` or a migration section in `docs/orm.md`) listing this next to the other behaviours that change silently under a port, rather than only in the reference entry for `take`.
- **A helper in the public surface** — a `page` / `perPage` → `{ take, skip }` function that truncates and clamps once, that the docs point ports at. It makes the correct spelling shorter than the incorrect one, which is the only thing that reliably moves call sites.
- **The mechanical option**, if it is worth the weight: a lint rule flagging a non-literal `take` / `skip` that is not obviously integral. The failure mode is invisible by construction, so a mechanical check earns more here than a doc line does.

## Acceptance

- The fractional-`take` divergence names the query-string shape as the thing to look for, not only the dialect reasoning behind the refusal.
- The Prisma-migration path lists it, so it is found while planning a port rather than while reading a stack trace.
- Optionally the helper, so there is one obvious correct spelling to point at.
- The refusal itself unchanged.

### LINKED ISSUE #357: Moving a write onto the ORM silently kills any `code === "P2002"` guard the application already had
The decision not to mirror Prisma's error codes is documented at `orm/errors.ts:190`:

> DECISION — gemi defines its own error rather than mirroring Prisma's codes.
>
> Prisma raises `PrismaClientKnownRequestError` with code `P2002` […] Mirroring that would ease a migration for code already branching on `P2002` — but **nothing in this repository does (checked)**, and carrying a `P` code implies the rest of the taxonomy is implemented too […]

The reasoning holds and I am not asking for `P2002`. The premise is what needs revisiting: *this* repository has nothing branching on `P2002` because it never used Prisma at run time. The first application ported onto the ORM had **four** such guards in production code, and they are the highest-stakes code in the app — the concurrency recoveries.

## What the switch actually does to one

A guard, written the ordinary way:

```ts
function isUniqueConstraintError(error: unknown): boolean {
  return typeof error === "object" && error !== null
    && "code" in error && (error as { code?: unknown }).code === "P2002"
}
```

Duck-typed **deliberately** — the comment above it says so — because a fake `{ code: "P2002" }` is all a unit test needs to drive the race, and `instanceof` across module boundaries is a nuisance in tests.

Now swap `prisma.aICredit.create(…)` for `AICredit.create(…)` underneath it. `UniqueConstraintError` sets `model`, `operation`, `fields`, `constraint`, `name` and no `code` anywhere on the prototype chain. So:

- **No type error.** The guard takes `unknown`; that is the whole point of it.
- **No runtime error.** The guard returns `false` and the `throw error` arm fires — the arm written to absorb exactly this race.
- **No test failure.** The tests still inject `{ code: "P2002" }`, which production no longer produces. The suite goes on passing green while guarding nothing.
- **No symptom until the race happens**, which is by definition rare and load-dependent.

Four independent signals, all silent. In the port this reached review in a merge-ready PR and was caught by a human reading the diff — not by `tsc`, not by ~2,700 unit tests. The two dead guards sat on a credit-balance read and an idempotent credit purchase: under concurrency the first would 500 instead of returning the winner's row (the exact regression its comment says it exists to prevent), and the second would report failure for a purchase that had in fact landed.

## Why this is different from the other porting divergences

A refused argument is loud and immediate. A changed *error taxonomy* is the opposite: the code still compiles, still runs, still passes, and the branch that stops working is the one that only executes under conditions your tests do not reproduce. It is also progressive — an app mid-port has some writers on Prisma and some on the ORM, so for a while the same guard must match **both** taxonomies, and no single spelling of it is right for the whole repo at once.

That is the shape the port converged on, independently, in two services:

```ts
export const isUniqueConstraintError = (error: unknown) =>
  error instanceof UniqueConstraintError ||
  (isRecord(error) && error.code === "P2002")   // still-Prisma writers; drop when the port finishes
```

Every port of a Prisma app with any concurrency handling will write that function. It is a bridge with a known lifetime, and it is worth naming rather than leaving each port to discover the hard way.

## What would close it

- **Say it where a porter looks.** The decision currently lives in a source comment on the class. `docs/orm.md`'s error table (line 1575) describes what `UniqueConstraintError` *is*; neither it nor any migration note says *"if you have `code === "P2002"` anywhere, it stops matching the moment that write moves onto the ORM, silently"*.
- **Ship or document the bridge predicate**, so there is one obvious thing to swap to and one obvious thing to delete when the port completes — rather than each app inventing it, and inventing it once per service.
- **The mechanical check, which is worth the most here.** A lint rule or codemod flagging a `"P2002"` literal in a file that also imports from the ORM's model surface. The failure mode is invisible by construction; a doc line does not find existing guards, and a grep only helps someone who already knows to grep.
- **The testing half, which is easy to miss.** The guards go dead *and their tests keep passing*, because the tests mock the old error. Anything gemi offers for testing writes should raise a real `UniqueConstraintError`, so a stale `{ code: "P2002" }` mock fails loudly instead of certifying a dead branch.

## Acceptance

- The Prisma-migration path states the taxonomy switch and its silence, as a thing to search for before porting rather than a property to infer from the class comment.
- A named, documented bridge predicate for the mid-port window where both taxonomies are live.
- A mechanical way to find existing `P2002` guards — a lint rule, a codemod, or at minimum a documented grep in a migration checklist.
- The `errors.ts` decision comment updated: the "nothing in this repository does (checked)" premise is true of this repo and false of the applications it is for, and the comment currently reads as though it settled the question for both.

nstfkc:
Closes #354, closes #355, closes #356, closes #357.

Four defects reported from the first Prisma application ported onto the ORM. Two are missing implementations that make ported code fail at run time; two are silent divergences where the port compiles, passes its whole suite, and breaks only under conditions the tests do not reproduce. They land together because they overlap in the files they touch — `docs/orm.md`, `UPGRADE.md`, `orm/index.ts` and two test files are each touched by more than one.

## #354 — a to-one whose foreign key is on the child

`planForeignSide` refused `update` / `delete` / `upsert` by name whenever `relation.kind !== "many"`, so Prisma's ordinary one-to-one had no supported spelling for any of them.

It turned out to be translation rather than implementation. `conjoin` already returns the parent restriction alone when the caller's filter is absent, and `listOf(null)` is `[]` — so **a to-one is a to-many of one**, and the existing step bodies are already correct. `toOneOperand` rewrites the spellings and the operands fall through to them:

| operand | normalises to |
| --- | --- |
| `update: <data>` / `{ data }` / `{ where?, data }` | `{ where, data }` |
| `upsert: { create, update, where? }` | `{ where, create, update }` |
| `delete` / `disconnect`: `true` | `{}` — the link alone |
| `delete` / `disconnect`: `false` | `null` — no statement |
| `delete` / `disconnect`: `<filter>` | itself |

Measured off the generated client rather than reasoned about: `ProfileUpdateToOneWithWhereWithoutUserInput` carries an **optional** `where` and it is a `WhereInput`, not a unique key — so `assertNamedRows` cannot serve this path and `assertToOneFilter` replaces it there. A miss raises `RecordNotFoundError`, since Prisma answers P2025 and the differential harness compares failures by *kind*; the to-many refusal would have classified as `other` and scored a refusal as agreement with a miss.

### Three adjacent defects in the same code

- **A live wrong write.** `canonicalShape` records a non-literal boolean as `"boolean"` and `disconnect` is not in `LITERAL_KEYS`, so `disconnect: true` and `disconnect: false` shared one plan entry. The owning side refuses `false` at compile time and then pushes a constant `() => null` — so a cached `true` plan served `false` and **nulled the foreign key on a call that asked for nothing**, because the compile-time refusal never re-runs on a cache hit. Same bug class `LITERAL_KEYS` already documents for `omit` and `skipDuplicates`; it cannot join that set, which matches by name at any depth and would record a to-many `disconnect: { id: 1 }` verbatim, so the guard is boolean-only.
- **`disconnect: true` on the foreign side** was refused by `assertNamedRows` with a message about unique keys. Same operand family, closed by the same normaliser.
- **Arrays on a to-one** — `create: [...]`, `connect: [...]` and `connectOrCreate: [...]` compiled and wrote or repointed several rows through a relation that holds one.

`templates/saas-starter` had no to-one whose key is on the child — every child-side relation there is a list, which is exactly why #116 never reached this path — so `Profile` joins the schema with a nullable `@unique userId`, making all four operands reachable. Its migration is `prisma migrate diff` output, not hand-written.

## #355 — `update` with an empty `data`

Refused where Prisma reads the row back. The relation-only branch already emits the exact select this needs, through the same `plan()` call, honouring `select` / `include` / `_count`, with `ORTHROW` still raising on a miss. The fix deletes a condition.

`updateMany({ data: {} })` is measured and matched rather than documented away: `{ count: 0 }` as a **constant**, verified against a filter matching two of three rows, one matching none, and no `where` at all.

## #356 — the fractional `take` divergence had no porter-facing surface

**The refusal is unchanged and correct.** What was missing was any way to find the call sites it breaks: `tsc` cannot express "integer", every unit test passes integer literals, and the failing input is external. `paginate()` ships from `gemi/orm`, taking `unknown` because a query string is where the values come from, truncating and clamping so its output *cannot* be a value `assertPageArgument` refuses — asserted by running the real validator over a table of hostile inputs.

gemi's own documentation taught the breaking shape: `docs/routing.md`, `docs/controllers.md`, `docs/navigation.md` and `PartialRenderController` all carried `Number(req.search.get("limit")) || 25`. All four are rewritten. The divergence note now names that shape and the `page`-reaches-the-ORM-multiplied case, and the stale error block quoting `UnsupportedQueryError` is replaced by the `InvalidArgumentError` the code actually throws.

## #357 — the error taxonomy switches silently

`errors.ts` argued against mirroring `P2002` from the premise that *"nothing in this repository does (checked)"*. True of this repo, false of the applications it is for: a real port had four such guards, all in concurrency-recovery code, all dead after the switch — no type error, no runtime error, no test failure (the tests inject `{ code: "P2002" }`, which production no longer produces), and no symptom until the race. The conclusion stands; the premise is restated.

`isUniqueConstraintError` ships — `instanceof` **or** the name, which is why it is worth shipping over telling people to write `instanceof`: it survives duplicate module copies, the actual reason the ported app duck-typed. The `|| code === "P2002"` bridge for the mid-port window is documented in `UPGRADE.md` with a delete-when-done marker and deliberately **not** exported.

`bunx gemi migrate` gains two annotation passes — a `"P2002"` literal beside a model import, and a non-integer-literal `take` / `skip`.

## `@updatedAt` follows the column, not the call

Found by adversarial review of the above, then measured directly by seeding the column to the epoch and reading it back (Prisma 6.19.2, SQLite):

| call | Prisma |
| --- | --- |
| `data: {}` | epoch — not stamped |
| `data: { profile: { create: … } }` | epoch — not stamped |
| `upsert` hit, `update: {}` | epoch — not stamped |
| `updateMany({ data: {} })` | epoch, `{ count: 0 }` |
| `data: { name: "real" }` | **stamped** |
| `data: { organization: { connect: … } }` | **stamped** |

Stamping unconditionally was load-bearing in the wrong direction twice. It made #355's read dead on every model carrying the attribute — the majority case, and the one the issue was filed from — because the stamp *was* the assignment keeping the count off zero. And it put a timestamp Prisma does not write on every nested to-one write #354 adds, where the differential harness could not see it: `updatedAt` is volatile, so both sides compare as the descriptor `"date"` and two different instants match. **Six tests failed on the fix, and all six had been pinning gemi's divergence as though it were the specification.**

An owning-side `disconnect` stamps here and does not in Prisma, whose `connect` one operand over does. That is Prisma disagreeing with itself; matching it would mean special-casing one operand to reproduce an inconsistency. Recorded as a known divergence instead.

## Still open, pinned rather than described

Each asserts **both** clients' answers, so it fails the day either side moves:

- the owning side takes only `disconnect: true`, where Prisma also takes `false` and a filter;
- `create` on an occupied to-one collides, where Prisma detaches the incumbent.

## One judgement call worth a look

`expectSameWrite` now compares relation arrays as a multiset rather than positionally. Neither client emits an `ORDER BY` and Postgres relocates an updated tuple, so the positional assertion was passing on physical storage order and had begun failing at random. It is scoped to schema-declared relations, so Json arrays and scalar lists keep their ordered comparison, and a test pins that a genuinely different *set* still fails. It is the one assertion in this diff that got **smaller**.

## Verification

| | |
| --- | --- |
| `packages/gemi` | 3,014 passed |
| `templates/saas-starter` (SQLite) | 739 passed |
| Postgres (`app/models/postgres.sh`) | 1,116 passed |
| `tsc --noEmit` | clean |
| `oxlint orm/ bin/migrate/` | 0 warnings, 0 errors |

Every measurement is pinned as a differential case against a real Prisma client, so each answer is checked rather than remembered.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -193,7 +193,7 @@ import { useSearchParams } from "gemi/client";
 
 function Pagination() {
   const searchParams = useSearchParams();
-  const page = Number(searchParams.get("page")) || 1;
+  const page = Math.max(1, Math.trunc(Number(searchParams.get("page"))) || 1);
 
   function goTo(next: number) {
     searchParams.set("page", String(next));
@@ -202,6 +202,13 @@ function Pagination() {
   // …
 }
 ```
+
+Truncate a page number where you read it, rather than trusting the query string: it holds
+whatever the last link, or a hand edit, left there, and the value is incremented and sent back
+to the server, where a fractional page arrives multiplied — as a fractional `skip`, which the
+ORM refuses. The server-side counterpart is `paginate` from `gemi/orm` (see
+[ORM](./orm.md#querying)); it belongs to the query layer, so on the client the whole of the job
+is the line above.
 
 `.set` accepts either `(key, value)` or an object of updates, and the value may be a
 function of the current value. `.push("hard")` forces a full data reload; the default
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -193,7 +193,11 @@ import { useSearchParams } from "gemi/client";
 
 function Pagination() {
   const searchParams = useSearchParams();
-  const page = Math.max(1, Math.trunc(Number(searchParams.get("page"))) || 1);
+  // `Number.isFinite` first: `?page=1e400` is `Infinity`, which `Math.trunc`
+  // returns unchanged and `Math.max` happily keeps — and it would go straight
+  // back into the URL through `goTo` below.
+  const asked = Number(searchParams.get("page"));
+  const page = Number.isFinite(asked) ? Math.max(1, Math.trunc(asked)) : 1;
 
   function goTo(next: number) {
     searchParams.set("page", String(next));
--- a/packages/gemi/bin/migrate/index.ts
+++ b/packages/gemi/bin/migrate/index.ts
@@ -14,6 +14,7 @@ import type { ClassDecl, ClassMember } from "./classBody";
 import { isUsed, parseImports, printImport, stripLiterals } from "./imports";
 import type { ImportDecl } from "./imports";
 import { renameIdentifier } from "./lex";
+import { annotateOrmPortHazards } from "./orm";
 import {
   DELETED_EXPORTS,
   EXTRACTION_TARGETS,
@@ -195,13 +196,17 @@ export async function runMigrate(options: MigrateOptions) {
   // 4. Provider files whose contents now live in app/config are superseded.
   for (const file of migratedFiles) changes.push({ file, contents: null });
 
-  // 5. Renames across the rest of the app, plus retired fields in app/config.
+  // 5. Renames across the rest of the app, plus retired fields in app/config
+  //    and the two ORM-port hazards. The ORM pass runs over every file, not
+  //    only `app/config`: what it looks for is a guard or a query argument, and
+  //    those live wherever the app writes and reads.
   const touched = new Set(changes.map((change) => change.file));
   for (const file of walk(appDir)) {
     if (touched.has(file)) continue;
     const source = readFileSync(file, "utf8");
     let rewritten = rewriteReferences(source, rel(rootDir, file), notes);
     rewritten = annotateRetiredFields(rewritten, file, rootDir, notes);
+    rewritten = annotateOrmPortHazards(rewritten, rel(rootDir, file), notes);
     if (rewritten !== source) changes.push({ file, contents: rewritten });
   }
 
--- a/packages/gemi/bin/migrate/index.ts
+++ b/packages/gemi/bin/migrate/index.ts
@@ -196,18 +196,60 @@ export async function runMigrate(options: MigrateOptions) {
   // 4. Provider files whose contents now live in app/config are superseded.
   for (const file of migratedFiles) changes.push({ file, contents: null });
 
-  // 5. Renames across the rest of the app, plus retired fields in app/config
-  //    and the two ORM-port hazards. The ORM pass runs over every file, not
-  //    only `app/config`: what it looks for is a guard or a query argument, and
-  //    those live wherever the app writes and reads.
+  // 5. Renames across the rest of the app, plus retired fields in app/config.
   const touched = new Set(changes.map((change) => change.file));
   for (const file of walk(appDir)) {
     if (touched.has(file)) continue;
     const source = readFileSync(file, "utf8");
     let rewritten = rewriteReferences(source, rel(rootDir, file), notes);
     rewritten = annotateRetiredFields(rewritten, file, rootDir, notes);
-    rewritten = annotateOrmPortHazards(rewritten, rel(rootDir, file), notes);
     if (rewritten !== source) changes.push({ file, contents: rewritten });
+  }
+
+  // 6. The two ORM-port hazards, over every file this run will leave behind —
+  //    which is a wider set than the loop above walks, and deliberately so.
+  //
+  //    Step 4 supersedes the provider files by moving their bodies into
+  //    `app/config/*.ts`, and steps 2 and 3 rewrite `Kernel.ts` and write
+  //    `AppServiceProvider.ts`. All of those land in `changes`, so the loop
+  //    above skips them — and they hold *user* code, carried over verbatim from
+  //    a provider's `boot()`. On the one run that matters, the first run on an
+  //    old-layout app, a `"P2002"` guard or a query-string `take` inside any of
+  //    them would never have been looked at. A second run finds them, but only
+  //    because by then the passes that moved them have become no-ops, which is
+  //    not a property to rely on.
+  //
+  //    So this runs last and over `changes` as well as over the tree: what the
+  //    pass looks for is a guard or a query argument, and those live wherever
+  //    the app writes and reads, including in a file this command just created.
+  const pending = new Map<string, number>();
+  const removed = new Set<string>();
+  changes.forEach((change, index) => {
+    if (change.contents === null) removed.add(change.file);
+    else pending.set(change.file, index);
+  });
+
+  const swept = new Set<string>();
+  for (const file of walk(appDir)) {
+    if (removed.has(file)) continue;
+    swept.add(file);
+    const index = pending.get(file);
+    const before =
+      index === undefined ? readFileSync(file, "utf8") : changes[index]!.contents!;
+    const after = annotateOrmPortHazards(before, rel(rootDir, file), notes);
+    if (after === before) continue;
+    if (index === undefined) changes.push({ file, contents: after });
+    else changes[index]!.contents = after;
+  }
+
+  // A file `changes` creates does not exist yet, so `walk` cannot yield it.
+  for (const change of changes) {
+    if (change.contents === null || swept.has(change.file)) continue;
+    change.contents = annotateOrmPortHazards(
+      change.contents,
+      rel(rootDir, change.file),
+      notes,
+    );
   }
 
   // Apply.
--- a/packages/gemi/bin/migrate/orm.ts
+++ b/packages/gemi/bin/migrate/orm.ts
@@ -16,6 +16,26 @@
 // integer or whether the `catch` this `"P2002"` sits in wraps an ORM call or a
 // Prisma one. A rewrite would need that. A sentence beside the line does not,
 // and a sentence beside the line is what both issues asked for.
+//
+// **Except in a file that can hold JSX, where a `//` line is not a comment.**
+// Inside JSX children it is page text, and the `{ code: "P2002" }` this module's
+// own message contains becomes an expression container — so splicing one in
+// stops the file parsing. Measured, on the obvious shape:
+//
+//     {error?.code === "P2002" ? <p>Email taken</p> : null}
+//
+// annotated and fed to Bun's `tsx` loader answers `Expected "}" but found ":"`.
+// A migration command that leaves the app unable to build is worse than one that
+// says nothing, and the `take` pass is exposed further still — it has no import
+// gate, and its message parses in children position, so it would have rendered a
+// paragraph of migration advice onto the page.
+//
+// The gate is the file extension, and that is a guarantee rather than a guess:
+// TypeScript permits JSX only in `.tsx` / `.jsx`, so a `.ts` file cannot contain
+// any. Those files are *reported* instead, with line numbers, which is what the
+// splice was carrying anyway. Neither pass can hit inside a string or template
+// literal — `skipAtomic` steps over both before either finder looks — so on a
+// file that cannot hold JSX every hit is ordinary code and the splice is safe.
 
 import { parseImports } from "./imports";
 import { matchDelims, skipAtomic, skipTrivia } from "./lex";
@@ -91,6 +111,7 @@ export function annotateOrmPortHazards(
   label: string,
   notes: Notes,
 ): string {
+  const spliceable = !JSX_FILE.test(label);
   let result = source;
 
   // The import gate is #357's, and it is what keeps this pass worth reading. A
@@ -105,15 +126,15 @@ export function annotateOrmPortHazards(
   // here. UPGRADE.md carries the plain grep for that, because a codemod that
   // followed call graphs would need the parser this command does not have.
   if (importsModelSurface(result)) {
-    const hits = prismaErrorCodeLiterals(result);
-    const marked = annotateLines(result, hits, P2002_HAZARD);
-    result = marked.text;
-    if (marked.annotated > 0) {
-      notes.push({
-        file: label,
-        message: `\`"P2002"\` (${occurrences(hits.length)}) — ${P2002_HAZARD}`,
-      });
-    }
+    result = runPass(
+      result,
+      prismaErrorCodeLiterals(result),
+      '`"P2002"`',
+      P2002_HAZARD,
+      label,
+      notes,
+      spliceable,
+    );
   }
 
   // No import gate on this one, on purpose. A fractional `skip` is produced
@@ -121,25 +142,83 @@ export function annotateOrmPortHazards(
   // file holding `take: limit` need not import a model at all — it is often a
   // repository or a service that takes the numbers and passes them along. The
   // gate would remove exactly the hop that makes the value hard to trace.
-  {
-    const hits = nonIntegerPageArguments(result);
-    const marked = annotateLines(result, hits, PAGE_ARGUMENT_HAZARD);
-    result = marked.text;
-    if (marked.annotated > 0) {
-      notes.push({
-        file: label,
-        message:
-          `\`take\` / \`skip\` (${occurrences(hits.length)}) — ` +
-          PAGE_ARGUMENT_HAZARD,
-      });
-    }
-  }
+  result = runPass(
+    result,
+    nonIntegerPageArguments(result),
+    "`take` / `skip`",
+    PAGE_ARGUMENT_HAZARD,
+    label,
+    notes,
+    spliceable,
+  );
 
   return result;
 }
 
+/**
+ * TypeScript accepts JSX only in these two extensions, which is what makes the
+ * test a guarantee rather than a heuristic: a `.ts` file cannot contain JSX, so
+ * it cannot contain a position where a `//` line means something else.
+ */
+const JSX_FILE = /\.[jt]sx$/i;
+
+/**
+ * One pass: splice where that is safe, report where it is not.
+ *
+ * The report is not a lesser outcome. It carries the line numbers the splice
+ * carried by sitting on them, and it is what the `--dry-run` reader sees either
+ * way; what it loses is surviving into the diff, which is worth less than the
+ * file continuing to parse.
+ */
+function runPass(
+  src: string,
+  hits: number[],
+  what: string,
+  message: string,
+  label: string,
+  notes: Notes,
+  spliceable: boolean,
+): string {
+  if (!hits.length) return src;
+
+  if (!spliceable) {
+    notes.push({
+      file: label,
+      message:
+        `${what} on ${lineList(src, hits)} — reported rather than annotated: ` +
+        `a \`//\` line inside JSX children is page text, not a comment, so ` +
+        `writing one here would stop the file parsing. ${message}`,
+    });
+    return src;
+  }
+
+  const marked = annotateLines(src, hits, message);
+  if (marked.annotated > 0) {
+    notes.push({
+      file: label,
+      message: `${what} (${occurrences(hits.length)}) — ${message}`,
+    });
+  }
+  return marked.text;
+}
+
 const occurrences = (count: number) =>
   count === 1 ? "1 occurrence" : `${count} occurrences`;
+
+/** `line 12` / `lines 12, 40` — 1-based, deduplicated, in source order. */
+function lineList(src: string, offsets: number[]): string {
+  const numbers: number[] = [];
+  for (const offset of offsets) {
+    let line = 1;
+    for (let i = 0; i < offset && i < src.length; i++) {
+      if (src[i] === "\n") line++;
+    }
+    if (numbers[numbers.length - 1] !== line) numbers.push(line);
+  }
+  return numbers.length === 1
+    ? `line ${numbers[0]}`
+    : `lines ${numbers.join(", ")}`;
+}
 
 // ---------------------------------------------------------------------------
 // #357 — a `"P2002"` literal in a file that also imports the model surface
--- a/packages/gemi/bin/migrate/orm.ts
+++ b/packages/gemi/bin/migrate/orm.ts
@@ -1,0 +1,356 @@
+// The two ORM-port passes: what a Prisma application walks into when a read or
+// a write moves onto gemi's ORM and nothing anywhere says so.
+//
+// Both came off the same port (#356, #357) and both share the shape that makes
+// a documentation line insufficient. The code keeps compiling, keeps running
+// and keeps passing its tests; the thing that broke only surfaces under a
+// condition the application itself never produces — a race, or a hand-edited
+// URL. `tsc` cannot see either (`Number(x)` is a `number`, and a guard over
+// `unknown` is supposed to accept anything), and a grep only helps somebody who
+// already knows what to grep for. So the check has to sweep the tree unprompted,
+// which is the one thing this command already does.
+//
+// Annotate-only, like every other pass in this command. `lex.ts` is a hand-
+// rolled scanner rather than a parser: it can tell a string literal from a
+// comment and match a bracket, and it cannot tell whether `limit` holds an
+// integer or whether the `catch` this `"P2002"` sits in wraps an ORM call or a
+// Prisma one. A rewrite would need that. A sentence beside the line does not,
+// and a sentence beside the line is what both issues asked for.
+
+import { parseImports } from "./imports";
+import { matchDelims, skipAtomic, skipTrivia } from "./lex";
+import { TODO } from "./tables";
+
+/**
+ * The report sink `runMigrate` collects into. Spelled structurally rather than
+ * imported, because `index.ts` imports this module and naming its `Note` here
+ * would close the cycle.
+ */
+type Notes = { file: string; message: string }[];
+
+/**
+ * What a `"P2002"` guard becomes once the write under it is a gemi write.
+ *
+ * The sentence has to carry the *silence*, not just the fact. Every other thing
+ * this command flags announces itself — a missing export is a compile error, a
+ * renamed field is a type error. This one has four independent signals and all
+ * four stay green, so a reader who is told only "the code changed" will
+ * reasonably conclude their guard is fine because their tests pass. It is the
+ * tests passing that is the symptom.
+ */
+const P2002_HAZARD =
+  '`"P2002"` stops matching the moment the write it guards moves onto the ORM. ' +
+  "gemi raises `UniqueConstraintError`, which carries `model`, `operation`, " +
+  "`fields` and `constraint` and no `code` anywhere on its prototype chain — so " +
+  "the check returns false, the recovery branch it exists to run never runs, and " +
+  "nothing reports it: it compiles, it runs, and a test that mocks " +
+  '`{ code: "P2002" }` still passes while the branch it certifies is dead. Use ' +
+  "`isUniqueConstraintError` from `gemi/orm`. While some writes in this app are " +
+  "still on Prisma, keep the old check as a second arm — " +
+  '`isUniqueConstraintError(e) || (isRecord(e) && e.code === "P2002")` — and ' +
+  "delete that arm when the last Prisma writer goes. See UPGRADE.md.";
+
+/**
+ * What a non-literal `take` / `skip` might be.
+ *
+ * Deliberately phrased as a request to confirm rather than a claim of a bug.
+ * Most non-literal `take`s are fine — they are a clamped constant, a page size
+ * the app itself computed, a value already truncated — and this pass cannot
+ * tell those from the broken ones, because the difference is a runtime value.
+ * An annotation that asserted a defect would be wrong most of the times it
+ * fires, and a marker that is wrong most of the time is one people learn to
+ * skip past, which costs more than the pass earns.
+ *
+ * So it names the shape that actually breaks, and it names `page` louder than
+ * `take`: `page` never reaches the ORM at all, it reaches it multiplied, as a
+ * fractional `skip`, which is why the ported application's reviewers found the
+ * `take` sites and missed the `page` ones.
+ */
+const PAGE_ARGUMENT_HAZARD =
+  "`take` and `skip` must be integers — gemi refuses a fractional one with " +
+  "`InvalidArgumentError` where Prisma truncated it, because SQLite answers " +
+  "`limit 1.5` with an opaque driver error and Postgres rounds it up to 2 rows. " +
+  "This value is not an integer literal, which is usually fine: confirm it is " +
+  "truncated where it enters the app rather than reading this as a bug. " +
+  '`Number(req.search.get("limit"))` is the shape that breaks, and a fractional ' +
+  "`page` is the one to look hardest for — it never reaches the ORM itself, it " +
+  "reaches it multiplied, as a fractional `skip`. `paginate({ page, perPage })` " +
+  "from `gemi/orm` truncates and clamps in one place and cannot produce a value " +
+  "the ORM refuses.";
+
+/**
+ * Runs both passes over one file under `app/`.
+ *
+ * Sequential rather than merged: each pass re-scans the text the previous one
+ * produced, which is free (these files are small) and removes the only way the
+ * two could interfere — an offset computed against the pre-insertion text being
+ * applied to the post-insertion one.
+ */
+export function annotateOrmPortHazards(
+  source: string,
+  label: string,
+  notes: Notes,
+): string {
+  let result = source;
+
+  // The import gate is #357's, and it is what keeps this pass worth reading. A
+  // `"P2002"` in a file with no model import is either still-Prisma code, where
+  // the guard is correct exactly as written, or an unrelated string; annotating
+  // those teaches the reader that the marker means nothing. A file mid-port
+  // imports both `@prisma/client` and the app's models, and that file *is* the
+  // one that needs the bridge — so the gate asks only about the model side.
+  //
+  // Its false negative is real and deliberate: a guard living in a shared
+  // `lib/errors.ts` that imports nothing from the model surface is not found
+  // here. UPGRADE.md carries the plain grep for that, because a codemod that
+  // followed call graphs would need the parser this command does not have.
+  if (importsModelSurface(result)) {
+    const hits = prismaErrorCodeLiterals(result);
+    const marked = annotateLines(result, hits, P2002_HAZARD);
+    result = marked.text;
+    if (marked.annotated > 0) {
+      notes.push({
+        file: label,
+        message: `\`"P2002"\` (${occurrences(hits.length)}) — ${P2002_HAZARD}`,
+      });
+    }
+  }
+
+  // No import gate on this one, on purpose. A fractional `skip` is produced
+  // where the query string is read and consumed several layers away, so the
+  // file holding `take: limit` need not import a model at all — it is often a
+  // repository or a service that takes the numbers and passes them along. The
+  // gate would remove exactly the hop that makes the value hard to trace.
+  {
+    const hits = nonIntegerPageArguments(result);
+    const marked = annotateLines(result, hits, PAGE_ARGUMENT_HAZARD);
+    result = marked.text;
+    if (marked.annotated > 0) {
+      notes.push({
+        file: label,
+        message:
+          `\`take\` / \`skip\` (${occurrences(hits.length)}) — ` +
+          PAGE_ARGUMENT_HAZARD,
+      });
+    }
+  }
+
+  return result;
+}
+
+const occurrences = (count: number) =>
+  count === 1 ? "1 occurrence" : `${count} occurrences`;
+
+// ---------------------------------------------------------------------------
+// #357 — a `"P2002"` literal in a file that also imports the model surface
+// ---------------------------------------------------------------------------
+
+/**
+ * Offsets of every `"P2002"` string literal, template included.
+ *
+ * Only `P2002`, not the `P` codes in general. It is the one whose replacement
+ * gemi actually ships a predicate for, and a marker that fires on `P2025` while
+ * having nothing to offer instead would be a complaint rather than a migration
+ * step.
+ *
+ * Comments are stepped over rather than searched, and that is load-bearing for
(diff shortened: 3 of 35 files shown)


packages/gemi/bin/migrate/orm.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  const out: string[] = [];
+  let annotated = 0;
+  for (let index = 0; index < lines.length; index++) {
+    const line = lines[index]!;
+    if (targets.has(index)) {
+      const previous = out[out.length - 1] ?? "";
+      if (!previous.includes(message)) {
+        out.push(`${/^[ \t]*/.exec(line)![0]}${TODO} ${message}`);
```
nstfkc, line 427: **`annotateOrmPortHazards` splices `//` comments into `.tsx` files, breaking the parse**

`annotateOrmPortHazards` inserts a `//` line comment above the hit line, but `runMigrate` feeds it the `.tsx` files that `walk(appDir)` yields. Inside JSX children the comment is not a comment — it becomes JSX text, and the `{ code: "P2002" }` inside the message becomes an expression container, so the file stops parsing.

**Failure scenario:** a ported app has `app/views/SignupForm.tsx` containing `import { User } from "@/app/models"` and, inside JSX, `{error?.code === "P2002" ? <p>Email taken</p> : null}`. `bunx gemi migrate` (`index.ts:209` runs the new pass over every file from `walk(appDir)`, which yields `.tsx`) rewrites the file with the TODO line spliced between `<div>` and that child. Verified against the branch by calling `annotateOrmPortHazards` on exactly that source and feeding the result to Bun's `tsx` transpiler: `PARSE FAILED: Expected "}" but found ":"`. The user's app no longer builds after running the migration command.

The take/skip pass is worse-exposed because it has no import gate: its message contains `paginate({ page, perPage })`, which in children position parses but yields `Cannot find name 'page'` plus the whole paragraph rendered as visible text on the page.

<sub>Related sites with the same root cause: `packages/gemi/bin/migrate/orm.ts:208`.</sub>

packages/gemi/orm/types.ts
```diff
@@ -648,17 +648,41 @@ type NestedUpdate<R extends RelationInfo> = NestedCreate<R> &
         upsert?: NestedUpsert<R["target"]> | NestedUpsert<R["target"]>[];
       }
     : {
-        disconnect?: boolean | WhereUniqueInput<R["target"]>;
-        delete?: boolean | WhereUniqueInput<R["target"]>;
-        update?: UpdateInput<R["target"]> | NestedUpdateOne<R["target"]>;
-        upsert?: Omit<NestedUpsert<R["target"]>, "where">;
+        disconnect?: boolean | WhereInput<R["target"]>;
+        delete?: boolean | WhereInput<R["target"]>;
+        update?: UpdateInput<R["target"]> | NestedUpdateToOne<R["target"]>;
```
nstfkc, line 653: **Owning-side to-one `update` accepts a `where` it then ignores, writing unconditionally**

`NestedUpdateToOne` (optional, non-unique `where`) now types owning-side to-one relations too, but `planOwningSide`'s update step never reads `operandAt.where` — it writes the linked row unconditionally, while the foreign side this PR just implemented conjoins the same filter and raises on a miss.

**Failure scenario:** `User.update({ where: { id: 1 }, data: { organization: { update: { where: { name: "Acme" }, data: { name: "Renamed" } } } } })` now type-checks (before this diff the to-one `update` operand was `NestedUpdateOne` with a *required* `where: WhereUniqueInput`). At runtime `planOwningSide`'s step builds `where: { [referenced]: linked }` from the parent's foreign key alone (`nested-writes.ts:527-540`) and issues `Organization.updateMany` with the caller's filter dropped, so the linked organization is renamed even when its name is not "Acme". Prisma answers P2025 for the same call, and the identical operand written through `User.profile` (foreign side) is conjoined and raises `RecordNotFoundError`. One spelling, two answers — and the owning one is a silent wrong write.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2205,3 +2468,87 @@ function listOf(value: unknown): unknown[] {
...
+    // arrived at from the opposite direction: there the key was too coarse for
+    // the code, here the code was finer than the key. Both are decided by
+    // asking what `canonicalShape` actually records.
+    if (
+      operand !== null &&
+      typeof operand === "object" &&
+      (operand as Record<string, unknown>).data !== undefined
+    ) {
+      const record = operand as Record<string, unknown>;
+      return { where: record.where, data: record.data };
```
nstfkc, line 2682: **`toOneOperand` silently drops every key but `where`/`data`, so a typo'd filter becomes an unconditional write**

`toOneOperand`'s wrapper branch rebuilds the operand as `{ where, data }` and silently drops every other key, so a misspelled `where` turns a filtered nested update into an unconditional one instead of being refused.

**Failure scenario:** `User.update({ where: { id: 1 }, data: { profile: { update: { data: { bio: "x" }, wehre: { bio: "old" } } } } })`. The `data` key is present, so the wrapper branch fires and returns `{ where: record.where /* undefined */, data: record.data }` — `wehre` is discarded without a word. `conjoin(undefined, link)` then reduces to the parent link alone, so the connected `Profile` is updated unconditionally.

Prisma rejects the same call (`Unknown argument 'wehre'`), and gemi's own `assertConnectOrCreateOperand` refuses extra keys by name for exactly this reason. The result is a write that should not have happened, with no error anywhere: the caller wrote a conditional nested update guarded by a filter and got an unguarded one that also cannot report the P2025 the non-matching filter was supposed to produce.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1945,24 +2149,28 @@ function assertNamedUpdates(
...
         );
       }
     }
 
-    matchUniqueKey(child, record.where, {
-      model: schema.name,
-      operation,
-      argument: `data.${relation.name}.update.where`,
-    });
+    if (many) {
```
nstfkc, line 2212: **To-one `update`/`upsert` never validate `where`, so a non-object filter is silently dropped**

On the to-one path `matchUniqueKey` is skipped for the `where` inside `update` and `upsert`, and nothing replaces it — `assertToOneFilter` only guards `delete`/`disconnect`, whose operand *is* the filter — so a `where` that is not an object is silently dropped by `conjoin` instead of refused.

**Failure scenario:** `User.update({ where: { id: 1 }, data: { profile: { upsert: { where: [{ bio: "old" }], create: { bio: "new" }, update: { bio: "touched" } } } } })` compiles. At execution `conjoin([{bio:"old"}], { userId: 1 })` hits its `Array.isArray(filter)` arm and returns the parent link alone, so the lookup finds the connected Profile and the **update** branch overwrites `bio` with `"touched"`. Prisma applies the filter, finds no match, runs the **create** branch and raises P2002 on `Profile.userId`.

Same silent drop for `profile: { update: { where: 7, data: {...} } }` — a caller carrying the to-many `where: { id }` habit over, or a filter variable that arrived as `null`/a number: the update lands on whatever child is connected rather than being refused. Before this PR the whole operand was refused by name, and on the to-many sibling `matchUniqueKey` still rejects a non-object; the to-one branch validates the `where` nowhere at all.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2205,3 +2468,87 @@ function listOf(value: unknown): unknown[] {
...
+    // asking what `canonicalShape` actually records.
+    if (
+      operand !== null &&
+      typeof operand === "object" &&
+      (operand as Record<string, unknown>).data !== undefined
+    ) {
+      const record = operand as Record<string, unknown>;
+      return { where: record.where, data: record.data };
+    }
+    return { where: undefined, data: operand };
```
nstfkc, line 2684: **Non-object to-one `update` operands escape plan-time validation and fail mid-transaction**

A to-one nested `update` whose operand is not an object (`null`, a number, a string) escapes plan-time validation, because `toOneOperand` wraps it as `{ where: undefined, data: <scalar> }` and `assertNamedUpdates(many=false)` only asks that `data !== undefined`.

**Failure scenario:** `User.update({ where: { id: 1 }, data: { profile: { update: null } } })`. `planOne` only filters `undefined`, so `null` reaches `planForeignSide`; `toOneOperand` falls through both object tests and returns `{ where: undefined, data: null }`; `assertNamedUpdates` sees an object whose `data` key is present (value `null`) and passes. Nothing is refused at compile time.

The parent `UPDATE` then executes, and the `after` step calls `executor.exec("Profile", "updateMany", { where, data: null })`, which hits the *new* `args?.data === null` guard in `compileUpdate` and throws `InvalidArgumentError("data", "Profile", "updateMany", …)`. `update: 5` behaves the same way via `updateAssignments`' "Expected an object."

So the caller gets a mid-transaction failure that has to unwind an already-written parent row — exactly what this file's other operands document plan-time validation as existing to prevent — and the error names a model (`Profile`) and an operation (`updateMany`) that appear nowhere in the query they wrote, instead of `data.profile.update` on `User.update`. The to-many sibling refuses the same operand at compile time with the correct path.

templates/saas-starter/app/models/differential.ts
```diff
@@ -482,10 +578,23 @@ export async function createDifferential(options: {
...
+        // touched within the Postgres heap.
+        //
+        // Only here, not in `expectSame` above. A read leaves the heap alone, so
+        // the two clients there see the same physical order and comparing it
+        // positionally still holds — and that comparison is worth keeping,
+        // because a relation strategy that returned children in a different
+        // order from Prisma's would be a real difference to an application even
+        // though no `ORDER BY` promises otherwise. It is only a *write* that
+        // makes the order move under the comparison.
+        const comparable = (value: unknown) =>
```
nstfkc, line 635: **`expectSameWrite`'s new sort masks ordering divergences in includes that asked for `orderBy`**

`expectSameWrite` now sorts every schema-declared relation array before comparing, which also discards the order of nested includes that explicitly asked for one via `orderBy` — an assertion the positional `toEqual` used to make.

**Failure scenario:** six write cases in `writes.differential.test.ts` (lines ~731-785, ~2164) pass `include: { tags: { orderBy: { id: "asc" } } }`. If gemi's implicit-m-n include strategy stopped honouring the nested `orderBy` — returned tags in join-table insertion order, or descending — `stabilizeRelations` sorts both sides by `JSON.stringify(row)` first, so the two lists compare equal and the suite stays green.

An application calling `Post.create({ data: { tags: { connect: [...] } }, include: { tags: { orderBy: { id: "asc" } } } })` then receives `[{id:10},{id:2}]` where Prisma returns `[{id:2},{id:10}]`, and the differential harness that exists to catch exactly that divergence reports agreement. The JSON-string comparator makes the masking total: `"{\"id\":10" < "{\"id\":2"`, so both sides are re-ordered identically-wrongly.

The previous positional comparison caught this; the flake being fixed only ever affected *unordered* includes, so the sort should have been skipped where the case named an `orderBy`.

packages/gemi/orm/page-args.ts
```diff
@@ -0,0 +1,188 @@
...
+    perPage?: number;
+    /** The largest page size a *request* may ask for. Default 100. */
+    maxPerPage?: number;
+  } = {},
+): { take: number; skip: number } {
+  // The options are code rather than user input, so it is tempting to trust
+  // them. They are clamped anyway, because the guarantee this function makes is
+  // unconditional — `paginate(args, { perPage: 0 })` returning `take: 0` would
+  // be an empty page for ever, with the helper's own name on it.
+  const maxPerPage = Math.max(1, toWholeNumber(options.maxPerPage) ?? DEFAULT_MAX_PER_PAGE);
```
nstfkc: **`maxPerPage: Infinity` is silently ignored and falls back to the default 100**

`maxPerPage: Infinity` — the natural spelling of "no ceiling" — is non-finite, so `toWholeNumber` returns `null` and the ceiling silently falls back to 100 instead of being lifted.

**Failure scenario:** `paginate({ perPage: "1000" }, { maxPerPage: Infinity })` returns `{ take: 100, skip: 0 }` — verified by running it. `toWholeNumber` rejects non-finite input by design (the comment on that check is about hostile query strings), but the same function is reused for the trusted options object, so an internal/admin endpoint written as `paginate(args, { maxPerPage: Infinity })` or `{ maxPerPage: Number.MAX_VALUE }` to opt out of the DoS ceiling gets the default 100 back with no error.

The endpoint appears to work and returns a plausible first page, so the truncation is only found when someone notices a 1000-row export is missing 900 rows.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2205,3 +2468,87 @@ function listOf(value: unknown): unknown[] {
...
+ *     child exactly as it was, foreign key included.
+ *
+ * So `true` becomes `{}`, which `conjoin` reduces to the parent link alone —
+ * precisely the one connected row — and `false` becomes `null`, which `listOf`
+ * reduces to `[]`, which is no statement.
+ */
+function toOneOperand(key: string, operand: unknown): unknown {
+  if (key === "update") {
+    // A `data` key is the wrapper; anything else is the payload itself. The
+    // same test `planOwningSide`'s `update` makes, deliberately — the two
```
nstfkc, line 2655: **The comment claims parity with `planOwningSide`, but the two disagree on `data: undefined`**

The comment claims `toOneOperand` makes "the same test `planOwningSide`'s `update` makes", but the owning side still branches on `"data" in operandAt` while the new normaliser branches on `.data !== undefined`, so the two sides disagree on `update: { data: undefined }`.

**Failure scenario:** `User.update({ where: { id: 1 }, data: { organization: { update: { data: maybeUndefined } } } })` where the value is `undefined` on one code path — the ordinary way conditional writes are expressed, and the reason `suppliedFields` skips `undefined` in the first place. The owning-side step (`nested-writes.ts:527`) sees `"data" in operandAt` as true, passes `data: undefined` to `Organization.updateMany`, and the caller gets `UnsupportedQueryError: updateMany requires 'data'`.

The same call through `User.profile` normalises to `{ where: undefined, data: {} }` and is a no-op returning the parent, which is also what Prisma does (it treats an explicit `undefined` as absent). The PR fixed this asymmetry on the foreign side and left the side it cites as precedent unchanged.

docs/navigation.md
```diff
@@ -193,7 +193,7 @@ import { useSearchParams } from "gemi/client";
 
 function Pagination() {
   const searchParams = useSearchParams();
-  const page = Number(searchParams.get("page")) || 1;
+  const page = Math.max(1, Math.trunc(Number(searchParams.get("page"))) || 1);
```
nstfkc: **The recommended client-side page clamp returns `Infinity` for the input the docs say it fixes**

The recommended client-side page clamp returns `Infinity` for `?page=1e400`, the exact input the paragraph beside it says `Math.trunc` cannot repair.

**Failure scenario:** `Math.max(1, Math.trunc(Number(searchParams.get("page"))) || 1)` with `?page=1e400`: `Number` is `Infinity`, `Math.trunc(Infinity)` is `Infinity`, `Infinity || 1` is `Infinity`, `Math.max(1, Infinity)` is `Infinity`. So the snippet the docs present as the whole of the client-side job (and `UPGRADE.md` repeats verbatim under "On the client") hands the component `page === Infinity`; the very next lines in the same example do `searchParams.set("page", String(next))`, which writes `?page=Infinity` into the URL, and any `page + 1` / `page - 1` control renders `Infinity`/`NaN`.

`UPGRADE.md`'s own prose two sections earlier says `1e400` "is `Infinity`, which `Math.trunc` cannot fix either" — the recommended fix does not cover the case it calls out. A `Number.isFinite` test is what the server-side `toWholeNumber` uses for precisely this.

packages/gemi/bin/migrate/index.ts
```diff
@@ -195,13 +196,17 @@ export async function runMigrate(options: MigrateOptions) {
   // 4. Provider files whose contents now live in app/config are superseded.
   for (const file of migratedFiles) changes.push({ file, contents: null });
 
-  // 5. Renames across the rest of the app, plus retired fields in app/config.
+  // 5. Renames across the rest of the app, plus retired fields in app/config
+  //    and the two ORM-port hazards. The ORM pass runs over every file, not
+  //    only `app/config`: what it looks for is a guard or a query argument, and
```
nstfkc: **The ORM-port passes sit inside the `touched` skip, so they never see provider-derived files**

The new ORM-port passes are placed inside the loop that skips files earlier passes already rewrote, so they never see those files.

**Failure scenario:** `annotateOrmPortHazards` is documented as needing to "sweep the tree unprompted" (`orm.ts:47`), but it runs after `if (touched.has(file)) continue;`, and `touched` holds every `app/config/*.ts` the config pass emitted, every extracted file, `Kernel.ts` and `AppServiceProvider.ts`.

On the one run that matters — the migration run on an old-layout app — a `"P2002"` guard or a non-integer `take` inside any of those files is never annotated and never reported in `notes`, and the user has no signal that those files were skipped. A second run does scan them, but only because the earlier passes have become no-ops by then.

nstfkc (commented):
## Code review — high effort, workflow-backed

4 finder angles produced 30 candidates; every distinct location was verified by an independent adversarial verifier. 21 survived, ranked and capped to the 10 reported below (9 refuted, the rest folded in as same-root-cause).

Two clusters dominate:

1. **The migrate command's new ORM-port annotation pass** splices `//` TODO lines into `.tsx` files by line number, so `bunx gemi migrate` can leave a user's app failing to parse or rendering the TODO text as page content — and it fires on ordinary English prose and destructuring renames.
2. **Nested to-one write validation**: `toOneOperand` silently drops the caller's `where` (turning filtered nested updates into unconditional writes on both owning and foreign sides) and makes several plan-time refusals unreachable, pushing failures into mid-transaction errors that name models the caller never wrote.

The rest cover a silently-ignored `maxPerPage: Infinity`, an owning/foreign asymmetry on `data: undefined`, a docs clamp returning `Infinity`, a differential harness that now masks ordering divergences, and the annotation pass never seeing provider-derived files on the run that matters.

nstfkc:
All ten fixed in 683b0e1. Every one reproduced first — none was taken on trust, and none was refuted.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | `//` spliced into `.tsx` breaks the parse | Extension gate; those files are **reported with line numbers** instead |
| 10 | Passes sit inside the `touched` skip | They now sweep every file the run leaves behind, including ones it creates |
| 3 | `toOneOperand` drops unknown keys | `assertToOneWriteOperand`, run **before** the rewrite |
| 4 | To-one `where` never validated | Same check — a non-object `where` is refused, not dropped |
| 5 | Non-object operand escapes to run time | Same check — refused at plan time, with the caller's own argument path |
| 2 | Owning side ignores its `where` | Conjoined and looked up, raising `RecordNotFoundError` as Prisma's P2025 |
| 8 | `"data" in` vs `.data !== undefined` | Owning side reads the value; both sides agree |
| 7 | `maxPerPage: Infinity` ignored | Honoured; the *request* still cannot spell it, and that is asserted |
| 6 | Sort masks ordered includes | Selection threaded through; a node naming `orderBy` stays positional |
| 9 | Docs clamp returns `Infinity` | `Number.isFinite` first, in `navigation.md` and `UPGRADE.md` |

Three notes where the fix is not quite the one suggested.

**On #1** — I did not try to detect JSX. The gate is the file extension, which is a *guarantee* rather than a heuristic: TypeScript permits JSX only in `.tsx` / `.jsx`. And because neither finder can hit inside a string or template literal — `skipAtomic` steps over both before either looks — every hit in a file that cannot hold JSX is ordinary code, so the splice is safe there and stays. The `.tsx` path reports line numbers, which is what the splice was carrying by sitting on the line. The new test asserts the file still **parses** rather than that it is unchanged, since parsing is the property that broke.

**On #10** — fixed, but the P2002 half still does not reach the generated `app/config/*.ts`, and the reason is the import gate rather than the sweep: the generated config imports `gemi/services`, not the model surface, because the provider's model import is not carried across. That is the gate's documented false negative arriving by a new route. I asserted it rather than leaving it, because the neighbouring `take` hazard in the same file *is* found now and the asymmetry would otherwise read as a bug in the sweep.

**On #6** — you were right that the flake only ever affected unordered includes, so the sort should have been skipped where the case named an `orderBy`. That is what it does now. Worth recording that the masking was total, exactly as you said: both sides get re-sorted by the same comparator, so `[{id:10},{id:2}]` and `[{id:2},{id:10}]` compare equal.

**On #2, one thing worth a second look.** Honouring the owning-side `where` costs a `findFirst` before the `updateMany`, and only when a `where` is actually present — that is shape-stable, so `update: { … }` still compiles to one statement. I took that over narrowing the type back, because the operand is legal Prisma and the foreign side already implements it; refusing it would have re-opened the asymmetry from the other end. Say if you'd rather have the narrower type.

Verification after the fixes: packages/gemi 3,032 · saas-starter 745 · Postgres 1,122 · `tsc` clean · `oxlint` clean.

nstfkc:
Both points were right; checked each before acting.

**On `Closes #354`** — filed [#359](https://github.com/nstfkc/gemi/issues/359) (owning-side `disconnect: false`) and [#360](https://github.com/nstfkc/gemi/issues/360) (`create` on an occupied to-one), each carrying the measurement and pointing at the test that pins both clients' answers; the divergence describe now carries the issue numbers back. `#354` keeps its `Closes`: its stated acceptance is met, and both of these are separate defects this PR's own review surfaced rather than part of what it asked for. The record now survives the merge either way.

Worth noting on #360 — Prisma *orphans* the row it displaces rather than deleting it. That is the half a guessed fix would get wrong, so it is in the issue's acceptance as a read-the-table assertion.

**On `@updatedAt`** — you were right that it is the hunk to read hardest, and checking it turned up something sharper than "read it carefully": it was documented only as a *rule*, in `docs/orm.md` reference prose and in source comments. Nothing said it is a **behaviour change**, and it is the only entry in the 0.51 notes that changes behaviour for apps already on gemi's ORM rather than only for apps porting off Prisma.

The spelling to grep for is a write with an empty payload used as a touch:

```ts
await User.update({ where: { id }, data: {} })   // no longer moves updatedAt
```

That never worked under Prisma, so no ported app can depend on it — but an app written against the ORM directly could. `UPGRADE.md` now carries the measurement table, that shape, what to write instead, and the one divergence that stays.

f5c91d2. packages/gemi 3,032 · saas-starter 745 · docs guards green.

PR #362, opened 2026-08-08
Close #359, #360 and #361: the owning side's disconnect grammar, and writing onto an occupied to-one

### LINKED ISSUE #359: An owning-side to-one `disconnect` takes only `true`, where Prisma also takes `false` and a filter
#354 implemented the boolean spelling of `disconnect` on the side whose **child** holds the foreign key. The owning side — where the key is on the row being updated — still takes `true` and nothing else, so the two sides of the same relation answer the same call differently.

Prisma's operand here is `ProfileWhereInput | boolean`, read off the generated client (6.19.2, `ProfileUpdateOneWithoutUserNestedInput`). It accepts `true`, `false` and a filter. gemi implements `true`.

## What it costs a caller

The spelling that breaks is the one a conditional detach is naturally written as:

```ts
await Profile.update({
  where: { id },
  data: { user: { disconnect: shouldDetach } },
})
```

On the branch where `shouldDetach` is `false`, Prisma writes nothing and returns the row with its link intact. gemi refuses the call. So the branch that asked for *nothing* is the one that fails, and it fails at run time — `shouldDetach` is a `boolean`, so nothing upstream can see it coming.

This is strictly better than what it replaced. Before #358, `disconnect: true` and `disconnect: false` shared one plan-cache entry, so a warm `true` plan served `false` and **nulled the foreign key on a call that asked for nothing** — the compile-time refusal never re-ran on a cache hit. That is fixed; the refusal now fires reliably. It is still a refusal where Prisma answers.

## Measured

```
prisma.profile.update({ where: { id: 1 }, data: { user: { disconnect: false } } })
  -> returns the row, userId still 1, nothing written

Profile.update(<the same args>)
  -> UnsupportedQueryError: only 'disconnect: true' is implemented
```

Pinned in `templates/saas-starter/app/models/writes.differential.test.ts`, in the `describe("where the two clients still disagree — each of these is a bug")` block. It asserts **both** clients' answers, so it fails the day either side moves — including the day this is fixed, which is what closing this issue looks like.

## The filter arm is missing too

`disconnect: { name: "Acme" }` — detach only if the linked row matches — is on Prisma's operand and unimplemented here. It has no caller demand behind it that I know of, unlike the boolean, but it is the same gap and the same one-line grammar.

## Acceptance

- The owning-side `disconnect` takes `false` as a no-op and a `WhereInput` as a conditional detach, matching the foreign side #358 implemented.
- `false` writes nothing — asserted by value, not only by "did not throw", since the failure this replaces was a write that should not have happened.
- The divergence case moves out of the "still disagree" describe into `CASES`, which is what closing it looks like: the harness takes over from the prose.
- The refusal message on the remaining unimplemented arm, if any, keeps saying what *is* implemented rather than asserting what Prisma offers — it claimed Prisma took only `true` until #358 measured otherwise.

### LINKED ISSUE #360: A nested `create` on an occupied to-one collides, where Prisma detaches the incumbent
A nested `create` on a to-one whose foreign key lives on the child fails with a unique violation when the parent already has one. Prisma answers the same call: it **detaches** the incumbent first, leaving it orphaned rather than deleted, and links the new row.

```prisma
model User    { profile Profile? }
model Profile { userId Int? @unique
                user   User?  @relation(fields: [userId], references: [id]) }
```

```ts
await User.update({
  where: { id: 1 },
  data: { profile: { create: { bio: "second" } } },
})
```

## Measured

Prisma 6.19.2, SQLite, with user 1 already carrying a profile:

```
prisma            -> ok.  Profile rows afterwards:
                       ("seed",   null)    the incumbent, orphaned not deleted
                       ("loose",  null)
                       ("second", 1)       the new row takes the link

gemi              -> UniqueConstraintError on Profile.userId
                     nothing written; the failed insert rolls back
```

That orphaning is the surprising half and is why this is worth implementing deliberately rather than guessing at: Prisma does **not** delete the row it displaces. It nulls its foreign key and leaves it in the table. A fix that deleted instead would be a data-loss bug wearing the same test.

## Why it survived #354

#354 was about `update` / `delete` / `upsert` — the three operands refused **by name** on that side. `create` was never refused there, so nothing pointed at it: it compiled, and against an *empty* to-one it works correctly. Only the occupied case diverges, and no fixture had a to-one whose foreign key was on the child until #358 added one.

The neighbouring operand already does the work. `set` exists to clear the links that are in the way before writing the new ones, and it does exactly the right thing here — read the linked rows through the child's own `findMany`, clear the ones the caller can see, then write. The scoping consequence carries over with it: `set` means *replace the set I can see*, so a child hidden by a policy stays attached rather than being silently detached, and this should mean the same.

## Pinned

`templates/saas-starter/app/models/writes.differential.test.ts`, in the `describe("where the two clients still disagree — each of these is a bug")` block. It asserts **both** clients' answers — including that gemi rolls back cleanly today — so it fails the day either side moves.

## Acceptance

- A nested `create` on an occupied to-one detaches the incumbent and links the new row, matching Prisma.
- The displaced row is **orphaned, not deleted** — asserted by reading the table, since the difference between the two is silent data loss.
- The whole thing stays one unit, as every other nested write here is: a failure anywhere rolls back the detach as well as the insert.
- A child the caller's policies hide is not detached, for the same reason `set` does not detach one.
- The case moves out of the "still disagree" describe into `CASES`.

### LINKED ISSUE #361: A nested `connect` on an occupied to-one collides, where Prisma detaches the incumbent
A nested `connect` on a to-one whose foreign key lives on the child fails with a unique violation when the parent already has one. Prisma answers the same call: it **detaches** the incumbent and links the named row — the same displacement `create` does, which #360 implemented.

```prisma
model User    { profile Profile? }
model Profile { userId Int? @unique
                user   User?  @relation(fields: [userId], references: [id]) }
```

```ts
// user 1 already holds profile 1; profile 2 is linked to nobody
await User.update({
  where: { id: 1 },
  data: { profile: { connect: { id: 2 } } },
})
```

## Measured

Prisma 6.19.2, SQLite:

```
prisma  -> ok.  Profile rows afterwards:
              ("seed",  null)   the incumbent, orphaned not deleted
              ("loose", 1)      the named row takes the link

gemi    -> UniqueConstraintError on Profile.userId
           nothing written; the failed update rolls back
```

## Why it survived #360

#360 was `create`, and the fix is `create`'s: it clears the links that are in the way through `clearLinks` before inserting. `connect` runs the neighbouring branch of the same function and does not clear, so the child's `@unique` foreign key rejects the repoint.

The two neighbouring shapes already agree and are in `CASES`, which is why nothing pointed at this one:

- `connect` onto an **empty** to-one — attaches, both clients.
- `connect` that **steals another parent's child** — repoints it, both clients, leaving the first parent with none.

Only the *occupied* case diverges: the displacement is needed on the receiving side rather than on the giving one.

Note the operand set here is not uniform, and that is Prisma's rather than something to generalise — measured on the same fixture:

```
create           on an occupied to-one  ->  detaches the incumbent, links the new row
connect          on an occupied to-one  ->  detaches the incumbent, links the named row
connectOrCreate  (create branch)        ->  P2002 on `userId`
upsert           (create branch)        ->  P2002 on `userId`
```

So a fix belongs on `connect` by name, not on "anything that links a to-one".

## Pinned

`templates/saas-starter/app/models/writes.differential.test.ts`, in the `describe("where the two clients still disagree")` block. It asserts **both** clients' answers — including that gemi rolls back cleanly today — so it fails the day either side moves, which is what closing this looks like.

## Acceptance

- A nested `connect` on an occupied to-one detaches the incumbent and links the named row, matching Prisma.
- The displaced row is **orphaned, not deleted** — asserted by reading the table, since the difference between the two is silent data loss.
- A child the caller's policies hide is not detached, for the same reason `set` and `create` do not detach one — `clearLinks` already carries that rule and its scoping consequence.
- The whole thing stays one unit: a failure anywhere rolls back the detach as well as the repoint.
- `connectOrCreate` and `upsert` keep colliding, and the cases pinning that (`M11c`, `M11d`) stay green — they are the control that says the displacement belongs to the operand.
- The case moves out of the "still disagree" describe into `CASES`.

nstfkc:
Three to-one defects from the first Prisma differential port, each verified against a generated client (6.19.2) before anything was written, each graduated out of the "still disagree" describe and into the differential's case tables — which is what closing one looks like here.

#361 was filed by this branch and fixed on it after the fact, on request; its measurements and acceptance criteria are on the issue.

## #359 — the owning side took `true` and nothing else

Prisma's operand is `XWhereInput | boolean` on **both** sides of a to-one. #358 implemented all three arms where the child holds the key; the side that holds it itself still refused everything but `true`, so one grammar had two answers depending on which table the column is on. The spelling that broke is the one a conditional detach is naturally written as — `disconnect: shouldDetach` — and it broke on the branch that asked for *nothing*, at run time.

All three arms now go through `toOneOperand` and `assertToOneFilter`, the same two helpers the foreign side uses, reused rather than restated:

| operand | what happens |
| --- | --- |
| `true` | one bound column set to null, no step |
| `false` | no contribution and no step at all — the statement degenerates to a read |
| `{}` / `{ name: "Ada" }` | detach only if the linked row matches; **silent** when it does not |

The filter arm costs two reads, and neither is avoidable from this side: the condition is a property of the linked row, and which row that is lives in a column of the row being updated, which the arguments do not carry. The parent read is pre-scoped; the child read goes through its own `$exec` un-pre-scoped, so a linked row the caller's policies hide reads as one that does not match and the link survives. `true` has no lookup to scope and clears the column either way — pinned, because the pair reads as an inconsistency otherwise.

The boolean stays a **compile-time** decision here where the foreign side re-reads it at bind time, and that is forced: `false` emits no assignment and `true` emits one, so they cannot share a statement text. #358's plan-key guard is what makes it sound, and `plan.writes.discrimination` moved its pin off the (now gone) refusal and onto the emitted SQL.

## #360 and #361 — writing onto an occupied to-one collided

`create` and `connect` onto a to-one that already has a child were a `UniqueConstraintError` on calls Prisma answers. Prisma takes the link and leaves the incumbent in the table with a null foreign key — **orphaned, not deleted**, which is the half a fix in the other direction turns into silent data loss wearing the same green test. The differential reads the table back for exactly that reason.

`set`'s clearing half became a shared `clearLinks`, and which operands call it is one `displaces` at the top of `planForeignSide`, because the answer turned out to be per *branch*:

| operand | on an occupied to-one |
| --- | --- |
| `create` | displaces |
| `connect` | displaces |
| `connectOrCreate`, hit | displaces — it *is* a connect |
| `connectOrCreate`, miss | collides on the child's unique key |
| `upsert`, create branch | collides on the child's unique key |

Not derivable, so it is measured and consulted per site. The two collide rows are pinned as `M11c` / `M11d`, which is what stops the rule quietly widening to "anything that ends with a child pointing here". #361's own acceptance criteria say `connectOrCreate` keeps colliding — that was written from the miss branch, the only one measured when it was filed, and is still true of it; the hit branch is measured to displace, and leaving it out would have put two spellings of one operation on two answers.

Four `connect` shapes, and only one diverged — the other three are what kept it hidden: an empty to-one attaches, another parent's child is repointed, and the row already linked here comes out unchanged (the clear nulls it and the repoint puts the key straight back). A miss detaches nothing, because the repoint raises and the clear is inside the transaction the nested steps already run in.

`clearLinks` names the column as ORM-authored (#98), which `set`'s clear was not. The *link* half deliberately still does not: naming it too makes `set` succeed and succeed wrongly on a child scoped by its own foreign key — the clear puts the row outside the very scope the link selects it by, so it ends up detached and never re-attached. A silent half-write in place of a loud refusal, so the refusal stays until #99. Asserted rather than left as a comment.

## Two divergences found on the way, neither folded in

**Prisma ignores the owning-side `disconnect` operand on a many-to-one**, measured on three relations: `disconnect: false` and a filter matching nothing both null the foreign key, though the generated input type is the same `WhereInput | boolean` it honours on a one-to-one — and `delete: false` on the same relation is a correct no-op, which rules out "booleans are ignored here". gemi does not reproduce it. Matching Prisma is this suite's whole point and it is overruled exactly here, because the behaviour to match is a silent destructive write on the call that asked for nothing — word for word the defect #358 removed from this ORM's own plan cache. Pinned with both answers, and stated in the docs, since it means `disconnect: shouldDetach` is safe here and is not safe through the Prisma client.

**The owning side linking into an occupied to-one** — `Profile.update({ data: { user: { connect } } })` where that user's profile is taken — collides here and displaces in Prisma. It is a different piece of work rather than the same one twice: the row to clear is a *sibling of the model being written*, not a row of the child model, and nothing in `planOwningSide` writes one. Filed as #363 with the two questions that answers, pinned here with both clients' answers.

## Verification

`packages/gemi` 3,054 · `saas-starter` 770 · the differential suite against a real Postgres 211 · `tsc` clean on `packages/gemi` (the template's 277 errors are pre-existing and unchanged) · `oxlint` clean on every file touched.

Closes #359, closes #360, closes #361.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3884,7 +3884,7 @@ await List.create({
 | --- | --- |
 | `create` | Writes new related rows. One statement each. On a **to-one that already has a child** the new row takes the link and the old one is detached — orphaned, not deleted. |
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
-| `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
+| `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. On a **to-one that already has a child** it displaces the incumbent, as `create` does. |
 | `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
 | `disconnect` | Clears the link. A unique key, or a list of them, on a to-many; `true`, `false` or a filter on a to-one, whichever side holds the key. The column that goes null is on whichever side holds it, and it has to be nullable. |
 | `delete` | Deletes the named rows outright, not just the link. A unique key or a list of them on a to-many; `true`, or a filter, on a to-one whose child holds the key. |
@@ -3969,16 +3969,29 @@ With no policy on the child that is Prisma's `set` exactly. Two of its behaviour
 because the name does not suggest them — it will repoint a row belonging to another parent, exactly
 as `connect` does, and it silently ignores a named row that does not exist.
 
-**A nested `create` on a to-one that already has a child displaces it.** The new row takes the link
-and the incumbent is left in the table with a null foreign key — detached, not deleted, which is
-Prisma's answer and the one worth stating because deleting would be the silent version of the same
-call. What it can displace is what your policies let you see: an incumbent hidden from you is not
-detached, and the insert then collides with it on the child's unique key. `connectOrCreate` and
-`upsert` do *not* displace — they collide, on both clients — so the behaviour belongs to `create`
-rather than to the relation.
+**Writing onto a to-one that already has a child displaces it.** The new or newly-linked row takes
+the link and the incumbent is left in the table with a null foreign key — detached, not deleted,
+which is Prisma's answer and the one worth stating because deleting would be the silent version of
+the same call. What can be displaced is what your policies let you see: an incumbent hidden from you
+is not detached, and the write then collides with it on the child's unique key.
 
-`connect` onto an occupied to-one is the gap in that sentence: Prisma displaces there too, and gemi
-raises `UniqueConstraintError` instead. Detach the incumbent first, or write the child's foreign key
+Which operands displace is Prisma's, measured rather than derived, and it is not "everything that
+ends with a child pointing here":
+
+| operand | on an occupied to-one |
+| --- | --- |
+| `create` | displaces the incumbent |
+| `connect` | displaces the incumbent |
+| `connectOrCreate`, hit | displaces — it *is* a connect |
+| `connectOrCreate`, miss | collides on the child's unique key |
+| `upsert`, create branch | collides on the child's unique key |
+
+So the two branches of one `connectOrCreate` answer differently: linking an existing row displaces,
+and the `create` *inside* a compound operand does not, though a bare `create` does.
+
+One gap remains, on the other side of the key: pointing **this** row at a partner who already has
+one — `Profile.update({ data: { user: { connect: { id } } } })` where that user's profile is taken —
+collides here and displaces in Prisma. Detach the incumbent first, or write the foreign key
 directly.
 
 **On a to-one, `disconnect` and `delete` take `true` or a filter — and `false` is a no-op rather
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3882,11 +3882,11 @@ await List.create({
 
 | Operand | What it does |
 | --- | --- |
-| `create` | Writes new related rows. One statement each. |
+| `create` | Writes new related rows. One statement each. On a **to-one that already has a child** the new row takes the link and the old one is detached — orphaned, not deleted. |
 | `createMany` | The same rows in **one** statement. To-many only, and the rows go inside `data`. |
 | `connect` | Points at an existing row — a bound column when it names the referenced key, a lookup otherwise. |
 | `connectOrCreate` | Looks the row up by a unique key and creates it only if it is not there. **A hit ignores `create` entirely** — it is connect-*or*-create, not upsert. |
-| `disconnect` | Clears the link. A unique key, or a list of them, on a to-many; `true` on a to-one, and a filter as well when the child is the side holding the key. The column that goes null is on whichever side holds it, and it has to be nullable. |
+| `disconnect` | Clears the link. A unique key, or a list of them, on a to-many; `true`, `false` or a filter on a to-one, whichever side holds the key. The column that goes null is on whichever side holds it, and it has to be nullable. |
 | `delete` | Deletes the named rows outright, not just the link. A unique key or a list of them on a to-many; `true`, or a filter, on a to-one whose child holds the key. |
 | `update` | Writes your columns to the named row. The child's own `onUpdate` and scope rules apply to the payload. |
 | `set` | Replaces the whole set — detaches what is linked now, attaches what you name. |
@@ -3969,6 +3969,18 @@ With no policy on the child that is Prisma's `set` exactly. Two of its behaviour
 because the name does not suggest them — it will repoint a row belonging to another parent, exactly
 as `connect` does, and it silently ignores a named row that does not exist.
 
+**A nested `create` on a to-one that already has a child displaces it.** The new row takes the link
+and the incumbent is left in the table with a null foreign key — detached, not deleted, which is
+Prisma's answer and the one worth stating because deleting would be the silent version of the same
+call. What it can displace is what your policies let you see: an incumbent hidden from you is not
+detached, and the insert then collides with it on the child's unique key. `connectOrCreate` and
+`upsert` do *not* displace — they collide, on both clients — so the behaviour belongs to `create`
+rather than to the relation.
+
+`connect` onto an occupied to-one is the gap in that sentence: Prisma displaces there too, and gemi
+raises `UniqueConstraintError` instead. Detach the incumbent first, or write the child's foreign key
+directly.
+
 **On a to-one, `disconnect` and `delete` take `true` or a filter — and `false` is a no-op rather
 than a synonym for `true`.** There is one row and nothing to name, so `true` is how you say *the
 one that is linked*. A filter narrows rather than picks: `delete: { status: "draft" }` acts only if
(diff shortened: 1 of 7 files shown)


nstfkc (commented):
Reviewed at `c49745d`. Three behaviour changes plus the suites that pin them; the shape of the work reads well. Four things below, none of them a merge blocker — the first is the one I'd want a pin or a refusal for.

**Verified independently:** `packages/gemi` 3054 passed / 1 skipped / 120 files. And the owning-side filter arm's parent `findFirst` is only sound if exactly one row can match — it is, because `create`/`createMany` are cut off by `EXISTING_ROW_ONLY` × `CREATE_ONLY_STATEMENTS`, `updateMany` refuses nested writes, and `compileUpsert` refuses them on both branches. That's worth a line in the docblock: the arm becomes silently wrong the day any of those three opens up.

---

## 1. `connect` breaks the case pinned as "changes nothing" — under a policy

`M12b` holds without policies. With a to-one child scoped on its own foreign key — the same `{ folderId: 2 }` shape the #98 test uses for `Note`, applied to `Cover` — against the existing fixture:

```
Folder.update({ where: { id: 2 }, data: { cover: { connect: { id: 30 } } } })
  raised:     RecordNotFoundError: No Cover found (Cover.update).
  link after: 2      (rolled back — nothing corrupted)
```

`clearLinks` nulls `folderId` on the very row the caller named, which drops it out of the scope the repoint then selects it by, so the `update` matches nothing and raises. Before this commit the clear did not run, the row kept `folderId = 2`, and the repoint matched — so a call that worked now fails.

This is the crossing the `clearLinks` docblock already reasons about for `set`. The difference is which mechanism catches it: `set`'s link half is deliberately not ORM-named, so the scope-escape guard refuses it first; `connect`'s link half *is* named (#98), so it reaches the where-clause instead. Loud rather than silent, and nothing is written either way — but the shape is reachable and unasserted, and the error names a model and an operation the caller never wrote.

Narrow, and arguably #99's to fix. It deserves what `set` got: a pin in `nested-write-policies.test.ts` recording that this is the answer, or the operand refused until #99.

## 2. `connectOrCreate`'s hit branch does not pass the ORM-authored list

It calls `clearLinks` and then repoints without `[childField]`, which `connect` passes ten lines down. So the same child-scoped-on-its-FK model answers `ScopeEscapeError` through one spelling and `RecordNotFoundError` through the other — two spellings of one operation on two answers, which is the argument the commit message itself makes for fixing the hit branch at all. Pre-existing (#98 skipped `connectOrCreate`); this change is what puts the two next to each other.

## 3. The `disconnect` filter arm stamps `@updatedAt` on the no-match branch

Compiled against the `user` fixture:

```
true  :: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
false :: select … from "User" where "id" = ?
filt  :: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
```

`false` degenerating to a read is exactly right. But the filter arm's statement text is fixed at compile time, so `disconnect: { no match }` — and `disconnect: <filter>` with nothing linked — writes the foreign key back to its own value **and stamps `@updatedAt`**, where Prisma emits no statement at all.

`updateAssignments`' KNOWN DIVERGENCE block covers only `disconnect: true`, where a column genuinely changes; this is a new case of it where nothing does, and `docs/orm.md` calls the same branch "silently nothing". `O5`/`O6` cannot catch it: `Profile` has no `@updatedAt`, and the harness compares timestamps as the `"date"` descriptor. Either extend that docblock and the docs paragraph, or add an owning-side case on a model that carries the attribute.

## 4. Stale docblock in `plan.ts`

The `disconnect`/`delete` boolean guard still says, present tense, that "`planOwningSide` refuses `operand !== true` at compile time and then pushes a constant `value: () => null`". Both halves are now false. The pin in `plan.writes.discrimination` moved off the dead refusal; the comment explaining why the guard exists still describes it, and its new prose is the right replacement text.

---

The rest reads as solid: `clearLinks` sharing one scoping rule across `set` / `create` / `connect`, the deliberate asymmetry of naming the clear as ORM-authored but not the link, orphan-not-delete asserted by reading the table back rather than by the call returning, the per-branch `displaces` for `connectOrCreate`, and declining to reproduce Prisma's many-to-one operand-dropping with the reason stated.

PR #365, opened 2026-08-08
Close #364: a delete that includes and omits handed the column back

### LINKED ISSUE #364: A `delete` carrying both `include` and `omit` returns the omitted column, where Prisma drops it
A `delete` that names both operands returns the column the caller omitted. Prisma refuses only `select` + `omit` — the pair that describes two different column lists — and takes `include` + `omit`, dropping the column and returning the relations beside it. So this is legal, Prisma-compatible code that gemi answers with more data than was asked for, and the column a caller omits is by definition the one it had a reason not to hand out.

```ts
await User.delete({
  where: { id },
  include: { session: true },
  omit: { password: true },
})
```

Found by an application migrating off Prisma: the call above is its `UserController.delete`, unchanged across the swap, and after it the `DELETE /api/user/delete` response carries the user's stored password hash. Its assertion is `expect(body).not.toHaveProperty("password")`.

## Measured

Prisma 6.19.2, SQLite, the template's own differential schema:

```
prisma.user.delete({ where: { publicId: "p1" },
                     include: { _count: { select: { accounts: true } } },
                     omit: { password: true } })
  -> 15 keys, no `password`, `_count.accounts` = 4

User.delete(<the same args>)
  -> 16 keys. `password` is back.
```

Pinned as `delete with an omit beside an include` in `templates/saas-starter/app/models/writes.differential.test.ts`, beside the `delete with _count` case it mirrors, and as two runtime cases in `templates/saas-starter/app/models/delete-include.test.ts`.

## Where it goes wrong

`Model.$exec`, in the read-first branch that `delete` takes as soon as there is anything to read — `packages/gemi/orm/Model.ts:965`:

```ts
if (op === "delete" && (plan.relations !== undefined || plan.counts)) {
  const { where } = effective;
  const projection =
    effective.select !== undefined
      ? { select: effective.select }
      : { include: effective.include };
```

That projection is rebuilt from two of the three operands. The `findFirst` it feeds accepts `omit` like any other read, and the row it returns is what the caller gets back — `return before`, a few lines down — so nothing downstream ever narrows it again.

A `delete` with **no** relation and no `_count` never reaches this branch. It compiles to one statement, and its `RETURNING` list comes from `resolveSelection` (`packages/gemi/orm/compile/select.ts:22`) by way of `returningClause`, which honours `omit` exactly as a read does. That is why the leak needs both operands to show up: `omit` alone is correct, `include` alone is correct, and the pair is not.

`delete`'s operand set already names all four — `packages/gemi/orm/compile/write.ts:88`, `delete: new Set(["where", "select", "include", "omit"])` — so the call compiles without complaint, and the guard against `select` + `omit` at `select.ts:55` fires before this branch is reached. Nothing raises anywhere.

## Scope

`delete` alone. The read-first path is the only place in `Model.$exec` that reconstructs a projection from `effective`, and it is guarded on `op === "delete"`; grepping `effective.select` / `effective.include` finds no other site. `update`'s read-only form — a `data` with no column to set — projects through `returningClause`'s `selected`, which is the same `resolveSelection` list, so it carries the `omit` already. `recountAfterSteps` re-reads with `select: { _count }` and writes back only `row._count`, so it cannot widen a row either.

## Acceptance

- A `delete` carrying `include` (or a bare `_count`) beside an `omit` returns the row without the omitted column, and with the relations the `include` asked for.
- Asserted against Prisma by the differential harness, not only by shape — this is a difference in *which keys come back*, which is precisely what that harness compares.
- The `select` + `omit` refusal still fires: the pre-read must not become a fifth way to reach a projection that names both.

nstfkc:
Closes #364.

`Model.$exec` sends a `delete` down a read-first path as soon as there is a relation or a `_count` to read: read the row inside a transaction, delete it, hand the caller what was read. That pre-read's projection was rebuilt from two of the three projection operands — `select` if there was one, `include` otherwise — and `omit` was simply not carried across. The pre-read is `return`ed verbatim, so nothing narrows it afterwards.

The result is a leak that needs both operands to appear:

| call | today | after |
| --- | --- | --- |
| `delete({ where, omit })` | correct | unchanged — one statement, `RETURNING` built by `resolveSelection` |
| `delete({ where, include })` | correct | unchanged |
| `delete({ where, include, omit })` | **every column, including the omitted one** | the row without it |
| `delete({ where, select, omit })` | refused | refused, in `resolveSelection`, before this branch |

Prisma refuses only `select` + `omit` — the pair that describes two different column lists — and answers `include` + `omit` by dropping the column and returning the relations beside it. gemi's own operand set for `delete` already names all four, and `select.ts`'s guard already fires on the one illegal pair, so the call compiles clean and nothing raises. Found by an application migrating off Prisma whose `UserController.delete` is `{ where, include: { session: true }, omit: { password: true } }`: the `DELETE /api/user/delete` response was carrying the stored password hash.

## The fix

`omit` is forwarded into the `include` arm of the projection, and only that arm. It cannot legally ride with the other one — `select` + `omit` throws in `resolveSelection`, and the plan is compiled before this branch runs, so a call naming both never reaches here. Spreading it into both would be dead code in one of them and a second, weaker copy of that rule in a file that does not own it.

`findFirst` takes `omit` like any other read, so the pre-read narrows exactly as the single-statement `delete` would have.

## Scope

`delete` alone, and that is checked rather than assumed. This is the only site in `$exec` that reconstructs a projection out of `effective`, and it is guarded on the operation. `update`'s read-only form — a `data` with nothing to set — projects through `returningClause`'s `selected`, which is the same `resolveSelection` list, so it carried the `omit` already; `recountAfterSteps` re-reads with `select: { _count }` and writes back only `row._count`, so it cannot widen a row either.

## Pinned twice

`writes.differential.test.ts` gets `delete with an omit beside an include`, sitting next to the `delete with _count` case it mirrors. That one matters more than the others: it compares the returned key set against a **generated Prisma client (6.19.2)** rather than against an expectation written here, which is the right authority for a claim about what Prisma does with these two operands. Before the fix it reported the difference by name:

```
delete with an omit beside an include: returned value:
expected { _count: { accounts: 4 }, …(15) } to deeply equal { _count: { accounts: 4 }, …(14) }
+   "password": null,
```

`delete-include.test.ts` gets the runtime pair — one reaching the branch through `plan.relations`, one through `plan.counts`, since either half of the condition gets there and the `_count` half has no relation plan behind it.

## Verification

```
packages/gemi          bun --bun vitest run
  Test Files  120 passed (120)
       Tests  3052 passed | 1 skipped (3053)

packages/gemi          bun --bun vitest run orm
  Test Files  64 passed (64)
       Tests  2368 passed (2368)

templates/saas-starter bun --bun vitest run app/models
  Test Files  1 failed | 19 passed | 3 skipped (23)
       Tests  1 failed | 744 passed | 133 skipped (878)
```

The Postgres halves skip without a `POSTGRES_URL`, as they do locally for everyone. The one failure is `app/models/generated.test.ts`, which needs `gemi-orm-generator` linked on `PATH` and does not have it in a fresh worktree; it fails identically on `main` with these changes stashed, so it is environmental rather than anything on this branch.

https://claude.ai/code/session_01ChxCkNb4bLwd5aYSgAPByU

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -963,27 +963,33 @@ export abstract class Model {
     // plan, so `relations` stays empty for an `include: { _count: … }` on its
     // own — hence the separate flag rather than a wider check.
     if (op === "delete" && (plan.relations !== undefined || plan.counts)) {
-      const { where } = effective;
-      // All three operands, not two. The row this reads is the row the caller
-      // gets back — `return before`, below — so anything left out here is
-      // silently un-narrowed, and `omit` left out means the column the caller
-      // asked to drop comes back. A `delete` with no relation to read never
-      // reaches this branch and projects through `resolveSelection`, which
-      // honours `omit`; forwarding it here is what makes the two paths answer
-      // the same question the same way (#364).
+      // **Everything but the `where`, rather than the operands named one by
+      // one.** The row this reads is the row the caller gets back — `return
+      // before`, below — so an operand left out here is silently un-narrowed,
+      // and nothing downstream will notice. That is #364: this rebuilt the
+      // projection as `select` *or* `include`, `omit` was not among the names,
+      // and a `delete` that asked to drop a column got it back as soon as the
+      // same call also carried an `include`. A `delete` with no relation to read
+      // never reaches this branch and projects through `resolveSelection`, which
+      // had honoured `omit` all along — one operation, two paths, two answers.
       //
-      // `omit` rides with the `include` arm only because it cannot legally ride
-      // with the other: `select` + `omit` is refused in `resolveSelection`, and
-      // the plan above has already been compiled, so a call naming both threw
-      // before reaching this line. Spreading it into both arms would be dead in
-      // one of them and would put a second, weaker copy of that rule here.
-      const projection =
-        effective.select !== undefined
-          ? { select: effective.select }
-          : {
-              include: effective.include,
-              ...(effective.omit !== undefined ? { omit: effective.omit } : {}),
-            };
+      // **Do not turn this back into named arms.** The rest is what makes the
+      // next projection operand — Prisma's `relationLoadStrategy`, a `distinct`
+      // on a write, whatever it turns out to be — arrive here for free instead
+      // of arriving as the same bug with a different column name. The `upsert`
+      // find-then-write branch a few hundred lines up spreads for the same
+      // reason and was never affected by #364.
+      //
+      // Bounded, not hopeful. `assertArgs` (`compile/write.ts`) rejects any
+      // operand outside `WRITE_ARGS.delete` — `where`, `select`, `include`,
+      // `omit` — and the plan above is already compiled, so the rest holds
+      // projection operands and nothing else. That set is a strict subset of
+      // `findFirst`'s, so every one of them is a legal read operand. `omit` and
+      // `select` cannot both be in it: `resolveSelection` refuses that pair, and
+      // a warm plan cannot smuggle it past, since both are in `LITERAL_KEYS`
+      // (`plan.ts`) and so key separately. An `undefined` operand is dropped from
+      // the plan key too, so carrying one costs no second cache entry.
+      const { where, ...projection } = effective;
 
       // The pool, not `conn`: when a transaction is already open `withTransaction`
       // savepoints against the ambient handle and ignores this argument, and
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -964,10 +964,26 @@ export abstract class Model {
     // own — hence the separate flag rather than a wider check.
     if (op === "delete" && (plan.relations !== undefined || plan.counts)) {
       const { where } = effective;
+      // All three operands, not two. The row this reads is the row the caller
+      // gets back — `return before`, below — so anything left out here is
+      // silently un-narrowed, and `omit` left out means the column the caller
+      // asked to drop comes back. A `delete` with no relation to read never
+      // reaches this branch and projects through `resolveSelection`, which
+      // honours `omit`; forwarding it here is what makes the two paths answer
+      // the same question the same way (#364).
+      //
+      // `omit` rides with the `include` arm only because it cannot legally ride
+      // with the other: `select` + `omit` is refused in `resolveSelection`, and
+      // the plan above has already been compiled, so a call naming both threw
+      // before reaching this line. Spreading it into both arms would be dead in
+      // one of them and would put a second, weaker copy of that rule here.
       const projection =
         effective.select !== undefined
           ? { select: effective.select }
-          : { include: effective.include };
+          : {
+              include: effective.include,
+              ...(effective.omit !== undefined ? { omit: effective.omit } : {}),
+            };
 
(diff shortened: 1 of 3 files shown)


nstfkc:
Both review points addressed in 2cb58d0.

**The differential now covers both halves of the branch.** It only exercised `plan.counts` — an `include: { _count }` beside an `omit`. The `plan.relations` half was pinned by a hand-written expectation and never seen by Prisma. Split and named for the route each takes:

```
delete with an omit beside a _count            include: { _count: { select: { accounts: true } } }
delete with an omit beside a relation include  include: { accounts: { orderBy: { id: "asc" } } }
```

Both pass. With the `Model.ts` hunk reverted, the four cases fail together — the two differential ones and the two runtime ones: `4 failed | 200 passed | 1 skipped (205)`.

**On explicit arms vs. the rest spread — you were right, and I have changed it.** `const { where, ...projection } = effective;`

My argument was that the explicit form gave the `select` + `omit` reasoning somewhere to live. That was the weaker half of the trade: the reasoning can live in a comment above a spread just as well, and it now does, while immunity to a fourth operand cannot be written into a comment at all. Named arms *are* how #364 happened — `omit` was not among the names, and nothing failed until someone combined it with an `include`.

What made it a clear call rather than a taste one is that the spread here is bounded rather than hopeful, which I checked before switching:

- `assertArgs` rejects any operand outside `WRITE_ARGS.delete` = `where | select | include | omit`, and the plan is compiled before this branch runs — so the rest holds projection operands and nothing else.
- That set is a **strict subset** of `findFirst`'s, so every key the rest can carry is a legal read operand. There is no shape where the spread turns a working `delete` into a throw.
- `applyPolicies` only ever spreads-and-replaces `where` / `data` / `create` / `update`; it never adds a top-level key, so post-policy `effective` stays within the same bound.
- `undefined` values are filtered out of the plan key (`plan.ts`), so carrying an explicitly-`undefined` operand costs no second cache entry — the property the conditional spread was protecting comes for free.

Thanks for the two corrections to my model of the code; both landed in the comment rather than only in my head.

**The `upsert` branch spreading from `args`.** You are right that my "only site" claim held for a reason I had not identified, and the comment now cites that branch by name as the precedent — it spreads, and it was never affected by #364. That is the argument for the spread made concretely rather than in the abstract.

**The plan-cache leg.** My "dead code in one arm" reasoning went from compile ordering only, and a cache hit skips compilation — so if the key ignored `omit`, a warm `{select}` plan would have served `{select, omit}` and the call *would* have reached this branch with both. It cannot, because `omit` is in `LITERAL_KEYS`. The comment now names that as the load-bearing fact rather than the ordering.

## Tallies

```
packages/gemi   bun --bun vitest run
  Test Files  120 passed (120)
       Tests  3052 passed | 1 skipped (3053)

packages/gemi   bun --bun vitest run orm
  Test Files  64 passed (64)
       Tests  2368 passed (2368)

templates/saas-starter   bun --bun vitest run app/models
  Test Files  1 failed | 19 passed | 3 skipped (23)
       Tests  1 failed | 745 passed | 133 skipped (879)
```

The one failure is `generated.test.ts`, environmental as established. `oxlint` clean on `orm/Model.ts`.

PR #368, opened 2026-08-08
Close #336: type the JSON path filter the compiler already implements

### LINKED ISSUE #336: `JsonFilter` does not type the path filter the compiler implements, so malformed filters compile
Found migrating Folio's analytics reads onto the named `analytics` connection from #329. This one blocks a whole controller rather than a line.

`JsonFilter` (`orm/types.ts`) offers `equals` and `not` on the column as a whole:

```ts
type JsonFilter = {
  equals?: JsonValue | DbNullValue | JsonNullValue | AnyNullValue;
  not?: JsonValue | DbNullValue | JsonNullValue | AnyNullValue;
};
```

There is no way to filter on a key *inside* the document. Prisma's spelling is `path`:

```ts
where.OR = [
  { payload: { path: ["operation"], equals: search.operation } },
  { name: search.operation, payload: { path: ["operation"], equals: AnyNull } },
];
```

`AsyncJob.payload` is a `Json` column holding `{ operation, … }`, and this is the admin job list's operation filter. The second arm is the interesting half — "the payload has no `operation` key, so fall back to matching the job's `name`" — which is what `AnyNull` at a path means and what keeps older rows written before the key existed reachable from the same filter.

Without it the whole `where` has to stay on Prisma, so the controller cannot move to the analytics pool at all.

## Note on `AnyNull`

`AnyNull` is already exported from `gemi/orm` and `types.ts:236` says it "is accepted here and nowhere else". At a path it means something a whole-column filter cannot express: *the key* is absent-or-null, as against *the column* being absent-or-null. Worth deciding deliberately rather than inheriting.

## Suggested shape

```ts
type JsonFilter = {
  path?: string[];
  equals?: …;
  not?: …;
  // and, if the operators come along: string_contains, array_contains, gt/gte/lt/lte
};
```

Postgres has `#>` / `#>>` for the path and `jsonb_typeof` for the null distinction, so the compiler has what it needs; `path` on SQLite would want `json_extract`, which may be a reason to gate it by dialect the way `mode: "insensitive"` already is.

I am not asking for the full operator set — `path` + `equals`/`not` is what this call site needs, and it is the shape most uses take.

Found on `gemi@0.51.0-rc.6`.

nstfkc:
A type-only change. `compileJsonFilter` already implements the whole feature — both path grammars, the `#>`/`#>>` vs `json_extract` split, the bound path, the bare-path refusal reproducing Prisma's *"A JSON path cannot be set without a scalar filter."*, the sentinel refusals — and `read.test.ts`'s 22 `json path filters` tests cover it. Nothing in the runtime is touched, and nothing in it moved.

What was broken is the compile-time surface, and it is the fourth of the same kind as #326, #333 and #337. It is also the worst of the four: those three *refused* an argument the compiler accepts, so a caller got an error and knew. This one accepted anything, silently.

| written | before | after | compiler |
| --- | --- | --- | --- |
| `{ path: ["operation"], equals: "video" }` | compiles | compiles | works |
| `{ path: 123, notAFilter: true }` | **compiles** | error | raises |
| `{ path: [], string_contains: 4 }` | **compiles** | error | raises |
| `{ path: ["operation"], equalz: "video" }` | **compiles** | error | raises |
| `{ metadata: { a: 1 } }` | **compiles** | error | raises |
| `{ metadata: { not: { a: 1 } } }` | **compiles** | error | raises |
| `{ path: ["a"], equals: DbNull }` | **compiles** | error | raises |

Measured with `tsc --noEmit --pretty false` on both sides, not read off the type.

## The absorption is the defect; the operators are the easy half

`JsonFilter` was `{ equals?, not? }` — no `path`, none of the ten operators `JSON_FILTERS` accepts. But adding them changes nothing on its own, and that is the part worth being explicit about.

`FieldFilter`'s `Json` arm unioned in `JsonValue`, which contains `{ [key: string]: JsonValue }`. `{ path: 123, notAFilter: true }` is a perfectly good JSON *document*, so it landed in the value arm; a union permits a property present in any member, so excess-property checking never fired either. A `JsonPathFilter` added beside that arm is just one more member the nonsense does not have to match.

**The object member does not need excluding, narrowing or discriminating. It needs deleting.** `where` has no bare-document branch for a `Json` column at all: `isFilterObject` reports any non-`Date`, non-array object as a filter, so `{ metadata: { a: 1 } }` reaches the scalar operator loop and raises *"A scalar filter takes contains, endsWith, equals, …"* on the key `a`. Only scalars and arrays fall through to `equals`. Measured against `compileRead` on Postgres:

| operand | compiler |
| --- | --- |
| `{ metadata: { a: 1 } }` | `InvalidArgumentError` on `where.metadata.a` |
| `{ metadata: [1, 2] }`, `{ metadata: "s" }` | `"metadata" = $1::text::jsonb` |
| `{ metadata: { equals: { a: 1 } } }` | `"metadata" = $1::text::jsonb` |
| `{ metadata: {} }` | `true` |

So the value arm is `Exclude<JsonValue, JsonObject>`, with no index signature left anywhere in the union — which is what lets excess-property checking work at all, and what makes `path` discriminate for free. A document goes under `equals`, which is the only spelling `docs/orm.md` has ever shown. `{}` still compiles, matching `where true`.

`JsonFilter` keeps a `path?: never`, and that one is load-bearing: without it `{ path: […], equals: DbNull }` satisfied `JsonFilter` structurally with `path` counting as "known" because `JsonPathFilter` declares it, so the sentinel-at-a-path case went on compiling. Found by a test, not by reading.

An earlier revision of this branch kept the object member with `path?: never` intersected onto it — the trick worked, and its entire effect was to go on admitting operands that always throw. Recorded here because the review that caught it is the reason the change is smaller than it was.

## `equals` binds a value; `not` recurses

`compileNot` hands any object operand back through `compileFieldFilter`, so `{ not: { path: […], equals: 1 } }` is a real negated path filter and `{ not: { a: 1 } }` raises, while `{ equals: { a: 1 } }` binds a document. `not` therefore takes the filter union — `JsonValueOperand | sentinels | JsonFilter | JsonPathFilter` — and `equals` keeps `JsonValue`. The asymmetry is the compiler's; it is now written down.

## Flat across the dialects, and why

The operator set and the path grammar are both dialect-dependent — Postgres takes `["a","b"]` and all ten operators, SQLite takes `"$.a.b"` and has neither `array_contains` nor the four comparisons — and `compileJsonFilter` refuses the wrong dialect's grammar at run time, naming the dialect.

Typed as one union of both anyway. The generated artifact is **dialect-agnostic**: `ModelTypeInfo` records no dialect, and the dialect is chosen from `DATABASE_URL` at connect time, so there is nothing for a dialect-shaped type to read. Shaping it means threading a parameter the generator does not know through every model, every `WhereInput` and every call site — and it is not how this file already handles the identical divergence. `mode: "insensitive"` is Postgres-only and is a flat property with a one-line comment; `ListFilter` is offered on both dialects though SQLite has no array type to answer it, with `assertListDialect` refusing at run time. The refusal stays where it can name the dialect and say "it works on postgres", which a type cannot do.

`path` is `string | readonly (string | number)[]`. `readonly` because hoisting a path — `const PATH = ["operation"] as const`, or the whole filter object written `as const` — is the natural thing to do and produced a compile error against a query that runs; `assertPathShape` uses `Array.isArray` and `.every`, both of which a frozen array answers.

Numbers because `assertPathShape` accepts them, where Prisma's generated `path` is `string[]`. That is a superset of the oracle, and this file refuses to be one on dialect operators for a stated reason — *"implementing them would make gemi answer a query the oracle cannot"* (`where.ts:945`). The two are consistent because that rule governs what the **compiler** answers and a type cannot enforce it: excluding numbers here would not remove the superset, since `assertPathShape` would still take `["items", 0]` from a `string[]`-typed variable or from `any`. It would only stop the type describing the compiler, which is #336 again in the other direction. If the superset should go, the place is `assertPathShape` — filed as #371 with both answers written out, rather than decided quietly in a type.

## What is deliberately still refused

The sentinels are absent from `JsonPathFilter`, and their absence is what keeps folio's `{ payload: { path: […], equals: AnyNull } }` — #336's original example — from typechecking. `#>>` yields SQL NULL for an absent key and for a JSON `null` alike, so the distinction the sentinels exist to draw is gone before the comparison happens. They stay on `JsonFilter`, which compiles against the column, where it survives. The same shape one level down — `{ path: […], equals: { not: DbNull } }` — falls out of the same scalar operand type, and both are pinned.

`null` is refused too, and **this is the one place the type goes further than the compiler.** `{ path: […], equals: null }` compiles to `("metadata" #>> $1) = $2` bound to NULL, which is NULL rather than true on both dialects — a predicate no row can satisfy. Once the sentinels are refused there is no other reading of it, and the same operand under `string_contains` is already a runtime refusal, in this file's own words because the pattern "runs and returns the wrong rows". Marked as a divergence rather than smuggled in, and filed against the runtime as #371.

Every operator stays optional, so `{ path: ["a"] }` alone still typechecks and still raises the bare-path error. Requiring one means a ten-way union in exactly the position where the compiler reports a misspelled operator, and the runtime message already names the whole set. Prisma's generated input has the same shape. The empty path stays a run-time refusal, for the same reason a non-empty tuple type would break `path: somePathVariable`.

## Verification

`packages/gemi/orm` **2,305 pass · 1 fail**, the 22 `json path filters` tests among them; the failure is `defaults.test.ts`'s `vi.resetModules` under `bun test`, pre-existing and identical on a clean tree at the same commit. `saas-starter` type tests **116 pass**, up from 102 — fourteen new, covering both path grammars, all ten operators, each nonsense form, the sentinel and its one-level-down shape, `null` at a path, the bare document, `not`'s recursion, the `as const` forms, a `path` key *inside* a document staying an ordinary key, and the same narrowing reached through a to-one. `saas-starter` runtime **296 pass**, byte-identical to baseline. `tsc --noEmit` clean on `packages/gemi`; `oxlint` and `oxfmt` clean on both files touched. `docs/llms-full.txt` re-embedded, which `docs-scope.test.ts` checks verbatim.

Closes #336. Files #371 for the three runtime operand gaps this surfaced.

https://claude.ai/code/session_01ChxCkNb4bLwd5aYSgAPByU

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3738,15 +3738,25 @@ document, which is a filter on the column rather than on a path.
 **What the type checks, and what it leaves to run time.** The path filter is typed: `path` takes
 either dialect's grammar, the ten operators are the ones above, and their operand types are the
 ones the compiler will compile. A misspelled operator, a `path` that is not one, a non-string under
-`string_contains` and a sentinel at a path are all compile errors. The type carries no dialect —
-the generated artifact does not know which database it will be pointed at — so the table above is
-offered in full on both and the wrong half is refused at run time, naming the dialect. The empty
-path is refused at run time too.
+`string_contains`, a sentinel at a path and a `null` at a path are all compile errors. The type
+carries no dialect — the generated artifact does not know which database it will be pointed at —
+so the table above is offered in full on both and the wrong half is refused at run time, naming the
+dialect. The empty path is refused at run time too.
 
-One consequence worth naming: a JSON document with a **top-level `path` key** cannot be written as
-the bare-value shorthand, because `path` is what tells the compiler the operand is a path filter.
-It never could — `{ metadata: { path: "/a" } }` reached the path compiler and raised — and now says
-so at compile time. Write the document explicitly: `{ metadata: { equals: { path: "/a" } } }`.
+**An object in a `Json` column's `where` is a filter, never a document.** `{ metadata: { a: 1 } }`
+does not ask for the document `{ a: 1 }` — it asks for the operator `a`, which does not exist, and
+raises `InvalidArgumentError`. Only a scalar or an array is read as a value. That has always been
+true and is now a compile error rather than a run-time one, `path` keys included:
+
+```ts
+await User.findMany({ where: { metadata: { equals: { a: 1 } } } })          // the document
+await User.findMany({ where: { metadata: { equals: { path: "/a" } } } })    // and this one
+await User.findMany({ where: { metadata: [1, 2] } })                        // arrays are values
+```
+
+`not` is the exception, and it is a filter too: it takes a whole nested filter, so
+`{ metadata: { not: { path: ["a"], equals: 1 } } }` is a negated path filter, while
+`{ metadata: { equals: { … } } }` binds its operand as a value.
 
 ### Scalar lists (Postgres only)
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3735,6 +3735,19 @@ merely unsupported:
 An empty path — `[]` or `""` — is refused on both dialects. On Postgres it would extract the whole
 document, which is a filter on the column rather than on a path.
 
+**What the type checks, and what it leaves to run time.** The path filter is typed: `path` takes
+either dialect's grammar, the ten operators are the ones above, and their operand types are the
+ones the compiler will compile. A misspelled operator, a `path` that is not one, a non-string under
+`string_contains` and a sentinel at a path are all compile errors. The type carries no dialect —
+the generated artifact does not know which database it will be pointed at — so the table above is
+offered in full on both and the wrong half is refused at run time, naming the dialect. The empty
+path is refused at run time too.
+
+One consequence worth naming: a JSON document with a **top-level `path` key** cannot be written as
+the bare-value shorthand, because `path` is what tells the compiler the operand is a path filter.
+It never could — `{ metadata: { path: "/a" } }` reached the path compiler and raised — and now says
+so at compile time. Write the document explicitly: `{ metadata: { equals: { path: "/a" } } }`.
+
 ### Scalar lists (Postgres only)
 
 ```ts
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -755,15 +755,25 @@ document, which is a filter on the column rather than on a path.
 **What the type checks, and what it leaves to run time.** The path filter is typed: `path` takes
 either dialect's grammar, the ten operators are the ones above, and their operand types are the
 ones the compiler will compile. A misspelled operator, a `path` that is not one, a non-string under
-`string_contains` and a sentinel at a path are all compile errors. The type carries no dialect —
-the generated artifact does not know which database it will be pointed at — so the table above is
-offered in full on both and the wrong half is refused at run time, naming the dialect. The empty
-path is refused at run time too.
+`string_contains`, a sentinel at a path and a `null` at a path are all compile errors. The type
+carries no dialect — the generated artifact does not know which database it will be pointed at —
+so the table above is offered in full on both and the wrong half is refused at run time, naming the
+dialect. The empty path is refused at run time too.
 
-One consequence worth naming: a JSON document with a **top-level `path` key** cannot be written as
-the bare-value shorthand, because `path` is what tells the compiler the operand is a path filter.
-It never could — `{ metadata: { path: "/a" } }` reached the path compiler and raised — and now says
-so at compile time. Write the document explicitly: `{ metadata: { equals: { path: "/a" } } }`.
+**An object in a `Json` column's `where` is a filter, never a document.** `{ metadata: { a: 1 } }`
+does not ask for the document `{ a: 1 }` — it asks for the operator `a`, which does not exist, and
+raises `InvalidArgumentError`. Only a scalar or an array is read as a value. That has always been
+true and is now a compile error rather than a run-time one, `path` keys included:
+
+```ts
+await User.findMany({ where: { metadata: { equals: { a: 1 } } } })          // the document
+await User.findMany({ where: { metadata: { equals: { path: "/a" } } } })    // and this one
+await User.findMany({ where: { metadata: [1, 2] } })                        // arrays are values
+```
+
+`not` is the exception, and it is a filter too: it takes a whole nested filter, so
+`{ metadata: { not: { path: ["a"], equals: 1 } } }` is a negated path filter, while
+`{ metadata: { equals: { … } } }` binds its operand as a value.
 
 ### Scalar lists (Postgres only)
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -752,6 +752,19 @@ merely unsupported:
 An empty path — `[]` or `""` — is refused on both dialects. On Postgres it would extract the whole
 document, which is a filter on the column rather than on a path.
 
+**What the type checks, and what it leaves to run time.** The path filter is typed: `path` takes
+either dialect's grammar, the ten operators are the ones above, and their operand types are the
+ones the compiler will compile. A misspelled operator, a `path` that is not one, a non-string under
+`string_contains` and a sentinel at a path are all compile errors. The type carries no dialect —
+the generated artifact does not know which database it will be pointed at — so the table above is
+offered in full on both and the wrong half is refused at run time, naming the dialect. The empty
+path is refused at run time too.
+
+One consequence worth naming: a JSON document with a **top-level `path` key** cannot be written as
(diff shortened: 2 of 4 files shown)


nstfkc:
Both blocking items addressed in `88889e8`, plus two of the three non-blocking ones. The review was right about the important thing, and it cost the change its cleverest construct — correctly.

## 1. The object arm is gone

You are right that there is no bare-document branch. I re-measured against `compileRead` before touching anything, Postgres, `read.test.ts`'s `jsonUser` fixture:

| operand | compiler |
| --- | --- |
| `{ metadata: { a: 1 } }` | `InvalidArgumentError` on `where.metadata.a` |
| `{ metadata: { equalz: "video" } }` | `InvalidArgumentError` on `where.metadata.equalz` |
| `{ metadata: [1, 2] }`, `{ metadata: "s" }` | `"metadata" = $1::text::jsonb` |
| `{ metadata: { equals: { a: 1 } } }` | `"metadata" = $1::text::jsonb` |
| `{ metadata: {} }` | `true` |

So `JsonValueOperand` is now just `Exclude<JsonValue, JsonObject>`, the intersection is gone with the paragraph explaining it, and the two literals I had pinned as valid are `@ts-expect-error` naming the error the compiler raises. You are also right about what that test *was* — this PR's own defect class aimed the other way, with my name on it.

`JsonFilter`'s `path?: never` stays, and thank you for confirming it is doing real work; it was found by a test rather than by reading, which is the only reason it is there.

**One correction.** It does not also refuse `{ metadata: {} }`. `JsonFilter`'s properties are all optional, so `{}` still satisfies it — verified with `tsc --pretty false`, no diagnostic. That is the right answer anyway, since the compiler emits `where true` for it, but the empty object is not free.

## 2. `readonly` path

Fixed as you suggested — `string | readonly (string | number)[]`. Confirmed the regression and the pre-existing refusal both against baseline `types.ts` with `tsc --pretty false`: on `main` the `as const` filter object compiled and `const PATH = ["operation"] as const` was the sole diagnostic; on the branch as pushed, both failed; now neither does. A mutable array still assigns to a `readonly` parameter, so nothing that compiled stops.

## Non-blocking, taken

**`not` recurses, so it takes the filter union.** Measured: `{ not: { path: ["a"], equals: 1 } }` → `not ("metadata" #>> $1) = $2`; `{ not: { equals: { a: 1 } } }` → `not ("metadata" = $1::text::jsonb)`; `{ not: { a: 1 } }` raises on `a`. So `not?: JsonValueOperand | sentinels | JsonFilter | JsonPathFilter` and `equals` keeps `JsonValue`. Modelling it as `| JsonPathFilter` alone would have refused `{ not: { equals: … } }`, which works — the recursion is what makes it come out right. The asymmetry between the two operators is the compiler's, and it is now written down where someone will find it.

**`null` dropped from `JsonPathScalar`.** Taken, and flagged in the type as the one place this change goes *further* than the compiler. The argument that convinced me is the one you make: with the sentinels refused there is no reading of `equals: null` left, since the two questions it could ask are exactly the two they ask — and `string_contains: null` is already a runtime refusal in this file's own words, "the pattern runs and returns the wrong rows". `= NULL` is that with different arithmetic. A `null` arriving dynamically still reaches the binder, so the type is not a guarantee and does not claim to be.

**`undefined` at a path** — agreed, not fixed here.

## The superset argument

You are right that I stated the conclusion and not the reasoning, and that as written it sat next to a comment refusing to be a superset for the opposite reason. Reworked rather than reversed, and the distinction is this: the rule at `where.ts:945` governs what the **compiler answers**, and a type cannot enforce it. Excluding numbers from `JsonPath` would not remove the superset — `assertPathShape` would still accept `["items", 0]` from a `string[]`-typed variable, from `any`, from JSON off the wire — it would only stop the type describing what the compiler does, which is #336 again in the other direction. A type-only change cannot decide this; `assertPathShape` can.

So it stays typed, and the decision is filed rather than made quietly.

**#371** covers all three runtime items — the `undefined` operand, the `null` operand, and the numeric path segment — each with the measurement and, for the segment, both possible answers written out. Both comments in `types.ts` that say "the compiler should follow" now point at it.

## Verification

`packages/gemi/orm` **2,305 pass · 1 fail** (`vi.resetModules`, pre-existing, identical on a clean tree). Type tests **116 pass**, up from 111 — five new: the bare-document refusal, `null` at a path, `not` carrying a whole filter, the `as const` forms, and a `path` key *inside* a document staying an ordinary key, since that is the property the whole discriminant rests on and it deserved a pin rather than a review comment. `saas-starter` runtime **296 pass · 1 fail**, byte-identical to baseline at this commit. `tsc` clean on `packages/gemi`; `oxlint` and `oxfmt` clean on both files touched. `docs/orm.md` and its `llms-full.txt` copy updated — the old paragraph claimed the bare-document shorthand worked for everything except `path`, which was the same wrong belief.

PR #377, opened 2026-08-09
Close #374: the plan-key guard explains itself by its current reason

### LINKED ISSUE #374: The plan-key guard for a `disconnect` boolean still explains itself by a refusal #359 removed
`plan.ts`, the guard that gives a `disconnect` / `delete` boolean its own plan key, still explains itself in terms of a refusal #359 removed:

> On the owning side `planOwningSide` refuses `operand !== true` at compile time and then pushes a constant `value: () => null` — and a cache *hit* skips compilation, so the refusal never re-runs.

Both halves are now false. Nothing is refused; `true` pushes the constant, `false` pushes no contribution at all, and the two spellings therefore compile to **different statement text** — an `update … set "organizationId" = ?, "updatedAt" = ?` against a bare `select`.

That matters more than a stale sentence usually does, because the paragraph is the argument for why the guard has to exist. Read today it says the guard protects a refusal, which would make it removable now that the refusal is gone. The real reason is stronger and is already written down one directory over — `plan.writes.discrimination.test.ts` was updated in the same PR:

> the boolean is decided *once*, when the plan is built, because it decides the SET list — so unlike the foreign side, where `toOneOperand` re-reads it at bind time, there is no second guard here and the key is the whole of it.

## Acceptance

- The docblock describes the guard's current reason: the boolean is structural on the owning side, so a cache hit that got it wrong emits the wrong SET list and no bind-time check can rescue it.
- The historical note stays — "a plan compiled from `true` served `false` and nulled a foreign key" is why the guard was added and is worth keeping — but reads as history rather than as present behaviour.

nstfkc:
Comments only. The guard itself — `shapeOfMember`'s `disconnect` / `delete` boolean key — is unchanged, and so is every statement gemi emits.

## What was wrong

`plan.ts` argued for the guard from a refusal that no longer exists:

> That was a live wrong write rather than a wrong refusal. On the owning side `planOwningSide` **refuses** `operand !== true` at compile time and then pushes a constant `value: () => null` — and a cache *hit* skips compilation, so the refusal never re-runs.

Both halves are false on `main`. `724a83f` (#359) replaced that refusal with all three arms — `true` pushes the constant, `false` pushes **no contribution at all**, a filter gets a `before` step — and only `nested-writes.ts` and `plan.writes.discrimination.test.ts` were updated with it. `plan.ts` was not; `git log -S` puts its guard in `e07a3ca`, the #358 commit, and nothing has touched the docblock since.

That matters more than a stale sentence usually does, because the paragraph *is* the argument for the guard's existence. Read as written it says the guard protects a refusal — so with the refusal gone, the guard looks like dead weight the next reader can delete.

## The measured behaviour it should have described

`packages/gemi/orm`, sqlite dialect, the `user` fixture, printed from a scratch test rather than quoted from the source:

```
TRUE  text: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
FALSE text: select "id", "publicId", … from "User" where "id" = ?
TRUE  key : sqlite:batched:User:update:{data:{organization:{disconnect:true}},where:{id:number}}
FALSE key : sqlite:batched:User:update:{data:{organization:{disconnect:false}},where:{id:number}}
```

Two statements, not two bind lists — so the boolean is *structural* on this side, decided once when the plan is compiled.

**And the guard is what stands between those two.** With the `typeof value === "boolean"` branch short-circuited to `false` and nothing else changed, the same scratch test reports:

```
TRUE  text: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
FALSE text: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
TRUE  key : sqlite:batched:User:update:{data:{organization:{disconnect:boolean}},where:{id:number}}
FALSE key : sqlite:batched:User:update:{data:{organization:{disconnect:boolean}},where:{id:number}}

same plan object? true
params bound from false args: [ null, 1786262472010, 1 ]
```

One entry, and `disconnect: false` binds `organizationId = null`. The foreign key is cleared on a call that asked for nothing — the original #358 damage, reproduced today against the post-#359 code. So removing the refusal did not retire the guard; it made the guard load-bearing in a *second* way, because the mis-serve now also runs the wrong statement text and not only the wrong bound value.

Prisma's answer for the same call, which is what `disconnect: false` is implemented to match: the row comes back with its link intact. That is asserted by value in the template's differential suite, unchanged here.

## The change

`packages/gemi/orm/plan.ts` — the docblock above the guard, in three paragraphs instead of one:

1. **Why the guard exists now.** The boolean decides the SET list rather than a value bound into one, the two spellings are two statement texts, and a cache hit skips compilation — so a hit that got the boolean wrong runs the other spelling's statement and there is no later check to catch it.
2. **The contrast with the foreign side**, which is what makes the point land rather than sound like belt-and-braces: where the child holds the key the operand becomes a statement the *step* runs and `toOneOperand` re-reads the boolean at bind time (`listOf(null)` is no statement), so a mis-served plan there still writes nothing. `delete` shares the guard for that side's sake — on the owning side `delete` is refused outright (`nested-writes.ts`, `data.<rel>.delete`), so it has no boolean of its own to protect.
3. **The history, marked as history**: #358 added the guard while the refusal was still there, the refusal was never protection because a cache hit does not re-run the compiler, and a plan compiled from `true` served `false` and nulled the foreign key. The "measured against the real client" sentence is kept verbatim — it is the thing that makes the paragraph evidence rather than a story.

The `LITERAL_KEYS` paragraph below it is untouched; it was already accurate.

`packages/gemi/orm/plan.writes.discrimination.test.ts` — one clause, `refuses` → `refused … back then`, in the paragraph that is already narrating the original bug in the past tense and is explicitly corrected two paragraphs later by *"The texts have since diverged"*. Same stale refusal, same sentence, and the issue cites this file as the one that got it right — leaving the present tense in would leave the file contradicting itself.

## Checks

```
bun install                                            # in this worktree, since node_modules is not shared
bun --bun vitest run orm         # packages/gemi  → 64 files, 2421 tests, all passing
bun run typecheck                # packages/gemi  → clean
bunx oxlint orm --deny-warnings  # packages/gemi  → 0 warnings, 0 errors
```

`plan.writes.discrimination.test.ts` passes unchanged in substance (13 tests) — the assertions it makes are the pin for everything claimed above, and this PR does not move them.

Not run: the `templates/saas-starter` differential suites. Nothing here is executable, and the ORM package suite already covers both halves of the claim — the plan key and the two emitted texts. The Prisma-side `disconnect: false` behaviour quoted above is asserted there today and unaffected.

Closes #374.

### DIFF
--- a/packages/gemi/orm/plan.ts
+++ b/packages/gemi/orm/plan.ts
@@ -460,14 +460,32 @@ function shapeOfMember(
    * **On the owning side the boolean is structural — it decides the SET list
    * rather than a value bound into one**, and that is the whole of why the
    * guard has to exist. `disconnect: true` pushes a contribution for the
-   * foreign key, so the statement is `update "User" set "organizationId" = ?,
-   * "updatedAt" = ? …`; `disconnect: false` pushes no contribution at all, so
-   * no assignment is left and `compileUpdate` degenerates to a bare `select`
-   * of the row — which is the no-op Prisma answers. Two different statement
-   * *texts*, chosen once, when the plan is built. A cache **hit** skips
-   * compilation, so a hit that got the boolean wrong runs the other
-   * spelling's statement, and nothing downstream can rescue it: on this side
-   * the compiled plan *is* the decision.
+   * foreign key and `disconnect: false` pushes none, so the two spellings are
+   * two assignment lists and therefore two statement *texts*, chosen once,
+   * when the plan is built. Measured on `User.organization`, sqlite:
+   *
+   *     data: { name, organization: { disconnect: true } }
+   *       -> update "User" set "name" = ?, "organizationId" = ?, "updatedAt" = ? …
+   *     data: { name, organization: { disconnect: false } }
+   *       -> update "User" set "name" = ?, "updatedAt" = ? …
+   *
+   * When the `disconnect` is the **only** member of `data` the `false` arm
+   * leaves no assignment at all, and `compileUpdate` degenerates further, to a
+   * bare `select` of the row. That is the special case, not the rule — say it
+   * the other way round and a reader takes `{ name, disconnect: false }` to
+   * drop the `name` write too.
+   *
+   * A cache **hit** skips compilation, so a hit that got the boolean wrong
+   * runs the other spelling's statement, and nothing downstream can rescue it:
+   * on this side the compiled plan *is* the decision.
+   *
+   * Note the worked example is deliberately *not* offered as parity with
+   * Prisma. On a **many-to-one** — which `User.organization` is — Prisma
+   * discards the operand and nulls the key for `disconnect: false`, and gemi
+   * declines to reproduce that, since it is the same silent destructive write
+   * #358 removed from this cache. `writes.differential.test.ts` pins the
+   * divergence ("where the two clients still disagree"). The guard's argument
+   * needs no agreement from Prisma: two statement texts is the whole of it.
    *
    * **The foreign side is not like it, and the contrast is what makes the key
    * load-bearing here and merely prudent there.** Where the child holds the
@@ -485,11 +503,15 @@ function shapeOfMember(
    * skips compilation and so never re-runs it. A plan compiled from
    * `disconnect: true` served `disconnect: false` and nulled the foreign key on
    * a call that asked for nothing. Measured against the real client, so the
-   * damage was the gap and not the guess: `disconnect: false` with a child
-   * present leaves the row exactly as it was, foreign key intact. #359 then
-   * implemented all three arms and removed that refusal, which did not retire
-   * the guard — it made it load-bearing a second way, since the two spellings
-   * now differ in statement text and not only in what a contribution binds.
+   * damage was the gap and not the guess. Measured on the one-to-one the
+   * differential suite pins the grammar against, `Profile.user`:
+   * `disconnect: false` with a child present leaves the row exactly as it was,
+   * foreign key intact. That is the reference for the no-op — the many-to-one
+   * of the worked example above is where the two clients part company, and
+   * gemi answers one grammar one way on both shapes. #359 then implemented all
+   * three arms and removed that refusal, which did not retire the guard — it
+   * made it load-bearing a second way, since the two spellings now differ in
+   * statement text and not only in what a contribution binds.
    *
    * **It cannot simply join {@link LITERAL_KEYS}, which is the same bug class
    * documented there twice already** — for `omit` and for `skipDuplicates`.
--- a/packages/gemi/orm/plan.ts
+++ b/packages/gemi/orm/plan.ts
@@ -457,14 +457,39 @@ function shapeOfMember(
    * nothing above tells them apart: they sit under `data`, which is a
    * {@link VALUE_KEYS} subtree, so both shape to `boolean` and share one entry.
    *
-   * That was a live wrong write rather than a wrong refusal. On the owning side
-   * `planOwningSide` refuses `operand !== true` at compile time and then pushes
-   * a constant `value: () => null` — and a cache *hit* skips compilation, so
-   * the refusal never re-runs. A plan compiled from `disconnect: true` served
-   * `disconnect: false` and nulled the foreign key on a call that asked for
-   * nothing. Measured against the real client, so the damage is the gap and not
-   * the guess: `disconnect: false` with a child present leaves the row exactly
-   * as it was, foreign key intact.
+   * **On the owning side the boolean is structural — it decides the SET list
+   * rather than a value bound into one**, and that is the whole of why the
+   * guard has to exist. `disconnect: true` pushes a contribution for the
+   * foreign key, so the statement is `update "User" set "organizationId" = ?,
+   * "updatedAt" = ? …`; `disconnect: false` pushes no contribution at all, so
+   * no assignment is left and `compileUpdate` degenerates to a bare `select`
+   * of the row — which is the no-op Prisma answers. Two different statement
+   * *texts*, chosen once, when the plan is built. A cache **hit** skips
+   * compilation, so a hit that got the boolean wrong runs the other
+   * spelling's statement, and nothing downstream can rescue it: on this side
+   * the compiled plan *is* the decision.
+   *
+   * **The foreign side is not like it, and the contrast is what makes the key
+   * load-bearing here and merely prudent there.** Where the child holds the
+   * key the operand becomes a *statement the step runs*, and `toOneOperand`
+   * re-reads the boolean out of the call at bind time — `listOf(null)` is no
+   * statement — so a plan mis-served there still writes nothing. `delete`
+   * shares this guard for that side's sake; on this side it is refused
+   * outright, since deleting the far row through a `data` key would remove a
+   * row the statement is not about.
+   *
+   * **The history, which is why the guard was added rather than why it stays.**
+   * When #358 added it, `planOwningSide` still refused `operand !== true` at
+   * compile time and then pushed a constant `value: () => null` for the one arm
+   * it took — and the refusal was no protection at all, because a cache *hit*
+   * skips compilation and so never re-runs it. A plan compiled from
+   * `disconnect: true` served `disconnect: false` and nulled the foreign key on
+   * a call that asked for nothing. Measured against the real client, so the
+   * damage was the gap and not the guess: `disconnect: false` with a child
+   * present leaves the row exactly as it was, foreign key intact. #359 then
+   * implemented all three arms and removed that refusal, which did not retire
+   * the guard — it made it load-bearing a second way, since the two spellings
+   * now differ in statement text and not only in what a contribution binds.
    *
    * **It cannot simply join {@link LITERAL_KEYS}, which is the same bug class
    * documented there twice already** — for `omit` and for `skipDuplicates`.
--- a/packages/gemi/orm/plan.writes.discrimination.test.ts
+++ b/packages/gemi/orm/plan.writes.discrimination.test.ts
@@ -279,10 +279,11 @@ describe("plan cache discrimination — writes", () => {
    * `disconnect: true` and `disconnect: false` shaped alike — a bare `boolean`
    * under `data`, which is a value subtree — so they were one cache entry. On
    * the owning side that is not a wasted entry, it is a wrong statement:
-   * `planOwningSide` refuses `operand !== true` **at compile time** and then
-   * pushes a constant `value: () => null`, and a cache hit skips compilation
-   * entirely. So a plan compiled from `disconnect: true` served `disconnect:
-   * false` and nulled the foreign key on a call that asked for nothing.
+   * `planOwningSide` refused `operand !== true` **at compile time** back then
+   * and then pushed a constant `value: () => null`, and a cache hit skips
+   * compilation entirely. So a plan compiled from `disconnect: true` served
+   * `disconnect: false` and nulled the foreign key on a call that asked for
+   * nothing.
    *
    * **Nothing above could catch it**, and that is worth stating precisely
    * because it is what a reader will assume was covered. `plan-key.invariants`
--- a/packages/gemi/orm/plan.writes.discrimination.test.ts
+++ b/packages/gemi/orm/plan.writes.discrimination.test.ts
@@ -279,17 +279,21 @@ describe("plan cache discrimination — writes", () => {
    * `disconnect: true` and `disconnect: false` shaped alike — a bare `boolean`
    * under `data`, which is a value subtree — so they were one cache entry. On
    * the owning side that is not a wasted entry, it is a wrong statement:
-   * `planOwningSide` refused `operand !== true` **at compile time** back then
-   * and then pushed a constant `value: () => null`, and a cache hit skips
+   * back then, `planOwningSide` refused `operand !== true` **at compile time**
+   * and pushed a constant `value: () => null`, and a cache hit skips
    * compilation entirely. So a plan compiled from `disconnect: true` served
    * `disconnect: false` and nulled the foreign key on a call that asked for
    * nothing.
    *
    * **Nothing above could catch it**, and that is worth stating precisely
    * because it is what a reader will assume was covered. `plan-key.invariants`
-   * asserts *same key ⇒ same SQL text*, and at the time these two had
+   * asserts *same key ⇒ same SQL text*, and neither side gave it anything to
+   * catch. On the **foreign** side the two spellings genuinely had
    * byte-identical text — the difference was in which value a `contribution`
-   * binds and in whether a step runs at all.
+   * binds and in whether a step runs at all. On the **owning** side, at the
+   * time, `disconnect: false` threw at compile time and so had no text to
+   * compare with; it acquired one, and a differing one, only when #359 gave it
+   * an arm.
    *
    * **The texts have since diverged, and the pin below moved onto them**: #359
    * implemented `disconnect: false` as the no-op Prisma answers, so it now


packages/gemi/orm/plan.ts
```diff
@@ -457,14 +457,39 @@ function shapeOfMember(
...
-   * a constant `value: () => null` — and a cache *hit* skips compilation, so
-   * the refusal never re-runs. A plan compiled from `disconnect: true` served
-   * `disconnect: false` and nulled the foreign key on a call that asked for
-   * nothing. Measured against the real client, so the damage is the gap and not
-   * the guess: `disconnect: false` with a child present leaves the row exactly
-   * as it was, foreign key intact.
+   * **On the owning side the boolean is structural — it decides the SET list
+   * rather than a value bound into one**, and that is the whole of why the
+   * guard has to exist. `disconnect: true` pushes a contribution for the
+   * foreign key, so the statement is `update "User" set "organizationId" = ?,
```
nstfkc: **should fix** — the worked example is the one relation where the parity claim on line 466 is measurably false.

`User.organization` is a **many-to-one** — `organization.users` is `kind: "many"` in `orm/fixtures.ts` — and that is exactly the shape `templates/saas-starter/app/models/writes.differential.test.ts` pins as *the only deliberate divergence in the file*, naming `User.organization` first:

```
disconnect: false               ->  the foreign key is nulled
disconnect: { name: "Nobody" }  ->  the foreign key is nulled
```

with the test asserting `fromPrisma.organizationId === null` against `fromGemi.organizationId === 1`. So on this relation the bare `select` is *not* "the no-op Prisma answers" — gemi overrules Prisma here on purpose, because reproducing it would be re-introducing the silent destructive write #358 removed.

The phrase itself is house wording and is fine in the abstract (`plan.writes.discrimination.test.ts:295` uses it too, unattached to a relation). What makes it wrong here is that it sits two clauses after `"User"` / `"organizationId"`, so a reader binds it to the counterexample.

Two ways out, either is a one-line edit: switch the worked example to `Profile.user` (the one-to-one the `OWNING_CASES` table pins, where `false` genuinely is Prisma's answer), or keep `User.organization` and drop the parity clause — the guard's argument is "two statement texts", and it does not need Prisma's agreement to stand.

Note this also bears on the sentence kept verbatim at line 488 (*"`disconnect: false` with a child present leaves the row exactly as it was, foreign key intact"*): that measurement is `Profile.user`'s, and under a `User`/`organizationId` example it reads as a claim about the many-to-one.

packages/gemi/orm/plan.ts
```diff
@@ -457,14 +457,39 @@ function shapeOfMember(
...
+   * "updatedAt" = ? …`; `disconnect: false` pushes no contribution at all, so
+   * no assignment is left and `compileUpdate` degenerates to a bare `select`
```
nstfkc: **should fix** — "no assignment is left" is only true when the `disconnect` is the *only* thing in `data`.

Measured on this branch, sqlite, `user` fixture:

```
data: { organization: { disconnect: false } }
  -> select "id", "publicId", … from "User" where "id" = ?

data: { name: "x", organization: { disconnect: false } }
  -> update "User" set "name" = ?, "updatedAt" = ? where "id" = ? returning …
```

As written, a reader takes the second case to compile to a `select` — i.e. to silently drop the `name` write. Both siblings carry the qualifier this one drops: `nested-writes.ts` says *"a `data` carrying **only this** compiles to the read `compileUpdate` emits for an empty assignment list"*, and the differential's `O8 owning disconnect false beside a column write` exists specifically because *"**without something else in `data`** the statement degenerates to a read"*.

Worth noting the argument gets *stronger*, not weaker, from the fix: in the mixed case the two spellings are still two texts (`set "name" = ?, "organizationId" = ?, "updatedAt" = ?` against `set "name" = ?, "updatedAt" = ?`), so "the boolean decides the SET list" is the general statement and the bare `select` is the special case of it.

packages/gemi/orm/plan.writes.discrimination.test.ts
```diff
@@ -279,10 +279,11 @@ describe("plan cache discrimination — writes", () => {
    * `disconnect: true` and `disconnect: false` shaped alike — a bare `boolean`
    * under `data`, which is a value subtree — so they were one cache entry. On
    * the owning side that is not a wasted entry, it is a wrong statement:
-   * `planOwningSide` refuses `operand !== true` **at compile time** and then
-   * pushes a constant `value: () => null`, and a cache hit skips compilation
-   * entirely. So a plan compiled from `disconnect: true` served `disconnect:
-   * false` and nulled the foreign key on a call that asked for nothing.
+   * `planOwningSide` refused `operand !== true` **at compile time** back then
```
nstfkc: **nit** — the tense fix is right and I would keep the hunk, but `refused … **at compile time** back then and then pushed` puts the time marker after the bolded phrase and then repeats "then" in the next clause. `back then, `planOwningSide` refused `operand !== true` **at compile time** and pushed a constant …` reads the way the rest of the paragraph does. Purely prose; no objection to landing as is.

nstfkc (commented):
**Verdict: minor comments.** The diagnosis is right, the guard really is load-bearing, and the central claim is *measured* rather than remembered — I reproduced it. Two clauses in the new paragraph are inaccurate as written, both in the first paragraph, both a sentence-level fix. Nothing here is behavioural.

## What I verified

**The guard is load-bearing, reproduced against this branch.** Short-circuited it (`if (false && (key === "disconnect" ...`), nothing else changed:

```
TRUE  text: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
FALSE text: update "User" set "organizationId" = ?, "updatedAt" = ? where "id" = ? returning …
same plan object? true
params bound from false args: [ null, 1786266263565, 1 ]
```

and three tests in `plan.writes.discrimination.test.ts` go red (`true, false and absent are three keys`, `a warm 'true' plan does not serve 'false' on the owning side`, `...and the two booleans are two keys here as well`). With the guard restored, the two spellings are two keys and two texts, exactly as the PR body prints them. So the paragraph's thesis — removing the refusal did not retire the guard — holds, and the existing pins are genuinely mutation-sensitive.

**The history is accurate.** `git show e07a3ca:packages/gemi/orm/compile/nested-writes.ts` still carries `if (operand !== true) throw new UnsupportedQueryError(...)` at line 454, and `724a83f` (#359) is what replaced it with the three arms. `git log -S` does put the guard in `e07a3ca` = PR #358, and there is no *issue* #358 to confuse it with, so the `#358` / `#359` mix of a PR number and an issue number reads unambiguously and matches how `nested-writes.ts` already cites both.

**The foreign-side contrast is true in both directions**, which the paragraph asserts only in one. `planForeignSide` pushes its `after` step unconditionally for `disconnect`/`delete` (`nested-writes.ts` ~1600 onward) and rebinds `at = (args) => toOneOperand(key, spelled(args))`, so a `true` plan served `false` runs the step over `listOf(null) === []` and writes nothing, and a `false` plan served `true` still writes. Nothing hides behind the "merely prudent".

**`delete` on the owning side is refused outright**, and the reason the new prose gives ("would remove a row the statement is not about") is verbatim the reason in the thrown `UnsupportedQueryError`. So the second paragraph's answer to "why is `delete` in the guard at all" checks out.

**Nothing else in the tree carries the dead refusal.** `grep` for `operand !== true` across `.ts` and `.md` returns nothing outside history; `docs/orm.md` §to-one already describes all three arms correctly. This PR is not leaving a third stale site behind.

**Commands run** (fresh worktree, `bun install` in-tree so nothing resolved to the main checkout):

```
bun --bun vitest run orm         # packages/gemi → 64 files, 2421 tests, all pass
bun run typecheck                # clean
bunx oxlint orm --deny-warnings  # 0 warnings, 0 errors
gh pr checks 377                 # sqlite pass, postgres pass, on head 9af2751
```

**Not checked:** I did not re-measure Prisma 6.19.2 myself, and I did not run the template differential suites — nothing here is executable and CI ran both dialect jobs green on the head commit. Every Prisma claim below is checked against this repo's own recorded measurements, not against a live client.

## Findings

Two inline, both **should fix**, both in the first new paragraph; one **nit** on the test file. Detail is on the lines.

1. **should fix** — `plan.ts:463`: the worked example is `User.organization`, which is a *many-to-one* (`Organization.users` is a list in the fixture), and that is precisely the relation the differential suite pins as the one place gemi deliberately diverges: Prisma **nulls** the key for `disconnect: false` there. So the parity clause two lines down is false for the example it is attached to.
2. **should fix** — `plan.ts:465`: "no assignment is left" is only true when the `disconnect` is the only thing in `data`; measured counterexample on the line.
3. **nit** — `plan.writes.discrimination.test.ts:282`: prose only.

## Not findings, for the record

- **The test-file hunk is in scope**, and I would keep it. It is the same dead refusal in the same present tense, in a paragraph that already narrates in the past and is corrected two paragraphs later by *"The texts have since diverged"*. Leaving it would have left that file contradicting itself while the PR fixed its sibling.
- **Quoting a fixture-specific statement in `plan.ts` is fine** in principle — the file already quotes concrete examples — and it is the measured text. It only becomes a problem because of which relation was picked; see finding 1.
- **One inaccuracy in the same test-file docblock was left untouched and I agree with leaving it, but it is now the only false clause in a paragraph this PR just corrected**, so flagging it rather than losing it (line 290, outside the diff so it cannot be an inline comment): *"`plan-key.invariants` asserts same key ⇒ same SQL text, and at the time these two had byte-identical text"*. On the owning side, at the time, `disconnect: false` **threw** at compile time and so had no text at all — the byte-identical pair was the foreign side's. If a follow-up touches this file anyway, one clause fixes it; it is not worth a round trip on its own.

packages/gemi/orm/plan.ts
nstfkc: **Addressed** — you are right, and I checked both halves rather than taking it. `writes.differential.test.ts`'s `describe("where the two clients still disagree")` — the file's only deliberate divergence — names `User.organization` first and asserts `fromPrisma.organizationId === null` against `fromGemi.organizationId === 1`; `OWNING_CASES` pins the grammar against `Profile.user`, the one-to-one, where Prisma does honour `false`. So the parity clause was attached to the counterexample.

**I took your second exit, because the first is not available in this file.** `packages/gemi/orm/fixtures.ts`'s `profile` has `userId` `nullable: false`, so the operand never reaches a statement:

```
PROFILE THREW: UnsupportedQueryError | gemi ORM does not support 'data.user.disconnect' yet
(Profile.update). 'Profile.userId' is required, so there is no value to leave behind — a
disconnected row would have to be deleted or repointed instead.
```

The template's `Profile.userId` *is* nullable, which is how `OWNING_CASES` uses it, but this docblock quotes text measured against this package's fixtures and I did not want a worked example that cannot be reproduced from the file it sits in.

So `User.organization` stays, the parity clause is gone, and a short paragraph now says outright that the example is **not** offered as parity — naming the many-to-one divergence, saying gemi declines to reproduce it because it is the same silent destructive write #358 removed, and pointing at your `describe` by name. As you say, the guard's argument does not need Prisma's agreement: two statement texts is the whole of it.

**And the line-488 sentence too**, which you were right to flag in the same breath. It now names the relation it was measured on — *"Measured on the one-to-one the differential suite pins the grammar against, `Profile.user`: `disconnect: false` with a child present leaves the row exactly as it was, foreign key intact. That is the reference for the no-op — the many-to-one of the worked example above is where the two clients part company, and gemi answers one grammar one way on both shapes."*

If `profile.userId` in `orm/fixtures.ts` ever goes nullable, the example could move there and that paragraph could shrink.

packages/gemi/orm/plan.ts
nstfkc: **Addressed** — and re-measured here rather than taken on trust. Same numbers you got, sqlite, `user` fixture:

```
data: { organization: { disconnect: false } }
  -> select "id", "publicId", … from "User" where "id" = ?
data: { name, organization: { disconnect: true } }
  -> update "User" set "name" = ?, "organizationId" = ?, "updatedAt" = ? …
data: { name, organization: { disconnect: false } }
  -> update "User" set "name" = ?, "updatedAt" = ? …
```

Rewritten the way you point out makes the argument stronger, not weaker: the **mixed** case is now the worked example, so the general statement — the boolean decides the SET list — is what the reader sees first, and the two spellings are two texts there without needing the degenerate case at all. The bare `select` follows as the special case, explicitly conditioned on the `disconnect` being the *only* member of `data`, with a sentence saying why the order matters (*"say it the other way round and a reader takes `{ name, disconnect: false }` to drop the `name` write too"*).

That also puts `plan.ts` in line with the qualifier both siblings already carry — `nested-writes.ts`'s *"a `data` carrying only this"* and `O8 owning disconnect false beside a column write`.

packages/gemi/orm/plan.writes.discrimination.test.ts
nstfkc: **Addressed**, taken as suggested — the time marker leads and the doubled "then" is gone:

> back then, `planOwningSide` refused `operand !== true` **at compile time** and pushed a constant `value: () => null`, and a cache hit skips compilation entirely.

**And I fixed line 290 after all**, the one you flagged in "not findings" as not worth a round trip on its own — precisely because of your *"if a follow-up touches this file anyway, one clause fixes it"*, and this hunk lands eight lines above it. You are right about the mechanism: on the owning side, at the time, `disconnect: false` **threw** and so had no text to compare with; the byte-identical pair was the foreign side's. It now reads:

> `plan-key.invariants` asserts *same key ⇒ same SQL text*, and neither side gave it anything to catch. On the **foreign** side the two spellings genuinely had byte-identical text — the difference was in which value a `contribution` binds and in whether a step runs at all. On the **owning** side, at the time, `disconnect: false` threw at compile time and so had no text to compare with; it acquired one, and a differing one, only when #359 gave it an arm.

That was the last false clause left in a paragraph this PR exists to correct, so leaving it would have been the same defect the issue was filed for, one paragraph down.

I did **not** touch `plan.writes.discrimination.test.ts:295`'s *"the no-op Prisma answers"*, per your own ruling on it — house wording, unattached to a relation, and outside this diff.

nstfkc:
## Review round 1 — all three addressed, nothing declined, nothing deferred

Pushed as `bf2a5ce`. Still comments only: the guard, and every statement gemi emits, are unchanged.

Inline replies are on each thread; the short version:

**1. `plan.ts:463` — the worked example was the relation the parity clause is false for.** Right, and verified both halves: `writes.differential.test.ts`'s "where the two clients still disagree" names `User.organization` first and asserts `fromPrisma.organizationId === null` against `fromGemi.organizationId === 1`, while `OWNING_CASES` pins the grammar against the one-to-one `Profile.user`, where Prisma does honour `false`.

I took the **second** of the two exits offered, because the first is not available here — `orm/fixtures.ts`'s `profile.userId` is `nullable: false`, so `assertDisconnectable` refuses a `Profile.user` disconnect outright:

```
PROFILE THREW: UnsupportedQueryError | gemi ORM does not support 'data.user.disconnect' yet
(Profile.update). 'Profile.userId' is required, so there is no value to leave behind …
```

The template's `Profile.userId` is nullable, which is how `OWNING_CASES` uses it, but this docblock quotes text measured against *this package's* fixtures, and a worked example that cannot be reproduced from the file it sits in is a worse trade. So the measured example stays, the parity clause is gone, and a paragraph now says outright that the example is not offered as parity — naming the divergence, saying gemi declines to reproduce it because it is the same silent destructive write #358 removed, and pointing at the pin by name.

The line-488 sentence flagged alongside it is fixed the same way: it now names the relation it was measured on (`Profile.user`) and adds that the many-to-one above is where the two clients part company.

**2. `plan.ts:465` — "no assignment is left" was unqualified.** Right, and re-measured:

```
data: { organization: { disconnect: false } }
  -> select "id", … from "User" where "id" = ?
data: { name, organization: { disconnect: true } }
  -> update "User" set "name" = ?, "organizationId" = ?, "updatedAt" = ? …
data: { name, organization: { disconnect: false } }
  -> update "User" set "name" = ?, "updatedAt" = ? …
```

Rewritten in the direction the comment notes makes the argument stronger: the **mixed** case is now the worked example, so "the boolean decides the SET list" is the general statement the reader meets first and it stands without the degenerate case at all. The bare `select` follows as the special case, explicitly conditioned on the `disconnect` being the only member of `data` — the qualifier both siblings already carry (`nested-writes.ts`'s "a `data` carrying only this", and `O8 owning disconnect false beside a column write`).

**3. `plan.writes.discrimination.test.ts:282` (nit)** — taken as suggested.

**Plus the one filed under "not findings".** Line 290's *"at the time these two had byte-identical text"* is fixed, on its own stated condition that a follow-up touching the file makes it free — and hunk 3 lands eight lines above it. It is now split by side: byte-identical on the foreign side, no text at all on the owning side (it threw), a differing one only once #359 gave it an arm. That was the last false clause in a paragraph this PR exists to correct.

`plan.writes.discrimination.test.ts:295`'s "the no-op Prisma answers" is left alone, per the review's own ruling — house wording, unattached to a relation, outside the diff.

### Checks

```
bun run typecheck                # packages/gemi  → clean
bunx oxlint orm --deny-warnings  # packages/gemi  → 0 warnings, 0 errors
bun --bun vitest run orm         # packages/gemi  → 64 files, 2421 tests, all passing
```

**Mutation check re-run**, since the prose's whole claim is that these pins exist. Guard short-circuited to `if (false && …)` → 3 red (`true, false and absent are three keys`; `a warm 'true' plan does not serve 'false' on the owning side`, on `expect(off).toMatch(/^select /)`; `…and the two booleans are two keys here as well`). Restored → 13/13 green.

Template differential suites still not run, same reason as before: nothing here is executable. Both Prisma facts the new prose now cites *by relation* are already asserted by value in `writes.differential.test.ts` and untouched by this diff.

### Housekeeping

- `Closes #374` was already in the body and `closingIssuesReferences` was already populated — the clerical defect affecting #381 and #386 does not apply here. Re-verified.
- No cross-PR work lands here. #378 owns the `connectOrCreate` hit-branch short-circuit and #385 owns its numeric JSON path segments; #377 carries no cross-PR comment and is first in the merge order because it interacts with nothing.
- The `childField` → `childFields` rename is #386's and is merge-time work — #386 is not on `main`, so the new names cannot be referenced, and nothing in this PR names either spelling.

Issue #372, opened 2026-08-08
A nested `connect` of the row already linked fails when the child scopes on its own foreign key


nstfkc:
#361's `clearLinks` nulls the very row the caller named before repointing it. Without policies that is a no-op — the repoint puts the key straight back, which is what `M15b` pins. With a child scoped **on its own foreign key**, the clear puts the row outside the scope the repoint then selects it by, so the repoint matches nothing and raises.

```ts
// Cover.folderId is the unique FK; Cover's policy scopes on { folderId: 2 }
await Folder.update({
  where: { id: 2 },
  data: { cover: { connect: { id: 30 } } },   // cover 30 is already linked to folder 2
})
```

## Measured

Against `nested-write-policies.test.ts`'s own fixture, `main` @ `5412fcd`:

```
gemi -> RecordNotFoundError: No Cover found (Cover.update).
        Cover 30 still has folderId = 2   (the clear rolled back with the statement)
```

Before #361 the clear did not run, `folderId` stayed `2`, and the repoint matched — so this is a call that worked and now does not. Nothing is written either way, and the error names a model and an operation the caller never wrote.

## Why it is the shape it is

This is the crossing the `clearLinks` docblock already reasons about for `set`, reached by a different route. `set`'s link half is deliberately **not** named as ORM-authored, so `assertNoScopeEscape` refuses it before the where-clause can miss. `connect`'s link half **is** named (#98), so it passes the guard and fails on the scope instead.

Excluding the named row from the clear is the obvious fix and is what the #361 commit message rules out on cost grounds: it needs the caller's key resolved against the child before the clear can be written, which is a lookup on every `connect` to save one on a rare one. That trade was made without this case in view — a lookup that also removes a live failure is a different trade from one that only saves a redundant write.

## Scope

Only a to-one (`displaces`), only where the child's key is nullable, and only where the child's policy scopes on that same key. `set` is refused in this shape and `disconnect` has no link half, so `connect` is the only operand that reaches it. `connectOrCreate`'s hit branch reaches it too but fails differently — see the sibling issue.

## Acceptance

- `connect` naming the row that is already linked succeeds under an FK-scoped child policy, as it does without one, and `M15b` keeps its meaning at the policy level rather than only in the differential.
- Whatever the fix, an incumbent the caller cannot see is still not detached — the rule `clearLinks` exists for.
[...]

PR #378, opened 2026-08-09
Close #373: connectOrCreate's hit branch names the foreign key as the ORM's

### LINKED ISSUE #373: `connectOrCreate`'s hit branch does not name the foreign key as ORM-authored, where `connect` does
`connect`'s repoint passes `[childField]` as the ORM-authored column list, so a child whose policy scopes on its own foreign key is not refused for a write the caller never made (#98). `connectOrCreate`'s hit branch does the same repoint, ten lines up, and does not.

```ts
// packages/gemi/orm/compile/nested-writes.ts — connectOrCreate, the `found` branch
await executor.exec(
  relation.model,
  "update",
  { where: item.where, data: { [childField]: parent[parentField] }, select: … },
  false,
  //  <- no [childField] here; `connect` below passes it
);
```

## Measured

Same fixture, same call, `main` @ `5412fcd` — a `Cover` scoped on `{ folderId: 2 }`:

```
connect          -> RecordNotFoundError: No Cover found (Cover.update).
connectOrCreate  -> ScopeEscapeError: Cover.update writes 'folderId', which Cover's
                    policy also scopes on — …
```

Two spellings of one operation on two answers, which is the defect #359 was filed for. #361's own reasoning for fixing the hit branch says it plainly: the hit branch **is** a connect. It should carry the same ORM-authored list as one.

## Note on ordering

The two are related but not the same fix. Passing `[childField]` here moves `connectOrCreate` off `ScopeEscapeError` and onto the `RecordNotFoundError` #372 is about — the right direction, and the two should probably land together so the hit branch is not fixed onto a second wrong answer.

## Acceptance

- `connectOrCreate`'s hit branch repoints through the same ORM-authored column list as `connect`.
- The miss branch is untouched: it creates the far row, so there is no caller-supplied key to distinguish.
- Pinned in `nested-write-policies.test.ts` alongside the existing `connect` / `disconnect` pair, which already records that #98 is a property of the guard rather than of one operand.

nstfkc:
Closes #373.

`connect` on the side whose **child** holds the foreign key passes `[childField]` to `executor.exec` as the ORM-authored column list, so a child whose policy scopes on that key is not refused for a write the caller never made (#98). `connectOrCreate`'s hit branch runs the neighbouring branch of the same function, issues the same repoint, and did not pass it.

One line of code. The rest of this is the evidence that the two really are one operation, and the two pins that stop them separating again.

## The divergence, measured

`nested-write-policies.test.ts`'s own fixture, `main` @ `5412fcd`, a child whose whole policy is `{ scope: () => ({ folderId: 2 }) }` — the plausible "my notes" scope #98 was filed about. Both relation kinds, because they fail differently and only one of them was in the issue:

| | `connect` | `connectOrCreate`, hit |
| --- | --- | --- |
| **to-many** (`Folder.notes`) | ok, `folderId` = 2 | `ScopeEscapeError` |
| **to-one** (`Folder.cover`) | `RecordNotFoundError` | `ScopeEscapeError` |

```
ScopeEscapeError: Note.update writes 'folderId', which Note's policy also
scopes on — so this update can move a row outside the scope that selected it…

  Add an onUpdate to that policy…
```

The caller wrote `connectOrCreate: { where: { id: 20 }, create: { … } }`. `folderId` is in `data` because the nested step put it there, and the message sends them to their own `data`, where there is nothing to find. That is #98's failure verbatim, on the one operand #98 did not reach.

The to-many row is the part worth reading twice: it is not two spellings of a refusal, it is a **working operand and a broken one**. On a to-one the two land differently only because `clearLinks` runs first and moves the row out of scope — the failure #372 is about.

## Why the hit branch is a `connect`, rather than merely resembling one

Measured, not remembered. Prisma 6.19.2, SQLite, `log: [{ emit: "event", level: "query" }]`, the same call spelled both ways against the template's own schema (`User.profile`, `Profile.userId @unique`), profile 10 unattached and user 1 empty:

```
=== connect ===
BEGIN IMMEDIATE
SELECT Profile.id, userId FROM Profile WHERE id = ?              [10]
SELECT User.*                FROM User    WHERE id = ?           [1]
SELECT Profile.id, userId FROM Profile WHERE userId IN (?)       [1]
SELECT Profile.id, userId FROM Profile WHERE userId IN (?)       [1]
UPDATE Profile SET userId = ? WHERE id IN (?)                    [1, 10]
SELECT User.*                FROM User    WHERE id = ?           [1]
COMMIT

=== connectOrCreate (hit) ===
BEGIN IMMEDIATE
SELECT User.*                FROM User    WHERE id = ?           [1]
SELECT Profile.id, userId FROM Profile WHERE id = ?              [10]
SELECT Profile.id, userId FROM Profile WHERE userId IN (?)       [1]
UPDATE Profile SET userId = ? WHERE id IN (?)                    [1, 10]
SELECT User.*                FROM User    WHERE id = ?           [1]
COMMIT
```

Byte-identical write, identical bindings, identical table afterwards (`[{"id":10,"userId":1}]`). The operands differ in how many times the incumbent is looked at and in what order — not in what is written. So the hit branch is not "like" a connect; it is the same statement, and provenance is a property of the statement.

This is also the fact `planForeignSide`'s existing docblock already relies on for displacement — *"the hit branch **is** a connect, so it displaces what is linked exactly as the bare operand does (#361)"*. The marker follows from the same sentence; it was simply not carried over when that landed.

**Prisma has no opinion on the rest of this.** There is no policy layer there, so the change is invisible to the differential harness — which is the point of running it anyway: 792 → 793 tests, the extra one being the case added here, and no differential case moved.

## The diff

`packages/gemi/orm/compile/nested-writes.ts`, the `found` branch of `connectOrCreate` in `planForeignSide` — `[childField]` added as `exec`'s fifth argument, with the reason at the argument. The `false` above it also picked up the comment its twin has, since the two arguments answer opposite questions about the same call and only one of them was explained.

The miss branch is deliberately untouched, which is the acceptance criterion and also just true: it `create`s the far row, so there is no caller-supplied payload for `folderId` to be confused with. `assertNoScopeEscape` does not fire on an insert at all — `onCreate` is the mechanism there.

Nothing else on this path is missing it. Every `executor.exec` in the file that writes `[childField]` was checked: `connect` (line 2045), `disconnect` (1669) and `clearLinks` (2587) carry it; `connectOrCreate`'s hit branch did not, and now does.

`set`'s link half (line 1516) still does not, and stays that way. That is a decision rather than an omission, and it is the counter-example that keeps this change honest: `set` writes the key **twice** — `null` through `clearLinks`, then the parent's key — and only the clear is marked. Marking both makes the call *succeed* and leave the rows detached, because the clear moves them outside the very scope the link then selects them by. A loud refusal beats a silent half-write until #99. The existing pin asserts exactly that and is unchanged.

## Tests

Both mutation-checked: reverting the one-line fix fails these two and nothing else in either suite.

**`a foreign-key scope allows relation operands the ORM keyed`** gains `connectOrCreate` between the existing `connect` and `disconnect`. That test's whole point, as its docblock says, is that the exemption is *"a property of the guard rather than of one operand"* — the third operand belongs in it. Position is forced by the same ordering note already there: `disconnect` nulls `folderId` and puts the row outside the scope the hit branch has to find it through. The row **count** is asserted alongside the link, because a lookup that missed would fall through to `create` and pass on the other branch — a green test for the wrong operand.

**`connectOrCreate's hit branch answers exactly as connect does`**, new, in the `a to-one whose key is on the child` describe. It compares the two outcomes against **each other** rather than against a literal, because the answer they now share is not yet the right one: `clearLinks` nulls the row the caller named before the repoint selects it by the scope, so both miss with `RecordNotFoundError`. That is #372's, and the pin for it belongs to #372 — this one asserts only that the two spellings cannot diverge, which stays true when #372 lands and both succeed. It does name one thing: the shared answer is not a `ScopeEscapeError`, since that is the divergence being fixed. The table is compared as well as the verdict — two calls can agree on "threw" while one of them wrote something first — and that comparison doubles as the non-vacuity check, since the surviving single row is what says the *hit* branch ran.

### Ordering against #372

Landing this alone moves the to-one case from `ScopeEscapeError` onto `RecordNotFoundError`. That is not a fix of the to-one case and is not claimed as one — it is the removal of a second, wrong answer, so the to-one now has exactly the one bug #372 describes instead of two. The to-many case is fixed outright. #372's fix needs no change here: it acts on `clearLinks`, and the equality pin above passes either way.

## Docs

`docs/orm.md` claimed *"`connect`, `disconnect` and the rest go through without an `onUpdate`"*. "The rest" was false in two directions — `connectOrCreate` did not, and `set` still does not. Now it names the three that do, with a paragraph saying why `set` is the exception and what would go wrong if it were not. `docs/llms-full.txt` carries the page verbatim and `orm/docs-scope.test.ts` asserts it; the embedded copy is re-synced.

## Coverage

```
                       Tests  3110 passed | 1 skipped (3111)     (base 3110)

saas-starter      Test Files  20 passed | 3 skipped (23)
                       Tests  793 passed | 133 skipped (926)     (base 792)
```

`test:types` 116 passed, no type errors. `tsc --noEmit` clean on `packages/gemi`, and on `tsconfig.generated.json`. `oxlint orm --deny-warnings` clean; `bun run lint` 79 warnings / 0 errors, unchanged. `oxfmt --check` reports the same two files as on the base commit — both were already non-conforming there, so no reformatting is included.

Postgres was **not** run locally: the change is in the plan step rather than in any SQL, and no statement text moves. CI's `postgres` job covers it.

The worktree symlink trap was checked before trusting any of the above — `readlink -f templates/saas-starter/node_modules/gemi` resolves inside this worktree, not the main checkout.

---

## Review round 1

The blocking finding was right, and chasing it found that the to-one pin was weaker than it read.

**The hit branch now short-circuits the row already linked here.** #379 gives the bare `connect` an
`alreadyLinked` guard and leaves this branch clear-then-repointing, which pulls the two spellings
apart again. The repair belongs here and is free here: the `findUnique` above already selects the
child's foreign key, so the comparison costs no statement. Gated on `displaces` — on a to-many
there is no clear, so the repoint writes the value already there, and skipping it would be an
unmeasured change to a branch this issue is not about (and would take the to-many pin's mutation
check with it).

**The to-one fixture was degenerate, and that is why it was fragile.** Under `scope: () => ({
folderId: 2 })` the only cover the policy can see is one already pointing at folder 2 — so the only
reachable `connect` is of the already-linked row, the marked `update` never ran, and the assertion
turned on which way two unrelated refusals happened to fall. Measured: with the fix reverted *and*
the short-circuit in place, that pin passed.

The scope is now `{ folderId: { in: [2, 3] } }` (`scopeOwnedFields` takes top-level keys, so the
guard still fires on `folderId`) and cover 56 starts on folder 3. Both spellings then perform a real
repoint through the guard and both resolve — pinned against the literal `"resolved"`, not only
against each other. That makes the pin **independent of #372/#379**: neither spelling is already
linked, so an already-linked short-circuit on either side cannot separate them. The already-linked
case is pinned separately, immediately after, as what the new short-circuit buys.

**Prisma parity for the short-circuit** is recorded as differential case `M15e to-one
connectOrCreate of the row already linked changes nothing`. Net nothing on both sides, so it would
be green either way — it is the parity record, since the divergence the short-circuit fixes is only
visible through a policy and the harness has none.

**Two reasons corrected, both re-measured on this fixture rather than repeated.** A miss under a
`scope` with no `onCreate` is refused by `assertCreateCovered` before anything is written:

```
connectOrCreate { where: { id: 999 }, create: { label: "made" } }
  -> UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
     Note has a policy that scopes reads but no 'onCreate' …
```

So `.resolves.toBeDefined()` is what catches a fall-through and the row count is defence in depth;
and `docs/orm.md` now names `connectOrCreate`'s **hit** branch rather than the operand whole, with
the refusal quoted. `docs/llms-full.txt` re-synced (`docs-scope.test.ts` asserts byte equality).

**Deferred:** #386's `childField` → `childFields` rename. #386 is not on `main`, so the new names
cannot be referenced from here; the two references this branch adds (the ORM-authored list and the
new `sameKey` comparison) become `childFields` and `childFields.every(...)` at merge time.

### Coverage after the round

```
packages/gemi     typecheck                        clean
packages/gemi     bun --bun vitest run             120 files, 3110 passed | 1 skipped
packages/gemi     oxlint orm --deny-warnings       0 warnings, 0 errors
saas-starter      TZ=UTC vitest run app/models     20 passed | 3 skipped, 794 passed | 133 skipped
saas-starter      bun run test:types               116 passed, no type errors
```

794 rather than 793: the new differential case. Both mutation checks re-run — reverting
`[childField]` fails the to-many pin *and* the to-one equality (`ScopeEscapeError` where `connect`
resolves); reverting the short-circuit fails the already-linked pin with `RecordNotFoundError`.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -972,14 +972,20 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
-still use the relation operands that write that key. `connect`, `disconnect` and the rest go
-through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
+still use the relation operands that write that key. `connect`, `connectOrCreate` and `disconnect`
+go through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
 rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
 those from the check.
 
 The exemption is provenance, not permission. A column *you* name is refused exactly as before —
 writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
 `onUpdate` to say so deliberately.
+
+`set` is the one operand still refused here, and deliberately. It writes the key twice — `null` to
+clear the old links, then the parent's key to make the new ones — and only the clear is recorded as
+the ORM's. Exempting both would make the call *succeed* and leave the rows detached, because the
+clear moves them outside the very scope the link then selects them by. A refusal is the better
+answer until that is fixed.
 
 `disconnect` and `delete` only exist on an `update` — a `create` has nothing linked to it yet, and
 Prisma reports them as an unknown argument there too. They differ on a row that is **not** linked to
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -972,10 +972,22 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
-still use the relation operands that write that key. `connect`, `connectOrCreate` and `disconnect`
-go through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
-rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
-those from the check.
+still use the relation operands that write that key. `connect`, `disconnect` and `connectOrCreate`'s
+**hit** branch go through without an `onUpdate`, because the scope-escape guard judges **the columns
+you supplied** rather than the ones the nested step put there: the ORM records which columns it
+wrote and exempts those from the check.
+
+The hit branch, and not `connectOrCreate` whole, because its other branch is a `create` — and a
+policy that carries a `scope` with no `onCreate` is refused there by a different rule, before the
+scope-escape guard is reached:
+
+```
+UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
+Note has a policy that scopes reads but no 'onCreate' …
+```
+
+So on a model scoped this way, `connectOrCreate` links an existing row and cannot fall back. Give
+the policy an `onCreate` if you want the fallback as well.
 
 The exemption is provenance, not permission. A column *you* name is refused exactly as before —
 writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
--- a/templates/saas-starter/app/models/nested-write-policies.test.ts
+++ b/templates/saas-starter/app/models/nested-write-policies.test.ts
@@ -1089,6 +1089,87 @@ describe("policies on nested writes", () => {
         [55, null],
       ]);
     });
+
+    /**
+     * **The same call spelled two ways has to give one answer** (#373).
+     *
+     * `connectOrCreate`'s hit branch repoints through the very `update` the
+     * bare `connect` uses — the neighbouring branch of `planForeignSide` — and
+     * used not to name `folderId` as the ORM's. Under a child scoped on its own
+     * foreign key that split the two apart:
+     *
+     *     connect          RecordNotFoundError: No Cover found (Cover.update).
+     *     connectOrCreate  ScopeEscapeError: Cover.update writes 'folderId', …
+     *
+     * — a refusal describing a write the caller never made, on the spelling
+     * that only differs by carrying a fallback it does not take.
+     *
+     * **Asserted as an equality between the two outcomes, not against a
+     * literal**, because the answer they now share is not yet the right one:
+     * `clearLinks` nulls the row the caller named before the repoint selects it
+     * by the scope, so both miss. That is #372, and it belongs to #372's pin —
+     * this one says only that the two spellings cannot diverge, which stays
+     * true when that lands and both succeed. What it does rule out by name is
+     * the guard firing on one of them, since that is the divergence this fixes.
+     *
+     * The `set` refusal in *"a foreign-key scope allows relation operands the
+     * ORM keyed"* below is the deliberate counter-example: its link half is
+     * *not* named as the ORM's, so it is still refused. That is a choice about
+     * a half-write, not an oversight — see the comment there.
+     */
+    test("connectOrCreate's hit branch answers exactly as connect does", async () => {
+      class KeyScoped extends Model {
+        static $schema = coverSchema;
+        static $policies = [{ scope: () => ({ folderId: 2 }) } as ModelPolicy];
+      }
+      register("Cover", KeyScoped);
+
+      try {
+        await seedFolder();
+        await raw.unsafe(
+          `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
+            `VALUES (56, 2, 'linked', 7)`,
+        );
+
+        const outcome = async (operand: unknown) => {
+          try {
+            await Model.asUser(OURS, () =>
+              Folder.$exec("update", {
+                where: { id: 2 },
+                data: { code: "ours", cover: operand },
+              }),
+            );
+            return "resolved";
+          } catch (error) {
+            return (error as Error).constructor.name;
+          }
+        };
+
+        const table = async () => {
+          const rows: any = await raw.unsafe(
+            `SELECT "id", "folderId" FROM "Cover" ORDER BY "id"`,
+          );
+          return [...rows].map((cover: any) => [cover.id, cover.folderId]);
+        };
+
+        const bare = await outcome({ connect: { id: 56 } });
+        const after = await table();
+
+        const paired = await outcome({
+          connectOrCreate: { where: { id: 56 }, create: { caption: "made" } },
+        });
+
+        expect(paired).toBe(bare);
+        // The table too, and not only the verdict: the two could agree on
+        // "threw" while one of them had already written something the other had
+        // not. This also carries the non-vacuity — cover 56 is still the only
+        // row, so the hit branch really was the branch taken.
+        expect(await table()).toEqual(after);
+        expect(paired).not.toBe("ScopeEscapeError");
+      } finally {
+        register("Cover", Cover);
+      }
+    });
   });
 
   /**
@@ -1516,6 +1597,31 @@ describe("policies on nested writes", () => {
       // While both were refusals neither wrote anything and the order did not
       // matter.
       await expect(attempt({ connect: { id: 20 } })).resolves.toBeDefined();
+
+      /**
+       * **...and the third operand that writes the key: `connectOrCreate`'s hit
+       * branch** (#373), which is a `connect` spelled with a fallback.
+       *
+       * It repoints through the same `update` the bare `connect` uses — the
+       * neighbouring branch of `planForeignSide` — and used to omit the
+       * ORM-authored column list that one passes. So one operation spelled two
+       * ways answered a foreign-key scope two different ways: `connect` went
+       * through and this raised `ScopeEscapeError` about a `folderId` the
+       * caller never wrote.
+       *
+       * Between `connect` and `disconnect` for the ordering reason above — the
+       * row has to still be inside `{ folderId: 2 }` for the hit branch to find
+       * it — which is also why the row count is asserted: a miss here would
+       * *create* a second note and pass on the operand's other branch.
+       */
+      await expect(
+        attempt({
+          connectOrCreate: { where: { id: 20 }, create: { label: "made" } },
+        }),
+      ).resolves.toBeDefined();
+      expect(await linkOf()).toBe(2);
+      expect(await raw.unsafe(`SELECT "id" FROM "Note"`)).toHaveLength(1);
+
       await expect(attempt({ disconnect: { id: 20 } })).resolves.toBeDefined();
       expect(await linkOf()).toBeNull();
     } finally {
--- a/templates/saas-starter/app/models/nested-write-policies.test.ts
+++ b/templates/saas-starter/app/models/nested-write-policies.test.ts
@@ -1096,21 +1096,29 @@ describe("policies on nested writes", () => {
      * `connectOrCreate`'s hit branch repoints through the very `update` the
      * bare `connect` uses — the neighbouring branch of `planForeignSide` — and
      * used not to name `folderId` as the ORM's. Under a child scoped on its own
-     * foreign key that split the two apart:
+     * foreign key that split the two apart, the operand that carries a fallback
+     * it does not take raising about a write the caller never made:
      *
-     *     connect          RecordNotFoundError: No Cover found (Cover.update).
+     *     connect          resolved
      *     connectOrCreate  ScopeEscapeError: Cover.update writes 'folderId', …
      *
-     * — a refusal describing a write the caller never made, on the spelling
-     * that only differs by carrying a fallback it does not take.
+     * **The scope is `{ in: [2, 3] }` rather than the bare `{ folderId: 2 }` an
+     * earlier draft used, and the widening is what makes the test mean
+     * anything.** Under `{ folderId: 2 }` the only cover this policy can see is
+     * one already pointing at folder 2 — so the only reachable `connect` is of
+     * the row already linked, and *that* case is decided before the repoint is
+     * reached (by `clearLinks` on the way in, and now by the hit branch's
+     * already-linked short-circuit). The marked `update` this issue is about
+     * never ran, and the whole assertion turned on which way two unrelated
+     * refusals happened to fall. A scope that admits two folders lets a cover be
+     * *visible on folder 3 and repointed to folder 2*, which is a real repoint
+     * through the guard, with both spellings resolving.
      *
-     * **Asserted as an equality between the two outcomes, not against a
-     * literal**, because the answer they now share is not yet the right one:
-     * `clearLinks` nulls the row the caller named before the repoint selects it
-     * by the scope, so both miss. That is #372, and it belongs to #372's pin —
-     * this one says only that the two spellings cannot diverge, which stays
-     * true when that lands and both succeed. What it does rule out by name is
-     * the guard firing on one of them, since that is the divergence this fixes.
+     * It also makes the pin **independent of #372/#379**: neither spelling is
+     * already linked here, so an already-linked short-circuit on either side
+     * cannot pull the two apart. The equality is asserted against a literal as
+     * well — `"resolved"` — so a future change that made *both* raise would be
+     * caught rather than agreed with.
      *
      * The `set` refusal in *"a foreign-key scope allows relation operands the
      * ORM keyed"* below is the deliberate counter-example: its link half is
@@ -1120,16 +1128,27 @@ describe("policies on nested writes", () => {
     test("connectOrCreate's hit branch answers exactly as connect does", async () => {
       class KeyScoped extends Model {
         static $schema = coverSchema;
-        static $policies = [{ scope: () => ({ folderId: 2 }) } as ModelPolicy];
+        static $policies = [
+          { scope: () => ({ folderId: { in: [2, 3] } }) } as ModelPolicy,
+        ];
       }
       register("Cover", KeyScoped);
 
       try {
         await seedFolder();
+        // The second folder the scope admits, so a visible cover can start off
+        // somewhere other than where it is being connected.
         await raw.unsafe(
-          `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
-            `VALUES (56, 2, 'linked', 7)`,
+          `INSERT INTO "Folder" ("id", "code", "orgId") VALUES (3, 'spare', 7)`,
         );
+
+        const park = async () => {
+          await raw.unsafe(`DELETE FROM "Cover"`);
+          await raw.unsafe(
+            `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
+              `VALUES (56, 3, 'parked', 7)`,
+          );
+        };
 
         const outcome = async (operand: unknown) => {
           try {
@@ -1152,20 +1171,45 @@ describe("policies on nested writes", () => {
           return [...rows].map((cover: any) => [cover.id, cover.folderId]);
         };
 
+        await park();
         const bare = await outcome({ connect: { id: 56 } });
         const after = await table();
 
+        await park();
         const paired = await outcome({
           connectOrCreate: { where: { id: 56 }, create: { caption: "made" } },
         });
 
+        expect(bare).toBe("resolved");
         expect(paired).toBe(bare);
         // The table too, and not only the verdict: the two could agree on
-        // "threw" while one of them had already written something the other had
-        // not. This also carries the non-vacuity — cover 56 is still the only
-        // row, so the hit branch really was the branch taken.
+        // "resolved" while one of them wrote something the other did not.
         expect(await table()).toEqual(after);
-        expect(paired).not.toBe("ScopeEscapeError");
+        // ...and the non-vacuity, which is *this* line and not the count: the
+        // cover really moved off folder 3, so the marked `update` ran rather
+        // than the operand falling through to its create branch.
+        expect(after).toEqual([[56, 2]]);
+
+        /**
+         * **The hit branch on the row already linked here writes nothing.**
+         *
+         * Cover 56 now points at folder 2, so this is the case the paragraph
+         * above says `{ folderId: 2 }` could only ever reach. Without the
+         * short-circuit the branch clears the link and then repoints the row it
+         * just nulled — and a null `folderId` is outside `{ in: [2, 3] }`, so
+         * the repoint cannot select it back and the operand raises
+         * `RecordNotFoundError` on a call that should change nothing.
+         *
+         * The bare `connect` still does exactly that (#372), which is why this
+         * asserts the hit branch alone rather than another equality: the two
+         * spellings are *allowed* to differ here until #372 lands, and pinning
+         * their agreement would go red the moment it does.
+         */
+        const again = await outcome({
+          connectOrCreate: { where: { id: 56 }, create: { caption: "made" } },
+        });
+        expect(again).toBe("resolved");
+        expect(await table()).toEqual([[56, 2]]);
       } finally {
         register("Cover", Cover);
       }
@@ -1610,9 +1654,23 @@ describe("policies on nested writes", () => {
        * caller never wrote.
        *
        * Between `connect` and `disconnect` for the ordering reason above — the
-       * row has to still be inside `{ folderId: 2 }` for the hit branch to find
-       * it — which is also why the row count is asserted: a miss here would
-       * *create* a second note and pass on the operand's other branch.
+       * row has to still be inside `{ folderId: 2 }` for the hit branch to
+       * find it.
+       *
+       * **What catches a miss is the line above, not the row count.** A miss
+       * under this policy is not a create at all: `KeyScoped` has a `scope` and
+       * no `onCreate`, so `assertCreateCovered` refuses the create branch
+       * outright, and `.resolves.toBeDefined()` fails on the
+       * `UnsupportedQueryError` — measured on this fixture:
+       *
+       *     connectOrCreate { where: { id: 999 }, create: { label: "made" } }
+       *       -> UnsupportedQueryError: gemi ORM does not support 'create' yet
+       *          (Note.create). Note has a policy that scopes reads but no
+       *          'onCreate' …
+       *
+       * The count stays as defence in depth — it is what would catch a create
+       * branch that *did* run, under some future policy or fix, while the
+       * operand still resolved.
        */
       await expect(
         attempt({
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3955,10 +3955,22 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
-still use the relation operands that write that key. `connect`, `connectOrCreate` and `disconnect`
-go through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
-rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
-those from the check.
+still use the relation operands that write that key. `connect`, `disconnect` and `connectOrCreate`'s
+**hit** branch go through without an `onUpdate`, because the scope-escape guard judges **the columns
+you supplied** rather than the ones the nested step put there: the ORM records which columns it
+wrote and exempts those from the check.
+
+The hit branch, and not `connectOrCreate` whole, because its other branch is a `create` — and a
+policy that carries a `scope` with no `onCreate` is refused there by a different rule, before the
+scope-escape guard is reached:
+
+```
+UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
+Note has a policy that scopes reads but no 'onCreate' …
+```
+
+So on a model scoped this way, `connectOrCreate` links an existing row and cannot fall back. Give
+the policy an `onCreate` if you want the fallback as well.
 
 The exemption is provenance, not permission. A column *you* name is refused exactly as before —
 writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3955,14 +3955,20 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
-still use the relation operands that write that key. `connect`, `disconnect` and the rest go
-through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
+still use the relation operands that write that key. `connect`, `connectOrCreate` and `disconnect`
+go through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
 rather than the ones the nested step put there: the ORM records which columns it wrote and exempts
 those from the check.
 
 The exemption is provenance, not permission. A column *you* name is refused exactly as before —
 writing `folderId` yourself on a model scoped by it is still a scope escape, and still needs an
 `onUpdate` to say so deliberately.
+
+`set` is the one operand still refused here, and deliberately. It writes the key twice — `null` to
+clear the old links, then the parent's key to make the new ones — and only the clear is recorded as
+the ORM's. Exempting both would make the call *succeed* and leave the rows detached, because the
+clear moves them outside the very scope the link then selects them by. A refusal is the better
+answer until that is fixed.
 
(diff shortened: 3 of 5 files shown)


templates/saas-starter/app/models/nested-write-policies.test.ts
```diff
@@ -1089,6 +1089,87 @@ describe("policies on nested writes", () => {
...
+        };
+
+        const bare = await outcome({ connect: { id: 56 } });
+        const after = await table();
+
+        const paired = await outcome({
+          connectOrCreate: { where: { id: 56 }, create: { caption: "made" } },
+        });
+
+        expect(paired).toBe(bare);
```
nstfkc, line 1184: **blocking** — this assertion fails once #379 (#372) merges, and the PR body says explicitly that it will not: *"#372's fix needs no change here: it acts on `clearLinks`, and the equality pin above passes either way."*

Measured, not predicted. I merged `origin/fix/372-connect-already-linked-writes-nothing` into this branch, kept **this** PR's test file, and reran:

```
FAIL  connectOrCreate's hit branch answers exactly as connect does
AssertionError: expected 'RecordNotFoundError' to be 'resolved'
 > 1162|         expect(paired).toBe(bare);
```

The reason is that #379 adds its `alreadyLinked` short-circuit to the **`connect` branch only** — `planForeignSide`'s `out.after.push({ operation: "connect" })` gets the `pending` loop; the `connectOrCreate` hit branch keeps its unconditional `clearLinks` + repoint. So after both land the two spellings diverge *again*, just in the other direction: `connect` writes nothing and resolves, `connectOrCreate` clears the named row out of `{ folderId: 2 }` and then cannot select it back, so it raises `RecordNotFoundError`.

That is the invariant this test exists to protect, so the test is doing its job — but it means the batch cannot merge #378 and #379 and stay green. The fix belongs in the hit branch, not in this test: it already reads `found[childField]` from the `findUnique` above it, so the short-circuit is free there in a way it is not for `connect`. Either give the hit branch the same `alreadyLinked` guard as part of addressing this review, or land the two as one commit that does.

(The to-many pin at line 1623 survives the merge — I checked; only this one goes red.)

templates/saas-starter/app/models/nested-write-policies.test.ts
```diff
@@ -1089,6 +1089,87 @@ describe("policies on nested writes", () => {
...
+        // The table too, and not only the verdict: the two could agree on
+        // "threw" while one of them had already written something the other had
+        // not. This also carries the non-vacuity — cover 56 is still the only
```
nstfkc: **should fix** — the stated reason for non-vacuity does not hold. Everything in this test rolls back: `connect` raises after `clearLinks` has nulled cover 56, and the table read `after` is `[[56, 2]]` again. So "cover 56 is still the only row" is guaranteed *whichever* branch ran, and this assertion cannot distinguish the hit branch from the create branch.

What actually carries the non-vacuity is the **verdict**, on the line above. I measured the miss under this exact policy (`{ scope: () => ({ folderId: 2 }) }`, no `onCreate`) and it is not a create at all:

```
connectOrCreate { where: { id: 999 }, create: { … } }
  -> UnsupportedQueryError: … Note has a policy that scopes reads but no
     'onCreate', and create has no where clause for a scope to narrow.
  -> table unchanged: []
```

`UnsupportedQueryError !== RecordNotFoundError`, so `expect(paired).toBe(bare)` is what would catch a lookup that fell through — the table comparison would pass. Worth rewording: the table check is a real defence against one call writing where the other did not, which is the *first* half of the comment and true; the "non-vacuity" clause attributes it to the wrong assertion.

templates/saas-starter/app/models/nested-write-policies.test.ts
```diff
@@ -1516,6 +1597,31 @@ describe("policies on nested writes", () => {
...
+       * neighbouring branch of `planForeignSide` — and used to omit the
+       * ORM-authored column list that one passes. So one operation spelled two
+       * ways answered a foreign-key scope two different ways: `connect` went
+       * through and this raised `ScopeEscapeError` about a `folderId` the
+       * caller never wrote.
+       *
+       * Between `connect` and `disconnect` for the ordering reason above — the
+       * row has to still be inside `{ folderId: 2 }` for the hit branch to find
+       * it — which is also why the row count is asserted: a miss here would
+       * *create* a second note and pass on the operand's other branch.
```
nstfkc: **nit** — same as the to-one docblock: "a miss here would *create* a second note and pass on the operand's other branch" is not what happens under this policy. `KeyScoped` has a `scope` and no `onCreate`, so a miss is refused by `assertCreateCovered` before anything is written — measured on this fixture:

```
UnsupportedQueryError: … Note has a policy that scopes reads but no 'onCreate' …
```

which means `.resolves.toBeDefined()` on line 1620 already fails on a miss, and `toHaveLength(1)` is defence-in-depth rather than the thing that catches it. Good to keep — just say what it is guarding against.

docs/orm.md
```diff
@@ -972,15 +972,21 @@ same answer you would get if it truly did not exist. That is deliberate. `connec
 tenant.
 
 A child whose policy *scopes on the foreign key itself* — `scope: () => ({ folderId: mine })` — can
-still use the relation operands that write that key. `connect`, `disconnect` and the rest go
-through without an `onUpdate`, because the scope-escape guard judges **the columns you supplied**
+still use the relation operands that write that key. `connect`, `connectOrCreate` and `disconnect`
```
nstfkc: **should fix** — `connectOrCreate` only goes through on a **hit**. Its miss branch is a `create`, and on the very model this paragraph is about — a child scoped on its own foreign key — a scope with no `onCreate` is refused outright by `assertCreateCovered`:

```
UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
Note has a policy that scopes reads but no 'onCreate' …
```

So a reader of this paragraph concludes the operand works under a foreign-key scope, and half of it does not. Literally the sentence is still true — the refusal is about `onCreate`, not `onUpdate` — but the surrounding claim is *"can still use the relation operands that write that key"*, and that is only true of `connect`, `disconnect`, and `connectOrCreate`'s hit. Naming the hit branch here costs three words and is the kind of precision the `set` paragraph you added just below already models.

Same edit needed in `docs/llms-full.txt` line 3958 to keep `docs-scope.test.ts` byte-equal.

(The `set` paragraph itself reads slightly beyond #373's scope, but the sentence it replaces — *"and the rest"* — was wrong in exactly that direction, so correcting it here is right.)

nstfkc (commented):
**Verdict: needs changes.** The one-line fix is correct, minimal and properly mutation-checked — I reproduced all of it. But the new to-one pin **goes red the moment #379 lands**, which is the one thing the PR body says it cannot do ("#372's fix needs no change here … the equality pin above passes either way"). That is measured below, not predicted. Two smaller items: a docblock and a docs sentence that each give a *reason* that does not hold.

## What I ran

Fresh worktree, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → resolves inside this worktree (not the main checkout). `build:bin`, both `gemi-orm-generator` symlinks, both `prisma generate` runs; `git status` clean afterwards.

```
packages/gemi     bun run typecheck                clean
packages/gemi     bun --bun vitest run             120 files, 3110 passed | 1 skipped
saas-starter      TZ=UTC bun --bun vitest run app/models
                                                   20 passed | 3 skipped, 793 passed | 133 skipped
packages/gemi     bunx oxlint orm --deny-warnings  0 warnings, 0 errors
gh pr checks 378                                   postgres pass, sqlite pass
```

**Mutation check — confirmed.** Deleting `[childField]` from the hit branch (line 1968) and rerunning `nested-write-policies.test.ts` gives `2 failed | 39 passed`, exactly the two pins this PR adds, with `ScopeEscapeError: Note.update writes 'folderId'` raised from `nested-writes.ts:1947`. Both tests really do fail on unfixed code.

**`oxfmt --check`** flags both touched source files. I ran `oxfmt` into a scratch copy and diffed: every hunk it wants is pre-existing 80-column wrapping across the whole file (the import block, `planOwningSide(...)` at line 315, and so on). None of it is a new line. The PR's decision not to reformat is right.

**The "nothing else on this path is missing it" claim checks out.** I walked every `executor.exec` in `nested-writes.ts` whose `data` writes `[childField]`: `disconnect` (1669), `connect` (2058) and `clearLinks` (2600) carry the list; the nested `create` (1797), `createMany` (1853), `upsert`'s create (1437) and `connectOrCreate`'s miss (1982) are inserts and correctly do not; `set`'s link half (1516) is the deliberate exception. The exemption also cannot over-permit here — the hit branch's `data` is exactly `{ [childField]: … }`, so there is no caller column for the marker to shadow.

**The owning side is not a second instance of this bug.** I checked: `planOwningSide` contributes `fkField` to the parent's own statement during `getOrCompile`, which runs *after* `applyPolicies`, so `assertNoScopeEscape` never sees it. Nothing to carry over.

## What I could not check

- **Postgres locally** (no container). The change is in the plan step and no statement text moves, and CI's `postgres` job is green, so I am satisfied.
- I did not re-derive the Prisma query logs in the PR body. The differential suite going 792 → 793 with no case moving is consistent with them.

## Merge note for whoever lands this batch

Beyond the blocking finding, #378 and #379 also **conflict textually** in `nested-write-policies.test.ts`: both add a test at the same spot in the `a to-one whose key is on the child` describe, both with a `KeyScoped` cover 56 fixture. `git merge` leaves three conflict hunks there. Not a defect in either PR — just don't expect these two to compose cleanly the way the session file suggests.

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs. Two findings here.

---

## 1. #378 × #379 — the known coupling, and it does break (blocking for whichever lands second)

`orm.md`'s note predicted #378's pin would survive #379. **It does not.** Measured, not reasoned.

**Textual first.** Both PRs insert tests at the same point in
`templates/saas-starter/app/models/nested-write-policies.test.ts`, both opening `class KeyScoped`
with the same body and both seeding `Cover` id 56. `git merge` produces a 3-hunk add/add conflict.
That part is trivial — take both — but it does mean neither can be a fast-forward.

**Then the semantic half.** I resolved the conflict as a union (all four tests kept, both
`nested-writes.ts` hunks auto-merged with no conflict) and ran the suite:

```
FAIL  connectOrCreate's hit branch answers exactly as connect does
AssertionError: expected 'RecordNotFoundError' to be 'resolved'
  ❯ nested-write-policies.test.ts:1341  expect(paired).toBe(bare)
```

#379 short-circuits **only** the bare `connect` branch. So `bare` resolves via `alreadyLinked`
while `connectOrCreate`'s hit branch still clear-then-repoints and raises. #378's pin is an
equality between the two spellings, and #379 is precisely what pulls them apart.

Note what this means beyond the test: **#378's stated guarantee — "the same call spelled two ways
has to give one answer" — stops holding on `main` once both land.** The pin is not being
over-strict; it is reporting a real re-divergence.

**The fix is the one `orm.md` anticipated** ("expect `connectOrCreate`'s hit branch to want the
same short-circuit via the `found[childField]` it already reads"). The hit branch already selects
the child's foreign key into `found`, so it is three lines before the `displaces` clear:

```ts
if (found) {
  if (displaces && sameKey(found[childField], parent[parentField])) continue;
  ...
```

I applied exactly that on the merged pair: **44/44 in that file, 3110 passed in `packages/gemi`,
796 passed in the template suite** — no regressions. (On a tree that also has #386, the same repair
has to be spelled compositely — see below.)

**Suggested ordering:** merge #379 and #378 adjacently, #379 first, and land the hit-branch
short-circuit as part of #378 (it is #378's invariant that fails without it).

---

## 2. #378 × #386 and #379 × #386 — a clean merge that does not compile

This is the `#92`/`#102` shape the `ci.yml` docblock exists for, and it is the sharpest one in the
batch: **`git merge` reports no conflict at all, and `tsc` then fails.**

#386 deletes the single-field bindings in `planForeignSide` —

```
```

— and both #378 and #379 add *new* references to the old names in hunks #386 does not touch, so
they merge silently and then:

```

```

Caught by `bun run typecheck`, so it is loud rather than silent — but it *will* stop the batch.

**What each PR owes on rebase.** #378's is mechanical; #379's needs the helper generalised, because
`sameKey` stays scalar-only under #386:

- **#378** — `[childField]` as the ORM-authored column list becomes `childFields` (the whole
  array), matching what #386 passes at every other `exec` in the function.
- **#379** — the call site becomes
  `alreadyLinked(relation.model, childFields, item, valuesOf(parent, parentFields), executor)`, and
  `alreadyLinked` itself takes `readonly string[]` / `readonly unknown[]`, selects
  `keySelect(fields)`, and compares
  `fields.every((f, i) => sameKey(named[f], values[i]))`. A partial tuple match must **not** count
  as already-linked; `sameKey` already returns false on a null either side, which lines up with
  #386's `isLinked` rule.

---

### What I verified by actually merging

I built the full ten-PR integration branch locally (all ten merged, both textual conflicts
resolved, the four repairs above applied) and ran the whole of CI's matrix against it:

| check | result |
| --- | --- |
| live Postgres via `postgres.sh` | **1247 passed** (the only 2 failures are the harness's designed one-dialect-at-a-time refusal) |

So the batch *is* integrable — it just does not integrate by itself.

templates/saas-starter/app/models/nested-write-policies.test.ts
nstfkc: **Addressed** — and you were right about more than the merge.

The short-circuit is in, on this branch, gated on `displaces`:

```ts
if (displaces && sameKey(found[childField], parent[parentField])) {
  continue;
}
```

`displaces` because that is the only shape the clear runs in. On a to-many there is no clear, so the repoint just writes the value already there — skipping it there would be an unmeasured change, and it would also kill the to-many pin's mutation check, since that fixture connects note 20 while it is already on folder 2.

**But landing it alone turned this pin red in the other direction, and that was the useful part.** Measured on this branch, no #379:

```
AssertionError: expected 'resolved' to be 'RecordNotFoundError'
```

`connectOrCreate` short-circuits and resolves; `connect` still clear-then-repoints and raises. Which sent me back to the fixture, and the fixture was the actual problem: under `scope: () => ({ folderId: 2 })` **the only cover this policy can see is one already pointing at folder 2**. So the only reachable `connect` here was of the already-linked row, the marked `update` never ran at all, and the equality was between two refusals that have nothing to do with #373. The proof is a mutation check that should not have passed:

```
short-circuit in place + [childField] reverted  ->  1 failed | 40 passed
                                                    (this pin passed)
```

So it is now `{ folderId: { in: [2, 3] } }` — still a scope on the child's own foreign key, and `scopeOwnedFields` takes top-level keys so the guard still fires on `folderId` — with cover 56 starting on folder **3**. Both spellings then do a real repoint through the guard and both resolve, pinned against the literal `"resolved"` as well as against each other.

That makes the pin **independent of #372/#379 at the root**, rather than repaired against them: neither spelling is already linked here, so an already-linked short-circuit on either side cannot pull them apart. The already-linked case is pinned separately, right after, as what the new short-circuit buys — and reverting the short-circuit fails exactly that assertion with `RecordNotFoundError`. It deliberately does not assert `connect`'s answer there, since that one is #372's and is allowed to change.

Prisma parity for the short-circuit went in as differential case `M15e to-one connectOrCreate of the row already linked changes nothing`.

Full run after: `packages/gemi` 3110 passed, template `app/models` 794 passed, `test:types` 116, typecheck and `oxlint` clean.

templates/saas-starter/app/models/nested-write-policies.test.ts
nstfkc: **Addressed** — you are right that the clause attributed the non-vacuity to the wrong assertion, and the fix went further than a reword because the fixture behind it was the real problem.

Confirmed your measurement myself on this fixture (temporarily pointing the operand at a missing id):

```
connectOrCreate { where: { id: 999 }, create: { label: "made" } }
  -> UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
     Note has a policy that scopes reads but no 'onCreate' …
     at assertCreateCovered (packages/gemi/orm/policy.ts:1232)
```

So the miss is refused before anything is written, and "cover 56 is still the only row" was guaranteed either way — exactly as you said.

The deeper version of the same point: under `{ folderId: 2 }` the *hit* branch was equally uninformative, because the only cover the policy can see is one already pointing at folder 2. The marked `update` never ran. The test now uses `{ folderId: { in: [2, 3] } }` with cover 56 parked on folder 3, so the repoint is real, and the non-vacuity is carried by a line that could not hold otherwise:

```ts
expect(after).toEqual([[56, 2]]);
```

The cover moved off folder 3 — which only the hit branch's `update` can do. The table comparison stays for the reason you called true (one call writing where the other did not), and the docblock now says that and only that.

templates/saas-starter/app/models/nested-write-policies.test.ts
nstfkc: **Addressed.** Re-measured on this exact fixture rather than taking it on trust — pointing the operand at id 999 and rerunning:

```
AssertionError: promise rejected "UnsupportedQueryError: …" instead of resolving
  ❯ nested-write-policies.test.ts  ).resolves.toBeDefined();

Caused by: UnsupportedQueryError: gemi ORM does not support 'create' yet
(Note.create). Note has a policy that scopes reads but no 'onCreate' …
  ❯ assertCreateCovered  packages/gemi/orm/policy.ts:1232
```

Your reading is exactly right. The docblock now says the line above is what catches a miss, quotes that refusal, and keeps the count as defence in depth against a create branch that *did* run — under some future policy or fix — while the operand still resolved.

docs/orm.md
nstfkc: **Addressed**, in both files. The paragraph now reads:

> A child whose policy *scopes on the foreign key itself* … can still use the relation operands that write that key. `connect`, `disconnect` and `connectOrCreate`'s **hit** branch go through without an `onUpdate` …
>
> The hit branch, and not `connectOrCreate` whole, because its other branch is a `create` — and a policy that carries a `scope` with no `onCreate` is refused there by a different rule, before the scope-escape guard is reached:
>
> ```
> UnsupportedQueryError: gemi ORM does not support 'create' yet (Note.create).
> Note has a policy that scopes reads but no 'onCreate' …
> ```
>
> So on a model scoped this way, `connectOrCreate` links an existing row and cannot fall back. Give the policy an `onCreate` if you want the fallback as well.

Three words was the right price; I spent a paragraph, because "half of it does not work" is the thing a reader of that page needs and the *reason* is a different guard, which the sentence would otherwise leave them hunting for. `docs/llms-full.txt` re-synced with the identical text — `orm/docs-scope.test.ts` is green, so the byte equality holds.

(And thanks for the note on the `set` paragraph. Agreed on the reasoning: "and the rest" was wrong in that direction too, so leaving it would have been the larger error.)

nstfkc:
Heads-up from #379, not a review of this PR and nothing touched on your branch: **we have both named a differential case `M15e`.** Yours is `"M15e to-one connectOrCreate of the row already linked changes nothing"` at `writes.differential.test.ts:776`; #379 added `"M15e to-one connect of the already-linked row named by its unique foreign key"` beside `M15b` at :764, in review round 1. The insertion points differ, so `git merge` takes both without a conflict and the suite quietly ends up with two `M15e`s — vitest does not object to duplicate names, so it does not fail either. `orm.md`s ordering merges #379 first, so the cheap fix is `M15f` here on your rebase. Details: https://github.com/nstfkc/gemi/pull/379#issuecomment-5230925094

nstfkc:
## Review round 1 — disposition of every comment

Pushed as `c7b3f84`. **5 addressed, 1 deferred, 0 declined.** The body is fixed too: the closing keyword was in the **title only**, so `closingIssuesReferences` was empty. It now opens with `Closes #373.` and resolves.

| # | comment | disposition |
| --- | --- | --- |
| 1 | *blocking* — the to-one pin goes red once #379 lands | **addressed**, and the fixture behind it rebuilt |
| 2 | *should fix* — the non-vacuity clause names the wrong assertion | **addressed** |
| 3 | *nit* — "a miss would create a second note" is not what happens | **addressed** |
| 4 | *should fix* — `docs/orm.md` should name the *hit* branch | **addressed**, both files |
| X1 | [cross-PR] #378 × #379 — the hit branch needs the same short-circuit | **addressed** — it is here, and the pin no longer depends on it |
| X2 | [cross-PR] #386 renames `childField` → `childFields` | **deferred to merge time** |

---

### 1 + X1 — the short-circuit, and the pin that was weaker than it read

The short-circuit is in, gated on `displaces`:

```ts
if (displaces && sameKey(found[childField], parent[parentField])) {
  continue;
}
```

Free here in a way it is not for the bare `connect`, exactly as you said: the `findUnique` above already selects the child's foreign key. `displaces` gates it because the clear is what makes the redundant repoint *wrong* — a to-many has no clear, so the repoint writes the value already there, and skipping it there would be an unmeasured change that also takes the to-many pin's mutation check with it.

**Landing it alone turned the pin red in the other direction** — `expected 'resolved' to be 'RecordNotFoundError'` — and that is what exposed the real defect. Under `scope: () => ({ folderId: 2 })` **the only cover the policy can see is one already pointing at folder 2**. So the only reachable `connect` was of the already-linked row, the marked `update` never ran, and the equality was between two refusals that have nothing to do with #373. Measured, and it should not have been possible: with the short-circuit in and `[childField]` reverted, that pin *passed*.

The fixture is now `{ folderId: { in: [2, 3] } }` — still a scope on the child's own foreign key, and `scopeOwnedFields` takes top-level keys so the guard still fires — with cover 56 starting on folder **3**. Both spellings then do a real repoint through the guard and both resolve, pinned against the literal `"resolved"` as well as against each other.

That answers X1 at the root rather than patching it: **neither spelling is already linked in this fixture, so an already-linked short-circuit on either side cannot separate them.** The pin is #372/#379-invariant now. The already-linked case is pinned immediately after as what the short-circuit buys, and it deliberately does *not* assert `connect`'s answer there — that one is #372's and is allowed to change.

Prisma parity for the short-circuit went in as differential case `M15e to-one connectOrCreate of the row already linked changes nothing`.

### 2 + 3 — two reasons that did not hold

Both re-measured on this fixture rather than repeated. A miss under a `scope` with no `onCreate` is refused by `assertCreateCovered` before anything is written, so the verdict is what catches a fall-through and the row count is defence in depth. The to-one docblock's non-vacuity now points at `expect(after).toEqual([[56, 2]])` — the cover moved off folder 3, which only the hit branch's `update` can do.

### 4 — docs

`docs/orm.md` names `connectOrCreate`'s **hit** branch and quotes the refusal its miss branch gets under a scope with no `onCreate`, with a line telling the reader what to do about it. `docs/llms-full.txt` re-synced; `orm/docs-scope.test.ts` green.

### X2 — deferred, and it cannot be done here

#386 is not on `main`, so `childFields`/`parentFields` do not exist on this base and cannot be referenced. This branch now adds **two** references to the old names rather than one. At merge time they become:

```ts
if (
  displaces &&
  childFields.every((f, i) => sameKey(found[f], parent[parentFields[i]]))
) {
  continue;
}
…
childFields,   // the ORM-authored list
```

Same rule as your note on #379's `alreadyLinked`: a partial tuple match must not count as already-linked, and `sameKey` returning false on a null either side lines up with #386's `isLinked`.

### Confirmed in the review and left alone

The `oxfmt --check` finding (pre-existing on the base) and the walk of every `exec` writing `[childField]` — nothing to do on either. The textual add/add conflict with #379 in `nested-write-policies.test.ts` still stands (both open a `KeyScoped` class and seed `Cover` 56); take both, and the semantic half no longer follows from it.

### Coverage after the round

```
packages/gemi     typecheck                        clean
packages/gemi     bun --bun vitest run             120 files, 3110 passed | 1 skipped
packages/gemi     oxlint orm --deny-warnings       0 warnings, 0 errors
saas-starter      TZ=UTC vitest run app/models     20 passed | 3 skipped, 794 passed | 133 skipped
saas-starter      bun run test:types               116 passed, no type errors
```

794 rather than 793 — the new differential case. Mutation checks re-run both ways: reverting `[childField]` fails the to-many pin **and** the to-one equality (`ScopeEscapeError` where `connect` resolves); reverting the short-circuit fails the already-linked pin with `RecordNotFoundError`. Postgres not run locally — the change is in the plan step and no statement text moves. Worth noting for whoever reads the CI badge: `nested-write-policies.test.ts` builds its own `sqlite://` workspace, so the Postgres job never exercised these pins either way; the new differential case does run there.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `c7b3f848` for
this PR.

### Conflict 1 (#378 × #379) — **resolved**, and your short-circuit is what resolves it

Measured, not read. The add/add in `nested-write-policies.test.ts` is still textual (both sides open
a `class KeyScoped`), but it is now a **keep-both**: the two tests no longer describe the same
scenario, because you rebuilt your fixture as `{ folderId: { in: [2, 3] } }` with cover 56 parked on
folder 3 while #379's stays `{ folderId: 2 }`. Resolved that way, all four tests in that `describe`
coexist and pass — 44/44 in the file.

`expect(paired).toBe(bare)` — the assertion that failed on the previous pass with
`bare = "resolved"` / `paired = "RecordNotFoundError"` — now passes. Mutation-checked on the merged
tree: disabling the hit-branch short-circuit to `if (false && displaces && …)` fails exactly that
test with `expected 'RecordNotFoundError' to be 'resolved'`, and restoring it goes green. So the
finding was real, the repair is the right one, and it is load-bearing on the integrated tree rather
than only on your branch.

### Still open on this PR: the duplicate `M15e`

#379 flagged this and it has not moved — both branches still add a case labelled `M15e` to the same
`CASES` array in `writes.differential.test.ts`, at different insertion points, so git takes both
without a conflict:

```
831:  ["M15e to-one connect of the already-linked row named by its unique foreign key", …   (#379)
858:  ["M15e to-one connectOrCreate of the row already linked changes nothing", …           (#378)
```

Measured on the merged tree: **both run and both pass**, so this is a naming defect rather than a
failing one — which is why nothing catches it. The cost is real but indirect: `vitest -t M15e`
selects both, and #379 has two prose cross-references (`nested-write-policies.test.ts:1118` and
`nested-writes.ts:2298`) that name `M15e` as a specific case and are now ambiguous. `M10` collides
six ways on `main` already, so the file's convention tolerates families — but these two are
unrelated cases, not a family. **Renaming yours to `M15f` on the rebase is still the right call**,
and it is a one-line change with no test consequence.

### Still merge-time-only: #386's rename (conflict 2)

Unchanged in kind, larger in size. Two of the five `TS2552` errors on the all-ten tree are this
PR's, both in the hit branch you added this round:

- `nested-writes.ts:2182` — `sameKey(found[childField], parent[parentField])`
- `nested-writes.ts:2222` — the `ormAuthored` argument `[childField]`

Both are mechanical against #386's composite spelling, and I verified the resolution compiles and
passes:

```ts
if (displaces && childFields.every((f, i) => sameKey(found[f], parent[parentFields[i]]))) continue;
…
childFields,   // was [childField]; `ormAuthored` is already `readonly string[]`
```

Note the surrounding code in your own branch already reads `keySelect(childFields)` after #386, so
this is genuinely two lines.

Closing keyword verified: `closingIssuesReferences` resolves to #373.

Issue #363, opened 2026-08-08
An owning-side `connect` into an occupied to-one collides, where Prisma detaches the incumbent


nstfkc:
The mirror of #361, on the other side of the key. Where **this row** holds the foreign key and the row it is being pointed at already has a partner, gemi fails with a unique violation. Prisma answers the same call: it **detaches** the incumbent — a *sibling row of the model being written* — and takes the link.

```prisma
model User    { profile Profile? }
model Profile { userId Int? @unique
                user   User?  @relation(fields: [userId], references: [id]) }
```

```ts
// profile 1 points at user 1; user 2 already holds profile 3
await Profile.update({
  where: { id: 1 },
  data: { user: { connect: { id: 2 } } },
})
```

## Measured

Prisma 6.19.2, SQLite:

```
prisma  -> ok.  Profile rows afterwards:
              (1, "seed",   2)     the row being written takes the link
              (3, "theirs", null)  user 2's old profile, orphaned not deleted

gemi    -> UniqueConstraintError on Profile.userId
           nothing written
```

`connectOrCreate` hitting an existing row does the same thing; its *miss* branch creates a new far row and so has no incumbent to displace.

## Why it is not #361's fix

#361 displaces on the side where the **child** holds the key, and the row it clears is a row of the *child* model — which is what `clearLinks` does, through the child's own `findMany` / `updateMany`, so the child's policies decide what "currently linked" means.

Here the collision is between two rows of the model the statement is already writing: pointing `Profile 1` at user 2 collides with `Profile 3`, which is the same table. Nothing in `planOwningSide` writes a sibling row today, and the two questions it raises have no answer yet in this file:

- **Which rows may be cleared.** The clear would go through this model's own `$exec` un-pre-scoped, so its policies narrow it — but the row being *updated* has already been through them once, so the call would consult one model's scope twice for two different purposes.
- **How the target key is known.** `connect: { publicId: "…" }` resolves the referenced value with a lookup, so the clear cannot be written until that has run — which orders a write between two reads that are currently independent.

The relation also has to be known to be **one-to-one** rather than many-to-one for any of this to apply, and from this side that is a property of the schema — `uniques` containing the foreign-key column — rather than of `relation.kind`, which says "one" for both.

## Pinned

`templates/saas-starter/app/models/writes.differential.test.ts`, in the `describe("where the two clients still disagree")` block. It asserts **both** clients' answers, so it fails the day either side moves.

## Acceptance

[...]

PR #379, opened 2026-08-09
Close #372: a nested connect of the row already linked writes nothing

### LINKED ISSUE #372: A nested `connect` of the row already linked fails when the child scopes on its own foreign key

nstfkc:
#361 gave the foreign-side `connect` a `clearLinks` so it could displace an incumbent. On the one shape where the caller names the row that is **already** linked, the clear nulls that row and the repoint puts the key straight back. The committed state agrees with Prisma — which is all `M15b` can see — and the two statements contradict each other, because **the repoint has to re-select the row by the column the clear just nulled**.

```
gemi @ 54d4278, Cover scoped { folderId: 2 }, cover 30 already holding folder 2

Folder.$exec("update", { where: { id: 2 }, data: { cover: { connect: { id: 30 } } } })
  -> RecordNotFoundError: No Cover found (Cover.update).
     table unchanged: [[30, 2]]   (the clear went back with the statement)

the same call, tenant-scoped Cover  -> ok
the same call, unpolicied Cover     -> ok
```

The clear moves the row out of `{ folderId: 2 }`; the repoint has to select it *by* `{ folderId: 2 }`; so the repoint matches nothing. Nothing is written either way and the error names a model and an operation the caller never wrote. Before #361 the clear did not run and this call succeeded.

**A policy is one way to put that column into the repoint's `where`, not the only one** — this is a correction from review, and the earlier framing of this PR was narrower than the defect. Anything that makes the `where` depend on the old value reaches the same crossing, including a caller who simply names the row *through* the unique foreign key, with no policy registered anywhere:

```
Cover unpolicied, cover 30 holds folderId 2

Folder.$exec("update", { where: { id: 2 }, data: { cover: { connect: { folderId: 2 } } } })
  -> RecordNotFoundError: No Cover found (Cover.update).   table unchanged: [[30, 2]]
```

`assertNamedRows` accepts that operand because `folderId` is in `uniques` — the same fact that makes the relation a to-one and makes `displaces` true. That shape is `error` against Prisma's `ok` with nothing policy-shaped in it, which is exactly what `writes.differential.test.ts` compares, so it is now pinned there as `M15e` (see **Tests**).

## What Prisma actually does, and how that was established

A scratch `Folder` / `Cover` schema (`Cover.folderId Int? @unique`, `Folder.cover Cover?`) pushed to a throwaway SQLite file, a generated 6.19.2 client, `log: [{ emit: "event", level: "query" }]`, and the table read back after each call. Not from the docs and not from `M15b`, which reads state and cannot see this.

`folder.update({ where: { id: 2 }, data: { cover: { connect: { id: 30 } } } })` where cover 30 already holds folder 2:

```
BEGIN IMMEDIATE
select Cover.id, Cover.folderId  where id = 30             resolve the operand
select Folder.id, Folder.code    where id = 2              the row being written
select Cover.id, Cover.folderId  where folderId in (2)     the incumbent
select Folder.id, Folder.code    where id = 2              the return read
COMMIT
```

**No `UPDATE` anywhere in it.** The same call onto an *occupied* folder — the M15 shape — logs the incumbent read twice, then `update Cover set folderId = null where id in (30)`, then `update Cover set folderId = 2 where id in (31)`. So the two shapes differ in Prisma by two statements, and gemi issued the same two for both.

The rest of the matrix, each measured rather than reasoned from that one:

| call | Prisma |
| --- | --- |
| `connect` the row already linked here | **no writes at all** — four selects between BEGIN and COMMIT |
| `connect` a loose row onto an occupied folder | incumbent cleared, named row repointed |
| `connect` onto an empty folder | repoint only, nothing cleared |
| `connect` a row linked to *another* folder | repoint only, nothing cleared |
| `connectOrCreate` hitting the row already linked here | **no writes at all** |

Two things in that log decided the shape of the fix. The first statement is Prisma resolving the operand's row **with its foreign key in the projection** — on every `connect`, not only this one; and the third shows that Prisma's own no-op path still costs it two selects, so what it skips is the pair of `UPDATE`s and nothing else.

**Correction from review**: an earlier wording of that sentence said gemi's skip is "of the writes, not of the reads". It is not — gemi's `return` sits *before* `clearLinks`, which is a `findMany` as well as an `updateMany`, so on the no-op path gemi issues no incumbent read where Prisma issues one. That is deliberate and safe rather than an oversight (`displaces` requires a nullable **unique** child key, so the named row is provably the only possible holder and the read could only ever return it), and the code comment at the `return` says so. Divergence in reads is not what the differential compares; it is recorded here so the next person holding the two logs side by side is not hunting for a select that was dropped on purpose.

## The diff

`alreadyLinked` sits beside `clearLinks`: one `findUnique` on the named row selecting the foreign key, compared with `sameKey` — the same helper `displaceSibling` uses on the other end of the key. When it answers true the operand contributes nothing: no clear, no repoint, and `pending.length === 0` returns before either.

**Only where `displaces` holds**, which is a to-one with a nullable child key. Every many-to-one and every to-many `connect` — nearly all of them — costs exactly what it did, and the array spelling is refused on a to-one above, so this is one extra select at most and never one per item.

**Un-pre-scoped**, so the child's own policies decide what the lookup can see. That is the same rule the repoint beside it follows and it is what makes the miss answer fall out rather than needing a branch: a row that does not exist and a row the caller cannot see both read as absent, both are treated as "not linked here", and both are left to the repoint, which raises `RecordNotFoundError` exactly as before. `M15c` is unchanged.

The direction matters more than the cost. A lookup that ignored the scope would answer "already linked" for another tenant's row and make the whole operand a silent no-op — a `connect` that reports success, links nothing, and tells the caller a row they cannot see is theirs. That is pinned, not just commented.

### Why not exclude the named row from the clear

That is what #361's commit message ruled out on cost grounds, and it reaches the same green table. It is the wrong repair twice over: it still issues the repoint, so a child carrying `@updatedAt` gets a stamp on a call Prisma does not write at all (#370's divergence class, arriving somewhere it does not have to), and it needs the caller's key resolved against the child *anyway* to build the `NOT`. The lookup it was declining is the one Prisma issues on every `connect` regardless.

#361 declined this trade for the fact it had in view — "a lookup on every `connect` to save one write on a rare one". Two things changed it: the write it saves is a live failure rather than a redundancy, and the lookup is not on every `connect`.

## What is *not* in this PR

`connectOrCreate`'s hit branch reaches the same crossing and fails differently — `ScopeEscapeError`, because that branch does not pass `[childField]` as the ORM-authored column list where `connect` does. That is #373, still open, and it is left alone here deliberately: fixing it is a one-line change on a different statement, and its hit branch already has the row in hand (`found[childField]`), so it needs no lookup of its own. After #373 lands, the branch should short-circuit the same way — verified still failing on this branch:

```
connectOrCreate  [fk-scoped]  -> ScopeEscapeError: Cover.update writes 'folderId', …
```

## Tests

Three in `nested-write-policies.test.ts`, in the `a to-one whose key is on the child` describe beside the existing `connect` displacement pair, on `Cover` — which needed no new fixture. Each was mutation-checked: reverting the code path it covers makes it, and only it, fail.

| test | what it pins | mutation that fails it |
| --- | --- | --- |
| `connect of the row already linked here survives a key scope` | the issue's own call, under `{ folderId: 2 }` | the short-circuit (`displaces &&` → `false &&`) |
| `that connect writes nothing, where an occupied folder writes twice` | that it succeeds by issuing **no write**, not by writing a redundant one | the same |
| `the no-op does not fire for a linked row the caller cannot see` | the lookup is scoped | the lookup's `false` → `true` (pre-scoped) |

The second counts through an `onUpdate` hook rather than through SQL, and that is not merely convenient: every write to `Cover` on this path goes through `Cover`'s own `$exec` and so through its policies, and `MUTATING` in `policy.ts` carries `update` and `updateMany` alike, so the hook sees the clear and the repoint both. Intercepting statements would have to reach inside the transaction the nested steps run in, where `database.sql` is no longer the object issuing them. Its second half is the control — the same operand naming a *loose* row counts 2 — so the zero is a measurement rather than a hook that never fires.

**And one in `writes.differential.test.ts`, added in review.** The first version of this PR argued the differential needed no new case, on the reasoning that `M15b` reads the table and the table is unchanged. That is true of `M15b`'s spelling and false of the defect: name the same row through its unique foreign key and it is `error` against `ok`, which is precisely what this harness compares (`M15c` already leans on it matching raised errors). `Profile.userId Int? @unique` makes it one entry beside `M15b`:

```ts
["M15e to-one connect of the already-linked row named by its unique foreign key", "update", {
  where: { id: 1 }, data: { profile: { connect: { userId: 1 } } },
}, ["User", "Profile"]],
```

Mutation-checked like the rest: with `displaces &&` → `false &&`, **`M15e` fails and `M15`, `M15b`, `M15c`, `M15d` all stay green** —

```
{ at: 'User.update({"where":{"id":1},"data":{"profile":{"connect":{"userId":1}}}})',
- kind: "",         threw: false      (prisma)
+ kind: "notFound", threw: true       (gemi, short-circuit reverted) }
```

It earns its place twice. It is a pin `M15b` structurally cannot be, and it is the only one of the four #372 tests that runs under **Postgres** — `nested-write-policies.test.ts` builds its own `sqlite://` workspace in `beforeAll`, so CI's postgres job never reaches the other three. `M15c`, `M15`, `M15d`, `M14`–`M14d`, `M16` and the whole owning-side `O12`–`O15` block are untouched.

## Verification

`packages/gemi` 3,111 passed / 1 skipped · `saas-starter` 799 passed / 133 skipped, SQLite · `bun run test:types` 116 passed, no type errors · the same suites against a real Postgres 16 via `app/models/postgres.sh`, **1,178 passed** — one more than before, which is `M15e` under the second dialect (the two SQLite differential suites fail-rather-than-skip while `TEST_POSTGRES_URL` is set, which is the harness's own design) · `tsc --noEmit` clean on `packages/gemi` · `oxlint` clean on all three files touched.

## Left to other PRs

- **`connectOrCreate`'s hit branch short-circuit** belongs to **#378**, whose own pin (`expect(paired).toBe(bare)`) is the thing that fails without it. See the cross-PR comment below; the repair is three lines using the `found[childField]` that branch already reads, and it is #378's invariant, not this one's.
- **#386's `childField` → `childFields` rename** is merge-time work. #386 is not on `main`, so those names do not exist on any tree this branch can be built against. `alreadyLinked` will need real generalisation there — `sameKey` stays scalar-only under #386, and a partial tuple match must not count as already-linked.

Closes #372.

### DIFF
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -2008,11 +2008,26 @@ function planForeignSide(
    * right in the table and wrong in the statements** (#372). `clearLinks` nulls
    * the very row the caller named and the repoint puts the key straight back,
    * so the committed state agrees — and that is all `M15b` could see. The two
-   * statements in between are not harmless: with a child whose policy scopes on
-   * *its own foreign key* (`{ folderId: 2 }`), the clear puts the row outside
-   * the scope the repoint then has to select it by, so the repoint matches
-   * nothing and the whole call raises `RecordNotFoundError` naming a model and
-   * an operation the caller never wrote. A call that worked before #361.
+   * statements in between are not harmless. **The repoint has to re-select the
+   * row by the column the clear just changed, so it fails whenever anything in
+   * that `where` depends on the old value** — and there are two ways to get
+   * one, not just the policy shape #372 was reported through:
+   *
+   *   - a child whose policy scopes on *its own foreign key*
+   *     (`{ folderId: 2 }`) — the scope is `AND`ed into the repoint's `where`,
+   *     and the clear has just moved the row out of it;
+   *   - **no policy anywhere**, when the caller names the row *by* that foreign
+   *     key (`connect: { folderId: 2 }`). `assertNamedRows` accepts that
+   *     operand because the column is in `uniques`, which is the same fact that
+   *     makes the relation a to-one in the first place — and then the caller's
+   *     own `where` is the thing the clear invalidates.
+   *
+   * Either way the repoint matches nothing and the whole call raises
+   * `RecordNotFoundError` naming a model and an operation the caller never
+   * wrote. A call that worked before #361. The second shape is what makes this
+   * a plain correctness bug rather than a policy interaction, and it is why the
+   * differential can pin it at all: `M15e` in `writes.differential.test.ts` is
+   * that spelling, `error` against `ok`, with no policy in it.
    *
    * So the named row is resolved first and the operand short-circuits when it
    * already points here — {@link alreadyLinked}. That is Prisma's own shape
@@ -2643,9 +2658,15 @@ async function clearLinks(
  * and the same answer read from the other end of the key: on the owning side
  * the row that must not be cleared is the one being written, here it is the one
  * being named. Both exist because clearing a row and then selecting it by the
- * column just cleared is a contradiction the moment a policy scopes on that
- * column, and both match Prisma, which issues no write for an already-linked
- * `connect` from either end.
+ * column just cleared is a contradiction the moment anything in that `where`
+ * depends on the old value — a policy scoping on the column, or a caller who
+ * named the row by it.
+ *
+ * **Prisma issues no write for an already-linked `connect` from either end, and
+ * only this end matches that outright.** The owning side skips the *clear* and
+ * still emits the repoint — {@link displaceSibling}'s docblock says so in plain
+ * terms, and that residual divergence is #370's class, not something #363/#375
+ * closed. Do not read the symmetry above as parity.
  *
  * **Un-pre-scoped, so the child's own policies decide what this can see** — the
  * same rule every other statement in this step follows, and it is what makes
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -2001,19 +2001,42 @@ function planForeignSide(
    *
    *     connect onto an empty to-one          attaches           agreed already
    *     connect a child of another parent     repoints it        agreed already
-   *     connect the row already linked here   no net change      agreed already
+   *     connect the row already linked here   writes nothing     see below
    *     connect onto an occupied to-one       displaces          the divergence
    *
-   * The third is worth its own line, because the clear and the link cross on
-   * it: `clearLinks` nulls the very row the caller named, and the update below
-   * puts the key straight back. One statement wasted on a call that changes
-   * nothing, and the alternative — excluding the named row from the clear —
-   * would need the caller's key resolved against the child before the clear can
-   * be written, which is a lookup on every `connect` to save one on a rare one.
+   * **The third is the one the clear and the link cross on, and #361 got it
+   * right in the table and wrong in the statements** (#372). `clearLinks` nulls
+   * the very row the caller named and the repoint puts the key straight back,
+   * so the committed state agrees — and that is all `M15b` could see. The two
+   * statements in between are not harmless: with a child whose policy scopes on
+   * *its own foreign key* (`{ folderId: 2 }`), the clear puts the row outside
+   * the scope the repoint then has to select it by, so the repoint matches
+   * nothing and the whole call raises `RecordNotFoundError` naming a model and
+   * an operation the caller never wrote. A call that worked before #361.
+   *
+   * So the named row is resolved first and the operand short-circuits when it
+   * already points here — {@link alreadyLinked}. That is Prisma's own shape
+   * rather than a repair invented for the scope: measured on 6.19.2/SQLite with
+   * query logging, `folder.update({ data: { cover: { connect: { id: 30 } } } })`
+   * where cover 30 already holds folder 2 logs **four selects and no `UPDATE`**
+   * between its `BEGIN` and `COMMIT`, where the same call onto an occupied
+   * folder logs the incumbent clear and the repoint. The lookup itself is
+   * Prisma's too — every `connect` above begins by selecting the operand's row
+   * with its foreign key in the projection.
+   *
+   * #361 declined this as "a lookup on every `connect` to save one write on a
+   * rare one", which was the right trade for the fact it had in view. Two
+   * things changed it: the write it saves is a live failure rather than a
+   * redundancy, and the lookup is not on every `connect` — only where
+   * `displaces` holds, which is a to-one, so every many-to-one and every
+   * to-many `connect` costs exactly what it did.
    *
    * **A miss detaches nothing**, which is Prisma's answer (P2025, nothing
-   * written) and is not a special case here: the `update` below raises, and the
-   * clear is inside the transaction the nested steps already run in.
+   * written) and is not a special case here: a row the lookup cannot find is
+   * left to the repoint, which raises, and the clear is inside the transaction
+   * the nested steps already run in. That covers the row that does not exist
+   * and the row this caller's policies hide alike — both read as absent, and
+   * both get the answer they had before this change.
    */
   out.after.push({
     relation: relation.name,
@@ -2022,11 +2045,35 @@ function planForeignSide(
       const parent = rows[0];
       if (!parent) return;
 
+      // `displaces` is only ever true on a to-one, where the array spelling is
+      // refused above — so this is one extra select at most, never one per item.
+      const pending: unknown[] = [];
+      for (const item of listOf(at(args))) {
+        if (
+          displaces &&
+          (await alreadyLinked(
+            relation.model,
+            childField,
+            item,
+            parent[parentField],
+            executor,
+          ))
+        ) {
+          continue;
+        }
+        pending.push(item);
+      }
+
+      // Nothing left to link means nothing to displace either: the only row that
+      // could hold this parent's key is the one the caller named, and it holds
+      // it already. Returning here is what makes the call write nothing at all.
+      if (pending.length === 0) return;
+
       if (displaces) {
         await clearLinks(relation.model, childField, parent[parentField], executor);
       }
 
-      for (const item of listOf(at(args))) {
+      for (const item of pending) {
         await executor.exec(
           relation.model,
           "update",
@@ -2586,6 +2633,53 @@ async function clearLinks(
     false,
     [field],
   );
+}
+
+/**
+ * **Whether the row a `connect` named already points at this parent** — the
+ * question that turns a clear-then-repoint into no statements at all (#372).
+ *
+ * The foreign-side twin of the `sameKey` test inside {@link displaceSibling},
+ * and the same answer read from the other end of the key: on the owning side
+ * the row that must not be cleared is the one being written, here it is the one
+ * being named. Both exist because clearing a row and then selecting it by the
+ * column just cleared is a contradiction the moment a policy scopes on that
+ * column, and both match Prisma, which issues no write for an already-linked
+ * `connect` from either end.
+ *
+ * **Un-pre-scoped, so the child's own policies decide what this can see** — the
+ * same rule every other statement in this step follows, and it is what makes
+ * the miss answer fall out rather than needing a branch. A row that does not
+ * exist and a row this caller cannot see both read as absent, so both are
+ * treated as "not linked here" and left to the repoint, which raises
+ * `RecordNotFoundError` for them exactly as it did before.
+ *
+ * Getting that direction backwards would be the expensive mistake: a lookup
+ * that ignored the scope could answer "already linked" for another tenant's row
+ * and make the whole operand a silent no-op, which is a `connect` that reports
+ * success and links nothing.
+ *
+ * `findUnique` rather than `findFirst` because `assertNamedRows` has already
+ * run `matchUniqueKey` over this operand — the same lookup, and the same
+ * spelling, that `connectOrCreate`'s hit branch uses one screen up.
+ */
+async function alreadyLinked(
+  model: string,
+  field: string,
+  where: unknown,
+  value: unknown,
+  executor: RelationExecutor,
+): Promise<boolean> {
+  const named = (await executor.exec(
+    model,
+    "findUnique",
+    { where, select: { [field]: true } },
+    false,
+  )) as Record<string, unknown> | null;
+
+  if (named === null || named === undefined) return false;
+
+  return sameKey(named[field], value);
 }
 
 /**
--- a/templates/saas-starter/app/models/nested-write-policies.test.ts
+++ b/templates/saas-starter/app/models/nested-write-policies.test.ts
@@ -1108,8 +1108,17 @@ describe("policies on nested writes", () => {
      * The fix resolves the named row first and short-circuits, which is also
      * what Prisma does — the same call on a generated 6.19.2 client logs four
      * selects and no `UPDATE`. So the assertion is that this succeeds and the
-     * link survives, which is exactly what the same call answers with no policy
-     * at all.
+     * link survives, which is what this call answers when the operand is
+     * `{ id: 56 }` and no policy is registered.
+     *
+     * **The scope is one way to reach the crossing, not the only one.** The
+     * repoint fails whenever anything in its `where` depends on the value the
+     * clear just nulled, and a caller who spells the operand as
+     * `connect: { folderId: 2 }` supplies that himself — no policy needed.
+     * That shape is pinned in the differential as `M15e`, where it shows up as
+     * `error` against `ok`; this file cannot host it, because it is
+     * SQLite-only by construction (its `beforeAll` builds its own `sqlite://`
+     * workspace, so the Postgres job never reaches these three tests).
      *
      * Read this next to `"a foreign-key scope allows relation operands the ORM
      * keyed"` further down, which is the same scope on the to-many side and
--- a/templates/saas-starter/app/models/nested-write-policies.test.ts
+++ b/templates/saas-starter/app/models/nested-write-policies.test.ts
@@ -1089,6 +1089,185 @@ describe("policies on nested writes", () => {
         [55, null],
       ]);
     });
+
+    /**
+     * **`connect` naming the row that is already linked here, under a child
+     * scoped on its own foreign key** — #372, and the case #361's displacement
+     * broke without any test being able to see it.
+     *
+     * `M15b` pins this call in the differential and stays green either way,
+     * because it reads the *table*: `clearLinks` nulled the named row and the
+     * repoint put the key straight back, so the committed state agreed with
+     * Prisma's. What it could not see is that the two statements contradict each
+     * other the moment a policy scopes on the column they both write. The clear
+     * moves the row out of `{ folderId: 2 }`; the repoint has to select it *by*
+     * `{ folderId: 2 }`; so the repoint matched nothing and the call raised
+     * `RecordNotFoundError: No Cover found (Cover.update)` — a model and an
+     * operation the caller never wrote, on a call that worked before #361.
+     *
+     * The fix resolves the named row first and short-circuits, which is also
+     * what Prisma does — the same call on a generated 6.19.2 client logs four
+     * selects and no `UPDATE`. So the assertion is that this succeeds and the
+     * link survives, which is exactly what the same call answers with no policy
+     * at all.
+     *
+     * Read this next to `"a foreign-key scope allows relation operands the ORM
+     * keyed"` further down, which is the same scope on the to-many side and
+     * where `set` is still refused: the two answers are different because
+     * `set`'s *link* half deliberately does not name its column as ORM-authored,
+     * so it is stopped by the guard before it can reach the scope. `connect`
+     * passes the guard (#98) and used to fail on the scope instead.
+     */
+    test("connect of the row already linked here survives a key scope", async () => {
+      class KeyScoped extends Model {
+        static $schema = coverSchema;
+        static $policies = [{ scope: () => ({ folderId: 2 }) } as ModelPolicy];
+      }
+      register("Cover", KeyScoped);
+
+      try {
+        await seedFolder();
+        await raw.unsafe(
+          `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
+            `VALUES (56, 2, 'already', 7)`,
+        );
+
+        await Model.asUser(OURS, () =>
+          Folder.$exec("update", {
+            where: { id: 2 },
+            data: { cover: { connect: { id: 56 } } },
+          }),
+        );
+
+        const covers: any = await raw.unsafe(
+          `SELECT "id", "folderId" FROM "Cover" ORDER BY "id"`,
+        );
+        expect([...covers].map((cover: any) => [cover.id, cover.folderId])).toEqual([
+          [56, 2],
+        ]);
+      } finally {
+        register("Cover", Cover);
+      }
+    });
+
+    /**
+     * ...and it succeeds by **writing nothing**, which is the half the table
+     * cannot show and the half that fixes the case above.
+     *
+     * Two repairs reach the assertion in the test before this one. Excluding the
+     * named row from the clear leaves the repoint in place, writing the value
+     * that is already there; short-circuiting the whole operand issues neither.
+     * Both leave `[[56, 2]]` behind, and only the second is what Prisma does —
+     * measured on 6.19.2/SQLite with `log: [{ emit: "event", level: "query" }]`,
+     * the call is `BEGIN IMMEDIATE`, four selects, `COMMIT`, with no `UPDATE`
+     * anywhere in it, where the same `connect` onto an *occupied* folder logs
+     * the incumbent clear and the repoint.
+     *
+     * The difference is reachable rather than theoretical: a child carrying
+     * `@updatedAt` would have its stamp bumped by the repoint on a call Prisma
+     * does not write at all. `Cover` has no such column — which is why this is
+     * asserted on the writes rather than on the row.
+     *
+     * **Counted through `onUpdate` rather than through the SQL**, which is not
+     * merely convenient: every write to `Cover` on this path goes through
+     * `Cover`'s own `$exec` and therefore through its policies, so a hook that
+     * counts sees the clear's `updateMany` and the repoint's `update` alike —
+     * `MUTATING` in `policy.ts` carries both. Intercepting statements would have
+     * to reach inside the transaction the nested steps run in, where
+     * `database.sql` is no longer the object issuing them.
+     *
+     * The occupied-folder call is in the same test as the control, so the
+     * counter is shown to count before it is trusted to count zero.
+     */
+    test("that connect writes nothing, where an occupied folder writes twice", async () => {
+      let updates = 0;
+      class Counted extends Model {
+        static $schema = coverSchema;
+        static $policies = [
+          {
+            scope: (context: any) => ({ orgId: context.user.orgId }),
+            onUpdate: (_context: any, data: any) => {
+              updates++;
+              return data;
+            },
+          } as ModelPolicy,
+        ];
+      }
+      register("Cover", Counted);
+
+      try {
+        await seedFolder();
+        await raw.unsafe(
+          `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
+            `VALUES (57, 2, 'already', 7)`,
+        );
+
+        await Model.asUser(OURS, () =>
+          Folder.$exec("update", {
+            where: { id: 2 },
+            data: { cover: { connect: { id: 57 } } },
+          }),
+        );
+
+        expect(updates).toBe(0);
+
+        // The control: the same operand where the named row is *not* the one
+        // linked here. Two writes — the incumbent's clear and the repoint — so
+        // the zero above is a measurement rather than a hook that never fires.
+        await raw.unsafe(
+          `INSERT INTO "Cover" ("id", "folderId", "caption", "orgId") ` +
+            `VALUES (58, NULL, 'loose', 7)`,
+        );
+
+        await Model.asUser(OURS, () =>
+          Folder.$exec("update", {
+            where: { id: 2 },
+            data: { cover: { connect: { id: 58 } } },
+          }),
+        );
+
+        expect(updates).toBe(2);
+      } finally {
+        register("Cover", Cover);
+      }
+    });
+
+    /**
+     * **The short-circuit is scoped, and getting that backwards is the expensive
+     * mistake** — so it is pinned rather than left to the code comment.
+     *
+     * Cover 50 holds folder 2 and belongs to the other tenant. The lookup that
+     * decides "already linked here" runs through `Cover`'s own `$exec`
+     * un-pre-scoped, so this caller's policies apply and the row reads as
+     * absent — the same rule the repoint beside it follows. The operand is
+     * therefore *not* short-circuited, the clear finds nothing it may touch, and
+     * the repoint raises.
+     *
+     * An unscoped lookup would answer "already linked" here and make the whole
+     * `connect` a silent no-op: a call that reports success, links nothing, and
+     * tells the caller a row they cannot see is theirs. That is strictly worse
+     * than the miss, which is why the direction is asserted.
+     */
+    test("the no-op does not fire for a linked row the caller cannot see", async () => {
+      await seedFolder();
+      await hiddenCover();
+
+      await expect(
+        Model.asUser(OURS, () =>
+          Folder.$exec("update", {
+            where: { id: 2 },
+            data: { cover: { connect: { id: 50 } } },
+          }),
+        ),
+      ).rejects.toThrow(RecordNotFoundError);
+
+      const covers: any = await raw.unsafe(
+        `SELECT "id", "folderId" FROM "Cover" ORDER BY "id"`,
+      );
+      expect([...covers].map((cover: any) => [cover.id, cover.folderId])).toEqual([
+        [50, 2],
+      ]);
+    });
   });
 
   /**
--- a/templates/saas-starter/app/models/writes.differential.test.ts
+++ b/templates/saas-starter/app/models/writes.differential.test.ts
@@ -736,9 +736,11 @@ const CASES: Case[] = [
 
   // M15 — `connect` onto a to-one that **already has a child** (#361), the
   // neighbour of M14 and the same displacement. Four shapes, and only this one
-  // ever diverged: the three above it — an empty to-one, a steal from another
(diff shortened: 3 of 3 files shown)


packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2001,19 +2001,42 @@ function planForeignSide(
...
-   * it: `clearLinks` nulls the very row the caller named, and the update below
-   * puts the key straight back. One statement wasted on a call that changes
-   * nothing, and the alternative — excluding the named row from the clear —
-   * would need the caller's key resolved against the child before the clear can
-   * be written, which is a lookup on every `connect` to save one on a rare one.
+   * **The third is the one the clear and the link cross on, and #361 got it
+   * right in the table and wrong in the statements** (#372). `clearLinks` nulls
+   * the very row the caller named and the repoint puts the key straight back,
+   * so the committed state agrees — and that is all `M15b` could see. The two
+   * statements in between are not harmless: with a child whose policy scopes on
```
nstfkc: **should fix** — the premise is narrower than the defect, and this comment is the place a future reader decides whether the short-circuit is still needed.

"with a child whose policy scopes on *its own foreign key*" is one way to reach the crossing, not the only one. The other needs no policy at all: name the already-linked row **through the unique foreign key itself**, and the caller's own `where` is what the clear invalidates.

```
gemi @ 54d4278, Cover unpolicied, cover 30 holds folderId 2

Folder.$exec("update", { where: { id: 2 }, data: { cover: { connect: { folderId: 2 } } } })
  ->  RecordNotFoundError: No Cover found (Cover.update).   table unchanged: [[30, 2]]
```

`clearLinks` nulls `folderId`, the repoint is `update where folderId = 2`, and it matches nothing — same mechanism, no scope involved. `assertNamedRows` accepts it because `folderId` is in `uniques`, which is the same fact that makes the relation a to-one in the first place. On this branch it answers `ok`, so the fix already covers it; only the reasoning here does not.

Suggest generalising to something like: *the repoint has to re-select the row by a column the clear just changed, so it fails whenever anything in that `where` depends on the old value — a child policy scoping on the foreign key, or a caller who named the row by the foreign key.* The second half is what makes this a plain correctness bug rather than a policy-interaction one, and it is why it belongs in the differential as well as here (see the comment on the test file).

templates/saas-starter/app/models/nested-write-policies.test.ts
```diff
@@ -1089,6 +1089,185 @@ describe("policies on nested writes", () => {
...
+     * other the moment a policy scopes on the column they both write. The clear
+     * moves the row out of `{ folderId: 2 }`; the repoint has to select it *by*
+     * `{ folderId: 2 }`; so the repoint matched nothing and the call raised
+     * `RecordNotFoundError: No Cover found (Cover.update)` — a model and an
+     * operation the caller never wrote, on a call that worked before #361.
+     *
+     * The fix resolves the named row first and short-circuits, which is also
+     * what Prisma does — the same call on a generated 6.19.2 client logs four
+     * selects and no `UPDATE`. So the assertion is that this succeeds and the
+     * link survives, which is exactly what the same call answers with no policy
```
nstfkc: **should fix** — "which is exactly what the same call answers with no policy at all" is true for `connect: { id: 56 }` and false for `connect: { folderId: 2 }`, and the PR body's "the differential needs no new case" follows from the same gap.

With no policy at all, naming the already-linked row by its unique foreign key raised on the base commit and returns `ok` on this branch (measured, see the summary). Prisma answers it with four selects and no `UPDATE` — also measured, on the same 6.19.2/SQLite scratch schema this PR's evidence came from:

```
folder.update({ where: { id: 2 }, data: { cover: { connect: { folderId: 2 } } } })
  BEGIN IMMEDIATE / select Cover where folderId = ? / select Folder where id = ?
  / select Cover where folderId in (?) / select Folder where id = ? / COMMIT
```

So this shape is `error` on one client and `ok` on the other with no policies in it, which is precisely what `writes.differential.test.ts` compares — `M15c` already relies on the harness matching raised errors. `differential.prisma` has `Profile.userId Int? @unique`, so it costs one entry beside `M15b`:

```ts
["M15e to-one connect of the already-linked row named by its unique foreign key", "update", {
  where: { id: 1 }, data: { profile: { connect: { userId: 1 } } },
}, ["User", "Profile"]],
```

Worth adding: the three tests here all pin the *policy* reading of #372, and this file is SQLite-only (its `beforeAll` builds its own `sqlite://` workspace), so nothing in the PR pins the fix on the differential side or under Postgres. `M15e` gives it both.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2588,6 +2635,53 @@ async function clearLinks(
...
+/**
+ * **Whether the row a `connect` named already points at this parent** — the
+ * question that turns a clear-then-repoint into no statements at all (#372).
+ *
+ * The foreign-side twin of the `sameKey` test inside {@link displaceSibling},
+ * and the same answer read from the other end of the key: on the owning side
+ * the row that must not be cleared is the one being written, here it is the one
+ * being named. Both exist because clearing a row and then selecting it by the
+ * column just cleared is a contradiction the moment a policy scopes on that
+ * column, and both match Prisma, which issues no write for an already-linked
```
nstfkc: **nit** — "both match Prisma" over-claims for the owning side, and `displaceSibling`'s own docblock forty lines down says so in plain terms:

> **The residual divergence runs the other way and is not fixed here.** Prisma makes the whole call a no-op; gemi still emits the repoint `UPDATE` […]

So on the owning side only the *clear* is skipped; the statement still fires. This side is the one that now matches Prisma exactly. The sentence reads as if #363/#375 reached parity, which is the claim #370 exists to keep honest.

Something like: *…and Prisma issues no write for an already-linked `connect` from either end. This side now matches that outright; the owning side skips only the clear and still emits the repoint — see `displaceSibling`.*

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -2022,11 +2045,35 @@ function planForeignSide(
...
+        ) {
+          continue;
+        }
+        pending.push(item);
+      }
+
+      // Nothing left to link means nothing to displace either: the only row that
+      // could hold this parent's key is the one the caller named, and it holds
+      // it already. Returning here is what makes the call write nothing at all.
+      if (pending.length === 0) return;
```
nstfkc, line 2085: **nit**, and only about the PR body rather than this line: it says *"the third shows the incumbent read still happening on the no-op path, so the skip is of the **writes**, not of the reads"*. This `return` is before `clearLinks`, which is a `findMany` **and** an `updateMany` — so on the no-op path gemi issues no incumbent read at all, where Prisma's log shows one.

That is the right call (the unique key means the named row is the only possible holder, so the read could only ever return it), and the comment on lines 2067–2069 states it correctly. It is the PR body's sentence that describes something the code does not do — worth correcting there so the next reader comparing the two logs is not looking for a select that was deliberately dropped.

nstfkc (commented):
**Verdict: minor comments — the fix is correct and the tests are real, but the diff's account of *when* the bug bites is narrower than the bug, and that costs the PR a differential case it claims not to need.**

## What I verified

Set the worktree up properly first (`bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → this worktree, `build:bin` + reinstall, both `prisma generate`s), then:

| | |
| --- | --- |
| `packages/gemi` → `bun --bun vitest run` | 3,110 passed / 1 skipped |
| `templates/saas-starter` → `bun --bun vitest run` | 798 passed / 133 skipped |
| `packages/gemi` → `bun run typecheck` | clean |
| `oxlint` on both files | 0 warnings, 0 errors |
| `gh pr checks 379` | `sqlite` pass, `postgres` pass |

**Mutation-checked both claims in the PR body rather than taking them.** `displaces &&` → `false &&`: the two `connect` tests fail and the third passes. `alreadyLinked`'s `false` → `true` (pre-scoped): only *the no-op does not fire for a linked row the caller cannot see* fails, with `promise resolved "{ id: 2, code: 'ours', orgId: 7 }" instead of rejecting` — i.e. the silent-no-op failure mode it exists for. The mutation table in the PR body is accurate.

**Probed the edges the session file said it had not.** All on this branch, against the `Folder`/`Cover` fixture with no policy unless stated:

```
parent `create` + connect of a row linked elsewhere   ->  ok, repointed  (unchanged)
child `redact`s folderId to null                      ->  ok, degrades to clear-then-repoint as predicted
connect the already-linked row, write count           ->  0 writes
connectOrCreate hitting the already-linked row        ->  ok but 2 writes   (#373's residue, still live here)
```

The early `return` when `pending` is empty is safe on every path I could reach: `assertNamedRows` refuses `connect: null` and `connect: {}` at plan time, the array spelling is refused on a to-one, and on a to-many `displaces` is `false`, so `pending.length === 0` is only ever the short-circuit — it can never swallow a clear that used to fire. `displaces` requires a nullable **unique** child key, so the named row is provably the only holder and skipping the clear cannot strand a second one.

## The one substantive finding

**The failure is not policy-only, and it is differential-visible.** Naming the already-linked row through its own unique foreign key reproduces it with **zero policies**, because then the caller's own `where` is the thing the clear invalidates — no scope needed:

```
gemi @ 54d4278 (this PR's base), Cover unpolicied, cover 30 holds folderId 2

Folder.$exec("update", { where: { id: 2 }, data: { cover: { connect: { folderId: 2 } } } })
  ->  RecordNotFoundError: No Cover found (Cover.update).   table unchanged: [[30, 2]]

same call on this branch                                  ->  ok, [[30, 2]]
```

Prisma 6.19.2 answers it, measured on the same scratch schema this PR's evidence came from (`log: [{ emit: "event", level: "query" }]`):

```
folder.update({ where: { id: 2 }, data: { cover: { connect: { folderId: 2 } } } })

BEGIN IMMEDIATE
select Cover.id, Cover.folderId  where folderId = ?      <- resolves the operand by the FK
select Folder.id, Folder.code    where id = ?
select Cover.id, Cover.folderId  where folderId in (?)
select Folder.id, Folder.code    where id = ?
COMMIT                                                    no UPDATE, table [[30, 2]]
```

So `"Does the differential need a new case? No"` is wrong for this spelling. It is an `error` vs `ok` divergence with no policy anywhere in it, which is exactly what `writes.differential.test.ts` compares — `M15c` already leans on the harness comparing raised errors. `Profile.userId Int? @unique` in `differential.prisma` makes it a one-liner beside `M15b`:

```ts
["M15e to-one connect of the already-linked row named by its unique foreign key", "update", {
  where: { id: 1 }, data: { profile: { connect: { userId: 1 } } },
}, ["User", "Profile"]],
```

That is a pin `M15b` structurally cannot be, and it would have caught #361 on its own. Details inline.

## What I could not check

- **Postgres.** `nested-write-policies.test.ts` builds its own `sqlite://` workspace in `beforeAll`, so the three new tests do not run under `postgres.sh` or under CI's postgres job — the green postgres check says nothing about them. The change is one `findUnique`, so I have no reason to expect divergence, but nothing measures it.
- **`@updatedAt` on a to-one child.** Same gap the session file admits and the same one #363/#375 hit: no fixture has the column, so "the repoint would bump a stamp Prisma leaves alone" stays reasoned. Unchanged by this PR either way.
- I did not re-derive the four-shape Prisma matrix in the PR body beyond the two rows above; both of those reproduced exactly as written.

## Scope

Clean — two files, both #372. `connectOrCreate` is correctly left to #373/#378, and I confirmed on this branch that its hit branch still issues 2 writes where Prisma issues none, so the follow-up `orm.md` predicts is real.

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs. Two findings here.

---

## 1. #378 × #379 — the known coupling, and it does break (blocking for whichever lands second)

`orm.md`'s note predicted #378's pin would survive #379. **It does not.** Measured, not reasoned.

**Textual first.** Both PRs insert tests at the same point in
`templates/saas-starter/app/models/nested-write-policies.test.ts`, both opening `class KeyScoped`
with the same body and both seeding `Cover` id 56. `git merge` produces a 3-hunk add/add conflict.
That part is trivial — take both — but it does mean neither can be a fast-forward.

**Then the semantic half.** I resolved the conflict as a union (all four tests kept, both
`nested-writes.ts` hunks auto-merged with no conflict) and ran the suite:

```
FAIL  connectOrCreate's hit branch answers exactly as connect does
AssertionError: expected 'RecordNotFoundError' to be 'resolved'
  ❯ nested-write-policies.test.ts:1341  expect(paired).toBe(bare)
```

#379 short-circuits **only** the bare `connect` branch. So `bare` resolves via `alreadyLinked`
while `connectOrCreate`'s hit branch still clear-then-repoints and raises. #378's pin is an
equality between the two spellings, and #379 is precisely what pulls them apart.

Note what this means beyond the test: **#378's stated guarantee — "the same call spelled two ways
has to give one answer" — stops holding on `main` once both land.** The pin is not being
over-strict; it is reporting a real re-divergence.

**The fix is the one `orm.md` anticipated** ("expect `connectOrCreate`'s hit branch to want the
same short-circuit via the `found[childField]` it already reads"). The hit branch already selects
the child's foreign key into `found`, so it is three lines before the `displaces` clear:

```ts
if (found) {
  if (displaces && sameKey(found[childField], parent[parentField])) continue;
  ...
```

I applied exactly that on the merged pair: **44/44 in that file, 3110 passed in `packages/gemi`,
796 passed in the template suite** — no regressions. (On a tree that also has #386, the same repair
has to be spelled compositely — see below.)

**Suggested ordering:** merge #379 and #378 adjacently, #379 first, and land the hit-branch
short-circuit as part of #378 (it is #378's invariant that fails without it).

---

## 2. #378 × #386 and #379 × #386 — a clean merge that does not compile

This is the `#92`/`#102` shape the `ci.yml` docblock exists for, and it is the sharpest one in the
batch: **`git merge` reports no conflict at all, and `tsc` then fails.**

#386 deletes the single-field bindings in `planForeignSide` —

```
```

— and both #378 and #379 add *new* references to the old names in hunks #386 does not touch, so
they merge silently and then:

```

```

Caught by `bun run typecheck`, so it is loud rather than silent — but it *will* stop the batch.

**What each PR owes on rebase.** #378's is mechanical; #379's needs the helper generalised, because
`sameKey` stays scalar-only under #386:

- **#378** — `[childField]` as the ORM-authored column list becomes `childFields` (the whole
  array), matching what #386 passes at every other `exec` in the function.
- **#379** — the call site becomes
  `alreadyLinked(relation.model, childFields, item, valuesOf(parent, parentFields), executor)`, and
  `alreadyLinked` itself takes `readonly string[]` / `readonly unknown[]`, selects
  `keySelect(fields)`, and compares
  `fields.every((f, i) => sameKey(named[f], values[i]))`. A partial tuple match must **not** count
  as already-linked; `sameKey` already returns false on a null either side, which lines up with
  #386's `isLinked` rule.

---

### What I verified by actually merging

I built the full ten-PR integration branch locally (all ten merged, both textual conflicts
resolved, the four repairs above applied) and ran the whole of CI's matrix against it:

| check | result |
| --- | --- |
| live Postgres via `postgres.sh` | **1247 passed** (the only 2 failures are the harness's designed one-dialect-at-a-time refusal) |

So the batch *is* integrable — it just does not integrate by itself.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: **Addressed** — 04cd84c, and you are right that the premise was the narrower thing.

I reproduced your spelling rather than taking it, and in a form that stays in the repo: with the short-circuit reverted (`displaces &&` → `false &&`), the new differential case fails against the real 6.19.2 client with no policy anywhere in it —

```
{ at: 'User.update({"where":{"id":1},"data":{"profile":{"connect":{"userId":1}}}})',
- kind: "",         threw: false      (prisma)
+ kind: "notFound", threw: true       (gemi, short-circuit reverted) }
```

and `M15`, `M15b`, `M15c`, `M15d` all stay green under that same mutation, so it is the shape and not the neighbourhood.

The docblock now leads with the mechanism instead of the symptom — *the repoint has to re-select the row by the column the clear just changed, so it fails whenever anything in that `where` depends on the old value* — and then names the two ways to get one: a child policy scoping on the foreign key, and a caller naming the row by it with nothing registered at all. It also records why `assertNamedRows` accepts the second (the column is in `uniques`, which is the same fact that makes the relation a to-one and makes `displaces` true), and points at `M15e`.

The PR body's opening paragraph carried the same narrow framing; that is corrected too, and marked as a correction rather than quietly rewritten.

templates/saas-starter/app/models/nested-write-policies.test.ts
nstfkc: **Addressed** — 04cd84c. Both halves: the over-claim here, and the "the differential needs no new case" that followed from it.

`M15e` is in, worded as you suggested and placed beside `M15b`:

```ts
["M15e to-one connect of the already-linked row named by its unique foreign key", "update", {
  where: { id: 1 }, data: { profile: { connect: { userId: 1 } } },
}, ["User", "Profile"]],
```

Mutation-checked before it was believed. `displaces &&` → `false &&`:

```
FAIL  M15e …
- kind: "",         threw: false     (prisma)
+ kind: "notFound", threw: true      (gemi)
M15, M15b, M15c, M15d — pass
```

That is the pin `M15b` structurally cannot be, and — as you note — the only one of the four #372 tests the Postgres job reaches. Confirmed rather than assumed: `postgres.sh` goes 1,177 → **1,178**, the extra one being `M15e` under the second dialect, and the two SQLite differential suites still fail-rather-than-skip by the harness's own design.

The comment on this test is fixed too. *"which is exactly what the same call answers with no policy at all"* now reads *"which is what this call answers when the operand is `{ id: 56 }` and no policy is registered"*, followed by a paragraph saying plainly that the scope is one route to the crossing and `connect: { folderId: 2 }` is another, that `M15e` is where that one lives, and why it cannot live here — this file builds its own `sqlite://` workspace in `beforeAll`.

I also corrected the `M15` block comment above `M15b`. It claimed the row-already-linked shape "agreed before the fix"; it agreed *in the table*, which is not the same thing, and saying so is the whole reason `M15e` needed writing.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: **Addressed** — 04cd84c. The sentence was doing exactly what #370 exists to stop.

It now reads:

> **Prisma issues no write for an already-linked `connect` from either end, and only this end matches that outright.** The owning side skips the *clear* and still emits the repoint — `displaceSibling`'s docblock says so in plain terms, and that residual divergence is #370's class, not something #363/#375 closed. Do not read the symmetry above as parity.

The sentence before it also stopped saying "the moment a policy scopes on that column", since that was the same over-narrow premise you flagged on the `connect` step: it now says "the moment anything in that `where` depends on the old value — a policy scoping on the column, or a caller who named the row by it."

packages/gemi/orm/compile/nested-writes.ts
nstfkc: **Addressed in the PR body** — you are right, and the code is left alone, because the code is the part that is correct.

The body's sentence now describes Prisma's log rather than gemi's ("Prisma's own no-op path still costs it two selects, so what it skips is the pair of `UPDATE`s and nothing else"), and a marked correction follows it saying what gemi actually does:

> gemi's `return` sits *before* `clearLinks`, which is a `findMany` as well as an `updateMany`, so on the no-op path gemi issues no incumbent read where Prisma issues one. That is deliberate and safe rather than an oversight (`displaces` requires a nullable **unique** child key, so the named row is provably the only possible holder and the read could only ever return it), and the code comment at the `return` says so. Divergence in reads is not what the differential compares; it is recorded here so the next person holding the two logs side by side is not hunting for a select that was dropped on purpose.

No change to lines 2067–2069 — as you say, they already state it correctly, and moving the `return` after the clear to buy log symmetry would trade a correct skip for a redundant read.

nstfkc:
## Review round 1 — disposition of every point

Pushed as 04cd84c. Four addressed, two deferred, nothing declined — the substantive finding was right, and it was right about exactly the part I had reasoned rather than measured.

| # | point | disposition |
| --- | --- | --- |
| inline, `nested-writes.ts` L2011 | the premise is narrower than the defect | **addressed** |
| inline, `nested-write-policies.test.ts` L1111 | `M15e` — the differential *does* need a case | **addressed** |
| inline, `nested-writes.ts` L2647 | "both match Prisma" over-claims for the owning side | **addressed** |
| inline, `nested-writes.ts` L2070 | the PR body describes a read the code deliberately drops | **addressed** (body) |
| cross-PR 1 | `connectOrCreate`'s hit-branch short-circuit | **deferred to #378** |
| cross-PR 2 | `childField` → `childFields` under #386 | **deferred to merge time** |

### The substantive finding — you were right, and it is now pinned rather than argued

The failure is not policy-only. `connect: { folderId: 2 }` reaches the same crossing with nothing registered, because then the caller's own `where` is what the clear invalidates, and `assertNamedRows` accepts it for the same reason the relation is a to-one at all — the column is in `uniques`.

I did not take that on the description. `M15e` is in beside `M15b`, and with the short-circuit reverted (`displaces &&` → `false &&`) it fails against the real 6.19.2 client while its four neighbours stay green:

```
FAIL  M15e to-one connect of the already-linked row named by its unique foreign key
{ at: 'User.update({"where":{"id":1},"data":{"profile":{"connect":{"userId":1}}}})',
- kind: "",         threw: false      (prisma)
+ kind: "notFound", threw: true       (gemi)
M15 / M15b / M15c / M15d — pass
```

It also closes the Postgres gap you named. `nested-write-policies.test.ts` builds its own `sqlite://` workspace in `beforeAll`, so CI's postgres job never reaches the other three tests; `M15e` runs under both dialects, measured — `postgres.sh` goes 1,177 → **1,178**.

Both docblocks now state the mechanism rather than the symptom (*the repoint has to re-select the row by the column the clear just changed, so it fails whenever anything in that `where` depends on the old value*), and the `M15` block comment stopped saying the already-linked shape "agreed before the fix" — it agreed *in the table*, which is the distinction the whole issue turns on.

### Deferred 1 — `connectOrCreate`'s hit branch belongs to #378

Not declined; it is a real re-divergence and the fix is the three lines you and `orm.md` both describe. It belongs to #378 because **it is #378's invariant that fails without it.** `expect(paired).toBe(bare)` is an equality between the two spellings, #378 owns that pin, and the repair uses `found[childField]` — a binding inside the branch #378 rewrites. Landing it here would put my edit in the exact hunk #378 changes, guaranteeing a second conflict in the file whose conflict you already had to resolve by hand, and would leave #378 shipping a pin that nothing in its own diff makes pass.

The merge order in `orm.md` has #379 immediately before #378 for precisely this reason, so #378 rebases onto this branch and applies it there — one author, one hunk, one rebase.

### Deferred 2 — the `#386` rename is merge-time work

Cannot be done on this branch: #386 is not on `main`, so `childFields`/`parentFields` do not exist on any tree this can be built against, and writing them would break the build here to fix a build that does not exist yet.

Recording the spec so whoever resolves it does not re-derive it — and noting that unlike #378's, it is **not** mechanical:

- call site → `alreadyLinked(relation.model, childFields, item, valuesOf(parent, parentFields), executor)`
- signature → `readonly string[]` / `readonly unknown[]`, selecting `keySelect(fields)`
- comparison → `fields.every((f, i) => sameKey(named[f], values[i]))`; `sameKey` stays scalar-only under #386
- **a partial tuple match must not count as already-linked** — a false positive there is the silent-no-op failure mode the third test in this PR exists to catch, arriving across a composite key. `sameKey` already returns false on a null either side, which lines up with #386's `isLinked` rule.

### Unchanged

The `@updatedAt` gap stays reasoned rather than measured, for the same reason #363/#375 could not close it: no fixture has the column on a to-one child. It is #370's documented class, and this PR neither widens nor narrows it.

### Verification

`packages/gemi` 3,110 passed / 1 skipped · `saas-starter` **799** passed / 133 skipped (was 798) · `bun run test:types` 116 passed, no type errors · Postgres 16 via `postgres.sh` **1,178** passed (the two SQLite differential suites fail-rather-than-skip while `TEST_POSTGRES_URL` is set — the harness's own design) · `tsc --noEmit` clean on `packages/gemi` · `oxlint` clean on all three files touched.

The PR body's closing keyword was already in the body, not only the title — `closingIssuesReferences` resolves to #372, re-checked after the edit.

nstfkc:
### One thing found while addressing the review, that nothing in the batch has flagged

**#378 and #379 now both add a differential case called `M15e`, and they will not conflict.**

```
this branch  writes.differential.test.ts:764
  ["M15e to-one connect of the already-linked row named by its unique foreign key", …]

origin/fix/373-connect-or-create-orm-authored-key:776
  ["M15e to-one connectOrCreate of the row already linked changes nothing", …]
```

Different insertion points in the same `CASES` array — mine goes in beside `M15b`, #378's after
`M15d` — so `git merge` takes both silently and the suite ends up with two cases named `M15e`.
Vitest does not object to duplicate test names, and `expectSameWrite` keys on nothing, so the
result is not a failure; it is a suite where `vitest -t "M15e"` runs two unrelated cases and any
future bisect of "which M15e regressed" starts from a wrong premise. That is the same shape as the
two conflicts in the cross-PR review — clean merge, wrong tree — only quieter, because this one
does not even fail.

**#379 merges first per the ordering in `orm.md`, so the cheap fix is for #378's case to become
`M15f` on its rebase.** I have not touched #378's branch. Flagging it here rather than renumbering
mine, because renumbering the one that lands first only moves the collision.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `04cd84ca` for
this PR.

### Conflict 1 (#378 × #379) — **resolved**

The textual add/add in `nested-write-policies.test.ts` survives but is now a clean **keep-both**
rather than a merge of two descriptions of one scenario: #378's round moved its fixture to
`{ folderId: { in: [2, 3] } }` with cover 56 parked on folder 3, while yours stays
`{ folderId: 2 }`. Resolved by keeping both tests, all four cases in that `describe` coexist and
pass (44/44 in the file).

The semantic half is fixed by #378's hit-branch short-circuit, exactly as you deferred it.
Mutation-checked on the merged tree: `if (false && displaces && …)` on the hit branch fails
#378's `expect(paired).toBe(bare)` with `expected 'RecordNotFoundError' to be 'resolved'`; restored,
green. **Your three tests are unaffected in either direction**, which is what your rebuild of the
pin was for.

### Your deferred item 2 was right, and I used your spec verbatim

`alreadyLinked` does need real generalisation under #386, not a rename — `sameKey` stays scalar-only.
The resolution that compiles and passes is the one recorded in your session file:

```ts
async function alreadyLinked(
  model: string, fields: readonly string[], where: unknown,
  values: readonly unknown[], executor: RelationExecutor,
): Promise<boolean> {
  const named = await executor.exec(model, "findUnique", { where, select: keySelect(fields) }, false);
  if (named === null || named === undefined) return false;
  return fields.every((f, i) => sameKey(named[f], values[i]));
}
```

with the call site becoming `alreadyLinked(relation.model, childFields, item, valuesOf(parent, parentFields), executor)`.
Three of the five `TS2552` errors on the all-ten tree are this PR's (`nested-writes.ts:2335`,
`:2337`, and the helper). `every` is the load-bearing detail you flagged — a *partial* tuple match
must not count as already-linked, or the silent-no-op failure your third test exists to catch
returns across a composite key. I left a comment saying so at the line.

### The `M15e` hazard you flagged is **still live**

Not fixed, and it will not fail anything by itself. Both branches still add a case labelled `M15e`
to the same `CASES` array in `writes.differential.test.ts`:

```
831:  ["M15e to-one connect of the already-linked row named by its unique foreign key", …   (this PR)
858:  ["M15e to-one connectOrCreate of the row already linked changes nothing", …           (#378)
```

Measured: git takes both silently, **both run, both pass**. So the damage is to reference rather
than to coverage — and it lands on this PR specifically, because two of your docblocks name `M15e`
as a particular case and are now ambiguous:

- `nested-write-policies.test.ts:1118` — "That shape is pinned in the differential as `M15e`"
- `nested-writes.ts:2298` — "`M15e` in `writes.differential.test.ts` is …"

#378 renaming to `M15f` fixes the collision; these two references then need no edit, since yours
keeps the name. I have asked #378 to do the rename on its rebase. Flagging here only so the
references are checked rather than assumed once it does.

Closing keyword verified: `closingIssuesReferences` resolves to #372.

PR #380, opened 2026-08-09
Close #371: the JSON path filter treats undefined as absent, and refuses null and a numeric segment

### LINKED ISSUE #371: Three gaps in the JSON path filter's operand handling: `undefined`, `null`, and a numeric path segment
Found while typing the path filter for #336 (PR #368). All three are in `compileJsonFilter` / `assertPathShape` / `assertJsonOperand` in `orm/compile/where.ts`, none is reachable from the type any more, and none is fixed by #368 — that change is type-only. Measured against `compileRead` on Postgres with the `read.test.ts` `jsonUser` fixture.

## 1. `undefined` under a path operator is not skipped

The scalar operator loop skips an operand that is `undefined` (`where.ts:1353`, `if (operand === undefined || key === "mode") continue`). `compileJsonFilter` builds `applied` from `Object.keys` alone (`:975`) and has no such guard, so:

```
{ metadata: { path: ["a"], equals: undefined } }   →  OK: ("metadata" #>> $1) = $2   // bound NULL
{ metadata: { path: ["a"] } }                      →  UnsupportedQueryError: A 'path' needs a filter beside it
```

The same query written two ways — one with the operator omitted, one with it present and undefined — gets a loud, correct refusal and a silently never-true predicate. This is the shape `where` is careful about everywhere else: an optional filter assembled from a form or a query string routinely carries `equals: undefined`, and the scalar path was given the guard for exactly that reason.

Prisma treats an `undefined` argument as absent, so it would raise the bare-path error here.

## 2. `null` under a path operator compiles to a predicate no row can satisfy

```
{ metadata: { path: ["a"], equals: null } }   →  OK: ("metadata" #>> $1) = $2   // bound NULL
```

`X = NULL` is NULL rather than true on both dialects, so the query runs, raises nothing, and matches no row. `assertJsonOperand` lets it through: the `stringly` branch refuses a non-string, and the object branch refuses an object, but `null` falls between them.

Once the sentinels are refused at a path — correctly, because `#>>` collapses an absent key and a JSON `null` into the same SQL NULL — there is no reading of `equals: null` left. The two questions it could be asking are exactly the two `DbNull` and `JsonNull` ask, and both are already refused here with a message pointing at the column.

The same operand under `string_contains` **is** refused, and the comment says why: without it "the pattern becomes `'%null%'`, which runs and returns the wrong rows". `= NULL` is the same failure with a different arithmetic — it runs and returns *no* rows.

#368 refuses `null` in the type (`JsonPathScalar = string | number | boolean`) and says in a comment that this is the one place the type goes further than the compiler, so this issue is what closes that gap. Until it does, a `null` arriving dynamically still reaches the binder.

## 3. `path: ["items", 0]` is a superset of the oracle

`assertPathShape` accepts a numeric segment (`:1249`, `typeof part === "string" || typeof part === "number"`); Prisma's generated `path` on Postgres is `string[]`. So gemi answers a path query Prisma's client will not express, which is the situation `where.ts:945-952` refuses on purpose for the SQLite operators — *"implementing them would make gemi answer a query the oracle cannot, which is precisely where a differential test stops being able to check anything."*

`#368` types it as `string | readonly (string | number)[]`, because a type-only change describing the compiler cannot remove a superset the compiler has — excluding numbers there would only stop the type saying what the runtime does, and `assertPathShape` would still accept `["items", 0]` from a `string[]`-typed variable or from `any`. If the superset should go, this is where.

Two answers, and I have no view on which:

- **Narrow `assertPathShape` to strings**, matching the oracle, and narrow `JsonPath` with it. `["items", "0"]` still reaches an array element on Postgres, since `#>` takes `text[]` and the segment arrives as text either way.
- **Keep it and pin it** as a measured, deliberate superset, the way the two-answer divergences in the differential suite are pinned.

Either is fine; leaving it undecided is what isn't, because #368 now has a comment pointing here.

nstfkc:
The three runtime gaps #368 surfaced and could not close, because #368 was type-only. All three are in the JSON path filter, none is reachable from the type any more, and each is the same failure class: **a query that runs, raises nothing, and answers wrong**.

## How this was established

A scratch schema — `model Doc { id Int @id, label String, metadata Json? }` — pushed to a throwaway `postgres:16`, a generated `@prisma/client` **6.19.2**, `log: [{ emit: "event", level: "query" }]`, and the rows read back after every call. Five seeded documents, chosen so the answers can differ:

```
a-is-1      { a: 1 }
a-is-null   { a: null }          <- the row that separates the two readings of `null`
no-a        { b: 2 }
items       { items: ["zero", "one"] }
col-null    metadata IS NULL
```

Not from the docs, and not from what #371's own text asserted — **one of its three claims turned out to be wrong**, and that changed the shape of the fix. See §2.

## 1. `undefined` under a path operator was not absent

`compileJsonFilter` built `applied` from `Object.keys` alone, which made it the one branch in `where.ts` that does not treat an `undefined` operand as absent: the scalar loop skips it (`if (operand === undefined || key === "mode") continue`), and the dispatch *into* this branch skips a `path` that is `undefined` (`if (filter.path !== undefined)`).

So the two spellings of the same query disagreed:

| gemi, before | |
| --- | --- |
| `{ path: ["a"] }` | `UnsupportedQueryError` — a path needs a filter beside it |
| `{ path: ["a"], equals: undefined }` | `("metadata" #>> $1) = $2` bound `["{\"a\"}", null]` |

The second is `= NULL`, which is NULL rather than true — it runs and matches nothing. Confirmed directly against the seeded table: `select label from "Doc" where ("metadata" #>> '{a}') = NULL` → **0 rows**. An optional filter assembled from a form or a query string carries exactly that operand for the field nobody filled in, which is why the scalar path was given the guard in the first place.

Prisma answers both spellings identically, measured:

| Prisma 6.19.2 | |
| --- | --- |
| `{ path: ["a"] }` | P2019 *"A JSON path cannot be set without a scalar filter."* |
| `{ path: ["a"], equals: undefined }` | the same P2019 |
| `{ path: ["a"], array_contains: undefined }` | the same P2019 |
| `{ path: ["a"], equals: 1, gt: undefined }` | compiles the `equals` alone, returns `a-is-1` |
| `{ path: undefined, equals: { a: 1 } }` | a filter on the **column** |

The last row was already right here — the dispatch tests `!== undefined` — and is now asserted so it stays right (`"metadata" = $1::text::jsonb`, bound `{"a":1}`). It was *claimed* asserted in the first revision of this body and was not; review caught that, and writing the assertion also corrected the SQL quoted here.

One ordering detail fell out and is pinned: dropping the key happens *before* the unknown-operator check, so `{ path: ["a"], equalz: undefined }` gets the bare-path refusal and `{ path: ["a"], equalz: 1 }` gets *"A JSON path filter takes …"*. Prisma resolves them in the same order — P2019 for the first, *"Unknown argument `equalz`. Did you mean `equals`?"* for the second.

## 2. `null` under a path operator — and the issue's reasoning for it was wrong

#371 argued that once the sentinels are refused there is "no reading of `equals: null` left", because the two questions it could be asking are the two `DbNull` and `JsonNull` ask. **Measured, that is false.** Prisma has a third reading and gives it:

```
{ metadata: { path: ["a"], equals: null } }
  SQL     ("Doc"."metadata"#>ARRAY[$1]::text[])::jsonb::jsonb = $2
  PARAMS  ["a", null]
  ROWS    ["a-is-null"]
```

Byte-identical SQL and identical rows to `equals: Prisma.JsonNull`. Prisma can ask it because it extracts with `#>` and compares as **jsonb**, where `'null'::jsonb = 'null'::jsonb` is true. `DbNull` is a genuinely different query there — `#> … IS NULL`, returning `no-a`, `items`, `col-null`.

gemi extracts with `#>>`, which yields *text*, and SQL NULL for an absent key and for a JSON null alike. Measured on the same table:

```
where ("metadata" #>> '{a}') = NULL           -> 0 rows     <- what gemi emitted
where ("metadata" #> '{a}') @> NULL           -> 0 rows     <- and under array_contains
where ("metadata" #> '{a}')::jsonb = 'null'   -> a-is-null  <- what Prisma emits
```

So this is **a divergence gemi refuses rather than a gap it closes**, and the distinction matters for the message. `null` is now an `InvalidArgumentError` under every path operator, saying that Prisma does answer it and gemi does not yet — the same shape as the object-operand refusal three lines below (*"Prisma accepts one here; gemi does not yet"*), and for the same reason: answering it means the `#> … ::jsonb` form on Postgres and `json_type` on SQLite, a second implementation of every operator per dialect rather than a guard.

`array_contains` is **included rather than exempt**, which is a scope decision worth naming. Its exemption from the object rule is because containment is the operator that takes a document; that has nothing to do with `null`. `jsonArrayContains` binds the operand raw, so `null` arrives as SQL NULL and `x @> NULL` is NULL — the same silent zero-row answer, arrived at differently. Prisma binds it as the JSON value and runs a real containment test. A `null` *inside* the document (`array_contains: [null]`) is an ordinary JSON value and still compiles; that negative is pinned, because refusing it would have been the easy over-reach.

Prisma's own behaviour under the remaining operators, for completeness — it is not uniform:

| Prisma 6.19.2 | |
| --- | --- |
| `equals: null`, `not: null` | compile, jsonb comparison, real answers |
| `array_contains: null` | compiles, real containment |
| `gt: null`, `lt: null` | **query-engine panic** — *"JSON target types only accept strings or numbers, found: null"* |
| `string_contains: null` | validation error, *"Argument `string_contains` must not be null"* |

gemi already refused the `string_*` row, with its own better message (`%null%` "runs and returns the wrong rows"); that message is kept and the new guard sits after it. The `gt`/`lt` row is a Prisma bug, so there is no oracle to match — refusing is the only answer that is not a crash or a wrong row.

## 3. `path: ["items", 0]` was a superset of the oracle

`assertPathShape` accepted `typeof part === "number"`. Prisma's generated `path` on Postgres is `string[]` and its client refuses a number **at run time as well as in the type**:

```
{ path: ["items", 0], equals: "zero" }    PrismaClientValidationError:
    Argument `path`: Invalid value provided. Expected String, provided Int.
{ path: ["items", "0"], equals: "zero" }  #>ARRAY[$1, $2]::text[]  ["items","0"]  -> items
```

That is the situation `compileJsonFilter`'s docblock refuses on purpose for the SQLite operators — *"implementing them would make gemi answer a query the oracle cannot, which is precisely where a differential test stops being able to check anything."* #371 offered two answers and had no view; this takes the first, because the rule is already written down in this file and because **nothing is lost**: `#>` takes a `text[]`, so `["items", "0"]` reaches the same element. gemi's own binder already proved it — before this change both spellings bound the identical `{"items","0"}`.

`JsonPath` narrows to `string | readonly string[]` with it. #368 typed it `string | readonly (string | number)[]` and said so explicitly: *"If the superset should go, the place is `assertPathShape`."* It has gone there, so the type follows — otherwise the type would describe a query the compiler refuses, which is #336 in the other direction. `readonly` stays, for the `as const` hoisting reason #368 gives.

SQLite is untouched: its grammar is a JSONPath string and the index lives inside it, `"$.items[0]"`.

## The diff

- **`orm/compile/where.ts`** — the three guards. The `undefined` filter in `applied`; the `null` refusal in `assertJsonOperand`, placed after the `string_*` branch so that message survives; the string-segment check in `assertPathShape`, with its own sentence rather than the grammar message, which would tell a caller who wrote an array to write an array. `assertJsonOperand`'s object check drops its now-dead `operand !== null &&`, with a line saying why it is gone.
- **`orm/types.ts`** — `JsonPath` narrowed; `array_contains` narrowed to `Exclude<JsonValue, null>`; the two docblocks that recorded these as open (*"the one place the type goes further than the compiler"*, *"If the superset should go…"*) rewritten to what is now true.
- **`docs/orm.md`** — the refusal list goes from three to four, the string-index rule is stated, and `undefined`-is-absent is stated. `docs/llms-full.txt` re-embedded, which `docs-scope.test.ts` compares verbatim.

## Tests

**`packages/gemi/orm/compile/read.test.ts` — 153 pass**, up from 140. Three new blocks, each mutation-checked: reverting the guard it covers makes it, and only it, fail.

| test | mutation that fails it |
| --- | --- |
| `an undefined operand is absent, so the path is bare` — four operators, the misspelled-key ordering, and one live operator beside an undefined one | dropping `filter[key] !== undefined` → 1 failure |
| `null at a path is refused` — six operators on Postgres, three on SQLite, `array_contains: [null]` still compiling, and nothing reaching the SQL | restoring the old `assertJsonOperand` → 8 failures, all in this block |
| `a path segment is a string, matching the oracle` — the number, a `null` segment, the string spelling's exact bind, and SQLite unaffected | re-admitting `typeof part === "number"` → 1 failure |

**`templates/saas-starter` differential — 2 new Postgres cases**, run against Prisma on a live `postgres:16`:

- `json path with a string array index` — `{ path: ["tags", "0"], equals: "a" }` against the seeded `metadata: { tags: ["a", "b"] }`;
- `json path with a string array index that misses` — `["tags", "1"]`, because a hit alone would also pass against a path that silently selected the whole array.

These do **not** pin the refusal — a result comparison cannot see one, and both would still pass with the numeric segment re-admitted. They pin the half the narrowing rests on: that the string spelling reaches the element on both clients, so nothing was removed but a spelling. The refusal is pinned at compile level, where it can be.

**`filters.test-d.ts` — 117 pass**, up from 116. The numeric segment becomes a `@ts-expect-error`, `array_contains: null` joins the `null` refusal, and `array_contains: [null]` is pinned as still valid. The positive path case is respelled `["items", "0", "id"]`.

## Verification

| | |
| --- | --- |
| `packages/gemi` unit | **3123 pass, 1 skipped**, 120 files, via `bun --bun vitest run` |
| `packages/gemi` typecheck | clean, and `tsc -p tsconfig.generated.json` clean |
| `oxlint orm --deny-warnings` | 0 warnings, 0 errors; `bun run lint` unchanged at 79 warnings |
| template, SQLite | **792 pass**, 20 files |
| template, Postgres (`-t postgres`, `--no-file-parallelism`, live `postgres:16`) | **875 pass**, 14 files |
| template type tests | **117 pass**, no type errors |

Both dialects were run locally against a real Postgres in a container, with the client regenerated for each provider the way `ci.yml` does it — the schema flip and the two generated clients are reverted in the tree.

## Loose ends a reviewer may want to argue with

- **`array_contains: null` being refused is the one judgement call beyond the issue's letter.** #371 names `assertJsonOperand`'s gap between its two branches, and `array_contains` returns before it. Left alone it goes on compiling `@> NULL` — zero rows, silently, where Prisma answers. Refusing it is four lines and one message; the alternative was to leave a known silent-wrong inside the function this PR is about.
- **`null` could be implemented on Postgres instead**, as `("metadata" #> $1)::jsonb = 'null'::jsonb`. Not done, because SQLite would then need `json_type` to answer the same question and the two would have to be kept in step — the exact split the object-operand refusal already declined. If it is ever wanted, it is one dialect method, not a guard change.
- **Narrowing `JsonPath` is a compile-time break** for anyone who wrote `path: ["items", 0]` since #368 landed — one day. The runtime refuses it either way now, so the type break is what makes the refusal visible before the query runs.
- **`oxfmt --check` reports these three files as unformatted**, and did so identically at the base commit. Not reformatted here; CI runs `oxlint`, not `oxfmt`.

Closes #371.

## Review round 1

Six points from review, all addressed; two cross-PR points deferred to the PRs that own them.

| point | disposition |
| --- | --- |
| the `null` message gave `array_contains` a mechanism it does not have | **fixed** — the message branches on the key. `array_contains` now names the raw bind and `x @> NULL`, and points at `array_contains: [null]` rather than at `equals: Prisma.JsonNull`, which would answer a containment question with an equality on the column |
| `path[1] is a undefined` / `a object`, and the array-index advice emitted for segments that are not indexes | **fixed** — the description is spelled out (`null`, `undefined`, `an object`, `an array`, `a number`), and the index sentence is gated on `typeof part === "number"` |
| `toThrow(InvalidArgumentError)` left the new guard's *placement* unpinned | **fixed** — every row asserts its message. Moving the `null` guard above the `string_*` one used to leave 153 green; it now fails 2 (`string_contains on postgres`, `string_ends_with on sqlite`) |
| `{ path: undefined, … }` claimed asserted, was not | **fixed** — asserted in the `undefined` block, and the SQL quoted in this body was wrong and is corrected |
| `it never becomes a bound NULL` passed for any throw | **fixed** — asserts the message alongside `text === ""` |
| the `undefined` fix also closes a plan-cache shape collision | **taken** — written into the `applied` comment, with the two `planKey` calls that collapse to one key |
| #385 ships `["a", 0]`, which this PR now refuses | **deferred to #385** — the three sites are in #385's files; respelling to `["a", "0"]` is its repair. Merge #380 first |
| #381 × `types.ts`: taking #381's mapped type reverts the `array_contains` narrowing | **deferred to merge time** — #381 is not on `main`, so the arm (`K extends "array_contains" ? Exclude<JsonValue, null> : …`) cannot be written against a type that does not exist here yet |

Mutation checks re-run on every test touched: reverting `filter[key] !== undefined` → 1 failure; the dispatch's `filter.path !== undefined` → `"path" in filter` → 1 failure (new); restoring the old `assertJsonOperand` in full → 8 failures; moving the `null` guard above the `string_*` one → 2 failures (new — it was 0 before this round); collapsing the `array_contains` message branch → 1 failure (new); reverting the segment description and the gating → 1 failure (new); re-admitting `typeof part === "number"` → 1 failure.

Green after: `packages/gemi` **3123 pass / 1 skipped**, typecheck clean, `oxlint orm --deny-warnings` clean, template SQLite **792 pass**, `test:types` **117 pass**. `git status` clean of generated artifacts after both `prisma generate` runs.

### DIFF
--- a/packages/gemi/orm/compile/read.test.ts
+++ b/packages/gemi/orm/compile/read.test.ts
@@ -1259,6 +1259,21 @@ describe("json path filters", () => {
       pgText({ where: { metadata: { path: ["a"], equalz: 1 } } }),
     ).toThrow(/A JSON path filter takes/);
 
+    // `path: undefined` is absent too — the dispatch into this branch tests
+    // `filter.path !== undefined`, so the filter falls through to the column
+    // arm rather than becoming a bare path. That half was already right before
+    // this change; asserted here so the rule holds on both sides of the
+    // dispatch. Prisma answers it the same way — `{ path: undefined,
+    // equals: { a: 1 } }` is a filter on the column there too.
+    const columnArgs = {
+      where: { metadata: { path: undefined, equals: { a: 1 } } },
+    };
+    expect(pgText(columnArgs)).toContain(`"metadata" = $1::text::jsonb`);
+    expect(pgText(columnArgs)).not.toContain("#>>");
+    expect(
+      compileRead(jsonUser, "findMany", columnArgs, postgres).bind(columnArgs),
+    ).toEqual([`{"a":1}`]);
+
     // ...and one live operator beside it is still a filter, so the guard drops
     // keys rather than the whole operand.
     const args = { where: { metadata: { path: ["a"], equals: 1, gt: undefined } } };
@@ -1420,31 +1435,49 @@ describe("json path filters", () => {
    * containment test.
    */
   describe("null at a path is refused", () => {
-    const operators = [
-      "equals",
-      "not",
-      "string_contains",
-      "array_contains",
-      "gt",
-      "lte",
+    /**
+     * **Each row asserts its message, not just `InvalidArgumentError`**, and
+     * that is what pins the two things the class alone cannot see.
+     *
+     * The `string_*` rows already threw before this change, from the non-string
+     * guard, with the `'%null%'` reasoning — so a class assertion passes either
+     * way and leaves the new guard's *placement* unpinned. It sits deliberately
+     * **after** that guard so the better message survives; moving it above used
+     * to leave all 153 tests green while `string_contains: null` silently swapped
+     * its message for the generic one.
+     *
+     * The `array_contains` row asserts the other sentence for the same kind of
+     * reason: that operator does not extract with `#>>`, so its refusal names
+     * the raw bind and `x @> NULL` instead of the text collapse.
+     */
+    const postgresOperators = [
+      ["equals", /null cannot be compared through a JSON path/],
+      ["not", /null cannot be compared through a JSON path/],
+      ["gt", /null cannot be compared through a JSON path/],
+      ["lte", /null cannot be compared through a JSON path/],
+      ["string_contains", /'%null%', which runs and returns the wrong rows/],
+      ["array_contains", /null cannot be tested for containment/],
     ] as const;
 
-    test.each(operators)("%s on postgres", (key) => {
-      expect(() =>
-        pgText({ where: { metadata: { path: ["a"], [key]: null } } }),
-      ).toThrow(InvalidArgumentError);
+    test.each(postgresOperators)("%s on postgres", (key, message) => {
+      const compile = () =>
+        pgText({ where: { metadata: { path: ["a"], [key]: null } } });
+      expect(compile).toThrow(InvalidArgumentError);
+      expect(compile).toThrow(message);
     });
 
     // SQLite offers only the six its dialect answers; the other four raise the
     // dialect refusal first, which is a different message and already pinned.
-    test.each(["equals", "not", "string_ends_with"] as const)(
-      "%s on sqlite",
-      (key) => {
-        expect(() =>
-          sqliteText({ where: { metadata: { path: "$.a", [key]: null } } }),
-        ).toThrow(InvalidArgumentError);
-      },
-    );
+    test.each([
+      ["equals", /null cannot be compared through a JSON path/],
+      ["not", /null cannot be compared through a JSON path/],
+      ["string_ends_with", /'%null%', which runs and returns the wrong rows/],
+    ] as const)("%s on sqlite", (key, message) => {
+      const compile = () =>
+        sqliteText({ where: { metadata: { path: "$.a", [key]: null } } });
+      expect(compile).toThrow(InvalidArgumentError);
+      expect(compile).toThrow(message);
+    });
 
     /**
      * The load-bearing negative: only the *bare* operand is refused. A `null`
@@ -1460,14 +1493,21 @@ describe("json path filters", () => {
       ]);
     });
 
-    /** Nothing of it reaches the SQL, which is the actual defect. */
+    /**
+     * Nothing of it reaches the SQL, which is the actual defect. The message is
+     * asserted alongside because a bare `text === ""` passes for *any* throw,
+     * including one from an unrelated future guard — it would say "no SQL was
+     * built" where it means "this guard built no SQL".
+     */
     test("it never becomes a bound NULL", () => {
       let text = "";
+      let message = "";
       try {
         text = pgText({ where: { metadata: { path: ["a"], equals: null } } });
-      } catch {
-        // expected
+      } catch (error) {
+        message = (error as Error).message;
       }
+      expect(message).toMatch(/null cannot be compared through a JSON path/);
       expect(text).toBe("");
     });
   });
@@ -1525,6 +1565,29 @@ describe("json path filters", () => {
     expect(() =>
       pgText({ where: { metadata: { path: ["a", null], equals: "x" } } }),
     ).toThrow(/path\[1\] is null/);
+
+    // Every other non-string reads grammatically too — `a ${typeof part}`
+    // alone gives "a undefined" and "a object", and the number is the only
+    // segment it happens to fit.
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a", undefined], equals: "x" } } }),
+    ).toThrow(/path\[1\] is undefined\./);
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a", { b: 1 }], equals: "x" } } }),
+    ).toThrow(/path\[1\] is an object\./);
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a", ["b"]], equals: "x" } } }),
+    ).toThrow(/path\[1\] is an array\./);
+
+    // And the array-index advice is advice for *one* mistake, so it is emitted
+    // for the number and withheld from the rest — a caller who wrote `null`
+    // did not write an index.
+    expect(() =>
+      pgText({ where: { metadata: { path: ["items", 0], equals: "x" } } }),
+    ).toThrow(/write \["items", "0"\] rather than \["items", 0\]/);
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a", null], equals: "x" } } }),
+    ).not.toThrow(/An array index is a key too/);
 
     // The string spelling reaches the element, and the bound path is identical
     // to what the numeric one used to produce — which is why this is a
--- a/packages/gemi/orm/compile/read.test.ts
+++ b/packages/gemi/orm/compile/read.test.ts
@@ -1226,6 +1226,52 @@ describe("json path filters", () => {
   });
 
   /**
+   * #371, the first of three. `applied` was built from `Object.keys` alone, so
+   * the one branch in this file that did **not** treat `undefined` as absent
+   * was the one where the operand goes on to be bound: the scalar loop skips it
+   * (`if (operand === undefined …) continue`) and the dispatch into this branch
+   * skips a `path` that is `undefined`.
+   *
+   * The two spellings therefore disagreed — the key omitted got a loud, correct
+   * refusal, and the key present-but-undefined compiled to `= NULL`, a
+   * predicate no row can satisfy. An optional filter assembled from a form or a
+   * query string carries exactly that.
+   *
+   * Prisma answers both identically: P2019 *"A JSON path cannot be set without
+   * a scalar filter."* Measured on a generated 6.19.2 client against Postgres.
+   */
+  test("an undefined operand is absent, so the path is bare", () => {
+    for (const key of ["equals", "not", "string_contains", "array_contains"]) {
+      expect(() =>
+        pgText({ where: { metadata: { path: ["a"], [key]: undefined } } }),
+      ).toThrow(/needs a filter beside it/);
+    }
+
+    // A *misspelled* key holding `undefined` is absent too, so the bare-path
+    // refusal wins over the unknown-operator one — which is the order Prisma
+    // resolves them in: `{ path: ["a"], equalz: undefined }` answers P2019
+    // there, and `{ path: ["a"], equalz: 1 }` answers *"Unknown argument
+    // `equalz`. Did you mean `equals`?"*.
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a"], equalz: undefined } } }),
+    ).toThrow(/needs a filter beside it/);
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a"], equalz: 1 } } }),
+    ).toThrow(/A JSON path filter takes/);
+
+    // ...and one live operator beside it is still a filter, so the guard drops
+    // keys rather than the whole operand.
+    const args = { where: { metadata: { path: ["a"], equals: 1, gt: undefined } } };
+    expect(pgText(args)).toContain(`("metadata" #>> $1) = $2`);
+    // `gt` compiles to `cast(… as real) >`, so its absence is visible.
+    expect(pgText(args)).not.toContain("as real");
+    expect(compileRead(jsonUser, "findMany", args, postgres).bind(args)).toEqual([
+      `{"a"}`,
+      "1",
+    ]);
+  });
+
+  /**
    * `array_contains` and the numeric comparisons are refused by Prisma's own
    * client on SQLite — *"Unknown argument"* — so implementing them would make
    * gemi answer a query the oracle cannot check. The message says it is a
@@ -1352,6 +1398,81 @@ describe("json path filters", () => {
   });
 
   /**
+   * #371, the second of three. `null` fell between `assertJsonOperand`'s two
+   * branches — the string branch refuses a non-string, the object branch tested
+   * `operand !== null` first — so it reached the binder and compiled to
+   * `("metadata" #>> $1) = $2` bound to NULL. `= NULL` is NULL rather than true
+   * on both dialects: the query runs, raises nothing, and matches no row.
+   *
+   * **Prisma does answer this one**, which is why it is a refusal rather than a
+   * gap being closed. Measured on a generated 6.19.2 client against Postgres:
+   * `{ path: ["a"], equals: null }` compiles to
+   * `("metadata"#>ARRAY[$1]::text[])::jsonb::jsonb = $2` and returns the rows
+   * whose `a` is the JSON value `null` — byte-identical SQL and identical rows
+   * to `equals: Prisma.JsonNull`. It can, because it extracts with `#>` and
+   * compares as `jsonb`; gemi extracts with `#>>`, which yields SQL NULL for an
+   * absent key and a JSON null alike. That is the same collapse the sentinel
+   * refusal above names, so the same answer follows.
+   *
+   * `array_contains` is refused too, and its mechanism is the other one:
+   * `jsonArrayContains` binds the operand raw, so `null` arrives as SQL NULL
+   * and `x @> NULL` is NULL. Prisma binds it as the JSON value and runs a real
+   * containment test.
+   */
+  describe("null at a path is refused", () => {
+    const operators = [
+      "equals",
+      "not",
+      "string_contains",
+      "array_contains",
+      "gt",
+      "lte",
+    ] as const;
+
+    test.each(operators)("%s on postgres", (key) => {
+      expect(() =>
+        pgText({ where: { metadata: { path: ["a"], [key]: null } } }),
+      ).toThrow(InvalidArgumentError);
+    });
+
+    // SQLite offers only the six its dialect answers; the other four raise the
+    // dialect refusal first, which is a different message and already pinned.
+    test.each(["equals", "not", "string_ends_with"] as const)(
+      "%s on sqlite",
+      (key) => {
+        expect(() =>
+          sqliteText({ where: { metadata: { path: "$.a", [key]: null } } }),
+        ).toThrow(InvalidArgumentError);
+      },
+    );
+
+    /**
+     * The load-bearing negative: only the *bare* operand is refused. A `null`
+     * inside a document under `array_contains` is an ordinary JSON value, and
+     * containment against `[null]` is a question Postgres answers.
+     */
+    test("a null inside an array_contains document still compiles", () => {
+      const args = { where: { metadata: { path: ["a"], array_contains: [null] } } };
+      expect(pgText(args)).toContain(`("metadata" #> $1) @> $2::jsonb`);
+      expect(compileRead(jsonUser, "findMany", args, postgres).bind(args)).toEqual([
+        `{"a"}`,
+        [null],
+      ]);
+    });
+
+    /** Nothing of it reaches the SQL, which is the actual defect. */
+    test("it never becomes a bound NULL", () => {
+      let text = "";
+      try {
+        text = pgText({ where: { metadata: { path: ["a"], equals: null } } });
+      } catch {
+        // expected
+      }
+      expect(text).toBe("");
+    });
+  });
+
+  /**
    * `[]` on Postgres extracts the whole document — `[].every` is vacuously
    * true, so it passed the grammar check — and `""` on SQLite raises inside
    * `json_extract` at execution time. Neither is a path.
@@ -1377,6 +1498,49 @@ describe("json path filters", () => {
       "{\"n\"}",
       2,
     ]);
+  });
+
+  /**
+   * #371, the third of three, and the one that was a *choice* rather than a
+   * defect. `assertPathShape` accepted `typeof part === "number"`, where
+   * Prisma's generated `path` on Postgres is `string[]` — so gemi answered a
+   * query the oracle could not express, which is the situation
+   * `compileJsonFilter`'s docblock refuses on purpose for the SQLite operators:
+   * "implementing them would make gemi answer a query the oracle cannot, which
+   * is precisely where a differential test stops being able to check anything."
+   *
+   * Prisma's refusal is a run-time one as well as a type one. Measured on a
+   * generated 6.19.2 client: `path: ["items", 0]` answers *"Argument `path`:
+   * Invalid value provided. Expected String, provided Int."* before any SQL is
+   * built.
+   *
+   * **Nothing is lost by narrowing**, which is what makes it the cheap answer:
+   * `#>` takes a `text[]`, so `["items", "0"]` reaches the same array element.
+   * Both spellings bind the identical path on Prisma and on gemi.
+   */
+  test("a path segment is a string, matching the oracle", () => {
+    expect(() =>
+      pgText({ where: { metadata: { path: ["items", 0], equals: "x" } } }),
+    ).toThrow(/path\[1\] is a number/);
+    expect(() =>
+      pgText({ where: { metadata: { path: ["a", null], equals: "x" } } }),
+    ).toThrow(/path\[1\] is null/);
+
+    // The string spelling reaches the element, and the bound path is identical
+    // to what the numeric one used to produce — which is why this is a
+    // narrowing rather than a removal.
+    const args = { where: { metadata: { path: ["items", "0"], equals: "x" } } };
+    expect(pgText(args)).toContain(`("metadata" #>> $1) = $2`);
+    expect(compileRead(jsonUser, "findMany", args, postgres).bind(args)).toEqual([
+      `{"items","0"}`,
+      "x",
+    ]);
+
+    // SQLite is untouched: its grammar is a JSONPath string, where an index is
+    // spelled inside the string and never reaches this check.
+    expect(() =>
+      sqliteText({ where: { metadata: { path: "$.items[0]", equals: "x" } } }),
+    ).not.toThrow();
   });
 
   test("two filters on one path are ANDed", () => {
--- a/packages/gemi/orm/compile/where.ts
+++ b/packages/gemi/orm/compile/where.ts
@@ -974,8 +974,23 @@ function compileJsonFilter(
   assertPathShape(schema, field, filter.path, context);
 
   const path: Binder = (args) => locate(args)?.path;
+  // **An `undefined` operand is an absent one**, which is the rule the scalar
+  // operator loop already follows (`if (operand === undefined …) continue`) and
+  // the rule the dispatch above follows for `path` itself
+  // (`filter.path !== undefined`). Building `applied` from `Object.keys` alone
+  // made this branch the one place that broke it: a filter assembled from a
+  // form or a query string carries `equals: undefined` for the field nobody
+  // filled in, and that compiled to `= NULL` — a predicate no row can satisfy —
+  // where the same query with the key simply omitted got the bare-path refusal
+  // below.
+  //
+  // Prisma answers both spellings identically. Measured on a generated 6.19.2
+  // client against Postgres: `{ path: ["a"], equals: undefined }` and
+  // `{ path: ["a"] }` both raise P2019 *"A JSON path cannot be set without a
+  // scalar filter."*, and `{ path: ["a"], equals: 1, gt: undefined }` compiles
+  // the `equals` alone.
   const applied = Object.keys(filter)
-    .filter((key) => key !== "path")
+    .filter((key) => key !== "path" && filter[key] !== undefined)
     .sort();
 
   if (applied.length === 0) {
@@ -1078,8 +1093,8 @@ function compileJsonFilter(
 /**
  * What each JSON filter will accept as its operand.
  *
- * Two defects, both of them the shape the scalar path is already careful about
- * and both silent:
+ * Three defects, all of them the shape the scalar path is already careful about
+ * and all silent:
  *
  *   - **The string filters had no non-string guard.** Their scalar siblings
  *     refuse one, with a comment explaining that `String(null)` makes the
@@ -1092,14 +1107,19 @@ function compileJsonFilter(
  *     binds `String(raw)` because `#>>` yields text, so `equals: { b: 1 }` bound
  *     the string `"[object Object]"` and matched nothing.
  *
- * The second is a **refusal rather than an implementation**, deliberately.
+ *   - **`null` fell between the two.** The string branch refuses a non-string
+ *     and the object branch tests `operand !== null` first, so `equals: null`
+ *     reached the binder and compiled to `= NULL` — see the guard below.
+ *
+ * The last two are a **refusal rather than an implementation**, deliberately.
  * Prisma does accept an object or an array there, and answering it properly
  * means the `#>` + `::jsonb` form `array_contains` already uses — which is
  * Postgres-only, so implementing it would give SQLite either a second wrong
  * answer or a silent divergence. Refusing names the limitation on both
  * dialects, which is what this file does everywhere else it cannot answer.
- * `array_contains` is exempt because containment is precisely the operator that
- * takes a document.
+ * `array_contains` is exempt from the object rule because containment is
+ * precisely the operator that takes a document — but not from the `null` one,
+ * because a bound SQL NULL is not a JSON `null` on either side of `@>`.
  */
 function assertJsonOperand(
   key: string,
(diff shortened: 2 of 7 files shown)


nstfkc (commented):
**Verdict: minor-comments.** The three guards are right, the mechanism is right, and the parity claim that reshaped the fix — that Prisma *does* answer `equals: null` at a path — is the kind of measurement this repo asks for. All three new test blocks are genuinely mutation-sensitive; I reverted each guard and confirmed the counts. Nothing here is blocking. What I found is four places where the *reasons* the change gives are not the reasons the code has, or are not pinned by anything.

## What I ran

Worktree at `47b3c1d`, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → the worktree's `packages/gemi` (verified, not the main checkout).

| | |
| --- | --- |
| `packages/gemi` full suite (`bun --bun vitest run`) | 3123 pass, 1 skipped, 120 files |
| `packages/gemi` typecheck | clean |
| template, SQLite (`TZ=UTC bun --bun vitest run app/models`) | 792 pass, 20 files, 3 skipped |
| template type tests (`--typecheck.only`) | 117 pass, no type errors |
| `gh pr checks 380` | `postgres` pass, `sqlite` pass |

**Mutation checks, all three reproduced.** Reverting `filter[key] !== undefined` → 1 failure (`an undefined operand is absent…`). Restoring the old `assertJsonOperand` — null guard removed *and* `operand !== null &&` put back — → 8 failures, all inside `null at a path is refused`. Re-admitting `typeof part === "number"` in the `findIndex` predicate → 1 failure (`a path segment is a string…`). The counts match the PR body exactly.

**What I could not check.** The two new differential cases are Postgres-only and I did not stand up a container; CI's `postgres` job is green and the PR body reports them running rather than skipping, so I took that on the author's evidence. I did not re-derive the Prisma 6.19.2 oracle — the seed, the SQL and the returned rows are written down precisely enough to be re-run, and the raw-psql half (`("metadata" #>> '{a}') = NULL` → 0 rows vs `("metadata" #> '{a}')::jsonb = 'null'` → the JSON-null row) is checkable from the text alone.

## One thing the PR does not claim, and could

The `undefined` fix also closes a **plan-cache shape collision**, which is a stronger argument for it than the one given. Measured:

```
planKey(postgres, "User", "findMany", { where: { metadata: { path: ["a"] } } })
planKey(postgres, "User", "findMany", { where: { metadata: { path: ["a"], equals: undefined } } })
  both -> postgres:batched:User:findMany:{where:{metadata:{path:[string]}}}
```

`canonicalShape` drops `undefined` members, so the two spellings were already **one cache entry** while they compiled to two different things — a refusal and a `= NULL` plan. Whichever compiled first decided for the other, so a bare `path` arriving second was served the never-true plan instead of the refusal it raises when compiled cold. After this change both throw, so there is nothing left to collide; that is worth a line in the `applied` comment, because it is the reason the cache is safe here and the next reader will otherwise have to re-derive it.

## Findings

Four inline, all **should fix** or **nit**, none blocking. Two more that have no single line:

- **should fix (PR body).** The body says of `{ path: undefined, equals: { a: 1 } }`: *"The last row was already right here — the dispatch tests `!== undefined` — and is asserted so it stays right."* It is not asserted anywhere: `grep -rn "path: undefined" --include="*.ts" packages templates` returns nothing, before or after this change. Either add the assertion (one line in the new `undefined` block) or drop the sentence — it is the only claim in the body that the tree does not back.
- **nit.** `it never becomes a bound NULL` swallows every exception and asserts `text === ""`. It passes for any throw, including one from an unrelated future guard. Asserting the message alongside would make it say what it means.

Scope is clean otherwise: every file changed is #371's, the `docs/llms-full.txt` re-embed is forced by `docs-scope.test.ts`, and the one step past the issue's letter — refusing `array_contains: null` — is disclosed in the body and argued rather than slipped in. I agree with taking it; the alternative was leaving one known silent-zero-rows answer inside the function the PR is about.

packages/gemi/orm/compile/where.ts
```diff
@@ -1125,11 +1145,54 @@ function assertJsonOperand(
...
+  // Answering any of them properly means the `#> … ::jsonb` form on Postgres
+  // and `json_type` on SQLite, which is a second implementation of the same
+  // operator per dialect rather than a guard — the same reason the object
+  // operand below is refused instead of compiled.
+  if (operand === null) {
+    throw new InvalidArgumentError(
+      `where.${field.name}.${key}`,
+      schema.name,
+      context.operation,
+      `null cannot be compared through a JSON path: the extraction yields ` +
```
nstfkc: **should fix** — this message gives `array_contains` a mechanism it does not have, and the comment fourteen lines above says so.

Measured on this branch:

```
{ path: ["a"], array_contains: null }
InvalidArgumentError | Invalid 'where.metadata.array_contains' (User.findMany).
  null cannot be compared through a JSON path: the extraction yields SQL NULL for
  an absent key and for a JSON null alike, and the comparison against it is NULL
  rather than true … Filter the column itself — { metadata: { equals: Prisma.JsonNull } }
  — or test the path against the value you mean.
```

But `array_contains` does not extract with `#>>`. It compiles `("metadata" #> $1) @> $2::jsonb` — the branch's own new test pins that string — so there is no text collapse, and the comment above this guard states it outright: *"It takes the `#>` form, so the collapse argument does not apply to it — but `jsonArrayContains` binds `null` as SQL NULL, and `x @> NULL` is NULL too."* That second clause is the true reason and it is the one the caller does not get.

The suggested alternative misses for the same reason: `{ metadata: { equals: Prisma.JsonNull } }` is a containment question answered with an equality on the column. A caller who wrote `array_contains: null` wants "does this array hold a JSON null", which is a different query from either.

This is the #374 shape — a comment (here, an error message) that names a mechanism other than the one at that line. Cheapest fix is to branch the sentence on `key === "array_contains"`: `@> NULL` is NULL because the operand binds as SQL NULL rather than as the JSON value, and Prisma runs a real containment test there.

packages/gemi/orm/compile/where.ts
```diff
@@ -1246,9 +1309,41 @@ function assertPathShape(
...
+    );
+    if (offending === -1) return;
+
+    const part = (path as unknown[])[offending];
+    throw new UnsupportedQueryError(
+      `${field.name}.path`,
+      schema.name,
+      context.operation,
+      `A JSON path is an array of strings, and path[${offending}] is ` +
+        `${part === null ? "null" : `a ${typeof part}`}. An array index is a ` +
```
nstfkc: **nit** — the ternary special-cases `null` only, so every other non-string segment reads ungrammatically. Measured:

```
path: ["a", undefined]   ->  A JSON path is an array of strings, and path[1] is a undefined.
path: ["a", { b: 1 }]    ->  A JSON path is an array of strings, and path[1] is a object.
```

`a undefined` / `a object`. The `number` case — the one the test pins — is the only one that reads.

Second, smaller: the sentence that follows is emitted unconditionally, so a caller who wrote `["a", null]` is told *"An array index is a key too — write `["items", "0"]` rather than `["items", 0]`"*, which is advice for a mistake they did not make. The index sentence is the whole point of the guard for a number and noise for anything else; gating it on `typeof part === "number"` keeps it where it earns its place.

packages/gemi/orm/compile/read.test.ts
```diff
@@ -1351,6 +1397,81 @@ describe("json path filters", () => {
...
+    const operators = [
+      "equals",
+      "not",
+      "string_contains",
+      "array_contains",
+      "gt",
+      "lte",
+    ] as const;
+
+    test.each(operators)("%s on postgres", (key) => {
```
nstfkc: **should fix** — `toThrow(InvalidArgumentError)` does not pin the thing the code and the PR body both call load-bearing: that the new `null` guard sits **after** the `string_*` branch so `string_contains`'s better message survives.

`string_contains: null` already threw `InvalidArgumentError` before this PR, from the non-string guard, with the `'%null%'` reasoning. So this row passes either way, and the placement is unpinned. Measured — I moved the whole `if (operand === null)` block above `if (stringly && typeof operand !== "string")` and ran the file:

```
Tests  153 passed (153)
```

Nothing failed, and `string_contains: null` silently swapped its *"the pattern becomes '%null%', which runs and returns the wrong rows"* message for the generic path-null one. That is a real regression the suite would not have caught.

Asserting the message rather than the class fixes both halves at once — the `string_*` rows keep `/%null%/` (or `/Expected a string/`) and the rest get `/null cannot be compared through a JSON path/`. It also turns the four rows that currently only assert a class into rows that assert the new guard actually fired.

(For the record, the block's mutation count is right: restoring the old `assertJsonOperand` in full gives 8 failures here, exactly as the session file says, and `string_contains on postgres` / `string_ends_with on sqlite` are the two that survive — which is the same fact from the other side.)

templates/saas-starter/app/models/filters.test-d.ts
```diff
@@ -295,17 +309,25 @@ describe("Json path filters", () => {
...
+   * The refusal is a divergence rather than a gap in the type: Prisma extracts
+   * with `#>` and compares as `jsonb`, so it reads this as the JSON value
+   * `null` and answers it. gemi's `#>>` yields SQL NULL for an absent key and a
+   * JSON null alike, which is the same collapse that keeps the sentinels off
+   * this filter.
    */
   test("null at a path is refused, having no reading left", async () => {
     // @ts-expect-error `= NULL` is never true; say which empty you mean, on the column
     await UserModel.findMany({ where: { metadata: { path: ["a"], equals: null } } });
+    // @ts-expect-error `x @> NULL` is NULL too — a null *inside* the document is fine
```
nstfkc, line 328: **should fix** — the test name on line 325 is now the opposite of the docblock directly above it.

`test("null at a path is refused, having no reading left", …)` came from #368, when the belief was that once the sentinels are refused nothing is left for `null` to mean. This PR's headline finding is that **that was wrong** — Prisma extracts with `#>`, compares as `jsonb`, and returns the JSON-null rows — and the docblock four lines up now says so: *"The refusal is a divergence rather than a gap in the type."* The name still asserts the disproved claim, and a test name is the line most likely to be read without the comment.

Something like `"null at a path is refused, because gemi cannot ask Prisma's question"` keeps it honest. Same defect class as #374 in this batch: the reason at the line no longer matches the line.

(The two new cases underneath are good — `array_contains: null` refused and `array_contains: [null]` still valid is exactly the pair that keeps the guard from over-reaching.)

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs. #380 is
involved in **two** cross-PR conflicts. Neither is a defect in this PR; both are things that break
when it meets another one.

---

## 1. #380 × #385 — a numeric path segment: this PR refuses it, #385 ships four

No textual conflict — the two touch disjoint files in `packages/gemi/orm` (#380: `compile/where.ts`,
`types.ts`; #385: `corpus.ts`, `plan.ts`, `plan-key.invariants.test.ts`). They merge silently and
the merged tree fails. Not in `orm.md`'s notes, and neither per-PR review could have seen it.

This PR's `assertPathShape` refuses a non-string segment. #385 ships `["a", 0]` in three places.
Measured on `main` + #380 + #385 (`packages/gemi` suite, **2 failed**):

```
FAIL  orm/plan-key.invariants.test.ts > a JSON path's depth is bound, not compiled in — postgres
UnsupportedQueryError: … A JSON path is an array of strings, and path[1] is a number.
  ❯ assertPathShape orm/compile/where.ts:1365
```

plus #385's corpus-wide scan reporting the same shape as an unexpected refusal under `findMany`
and `findFirst`. The three sites are `corpus.ts`, `plan-key.invariants.test.ts`, and a new `plan.ts`
docblock in #385 whose text names `["a", 0]` as one of four measured cache entries — a sentence
that stops being true.

Respelling to `["a", "0"]` costs nothing (`#>` takes a `text[]`, which is this PR's own argument)
and takes the package suite to **3185 passed**. The repair belongs in #385, so **merge #380 first**.

**Worth flagging beyond the mechanical fix:** #385 carries a blocking finding that its `LIST_KEYS`
collapse makes `assertPathShape` unreachable on a warm plan. This PR *adds* a refusal to
`assertPathShape`, so it strictly widens that hole — `["a", 0]` becomes a shape refused cold and
bound warm off the `["a","b"]` plan. Full detail on #385.

---

## 2. #380 × #381 — resolving `JsonPathFilter` the obvious way drops this PR's `array_contains` narrowing

Both rewrite the same declaration in `types.ts`. This PR keeps the object literal and narrows
`array_contains?: Exclude<JsonValue, null>`; #381 replaces the literal wholesale with a mapped type
over the compiler's tuple, in which `JsonPathOperand<"array_contains"> = JsonValue` — `null`
included.

`git merge` conflicts, so it is visible. The hazard is the resolution: #381's whole purpose is that
the mapped type replaces the literal, so taking #381's side reads as correct and silently reverts
this PR's narrowing. I resolved it that way and measured — `typecheck` clean, package suite
**3123 passed**, both green — with the loss caught only by the template's type tests:

```
FAIL  app/models/filters.test-d.ts > null at a path is refused, having no reading left
TypeCheckError: Unused '@ts-expect-error' directive.   (filters.test-d.ts:328, array_contains: null)
```

That is this PR's own new assertion. CI does run `test:types` (`ci.yml` line 215), so it is loud.

**The resolution that keeps both** is to take #381's mapped type and put the narrowing in the arm —
`K extends "array_contains" ? Exclude<JsonValue, null> : …` — which gives **117 passed, no type
errors**. This PR's property-level docblock on `array_contains` has nowhere to live under a mapped
type and should move onto that arm.

Your `JsonPath` narrowing to `readonly string[]` **does** survive #381 untouched (it keeps
`path: JsonPath` and does not redefine the alias) — confirmed by the numeric-segment
`@ts-expect-error` still firing. `compile/where.ts` auto-merges cleanly with #381's
`JSON_FILTERS` → `JSON_FILTER_NAMES` change, with no interaction I could find.

**Cheapest order: #380 before #381.**

---

Full ten-PR integration branch built locally with both repairs applied: `typecheck` clean,
`packages/gemi` **3185 passed / 1 skipped**, template sqlite **845 passed**, `test:types`
**117 passed**, live Postgres via `postgres.sh` **1247 passed** (the only 2 failures being the
harness's designed one-dialect-at-a-time refusal).

packages/gemi/orm/compile/where.ts
nstfkc: **Fixed** in `ff53a77` — you are right, and the comment above the guard was the tell.

The message now branches on the key:

```
{ path: ["a"], array_contains: null }
InvalidArgumentError | Invalid `where.metadata.array_contains` (User.findMany).
  null cannot be tested for containment through a JSON path: the operand binds
  as SQL NULL rather than as the JSON value, and 'x @> NULL' is NULL rather than
  true — the query would run and match nothing. Prisma binds it as the JSON
  value and runs a real containment test; gemi does not yet. A null *inside*
  the document is an ordinary value and still compiles:
  { metadata: { path: […], array_contains: [null] } }.
```

Both halves of your point are taken. The `#>>` collapse is gone from that arm, and the suggested alternative is gone with it: `{ equals: Prisma.JsonNull }` answers a containment question with an equality on the column, so the caller is pointed at `array_contains: [null]` — which is the query they can actually write, and which this PR already pins as still valid.

The comment above the guard now says *why* it branches rather than leaving the reader to notice that it should, and the `array_contains on postgres` test row asserts `/null cannot be tested for containment/` rather than the class. Mutation: collapsing the branch back to one message fails exactly that row.

packages/gemi/orm/compile/where.ts
nstfkc: **Fixed** in `ff53a77`, both halves.

The description is spelled out rather than interpolated from `typeof`:

```
path: ["a", null]        ->  … path[1] is null. Prisma's path is string[] too.
path: ["a", undefined]   ->  … path[1] is undefined. Prisma's path is string[] too.
path: ["a", { b: 1 }]    ->  … path[1] is an object. Prisma's path is string[] too.
path: ["a", ["b"]]       ->  … path[1] is an array. Prisma's path is string[] too.
path: ["items", 0]       ->  … path[1] is a number. An array index is a key too — write
                             ["items", "0"] rather than ["items", 0]; it reaches the same
                             element, because postgres takes the path as text. Prisma's
                             path is string[] and its client refuses a number here too.
```

`Array.isArray` before the `typeof === "object"` arm, because an array segment is a distinct enough mistake to name.

Your second point is the one I would have missed: the index sentence is gated on `typeof part === "number"` now, so a caller who wrote `null` is not told how to spell an index. The non-number arm keeps only the sentence that is true for all of them — and deliberately does *not* claim Prisma refuses each of these at run time, since the number is the only one I actually measured; `path: string[]` is a type fact I can state without an oracle.

Pinned in `a path segment is a string, matching the oracle`: the four descriptions, the index sentence present for the number, and `not.toThrow(/An array index is a key too/)` for `null`. Mutation — restoring the old interpolation and the ungated sentence — fails that test.

packages/gemi/orm/compile/read.test.ts
nstfkc: **Fixed** in `ff53a77`. This was the best finding of the round — the placement is the thing the PR body calls load-bearing and it was pinned by nothing, exactly as you measured.

Both `test.each` tables now carry the expected message beside the key, and each row asserts the class *and* the message:

| key | asserted |
| --- | --- |
| `equals`, `not`, `gt`, `lte` (pg) | `/null cannot be compared through a JSON path/` |
| `string_contains` (pg) | `/'%null%', which runs and returns the wrong rows/` |
| `array_contains` (pg) | `/null cannot be tested for containment/` |
| `equals`, `not` (sqlite) | `/null cannot be compared through a JSON path/` |
| `string_ends_with` (sqlite) | `/'%null%', which runs and returns the wrong rows/` |

I re-ran your mutation. Moving the whole `if (operand === null)` block above `if (stringly && typeof operand !== "string")`:

```
× string_contains on postgres
× string_ends_with on sqlite
Tests  2 failed | 151 passed (153)
```

Where it was 0 before. The `array_contains` row is new too, and comes from the sibling finding on `where.ts` — that operator's refusal names a different mechanism, so asserting the class alone would have hidden that as well.

For the record on the other side of your parenthesis: restoring the old `assertJsonOperand` in full still gives 8 failures here, and the two survivors are still the two `string_*` rows — they now survive because their message is the `%null%` one and it is asserted, rather than because a class matched by accident.

templates/saas-starter/app/models/filters.test-d.ts
nstfkc: **Fixed** in `ff53a77`. Renamed to your wording:

```ts
test("null at a path is refused, because gemi cannot ask Prisma's question", async () => {
```

The old name was inherited from #368 verbatim and I edited the docblock above it without looking up at the line that a reader actually sees in a failure report. That is the #374 shape and you are right to call it that.

The docblock already said the true thing (*"The refusal is a divergence rather than a gap in the type"*), so nothing else in the block needed changing. `test:types` still **117 passed, no type errors**.

Agreed on the pair underneath — `array_contains: null` refused and `array_contains: [null]` valid is the negative that keeps the guard honest, and it is the one that catches the #381 resolution hazard as an unused `@ts-expect-error`.

nstfkc:
## Review round 1 — every point, and where it went

`ff53a77`. Six addressed, two deferred, none declined. Inline replies are on the four line comments; this covers the two findings that had no single line, the plan-cache observation, and both cross-PR items.

### The two findings without a line

**should fix — the body claimed `{ path: undefined, equals: { a: 1 } }` was asserted, and `grep` said otherwise.** Correct, and worse than it looked. I added the assertion rather than dropping the sentence, and the assertion showed the SQL quoted in the body was also wrong:

```
{ path: undefined, equals: { a: 1 } }
  body claimed   "metadata"::jsonb = $1
  actually       "metadata" = $1::text::jsonb     bind [{"a":1}]
```

Now in the `undefined` block, with the bind, plus a `not.toContain("#>>")` so it says which arm it took. The body is corrected. And it is mutation-sensitive in its own right — changing the dispatch's `filter.path !== undefined` to `"path" in filter` fails it, where before this round nothing did.

**nit — `it never becomes a bound NULL` passed for any throw.** Fixed: it captures the message and asserts `/null cannot be compared through a JSON path/` beside `text === ""`. It was saying "no SQL was built" where it meant "this guard built no SQL".

### The plan-cache collision

Taken, and thank you for measuring it — it is a better argument for the `undefined` guard than the form-and-query-string one I gave, because it is a wrong answer to a *correctly spelled* query. It is now a paragraph in the `applied` comment carrying your two `planKey` calls verbatim, the shared key they produce, and the consequence: whichever spelling compiled first decided for the other, so a bare `path` arriving second was served the never-true plan instead of the refusal it raises cold. Both throw now, so the shared key has nothing left to collide on.

I did not add a test for it. `canonicalShape` dropping `undefined` is not this PR's behaviour and is pinned where it lives; what this PR changes is that both spellings now throw, which the `undefined` block already asserts from both directions.

### Cross-PR — deferred, both

**#380 × #385, the numeric segment.** Deferred to **#385**, which owns all three sites (`corpus.ts`, `plan-key.invariants.test.ts`, and the new `plan.ts` docblock). Respelling `["a", 0]` → `["a", "0"]` costs nothing — `#>` takes a `text[]`, which is this PR's own argument for the narrowing — but the change belongs in #385's branch and I am not reaching into it. Agreed on the order: **#380 before #385**.

On the second half of that note — that this PR widens #385's `LIST_KEYS` warm-key hole — I agree it is true and it is #385's finding to resolve. It is worth saying that the widening is of the *bypass*, not of the refusal: `assertPathShape` is where a path shape can be checked at all, so a fix that makes it reachable on a warm plan repairs this shape along with the ones already there.

**#380 × #381, `JsonPathFilter`.** Deferred to **merge time / #381**. The resolution you name is right — take #381's mapped type and put the narrowing in the arm, `K extends "array_contains" ? Exclude<JsonValue, null> : …`, with this PR's property docblock moving onto it — but it cannot be written here: #381 is not on `main`, so there is no mapped type to narrow. What I can do from this side is make sure the loss stays loud, and it does: `array_contains: null` in `filters.test-d.ts` is the `@ts-expect-error` that goes unused if the narrowing is reverted, and `ci.yml` line 215 runs `test:types`. Recorded in `.orm-sessions/issue-371.md` under `## Review round 1` so whoever resolves the conflict has the arm spelled out.

**#386's `childField` → `childFields` rename** touches nothing in this PR — no reference to either name in the diff — so there is nothing here to absorb.

### Verification after the round

| | |
| --- | --- |
| `packages/gemi` (`bun --bun vitest run`) | **3123 pass, 1 skipped**, 120 files |
| `packages/gemi` typecheck | clean |
| `oxlint orm --deny-warnings` | 0 warnings, 0 errors |
| template SQLite (`TZ=UTC bun --bun vitest run app/models`) | **792 pass**, 20 files |
| template `test:types` | **117 pass**, no type errors |
| `git status` after both `prisma generate` runs | clean — no generated drift |

Mutation checks, re-run on every test touched. Four of these were 0 before this round:

| mutation | failures |
| --- | --- |
| `applied` filter back to `key !== "path"` | 1 |
| dispatch `filter.path !== undefined` → `"path" in filter` | **1 (new)** |
| `assertJsonOperand` restored in full (null guard out, `operand !== null &&` back) | 8 |
| `null` guard moved above the `string_*` guard | **2 (new — was 0)** |
| `array_contains` message branch collapsed to one message | **1 (new)** |
| segment description + index-sentence gating reverted | **1 (new)** |
| `findIndex` predicate re-admits `typeof part === "number"` | 1 |

I did not stand up Postgres again this round. Nothing here changes a compiled query — the diff is three error messages, one comment and the tests — and the differential suite compares result sets, not messages. The two new differential cases are untouched; CI's `postgres` job covers them on this push.

Worktree at `ff53a77`, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → the worktree's `packages/gemi`.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `ff53a770` for
this PR.

### #380 × #381 narrowing hazard — **resolved**, by #381 rather than by the resolution

The previous pass warned that the *natural* resolution of the `types.ts` conflict — take #381's
mapped type, which is its whole point — silently reverts your `array_contains?: Exclude<JsonValue, null>`.
That is no longer possible to get wrong: #381's round moved the narrowing onto the arm of the mapped
type it introduces,

```ts
type JsonPathOperand<K extends JsonFilterName> = K extends "array_contains"
  ? Exclude<JsonValue, null>
  : …
```

so taking #381's side — which is what I did, and it is still the only sensible resolution — keeps
the property instead of dropping it.

Verified rather than assumed, because a green suite was the whole problem last time. On the merged
tree I reverted that arm to plain `JsonValue`: `packages/gemi` stays green (as before), and
`test:types` goes red with `Unused '@ts-expect-error' directive` at `filters.test-d.ts:372`.
Restored, 118 pass, no type errors. So the tripwire is live on the integrated tree and CI runs it
(`ci.yml:215`).

The `filters.test-d.ts` conflict predicted by #381 also materialised — add/add inside
`test("null at a path is refused, …")`, both sides adding an `array_contains: null`
`@ts-expect-error` with different wording. Keep either one line, as #381 said; I kept its side,
which also carries the "a null *inside* the document stays expressible" companion assertion.

### New, and it is yours × #385: your segment message breaks #385's refusal list

Disjoint files, silent merge, one test failure — the same class as the original #380 × #385, and
new since the last pass.

```
FAIL orm/plan-key.invariants.test.ts > every corpus entry compiles, or is refused by name — postgres
  findMany {"where":{"metadata":{"path":["a",null],"equals":"x"}}} —
  A JSON path is an array of strings, and path[1] is null. Prisma's path is string[] too.
```

#385's round added `{ path: ["a", null], equals: "x" }` to `corpus.ts`, and reported it needed no
`EXPECTED_REFUSALS` change. That was correct **on its own branch**: without your change a malformed
segment falls through to the grammar throw and is refused with *"a JSON path is an array of keys"*,
which is already listed. Your `assertPathShape` gives the segment case a sentence of its own, so the
entry matches nothing.

Nothing to change here — the fix belongs in #385's list, and I have posted the exact pattern there
(`/A JSON path is an array of strings, and path\[\d+\] is null/`, Postgres-only, deliberately not
the broader `/A JSON path is an array of strings/`, which would also excuse a numeric segment and
destroy the tripwire #385 built around `["a", "0"]`). Flagged here so the interaction is on record
against the change that causes it.

The **numeric-segment** half of the original conflict is fully resolved: #385 respelled all three
sites to `["a", "0"]`, and reverting one to `["a", 0]` fails this same invariant with your numeric
message — measured.

Closing keyword verified: `closingIssuesReferences` resolves to #371.

PR #381, opened 2026-08-09
Close #369: the caller-facing types read the compiler's operator lists

### LINKED ISSUE #369: The operator sets are encoded twice, once in the compiler and once in the types, and nothing forces them to agree
#326, #333, #337 and #336 are four instances of one mechanism, and the fifth is already possible today. Worth closing the class rather than fixing them one at a time.

## The mechanism

`orm/types.ts` and `orm/compile/*.ts` each **independently encode the same operator set**, and nothing forces them to agree.

`JSON_FILTERS` (`orm/compile/where.ts:918-930`) is a runtime `Set<string>`. `JsonPathFilter` (after #368) is a hand-written type listing the same ten names. Change one and nothing breaks — there is no compile error, no failing test, and no diff a reviewer would flag, because the two lists live in different files and neither imports the other.

Every one of the four issues is the same shape, and always in the same direction: **the compiler accepts something the type refuses to describe**, so a documented, tested, working feature has no caller-facing spelling. In #336's case the result was worse than a missing type — the filter was silently absorbed as an `equals` shorthand and interpreted correctly by accident, so it worked for anyone who spelled it right and failed silently for anyone who didn't.

## The fix

Make the runtime sets the single source and derive the types from them: declare them `as const` arrays and map over the tuple (`keyof` / mapped types) so the compiler's own list *is* the type's list. Sites that carry a duplicated list today:

- `JSON_FILTERS` — `orm/compile/where.ts:918-930`
- `HAVING_OPERATORS`
- `readOperators`
- the nested-write operand sets (`WRITE_ARGS` and friends)

That converts "the type didn't describe the compiler" from something a reviewer has to notice into a compile error. It is a mechanical change per site, not a redesign.

## What it does not cover

The dialect-set halves (`dialect.jsonFilters`) cannot be derived this way — the generated artifact is dialect-agnostic (`ModelTypeInfo` records no dialect; `DATABASE_URL` picks it at connect time), so there is nothing for a dialect-shaped type to read. Those stay the runtime's job, refused at runtime with a message naming the dialect, which is the existing precedent (`mode?: "default" | "insensitive"`, `ListFilter` / `assertListDialect`).

## Why now

#368 adds a tenth hand-maintained copy of the `JSON_FILTERS` list. It is correct today. Nothing keeps it correct.

nstfkc:
Closes #369.

`orm/types.ts` and `orm/compile/*.ts` each encoded the same operator sets, in files that neither imports the other. Six grammars now map over the compiler's own tuple, so the compiler's list **is** the type's list.

## The defect, reproduced as a mutation

Not argued from the issue text. On `main`, deleting one line from `JsonPathFilter` — the caller-facing type:

```diff
-  string_ends_with?: string;
```

leaves `packages/gemi` entirely green:

```
tsc --noEmit -p tsconfig.json    (no output)
vitest run orm                   64 files, 2421 tests, all passed
```

while `compileJsonFilter` goes on implementing `string_ends_with`, `docs/orm.md` goes on documenting it, and the differential case `"json path string_ends_with"` goes on asserting it against Prisma. That is #336 exactly, arrived at by hand: a working, tested, documented filter with no caller-facing spelling.

The one thing that catches it is `filters.test-d.ts`, added by #368 — and that file is a *fourth* hand-written copy of the same ten names, which the issue's last paragraph is about. It fails when the type shrinks. Nothing anywhere fails when the compiler's list grows: adding `"string_matches"` to `JSON_FILTERS` on `main` leaves typecheck and all 2421 tests green.

## The change

Each runtime set becomes an `as const` tuple that exports a string-literal union; `types.ts` type-imports the union and maps over it.

| the compiler's list | the type that used to restate it |
| --- | --- |
| `JSON_FILTER_NAMES` — `compile/where.ts` | `JsonPathFilter` |
| `LIST_FILTER_NAMES` — `compile/where.ts` | `ListFilter` |
| `OPERATORS` — `compile/group-by.ts` | `HavingComparison` |
| `MANY_FILTER_OPERATORS` / `ONE_FILTER_OPERATORS` — `relation-filters.ts` | `RelationFilter` |
| `SUPPORTED_STATEMENTS` / `EXISTING_ROW_ONLY_STATEMENTS` / `COLLECTION_ONLY_STATEMENTS` — `compile/nested-writes.ts` | `NestedCreate` / `NestedUpdate` |

Which operator exists is now stated once. What each operator's *operand* is still has to be stated per operator, and that is a smaller and much noisier surface to get wrong — a wrong operand type is a compile error at the call site, where a missing key was silence.

`import type` throughout, so nothing is emitted: `dist/orm/index.js` is byte-unchanged and `runtime-isolation.test.ts`'s property is untouched. The `.d.ts` half was checked rather than assumed — `bun run build:types` emits `dist/orm/compile/where.d.ts`, `group-by.d.ts`, `nested-writes.d.ts` and `relation-filters.d.ts` beside `types.d.ts`, and `types.d.ts` carries the four `import type` lines that resolve to them. There is no import cycle: nothing under `compile/` imports `../types`.

### The nested-write statements were the one non-mechanical site

The other four are one list against one key set. The write side is three key sets against four, and they turn out to fall out of subtraction rather than needing a fourth list:

```
NestedCreate<many>  = SUPPORTED − EXISTING_ROW_ONLY
NestedCreate<one>   = SUPPORTED − EXISTING_ROW_ONLY − COLLECTION_ONLY
NestedUpdate<many>  = SUPPORTED
NestedUpdate<one>   = SUPPORTED − COLLECTION_ONLY
```

*(Corrected in review — the first line said `= SUPPORTED`, which is `NestedUpdate<many>`'s row. The docblock in `types.ts` had it right, and the reviewer's independent mutual-assignability check against `54d4278`'s `types.ts` confirms all four arms, operands included.)*

`COLLECTION_ONLY_STATEMENTS` is new, and it is not a fifth hand-maintained list: both planners already refused `set` / `updateMany` / `deleteMany` on a to-one through an inline `key === … || key === …` chain, and `planForeignSide`'s docblock already named all four together — *"with no `createMany`, `set`, `updateMany` or `deleteMany` key at all"*. Both chains now read `NO_SET_OF_ROWS`, which is the tuple minus `createMany`, derived by `.filter` rather than written out.

`createMany` is carved out rather than folded in because its refusal is **not** the same refusal. It says "there is one related row at most and nothing to create many of" on the owning side and "there is nothing to create many of" on the foreign side, each next to its own operand check, and `nested-writes.ts` says in as many words that the two are duplicated on purpose because "both are reachable and neither subsumes the other". Merging them would have rewritten two pinned messages for tidiness.

`NestedUpdate` also stops being `NestedCreate<R> & (…)`. As an intersection there were two places a statement's operand could be declared, and one declared in both with different operands would intersect to something neither half meant. One mapped type over the unsubtracted set has one.

### A statement with no operand arm says so

`ManyNestedOperand` / `OneNestedOperand` fall through to a named type rather than `never`, following `AggregateNeedsArithmetic`'s precedent in the same file. Measured, by adding `"replace"` to `SUPPORTED_STATEMENTS`:

```
TypeCheckError: Property '"this nested-write statement is in SUPPORTED_STATEMENTS
but has no operand type here"' is missing in type '{}'
```

rather than *"not assignable to type 'undefined'"*.

`JsonPathOperand` deliberately does **not** do this — its default arm is `JsonPathScalar`, which is what five of the current six non-string operators take. A `never` there would give a new operator a key nothing can be passed under, which is a quieter kind of wrong than the one this removes.

## Evidence that the derivation holds

Five throwaway `@ts-expect-error` probes in the template, run through `bun run test:types` — the harness fails an `@ts-expect-error` whose error stops happening, so the directive flips in both directions:

```ts
// @ts-expect-error present only when JSON_FILTER_NAMES carries it
await UserModel.findMany({ where: { metadata: { path: ["a"], string_matches: "x" } } });
// @ts-expect-error present only when group-by OPERATORS carries it
await UserModel.groupBy({ by: ["globalRole"], having: { name: { startsWith: "a" } } });
// @ts-expect-error present only when MANY_FILTER_OPERATORS carries it
await UserModel.findMany({ where: { accounts: { atLeastOne: {} } } });
// @ts-expect-error present only when SUPPORTED_STATEMENTS carries it
await UserModel.update({ where: { id: 1 }, data: { accounts: { replace: undefined } } });
// @ts-expect-error ...and the to-one arm of the same
await AccountModel.update({ where: { id: 1 }, data: { user: { replace: undefined } } });
```

| state | result |
| --- | --- |
| this branch as it stands | 121 passed, no type errors — all five spellings refused |
| one name added to each of the four runtime lists, **`types.ts` untouched** | all five `@ts-expect-error` directives report *"Unused"* — the type followed |

and the other 116 assertions in `filters.test-d.ts` stayed green through both, so the widening was exactly the five keys and nothing else.

The probes are not committed. They demonstrate a property of the derivation, and the derivation is what makes them redundant: there is no second list left for them to guard.

## Runs

Everything CI runs, in the sqlite job's order:

| | |
| --- | --- |
| `tsc --noEmit -p tsconfig.json` | clean |
| `tsc --noEmit -p tsconfig.generated.json` | clean |
| `vitest run` (packages/gemi) | 120 files, 3110 passed, 1 skipped |
| `vitest run app/models` (template) | 20 files, 792 passed, 133 skipped |
| `bun run test:types` (template) | 6 files, 117 passed, no type errors |
| `bunx prisma generate` (template) | zero-line diff on the committed artifact |
| `gemi check models --ignore bench` | 7 model files against 14 registered models |
| `oxlint orm --deny-warnings` | 0 warnings, 0 errors |
| `bun run lint` | 79 warnings, 0 errors (unchanged) |

`bun install` was run in the worktree first and `readlink -f templates/saas-starter/node_modules/gemi` confirmed to resolve into it, so the template suites exercised this branch rather than the main checkout.

**Postgres was not run** — no service container here. The change is dialect-agnostic by construction (no runtime table changed membership, and `dialect.jsonFilters` is untouched), but that is an argument, not a measurement, and the `postgres` job is what will settle it.

**Compile time was measured**, because `types.ts`'s own header warns that eager evaluation would turn a cyclic schema into an instantiation-depth error. These map over fixed tuples of literals, not over `keyof M`, so the key set is finite whatever the schema is. Warm `test:types` over the template's fourteen mutually recursive models: 1.83 / 1.68 / 1.56s on this branch, 1.80 / 1.72s on `main`. The note is now in the header beside the warning it answers.

## What this does not cover, and why

- **The scalar `OPERATORS` set** in `compile/where.ts` is *not* derived, though it is the same shape. Its twelve names split across `BaseFilter` / `OrderingFilter` / `StringFilter`, and which arm a name belongs to is gated on the column's storage class — a fact the runtime set does not record, since the compiler applies one flat table and lets the operand's type decide. Deriving it would mean inventing a three-way partition here and asserting it sums back, which is a new source of truth rather than a removed one. Left with its comment.
- **The dialect halves** (`dialect.jsonFilters`, `dialect.listFilters`) stay the runtime's job, for the reason the issue gives: the generated artifact is dialect-agnostic, so a dialect-shaped type has nothing to read. Unchanged.
- **`READ_ARGS` / `WRITE_ARGS`** are operation *argument* names rather than operators, and they already have two cross-checks that read them directly — `generated-surface.test.ts` and `plan.coverage.test.ts`. Adding a third mechanism over the same pair is not what the issue asks for.
- ~~**`ListFilter` has no type-level regression test**, and cannot get one here.~~ **Withdrawn in review — it does now.** That was true of asking through a *model*: a `String[]` column is a validation error on SQLite, so scalar lists live in `postgres-only.prisma`, whose artifacts are uncommitted. It was not true of the type. `ListFilter` is reached from the public `FieldFilter` on the element type alone, so `Exclude<FieldFilter<string[]>, string[]>` is the arm, with no column, no Postgres and no generated artifact. `filters.test-d.ts` now pins it twice — the key set against `LIST_FILTER_NAMES`, and `ListFilterOperand`'s three-way operand split — and each mutation fails exactly one. The `Boolean` / `Bytes` gap the file records is unrelated and still real.
- **`NestedUpdate<one>` still offers `upsert` and `delete` on an owning-side to-one, which `planOwningSide` refuses.** Pre-existing and untouched: `RelationInfo` records `kind` but not which side holds the foreign key, so the type cannot tell an owning to-one from a foreign one. Worth its own issue rather than a guess here.

## Review round 1

Six inline comments and one cross-PR comment; five addressed, one was a verification with no change requested, and nothing declined or deferred. Commit `02b8b66`.

- **The derivation had no tripwire.** Reverting `types.ts` to `54d4278` left every derived union an export nothing imports, and the whole package stayed green — including `architecture.test.ts`, whose stated property is exactly that. Its scan matched `export function`, `export const` and `export class` and nothing else. It now matches `export type` and `export interface`, and **skips itself as a source**: its own new docblock names three of the unions, and that alone was enough to keep those three green under the revert while the other five went red. A comment is not a use. Reverting `types.ts` now names all eight. The six pre-existing type exports the wider scan turns up are allow-listed with a reason each — all six name a parameter or field of a public signature, which is a use this scan cannot see.
- **`ListFilter` is asserted** — see the struck-through bullet above.
- **`array_contains` narrows to `Exclude<JsonValue, null>`**, which is #380's finding landed on `JsonPathOperand`'s arm rather than left to a merge to preserve. A bare `null` reaches `dialect.jsonArrayContains` and binds as SQL NULL; `x @> NULL` is NULL, not false. The assertion is in `filters.test-d.ts` beside the existing `equals: null` one, with a companion line pinning that a null *inside* the document is still expressible. Reverting the arm to `JsonValue` reports the directive unused.
- **`OPERATOR_SQL` is `Readonly`**, so the lookup alias can no longer write to the object `HavingOperator` is derived from. The `| undefined` is unchanged and is the half that matters.
- **Two comments corrected.** `relation-filters.ts` claimed the two tuples are exported; they are not, and `architecture.test.ts` records that they were deliberately unexported — what crosses the boundary is the derived type. The `types.ts` header said `[K in <five literal strings>]`, which is the relation-filter count and no other; it now says a fixed tuple of literals, notes the range is two to eleven, and names all six mapped grammars rather than three.
- **The subtraction check needed nothing** beyond the table fix above.

### DIFF
--- a/packages/gemi/orm/compile/group-by.ts
+++ b/packages/gemi/orm/compile/group-by.ts
@@ -475,15 +475,35 @@ function assertPureAggregate(
   );
 }
 
-/** The comparison operators a `having` operand may use. */
-const OPERATORS: Record<string, string> = {
+/**
+ * The comparison operators a `having` operand may use.
+ *
+ * **`as const`, so `HavingComparison` in `../types` can be a mapped type over
+ * its keys rather than a second copy of them.** The type used to list the same
+ * six names in another file, with nothing making the two agree — the same
+ * mechanism as `JSON_FILTER_NAMES` in `./where.ts`, and the one that let #326,
+ * #333, #336 and #337 each ship a compiler and a type that disagreed.
+ *
+ * Narrower than a `where`'s set on purpose: `having` walks this table and
+ * nothing else, so `contains`, `in`, `startsWith` and `mode` are refused here.
+ */
+const OPERATORS = {
   equals: "=",
   not: "<>",
   lt: "<",
   lte: "<=",
   gt: ">",
   gte: ">=",
-};
+} as const;
+
+/** One of {@link OPERATORS}' keys. `HavingComparison` maps over this. */
+export type HavingOperator = keyof typeof OPERATORS;
+
+/**
+ * The same table, widened for lookup: the key comes from a caller's object and
+ * is `string`, not {@link HavingOperator} — that is the question being asked.
+ */
+const OPERATOR_SQL: Record<string, string | undefined> = OPERATORS;
 
 function comparison(
   lhs: string,
@@ -506,7 +526,7 @@ function comparison(
     const value = source[key];
     if (value === undefined) continue;
 
-    const operator = OPERATORS[key];
+    const operator = OPERATOR_SQL[key];
     if (!operator) {
       throw new UnsupportedQueryError(
         `${argument}.${key}`,
--- a/packages/gemi/orm/compile/group-by.ts
+++ b/packages/gemi/orm/compile/group-by.ts
@@ -502,8 +502,14 @@ export type HavingOperator = keyof typeof OPERATORS;
 /**
  * The same table, widened for lookup: the key comes from a caller's object and
  * is `string`, not {@link HavingOperator} — that is the question being asked.
+ *
+ * `Readonly`, because this alias points at the object {@link HavingOperator} is
+ * derived from: without it `OPERATOR_SQL.contains = "like"` would typecheck and
+ * mutate the source of truth. The `| undefined` is the load-bearing half — it
+ * is what makes the `if (!operator)` guard below honest rather than a cast that
+ * claims every string is a key.
  */
-const OPERATOR_SQL: Record<string, string | undefined> = OPERATORS;
+const OPERATOR_SQL: Readonly<Record<string, string | undefined>> = OPERATORS;
 
 function comparison(
   lhs: string,
--- a/packages/gemi/orm/relation-filters.ts
+++ b/packages/gemi/orm/relation-filters.ts
@@ -27,9 +27,14 @@ const MANY_FILTER_OPERATORS = ["every", "none", "some"] as const;
 const ONE_FILTER_OPERATORS = ["is", "isNot"] as const;
 
 /**
- * A **third** consumer of the same two tuples, and the reason they are exported
- * at all: `RelationFilter` in `types.ts` is a mapped type over these, rather
- * than a hand-written copy of the five names.
+ * A **third** consumer of the same two tuples, and the reason the two unions
+ * below exist: `RelationFilter` in `types.ts` is a mapped type over these,
+ * rather than a hand-written copy of the five names.
+ *
+ * The tuples themselves stay module-private — `architecture.test.ts` records
+ * that they were deliberately unexported, being used only by the function
+ * directly beneath them, and that is still true. What crosses the file
+ * boundary is the derived type, which is erased.
  *
  * The paragraph above says two places must agree about this shape and must not
  * import each other. The caller-facing type is the place that was left out of
--- a/packages/gemi/orm/relation-filters.ts
+++ b/packages/gemi/orm/relation-filters.ts
@@ -26,6 +26,21 @@
 const MANY_FILTER_OPERATORS = ["every", "none", "some"] as const;
 const ONE_FILTER_OPERATORS = ["is", "isNot"] as const;
 
+/**
+ * A **third** consumer of the same two tuples, and the reason they are exported
+ * at all: `RelationFilter` in `types.ts` is a mapped type over these, rather
+ * than a hand-written copy of the five names.
+ *
+ * The paragraph above says two places must agree about this shape and must not
+ * import each other. The caller-facing type is the place that was left out of
+ * that arrangement — it spelled `some`/`every`/`none` and `is`/`isNot` again,
+ * in a file that imports nothing from here, so nothing failed when the two
+ * drifted. That is the mechanism #369 exists to close, and the same one behind
+ * #326, #333, #336 and #337: a type free to describe a compiler it never reads.
+ */
+export type ManyRelationOperator = (typeof MANY_FILTER_OPERATORS)[number];
+export type OneRelationOperator = (typeof ONE_FILTER_OPERATORS)[number];
+
 export function relationFilterOperators(
   kind: "one" | "many",
 ): readonly string[] {
--- a/packages/gemi/orm/types.ts
+++ b/packages/gemi/orm/types.ts
@@ -57,12 +57,41 @@
 // eagerly evaluated — mapping over `keyof` at the top level, say, or wrapping one
 // in a conditional that has to resolve the whole shape — would turn a cyclic
 // schema into an instantiation-depth error rather than a slow compile.
+//
+// **The operator grammars below map over a tuple, and that is not the mapping
+// the paragraph above warns about.** `RelationFilter`, `NestedCreate` and
+// `NestedUpdate` are `[K in <five literal strings>]`, not `[K in keyof M]`: the
+// key set is fixed and finite whatever the schema is, and the property types
+// are deferred exactly as an interface's are. Measured rather than assumed —
+// the template's fourteen mutually recursive models typecheck in the same time
+// they did before (1.6–1.8s, warm, either way).
 
 import type {
   AnyNullValue,
   DbNullValue,
   JsonNullValue,
 } from "./json-null";
+// The operator and statement names below are **the compiler's own lists**, not
+// copies of them. Each filter grammar here used to spell its operators a second
+// time, in a file the compiler neither imports nor is imported by, so the two
+// could differ with nothing failing: #326, #333, #336 and #337 are four
+// instances of exactly that, and every one of them ran the same way — the
+// compiler answered a query the type had no spelling for. Mapping over the
+// runtime tuple turns the next such drift into a compile error here.
+//
+// `import type`, so this is erased entirely: no runtime edge is created from
+// the type layer into the compiler, and `orm/index.js` is unchanged.
+import type { HavingOperator } from "./compile/group-by";
+import type {
+  CollectionOnlyStatement,
+  ExistingRowStatement,
+  NestedWriteStatement,
+} from "./compile/nested-writes";
+import type { JsonFilterName, ListFilterName } from "./compile/where";
+import type {
+  ManyRelationOperator,
+  OneRelationOperator,
+} from "./relation-filters";
 
 /**
  * Any value a `Json` column can hold, on either dialect.
@@ -351,18 +380,30 @@ type StringFilter<V> = NonNullable<V> extends string
 type NestedFilter<V> = BaseFilter<V> & OrderingFilter<V> & StringFilter<V>;
 
 /**
+ * What each scalar-list filter compares against.
+ *
+ * `has` asks about one element and `isEmpty` about the array's length; the
+ * other three take a whole array. Split out so {@link ListFilter} can be a
+ * mapped type — the operand still has to be stated per operator, but *which
+ * operators exist* is no longer stated twice.
+ */
+type ListFilterOperand<K extends ListFilterName, E> = K extends "has"
+  ? E
+  : K extends "isEmpty"
+    ? boolean
+    : E[];
+
+/**
  * The filters that apply to a **scalar list** — `tags String[]`.
  *
- * A different set on a different left-hand side, mirroring `LIST_FILTERS` in
- * the where compiler. Postgres only: `SqliteDialect.listFilters` refuses the
- * column outright, because SQLite has no array type.
+ * A different set on a different left-hand side, and the set is
+ * `LIST_FILTER_NAMES` in the where compiler — read from there rather than
+ * mirrored, which is what this comment used to claim it did. Postgres only:
+ * `SqliteDialect.listFilters` refuses the column outright, because SQLite has
+ * no array type.
  */
 type ListFilter<E> = {
-  equals?: E[];
-  has?: E;
-  hasEvery?: E[];
-  hasSome?: E[];
-  isEmpty?: boolean;
+  [K in ListFilterName]?: ListFilterOperand<K, E>;
 };
 
 /**
@@ -493,9 +534,11 @@ type JsonPathScalar = string | number | boolean;
  * `where: { metadata: { path: …, equals: … } }` — a filter on a value *inside*
  * the document rather than on the column.
  *
- * The operator set is `JSON_FILTERS` in `compile/where.ts`, all ten of them, and
- * the operand types are `assertJsonOperand`'s. `array_contains` is the one that
- * takes a document, because containment is the operator that means one.
+ * The operator set **is** `JSON_FILTER_NAMES` in `compile/where.ts`, all ten of
+ * them — mapped over rather than copied, so the compiler's list is this type's
+ * list (#369). The operand types are `assertJsonOperand`'s. `array_contains` is
+ * the one that takes a document, because containment is the operator that means
+ * one.
  *
  * **`path` is required, and that is what makes the union above discriminate.**
  * `compileFieldFilter` dispatches on `filter.path !== undefined`: an operand
@@ -510,19 +553,31 @@ type JsonPathScalar = string | number | boolean;
  * reports a misspelled operator, and the runtime message already names the whole
  * set. Prisma's generated input has the same shape for the same reason.
  */
-type JsonPathFilter = {
-  path: JsonPath;
-  equals?: JsonPathScalar;
-  not?: JsonPathScalar;
-  string_contains?: string;
-  string_starts_with?: string;
-  string_ends_with?: string;
-  array_contains?: JsonValue;
-  lt?: JsonPathScalar;
-  lte?: JsonPathScalar;
-  gt?: JsonPathScalar;
-  gte?: JsonPathScalar;
+type JsonPathFilter = { path: JsonPath } & {
+  [K in JsonFilterName]?: JsonPathOperand<K>;
 };
+
+/**
+ * What each JSON path operator compares the extracted value against, which is
+ * `assertJsonOperand`'s rule restated as a type.
+ *
+ * Three answers for ten operators: the `string_*` three build a `like` pattern
+ * and so need a string; `array_contains` is the one operator whose operand is a
+ * *document*, because containment is the question that means one; the remaining
+ * six compare one bound scalar against one extracted value.
+ *
+ * The default arm is `JsonPathScalar` rather than `never` deliberately. An
+ * eleventh operator added to `JSON_FILTER_NAMES` lands here with the operand
+ * type five of the current six have, which is the likeliest right answer and is
+ * in any case *usable* — where a `never` would give the new operator a key
+ * nothing can be passed under, which is a quieter kind of wrong than the one
+ * this change removes.
+ */
+type JsonPathOperand<K extends JsonFilterName> = K extends "array_contains"
+  ? JsonValue
+  : K extends "string_contains" | "string_starts_with" | "string_ends_with"
+    ? string
+    : JsonPathScalar;
 
 /**
  * The bare-value half of a `Json` column's filter — **`JsonValue` with its
@@ -581,19 +636,25 @@ export type FieldFilter<V> =
       ? E[] | ListFilter<E>
       : V | NestedFilter<V>;
 
+/**
+ * The operator names are `relation-filters.ts`'s two tuples, which is the same
+ * pair `readOperators` iterates and `isOperatorForm` tests against — so the
+ * grammar the caller may write and the grammar the compiler recognises are one
+ * list (#369).
+ *
+ * `relation-filters.ts` already existed to stop the where compiler and the
+ * policy walk disagreeing about this shape; the type was the third party to
+ * that agreement and was not in it.
+ */
 type RelationFilter<R extends RelationInfo> = R["kind"] extends "many"
-  ? {
-      some?: WhereInput<R["target"]>;
-      every?: WhereInput<R["target"]>;
-      none?: WhereInput<R["target"]>;
-    }
+  ? { [K in ManyRelationOperator]?: WhereInput<R["target"]> }
   : /**
      * A to-one takes the nested `where` directly — `{ user: { email } }` means
      * `is` — and takes `null` for "there is no related row". `readOperators` in
      * the where compiler folds the two spellings together.
      */
     | WhereInput<R["target"]>
-      | { is?: WhereInput<R["target"]> | null; isNot?: WhereInput<R["target"]> | null }
+      | { [K in OneRelationOperator]?: WhereInput<R["target"]> | null }
       | (R["nullable"] extends true ? null : never);
 
 export type WhereInput<M extends ModelTypeInfo> = {
@@ -784,24 +845,99 @@ export type FindUniqueOrThrowArgs<M extends ModelTypeInfo> = FindUniqueArgs<M>;
 // ---------------------------------------------------------------------------
 
 /**
- * The nested write statements, from `SUPPORTED` in `compile/nested-writes.ts`.
+ * The statements a nested write may name **under a `create`**, which is
+ * `SUPPORTED` minus `EXISTING_ROW_ONLY` in `compile/nested-writes.ts` —
+ * subtracted rather than listed again, so the two files cannot answer
+ * differently (#369). Concretely: `connect`, `connectOrCreate`, `create`,
+ * `createMany`.
+ *
+ * The subtraction is the compiler's own reasoning, not an approximation of it.
+ * `assertNestedOperand` refuses an `EXISTING_ROW_ONLY` key whenever the
+ * operation is a `create`, because there is nothing linked to a row that does
+ * not exist yet.
+ */
+type NestedCreateStatement = Exclude<NestedWriteStatement, ExistingRowStatement>;
+
+/**
+ * What each statement's operand is on a **to-many**.
+ *
+ * The `| T[]` on nearly every arm is Prisma's own shape and the compiler's:
+ * `listOf` accepts a single object or an array of them everywhere a collection
+ * is being written.
+ *
+ * A statement added to `SUPPORTED_STATEMENTS` with no arm here resolves to
+ * {@link NestedStatementHasNoOperandType}, whose one property is a sentence —
+ * the same device `AggregateNeedsArithmetic` uses two sections down, and for
+ * the same reason: an optional `never` reports *"not assignable to type
+ * 'undefined'"*, which says nothing about what went wrong.
+ */
+type ManyNestedOperand<
+  K extends NestedWriteStatement,
+  M extends ModelTypeInfo,
+> = K extends "create"
+  ? CreateInput<M> | CreateInput<M>[]
+  : K extends "createMany"
+    ? { data: CreateInput<M>[]; skipDuplicates?: boolean }
+    : K extends "connect" | "set" | "disconnect" | "delete"
+      ? WhereUniqueInput<M> | WhereUniqueInput<M>[]
+      : K extends "connectOrCreate"
+        ? ConnectOrCreate<M> | ConnectOrCreate<M>[]
+        : K extends "deleteMany"
+          ? WhereInput<M> | WhereInput<M>[]
+          : K extends "update"
+            ? NestedUpdateOne<M> | NestedUpdateOne<M>[]
+            : K extends "updateMany"
+              ? NestedUpdateMany<M> | NestedUpdateMany<M>[]
+              : K extends "upsert"
+                ? NestedUpsert<M> | NestedUpsert<M>[]
+                : NestedStatementHasNoOperandType;
+
+/**
+ * ...and on a **to-one**, where four statements do not exist at all and none of
+ * the rest takes an array — there is one related row, so there is nothing for a
+ * list of operands to address, and `planForeignSide` refuses one by name.
+ *
+ * Three differ in more than that. `disconnect` and `delete` take a boolean —
+ * "the link, whatever it points at" — where a to-many has to say *which* rows;
+ * and `update` takes the bare `data` as well as `{ where?, data }`, because the
+ * parent's key already names the row.
+ */
+type OneNestedOperand<
+  K extends Exclude<NestedWriteStatement, CollectionOnlyStatement>,
+  M extends ModelTypeInfo,
+> = K extends "create"
+  ? CreateInput<M>
+  : K extends "connect"
+    ? WhereUniqueInput<M>
+    : K extends "connectOrCreate"
+      ? ConnectOrCreate<M>
+      : K extends "disconnect" | "delete"
+        ? boolean | WhereInput<M>
+        : K extends "update"
+          ? UpdateInput<M> | NestedUpdateToOne<M>
+          : K extends "upsert"
+            ? NestedUpsertToOne<M>
+            : NestedStatementHasNoOperandType;
+
+/** @see {@link ManyNestedOperand} */
+type NestedStatementHasNoOperandType = {
+  "this nested-write statement is in SUPPORTED_STATEMENTS but has no operand type here": never;
+};
+
+/**
+ * The nested write statements a `create`'s `data` may carry.
  *
  * A to-one cannot take the collection-shaped ones, which is the only structural
- * difference between the two sides.
+ * difference between the two sides — and `COLLECTION_ONLY_STATEMENTS` is the
+ * list both planners refuse there, subtracted rather than restated.
  */
 type NestedCreate<R extends RelationInfo> = R["kind"] extends "many"
-  ? {
-      create?: CreateInput<R["target"]> | CreateInput<R["target"]>[];
-      createMany?: { data: CreateInput<R["target"]>[]; skipDuplicates?: boolean };
-      connect?: WhereUniqueInput<R["target"]> | WhereUniqueInput<R["target"]>[];
-      connectOrCreate?:
-        | ConnectOrCreate<R["target"]>
-        | ConnectOrCreate<R["target"]>[];
-    }
+  ? { [K in NestedCreateStatement]?: ManyNestedOperand<K, R["target"]> }
   : {
-      create?: CreateInput<R["target"]>;
-      connect?: WhereUniqueInput<R["target"]>;
-      connectOrCreate?: ConnectOrCreate<R["target"]>;
+      [K in Exclude<
+        NestedCreateStatement,
+        CollectionOnlyStatement
+      >]?: OneNestedOperand<K, R["target"]>;
     };
 
 interface ConnectOrCreate<M extends ModelTypeInfo> {
@@ -809,27 +945,23 @@ interface ConnectOrCreate<M extends ModelTypeInfo> {
   create: CreateInput<M>;
 }
 
-type NestedUpdate<R extends RelationInfo> = NestedCreate<R> &
-  (R["kind"] extends "many"
-    ? {
-        set?: WhereUniqueInput<R["target"]> | WhereUniqueInput<R["target"]>[];
-        disconnect?:
-          | WhereUniqueInput<R["target"]>
-          | WhereUniqueInput<R["target"]>[];
-        delete?: WhereUniqueInput<R["target"]> | WhereUniqueInput<R["target"]>[];
-        deleteMany?: WhereInput<R["target"]> | WhereInput<R["target"]>[];
-        update?: NestedUpdateOne<R["target"]> | NestedUpdateOne<R["target"]>[];
-        updateMany?:
-          | NestedUpdateMany<R["target"]>
-          | NestedUpdateMany<R["target"]>[];
-        upsert?: NestedUpsert<R["target"]> | NestedUpsert<R["target"]>[];
-      }
-    : {
-        disconnect?: boolean | WhereInput<R["target"]>;
-        delete?: boolean | WhereInput<R["target"]>;
-        update?: UpdateInput<R["target"]> | NestedUpdateToOne<R["target"]>;
-        upsert?: NestedUpsertToOne<R["target"]>;
-      });
+/**
+ * An `update`'s `data` carries the whole of `SUPPORTED_STATEMENTS`, so this is
+ * the unsubtracted set rather than `NestedCreate` intersected with the rest.
+ *
+ * Stated as one mapped type for a reason beyond brevity: written as an
+ * intersection, the create-half and the update-half were two places a
+ * statement's operand could be given, and a statement listed in both with
+ * different operands would silently intersect to something neither file meant.
+ */
+type NestedUpdate<R extends RelationInfo> = R["kind"] extends "many"
+  ? { [K in NestedWriteStatement]?: ManyNestedOperand<K, R["target"]> }
+  : {
+      [K in Exclude<
+        NestedWriteStatement,
+        CollectionOnlyStatement
+      >]?: OneNestedOperand<K, R["target"]>;
+    };
 
 interface NestedUpdateOne<M extends ModelTypeInfo> {
   where: WhereUniqueInput<M>;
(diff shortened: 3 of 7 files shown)


packages/gemi/orm/relation-filters.ts
```diff
@@ -26,6 +26,21 @@
...
+ * than a hand-written copy of the five names.
+ *
+ * The paragraph above says two places must agree about this shape and must not
+ * import each other. The caller-facing type is the place that was left out of
+ * that arrangement — it spelled `some`/`every`/`none` and `is`/`isNot` again,
+ * in a file that imports nothing from here, so nothing failed when the two
+ * drifted. That is the mechanism #369 exists to close, and the same one behind
+ * #326, #333, #336 and #337: a type free to describe a compiler it never reads.
+ */
+export type ManyRelationOperator = (typeof MANY_FILTER_OPERATORS)[number];
```
nstfkc, line 46: **should fix — nothing enforces that these two have a consumer.**

These are the whole mechanism, and they are the part with no tripwire. I reverted `orm/types.ts` to `54d4278` on this branch and ran everything: typecheck clean, 3110 gemi tests pass, `oxlint orm --deny-warnings` reports 0/0, 792 template tests pass, 116 type assertions pass. `ManyRelationOperator` and `OneRelationOperator` — and `JsonFilterName`, `ListFilterName`, `HavingOperator`, the three nested-write unions — all become exports nothing imports, and nothing notices.

`architecture.test.ts`'s "no export is referenced only inside its own file" is the test that should catch precisely this, and it misses because its scan is `/^export (?:async )?function (\w+)|^export const (\w+)|^export class (\w+)/`. Adding `export type` / `export interface` to that regex makes the derivation load-bearing at test time.

I measured the blast radius so it does not have to be guessed at: of 99 exported types and interfaces under `orm/**`, six are referenced only in their own file — `PolicedModel` and `PolicyLookup` (`policy.ts`), `ProtocolSkewWarner`, `RelationNode`, `RelationOrdering`, `SchemaFields`. Six `MAY_BE_EXPORTED_UNUSED` entries or six unexports, and then reverting `types.ts` is a red test instead of a silent no-op.

packages/gemi/orm/types.ts
```diff
@@ -350,19 +379,31 @@ type StringFilter<V> = NonNullable<V> extends string
...
+ * `SqliteDialect.listFilters` refuses the column outright, because SQLite has
+ * no array type.
  */
 type ListFilter<E> = {
-  equals?: E[];
-  has?: E;
-  hasEvery?: E[];
-  hasSome?: E[];
-  isEmpty?: boolean;
+  [K in ListFilterName]?: ListFilterOperand<K, E>;
```
nstfkc, line 409: **should fix — this one *can* be type-tested here, and I did it.**

The body says `ListFilter` "cannot get [a type-level regression test] here" because a `String[]` lives in `postgres-only.prisma` whose artifacts are uncommitted. That is true of testing it *through a model*, but the type does not need a model. `FieldFilter` is already exported from `gemi/orm`, and its list arm is reachable with nothing but a type argument:

```ts
type ListArm = Exclude<FieldFilter<string[]>, string[]>;
type Expect<A, B> = [A] extends [B] ? ([B] extends [A] ? true : false) : false;
const listExact: Expect<
  ListArm,
  { equals?: string[]; has?: string; hasEvery?: string[];
    hasSome?: string[]; isEmpty?: boolean }
> = true;
```

Run through the template's `bun run test:types` on this branch: passes, and it is a real assertion — flipping `has?: string` to `has?: string[]` in the expected shape turns it red. No `String[]` column, no Postgres, no generated artifact.

So the one derivation the write-up says nothing in this repository can assert is assertable in about ten lines in `filters.test-d.ts`, which is also where the `Boolean` / `Bytes` gap note lives. Worth landing, given finding 1 — it would be the only committed thing pinning `ListFilterOperand`'s three-way split.

packages/gemi/orm/relation-filters.ts
```diff
@@ -26,6 +26,21 @@
 const MANY_FILTER_OPERATORS = ["every", "none", "some"] as const;
 const ONE_FILTER_OPERATORS = ["is", "isNot"] as const;
 
+/**
+ * A **third** consumer of the same two tuples, and the reason they are exported
```
nstfkc: **nit — the tuples are not exported.** "the reason they are exported at all" reads as being about `MANY_FILTER_OPERATORS` / `ONE_FILTER_OPERATORS`, and those are still module-private `const`s two lines up; what is exported is the two derived types. It matters more than usual here because `architecture.test.ts` carries a docblock recording that these two tuples were deliberately *un*exported ("used only by the function directly beneath them"), so a reader who follows this sentence and greps for `export const` finds the opposite of what it says. "…and the reason the two unions below exist" would be accurate.

packages/gemi/orm/types.ts
```diff
@@ -57,12 +57,41 @@
 // eagerly evaluated — mapping over `keyof` at the top level, say, or wrapping one
 // in a conditional that has to resolve the whole shape — would turn a cyclic
 // schema into an instantiation-depth error rather than a slow compile.
+//
+// **The operator grammars below map over a tuple, and that is not the mapping
+// the paragraph above warns about.** `RelationFilter`, `NestedCreate` and
+// `NestedUpdate` are `[K in <five literal strings>]`, not `[K in keyof M]`: the
```
nstfkc: **nit — `[K in <five literal strings>]` is wrong for two of the three types it names.** `NestedUpdate<many>` maps over all eleven of `SUPPORTED_STATEMENTS`; `NestedCreate<many>` over four; five is the relation-filter count (3 + 2) and nothing else. The argument the paragraph is making — finite key set, fixed whatever the schema is — is right and the measurement backs it; it is just the number that is off. `[K in a fixed tuple of literals]` says it without being checkable-wrong.

While here: the paragraph names `RelationFilter`, `NestedCreate` and `NestedUpdate`, but `ListFilter`, `JsonPathFilter` and `HavingComparison` became mapped types in this same change and are covered by the same argument. Naming all six keeps the next reader from wondering whether the other three were considered.

packages/gemi/orm/compile/group-by.ts
```diff
@@ -475,15 +475,35 @@ function assertPureAggregate(
...
+} as const;
+
+/** One of {@link OPERATORS}' keys. `HavingComparison` maps over this. */
+export type HavingOperator = keyof typeof OPERATORS;
+
+/**
+ * The same table, widened for lookup: the key comes from a caller's object and
+ * is `string`, not {@link HavingOperator} — that is the question being asked.
+ */
+const OPERATOR_SQL: Record<string, string | undefined> = OPERATORS;
```
nstfkc: **nit — the alias hands out write access to the source of truth.** The annotation is what widens the keys for the caller-supplied lookup, which is the right call, but it also drops `readonly`: `OPERATOR_SQL.contains = "like"` typechecks and mutates the same object `HavingOperator` is derived from. `Readonly<Record<string, string | undefined>>` keeps the `OPERATOR_SQL[key]` lookup working and costs nothing.

(The `| undefined` is right and is the bit that matters — it is what makes the `if (!operator)` guard honest rather than a cast that lies.)

packages/gemi/orm/types.ts
```diff
@@ -784,52 +845,123 @@ export type FindUniqueOrThrowArgs<M extends ModelTypeInfo> = FindUniqueArgs<M>;
...
+ * subtracted rather than listed again, so the two files cannot answer
+ * differently (#369). Concretely: `connect`, `connectOrCreate`, `create`,
+ * `createMany`.
+ *
+ * The subtraction is the compiler's own reasoning, not an approximation of it.
+ * `assertNestedOperand` refuses an `EXISTING_ROW_ONLY` key whenever the
+ * operation is a `create`, because there is nothing linked to a row that does
+ * not exist yet.
+ */
+type NestedCreateStatement = Exclude<NestedWriteStatement, ExistingRowStatement>;
```
nstfkc, line 872: **Verified, no change requested** — recording the measurement since this is the one non-mechanical site.

The subtraction is exact. I checked out `54d4278`'s `orm/types.ts` as a second module and compared all four arms through the public `CreateInput` / `UpdateInput` with a mutual-assignability helper (and a control pair that correctly reports `false`): `NestedCreate<many>`, `NestedCreate<one>`, `NestedUpdate<many>` and `NestedUpdate<one>` are each *mutually assignable* with their pre-change selves — so every operand arm survives, not merely the key sets. Independently, under the template's strict `test:types`, the four key sets come out as exactly `{create, createMany, connect, connectOrCreate}`, `{create, connect, connectOrCreate}`, all eleven, and `{create, connect, connectOrCreate, disconnect, delete, update, upsert}`.

And the subtraction really is the compiler's own reasoning rather than a restatement of the old key list: `nested-writes.ts:353` refuses an `EXISTING_ROW_ONLY` key exactly when `CREATE_ONLY_STATEMENTS.has(operation)`.

One thing the body gets wrong about this: its table says `NestedCreate<many> = SUPPORTED`. The docblock above this line has it right.

nstfkc (commented):
**Verdict: correct, and the mechanism is the right one — minor comments only, nothing blocking.** The derivation holds under measurement in both directions, and every one of the five grammars is *exactly* the type it replaced. What is missing is a committed tripwire: undoing this change is still silent.

## What I verified, and how

Environment first, since it is the trap this batch keeps paying for: `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`, `build:bin` + `gemi-orm-generator` symlinks + both `prisma generate` runs. So the template numbers below are this branch.

| | |
| --- | --- |
| `tsc --noEmit -p tsconfig.json` | clean |
| `bun --bun vitest run` (packages/gemi) | 120 files, 3110 passed, 1 skipped |
| `bun --bun vitest run app/models` (template) | 20 files, 792 passed, 133 skipped |
| `bun run test:types` (template) | 6 files, 116 passed, no type errors |
| `bunx prisma generate` ×2 | zero-line diff on the committed artifacts |
| `gh pr checks 381` | `sqlite` pass, `postgres` pass |

**The `postgres` job passing settles the one thing the PR says it could not measure.**

**The derivations are exactly the types they replaced — measured, not read.** I checked out `54d4278`'s `orm/types.ts` as a second module and compared the two through the public `CreateInput` / `UpdateInput` with a mutual-assignability helper (plus a control pair that correctly reports `false`):

```
NestedCreate<many>  new ≡ old      NestedCreate<one>  new ≡ old
NestedUpdate<many>  new ≡ old      NestedUpdate<one>  new ≡ old
```

So the subtraction algebra *and* every operand arm survive, not just the key sets. Separately, under the template's strict `test:types` (the package's own `tsconfig` has `strictNullChecks` off, so exactness assertions run there are blind to `V` vs `V | null` — worth knowing for anyone repeating this):

- `Exclude<FieldFilter<string[]>, string[]>` ≡ the old hand-written `ListFilter` — exactly;
- `Extract<FieldFilter<JsonValue>, { path: any }>` ≡ the old hand-written `JsonPathFilter`, all ten operands — exactly;
- both `RelationFilter` arms ≡ the old ones — exactly;
- `having` still takes `equals: null` / `not: null`, still refuses `lt: null`, `startsWith` and `in`.

**The derivation follows the compiler — reproduced.** I rebuilt the five `@ts-expect-error` probes (they are not committed, so I could not take them on trust) and added four key-set assertions. Branch as it stands: 125 passed. One name added to each of the four runtime lists with `types.ts` untouched: all five directives report *"Unused"* and all four nested-write key sets widen — 9 failures, and the other 116 assertions in `filters.test-d.ts` stay green. That is the property the PR claims, and it holds.

**Runtime equivalence** re-checked by hand: `NO_SET_OF_ROWS` is `{set, updateMany, deleteMany}`, exactly the three names both inline chains tested; and `NestedCreate = SUPPORTED − EXISTING_ROW_ONLY` really is what `nested-writes.ts:353` enforces (`EXISTING_ROW_ONLY.has(key) && CREATE_ONLY_STATEMENTS.has(operation)`), not an approximation of it.

**Declaration emit and publish** — `build:types` emits `dist/orm/compile/{where,group-by,nested-writes}.d.ts` and `dist/orm/relation-filters.d.ts`, and `dist/orm/types.d.ts` carries all five `import type` lines. `scripts/build-publish.ts` copies `dist/` whole and only rewrites the `exports` map, so the deep `.d.ts` ship with the package. No new publish risk.

## Findings

**1 — should fix. Nothing committed here fails if the fix is undone.** I measured it: `git checkout 54d4278 -- packages/gemi/orm/types.ts` on this branch leaves *everything* green — typecheck clean, 3110 gemi tests, `oxlint orm --deny-warnings` 0/0, 792 template tests, 116 type assertions. The five compile-side exports become dead and nothing says so, because `architecture.test.ts`'s export scan is `/^export (?:async )?function|^export const|^export class/` and these are `export type`.

The PR's answer — *"the derivation is what makes them redundant; there is no second list left to guard"* — is right about *drift* and wrong about *reversion*, and reversion puts the repo back in exactly the state #369 describes. There is a one-regex fix that is already the right test: extend that scan to `export type` / `export interface`. I measured the cost — of 99 exported types and interfaces under `orm/**`, only six are referenced solely in their own file (`PolicedModel`, `PolicyLookup` in `policy.ts`; `ProtocolSkewWarner`; `RelationNode`; `RelationOrdering`; `SchemaFields`), so it is six allow-list entries or six unexports, and after it `types.ts` ceasing to read `JsonFilterName` & co. is a red test rather than a silent revert.

**2 — should fix. "`ListFilter` cannot get a type-level test here" is measurably false**, and it is the one claim in the write-up that turned out not to survive checking. See the inline comment: `FieldFilter` is already exported from `gemi/orm`, and `FieldFilter<string[]>` reaches `ListFilter<string>` with no `String[]` column, no Postgres and no generated artifact. Ten lines in `filters.test-d.ts`.

**3 — should fix. The PR body's subtraction table contradicts the code.** It says `NestedCreate<many> = SUPPORTED`; the code says `Exclude<NestedWriteStatement, ExistingRowStatement>` — `connect`, `connectOrCreate`, `create`, `createMany`, not all eleven. The docblock on `NestedCreateStatement` and `.orm-sessions/issue-369.md` both have it right, so it is the body's table alone. Worth fixing before it becomes the merge commit's account of the change.

**4 — should fix. This will not close #369 on merge.** `closingIssuesReferences` on the PR is empty: "Close #369" is in the *title*, and GitHub only reads closing keywords from the body. This batch already has four issues left clerically open for the same class of reason (see `orm.md`'s note about #299) — one line in the body avoids a fifth.

**Note, not a finding — a sixth statement list the types still do not read.** `JOIN_TABLE_SUPPORTED` (`nested-writes.ts:2151`) admits five of the eleven statements through an implicit m-n, and `NestedUpdate<many>` offers all eleven regardless. Pre-existing, not derivable today (`RelationInfo` records `kind`, not whether the relation is a join-table one), and it fails the *safe* way — the type is wider than the compiler, so you get a runtime refusal rather than #336's silent absorption. But it does mean "which operator exists is now stated once" has an exception, and the body would be more accurate for naming it beside the `NestedUpdate<one>` / `upsert` carve-out. Same follow-up issue, probably.

## What I did not check

Live Postgres locally (no service container — CI's `postgres` job covers it, and it is green); the 133 skipped template tests, which are the Postgres-only ones; `bun run lint` across the whole package (I ran `oxlint orm` only).

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs.

## #380 × #381 — resolving `JsonPathFilter` the obvious way silently drops #380's `array_contains` narrowing

Both PRs rewrite the same declaration in `packages/gemi/orm/types.ts`, from opposite directions:

- **#380** keeps the hand-written object literal and narrows one property:
  `array_contains?: Exclude<JsonValue, null>`, with a docblock explaining that `jsonArrayContains`
  binds a bare `null` as SQL NULL and `x @> NULL` is NULL.
- **#381** replaces the whole literal with a mapped type over the compiler's own tuple:
  `{ path: JsonPath } & { [K in JsonFilterName]?: JsonPathOperand<K> }`, where
  `JsonPathOperand<"array_contains"> = JsonValue` — **`null` included**.

`git merge` conflicts on it, which is the good news. The hazard is the *resolution*: #381's entire
point is that the mapped type replaces the literal, so taking #381's side is the natural call — and
it reverts #380's narrowing without anything in the diff saying so.

I did exactly that and measured it. `bun run typecheck` in `packages/gemi` is **clean**, and the
package suite is **3123 passed** — both green with the narrowing gone. What catches it is the
template's type tests:

```
FAIL  app/models/filters.test-d.ts > null at a path is refused, having no reading left
 ❯ filters.test-d.ts:328
   328|     // @ts-expect-error `x @> NULL` is NULL too — a null *inside* the document is fine
   329|     await UserModel.findMany({ where: { metadata: { path: ["a"], array_contains: null } } });
```

That is #380's own new assertion firing. Worth noting **CI does run this** — `bun run test:types`,
`ci.yml` line 215 — so the loss is caught rather than shipped. It is still a red batch, and a
reviewer reading only the conflict resolution would have no reason to expect it.

### The resolution that keeps both

Take #381's mapped type, and put #380's narrowing in the arm:

```ts
type JsonPathOperand<K extends JsonFilterName> = K extends "array_contains"
  ? Exclude<JsonValue, null>          // #380 — `x @> NULL` is NULL
  : K extends "string_contains" | "string_starts_with" | "string_ends_with"
    ? string
    : JsonPathScalar;
```

Measured: `test:types` **117 passed, no type errors**; `typecheck` clean; `packages/gemi`
**3123 passed**. #380's property-level docblock on `array_contains` has nowhere to live under a
mapped type — it should move onto the `array_contains` arm, since it is the reasoning for the
`Exclude`.

Two things that compose fine and needed checking:

- **`JsonPath` survives.** #381 keeps `path: JsonPath` and does not redefine the alias, so #380's
  narrowing to `readonly string[]` carries through — confirmed by #380's numeric-segment
  `@ts-expect-error` still firing.
- **`compile/where.ts` auto-merges cleanly** (#380 changes `compileJsonFilter`'s operand handling,
  #381 turns `JSON_FILTERS` into `JSON_FILTER_NAMES` beside it) with no semantic interaction I could
  find.

**Cheapest order: #380 before #381**, so the resolution lands in #381 where the mapped type is
being introduced and its author owns the arm.

Both PRs merged locally, and then the full ten-PR integration branch with this resolution applied:
`typecheck` clean, `packages/gemi` **3185 passed**, template sqlite **845 passed**, `test:types`
**117 passed**, live Postgres via `postgres.sh` **1247 passed**.

packages/gemi/orm/relation-filters.ts
nstfkc: Addressed in `02b8b66`, and you were right that this is the part with no tripwire — I reproduced your revert before changing anything and got the same green run.

The scan now matches `export type` and `export interface`. One thing that fell out of doing it, which is worth recording because it made the first attempt a false pass: **`architecture.test.ts` had to stop counting itself as a source.** The docblock I wrote for the change names three of the unions in prose, and that alone was enough to make them "referenced" — reverting `types.ts` reported five of eight and silently excused `JsonFilterName`, `ListFilterName` and `HavingOperator`. A comment is not a use, so the test-file loop skips its own file now. With that, the revert reports all eight:

```
relation-filters.ts exports ManyRelationOperator, and nothing else uses it
relation-filters.ts exports OneRelationOperator, and nothing else uses it
compile/where.ts exports ListFilterName, and nothing else uses it
compile/where.ts exports JsonFilterName, and nothing else uses it
compile/group-by.ts exports HavingOperator, and nothing else uses it
compile/nested-writes.ts exports NestedWriteStatement, and nothing else uses it
compile/nested-writes.ts exports ExistingRowStatement, and nothing else uses it
compile/nested-writes.ts exports CollectionOnlyStatement, and nothing else uses it
```

Your blast radius measured exactly right — the same six, no others. I took the allow-list rather than the unexports, with a reason on each: all six name a *parameter or field of a public signature* (`PolicedModel` / `PolicyLookup` are what an application writes `$policies` and its resolver against, `ProtocolSkewWarner` is a callback shape a host would declare, and the three schema-walk shapes are the vocabulary `ModelSchema` and the include/order-by planners' results are built from). That is a real use the scan cannot see, because it only counts references inside `orm/**`. Unexporting them would be a separate judgement about the public surface, and not one this PR should be making.

`export type (\w+)` deliberately does not match `export type { … } from`: a re-export declares nothing, and counting it would record the wrong home file. Noted at the regex.

packages/gemi/orm/types.ts
nstfkc: You are right and I was wrong about the scope of the claim — landed in `02b8b66`. "Nothing in this repository can assert it" was true of asking *through a model* and I wrote it as though it were true of the type. `ListFilter` is reached from `FieldFilter`'s array branch on the element type alone, and `FieldFilter` is public, so a type argument is the whole of what it needs.

I split it into two assertions rather than one, because they fail on different drift and I wanted both directions pinned:

```ts
type ListArm = Exclude<FieldFilter<string[]>, string[]>;

expectTypeOf<keyof ListArm>().toEqualTypeOf<
  "equals" | "has" | "hasEvery" | "hasSome" | "isEmpty"
>();

expectTypeOf<ListArm>().toEqualTypeOf<{
  equals?: string[]; has?: string; hasEvery?: string[];
  hasSome?: string[]; isEmpty?: boolean;
}>();
```

Used `expectTypeOf` over a hand-rolled `Expect<A, B>` for two reasons: it is what the rest of the file uses, and mutual assignability does not actually catch a *widened* key set — an extra optional property is assignable in both directions, so `Expect<>` would have passed on the one drift this exists to catch. `toEqualTypeOf` is strict and does.

Both mutated, on the source rather than on the expectation:

| mutation | result |
| --- | --- |
| `"hasNone"` appended to `LIST_FILTER_NAMES` in `compile/where.ts` | red — *"Expected literal string isEmpty, Actual literal string hasNone"*, on the `keyof` assertion |
| `ListFilterOperand`'s `has` arm `E` → `E[]` | red — on the shape assertion |
| neither | 117 passed, no type errors |

The `Boolean` / `Bytes` note in that docblock is a different gap and stays; I reworded the `String[]` half of it to say there is no *column* here, and to point at the test one down.

packages/gemi/orm/relation-filters.ts
nstfkc: Correct, and worse than a wording slip given what `architecture.test.ts` records right next door — a reader following that sentence to `export const` finds the opposite. Fixed in `02b8b66` to "…and the reason the two unions below exist", with a second sentence saying outright that the tuples stay module-private, that the unexport is still the deliberate decision the test's docblock describes, and that what crosses the file boundary is the derived type, which is erased.

That claim is now also enforced rather than asserted: the same commit widens `architecture.test.ts`'s scan to `export type`, so if either tuple ever *is* exported and goes unused, the test says so.

packages/gemi/orm/types.ts
nstfkc: Both fixed in `02b8b66`. The number was wrong in the way you say — five is `MANY_FILTER_OPERATORS` + `ONE_FILTER_OPERATORS` and nothing else; the range across the six is two (`ONE_FILTER_OPERATORS`) to eleven (`SUPPORTED_STATEMENTS`). It now reads `[K in a fixed tuple of literals]`, which is the claim the argument actually rests on, with a parenthetical noting the sizes differ and that none of them is a function of the model.

And the paragraph names all six now — `ListFilter`, `JsonPathFilter`, `HavingComparison`, `RelationFilter`, `NestedCreate`, `NestedUpdate` — for the reason you give: three named out of six reads as three considered out of six.

Worth flagging one interaction, since it changed how I wrote it: naming the *unions* in prose inside `architecture.test.ts` turned out to make them "referenced" by the export scan I widened for finding 1, which silently excused three of the eight. That is why the scan now skips its own file. Naming the mapped types here is safe — they are not exported — but it is the same class of trap.

packages/gemi/orm/compile/group-by.ts
nstfkc: Right, and it is not only theoretical — the alias is the *only* handle on that object outside its own two lines, so a write through it would be a write to the thing `HavingOperator` is derived from, with the derived type still reporting the old table. Fixed in `02b8b66`: `Readonly<Record<string, string | undefined>>`. `OPERATOR_SQL[key]` still typechecks with `key: string`, `Object.keys(OPERATORS)` in the error message is untouched, and the docblock now says why both halves of the annotation are there — the `Readonly` for this, the `| undefined` for the `if (!operator)` guard, which is your point and the one I had already written down.

packages/gemi/orm/types.ts
nstfkc: Thank you for the mutual-assignability check against `54d4278`'s `types.ts` — that is the assertion I wanted and could not think how to write, since it pins the *operands* and not just the key sets. My own check was the four key lists by hand, which would have missed an operand arm that drifted.

And the body was wrong exactly where you say. `NestedCreate<many> = SUPPORTED` is `NestedUpdate<many>`'s row; `NestedCreate<many>` is `SUPPORTED − EXISTING_ROW_ONLY`, the four keys your `test:types` run reports (`create`, `createMany`, `connect`, `connectOrCreate`). The docblock in `types.ts` has it right, so this was a transcription error into the PR description and nothing more. Corrected in the body, with a note saying it was corrected rather than a silent edit.

Your `nested-writes.ts:353` observation is the better statement of why the subtraction is the compiler's reasoning and not a restatement of the old key list, and I have folded it into the reply rather than the code — the docblock above this line already cites the refusal, just not by line.

No code change from this one.

nstfkc:
## Review round 1 — disposition of every comment

Seven comments: six inline, one **[cross-PR review]**. **Five addressed, one cross-PR item addressed, one was a verification requesting no change. Nothing declined, nothing deferred.** Everything is in `02b8b66`, pushed to this branch.

| # | comment | disposition |
| --- | --- | --- |
| [3743288408](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288408) | *should fix* — nothing enforces the derived types have a consumer | **addressed** — `architecture.test.ts` scans `export type` / `export interface`, and skips itself |
| [3743288411](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288411) | *should fix* — `ListFilter` **can** be type-tested here | **addressed** — two assertions in `filters.test-d.ts`, both mutated |
| [3743288415](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288415) | *nit* — the tuples are not exported | **addressed** — comment corrected, and the claim is now enforced by the widened scan |
| [3743288417](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288417) | *nit* — "five literal strings" is wrong for two of three | **addressed** — reworded, and all six mapped grammars named |
| [3743288420](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288420) | *nit* — `OPERATOR_SQL` hands out write access | **addressed** — `Readonly<…>` |
| [3743288422](https://github.com/nstfkc/gemi/pull/381#discussion_r3743288422) | verified, no change requested | **no code change** — the body's table error it names is fixed |
| [cross-PR](https://github.com/nstfkc/gemi/pull/381#issuecomment-5230860032) | #380 × #381 — the natural resolution drops `array_contains`'s narrowing | **addressed here**, per its own "the author owns the arm" |

### The one that mattered

Your revert of `types.ts` to `54d4278` reproduced exactly: typecheck clean, 3110 tests, `oxlint --deny-warnings` 0/0, 792 template tests, 116 type assertions. A PR whose entire premise is *"the two lists can no longer drift because one is derived from the other"* had nothing failing when the derivation was removed. That is the same shape as the defect it fixes, one level up, and it is the finding I should have caught myself — my session file even records that `architecture.test.ts` "only scans `export const` / `export function` / `export class`, not `export type`, so it is unaffected", and I wrote that down as a reassurance rather than as the problem it was.

Widening the scan turned up one thing worth knowing for anyone else who touches that test: **it counted its own prose as a reference.** The docblock I added names three of the unions, and that alone kept those three green under the revert while the other five went red — a false pass produced by the fix's own documentation. The test-file loop now skips `architecture.test.ts`, and the revert names all eight.

### The cross-PR item

`JsonPathOperand`'s `array_contains` arm is `Exclude<JsonValue, null>` on this branch now, with `filters.test-d.ts` asserting the refusal beside the existing `equals: null` one, plus a companion line pinning that a `null` *inside* the document is still expressible. So there is no longer a resolution that can lose it: whichever side of the `types.ts` conflict is taken, the narrowing is in both.

The finding is **#380's** and the credit is in the docblock and in the test. Landing it here is what that comment asks for — the arm belongs to the mapped type this PR introduces, and #380 has nowhere to put a property-level docblock once the literal is gone. It also stands on its own on `main` today: `assertJsonOperand` waves `array_contains` through without a scalar check, so a bare `null` reaches `dialect.jsonArrayContains` and binds as SQL NULL, and `x @> NULL` is NULL rather than false. Reverting the arm to `JsonValue` reports the directive unused, so it cannot go quiet.

The rest of that comment I did **not** act on, deliberately: `JsonPath` and `compile/where.ts` are #380's own and untouched here, and the merge order (#380 before #381) is a merge-time decision, not something this branch can encode.

### Not in scope for this PR

- **#386's `childField` → `childFields` rename.** Not applicable here in any case — this branch touches no composite-relation code — and #386 is not on `main`, so there are no new names to reference. Merge-time work for whoever rebases after #386.
- **#378's `connectOrCreate` hit-branch short-circuit** and **#385's numeric JSON path segments** belong to those branches; nothing here reaches them.
- **`NestedUpdate<one>` offering `upsert` / `delete` on an owning-side to-one** is still unfixed and still pre-existing. `RelationInfo` records `kind` but not which side holds the foreign key. It wants an issue and a generator change, not a guess in `types.ts`.

### Runs, on `02b8b66`

`bun install` in the worktree first; `readlink -f templates/saas-starter/node_modules/gemi` resolves into it, so the template suites exercised this branch.

| | |
| --- | --- |
| `tsc --noEmit -p tsconfig.json` | clean |
| `tsc --noEmit -p tsconfig.generated.json` | clean |
| `vitest run` (packages/gemi) | 120 files, 3110 passed, 1 skipped |
| `vitest run app/models` (template) | 20 files, 792 passed, 133 skipped |
| `bun run test:types` (template) | 6 files, **117** passed (was 116), no type errors |
| `bunx prisma generate` ×2 | zero-line diff on the committed artifacts |
| `gemi check models --ignore bench` | 7 model files, 14 registered models |
| `oxlint orm --deny-warnings` | 0 warnings, 0 errors |
| `bun run lint` | 79 warnings, 0 errors (unchanged) |
| `bun run build:types` | emits, and `types.d.ts`'s four `import type` lines resolve |

Every test added or changed was mutated: the export scan against a full `types.ts` revert (eight names), the `ListFilter` key set against `"hasNone"` in `LIST_FILTER_NAMES`, the `ListFilter` operand split against `has: E → E[]`, and the `array_contains` refusal against reverting the arm to `JsonValue`. Each goes red on its own mutation and green with it restored.

**Postgres still not run** — no service container here, unchanged from the original description. Nothing in this round changes a runtime table's membership.

Also fixed the clerical one: the body carries `Closes #369` now, and `closingIssuesReferences` reports #369 rather than being empty.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `02b8b667` for
this PR.

### The #380 × #381 resolution hazard is **gone**, and your round is what removed it

The previous pass flagged this as the batch's quietest trap: the `types.ts` conflict resolved the
natural way — take this branch's mapped type, which is the whole point of the PR — silently reverted
#380's `array_contains?: Exclude<JsonValue, null>`, and typecheck plus the full package suite were
green with the narrowing gone.

Putting the narrowing on `JsonPathOperand`'s own arm makes that unreachable. Taking this side of the
conflict is still the only sensible resolution, and it now *keeps* the property:

```ts
type JsonPathOperand<K extends JsonFilterName> = K extends "array_contains"
  ? Exclude<JsonValue, null>
  : …
```

Measured rather than trusted, since a green suite was the failure mode: on the merged tree I
reverted that arm to plain `JsonValue`. `packages/gemi` stays green — 120 files, 3199 passed —
exactly as the last pass reported, and `test:types` goes red with
`Unused '@ts-expect-error' directive` at `filters.test-d.ts:372`. Restored, 118 pass, no type errors.
Your parenthetical at that arm ("so the two do not resolve to whichever side of the merge is taken")
is now a measured fact on the integrated tree.

### Your merge note (a) was accurate

`filters.test-d.ts` did conflict add/add inside `test("null at a path is refused, …")`, both sides
adding an `array_contains: null` `@ts-expect-error` with different wording. Kept this side's, which
also carries the "only the top level goes" companion line pinning `array_contains: [1, null]`. The
`types.ts` conflict was resolved to this branch's mapped type as you said it safely could be.

Note (b) also holds unchanged: this branch touches no composite-relation code, and #386's
`childField` → `childFields` rename produces no error attributable to it — all five `TS2552`s on the
all-ten tree are #378's and #379's.

### Nothing else on this PR

No new cross-PR interaction. `architecture.test.ts`'s widened export scan, the `group-by.ts`
`Readonly`, and the `ListFilter` assertion in `filters.test-d.ts` all survive the merge untouched
and green.

848 passed / 134 skipped · `test:types` 6 files / 118 passed, no type errors · `postgres.sh` 1251
passed / 520 skipped (plus the two SQLite-half differential suites the harness fails by design) ·
`oxlint orm --deny-warnings` 0/0 · both `prisma generate` runs leave `git status` clean.

Closing keyword verified: `closingIssuesReferences` resolves to #369 — the title-only defect is
fixed.

Issue #259, opened 2026-07-29
Prisma.AnyNull is neither mapped nor refused: written as {}, and filtered to the complement of the right rows


nstfkc:
`Prisma.AnyNull` is not refused: a write stores it as `{}`, and a filter returns the complement of the correct rows.

Raised in review of #239, which hardened `jsonNullKind` against forged sentinels. That PR does not introduce this — it has been true since #223, where `AnyNull` was left out of `SENTINELS` on the argument that mapping it *"would quietly accept something Prisma does not."* Omitting it does not refuse it either: it falls through to the ordinary data path.

## What happens today

`jsonNullKind(Prisma.AnyNull)` returns `null`, so:

| | Prisma | gemi |
| --- | --- | --- |
| write | `PrismaClientValidationError` | accepted, stored as the jsonb object `{}` |
| filter (`equals: AnyNull`) | the SQL-NULL rows **and** the JSON-null rows | the rows whose column holds `{}` |

`cast.ts` `fieldParam` reaches `JSON.stringify(value)`, and `JSON.stringify(Prisma.AnyNull)` is `{}`. In `where.ts`, `compileFieldFilter` sees no sentinel, so a bare `AnyNull` takes the `equals` shorthand and `{ equals: AnyNull }` compiles to `= $1::jsonb` bound with `{}`.

The write is the same silent shape #239 exists to close: a value carrying no data stored as a plausible-looking object, nothing raised, and both outcomes legal for the column. The filter is worse than merely wrong — `AnyNull` means *either kind of null*, so the correct answer is exactly the rows gemi excludes.

## What to do

Two halves, and they are separate decisions:

1. **Write.** Refuse it the way #225 refuses a bare sentinel in a filter — name it in the message and note that Prisma rejects it here too. This needs `AnyNull` recognised (a predicate alongside `jsonNullKind`, or a widened return) and a throw raised where the write arguments carry schema/operation context (`compile/write.ts`), not in `fieldParam`, which has neither.
2. **Filter.** Either implement it — `<col> is null or <col> = 'null'` — or refuse it. Either beats returning the complement.

Both need differential coverage against the real `Prisma.AnyNull` in `templates/saas-starter/app/models/` (`writes.differential.test.ts` for the refusal, `differential.test.ts` for the filter), since the package suite cannot construct the genuine object.

[...]

PR #382, opened 2026-08-09
Close #376: a raw `::jsonb` parameter binds as JSON text, as it did under Prisma

### LINKED ISSUE #376: A `$1::jsonb` parameter in raw SQL binds as a jsonb string, where Prisma bound the parsed object
Raw SQL that binds a JS string into a `jsonb` parameter behaves differently under `DB.execute`/`DB.query` than it did under Prisma's `$executeRaw`, and the difference silently writes a jsonb **string** where an object was intended. No error, no type complaint, and the SQL text reads correctly by eye.

Found migrating a large app off Prisma onto the ORM: 16 sites in one directory were affected, and it took an end-to-end test against a real Postgres to notice.

## The divergence

`$1::jsonb` resolves the **parameter's** type to `jsonb`, and Bun's driver then JSON-encodes the JS string it was handed. So `'{"a":1}'` arrives as the jsonb string `"{\"a\":1}"`, not the object. Prisma's `$executeRaw` sent parameters untyped and let Postgres parse the text.

Measured on Postgres 16:

```
$1::jsonb with a JS string    → {"folioAiWorkspace": "{\"version\":1,…}"}
'{...}'::jsonb || $1::jsonb   → [{"attribution":…}, "{\"version\":1,…}"]
```

The second line is the sharp end: `||` between an object and a jsonb *string* is **array concatenation, not a merge**. A statement written to merge metadata into a document instead appends the serialised text as a new array element. Downstream code that expects an object gets an array, or silently drops the row when a `isRecord()` guard fails.

A bare `$1` into a `jsonb` column has the identical fault.

## The fix, for anyone hitting this

`::text::jsonb` — pin the parameter to `text` so Postgres parses it, which is exactly what Prisma did implicitly and depends on no driver JSON behaviour at all:

```sql
-- wrong: the parameter is typed jsonb, so the driver JSON-encodes the string
UPDATE "AsyncJob" SET payload = payload || $1::jsonb WHERE id = $2

-- right
UPDATE "AsyncJob" SET payload = payload || $1::text::jsonb WHERE id = $2
```

`CAST(… AS text)::jsonb` works the same way.

## Why this is worth handling upstream rather than leaving to each caller

**Nothing catches it.** The types are fine — a string is a legal operand. A unit test with a fake `execute` passes, because the SQL text is what you would write by hand and the binding never happens. `oxlint` and `tsc` see nothing. It surfaces only against a real database, and then as *wrong data* rather than an error: the write succeeds, and the corruption is discovered later by whatever reads the column.

**It is a migration trap specifically.** The whole value proposition of modelling the API on Prisma's is that a port is mechanical. This is a case where a byte-identical statement changes meaning, which is the one thing a porter has no reason to check.

**It is invisible in review.** Two rounds of adversarial review passed the change that introduced it — including one that checked the compiler's JSON-filter behaviour in detail. Neither could have caught it; only e2e did.

## Suggested options, in rough order of preference

1. **Handle it in the `sql` tagged template.** If an interpolated value is a plain object or array destined for a JSON operand, bind it as text. That makes the Prisma-shaped statement do the Prisma thing, which is the promise the rest of the ORM keeps.
2. **Document it prominently** next to the raw-SQL escape hatch (`DB.query` / `DB.execute` / `sql`), with the `::text::jsonb` spelling and the `||` array-concatenation consequence spelled out — this is the kind of thing that belongs beside `docs/orm.md`'s existing "KNOWN DIVERGENCE" material, and it is precedented: `dialect/postgres.ts` already records Bun's `TZ`/protocol-skew divergence in exactly that register.
3. **A lint or a dev-mode warning** on `$n::jsonb` in a raw statement, pointing at `::text::jsonb`.

(2) is the minimum. (1) is what would have prevented it.

## Related

`packages/gemi/orm/dialect/postgres.ts` already documents two other places where Bun's driver diverges from what a Prisma-shaped caller expects — the parameterless-vs-parameterized timestamp decode, and `code` vs `errno` on server errors (*"Checking `code` alone — the obvious reading, and the one this started with — matches nothing at all"*). This is a third of the same kind, and the three together suggest the driver's parameter handling deserves one consolidated section rather than three scattered comments.

nstfkc:
The migration trap #376 reports, reproduced and closed at the choke point. A raw statement that Prisma ran as a merge ran as an append under `DB.execute`, with no error, no type complaint, and SQL text that reads correctly by eye.

## What is actually different, measured

Not the cast — the **parameter's type**. Prisma sends a JS string as `text` and lets Postgres parse it into whatever the expression wants. Bun describes the statement, is told the parameter is `jsonb`, and JSON-encodes the string it was handed. So the document arrives as a jsonb *string*.

Postgres 16 in a throwaway container, Bun 1.3.14, and a generated `@prisma/client` 6.19.2 against the same database with `log: [{ emit: "event", level: "query" }]`. Binding the JS string `'{"version":1}'` into a `jsonb` column, then reading `jsonb_typeof` and `payload::text` back:

| statement | prisma | bun |
| --- | --- | --- |
| `values ($1)` | **error** — `column "payload" is of type jsonb but expression is of type text` | `"{\"version\":1}"` — jsonb_typeof `string` |
| `values ($1::jsonb)` | `{"version": 1}` | `"{\"version\":1}"` |
| `values ($1 :: jsonb)` | — | `"{\"version\":1}"` |
| `values ($1::JSONB)` | — | `"{\"version\":1}"` |
| `values (cast($1 as jsonb))` | `{"version": 1}` | `"{\"version\":1}"` |
| `values ($1::pg_catalog.jsonb)` | — | `"{\"version\":1}"` |
| `values ($1::text::jsonb)` | `{"version": 1}` | `{"version": 1}` |

And the sharp end, the same two clients on `'{"attribution":"x"}'::jsonb || $1`:

```
prisma  $1::jsonb        {"attribution": "x", "version": 1}
bun     $1::jsonb        [{"attribution": "x"}, "{\"version\":1}"]     jsonb_typeof -> array
bun     $1::text::jsonb  {"attribution": "x", "version": 1}
```

Three rows of that table changed the shape of the fix:

- **The bare `$1` row.** Prisma *refuses* it with a string operand. So a statement that ran under Prisma with a string carried a cast — which makes "read the cast" cover exactly what a port can be carrying, rather than being a partial answer to a general problem. With an *object* operand a bare `$1` works on both clients and is untouched here.
- **The `cast(… as jsonb)` row.** It mis-stores identically, so handling only `::` would have left a second door open with the first one shut.
- **Prisma's own two paths disagree about a string**, which is what settled the rule below.

## The change

`renderFragment` — the single choke point every `DB.query` / `DB.execute` fragment passes through — reads the text that follows each parameter marker, before placeholders are assigned. Where the caller cast their own placeholder to `json`/`jsonb`, the placeholder is rewritten to cast through `text` and the value is bound as JSON text:

```
sql`… "payload" || ${JSON.stringify(patch)}::jsonb …`
  ->  … "payload" || $1::text::jsonb …
```

That is the pair `compile/cast.ts` already established for a typed `Json` column at #209 — the cast and the serialisation travel together, and neither is correct alone. The raw path had no field to ask, which is the whole reason it was missing; the cast the caller wrote is the declared type by another route.

**Rewriting a caller's SQL wants a justification.** The cast is preserved exactly — `::text::jsonb` and `::jsonb` denote the same type and the expression's value is unchanged — while the *parameter's* type, which is the thing that differs from Prisma and the thing no caller can otherwise reach through the driver, is set to what Prisma set it to. The raw path was never byte-preserving anyway: it is what turns a marker into `$1` or `?`.

### A string passes through, where `fieldParam` serialises one

The one place the two paths differ, and it is deliberate — **it is which of the two Prisma does on each path**:

```
prisma.doc.create({ data: { payload: '{"version":1}' } })      -> jsonb string
prisma.$executeRaw`… values (${'{"version":1}'}::jsonb)`       -> jsonb object
```

With a field, nobody wrote a cast and the value is a JSON *value*. With a caller-written cast, the caller has said the parameter is JSON *text* being cast, which is the only reading under which `${JSON.stringify(patch)}::jsonb` — the spelling a port contains — means what it meant. Somebody who wants the jsonb string writes the serialisation they mean.

Everything else follows from that one rule: an object or array is serialised (so `${patch}::jsonb`, which worked before because Bun serialises it, still works), a number and a boolean become JSON numbers and booleans where Prisma answers `cannot cast type bigint to jsonb` — gemi is more permissive there, and consistent with its own typed `Json` column, which is the more useful of the two to agree with. `null` stays SQL NULL rather than becoming a JSON null, matching both Prisma and `fieldParam`. `DbNull` / `JsonNull` mean what they mean on a column, because `JSON.stringify` turns them into `{}` — #259's mis-store reachable by a third route. A `bigint`, a function and a cycle raise `InvalidArgumentError` naming the operation rather than dying inside `JSON.stringify`.

### What is left alone, on purpose

`$1::pg_catalog.jsonb` (not what anyone writes), `$1::jsonb[]` (an array of documents — serialising the JS array into one JSON text would be wrong), `as jsonb)` with no `cast(` opening it, and a bare `$1`, whose residue is argued above and written down in the docs. `DB.sql` — Bun's own tag — does not pass through `renderFragment` and is untouched; the docs say so.

SQLite needs no branch: `::` is not SQLite syntax and there is no `jsonb` type, so the match cannot fire on a statement SQLite could run. It is still asserted on both dialects, because "cannot fire" is a claim.

## Tests

**Compile-level** (`orm/sql.test.ts`, 70 → 97): the emitted text and the bound values together, on both dialects — the five recognised spellings, the four left alone, a nested fragment retyped by the *outer* statement's cast (which is why this cannot be decided where the value is interpolated), two casts in one statement, the neighbouring parameter still going through `encodeUntyped`, and the refusals. The 70 that were there are unchanged and still pass; no existing expectation moved.

**Live Postgres** (`templates/saas-starter/app/models/raw-sql.test.ts`, 9 new, named `— postgres` so CI's `-t postgres` job runs them): `DB.execute` against `User.metadata`, read back through Bun with `jsonb_typeof` — the document stored as a document, `||` merging, every shape through both cast spellings, and a raw `where "metadata" = ${…}::jsonb` matching a row the *ORM* wrote, which is the property that makes raw SQL an escape hatch rather than a second representation.

Also there: **the driver behaviour the fix rests on, pinned**. `raw.unsafe` with `$1::jsonb` still stores the jsonb string and `||` still yields an array. Nothing in the ORM would notice Bun changing that, and if it does change, the retyping is solving a problem that no longer exists — this failing is the only way anyone finds out.

Reversed to check they bite: with the retyping disabled, 6 of the 9 new Postgres tests fail (`expected 'string' to be 'object'`, `expected 'array' to be 'object'`, and the filter matching nothing). The object and array cases pass either way, which is the measurement above showing through.

```
packages/gemi        3145 passed | 1 skipped        typecheck clean, oxlint orm --deny-warnings clean
template  sqlite      801 passed | 133 skipped
template  postgres    891 passed | 788 skipped      via app/models/postgres.sh app/models -t postgres
```

## Docs

`docs/orm.md` gains **`### A `::jsonb` cast you write types the parameter as `text`** in the Raw SQL section, with the measured table, the `||` consequence, the string rule, and the uncovered residue; mirrored into `llms-full.txt`, cross-linked from `## `Json` columns`. The paragraph that said "stringify it yourself if the column is JSON" is corrected — followed literally on Postgres, that advice *was* the bug.

`known-divergences.test.ts` pins the section and the source it derives from, the way it already pins the `TZ=UTC` one: the failure mode this whole issue describes is a fix nobody can see from the call site, so a page that stops describing it is a regression.

Closes #376.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -2045,6 +2045,16 @@ So gemi reads the cast. Where a parameter carries one, the placeholder is retype
 what a `Json` *column* has always compiled to. Recognised: `::jsonb`, `::json`, `::text::jsonb`,
 `cast(… as jsonb)`, in any case and with whitespace anywhere Postgres allows it.
 
+**On Postgres only, including the `cast(…)` spelling.** SQLite has no `jsonb` type and no `::`
+operator, so the mis-store cannot arise there — but `cast(x as json)` *does* parse and run on SQLite,
+where it means something else entirely: SQLite accepts any type name in a `CAST` and applies affinity
+rules, so `cast('{"a":1}' as json)` is `0`. gemi therefore leaves such a statement exactly as you
+wrote it on SQLite — retyping it would rewrite working SQL into a syntax error, and serialise for a
+cast that never meant JSON. There is no portable spelling to reach for: the two dialects do not share
+a JSON grammar at all, which is the same reason `path:` takes an array on Postgres and a `$.a.b`
+string on SQLite for the ORM's own filters. A raw statement that touches JSON is a per-dialect
+statement.
+
 At such a parameter, **a string is JSON text** — `${'{"a":1}'}::jsonb` stores the object — where a
 `Json` column takes the same string as a JSON string value. That is not an inconsistency to
 work around: it is which of the two Prisma does on each path. Pass the serialisation you mean:
@@ -2064,8 +2074,10 @@ same fault and there is nothing to read — no cast, no column type. It is also 
 cannot be carrying, because Prisma *refuses* it (`column "payload" is of type jsonb but expression is
 of type text`); write the cast. Schema-qualified spellings (`$1::pg_catalog.jsonb`) and `jsonb[]` are
 deliberately left alone — the second is an array of documents, and serialising the JS array into one
-would be wrong. `DB.sql`, Bun's own tagged template, is untouched by all of this: it binds through
-the driver directly, so `::text::jsonb` is yours to write there.
+would be wrong. A cast on an expression **around** the placeholder rather than on the placeholder
+itself is not read either — `(${'{"a":1}'})::jsonb` and `cast(coalesce(${v}) as jsonb)` still
+mis-store, so put the cast on the placeholder. `DB.sql`, Bun's own tagged template, is untouched by
+all of this: it binds through the driver directly, so `::text::jsonb` is yours to write there.
 
 ### The rowcount is a primitive, not a diagnostic
 
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -1982,13 +1982,14 @@ stops at this door.
 
 Values are handed to the driver as they arrive, with two exceptions that exist so a fragment means
 the same thing on both dialects: a `Date` binds as milliseconds on SQLite — which is what the ORM
-stores — and a boolean as `0`/`1`.
+stores — and a boolean as `0`/`1`. A third applies only where the statement itself says the
+parameter is JSON — see [below](#a-jsonb-cast-you-write-types-the-parameter-as-text).
 
-A plain object is **not** JSON-encoded for you. The compiler only knows to do that because a field
-says `Json`, and guessing from the value would turn a mistyped parameter into a successfully-written
-string. What the driver then does with it differs: Bun binds an object into a Postgres `jsonb`
-column correctly, and on SQLite binds it to something no row matches. Stringify it yourself if the
-column is JSON and you want the same answer on both.
+A plain object is otherwise **not** JSON-encoded for you. The compiler only knows to do that because
+a field says `Json`, and guessing from the value would turn a mistyped parameter into a
+successfully-written string. What the driver then does with it differs: Bun binds an object into a
+Postgres `jsonb` column correctly, and on SQLite binds it to something no row matches. Stringify it
+yourself if the column is JSON and you want the same answer on both.
 
 **The separator `join` takes is written into the SQL**, so it is not a free string: whitespace,
 commas, or `and` / `or`, and nothing else. Anything more expressive is legitimate but has to say so
@@ -2008,6 +2009,63 @@ through, which has none of them and is still an injection.
 interpolation slot is bound — including an object that merely *looks* like a fragment, which is what
 a parsed request body can be made to contain. That check is membership, not shape, precisely so that
 `{"text": "…", "binders": []}` arriving from outside the program is a parameter and not a statement.
+
+### A `::jsonb` cast you write types the parameter as `text`
+
+**Read this before porting a raw statement off Prisma's `$executeRaw`.** On Postgres this is the one
+place a byte-identical statement used to change meaning, and it changed it silently.
+
+```ts
+// Merges the patch into the document, on both clients.
+await DB.execute(
+  sql`update "AsyncJob" set "payload" = "payload" || ${JSON.stringify(patch)}::jsonb
+      where "id" = ${id}`,
+)
+```
+
+The divergence is in the **parameter's** type, not the cast's. Prisma sends a JS string as `text` and
+lets Postgres parse it into the target type. Bun asks the server what the statement wants, is told
+`jsonb`, and JSON-encodes the string it was handed — so the document arrives as a jsonb *string*.
+Measured on Postgres 16, binding the JS string `'{"version":1}'`:
+
+| statement | Prisma | Bun, unassisted |
+| --- | --- | --- |
+| `values ($1)` | error: `jsonb` vs `text` | `"{\"version\":1}"` — a jsonb string |
+| `values ($1::jsonb)` | `{"version": 1}` | `"{\"version\":1}"` |
+| `values (cast($1 as jsonb))` | `{"version": 1}` | `"{\"version\":1}"` |
+| `values ($1::text::jsonb)` | `{"version": 1}` | `{"version": 1}` |
+
+The `||` case is the sharp end: concatenating an object with a jsonb *string* is **array
+concatenation, not a merge**, so a statement written to merge metadata appends the serialised text as
+a new element. The write succeeds, nothing raises, the SQL reads correctly by eye, and the damage is
+found later by whatever reads the column.
+
+So gemi reads the cast. Where a parameter carries one, the placeholder is retyped through `text` —
+`$1::text::jsonb` — and the value is serialised to match, which is what Prisma did implicitly and
+what a `Json` *column* has always compiled to. Recognised: `::jsonb`, `::json`, `::text::jsonb`,
+`cast(… as jsonb)`, in any case and with whitespace anywhere Postgres allows it.
+
+At such a parameter, **a string is JSON text** — `${'{"a":1}'}::jsonb` stores the object — where a
+`Json` column takes the same string as a JSON string value. That is not an inconsistency to
+work around: it is which of the two Prisma does on each path. Pass the serialisation you mean:
+
+```ts
+sql`… = ${JSON.stringify(patch)}::jsonb`               // the object
+sql`… = ${JSON.stringify(JSON.stringify(patch))}::jsonb`  // the jsonb string, if you want that
+```
+
+Everything else follows from "the parameter is JSON text": an object or array is serialised for you,
+so `${patch}::jsonb` works too; a number and a boolean become JSON numbers and booleans; `null` stays
+SQL NULL, and `DbNull` / `JsonNull` mean what they mean on a column. A `bigint` has no JSON form and
+raises rather than binding something wrong.
+
+**What is not covered, and why that is the right line:** a bare `$1` into a `jsonb` column has the
+same fault and there is nothing to read — no cast, no column type. It is also the shape a Prisma port
+cannot be carrying, because Prisma *refuses* it (`column "payload" is of type jsonb but expression is
+of type text`); write the cast. Schema-qualified spellings (`$1::pg_catalog.jsonb`) and `jsonb[]` are
+deliberately left alone — the second is an array of documents, and serialising the JS array into one
+would be wrong. `DB.sql`, Bun's own tagged template, is untouched by all of this: it binds through
+the driver directly, so `::text::jsonb` is yours to write there.
 
 ### The rowcount is a primitive, not a diagnostic
 
@@ -2143,6 +2201,11 @@ On Postgres the parameter is cast — `$1::text::jsonb` — and the value serial
 is what lets a bare number or boolean through. Binding it raw makes the driver send an `integer` to
 a `jsonb` column, and serialising *without* the cast stores the jsonb string `"42"` instead of the
 number, which is worse than refusing it. SQLite needs neither: it stores JSON as text.
+
+A raw statement gets the same cast from the one you write yourself — see
+[A `::jsonb` cast you write types the parameter as `text`](#a-jsonb-cast-you-write-types-the-parameter-as-text),
+which is also where the one difference between the two paths is spelled out: at a cast you wrote, a
+JS string is JSON *text* rather than a JSON string value.
 
 > **Upgrading:** `Json` values written by a *pre-release* build of this ORM on Postgres were stored
 > as JSON **strings** rather than as objects — `{ a: 1 }` landed as `"{\"a\":1}"`. Reads undid it,
--- a/packages/gemi/orm/json-param.ts
+++ b/packages/gemi/orm/json-param.ts
@@ -1,0 +1,243 @@
+import {
+  type Fragment,
+  fromParamSegments,
+  paramSegments,
+} from "./compile/fragment";
+import { InvalidArgumentError } from "./errors";
+import { jsonNullKind } from "./json-null";
+
+/**
+ * A raw parameter the caller cast to `json`/`jsonb` carries **JSON text**.
+ *
+ * ## The statement that changes meaning between the two clients
+ *
+ * ```sql
+ * update "AsyncJob" set payload = payload || $1::jsonb where id = $2
+ * ```
+ *
+ * Byte-identical under Prisma's `$executeRaw` and under `DB.execute`, and it
+ * did two different things. Measured on Postgres 16 through Bun 1.3.14 and a
+ * generated `@prisma/client` 6.19.2, binding the JS string
+ * `'{"version":1}'` into a `jsonb` column:
+ *
+ *   form                      prisma                       bun
+ *   $1                        error: jsonb vs text         "{\"version\":1}"   (a jsonb string)
+ *   $1::jsonb                 {"version": 1}               "{\"version\":1}"
+ *   cast($1 as jsonb)         {"version": 1}               "{\"version\":1}"
+ *   $1::text::jsonb           {"version": 1}               {"version": 1}
+ *
+ * The mechanism is the parameter's *type*, not the cast's. Prisma sends a JS
+ * string as `text` and lets Postgres parse it into the target type; Bun asks
+ * the server what the statement wants, is told `jsonb`, and JSON-encodes the
+ * string it was handed — so the string becomes a jsonb **string** rather than
+ * the object it spells. `::text::jsonb` pins the parameter back to `text`,
+ * which is what Prisma did implicitly and what `compile/cast.ts` already emits
+ * for a typed `Json` column (#209).
+ *
+ * **The `||` case is the sharp end.** Concatenating an object with a jsonb
+ * *string* is array concatenation, not a merge:
+ *
+ *   payload || $1::jsonb        [{"attribution": …}, "{\"version\":1}"]
+ *   payload || $1::text::jsonb  {"attribution": …, "version": 1}
+ *
+ * A statement written to merge metadata into a document appends the serialised
+ * text as a new array element instead. The write succeeds, nothing raises, the
+ * SQL reads correctly by eye, and the corruption is found by whatever reads the
+ * column next.
+ *
+ * ## So the cast is read, and the parameter retyped
+ *
+ * A `::json`/`::jsonb` the caller wrote onto their own placeholder is the one
+ * thing in a raw statement that says "this parameter is JSON" — there is no
+ * field to ask, which is the whole difference between this and `fieldParam`.
+ * When it is there, the placeholder is rewritten to cast through `text` and the
+ * value is serialised to match, so the Prisma-shaped statement does the Prisma
+ * thing.
+ *
+ * **Rewriting a caller's SQL text needs a reason, and this is it:** the cast is
+ * preserved exactly — `::text::jsonb` and `::jsonb` denote the same type and
+ * the expression's result is unchanged — while the *parameter's* type, which is
+ * the thing that differs from Prisma and the thing no caller can otherwise
+ * reach, is set to what Prisma set it to. The raw path was never
+ * byte-preserving in any case: it is what turns a marker into `$1` or `?`.
+ *
+ * ## A string passes through, unlike the typed path — and Prisma agrees twice
+ *
+ * `fieldParam` serialises *every* value, so a `Json` column set to the JS string
+ * `"42"` stores the jsonb string `"42"`. Here a string is handed over as it
+ * arrived, so `${'{"a":1}'}::jsonb` stores the object. That is not an
+ * inconsistency to be tidied away: it is which of the two Prisma does, on each
+ * path, and both were measured against 6.19.2.
+ *
+ *   prisma.doc.create({ data: { payload: '{"version":1}' } })  -> jsonb string
+ *   prisma.$executeRaw`… values (${'{"version":1}'}::jsonb)`   -> jsonb object
+ *
+ * The distinction that explains both: with a field, nobody wrote a cast and the
+ * value is a JSON *value*; with a caller-written cast, the caller has said the
+ * parameter is JSON *text* being cast. Somebody who wants the jsonb string
+ * writes the serialisation they mean — `${JSON.stringify(v)}::jsonb`.
+ *
+ * ## What is left, and why the remainder is the right remainder
+ *
+ * A bare `$1` into a `jsonb` column is the same fault and is not detectable
+ * here: no cast, no field, nothing to read. It is also the shape a Prisma port
+ * cannot contain — Prisma *refuses* it (`column "payload" is of type jsonb but
+ * expression is of type text`), so a statement that ran under Prisma with a
+ * string operand had a cast on it. An object there works on both clients and is
+ * untouched. `docs/orm.md` carries the residue; this covers what a port can
+ * actually be carrying.
+ *
+ * Schema-qualified spellings (`$1::pg_catalog.jsonb`) and `jsonb[]` are
+ * deliberately not matched. The first is not what anyone writes and the second
+ * is a different operand — an array of documents, where serialising the JS
+ * array to one JSON text would be wrong.
+ */
+
+/**
+ * `::jsonb` / `::json` applied to the placeholder that precedes it.
+ *
+ * Whitespace is allowed on both sides of the operator because Postgres allows
+ * it and `$1 :: jsonb` mis-stores identically — measured, not assumed. The
+ * lookahead keeps `::jsonb[]` out: that operand is an array of documents, and
+ * the serialisation this triggers would flatten it into one.
+ *
+ * **`::text::jsonb` matches too**, and is then rewritten to itself. The cast is
+ * already right there — it is the spelling this whole note recommends, and the
+ * one a caller who has already been bitten will have written — but the *value*
+ * still has to be serialised, or an object bound to it reaches Postgres as
+ * `[object Object]`. Matching both spellings is what makes them mean the same
+ * thing, which is the only defensible relationship between them.
+ */
+const OPERATOR_CAST = /^\s*::\s*(?:text\s*::\s*)?(jsonb|json)\b(?!\s*\[)/i;
+
+/** The same cast written as a function: `cast($1 as jsonb)`. */
+const FUNCTION_CAST = /^\s*as\s+(jsonb|json)\s*\)/i;
+
+/**
+ * ...which is only that when a `cast(` opened immediately before the
+ * placeholder. Without this, `select ${x} as jsonb)` — nonsense, but nonsense
+ * the caller wrote — would have its parameter serialised on the strength of two
+ * keywords that mean nothing here.
+ */
+const CAST_OPEN = /\bcast\s*\(\s*$/i;
+
+/**
+ * The fragment with every JSON-cast parameter retyped through `text`, and the
+ * positions of the parameters that were.
+ *
+ * Returns the fragment unchanged, and an empty set, when there is no such cast
+ * — which is every statement in the compiler and nearly every raw one.
+ */
+export function retypeJsonParameters(fragment: Fragment): {
+  fragment: Fragment;
+  jsonParameters: ReadonlySet<number>;
+} {
+  const segments = paramSegments(fragment);
+  const jsonParameters = new Set<number>();
+  let rewritten: string[] | undefined;
+
+  // `segments[i]` is the text before parameter `i` and `segments[i + 1]` the
+  // text after it, so the last segment trails the final parameter and has no
+  // parameter of its own.
+  for (let index = 0; index < segments.length - 1; index++) {
+    const after = segments[index + 1];
+
+    const operator = OPERATOR_CAST.exec(after);
+    if (operator) {
+      rewritten ??= [...segments];
+      rewritten[index + 1] =
+        `::text::${operator[1]}${after.slice(operator[0].length)}`;
+      jsonParameters.add(index);
+      continue;
+    }
+
+    const fn = FUNCTION_CAST.exec(after);
+    // The preceding segment is read from the rewrite when there is one: a
+    // previous parameter's rewrite replaces the head of this segment and leaves
+    // its tail — the `cast(` this is looking for — where it was, but reading the
+    // live array is what keeps that a fact rather than an assumption.
+    if (fn && CAST_OPEN.test((rewritten ?? segments)[index])) {
+      rewritten ??= [...segments];
+      rewritten[index + 1] = ` as text)::${fn[1]}${after.slice(fn[0].length)}`;
+      jsonParameters.add(index);
+    }
+  }
+
+  if (!rewritten) return { fragment, jsonParameters };
+  return {
+    fragment: fromParamSegments(rewritten, fragment.binders),
+    jsonParameters,
+  };
+}
+
+/**
+ * The value for one such parameter: JSON text, or SQL NULL.
+ *
+ * `dialect.encodeUntyped` deliberately does not run on these. Its two
+ * conversions are for a value being compared against a column the ORM wrote — a
+ * `Date` as milliseconds, a boolean as `0`/`1` — and neither is JSON. A `Date`
+ * here becomes the ISO string `JSON.stringify` gives it, because that is what
+ * putting a `Date` in a JSON document means.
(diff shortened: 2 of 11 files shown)


packages/gemi/orm/sql.ts
```diff
@@ -344,18 +357,29 @@ export function renderFragment(
...
 
-  const rendered = render(fragment, dialect, { model: "DB", operation });
+  // The one thing a raw parameter can be *told* about its type: a `::jsonb` the
+  // caller wrote onto their own placeholder. Where there is one, the parameter
+  // is retyped through `text` and its value serialised, because Bun otherwise
+  // JSON-encodes a string bound to a `jsonb` parameter and stores the document
+  // as a jsonb *string* where Prisma stored the object. See `json-param.ts` —
+  // the mechanism, the measurements, and why a string is not serialised here
+  // while `fieldParam` serialises one.
+  const json = retypeJsonParameters(fragment);
```
nstfkc: **blocking** — this runs on both dialects, and on SQLite it can turn a working statement into a syntax error.

The PR body and session §6 argue no dialect branch is needed because "`::` is not SQLite syntax and SQLite has no `jsonb` type, so the match cannot fire on a statement SQLite could run." That is true of `OPERATOR_CAST`. It is **not** true of `FUNCTION_CAST`: SQLite's `CAST` accepts an arbitrary type name, so `cast(? as json)` parses and runs there today. Measured with `bun:sqlite` (Bun 1.3.14):

```
select cast(? as jsonb) as v                              -> [{"v":0}]      (runs)
select cast(? as json) as v                               -> [{"v":0}]      (runs)
select json_set('{"a":1}', '$.b', cast(? as json)) as v   -> {"a":1,"b":2}  (runs, and is meaningful
                                                             — the cast is what makes it a number
                                                             rather than the string "2")
select cast(? as text)::json as v                         -> ERROR unrecognized token: ":"
```

And through this code path, with `SqliteDialect`:

```
sql`select json_set('{"a":1}', '$.b', cast(${"2"} as json))`
  -> select json_set('{"a":1}', '$.b', cast(? as text)::json)   // SQLite: unrecognized token ":"
```

So a raw statement that runs on SQLite on `main` fails to parse on this branch. Narrow — but the new docs section is in the dialect-general Raw SQL chapter and lists ``cast(… as jsonb)`` under **Recognised** with no dialect caveat, which is an invitation to write exactly this.

The fix is here, where the dialect is already in hand: pass it to `retypeJsonParameters` and skip (or restrict to `OPERATOR_CAST`) on SQLite. That also makes the "cannot fire" reasoning something `tsc` and a test can hold, rather than a property of the regexes.

packages/gemi/orm/json-param.ts
```diff
@@ -0,0 +1,243 @@
...
+ * already right there — it is the spelling this whole note recommends, and the
+ * one a caller who has already been bitten will have written — but the *value*
+ * still has to be serialised, or an object bound to it reaches Postgres as
+ * `[object Object]`. Matching both spellings is what makes them mean the same
+ * thing, which is the only defensible relationship between them.
+ */
+const OPERATOR_CAST = /^\s*::\s*(?:text\s*::\s*)?(jsonb|json)\b(?!\s*\[)/i;
+
+/** The same cast written as a function: `cast($1 as jsonb)`. */
+const FUNCTION_CAST = /^\s*as\s+(jsonb|json)\s*\)/i;
```
nstfkc, line 154: **should fix** (the mechanism behind the `sql.ts` comment) — this is the branch that is reachable on SQLite.

`OPERATOR_CAST` is genuinely unreachable there: `::` does not parse. `FUNCTION_CAST` is not, because SQLite's `CAST` takes any type name — `cast(? as json)` and `cast(? as jsonb)` both parse and run — and the rewrite this pairs with emits `as text)::json`, which SQLite rejects.

If the gate lands in `renderFragment`, this docblock is the place to say *why* only one of the two spellings needed one: `::` is Postgres-only syntax, `cast(… as …)` is not, and "SQLite has no jsonb type" is about semantics rather than about what the parser accepts.

templates/saas-starter/app/models/raw-sql.test.ts
```diff
@@ -295,6 +295,152 @@ function suite(label: string, url?: string) {
       expect(await emails()).toHaveLength(4);
     });
 
+    // --- a jsonb parameter -------------------------------------------------
+
+    /**
+     * **A `::jsonb` cast in a raw statement, which is where a Prisma port goes
+     * wrong silently.** Postgres only: SQLite has no `jsonb` type and no `::`
+     * operator, so these statements are not expressible there at all.
```
nstfkc: **should fix** — "these statements are not expressible there at all" is false for one of the two spellings this very block exercises.

The `test.each` below runs `cast(${value} as jsonb)`, and `cast(x as jsonb)` *is* expressible on SQLite: `select cast(? as jsonb)` parses and returns a row (SQLite's `CAST` accepts an arbitrary type name and applies affinity rules). What is not expressible on SQLite is `::`.

This is the #374 shape — a comment that is wrong at the line the reason belongs to — and it is the same wrong belief that leaves the rewrite ungated in `renderFragment`, so the two want fixing together. Suggested wording: Postgres only, because `::` is not SQLite syntax and SQLite has no `jsonb` semantics for `cast` to reach — not because the text cannot be parsed there.

packages/gemi/orm/sql.test.ts
```diff
@@ -291,6 +292,177 @@ describe("values with no column type behind them", () => {
...
+ * sends a string parameter as `text` and lets Postgres parse it, while Bun is
+ * told the parameter is `jsonb` and JSON-encodes the string, so the document is
+ * stored as a jsonb *string*. `payload || $1::jsonb` then appends it as an array
+ * element instead of merging. `json-param.ts` carries the measurements; these
+ * pin the emitted statement, which is where the fix lives.
+ */
+describe("a parameter the caller cast to json", () => {
+  const document = `{"version":1}`;
+
+  test("the placeholder is retyped through text, on both dialects", () => {
```
nstfkc: **should fix** — this is the test that backs "asserted on both dialects rather than assumed", and it asserts the one spelling that could never have been a problem.

`::jsonb` on SQLite was already unparseable before the change, so `?::text::jsonb` being emitted there costs nothing. The spelling that matters is the function one — `cast(${document} as jsonb)`, tested below at the default (Postgres) dialect only — because that *does* parse and run on SQLite, and the rewrite makes it stop.

Adding the sqlite dialect to the function-cast row is what would have caught the `renderFragment` gap above, and once the gate lands it is the row that pins it: on SQLite the emitted text should be `cast(? as jsonb)` unchanged.

packages/gemi/orm/json-param.ts
```diff
@@ -0,0 +1,243 @@
...
+ * `dialect.encodeUntyped` deliberately does not run on these. Its two
+ * conversions are for a value being compared against a column the ORM wrote — a
+ * `Date` as milliseconds, a boolean as `0`/`1` — and neither is JSON. A `Date`
+ * here becomes the ISO string `JSON.stringify` gives it, because that is what
+ * putting a `Date` in a JSON document means.
+ */
+export function jsonTextParameter(value: unknown, operation: string): unknown {
+  // SQL NULL rather than the JSON value `null`, the same reading `fieldParam`
+  // takes for a nullable `Json` column: an absent value is not a stored null.
+  // Prisma binds a JS `null` at `$1::jsonb` to SQL NULL too — measured.
```
nstfkc: **nit** — `— measured` here has nothing behind it in the record.

Every other measurement claim in this file lands in the table above it or in session §2 (the `$1 :: jsonb` whitespace row, the `[object Object]` row for `::text::jsonb` with an object, the `cast(… as jsonb)` row). The JS `null` row is in none of them, and the PR body's table does not carry it either. The file is otherwise scrupulous about the difference between measured and reasoned, which is exactly why this one word stands out.

The *behaviour* is almost certainly right — with the parameter declared `text` (which the `values ($1)` error text shows Prisma does), `NULL::jsonb` is SQL NULL — so this is about the record, not the code. Either add the row to the header table or say "which follows from the parameter being declared `text`" instead of "measured".

docs/orm.md
```diff
@@ -2009,6 +2010,63 @@ interpolation slot is bound — including an object that merely *looks* like a f
...
+sql`… = ${JSON.stringify(patch)}::jsonb`               // the object
+sql`… = ${JSON.stringify(JSON.stringify(patch))}::jsonb`  // the jsonb string, if you want that
+```
+
+Everything else follows from "the parameter is JSON text": an object or array is serialised for you,
+so `${patch}::jsonb` works too; a number and a boolean become JSON numbers and booleans; `null` stays
+SQL NULL, and `DbNull` / `JsonNull` mean what they mean on a column. A `bigint` has no JSON form and
+raises rather than binding something wrong.
+
+**What is not covered, and why that is the right line:** a bare `$1` into a `jsonb` column has the
```
nstfkc, line 2072: **nit** — this paragraph names three uncovered shapes and omits the fourth the author already knows about.

Session §7.4 records that detection is textual and cannot see a cast applied to a *surrounding* expression, so `(${v})::jsonb` still mis-stores. I confirmed it — `cast(coalesce(${v}) as jsonb)` is likewise left alone. A reader who has been bitten once and comes here to learn what is covered gets `$1`, `pg_catalog`, and `jsonb[]`, and will reasonably read the parenthesised form as covered because the type name is right there.

One clause: a cast on an expression *around* the placeholder rather than on the placeholder itself is not read — put the cast on the placeholder.

nstfkc (commented):
**Verdict: minor-comments — the fix is right, the evidence is unusually good, and one dialect claim in it is measurably false.**

The mechanism holds up. `retypeJsonParameters` runs at the one choke point (`renderFragment`, whose only production caller is `facades/DB.ts`), the marker→binder alignment it depends on is guaranteed by `render`'s own count assertion, and the index it carries into `values.map` is sound because `bindValues` emits exactly one value per binder in order. The cast-and-serialise pairing is the same one `compile/cast.ts` established at #209, and the string-passes-through rule is argued from Prisma's *two* paths rather than from one.

### What I ran

```
bun install (in-tree) — readlink -f templates/saas-starter/node_modules/gemi
  -> …/wf_758c4f33-3da-6/packages/gemi                      (confirmed, not the main checkout)
packages/gemi   bun run typecheck                           clean
packages/gemi   bun --bun vitest run                        3145 passed | 1 skipped
packages/gemi   bunx oxlint orm --deny-warnings             0 warnings, 0 errors
template sqlite bun --bun vitest run app/models             801 passed | 133 skipped
template pg     ./app/models/postgres.sh app/models/raw-sql.test.ts -t postgres
                                                            36 passed | 18 skipped
gh pr checks 382                                            postgres pass, sqlite pass
```

**Mutation check, both levels — the tests bite.** I stubbed `retypeJsonParameters` out at the call site (`const json = { fragment, jsonParameters: new Set<number>() }`) and re-ran:

- `orm/sql.test.ts`: **18 of the 27 new tests fail**, including the DbNull row (`expected [ {} ] to deeply equal [ null ]` — i.e. the `JSON.stringify(DbNull) === "{}"` mis-store the header calls out).
- live Postgres `raw-sql.test.ts`: **exactly the 6 of 9 the PR body claims**, with the same messages. The 3 that pass either way are the object/array rows, which is the §2 measurement showing through, as stated. I reverted the stub and confirmed `git status` clean before reviewing further.

I also re-derived the residues by hand rather than taking them on trust: `${x}::jsonb []` (space before the bracket) is correctly left alone by the lookahead, `select cast(1 as text), ${x} as json` is correctly *not* matched (the `cast(` must terminate the preceding segment), `cast(coalesce(${x}) as jsonb)` is left alone as §7.4 says, a fragment reused twice is retyped only where the outer cast is, and `${x}${unsafeSql("::jsonb")}` *is* retyped — consistent, but nobody has written it down.

### Findings

1. **blocking** — `cast($1 as json)` / `cast($1 as jsonb)` is a *valid, runnable* SQLite statement, and the rewrite turns it into a syntax error. The "SQLite needs no branch" argument covers `::` but not the function spelling. Details inline on `orm/sql.ts` and `orm/json-param.ts`; measured with `bun:sqlite`.
2. **should fix** — the header comment on the new live tests states the same false claim ("not expressible there at all") about a spelling that same test exercises. This is the #374 shape: a comment that is wrong at the line the reason belongs to.
3. **should fix** — the compile test that was meant to back the dialect claim ("on both dialects") only asserts the `::` spelling on SQLite; the function spelling, which is the one that is valid there, is asserted on Postgres only. The claim's own test does not cover the case where the claim fails.
4. **nit** — one `— measured` in `json-param.ts` (the JS `null` row) has no corresponding row in the measurement table, in the PR body, or in the session file.
5. **nit** — the docs' "what is not covered" paragraph names three residues but not `(${v})::jsonb`, which §7.4 identifies and which still mis-stores.

### What I could not check

- **Prisma.** I ran no Prisma client. Every Prisma column in the tables is taken on the author's word; the session file records the harness (a scratch `Doc` model, `prisma db push`, query-event logging) but the scripts were session-scoped and are gone, so nothing in the repository reproduces them. What I *did* check is that the completeness argument's load-bearing half is sound in Postgres's own terms: with the parameter declared `text` — which is what Prisma declares, per the `values ($1)` error text — a jsonb-consuming position genuinely refuses it. `PREPARE p(text) AS SELECT jsonb_set('{"a":1}'::jsonb, '{a}', $1)` errors with *no function matches*, and adding `::jsonb` makes it prepare. So "a port cannot be carrying an uncast `$1` at a jsonb operand" survives a check the session did not run. (One wrinkle, harmless here: `PREPARE p(text) AS SELECT '{...}'::jsonb || $1` *does* prepare, because it resolves to `text || anynonarray`, not the jsonb merge — it would then fail at assignment to a jsonb column.)
- **Bun other than 1.3.14 / Postgres other than 16**, as the session says. The `raw.unsafe` pin is the right shape for that.
- I did not run the full unfiltered `postgres.sh` sweep, only `raw-sql.test.ts` under `-t postgres`.

### Not findings, but worth stating out loud

- `${'hello'}::jsonb` — a non-JSON string — previously stored the jsonb string `"hello"` and now raises `invalid input syntax for type json` at the database. Session §7.2 owns this and it is Prisma's behaviour for the same statement; the docs give the other spelling but do not say that the old one now errors. Pre-1.0, no release affected, so I would leave it.
- A `Date` at a json cast now binds `"1970-01-01T00:00:00.000Z"` and a `Uint8Array` binds `{"0":1,"1":2}`. The `Date` case is documented in `jsonTextParameter`'s docblock and is the right reading; the typed-array case is nobody's intent but also nobody's code.
- Scope is clean: ten files, all of them #376's, no drive-bys. `docs/llms-full.txt` and `docs/orm.md` are a byte-identical mirror (I diffed the two patches' `+`/`-` bodies — same md5). `paramSegments`/`fromParamSegments` are exported from `compile/fragment.ts` but not re-exported from `orm/index.ts`, so the public surface does not widen.

packages/gemi/orm/sql.ts
nstfkc: **Addressed** — you are right, and I re-measured it rather than taking it on trust. `bun:sqlite`, Bun 1.3.14, in this worktree:

```
select cast(? as jsonb) as v                             -> [{"v":0}]      runs
select cast(? as json) as v                              -> [{"v":0}]      runs
select json_set('{"a":1}', '$.b', cast(? as json)) as v  -> {"a":1,"b":2}   runs
select cast(? as text)::json as v                        -> ERROR unrecognized token: ":"
```

So §6's argument covered `OPERATOR_CAST` and not `FUNCTION_CAST`, exactly as you say, and a statement that runs on `main` failed to parse on this branch.

It is also **wrong** there, not merely unnecessary, and that decided the shape of the fix rather than leaving it a coin-flip between your two options. SQLite applies affinity rules to whatever type name a `CAST` names, and one matching none of INT/CHAR/CLOB/TEXT/BLOB/REAL/FLOA/DOUB gets NUMERIC — so `cast('{"a":1}' as json)` is `0`. It is a numeric coercion, not a JSON parse, so serialising a value for it would be a wrong answer even in a world where `::` parsed. Restricting SQLite to `OPERATOR_CAST` would leave the rewrite firing on the one spelling that cannot run there anyway; **skipping entirely** is the honest reading, and it leaves SQLite byte-identical to `main`.

`retypeJsonParameters(fragment, dialect)` now returns early on `!dialect.typesParametersFromStatement`. That is a new `SqlDialect` capability rather than a `dialect.name` check, per the rule the interface's own docblock states — Postgres true (the server describes the statement and Bun encodes to the type it is told), SQLite false, each with the measurement at its declaration.

Pinned twice, because the two halves are different claims: `sql.test.ts` asserts the emitted text and the unserialised value on SQLite, and `raw-sql.test.ts` runs `cast(? as jsonb)` against a real SQLite database — what a parser accepts is not something emitted text can settle, and the emitted text is what the first version got wrong. Both go red with the flag flipped to `true`, the live one with your exact `unrecognized token: ":"`.

`6a06abb`.

packages/gemi/orm/json-param.ts
nstfkc: **Addressed**, and this docblock is where the reason now lives — you put your finger on the right line.

`FUNCTION_CAST` carries the explanation: `::` is Postgres-only *syntax*, so `OPERATOR_CAST` genuinely cannot fire on a statement SQLite could run; `cast(… as …)` is not, SQLite takes any type name there, and the rewrite this pairs with emits `as text)::json`, which SQLite refuses. Stated as the distinction you named — "SQLite has no jsonb type" is a claim about semantics, and what matters here is what the parser accepts.

The header gained a matching section (`## Postgres only, and that is a dialect capability rather than a regex's luck`) with the `bun:sqlite` table, saying plainly that the first version's argument was half right and which half. The gate itself is `SqlDialect.typesParametersFromStatement`, which carries the fuller measurement at its declaration — including the affinity reading, which is the part that makes leaving SQLite alone *correct* rather than merely safe: `cast('{"a":1}' as json)` is `0` there, a numeric coercion rather than a JSON parse.

`6a06abb`.

templates/saas-starter/app/models/raw-sql.test.ts
nstfkc: **Addressed.** The claim was false and, as you say, it is the same wrong belief that left the rewrite ungated — so the two were fixed together.

The header now says Postgres only *for the narrower reason*: `::` is Postgres-only syntax and SQLite's parser rejects it, while `cast(x as jsonb)` — which the `test.each` below does exercise — parses and runs on SQLite and simply means something else there (any type name is accepted; one matching none of INT/CHAR/CLOB/TEXT/BLOB/REAL/FLOA/DOUB gets NUMERIC affinity, so `cast('{"a":1}' as json)` is `0`). It points at `SqlDialect.typesParametersFromStatement` and at the SQLite half of the pin, which is now in this same suite.

That pin is a new sqlite-gated test a few blocks down: `select cast(${"2"} as jsonb)` returns `2` and `select cast(${'{"a":1}'} as json)` returns `0`, through `DB.query` against a real SQLite database. It fails with your `unrecognized token: ":"` when the gate is flipped.

One aside worth recording, since it cost me a cycle: my first draft of that test used `select json_set('{"a":1}', '$.b', cast(? as json))`, which runs fine under `bun:sqlite` but returns **zero rows** through Bun's `SQL` client — with or without a parameter, and independently of this change. Unrelated to #376, but somebody will hit it.

`6a06abb`.

packages/gemi/orm/sql.test.ts
nstfkc: **Addressed** — this is the sharpest of the five, because it is the test whose *job* was to catch the `sql.ts` gap and it asserted the one row that could not.

The `::jsonb`-on-SQLite assertions are gone from the two rows that carried them (`the placeholder is retyped through text` and the value `test.each`): `::` does not parse there, so `?::text::jsonb` cost nothing either way and was never evidence for anything. `only the cast parameter is affected` moved to Postgres too, and now makes its point by the `Date` staying a `Date` rather than becoming the ISO string `JSON.stringify` gives one — which is the same distinction it was making through `encodeUntyped` on SQLite before.

In their place, three rows that pin the gate on the spelling that matters:

```
cast(${document} as jsonb)   ->  cast(? as jsonb)     unchanged
${document}::jsonb           ->  ?::jsonb             unchanged
cast(${document} as json)    ->  cast(? as json)      unchanged
```

plus a value assertion, since text alone cannot see the serialisation: `cast(${document} as jsonb)` binds the string untouched and `${42}::jsonb` binds `42`. All four fail with `typesParametersFromStatement` flipped to `true` on `SqliteDialect`.

And the half a compile test structurally cannot hold — whether SQLite's *parser* accepts it — is now a live sqlite test in `raw-sql.test.ts`, which fails with `unrecognized token: ":"` under the same mutation.

`6a06abb`.

packages/gemi/orm/json-param.ts
nstfkc: **Addressed**, and you are right that it is about the record rather than the code — that row genuinely was not run. I took your second option, because inventing a measurement after the fact is the worse of the two:

```
// Prisma binds a JS `null` at `$1::jsonb` to SQL NULL too — which follows
// from the parameter being declared `text` (the `values ($1)` error text in
// the table above is what shows that it is), since `NULL::text::jsonb` is SQL
// NULL. Reasoned, not measured: it is the one row of that table nobody ran.
```

Saying which it is, out loud, is the point — the file's whole value is that "measured" means something there.

docs/orm.md
nstfkc: **Addressed.** The reading you describe — type name right there, therefore covered — is exactly the one that gets somebody bitten a second time by the page that was meant to stop it, so it earns its clause:

> A cast on an expression **around** the placeholder rather than on the placeholder itself is not read either — `(${'{"a":1}'})::jsonb` and `cast(coalesce(${v}) as jsonb)` still mis-store, so put the cast on the placeholder.

Both spellings named, since you confirmed the second. Mirrored byte-for-byte into `llms-full.txt` (the two patches' `+`/`-` bodies still hash the same), and `json-param.ts`'s header carries the same residue with the reason it is not parsed — parentheses would need real parsing to reach a spelling nobody writes.

This section also gained a dialect paragraph from the blocking finding: `cast(… as jsonb)` was listed under **Recognised** with no caveat in a dialect-general chapter, which was the invitation you named.

`6a06abb`.

nstfkc:
## Review round 1 — `6a06abb`

All five findings addressed, none declined, one thing deferred that nobody raised. The blocking one was right, and it was right for a reason I had stated backwards rather than merely omitted, so the fix is a capability rather than a patch.

| # | finding | disposition |
| --- | --- | --- |
| 1 | **blocking** — `cast($1 as jsonb)` runs on SQLite and the rewrite makes it a syntax error | **addressed** — gated on a new `SqlDialect.typesParametersFromStatement` |
| 2 | **should fix** — `FUNCTION_CAST`'s docblock is where the reason belongs | **addressed** |
| 3 | **should fix** — the live tests' header claims these are inexpressible on SQLite | **addressed** |
| 4 | **should fix** — the "both dialects" compile test asserts only the `::` spelling on SQLite | **addressed**, and the SQLite rows are now the ones that pin the gate |
| 5 | **nit** — a `— measured` with nothing behind it | **addressed** — relabelled as reasoned, with the derivation |
| 6 | **nit** — the docs' residue list omits the cast-on-a-surrounding-expression case | **addressed** |

### The blocking one

Re-measured before fixing, `bun:sqlite` / Bun 1.3.14:

```
select cast(? as jsonb) as v                             -> [{"v":0}]      runs
select cast(? as json) as v                              -> [{"v":0}]      runs
select json_set('{"a":1}', '$.b', cast(? as json)) as v  -> {"a":1,"b":2}   runs
select cast(? as text)::json as v                        -> ERROR unrecognized token: ":"
```

Confirmed. §6's "the match cannot fire on a statement SQLite could run" was true of `::` and false of `cast(… as …)`, and a raw statement that runs on `main` failed to parse on this branch.

**It is also wrong there and not merely unnecessary**, which is what decided between the reviewer's two suggested fixes. SQLite applies affinity rules to any type name a `CAST` names, and one matching none of INT/CHAR/CLOB/TEXT/BLOB/REAL/FLOA/DOUB gets NUMERIC — so `cast('{"a":1}' as json)` is `0`. A numeric coercion, not a JSON parse. Serialising a value for it would be a wrong answer even if `::` had parsed, so restricting SQLite to `OPERATOR_CAST` is not enough: the retyping is skipped there entirely, and SQLite is byte-identical to `main`.

`retypeJsonParameters` now takes the dialect and returns early on `!dialect.typesParametersFromStatement`. A capability rather than a `dialect.name` check, because that is the rule `SqlDialect`'s own docblock states, and because — as the reviewer put it — "cannot fire" should be something `tsc` and a test can hold rather than a property of two regular expressions. Postgres declares it `true` (the server describes the statement and Bun encodes to the type it is told, which is the whole of #376); SQLite `false`. Both declarations carry their measurement.

### Tests

The compile rows that "asserted on both dialects" asserted, on SQLite, only the `::` spelling — the one that could never have been a problem, since it does not parse there. Replaced by the function spelling on SQLite, emitted unchanged, with a value assertion because text alone cannot see the serialisation.

And the half a compile test structurally cannot hold — what SQLite's *parser* accepts — is a live sqlite test in `raw-sql.test.ts`: `cast(${"2"} as jsonb)` returns `2`, `cast(${'{"a":1}'} as json)` returns `0`.

**Mutation check, both levels.** Flipping `SqliteDialect.typesParametersFromStatement` to `true`:

```
orm/sql.test.ts               all 4 new sqlite rows fail
raw-sql.test.ts (sqlite)      1 fails — SQLiteError: unrecognized token: ":"
```

Restored afterwards, and the postgres-side mutation from the original PR still holds.

### Verification

`bun install` in-tree first; `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`, not the main checkout. `bun run build:bin`, the `gemi-orm-generator` symlink into both `node_modules/.bin`, and both `prisma generate` runs, exactly as CI does them.

```
packages/gemi   bun run typecheck                          clean
packages/gemi   bun --bun vitest run                       3149 passed | 1 skipped   (was 3145)
packages/gemi   bunx oxlint orm --deny-warnings            0 warnings, 0 errors
template sqlite bun --bun vitest run app/models            802 passed | 133 skipped  (was 801)
template pg     ./app/models/postgres.sh app/models -t postgres
                                                           892 passed | 789 skipped  (was 891)
template        bun run test:types                         116 passed, no type errors
```

`docs/orm.md` and `docs/llms-full.txt` are still a byte-identical mirror — the two patches' `+`/`-` bodies hash the same.

### Deferred

**The `childField` → `childFields` rename from #386.** Not raised here, and not mine to do: #386 is not on `main`, so I cannot reference the new names without reaching into another PR's branch. This PR touches none of the affected code anyway — `dialect/index.ts` gains one capability member, and nothing in the relation resolver moves — so it should merge into #386's rename without conflict. Merge-time work either way.

Nothing else in this PR belongs to another one. #378's `connectOrCreate` hit-branch short-circuit and #385's numeric JSON path segments are cross-PR items assigned elsewhere and are untouched here.

### Not changed, and stated so it can be disagreed with in one place

The three "not findings" from the review body — `${'hello'}::jsonb` now erroring at the database, a `Date` binding as its ISO string, a `Uint8Array` binding as `{"0":1,…}` — are left as they are, on the reviewer's own reasoning: the first is Prisma's behaviour for the same statement and pre-1.0, and the other two are documented where they are decided. Session §7.2 owns the first if anyone wants to reopen it.

One aside worth recording, found while writing the live SQLite test and unrelated to this PR: `select json_set('{"a":1}', '$.b', …)` runs fine under `bun:sqlite` but returns **zero rows** through Bun's `SQL` client, with or without a parameter. The test uses a bare `cast(…)` instead.

`Closes #376` is in the body, and `closingIssuesReferences` resolves to #376 — checked, per the #381/#386 clerical failure.

PR #383, opened 2026-08-09
Close #366: the row a delete returns is redacted as a delete

### LINKED ISSUE #366: An operation-keyed `redact` does not fire on a `delete` that carries an `include`
Adding an `include` to a `delete` silently changes which `context.operation` a policy sees for the returned row, so an operation-keyed `redact` stops firing. Found while reviewing #365; it is the same shape as #364 — a column the model meant to hide reaches the caller the moment an `include` is added — but a different mechanism, and #365 does not fix it.

## Measured

On the template fixture, with a policy carrying:

```ts
redact: (ctx, row) => { if (ctx.operation === "delete") row.name = null }
```

```
delete({ where })                              -> { id: 6, name: null }    redacted
delete({ where, include: { children: true } }) -> { id: 7, name: "kept" }  NOT redacted
```

## Why

A `delete` whose plan has relations or counts takes the read-first path. The pre-read runs as `$exec("findFirst", …)` (`orm/Model.ts:993`), so `policyContext` (`orm/Model.ts:815`) is built with operation `"findFirst"` — and **that pre-read row is the one returned to the caller**. The inner `$exec("delete", { where })` does run with operation `"delete"`, but its row is discarded.

Logging both invocations shows `CTXOP: findFirst` on the row that is returned, then `CTXOP: delete` on the row that is thrown away.

## This contradicts a comment already in the file

`orm/Model.ts:940-943` states:

> A policy that varies by `context.operation` therefore sees "delete" …

Whatever that is true of for the included children, it is plainly false for the root row — which is the row the caller gets.

## Scope

- Only the read-first path, i.e. only a `delete` that carries an `include` or a `_count`. A plain `delete` redacts correctly.
- Not a regression from #365, and not introduced by it — #365 fixes `omit` on this same path, which is why this turned up next to it. `omit` and `redact` are two different hiding mechanisms and only the first is fixed.
- The template ships no policy that branches on `context.operation`, which is why no existing test catches it.

## Worth deciding, not just patching

The fix is not obviously "pass `delete` as the operation to the pre-read": the pre-read genuinely *is* a read, and a `before`-style policy that scopes reads should presumably still apply to it. The question is which operation the **returned** row should be redacted under, and whether the pre-read needs to run its scope under `findFirst` while redacting under `delete`. Worth settling deliberately — the comment at `Model.ts:940-943` asserts one answer and the code does the other, so at minimum one of them is wrong today.

nstfkc:
Closes #366.

`Model.$exec` sends a `delete` down a read-first path as soon as there is a relation or a `_count` to read: read the row inside a transaction, delete it, hand the caller **what was read**. The `delete` statement's own row is thrown away. Every other consequence of that split had already been made invisible to the caller — the read is scoped as the delete was, a miss raises `RecordNotFoundError` naming `delete`, and #364 taught the pre-read to carry the `omit`. `context.operation` was the last one still showing.

| call | today | after |
| --- | --- | --- |
| `delete({ where })` | redacted under `"delete"` | unchanged — one statement, no pre-read |
| `delete({ where, include })` | **redacted under `"findFirst"`** | `"delete"` |
| `delete({ where, include: { _count } })` | **redacted under `"findFirst"`** | `"delete"` |
| a nested read under either | `"findMany"` / `"findFirst"` | unchanged |

So a policy carrying `redact: (ctx, row) => { if (ctx.operation === "delete") row.secret = null }` protected the row until somebody added an `include` to the call, at which point the column came back. The plain `delete` beside it was always right, which is what makes this a divergence *within one operation* rather than an open question about what `delete` should mean.

## Measured

There is no Prisma oracle here — policies are gemi's, and the differential harness deliberately compares against the **pre-redaction** payload. The authority is internal consistency, so the evidence is the live-DB template suite rather than a generated client. Instrumenting the policy to record every `context.operation` it is handed, on the fixture in `delete-redact-operation.test.ts`:

```
delete({ where })                      -> { secret: null, label: "kept" }   CTXOP: delete
delete({ where, include: { children } })  -> { secret: "hunter2", … }       CTXOP: findFirst, delete
delete({ where, include: { _count } })    -> { secret: "hunter2", … }       CTXOP: findFirst, delete
```

Two invocations, and **the one that redacts the returned row is the first**. The second is the inner `delete` statement, running under `"delete"` correctly and having its row discarded — which is why nothing observable came of that half, and why the divergence is silent.

## The fix

`markRedactedAs` / `redactedAs` in `policy.ts`: a module-private `Symbol` on the options object, the third of its kind after `PRE_SCOPED` and `ORM_AUTHORED` and shut the same way — not exported from `orm/index.ts`, so an application cannot forge one and cannot make a query claim to be an operation it is not. The read-first branch marks its pre-read with `op` (not the literal `"delete"`; the branch is guarded on `op`, and taking it from the variable keeps the two from drifting if the guard ever admits a second read-before-write operation).

**Read only on a pre-scoped call**, and that guard is the whole reason this is a two-line change rather than a design question:

```ts
policy = policyContext(
  schema.name,
  (preScoped ? redactedAs(options) : undefined) ?? op,
  …
);
```

The issue asked whether the pre-read "needs to run its scope under `findFirst` while redacting under `delete`". It turns out there is nothing to arbitrate: the pre-read is `markPreScoped`, so `applyPolicies` **does not run on it at all** — no `before`, no `scope`, no `onCreate`/`onUpdate` sees this value. The `where` the `findFirst` executes is the delete's, predicate for predicate, rewritten once at the outer call. The only remaining reader of the context is `applyRedaction`, over the one row about to be returned. The `preScoped` guard makes that structural instead of merely true today: without it, a future non-pre-scoped caller could hand a `scope` an operation the statement is not, and `SCOPABLE` is keyed on exactly that.

That is also pinned rather than asserted. `delete-redact-operation.test.ts` carries a policy whose `scope` fires only for `findFirst`; a real `findFirst` returns `null`, and the `delete`'s pre-read is untouched by it.

## The comment that contradicted the code

The issue is right that `Model.ts:940-943` asserted one answer while the code did the other, and the comment was wrong in **both** halves rather than one:

> A policy that varies by `context.operation` therefore sees "delete" for the children of a row being deleted…

- The **root** row saw `findFirst`. That is #366 and it is fixed.
- The **children** never saw `delete` either. `applyNestedPolicies` scopes every nested node as `NESTED_READ`, a constant that is `findMany` — deliberately a constant, per its own docblock, because threading the root operation down is what produced an earlier bug. Measured: a batched child's `redact` is handed `findMany`; a folded (lateral, Postgres) child is handed `findFirst`, because `redactFolded` passes the enclosing read's name.

The fix does **not** extend the marker to the children, and the comment now says what is true. A nested read is a read of another model whatever statement encloses it, and telling a child's policy `"delete"` would deepen the batched/lateral disagreement rather than settle it.

**Loose end, deliberately left:** those two read names still differ by strategy — `findMany` batched, `findFirst` folded — for the *same* query on the *same* rows. That predates this change, is two reads disagreeing rather than a read reporting a write, and belongs to `redactFolded` rather than to #366. Both values are measured and written down at the test that skirts them; it asserts only that a child never reports the enclosing write, so it holds on either dialect.

## Pinned

`templates/saas-starter/app/models/delete-redact-operation.test.ts`, nine cases per dialect, on a `DelOpParent`/`DelOpChild` fixture of its own — the sibling `delete-include.test.ts` models carry no policies and giving them one would change every case in that file.

The policy branches on two operations in one hook, which is what makes the assertion about the *operation* rather than about redaction happening at all: on the returned row `secret` must be gone and `label` must **not** be, and before this change it was exactly the other way round. A test that only checked "`delete` redacts" would pass on an implementation that redacts unconditionally.

With the `Model.ts`/`policy.ts` hunks stashed, three of the nine fail together — the `include` case, the `_count` case, and the operation list, which reports `["findFirst", "delete"]` where it should report `["delete", "delete"]`.

`policy.test.ts` gets the marker's own contract, which is checkable with no database: it round-trips, it leaves the `strategy` the caller asked for alone, it is absent from unmarked options, and it cannot be spelled from outside the module — neither by a string key nor by a symbol of the caller's own carrying the same description.

## Verification

```
packages/gemi            bun --bun vitest run
  Test Files  120 passed (120)
       Tests  3115 passed | 1 skipped (3116)

packages/gemi            bun run typecheck                       clean
packages/gemi            bunx tsc --noEmit -p tsconfig.generated.json   clean
repo                     bun run lint      1 successful, 0 errors (79 pre-existing warnings)

templates/saas-starter   bun --bun vitest run app/models          (sqlite)
  Test Files  21 passed | 3 skipped (24)
       Tests  801 passed | 134 skipped (935)

templates/saas-starter   app/models/postgres.sh app/models/       (postgres 16, docker)
  Test Files  2 failed | 22 passed (24)
       Tests  1192 passed | 487 skipped (1679)
```

The two Postgres-run failures are `differential.test.ts` and `writes.differential.test.ts` refusing their **sqlite** halves by name — "This suite needs @prisma/client generated for the 'sqlite' provider… Only one dialect's suite can run at a time, which is why the other one is failing rather than skipping." That is the harness's own designed behaviour, not a result. Both files pass in the SQLite run above.

A worktree needs `bun install`, `bun run build:bin` in `packages/gemi`, the `gemi-orm-generator` link, and `bunx prisma generate --schema templates/saas-starter/prisma/differential.prisma` **from the repository root** before any of this is honest; without them `generated.test.ts` fails and the three differential files do not collect at all.

### DIFF
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -828,8 +828,12 @@ export abstract class Model {
       // `applyRedaction`. Reading the marker unconditionally would leave a
       // shape — a future non-pre-scoped caller — where a `before` or a `scope`
       // could be handed an operation the statement is not, and `SCOPABLE` is
-      // keyed on exactly that. The guard makes that shape unrepresentable
-      // rather than merely absent.
+      // keyed on exactly that. The guard does not make that shape
+      // unconstructible — `markRedactedAs(options, "delete")` on its own is
+      // still a value anyone can build — it makes it *unread*, so the
+      // consequence is absent rather than the shape. That is the whole of what
+      // is claimed, and it is enough: the marker changes nothing on any call
+      // that policies still rewrite.
       policy = policyContext(
         schema.name,
         (preScoped ? redactedAs(options) : undefined) ?? op,
@@ -976,11 +980,29 @@ export abstract class Model {
     //   `"delete"`, and its row is thrown away, so nothing observable came of
     //   that half.
     //
-    //   The marker covers the returned row and nothing else. Nested reads keep
-    //   seeing `findMany` — `NESTED_READ` in `policy.ts` is a constant on
-    //   purpose, because a read of another model is a read whatever statement
-    //   encloses it, and that is also what a batched relation read has always
-    //   reported here.
+    //   The marker covers the returned row and nothing else. Nested reads are
+    //   *scoped* as `findMany` on both strategies — `NESTED_READ` in `policy.ts`
+    //   is a constant on purpose, because a read of another model is a read
+    //   whatever statement encloses it — and the marker does not reach them: the
+    //   relation executor rebuilds its options from `{ strategy }` rather than
+    //   forwarding these.
+    //
+    //   Which name a nested row is *redacted* as is a second mechanism and it
+    //   does not currently agree with the first. A **batched** child runs its
+    //   own `$exec` under `NESTED_READ`, so it is redacted as `findMany`; a
+    //   **folded** (lateral, the Postgres default) child never enters its own
+    //   `$exec`, so `redactFolded` below redacts it on the parent's behalf and
+    //   builds its context from the enclosing `$exec`'s `op` — which on this
+    //   call is the pre-read, a `findFirst`. Measured, same query, same rows:
+    //
+    //     sqlite   (batched) child redact context.operation -> "findMany"
+    //     postgres (folded)  child redact context.operation -> "findFirst"
+    //
+    //   Pre-existing, not widened here, and the same on an ordinary `findFirst`
+    //   with an `include`. Filed as #388; the fix is to hand `redactFolded`
+    //   `NESTED_READ` too. Neither value is the write, which is the part #366
+    //   is about and the only part `delete-redact-operation.test.ts` pins under
+    //   the default strategy.
     //
     // Only when there is something to read. A plain `delete` still compiles to
     // one statement and opens no transaction.
--- a/packages/gemi/orm/Model.ts
+++ b/packages/gemi/orm/Model.ts
@@ -46,9 +46,11 @@ import {
   isPreScoped,
   markOrmAuthored,
   markPreScoped,
+  markRedactedAs,
   ormAuthoredFields,
   policiesFor,
   policyContext,
+  redactedAs,
   type PolicyEntry,
   type PolicyContext,
 } from "./policy";
@@ -812,7 +814,28 @@ export abstract class Model {
       // policy that never consults the user — a soft-delete scope, say — has
       // nothing to deny and must keep working with no request in sight. See
       // `policyContext`.
-      policy = policyContext(schema.name, op, currentUser(), system);
+      //
+      // `redactedAs` is the one thing that can make `context.operation` differ
+      // from `op`, and it exists to *remove* a difference the caller can see: a
+      // `delete` that has to read before it writes runs its pre-read as a
+      // `findFirst` and hands that row back, so without this a `redact` keyed on
+      // `"delete"` stopped firing the moment the call also carried an `include`
+      // (#366). The row is the delete's return value; it is redacted as one.
+      //
+      // **Only on a pre-scoped call, which is what keeps it a redaction
+      // concern.** `applyPolicies` below is skipped when `preScoped`, so on the
+      // one call that sets this marker the context reaches nothing but
+      // `applyRedaction`. Reading the marker unconditionally would leave a
+      // shape — a future non-pre-scoped caller — where a `before` or a `scope`
+      // could be handed an operation the statement is not, and `SCOPABLE` is
+      // keyed on exactly that. The guard makes that shape unrepresentable
+      // rather than merely absent.
+      policy = policyContext(
+        schema.name,
+        (preScoped ? redactedAs(options) : undefined) ?? op,
+        currentUser(),
+        system,
+      );
 
       // **The context is built even when pre-scoped; only the rewrite is
       // skipped.** `preScoped` means "the scope is already in these args" — not
@@ -939,9 +962,25 @@ export abstract class Model {
     //   model's policies once already; re-applying them would `AND` the same
     //   predicate twice — same rows, different plan key.
     // - **The read is scoped as the delete was**, including its nested policies.
-    //   A policy that varies by `context.operation` therefore sees "delete" for
-    //   the children of a row being deleted, which is the conservative reading:
-    //   these are the rows the delete is about to remove.
+    //   `effective` was rewritten by `applyPolicies` and `applyNestedPolicies`
+    //   under operation `"delete"`, and `markPreScoped` stops the pre-read
+    //   redoing either — so the `where` the `findFirst` runs is the delete's,
+    //   predicate for predicate.
+    // - **`markRedactedAs`, so the row is redacted as the delete it belongs
+    //   to.** This bullet used to claim `context.operation` was already
+    //   `"delete"` here. It was not, and the row it was most wrong about was the
+    //   root one — the one returned. The pre-read is a `findFirst`, so
+    //   `policyContext` was built with `"findFirst"` and a `redact` keyed on
+    //   `"delete"` silently stopped firing as soon as the same call carried an
+    //   `include` (#366). The `delete` statement below *does* run under
+    //   `"delete"`, and its row is thrown away, so nothing observable came of
+    //   that half.
+    //
+    //   The marker covers the returned row and nothing else. Nested reads keep
+    //   seeing `findMany` — `NESTED_READ` in `policy.ts` is a constant on
+    //   purpose, because a read of another model is a read whatever statement
+    //   encloses it, and that is also what a batched relation read has always
+    //   reported here.
     //
     // Only when there is something to read. A plain `delete` still compiles to
     // one statement and opens no transaction.
@@ -998,7 +1037,14 @@ export abstract class Model {
         const before = await this.$exec(
           "findFirst",
           { where, ...projection },
-          markPreScoped({ strategy: options?.strategy }) as never,
+          // `op` rather than the literal `"delete"`: the branch is guarded on
+          // it, so they are the same string, and taking it from the variable
+          // keeps the two from drifting if the guard ever admits a second
+          // operation that has to read before it writes.
+          markRedactedAs(
+            markPreScoped({ strategy: options?.strategy }),
+            op,
+          ) as never,
         );
 
         // `findFirst` shapes a miss to `null`; the caller asked for a delete, so
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -12,9 +12,11 @@ import {
   applyNestedPolicies,
   applyPolicies,
   applyRedaction,
+  markRedactedAs,
   policiesFor,
   policyContext,
   redactNullable,
+  redactedAs,
   type ModelPolicy,
   type PolicyContext,
 } from "./policy";
@@ -1903,5 +1905,57 @@ describe("a malformed $policies entry", () => {
     Owner.$policies = [{ scopes: () => ({}) }] as unknown as ModelPolicy[];
 
     expect(() => policiesFor(Owner)).toThrow(InvalidPolicyEntryError);
+  });
+});
+
+/**
+ * The marker that lets a read-first `delete` say which operation the row it is
+ * about to return belongs to — #366.
+ *
+ * The behaviour it produces needs a database, and lives in
+ * `templates/saas-starter/app/models/delete-redact-operation.test.ts`. What is
+ * checkable here is the property that makes it safe to have at all: the key is a
+ * module-private `Symbol`, so it cannot be spelled from outside this module and
+ * cannot become a way for an application to make a query claim to be an
+ * operation it is not — the same door `PRE_SCOPED` and `ORM_AUTHORED` are shut
+ * the same way.
+ */
+describe("the operation a returned row is redacted as", () => {
+  test("survives being carried on an options object", () => {
+    expect(redactedAs(markRedactedAs({ strategy: "batched" }, "delete"))).toBe(
+      "delete",
+    );
+  });
+
+  test("does not disturb the options it rides on", () => {
+    const marked = markRedactedAs({ strategy: "batched" }, "delete") as {
+      strategy?: string;
+    };
+
+    expect(marked.strategy).toBe("batched");
+    // Named keys only. `toEqual` compares symbol keys too, and the point here is
+    // that the strategy the caller asked for still reaches the pre-read — which
+    // is what carries a `{ strategy: "lateral" }` down from `$exec`'s options.
+    expect(Object.keys(marked)).toEqual(["strategy"]);
+  });
+
+  test("is absent from options that were never marked", () => {
+    expect(redactedAs({ strategy: "batched" })).toBeUndefined();
+    expect(redactedAs(undefined)).toBeUndefined();
+    expect(redactedAs(null)).toBeUndefined();
+  });
+
+  /**
+   * The forgery attempt, written out rather than argued: a string key is not the
+   * key, and neither is a symbol of the caller's own carrying the same
+   * description — `Symbol()` mints a fresh identity every call.
+   */
+  test("cannot be spelled from outside the module", () => {
+    expect(redactedAs({ redactedAs: "delete" })).toBeUndefined();
+    expect(
+      redactedAs({
+        [Symbol("gemi.orm.operationTheRowIsReturnedFor")]: "delete",
+      }),
+    ).toBeUndefined();
   });
 });
--- a/packages/gemi/orm/policy.test.ts
+++ b/packages/gemi/orm/policy.test.ts
@@ -12,6 +12,8 @@ import {
   applyNestedPolicies,
   applyPolicies,
   applyRedaction,
+  isPreScoped,
+  markPreScoped,
   markRedactedAs,
   policiesFor,
   policyContext,
@@ -1937,6 +1939,29 @@ describe("the operation a returned row is redacted as", () => {
     // that the strategy the caller asked for still reaches the pre-read — which
     // is what carries a `{ strategy: "lateral" }` down from `$exec`'s options.
     expect(Object.keys(marked)).toEqual(["strategy"]);
+  });
+
+  /**
+   * The property the one call site actually rests on, which the `strategy` case
+   * above does *not* cover: the spread inside `markRedactedAs` has to carry the
+   * `PRE_SCOPED` **symbol** it is wrapping.
+   *
+   * `markRedactedAs(markPreScoped({ strategy }), op)` is the only construction
+   * in the codebase, and it is load-bearing twice over. If that spread ever
+   * stopped carrying symbol keys the pre-read would be scoped a second time —
+   * and `redactedAs` is read only when `preScoped`, so the marker would go
+   * unread as well, silently restoring #366. Both failures are caught today, but
+   * only by the live-database suite and only as a `RecordNotFoundError` a long
+   * way from its cause. One line pins it where the mechanism is.
+   */
+  test("carries the pre-scoped marker it wraps", () => {
+    const marked = markRedactedAs(
+      markPreScoped({ strategy: "batched" }),
+      "delete",
+    );
+
+    expect(isPreScoped(marked)).toBe(true);
+    expect(redactedAs(marked)).toBe("delete");
   });
 
   test("is absent from options that were never marked", () => {
--- a/packages/gemi/orm/policy.ts
+++ b/packages/gemi/orm/policy.ts
@@ -1449,6 +1449,52 @@ export function ormAuthoredFields(options: unknown): readonly string[] {
 
 const EMPTY_FIELDS: readonly string[] = [];
 
+/**
+ * The operation whose *return value* this `$exec` is producing, when that is not
+ * the operation it runs.
+ *
+ * There is exactly one such call: the pre-read a read-first `delete` performs.
+ * `delete({ where, include })` cannot be one statement — the children have to be
+ * read before the row goes away — so `$exec` runs a `findFirst` inside a
+ * transaction and hands **that** row to the caller, discarding what the `delete`
+ * statement returned. Every other consequence of the split was already made
+ * invisible: the read is scoped as the delete was, a miss raises
+ * `RecordNotFoundError` naming `delete`, and #364 taught it to carry the `omit`.
+ * `context.operation` was the last one left visible, and it was visible in the
+ * worst direction — a `redact` keyed on `"delete"` stopped firing on the row the
+ * caller got back, the moment an `include` was added (#366).
+ *
+ * **Redaction only, which is why it is named for redaction.** By the time this
+ * is read the pre-read is already `markPreScoped`, so `applyPolicies` does not
+ * run on it at all — no `before`, no `scope`, no `onCreate` / `onUpdate` sees
+ * this value, and a scope written for reads cannot be turned off by it. The
+ * remaining reader of the context is `applyRedaction`, over the one row that is
+ * about to be returned.
+ *
+ * **Not the nested reads.** A relation read underneath is scoped and redacted as
+ * `findMany` — see {@link NESTED_READ}, which is a constant for the reason this
+ * is narrow: a nested read is a read of another model, whatever statement
+ * encloses it. This marker travels no further than the row it names.
+ *
+ * A module-private `Symbol`, for the same reason `PRE_SCOPED` and `ORM_AUTHORED`
+ * are: it is not exported from `orm/index.ts`, so an application cannot forge it
+ * and cannot use it to make a query claim to be an operation it is not.
+ */
+const REDACTED_AS = Symbol("gemi.orm.operationTheRowIsReturnedFor");
+
+export function markRedactedAs(
+  options: object | undefined,
+  operation: Operation,
+): object {
+  return { ...options, [REDACTED_AS]: operation };
+}
+
+export function redactedAs(options: unknown): Operation | undefined {
+  if (typeof options !== "object" || options === null) return undefined;
+  const operation = (options as Record<symbol, unknown>)[REDACTED_AS];
+  return typeof operation === "string" ? (operation as Operation) : undefined;
+}
+
 /** Resolves a model name to its policies, so this file need not import the registry. */
 export type PolicyLookup = (model: string) => {
   policies: readonly ModelPolicy[];
--- a/packages/gemi/orm/policy.ts
+++ b/packages/gemi/orm/policy.ts
@@ -1471,10 +1471,33 @@ const EMPTY_FIELDS: readonly string[] = [];
  * remaining reader of the context is `applyRedaction`, over the one row that is
  * about to be returned.
  *
- * **Not the nested reads.** A relation read underneath is scoped and redacted as
- * `findMany` — see {@link NESTED_READ}, which is a constant for the reason this
- * is narrow: a nested read is a read of another model, whatever statement
- * encloses it. This marker travels no further than the row it names.
+ * **Not the nested reads.** A relation read underneath is *scoped* as `findMany`
+ * — see {@link NESTED_READ}, which is a constant for the reason this marker is
+ * narrow: a nested read is a read of another model, whatever statement encloses
+ * it. This marker travels no further than the row it names, and the relation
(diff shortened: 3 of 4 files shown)


packages/gemi/orm/Model.ts
```diff
@@ -939,9 +962,25 @@ export abstract class Model {
...
+    //   `"delete"` here. It was not, and the row it was most wrong about was the
+    //   root one — the one returned. The pre-read is a `findFirst`, so
+    //   `policyContext` was built with `"findFirst"` and a `redact` keyed on
+    //   `"delete"` silently stopped firing as soon as the same call carried an
+    //   `include` (#366). The `delete` statement below *does* run under
+    //   `"delete"`, and its row is thrown away, so nothing observable came of
+    //   that half.
+    //
+    //   The marker covers the returned row and nothing else. Nested reads keep
+    //   seeing `findMany` — `NESTED_READ` in `policy.ts` is a constant on
```
nstfkc: **should fix** — this replacement comment is wrong in the same way the one it replaces was.

`Nested reads keep seeing findMany` is unqualified, and it is false on Postgres. `redactFolded` (`Model.ts:1539`) is handed the enclosing `$exec`'s `op`, not `NESTED_READ`, and the enclosing `$exec` here is the pre-read — a `findFirst`. Measured on this PR's own fixture with the child's `redact` instrumented, same query, same rows:

```
sqlite   (batched) childOperations: ["findMany","findMany"]
postgres (lateral) childOperations: ["findFirst","findFirst"]
```

The trailing clause (`that is also what a batched relation read has always reported here`) is technically the escape hatch, but it reads as reinforcement rather than as the qualification it actually is — a reader takes away "findMany", full stop. The claim about *scoping* is fine on both dialects (`applyNestedPolicies` does pass `NESTED_READ`); it is redaction that diverges, and the two are run together in one sentence.

Suggest saying what is true: a nested read is scoped as `findMany` on both strategies, and redacted as `findMany` when batched / as the enclosing read's name when folded — with `redactFolded` named as the reason and a note that it is pre-existing. The PR body and `delete-redact-operation.test.ts` already word it that way; this is the file where the batch already burned #374 on a stale comment.

packages/gemi/orm/policy.ts
```diff
@@ -1449,6 +1449,52 @@ export function ormAuthoredFields(options: unknown): readonly string[] {
...
+ * caller got back, the moment an `include` was added (#366).
+ *
+ * **Redaction only, which is why it is named for redaction.** By the time this
+ * is read the pre-read is already `markPreScoped`, so `applyPolicies` does not
+ * run on it at all — no `before`, no `scope`, no `onCreate` / `onUpdate` sees
+ * this value, and a scope written for reads cannot be turned off by it. The
+ * remaining reader of the context is `applyRedaction`, over the one row that is
+ * about to be returned.
+ *
+ * **Not the nested reads.** A relation read underneath is scoped and redacted as
```
nstfkc: **should fix** — same claim, second location, and here the two mechanisms are collapsed into one verb.

`scoped and redacted as findMany` is two statements with two different truth values:

- **scoped** as `findMany`: true on both strategies — `applyNestedPolicies` passes `NESTED_READ`.
- **redacted** as `findMany`: true only when the child went through its own `$exec` (batched). A folded child never does; `redactFolded` builds its context from the enclosing `op`, so under the default Postgres strategy the child of *this very pre-read* is redacted as `findFirst`. Measured on postgres:16 in docker: `childOperations: ["findFirst","findFirst"]`.

Worth being precise about here in particular, because this docblock is the contract for the marker — someone later deciding whether it is safe to widen `REDACTED_AS` will read this paragraph as the reason not to, and the reason is real but is not the one stated.

packages/gemi/orm/Model.ts
```diff
@@ -812,7 +814,28 @@ export abstract class Model {
...
+      // `"delete"` stopped firing the moment the call also carried an `include`
+      // (#366). The row is the delete's return value; it is redacted as one.
+      //
+      // **Only on a pre-scoped call, which is what keeps it a redaction
+      // concern.** `applyPolicies` below is skipped when `preScoped`, so on the
+      // one call that sets this marker the context reaches nothing but
+      // `applyRedaction`. Reading the marker unconditionally would leave a
+      // shape — a future non-pre-scoped caller — where a `before` or a `scope`
+      // could be handed an operation the statement is not, and `SCOPABLE` is
+      // keyed on exactly that. The guard makes that shape unrepresentable
```
nstfkc: **nit** — `unrepresentable` overstates what the guard buys, in a comment whose whole job is precision.

`markRedactedAs(someOptions, "delete")` without `markPreScoped` is still perfectly constructible; the guard makes it *unreadable*, so the consequence is absent rather than the shape. The stronger constructions do exist — `markRedactedAs` could set `PRE_SCOPED` itself, or assert `isPreScoped(options)` — and either would earn the word. As written, "the marker is only honoured on a pre-scoped call, so the bad shape has no effect" is the accurate sentence, and it is still a good argument for keeping the guard.

packages/gemi/orm/policy.test.ts
```diff
@@ -1905,3 +1907,55 @@ describe("a malformed $policies entry", () => {
...
+ * the same way.
+ */
+describe("the operation a returned row is redacted as", () => {
+  test("survives being carried on an options object", () => {
+    expect(redactedAs(markRedactedAs({ strategy: "batched" }, "delete"))).toBe(
+      "delete",
+    );
+  });
+
+  test("does not disturb the options it rides on", () => {
```
nstfkc, line 1932: **nit** — this pins that `strategy` survives, but not the property the call site actually depends on: that the spread in `markRedactedAs` preserves the **`PRE_SCOPED`** symbol it is wrapping.

`markRedactedAs(markPreScoped({ strategy }), op)` is the only call site, and if that spread ever stopped carrying symbol keys the pre-read would be re-scoped — the `label: "no-such-label"` fragment would fire on it and the delete would raise `RecordNotFoundError`. That is caught today, but only by the live-DB suite, and only as a confusing failure a long way from the cause. One line here pins it where the mechanism is:

```ts
expect(isPreScoped(markRedactedAs(markPreScoped({}), "delete"))).toBe(true);
```

`isPreScoped` and `markPreScoped` are both already exported from the module under test.

templates/saas-starter/app/models/delete-redact-operation.test.ts
```diff
@@ -0,0 +1,442 @@
...
+      register("DelOpChild", DelOpChild);
+    }, 120_000);
+
+    afterAll(async () => {
+      await raw?.unsafe(`DROP TABLE IF EXISTS "DelOpChild"`).catch(() => {});
+      await raw?.unsafe(`DROP TABLE IF EXISTS "DelOpParent"`).catch(() => {});
+      await raw?.close();
+      await database?.close();
+      if (previous) Application.setInstance(previous);
+      if (workspace) rmSync(workspace, { recursive: true, force: true });
```
nstfkc: **nit** — `workspace` is set only by the sqlite resolver, but this `afterAll` is shared by both suites, so the Postgres suite also tries to remove the SQLite temp dir.

Harmless today: describes in one file run in order, and `force: true` swallows the second call. But it is cross-suite mutable state in a helper that is otherwise parameterised cleanly, and the same shape would bite if the two suites ever ran in separate workers. Cleaner to let the resolver own its own teardown, or to pass a `cleanup?: () => void` into `suite()` beside `resolve`.

templates/saas-starter/app/models/delete-redact-operation.test.ts
```diff
@@ -0,0 +1,442 @@
...
+        include: { children: true },
+      });
+
+      // Not an equality: the value differs by dialect, because the default
+      // strategy does. Measured — `findMany` on SQLite (batched), `findFirst` on
+      // Postgres (lateral, where `redactFolded` passes the enclosing read's
+      // name). Neither is the write, which is the part that has to hold, and
+      // pinning either literal here would make this suite fail on one dialect
+      // for a divergence that is not #366's.
+      expect(childOperations).not.toContain("delete");
```
nstfkc, line 402: **nit** — this case is written exactly right for the divergence it skirts, and the comment above it is the clearest description of that divergence anywhere in the repo. But the loose end still has no issue number, and this PR leaves it slightly worse than it found it.

Before: parent `findFirst`, folded child `findFirst` — at least internally consistent. After: parent `delete`, folded child `findFirst`, and `findFirst` is now a name that corresponds to nothing the caller asked for and nothing else in the call reports. Confirmed both on postgres:16.

Not this PR's job to fix — I agree with the session file's reasoning for staying inside #366. But `.orm-sessions/issue-366.md` says "somebody should file it" and nobody has, and the `orm.md` note has no number to point at. Worth filing before this merges; the fix looks like a one-liner, giving `redactFolded` `NESTED_READ` rather than the enclosing `op`.

nstfkc (commented):
**Verdict: minor-comments.** The mechanism is right, the fix is minimal, the central test is mutation-checked, and I reproduced every number in the PR body. The findings are all about the *replacement* comments — which assert something I measured to be false on Postgres, in the same way the comment they replace was false.

## What I verified

Checked out `2e255d4` into a clean worktree, `bun install`, `readlink -f templates/saas-starter/node_modules/gemi` -> **this** worktree's `packages/gemi` (not the main checkout).

```
packages/gemi          bun run typecheck                       clean
packages/gemi          bun --bun vitest run                    120 files, 3115 passed | 1 skipped
templates/saas-starter bun --bun vitest run app/models         21 passed | 3 skipped, 801 passed | 134 skipped
templates/saas-starter ./app/models/postgres.sh app/models/    22 passed | 2 failed, 1192 passed | 487 skipped
gh pr checks 383                                               sqlite pass, postgres pass
```

The two Postgres failures are `differential.test.ts` / `writes.differential.test.ts` refusing their **SQLite** halves by name ("Only one dialect's suite can run at a time") — the harness's designed behaviour, both green in the SQLite run. Every number matches the PR body exactly.

**Mutation check, the one that mattered.** `git checkout 54d4278 -- packages/gemi/orm/Model.ts packages/gemi/orm/policy.ts`, then re-ran the new file:

```
FAIL  a delete with an include redacts under delete      expected "hunter2" to be null
FAIL  a delete with a _count redacts under delete        expected "hunter2" to be null
FAIL  neither invocation reports the read it was compiled as
      expected [ 'findFirst', 'delete' ] to deeply equal [ 'delete', 'delete' ]
Tests  3 failed | 6 passed | 1 skipped
```

Exactly the three the author claims, and the third fails with the pre-fix operation list rather than by accident. The `label` half of the two-branch policy is doing real work: it is what stops an unconditional redactor passing.

**Mechanism, read rather than taken on trust.** `policy` is assigned at `Model.ts:833` and read in exactly two places — `applyPolicies` at 854, guarded on `!preScoped`, and `applyRedaction` at 1217. The pre-read is `markPreScoped`, so the first is unreachable on the one call that sets the marker; the guard's claim holds. `redactedAs` cannot leak downward either: the relation executor at `Model.ts:921` rebuilds `base` from `{ strategy }` rather than forwarding `options`, so a nested `$exec` never carries it. The plan key comes from `op`, not from `policy.operation`, so no cached plan is shared across differing contexts. And the branch is guarded on `op === "delete"`, so `markRedactedAs(..., op)` really is the literal — the variable is the right call, not a hedge.

**The other direction, since a redaction fix can also un-redact.** A policy keyed on reads (`if (ctx.operation === "findFirst") ...`) now stops firing on the row a `delete({ where, include })` returns. That is not a new exposure: the plain `delete({ where })` beside it never fired that branch either, so the caller could already reach the same columns with one fewer argument. Bounded correctly.

**Scope** is tight — one helper pair, two lines of behaviour, comments, two test files. No drive-bys.

**What I could not check:** nothing about Prisma parity, and I agree there is nothing to check — policies are gemi's and `differential.ts` compares pre-redaction payloads. The PR says so plainly instead of implying a parity it does not have, which is the right call. I also did not audit the scalar-list Postgres suites' skip counts.

## Findings

Six, none blocking. Two **should fix** (the same wrong claim, in two files), four **nit**.

The one that matters: the comment being replaced was wrong, and the replacement asserts `Nested reads keep seeing findMany` without qualification. I instrumented the child policy and ran the PR's own fixture on both dialects:

```
sqlite   (batched) childOperations: ["findMany","findMany"]   parentOperations: ["delete","delete"]
postgres (lateral) childOperations: ["findFirst","findFirst"] parentOperations: ["delete","delete"]
```

True on SQLite, false on Postgres — the same shape of defect as the one this PR fixes, and #374 in this batch was exactly a stale comment. The PR body and the test file both state the divergence honestly; the two comments in the source do not, and the source comment is the one a future reader hits first. Details inline.

packages/gemi/orm/Model.ts
nstfkc: **Addressed** — 9542cc6.

You are right, and I re-measured rather than taking it on the reading. Forcing both strategies on the same fixture, with the child policy instrumented:

```
lateral  delete + include    parent ["delete","delete"]  child ["findFirst"]
lateral  findFirst + include parent ["findFirst"]        child ["findFirst"]
batched  delete + include    parent ["delete","delete"]  child ["findMany"]
```

postgres:16 in docker. (Worth knowing for anyone reproducing: `{ strategy: "lateral" }` on SQLite does **not** reproduce it — the lateral strategy declines per node for everything it cannot fold, and on SQLite that is everything, so the child still runs its own `$exec` and reports `findMany`. It needs a real Postgres.)

The comment now separates the two mechanisms explicitly, gives both measured values, names `redactFolded` as the reason, says it is pre-existing and the same on an ordinary `findFirst` with an `include`, and points at the issue. Filed as **#388**, so the loose end has a number now — that is your last comment on this PR, answered here too.

packages/gemi/orm/policy.ts
nstfkc: **Addressed** — 9542cc6.

Agreed, and your framing of *why it matters here specifically* is the part I had missed: this docblock is the contract, so it is what someone reads when deciding whether to widen `REDACTED_AS`, and the stated reason was not the real one.

The paragraph is now split. Scoping is `NESTED_READ` on both strategies; redaction is `findMany` only when the child ran its own `$exec`, and a folded child never does. The measured `findFirst` is in the docblock with the dialect named.

I also made it carry the argument it should have carried all along: the divergence is the reason **not** to widen this marker to nested rows, because doing so would hand a child's policy the enclosing *write*'s name on one strategy and not the other — strictly worse than two read names disagreeing. Filed as #388 and cited.

While I was in there I added the other half of "travels no further": the relation executor rebuilds its options from `{ strategy }` rather than forwarding these, so the marker cannot reach a nested `$exec` even by accident. You verified that in your summary; it belongs in the docblock rather than only in a review.

packages/gemi/orm/Model.ts
nstfkc: **Addressed** — 9542cc6, by rewording rather than by strengthening the mechanism.

You are right that `unrepresentable` is the wrong word: `markRedactedAs(options, "delete")` on its own is a value anyone can build. The guard makes it *unread*, so the consequence is absent, not the shape. The comment now says exactly that and no more.

I considered your two stronger constructions and declined both, for the same reason the marker is shaped the way it is:

- **`markRedactedAs` setting `PRE_SCOPED` itself** conflates two independent facts. `PRE_SCOPED` means "the args already carry this model's scope"; `REDACTED_AS` means "the row this returns belongs to another operation". They are true together at the one call site, but only by coincidence of that call site, and fusing them would make a future caller unable to state one without the other.
- **`markRedactedAs` asserting `isPreScoped(options)`** buys a throw at a construction that already has no effect, on a path that runs per query.

What I did instead is pin the coupling in the test — see the reply to your `policy.test.ts` comment. That catches the failure mode that actually costs something (the spread dropping the symbol) at the cost of one assertion, where an invariant check would have caught a shape that does nothing.

packages/gemi/orm/policy.test.ts
nstfkc: **Addressed** — 9542cc6. Good catch; this is the one finding that changes what the suite can detect rather than what a comment says.

Added as `carries the pre-scoped marker it wraps`, asserting both directions of the wrap:

```ts
const marked = markRedactedAs(markPreScoped({ strategy: "batched" }), "delete");

expect(isPreScoped(marked)).toBe(true);
expect(redactedAs(marked)).toBe("delete");
```

Your account of the failure understates it slightly, which is why I wrote the docblock around both halves: if the spread stopped carrying symbol keys the pre-read would be scoped a second time **and** `redactedAs` would go unread, because it is only consulted when `preScoped`. So the same mutation both raises `RecordNotFoundError` from the live suite and silently restores #366.

Mutation-checked. Replacing the spread with `{ ...Object.fromEntries(Object.entries(options ?? {})), [REDACTED_AS]: operation }`:

```
FAIL  orm/policy.test.ts > carries the pre-scoped marker it wraps
      expected false to be true
```

and the rest of the 3117-test package suite stays green under that mutation — which is the point: nothing else in `packages/gemi` was watching this.

templates/saas-starter/app/models/delete-redact-operation.test.ts
nstfkc: **Addressed** — 9542cc6, with your second suggestion (the resolver owning its teardown), since it removes the module-level binding entirely rather than moving it.

`resolve` now returns `{ url, cleanup? }`:

```ts
suite("… — sqlite", () => {
  const workspace = mkdtempSync(join(tmpdir(), "gemi-orm-delete-redact-"));
  return {
    url: `sqlite://${join(workspace, "delete-redact.db")}`,
    cleanup: () => rmSync(workspace, { recursive: true, force: true }),
  };
}, SQLITE_DDL);
```

`workspace` is now a local of the resolver, so the Postgres suite has no way to reach it — it returns no `cleanup` at all, and the tables are dropped by the shared `afterAll` either way. The `describe`-level `cleanup?.()` replaces the `if (workspace)`.

You are right that it was harmless today and right that it would not stay harmless: `force: true` was doing the work, and the shape only survives while both suites share one worker.

templates/saas-starter/app/models/delete-redact-operation.test.ts
nstfkc: **Addressed** — filed as **#388**, and cited from the four places that describe it.

https://github.com/nstfkc/gemi/issues/388 — "A folded child is redacted as the enclosing read, where a batched one is redacted as findMany". It carries the measurements, the mechanism (`redactFolded` at `Model.ts:1539`, called at `Model.ts:1201` with the enclosing `op` rather than `NESTED_READ`), the one-line fix you name, and the note that the pin belongs in the template suite as an equality *across strategies* rather than a per-dialect literal — which is the assertion the divergence would have failed.

Your point that this PR leaves it slightly worse is fair and I put it in the issue verbatim: before, parent and folded child both said `findFirst` and were at least internally consistent; after, the parent says `delete` and the folded child says `findFirst`, a name nothing else in the call reports. I still think that is the right trade — the alternative is leaving the returned row redacted as a read, which is #366 — but it is a real cost and it should be written down where the fix is, not only here.

The issue also records that the divergence is **not** specific to `delete`: a plain `findFirst({ include })` shows it too. Measured on postgres:16:

```
lateral  findFirst + include   child ["findFirst"]
batched  findFirst + include   child ["findMany"]
```

That widens the case for fixing it, since it means a policy keyed on `findMany` protects a child on SQLite and leaks it on Postgres for identical application code, with no `delete` anywhere in sight.

The two source comments and the two test cases that skirt the divergence now name #388, so the `orm.md` note has a number to point at.

nstfkc:
## Review round 1 — 9542cc6

Six comments, **six addressed, none declined, none deferred**. Every one is answered inline as well.

The verdict was right about where the weakness was: the fix and its central test are unchanged, and all six findings were about the *replacement* comments plus one missing assertion.

### The two `should fix` findings — the same claim, two files

`Nested reads keep seeing findMany` (`Model.ts`) and `scoped and redacted as findMany` (`policy.ts`) were unqualified, and they ran two mechanisms with two different truth values through one verb. **Scoping** is `NESTED_READ` on both strategies. **Redaction** is `findMany` only when the child ran its own `$exec`; a folded child never does, because `redactFolded` builds its context from the enclosing `$exec`'s `op`.

Re-measured rather than taken on the reading, forcing both strategies on this PR's own fixture with the child policy instrumented:

```
lateral  delete + include     parent ["delete","delete"]  child ["findFirst"]
lateral  findFirst + include  parent ["findFirst"]        child ["findFirst"]
batched  delete + include     parent ["delete","delete"]  child ["findMany"]
```

postgres:16 in docker. One thing worth writing down for anyone reproducing it: **`{ strategy: "lateral" }` on SQLite does not reproduce this.** The lateral strategy declines per node for anything it cannot fold, and on SQLite that is everything, so the child still goes through its own `$exec` and reports `findMany`. It needs a real Postgres — I lost a run to that.

Both comments now state the two mechanisms separately, carry both measured values with the dialect named, point at `redactFolded`, and say the divergence is pre-existing and equally true of an ordinary `findFirst`. The `policy.ts` docblock also now carries the argument it should have had all along: this divergence is the reason `REDACTED_AS` must **not** be widened to nested rows, since doing so would hand a child's policy the enclosing *write*'s name on one strategy and not the other — strictly worse than two read names disagreeing.

### The loose end now has a number: #388

Filed as #388 — mechanism (`redactFolded` at `Model.ts:1539`, called at `Model.ts:1201` with the enclosing `op` rather than `NESTED_READ`), the measurements, the one-line fix, and the note that the pin belongs in the template suite as an equality *across strategies* rather than a per-dialect literal, which is the assertion the divergence would have failed. Cited from both source comments and both test cases that skirt it, so the `orm.md` note has something to point at.

The review's fair objection — that this PR leaves the divergence slightly worse, parent `delete` beside a folded child `findFirst` where the two used to agree — is recorded in the issue rather than only in a review thread. The issue also records that it is not specific to `delete`: a plain `findFirst({ include })` shows it too, which means a policy keyed on `findMany` protects a child on SQLite and leaks it on Postgres for identical application code, with no `delete` anywhere in sight.

### The four nits

- **`unrepresentable`** overstated the `preScoped` guard. `markRedactedAs(options, "delete")` on its own is still constructible; the guard makes it *unread*, so the consequence is absent rather than the shape. Reworded to claim only that. I declined the two stronger constructions the comment offered, and said so on the thread: `markRedactedAs` setting `PRE_SCOPED` itself conflates two independent facts, and asserting `isPreScoped` buys a throw at a construction that already has no effect, on a per-query path. What I did instead is pin the coupling in the test, below.
- **`policy.test.ts` was not pinning what the call site rests on.** Added `carries the pre-scoped marker it wraps`. The failure mode is worse than the comment said: if that spread stopped carrying symbol keys the pre-read would be scoped a second time **and** `redactedAs` would go unread, because it is only consulted when `preScoped` — so one mutation both raises `RecordNotFoundError` from the live suite and silently restores #366.
- **Cross-suite `workspace`** in the template file: `resolve` now returns `{ url, cleanup? }` and the SQLite resolver owns its temp dir as a local. The Postgres suite returns no `cleanup` and has no way to reach the other's.
- **The unfiled loose end**: #388, above.

### Verification

Mutation checks re-run *after* the changes, not assumed to survive them.

New unit test — replace the spread in `markRedactedAs` with `{ ...Object.fromEntries(Object.entries(options ?? {})), [REDACTED_AS]: operation }`:

```
FAIL  orm/policy.test.ts > carries the pre-scoped marker it wraps   expected false to be true
```

and the other 3116 tests in the package stay green under that mutation, which is the point — nothing else was watching it.

Central test — `git checkout 54d4278 -- packages/gemi/orm/Model.ts packages/gemi/orm/policy.ts`, then the template file:

```
Tests  3 failed | 6 passed | 1 skipped (10)
  x a delete with an include redacts under delete
  x a delete with a _count redacts under delete
  x neither invocation reports the read it was compiled as
```

Still exactly three, so the `resolve` refactor did not weaken it. Restored, re-run green.

```
packages/gemi            bun run typecheck                     clean
packages/gemi            bun --bun vitest run                  120 files, 3116 passed | 1 skipped (3117)
templates/saas-starter   bun --bun vitest run app/models       21 passed | 3 skipped, 801 passed | 134 skipped
templates/saas-starter   ./app/models/postgres.sh app/models/  22 passed | 2 failed, 1192 passed | 487 skipped
repo                     oxlint <the four changed files>       0 warnings, 0 errors
```

Same two Postgres "failures" as before — `differential.test.ts` and `writes.differential.test.ts` refusing their **SQLite** halves by name, the harness's designed behaviour. Every other number is identical to the pre-review run except the package test count, which is +1 for the new assertion.

`prettier --check` still warns on `policy.ts` and `policy.test.ts`; it warns identically on `main`, and diffing prettier's output shows every hunk it wants is on lines this PR does not touch.

### Closing keyword

Checked, per the batch instruction. Already correct — `Closes #366.` is the first line of the body, and `gh pr view 383 --json closingIssuesReferences` returns issue 366. Nothing to fix here; #381 and #386 are the two that need it.

PR #384, opened 2026-08-09
Close #268: point the plan at main, and make the branch audit runnable

### LINKED ISSUE #268: Land the ORM stack: PR #45 needs a new base, and 117 of 120 ORM branches hold nothing unique

nstfkc:
#268 is two halves that have moved apart since it was written. The landing half is **done** and the plan document does not know it. The hygiene half is **not** something a commit can discharge — deleting 123 remote branches is irreversible from here — so what it gets instead is the check, which the issue itself identifies as the missing piece: *"the check is one command and nothing invokes it."*

## The landing half, re-measured

Everything the issue was blocked on has landed, and #45's "needs a new base" question was answered by not needing a base at all.

```
$ git merge-base --is-ancestor origin/refactor/laravel-container-architecture origin/main   # 0
$ git merge-base --is-ancestor origin/feat/database-layer origin/main                       # 0
$ git merge-base --is-ancestor origin/feat/orm origin/main                                  # 0
$ git merge-base --is-ancestor origin/next origin/main                                      # 0
```

`next` merged as **PR #273 on 2026-08-04**, and the twenty PRs merged since — #314 through #375 — all target `main`, as do all five open right now. The only PR in that window with any other base is **#313**, which merged into `next` one second before #273 did and is part of the same landing.

## What that did to the plan document

`plans/orm/README.md` still routed a reader to `next` in **three** places, and the number that matters is what `next` is now:

```
$ git rev-list --count origin/main..origin/next     # 0   — next is ahead by nothing
$ git rev-list --count origin/next..origin/main     # 83  — and behind by 83
$ git log -1 --format=%ci origin/next               # 2026-08-03 09:05:26 +0200
```

So "Branch from `next`" is not merely out of date, it is the pooling failure this plan is about, entered from the top: a contributor following it starts 83 commits behind the trunk and stacks onto a branch nothing will carry upward. The three places were **Stack position**, **Picking up an iteration** step 0 (which said "not from `main`", inverted), and the **Where this stack merges** entry under Open decisions, which described #273 as an open proposal.

The section says of itself that it has gone stale three times and that a date cannot rot. This is the fourth, and it is the counter-example to its own claim: the previous version *was* dated and was still wrong, because a date records when a branch was right and not when it stopped being. That is written into the section rather than quietly fixed, and it is the argument for the second half of this PR.

## The hygiene half: the check, not the deletion

Re-running the runbook against `origin/main` today rather than against `feat/orm` (the issue body) or `origin/next` (its second comment):

```
$ sh plans/orm/branch-audit.sh
trunk origin/main at 54d4278

HOLDS UNRESOLVED WORK — decide these before deleting anything:
  origin/fix/orm-soft-delete-field-constraint(1)

SUPERSEDED — the commits are not on the trunk, the content is (matched by subject):
  origin/feat/orm-json-coverage
  origin/feat/orm-json-filters

origin/main: 120 deletable, 2 superseded, 1 holding unresolved work
```

**123 branches now, not 120** — four have appeared since the issue was written — and the finding reproduces: the same three hold anything, and the same 120 hold only the merge commits their own child PRs created.

### The correction the audit needed

The runbook's second command is `git rev-list --no-merges --count`, and **a commit count cannot see a rebase**. That is not hypothetical here — it is why the issue's own second comment had to hand three branches over as unresolved when two of them were not:

```
$ git log --oneline -1 origin/feat/orm-json-filters
1fae1ea JSON path filters, and the dialect split they inherit from Prisma

$ git log origin/main --oneline --fixed-strings \
      --grep="JSON path filters, and the dialect split they inherit from Prisma"
cc22399 JSON path filters, and the dialect split they inherit from Prisma
```

`cc22399` is `1fae1ea` rebased by #299 — the same seven files plus one `docs/llms-full.txt` hunk, `641` insertions against `671`. Restricted to the seven they share, each adds **641 lines and they differ on exactly one of those lines** (`+  param,` where the other has a blank), the rest of the divergence being blob hashes, hunk offsets and context lines from the base moving underneath. That is more than enough for two different patch-ids, so `git cherry` reports `+` here too:

```
$ git cherry origin/main origin/feat/orm-json-filters
+ 1fae1eaec30557a8436fa7877fb44ec93b2c34fa
```

Both of the obvious tests therefore report this branch as holding unlanded work, and will for as long as the branch exists. Subjects survive a rebase where patch-ids do not, so the script matches on the subject and reports those separately — `SUPERSEDED` rather than `HOLDS`.

The match is a whole-line comparison (`grep -Fxq` against the trunk's subjects), **not** `--grep`. Review caught that the first version used `--grep`, which searches the whole message — subject *and* body — for a *substring*: a branch subjected `0.51.0-rc.1` matched `0.51.0-rc.12` on a trunk that had never seen it. That matters because `SUPERSEDED` is a decision and not a note — it is not counted in `n_holds`, so it sets the exit status, and it is concatenated into the printed `git push origin --delete` line. A false hit proposes a live branch for deletion. Fixed, with today's classification unchanged. What remains is the genuine heuristic, labelled as one in the output: two unrelated commits can share an identical subject.

`fix/orm-soft-delete-field-constraint` is correctly in the third bucket. Its subject has no counterpart on `main`, which matches #299's account of it: the content was *adapted* into `policies.test.ts` rather than carried, and #262 closed around it. It is the one branch that still wants a decision, and it is why the script exits 1 today.

### Why this is a script and not a CI job

The issue's second comment says a CI job is what would have caught every instance, and also that it is *"worth its own issue if this one closes on the deletion."* That is the right split and this PR keeps it:

- The script **exits 1 today**, on `fix/orm-soft-delete-field-constraint`. Wiring it into `.github/workflows/ci.yml` now lands a red check, and `ci.yml`'s own comments are emphatic that a board red for a reason nobody acts on stops being read.
- It cannot be made green from inside a PR. The branch genuinely holds a commit that is not reachable from `main`; the only thing that resolves it is deleting the branch, on the remote.

So the exit status is already the gate — `0` when nothing is unresolved, `1` when something is — and wiring it up is a one-line change once the deletion happens. The reason it is not wired yet is written at the top of the file rather than left for someone to rediscover.

The script **reports and does not delete**, for the reason the issue's comment gives for handing the list over rather than running it. When nothing is unresolved it prints the `git push origin --delete` line for the current set, so the re-verify-then-delete loop is one command instead of a shell loop reconstructed from a GitHub comment.

## Evidence that the script works, including the path that cannot happen yet

The all-clear branch is unreachable against `origin/main` today, so it was exercised against a scratch trunk that contains the three stragglers — `git merge -s ours`, which makes them ancestors without touching content:

```
$ git checkout -b tmp-audit-all-clear origin/main
$ git merge -s ours --no-edit origin/feat/orm-json-filters \
      origin/feat/orm-json-coverage origin/fix/orm-soft-delete-field-constraint
$ sh plans/orm/branch-audit.sh tmp-audit-all-clear
tmp-audit-all-clear: 123 deletable, 0 superseded, 0 holding unresolved work

Nothing is unresolved. To delete, re-run this and pipe the list rather
than trusting a copy of it — the set changes as branches land:

  git push origin --delete ci/orm-lint-zero ci/orm-postgres … tests/orm-where-mutants
$ echo $?
0
```

123 branch names, exit 0, and the branch was deleted afterwards. The other three paths:

| path | how | result |
| --- | --- | --- |
| nothing unresolved | scratch trunk above | exit 0, delete line printed |
| something unresolved | `origin/main` | exit 1, no delete line |
| no branches match at all | copy with the prefix pattern replaced | exit 0, no delete line, no spurious command |
| trunk does not exist | `origin/does-not-exist` | exit 2, `no such ref` on stderr |

The third row is the one that found a real bug. `branches=$(… | grep …)` under `set -eu` aborts when `grep` matches nothing, because the assignment takes the pipeline's status — so a clone where the branches had already been deleted, which is the state this whole exercise is trying to reach, would have exited silently instead of reporting that there was nothing left to do. `|| true`, with the reason at the line.

`shellcheck` clean, exit 0.

## Tests

`plans/orm/README.md` is not an inert document — `packages/gemi/orm/plan-record.test.ts` reads it and asserts it does not describe shipped API as forthcoming, and it is cited by name from `strategy.ts`, `lateral.ts`, `nested-writes.ts` and `order-relation.ts`.

```
                       Tests  3110 passed | 1 skipped (3111)
```

Which is the count on the base commit. `tsc --noEmit -p tsconfig.json` clean; `oxlint orm --deny-warnings` — 0 warnings, 0 errors. `prettier --check` reports the same result on `plans/orm/README.md` before and after this change, so nothing here is a formatting regression and reformatting it would be a large unrelated diff.

Run inside a worktree with its own `bun install`, since `templates/saas-starter/node_modules/gemi` otherwise resolves to the main checkout:

```
$ readlink -f templates/saas-starter/node_modules/gemi
…/.claude/worktrees/wf_6392b57a-151-10/packages/gemi
```

No Postgres run: nothing here touches a compiler, a dialect or a query.

## What this does not do, and what is left

**It does not delete the 123 branches.** That is the issue's headline second clause and it is not a commit's to make — it is irreversible from this side, and the issue's own comment declined to run it for that reason. The list is now regenerated by measurement on every run instead of living in a comment that goes stale, which is the most a PR can usefully contribute to it.

Two things a reviewer may reasonably want to happen next, both deliberately out of scope:

- **Delete the branches**, then flip the script into `ci.yml`. In that order, because the reverse lands a red check.
- **`fix/orm-soft-delete-field-constraint`** needs a decision before either. #262 is closed and #299 says its finding was carried into `policies.test.ts` in adapted form; if that is accepted, the branch is deletable and the script goes green with no code change at all.

Corroboration from elsewhere, arriving while this was being written: **#267 was closed today** on the finding that `1fae1ea`'s content reached `main` through #299. That is the same conclusion the `SUPERSEDED` bucket reaches mechanically, from a subject match rather than from reading the two commits — which is the argument for having the bucket.

Closes #268.


plans/orm/branch-audit.sh
```diff
@@ -0,0 +1,122 @@
...
+	# for the plain count to report 1 forever. Subjects survive a rebase where
+	# patch-ids do not, so the subject is what is matched.
+	#
+	# It is a heuristic, and the output labels it as one. Two unrelated commits
+	# can share a subject, so this narrows the list a person reads rather than
+	# deciding anything on its own.
+	carried=1
+	for sha in $(git rev-list --no-merges "$TRUNK..$branch"); do
+		subject=$(git log -1 --format='%s' "$sha")
+		if [ -z "$(git log "$TRUNK" --format='%h' --fixed-strings --grep="$subject" --max-count=1)" ]; then
```
nstfkc: **should fix** — the supersession test is a *substring* search over the whole commit message, not a subject comparison, and a false hit here is the one error that costs something: it moves a branch out of `HOLDS`, flips the exit status to 0, and puts the branch into the printed `git push origin --delete` line.

`--grep` matches anywhere in the message, subject *and* body. Measured against this repo's own history:

```
$ git log origin/main --format='%h|%s' --fixed-strings --grep="0.51.0-rc.1"
5412fcd|0.51.0-rc.12
1ec063a|Bump version from 0.51.0-rc.10 to 0.51.0-rc.11
6bdaf66|Bump version from 0.51.0-rc.1 to 0.51.0-rc.2
...
```

A branch whose only unique commit is subjected `0.51.0-rc.1` is reported SUPERSEDED against a trunk that has never seen it. That is not a contrived subject either — `git log origin/main --format=%s | awk 'length($0)<=22'` on this repo yields `add errors`, `app tests`, `auth v1`, `fix api route`, `add transition` and about thirty more, every one of which is a substring of something. And it matches bodies, which is why `--grep="JSON path filters"` also returns `9c8f3ef` (`Json against a real database, …`), whose subject contains no such string.

The narrowing you actually want is an exact subject match, and it is free — it reproduces today's classification exactly:

```
$ for sha in 1fae1ea 719bb4c; do
    s=$(git log -1 --format='%s' $sha)
    git log origin/main --format='%s' | grep -Fxq "$s" && echo "$sha SUPERSEDED" || echo "$sha HOLDS"
  done
1fae1ea SUPERSEDED     # JSON path filters, and the dialect split they inherit from Prisma
719bb4c HOLDS          # The unchecked `field` is an UnknownFieldError, not a `no such column`
```

So `git log "$TRUNK" --format='%s' | grep -Fxq "$subject"` gives the same two buckets today and drops the whole false-positive class. (`-x` is the load-bearing flag; `-F` you already have.) The heuristic that remains — two unrelated commits sharing an identical subject — is the one the comment above actually describes.

plans/orm/branch-audit.sh
```diff
@@ -0,0 +1,122 @@
...
+	# A commit count cannot see a rebase, and this repository lands work by
+	# rebasing often enough that the distinction decides the answer:
+	# `feat/orm-json-filters` holds `1fae1ea`, whose content is on the trunk as
+	# `cc22399` — #299's rebase of it, differing by a `docs/llms-full.txt` hunk
+	# and one line, which is enough for `git cherry` to call it unmerged and
+	# for the plain count to report 1 forever. Subjects survive a rebase where
+	# patch-ids do not, so the subject is what is matched.
+	#
+	# It is a heuristic, and the output labels it as one. Two unrelated commits
+	# can share a subject, so this narrows the list a person reads rather than
```
nstfkc: **should fix** — "this narrows the list a person reads rather than deciding anything on its own" is not what the code does, and this is the kind of comment/behaviour gap that was its own issue in this batch (#374).

SUPERSEDED is a *decision*, twice over: `n_holds` does not count it, so it is what makes the script exit 0 rather than 1, and line 119 concatenates `$superseded` into the `git push origin --delete` line alongside `$deletable`. A branch in this bucket is not narrowed for a reader — it is proposed for deletion.

That is probably still the right behaviour (a superseded branch *is* deletable), but the comment should say so, because the consequence is what decides how careful the match needs to be — see my note at line 83. Something like "…so this is a heuristic, and a false hit here puts a branch into the delete line: prefer an exact match" puts the reason at the line that owns it.

plans/orm/branch-audit.sh
```diff
@@ -0,0 +1,122 @@
...
+# fed, not one of them, and it is kept for the history it names.
+#
+# `|| true` because a `grep` that matches nothing exits 1, and under `set -eu`
+# that ends the script here — so a clone where the branches have already been
+# deleted, which is the state this whole exercise is trying to reach, would
+# abort instead of reporting that there is nothing left to do.
+branches=$(
+	git for-each-ref --format='%(refname:short)' refs/remotes/origin |
+		grep -E '^origin/(feat|fix|tests|docs|ci)/orm' |
+		grep -v '^origin/feat/orm$' || true
```
nstfkc, line 46: **should fix** — the `|| true` is the right fix for the abort you found, but it collapses two states that the header (lines 20–21, "so it can be a gate later without being rewritten") needs to keep apart.

After this line, *"every ORM branch has been deleted"* and *"this clone has no `refs/remotes/origin/*` beyond the checked-out branch"* produce byte-identical output and identical exit 0:

```
$ sh /tmp/copy-with-prefix-pattern-replaced.sh
trunk origin/main at 54d4278

origin/main: 0 deletable, 0 superseded, 0 holding unresolved work
$ echo $?
0
```

The first is the success this exercise is aiming at. The second is a broken invocation, and it is the *likely* one in the environment the header is aiming at: `ci.yml` uses `actions/checkout@v4` with its defaults, which fetches a single ref at depth 1 — I have not run it to confirm what `refs/remotes/origin` holds there, but if it holds only the checked-out branch then the gate is vacuously green forever, which is a worse failure than the red check you correctly declined to land. (On `pull_request` it would instead be exit 2 from the `origin/main` guard, which at least fails loudly.)

Cheap fix, and it keeps the `|| true`: after the assignment, treat an empty `$branches` as its own outcome — print "no ORM-named branches on `origin` — nothing to audit, or nothing fetched" and pick an exit status deliberately, rather than falling through to a 0/0/0 summary that reads like an all-clear. Worth doing now while the reasoning is in front of you, since the whole point of the header comment is that this file becomes a gate without being rewritten.

plans/orm/README.md
```diff
@@ -136,6 +153,19 @@ and the intermediates can be deleted rather than repaired one at a time. Without
 that second check the first number is alarming and misleading in the expensive
 direction: "nine PRs merged and nothing landed" reads as nine repairs.
 
+**A non-zero count is not proof the work is unlanded**, which is the one place
+the two commands above will mislead you in the other direction. They compare
+*commits*, and a rebase changes the commit while keeping the content:
+`feat/orm-json-filters` holds `1fae1ea`, whose content is on `main` as `cc22399`
+— #299 rebased it, adding a `docs/llms-full.txt` hunk and shifting every other
+one, though the seven shared files add 648 lines on each side and differ on
```
nstfkc: **should fix** — `648` is an artifact of how it was counted, and "differ on exactly one of them" reads as one *file* when the measurement is one *line*.

The seven shared files add **641** lines, not 648, and the diffstat says so — `git show --stat 1fae1ea` ends `7 files changed, 641 insertions(+)`, and `cc22399`'s `671` is exactly that plus the 30-line `docs/llms-full.txt`. 648 is what `grep -c '^+'` returns, because it counts the seven `+++ b/<path>` headers as added lines: 641 + 7 = 648.

```
$ git show --format= --unified=0 1fae1ea -- <the 7 shared files> | grep -c '^+'        # 648
$ git show --format=        1fae1ea -- <the 7 shared files> | grep '^+' | grep -vc '^+++'  # 641
$ ... same two numbers for cc22399
```

The PR body has both numbers — 641 for the total and 648 for "restricted to the seven they share" — and they describe the same set, since all seven of `1fae1ea`'s files are shared. Only one can be right.

On the second half: the two are identical on 640 of the 641 added lines and differ on one *line* — position 159, `  param,` against a blank — not on one of the seven files. Suggest "…add 641 lines on each side and differ on exactly one of those lines". The claim survives either way; it is a stronger claim than the sentence currently makes, and in a PR whose subject is a document that stated something unverified, the numbers in it are worth being exact.

plans/orm/README.md
```diff
@@ -110,8 +121,14 @@ feat/orm-08-eloquent-doorway      15 ahead of feat/orm-07-performance
 
 ### The runbook
 
-**Detect it.** One command, worth running before treating a stack as shipped,
-because every PR showing `MERGED` looks exactly like success:
+`plans/orm/branch-audit.sh` runs all of this. It is a file rather than a
```
nstfkc: **nit** — "runs all of this" overstates it by one step. The script never runs the **Detect it** command (`git rev-list --count origin/<trunkward-branch>..origin/<stack-tip>`); it starts at the second check, and it does not take a stack tip at all — only `$TRUNK` is a parameter, the branch set is fixed by the `^origin/(feat|fix|tests|docs|ci)/orm` glob. "runs the branch-hygiene half of this" or "runs the second and third checks across every ORM-named branch" would be accurate, and it also explains the next sentence better: the prose is kept precisely *because* the first step is still a by-hand command against an arbitrary pair.

nstfkc (commented):
**Verdict: minor comments.** The headline claims are all true and all reproduce, the document is strictly more correct than it was, and nothing here can affect runtime. Four findings, none blocking: three on the new script (one real latent defect in the supersession test) and one on a number in the README that is off by an artifact of how it was counted.

## What I re-measured, independently of the PR body

Every ancestry and topology claim in the rewritten **Stack position** section holds today:

```
$ for b in refactor/laravel-container-architecture feat/database-layer feat/orm next; do
    git merge-base --is-ancestor origin/$b origin/main; echo "$b -> $?"; done
refactor/laravel-container-architecture -> 0
feat/database-layer -> 0
feat/orm -> 0
next -> 0

$ git rev-list --count origin/main..origin/next     # 0
$ git rev-list --count origin/next..origin/main     # 83
$ git log -1 --format=%ci origin/next               # 2026-08-03 09:05:26 +0200
```

The "twenty PRs merged since that landing" figure is exact — `gh pr list --state merged --base main` filtered to `mergedAt > 2026-08-04T15:25:09Z` returns **20**. `git cherry origin/main origin/feat/orm-json-filters` does report `+ 1fae1ea`, so the paragraph's premise about patch-ids is real and not remembered. `feat/orm` is an ancestor of `next` as well, so the historical diagram's new `next` row is right too. So the part of #268 the PR says was still costing something genuinely was, and the fix is correct.

The script reproduces byte-for-byte against `origin/main` at `54d4278` — `120 deletable, 2 superseded, 1 holding unresolved work`, exit 1 — and I confirmed the two paths I could reach without a scratch trunk: `origin/does-not-exist` → exit 2, and a copy with the prefix pattern replaced → `0 deletable, 0 superseded, 0 holding`, exit 0. `shellcheck` is clean. The delete-line `sed` produces correct bare branch names (checked in isolation with synthetic input).

One thing the PR does *not* claim that I checked anyway, because `--no-merges` in the `deletable` test hides it: **no "deletable" branch carries an evil merge.** For every branch with zero unique non-merge commits I walked its unique merge commits and took `git show --format=` (empty iff the merge introduced no content of its own). Zero of the 120 have a non-empty one, so the `--no-merges` shortcut is not silently proposing to delete a conflict resolution. Worth having measured, since that is the one way the deletable bucket could be wrong.

## What I ran

```
cd packages/gemi && bun --bun vitest run   ->  120 files, 3110 passed | 1 skipped
cd packages/gemi && bun run typecheck      ->  clean
cd packages/gemi && bunx oxlint orm --deny-warnings  ->  0 warnings, 0 errors
shellcheck plans/orm/branch-audit.sh       ->  exit 0
gh pr checks 384                           ->  sqlite pass, postgres pass
```

The package suite matters here because `orm/plan-record.test.ts` reads `plans/orm/README.md` at test time; it passes. `bun install` in-tree first, `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`, `git status` clean afterwards.

## What I did not check

- **No mutation check, because there is nothing to mutate.** Reverting this diff changes no test's outcome — the README's routing claim and all 122 lines of the new script are guarded by nothing. That is not a fault of the PR (see the note below), but it means "the tests pass" carries no signal about this change at all, and I want that stated rather than implied.
- The template and Postgres suites. No `.ts`, no dialect, no schema, no query is touched; CI's postgres job passes.
- The all-clear path against the real remote. I did not rebuild the `merge -s ours` scratch trunk — it mutates `refs/remotes/*`, which is shared across worktrees in this repo, and other agents are working in sibling worktrees. I verified the delete-line formatting separately instead.

## Note on the closing keyword, for whoever merges

Merging this closes #268 with 123 branches still on `origin` and no follow-up issue filed. The PR body is completely straight about that under *What this does not do*, and I am **not** asking for the deletion or the `ci.yml` wiring — both are correctly out of scope and the sequencing argument (delete, then gate, never the reverse) is right. But the two remaining halves currently live only in a PR body that is about to be archived. Filing the follow-up issue before merge, or dropping `Closes #268.`, would keep them addressable. Your call — flagging it, not blocking on it.

plans/orm/branch-audit.sh
nstfkc: **Addressed** — `3489b8a`. You are right, and it is worse than a narrowing failure: the false hit is load-bearing in both directions.

Reproduced your two cases against this repo before changing anything:

```
$ git log origin/main --format="%h|%s" --fixed-strings --grep="0.51.0-rc.1" | head -5
5412fcd|0.51.0-rc.12
1ec063a|Bump version from 0.51.0-rc.10 to 0.51.0-rc.11
b657f97|Bump version from 0.51.0-rc.9 to 0.51.0-rc.10
cea0497|Remove the `gemi/runtime` export, which has resolved to nothing since 2024
6bdaf66|Bump version from 0.51.0-rc.1 to 0.51.0-rc.2
```

`cea0497` is the sharper half of it — its *subject* contains no such text at all, so that hit came from the body, exactly as your `9c8f3ef` example does for `JSON path filters`.

Now `grep -Fxq -e "$subject" "$trunk_subjects"`, with the trunk's subjects read once into a `mktemp` file rather than a `git log` per candidate commit (`trap` cleans it up). `-e` because a subject can begin with `-`.

**Classification against `origin/main` is unchanged**, which is the check that the fix is a narrowing and not a rewrite: `120 deletable, 2 superseded, 1 holding unresolved work`, exit 1.

I also ran this as a mutation check rather than asserting it, since the script has no test to revert. Synthetic branch under a private `refs/remotes/mutcheck/` namespace (not `origin/`, so nothing else in this repo could see it), one unique commit subjected `0.51.0-rc.1`:

```
--- FIXED (grep -Fxq) ---
HOLDS UNRESOLVED WORK:
  mutcheck/feat/orm-fake-substring(1)
0 deletable, 0 superseded, 1 holding      exit=1

--- REVERTED (--grep) ---
SUPERSEDED:
  mutcheck/feat/orm-fake-substring
0 deletable, 1 superseded, 0 holding      exit=0
  git push origin --delete mutcheck/feat/orm-fake-substring
```

So the pre-fix version does not merely misreport it — it exits 0 and prints the branch in the delete line. Ref deleted afterwards; `git for-each-ref refs/remotes/mutcheck` is empty.

plans/orm/branch-audit.sh
nstfkc: **Addressed** — `3489b8a`. The comment was describing a script that reports; the one I wrote decides, and you traced both routes correctly (`n_holds` sets the exit status, line 119 concatenates `$superseded` into the delete line).

The comment now says so at the line, and states the consequence as the reason the match has to be exact rather than close:

> It is a heuristic, and it decides rather than advises: a branch in this bucket is not counted as holding work, so it is what makes the exit status 0, and it is concatenated into the printed `git push origin --delete` line below. A false hit therefore proposes a live branch for deletion, which is why the match is exact and not merely close.

…followed by the `--grep`-is-a-substring paragraph, so the two notes read as one argument. The remaining heuristic is named separately and honestly: two unrelated commits can share an identical subject.

I took the output label with it, since it had the same gap — it announced a bucket without saying what happens to it:

```
SUPERSEDED — the commits are not on the trunk, the content is (exact subject
match), so these are proposed for deletion alongside the empty ones:
```

Agreed on the behaviour: a superseded branch is deletable, and that was always the intent. It was only the comment (and the label) that pretended otherwise. Fair to name #374 here — same shape of defect.

plans/orm/branch-audit.sh
nstfkc: **Addressed** — `3489b8a`. Right, and the framing that convinced me is yours: a gate that is vacuously green forever is a worse failure than the red check I declined to land, because the red one is at least visible.

The two states are now separated by how many refs `origin` has at all, and the broken one exits 2 like the missing-trunk guard rather than 0:

```sh
n_origin=$(git for-each-ref --format="%(refname:short)" refs/remotes/origin |
	grep -cv "^origin/HEAD$" || true)

if [ -z "$branches" ]; then
	if [ "$n_origin" -le 1 ]; then
		echo "refs/remotes/origin holds $n_origin branch(es) — nothing was fetched to audit." >&2
		echo "Run 'git fetch origin' (a single-ref or shallow checkout is not enough)." >&2
		exit 2
	fi
	printf "no ORM-named branches among %d on origin — nothing to audit.\n" "$n_origin"
	exit 0
fi
```

The `|| true` stays, for the reason it was added.

Both paths exercised:

```
$ sh /tmp/copy-with-prefix-pattern-replaced.sh          # genuine all-clear
trunk origin/main at 54d4278

no ORM-named branches among 195 on origin — nothing to audit.
exit=0

$ sh /tmp/copy-with-refs-remotes-repointed.sh           # unfetched clone
trunk origin/main at 54d4278

refs/remotes/nowhere holds 0 branch(es) — nothing was fetched to audit.
Run 'git fetch origin' (a single-ref or shallow checkout is not enough).
exit=2
```

(The second copy repoints `refs/remotes/origin` at an empty namespace, which is the same input a single-ref checkout produces without my needing to run CI to find out.)

The `<= 1` threshold is deliberately crude — it is asking "did anything get fetched", not "was the fetch complete". A partial-but-nonempty fetch would still under-report, and no cheap local test distinguishes that from branches genuinely having been deleted. Printing the denominator in the all-clear line (`among 195`) is the compromise: a reader or a log sees the number the answer was computed from. And you are right that on `pull_request` the `origin/main` guard fires first with exit 2 anyway; this covers the `push`-triggered case where it would not.

plans/orm/README.md
nstfkc: **Addressed** — `3489b8a`. Confirmed on both halves, and 648 was exactly the artifact you name.

```
$ git show --stat 1fae1ea | tail -1
 7 files changed, 641 insertions(+)
$ git show --stat cc22399 | tail -1
 8 files changed, 671 insertions(+)      # 641 + the 30-line docs/llms-full.txt
```

And the added-line sets, restricted to the seven shared files, with the `+++ b/<path>` headers excluded:

```
$ wc -l a.txt b.txt
     641 a.txt
     641 b.txt
$ diff a.txt b.txt
159c159
< +
---
> +  param,
```

641 on each side, one differing line at position 159. So the PR body was right and the README was wrong, and they were describing the same set — as you say, all seven of `1fae1ea`'s files are shared, so "restricted to the seven" is not a restriction at all here.

The sentence now reads:

> …#299 rebased it, adding a 30-line `docs/llms-full.txt` hunk and shifting every other one, though the seven files they share add **641** lines on each side (`641` and `671` in the two diffstats, differing by exactly that hunk) and differ on exactly one of those lines.

Which also fixes the second half — it was reading as one *file* — and now carries the two diffstat numbers so the arithmetic is checkable from the sentence. The PR body had the same `648` in it and has been corrected to match.

Point taken on why this one mattered more than its size suggests: this is a PR about a document that stated something nobody had verified.

plans/orm/README.md
nstfkc: **Addressed** — `3489b8a`. Correct on both counts: the script never runs **Detect it**, and it does not take a stack tip — only `$TRUNK` is a parameter and the branch set is fixed by the glob.

Rewritten to say which checks it runs and to make that the reason the prose stays, which is the connection the old sentence was hiding:

> `plans/orm/branch-audit.sh` runs the second and third checks below across every ORM-named branch on `origin`. It does **not** run the first one — **Detect it** takes an arbitrary pair of branches, and the script takes only a trunk, with the branch set fixed by an `origin/(feat|fix|tests|docs|ci)/orm` glob. That is why the prose stays: the first step is still a by-hand command, and the rest is written out for a reader who needs to run one of them against some other pair.

The "file rather than a paragraph" argument moved to its own paragraph after that, since it was doing unrelated work in the same sentence.

nstfkc:
## Review round 1 — all five addressed, none declined, none deferred

Pushed as `3489b8a` on the same branch. Every inline comment is answered at its own line; this is the disposition table plus the two things that came out of the review body rather than the diff.

| # | line | finding | disposition |
| --- | --- | --- | --- |
| 1 | `branch-audit.sh:83` | `--grep` is a whole-message **substring** search, not a subject comparison | **addressed** — `grep -Fxq` against the trunk's subjects |
| 2 | `branch-audit.sh:78` | the comment says SUPERSEDED narrows a list; the code makes it decide | **addressed** — comment and output label both say what it does |
| 3 | `branch-audit.sh:46` | `\|\| true` collapses "all deleted" with "nothing fetched" | **addressed** — separated, the broken one exits 2 |
| 4 | `README.md:161` | `648` is `grep -c '^+'` counting seven `+++` headers; 641 is the number | **addressed** — and "one of them" was reading as one *file* |
| 5 | `README.md:124` | "runs all of this" overstates by one step | **addressed** — names the checks it does run |

### The one that was more than a narrowing

Finding 1 is the only one with teeth, and it is worse than "the heuristic is loose". `SUPERSEDED` is a **decision**, as finding 2 says: it is excluded from `n_holds`, so it sets the exit status, and it is concatenated into the printed `git push origin --delete` line. A substring false hit therefore proposes a live branch for deletion and reports all-clear while doing it. Your `0.51.0-rc.1` case reproduces, and the body-match half is sharper than the example given — `cea0497` ("Remove the `gemi/runtime` export…") matched too, and its subject contains no such text at all.

Since the script has no test to revert, I ran the mutation check by construction instead. A synthetic branch under a private `refs/remotes/mutcheck/` namespace — deliberately not `origin/`, so no sibling worktree could see it — with one unique commit subjected `0.51.0-rc.1`:

```
--- FIXED (grep -Fxq) ---            --- REVERTED (--grep) ---
HOLDS UNRESOLVED WORK:               SUPERSEDED:
  mutcheck/feat/orm-fake-…(1)          mutcheck/feat/orm-fake-…
1 holding          exit=1            1 superseded       exit=0
                                       git push origin --delete mutcheck/feat/orm-fake-…
```

Ref deleted afterwards. **Classification against `origin/main` is unchanged** — `120 deletable, 2 superseded, 1 holding unresolved work`, exit 1 — which is the evidence that this narrowed the test without moving the answer.

### Your note on the closing keyword

Checked first, because two PRs in this batch are broken that way: **#384 is not one of them.** `gh pr view 384 --json closingIssuesReferences` returns `268`, so the keyword is in the body and fires.

On the substance — that merging archives the PR body along with the two remaining halves — you are right, and that is a real objection to leaving them there. I have **not** filed the follow-up issue and have **not** dropped `Closes #268.`; both are maintainer calls, and the branch deletion in particular was explicitly left to a human. What I did instead is put the remainder somewhere that survives the merge, in the plan document itself, under **The runbook**:

> **What is still owed, in this order.** The script reports and never deletes, so none of the following is discharged by running it, and it is written here rather than left in a merged PR description:
>
> 1. **Decide `fix/orm-soft-delete-field-constraint`.** …the only reason the script exits non-zero…
> 2. **Delete the branches**, from the list the script prints — regenerated by running it, not copied…
> 3. **Then** wire the script into `.github/workflows/ci.yml`. In that order…

So whoever picks this up finds it in the file the work is about, not in an archived description. If you would still rather have an issue number to point at, that is a one-line follow-up and I am happy for it to block the merge — I just did not want to open it unasked.

### Two things I am not doing, and why

- **The `childField` → `childFields` rename from #386.** Not applicable rather than deferred: this PR touches no TypeScript at all. Two files, `plans/orm/README.md` and `plans/orm/branch-audit.sh`.
- **`connectOrCreate`'s hit-branch short-circuit (#378) and the numeric JSON path segments (#385).** Assigned to those PRs by the cross-PR review, and neither has any surface here.

### Verification

Re-run in this worktree after `bun install` in-tree (`readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`; `git status` clean afterwards):

```
packages/gemi   bun --bun vitest run              ->  120 files, 3110 passed | 1 skipped
packages/gemi   bun run typecheck                 ->  clean
packages/gemi   bunx oxlint orm --deny-warnings   ->  0 warnings, 0 errors
                shellcheck plans/orm/branch-audit.sh  ->  exit 0
```

`orm/plan-record.test.ts` reads `plans/orm/README.md` at test time, which is why the package suite is the one that matters here. No template or Postgres run: still no `.ts`, dialect, schema or query in the diff.

All five script paths exercised, including the one this round added:

| path | how | result |
| --- | --- | --- |
| something unresolved | `origin/main` | exit 1, no delete line |
| nothing unresolved | scratch trunk via `git commit-tree` with the three stragglers as extra parents | 123 deletable, exit 0, delete line printed |
| no ORM branches, remote fetched | copy with the prefix pattern replaced | `no ORM-named branches among 195 on origin`, exit 0 |
| no ORM branches, nothing fetched | copy with `refs/remotes/origin` repointed at an empty namespace | exit 2 on stderr |
| trunk does not exist | `origin/does-not-exist` | exit 2 |

The scratch trunk is a bare `commit-tree` object this time rather than a `merge -s ours` branch, so nothing under `refs/` was written for it at all — your point about `refs/*` being shared across this repo's worktrees was well taken.

`prettier --check plans/orm/README.md` still reports identically before and after the change, as it did on the base commit.

PR #385, opened 2026-08-09
Close #301: the corpus walks JSON path filters, and depth stops minting plan entries

### LINKED ISSUE #301: corpus.ts is not extended for JSON path filters, and cannot be until a fixture grows a Json column
Split out of the review on #299, which agreed the omission is defensible for a
release cut and asked for an issue so it does not get lost.

## What is missing

`packages/gemi/orm/corpus.ts` says in its own header:

> **This is the thing to extend when a new argument is implemented.** It is
> deliberately broad rather than minimal: a case costs one line, and both
> failure modes it guards are silent.

#299 implemented a new argument — `where: { metadata: { path: …, equals: … } }` —
and did not extend it. So the two suites that walk the corpus,
`plan-key.invariants.test.ts` and `binding.invariants.test.ts`, do not see JSON
path filters at all.

## Why it was not just done

The corpus is compiled against `userWithProfile`, which has **no `Json` column**:

```ts
for (const args of READS) {
  for (const op of ["findMany", "findFirst"]) {
    attempt(op, args, () => compileRead(userWithProfile, op, args, dialect));
  }
}
```

A `path` entry there is refused with *"A 'path' filter reads inside a JSON
document, and 'metadata' is a String column"* — and refusals are **tracked
rather than swallowed**, deliberately, because silently dropping them is how
this suite was vacuous on its first run. So adding the case without adding the
column makes the suite report a dropped entry rather than cover anything.

Adding a `Json` column to `userWithProfile` is not a one-line change either: it
is a shared fixture, and every suite that reads all columns has its expected SQL
text in it. That is a re-baseline, which is why #299 left it.

## What is and is not at risk

**Not at risk — the direction that matters.** The #266 invariant is *different
SQL implies different plan key*, and it holds here by construction: the filter
names (`path`, `equals`, `string_contains`, …) are entries in the canonical
shape, so two JSON filters that compile differently cannot share an entry. The
sentinel case that made #266 possible is now refused outright inside a path
filter (#299), so the shapes that collided there cannot arise.

**At risk — the cheaper direction, and one known instance.** On Postgres a
`path` array's *length* reaches the key, because `canonicalShape` records arrays
element-wise:

```
{ path: ["a"],      equals: "x" }   ->  one entry
{ path: ["a", "b"], equals: "x" }   ->  a second entry, identical SQL
```

The SQL text is the same — the path binds as one `text[]` parameter — so this
mints one cache entry per distinct path depth holding the same statement.
Wasteful rather than wrong, and exactly the shape `plan.ts` already documents
and handles for `in` / `notIn` via `LIST_KEYS` and `collapsedList`:

> Recording them element-wise there is not *wrong*, but it mints one entry per
> distinct list length, each holding SQL identical to its neighbours'.

A `path` on a dialect that binds it whole is the same case and is not in
`LIST_KEYS`. Nothing measures it, because nothing walks a `path` shape.

## What this issue wants

1. Give a fixture a `Json` column — either `userWithProfile`, accepting the
   re-baseline, or a dedicated schema the invariant suites also walk.
2. Add `path` shapes to `READS`: both grammars, several depths, each filter,
   and the refusals.
3. Decide whether `path` belongs in `LIST_KEYS` on a dialect whose
   `jsonPathSyntax` is `"array"`, so depth stops minting entries. Note this is
   dialect-conditional in a way `in` is not — `plan.ts` asks the dialect through
   `collapsedList` already, so the seam exists.

Worth doing together: (1) is the cost, and (2) is what makes (3) measurable
rather than reasoned about.

nstfkc:
`corpus.ts` says in its own header that it is the thing to extend when a new argument is implemented. #299 implemented `where: { metadata: { path: …, equals: … } }` and could not extend it, because both invariant suites compile the corpus against `userWithProfile` and a `path` on a non-Json column is refused outright — a refused entry covers nothing, and refusals are tracked rather than swallowed, so adding one without the column would have made the suite *report a hole* rather than fill it.

## The cost #299 and #301 both assumed, measured

> Adding a `Json` column to `userWithProfile` is not a one-line change either: it is a shared fixture, and every suite that reads all columns has its expected SQL text in it. That is a re-baseline, which is why #299 left it.

**That is not true today, and it is the reason this PR is small.** `USER_COLUMNS` is spelled against `user`, not against `userWithProfile`, and `grep` finds four other files touching `userWithProfile` — `compile/refusals.test.ts`, `compile/write.test.ts`, `plan.writes.discrimination.test.ts` and the two invariant suites — none of which asserts a full column list against it. Adding the column and running the package:

```
Test Files  64 passed (64)          <- orm only, before and after
     Tests  2421 passed (2421)
```

One test failed, and it was mine: `architecture.test.ts` objecting to a `USER_WITH_PROFILE_COLUMNS` export nothing used. It is not in the diff.

The column is `metadata Json?`, which is what the template's real `User` model declares. `user` — the fixture whose docblock claims to be a faithful copy of it — has neither `metadata` nor `avatar`, so it has drifted; this moves `userWithProfile` toward the real model rather than away from it. Re-baselining `user` is a different change and is not here.

## What the corpus gained

33 read shapes and 2 writes. `path` is the only argument in the corpus whose **grammar** is dialect-specific — Postgres takes `path: ["a", "b"]` and refuses a string, SQLite takes `path: "$.a.b"` and refuses an array, which is Prisma's own split — so **every entry is refused on exactly one dialect**, and both spellings are present so neither half goes unwalked.

That makes the refusal bookkeeping load-bearing rather than incidental. `EXPECTED_REFUSALS` now names, per dialect:

| dialect | pattern | what would make it stop firing |
| --- | --- | --- |
| sqlite | `a JSON path is a JSONPath string` | SQLite accepting the array grammar |
| sqlite | `is not available on sqlite` | `jsonFilters` widening to Postgres's set |
| postgres | `a JSON path is an array of keys` | Postgres accepting the JSONPath grammar |

and four that fire on **both**, split out into a `SHARED_REFUSALS` list so "this is a dialect split" and "this is refused everywhere" no longer read alike: `is a String column`, `A JSON path cannot be empty`, `A 'path' needs a filter beside it`, `A JSON path filter takes`. Each of those four is in the corpus in *both* grammars — the grammar check runs before all of them, so a single spelling would fire on one dialect and be swallowed on the other. The suite asserts every listed pattern still fires, so none of this can go stale silently.

## The decision the issue asked for: `path` joins `LIST_KEYS`

The issue's third question was whether `path` belongs in `LIST_KEYS` on a dialect whose `jsonPathSyntax` is `"array"`. It does. Measured through `compileRead` + `planKey` on both dialects:

```
postgres
  { path: ["a"],           equals } -> where ("metadata" #>> $1) = $2   key ...path:[string]
  { path: ["a","b"],       equals } -> where ("metadata" #>> $1) = $2   key ...path:[string,string]
  { path: ["a","b","c"],   equals } -> where ("metadata" #>> $1) = $2   key ...path:[string,string,string]

sqlite
  { path: "$.a",     equals } -> where (json_extract("metadata", ?)) = ?   key ...path:string
  { path: "$.a.b",   equals } -> where (json_extract("metadata", ?)) = ?   key ...path:string
```

Three cache entries, one statement. SQLite was never affected — a path is a *string* there, so depth was never in the shape — which is why this is Postgres-only in effect and dialect-conditional in a way `in` is not.

> **Review round 1.** This table originally carried a fourth row, `{ path: ["a", 0] }`. #380 makes `assertPathShape` refuse a numeric segment and narrows `JsonPath` to `readonly string[]`, so that row stops compiling the moment the two land together — the cross-PR review measured it as two failures on a clean merge. Every numeric segment in this branch is now spelled `"0"`; `#>` takes a `text[]` and reaches the same element either way. The count above is what it now is, re-measured, not what it was.

### The gate, stated rather than hidden

The collapse is gated on `collapseLists`, which is `bindsListAsOneParameter` — a property about `in` lists, not about paths. It is the right condition on both implemented dialects because the array path *grammar* exists only where `jsonPathSyntax` is `"array"`, and the one dialect with that grammar binds the array whole. A dialect that spelled a path as an array but emitted one parameter per segment would make it wrong in the dangerous direction. There is no non-speculative seam to hang a flag on today; `plan-key.invariants.test.ts` walks both grammars on both dialects asserting same-key-implies-same-SQL, so that is what fails if one appears. Written out at the line.

### The name collision, measured rather than assumed

`LIST_KEYS` matches by name at any depth, so a *column* named `path` is in scope. Compiled against a scratch schema carrying `path` as a Json column and as a `String[]`:

```
postgres
  where: { path: ["a"] }        Json  ->  "path" = $1::text::jsonb      (1 and 2 elements alike)
  where: { path: { equals: [] } } list ->  "path" = $1
  data:  { path: ["a","b"] }    both  ->  values ($1, $2)
  where: { path: ["a","b"] }    list  ->  refused — "a bare array is not a filter on one"
sqlite   collapseLists is false; the branch is unreachable
```

Every position binds as one parameter, so the collapse is *correct* there rather than merely harmless — it closes a sliver of the bare-array-in-`data` churn `LIST_KEYS`' own comment leaves open.

## The hole the collapse opened, and the two tests that close it

A collapsed key is only safe if a warm plan binds the **new** call's path. Nothing in the repository could show that, and the reason is worth stating: `expectSame` calls `clearPlanCache()` at the top of every case, so all eleven Postgres JSON path differential cases compile fresh and a plan serving a stale path leaves every one of them green.

The compile-level suites were blind to it too, and provably: a compiler that reads `filter.path` at compile time and closes over it emits **byte-identical text** for two different paths and keys them identically, so both existing invariants pass. Mutating `compile/where.ts`

```diff
-  const path: Binder = (args) => locate(args)?.path;
+  const compiledIn = filter.path;
+  const path: Binder = () => compiledIn;
```

was caught by nothing on the base branch and by nothing in the first revision of this PR either.

**1. A corpus-wide property, in `binding.invariants.test.ts`:** compile from `args`, bind `twin`, and require the array the plan compiled *for* `twin` produces. Written over the whole corpus rather than over `path` alone, because nothing about the failure is specific to paths — `where: { name: "a" }` compiled with a captured `"a"` fails the same way and was equally unasserted.

The clock is frozen (`vi.useFakeTimers`) rather than the timestamps masked, and that is not tidiness: `encode` turns a `Date` into a number on SQLite and passes it through on Postgres, so a `value instanceof Date` mask covers one dialect, silently drops the other, and then flakes on whichever millisecond boundary falls between two binds — which it did, on the first run. `cuid` has no clock to freeze and is masked by kind.

**2. A live-database case, in `differential.test.ts`,** running two depths back to back through one warm cache. The pair is chosen to discriminate rather than merely agree: the shallow path matches Ada and the deep one matches nobody, so a plan answering the second with the first's path returns a row instead of none. Run in both orders, because "the first shape compiled decides for the second" is directional, and Prisma is still the oracle — `expectSame` pins both answers first.

### Mutation results

Every new assertion was checked by reverting the thing it covers.

| mutation | what fails |
| --- | --- |
| `path` out of `LIST_KEYS` | `a JSON path's depth is bound, not compiled in — postgres`: `expected 3 to be 1`, printing three keys over one statement |
| the segment-kind gate reverted to an unconditional `collapsedList` | `a path a cold compile refuses is not served by a warm plan — postgres`: `["a",null]` keys the same as `["a"]` (round 1) |
| `path` out of `LIST_KEYS`, on live Postgres | `a JSON path is one plan whatever its depth is`: `expected 2 to be 1` compiles |
| path captured at compile time | `a plan compiled for one call binds the next call's values` — **both** dialects, showing `reused ["$.a","ZZZZ"]` against `fresh ["ZZZZ","ZZZZ"]` |
| path captured at compile time, on live Postgres | the same case, `expected [ 1 ] to deeply equal []` — the deep path answered with the shallow path's row |
| a JSON refusal pattern removed from `EXPECTED_REFUSALS` | `every corpus entry compiles, or is refused by name` |

The two live-Postgres mutations are the ones worth reading: they are the wrong-rows failure the collapse makes reachable, demonstrated against a real server rather than argued.

## Verification

```
                       Tests  3116 passed | 1 skipped (3117)      (base 3110)

saas-starter      Test Files   20 passed | 3 skipped (23)
   sqlite              Tests  793 passed | 133 skipped (926)      (base 792)
                  type tests  116 passed, no errors
```

**Postgres**, brought up through `app/models/postgres.sh` (postgres:16 in Docker, `prisma db push`, `@prisma/client` 6.19.2 regenerated for the provider), run with CI's `-t postgres` filter:

```
Test Files  14 passed | 9 skipped (23)
     Tests  874 passed | 789 skipped (1663)      (base 873)
```

An unfiltered Postgres run is `2 failed | 21 passed`, both failures being the SQLite halves of the differential suites failing loudly by design under a Postgres-generated client — the shape CI's filter exists for, and the same numbers #375 recorded.

`tsc --noEmit` clean on `packages/gemi` and on `tsconfig.generated.json`. `oxlint orm --deny-warnings`: 0 warnings, 0 errors. `bun run lint`: 79 warnings, 0 errors — unchanged. Schema restored to `sqlite`, container removed, `git status` clean.

## Review round 1 (80f99a8)

**The collapse erased the segment *kinds*, not only the depth — blocking, fixed.** `collapsedList` maps any non-empty array to `[*]`, so `["a", null]`, `["a", undefined]`, `[["a"]]` and `[{ k: 1 }]` all keyed identically to `["a"]` while a cold compile refused every one of them. `[["a"]]` is the bad one: it flattens to `{"a"}` and returns the rows for `["a"]` — wrong rows, no error. A regression this branch introduced; on `main` those shapes key element-wise and are refused on every call.

`path` now collapses **only when every segment is a `string`**:

```ts
if (key !== "path" || value.every((segment) => typeof segment === "string")) {
  return collapsedList(value);
}
```

The reviewer proposed `string | number`, which is exactly today's accepted set. The predicate has to stay a *subset* of what `assertPathShape` accepts rather than an equal — a collapsed shape covering a refused argument is a correctness bug, an accepted argument left uncollapsed costs one cache entry — and #380 narrows `JsonPath` to `readonly string[]`, which would make `string | number` wider than the accepted set and re-open this hole for `["a", 0]`. `string` is right before and after.

New property, `a path a cold compile refuses is not served by a warm plan — %s`: five malformed shapes per dialect, each asserted refused cold (so the test cannot go vacuous), each keyed apart from the valid path, and each still throwing after `getOrCompile` has been warmed with `["a"]`, with `planCacheStats().hits === 0`. `{ path: ["a", null] }` is in the corpus too — it needs no `EXPECTED_REFUSALS` change, since it fires the array-grammar message on Postgres and the wrong-grammar one on SQLite.

**#380 × #385, done here.** `["a", 0]` respelled `["a", "0"]` in `corpus.ts`, `plan-key.invariants.test.ts` and the `plan.ts` docblock, per the cross-PR review's assignment. The measurement above is restated as three depths / three entries, which is what it now is.

**Three comment-accuracy corrections**, each verified by compiling every JSON corpus entry on both dialects rather than by reading. "The grammar check runs before all of them" is right for the bare-path and unknown-filter pairs and wrong for the String-column and empty-path pairs; "every entry below is refused on exactly one dialect" is true of the two grammar blocks and false of the refusal block; `path: 5` reaches the same throw and the same message as the wrong-grammar entries rather than a branch of its own. Entries kept, reasons rewritten.

**Not done here.** The `childField` → `childFields` rename from #386 — #386 is not on `main`, so this branch cannot reference the new names; it is merge-time work for whoever lands #386 last. The `connectOrCreate` hit-branch short-circuit belongs to #378 by the cross-PR review's own assignment.

## Noted, not fixed

- **The `user` fixture has drifted from the template's `User`.** Its docblock claims same fields, same order, same types; the template declares `metadata Json?` and `avatar Bytes?` and the fixture has neither. Fixing it is the re-baseline this PR found it did not need, and belongs to whoever wants the Bytes coverage too.
- **`expectSame` clears the plan cache per case**, so *no* differential case exercises a cache hit. That is right for a differential — it compares answers, not caching — but it means the plan cache's live behaviour is covered only by the handful of bespoke tests that ask for it explicitly. There is one for `create` in `writes.differential.test.ts` and now one for JSON paths; everything else is compile-level.

Closes #301.

### DIFF
--- a/packages/gemi/orm/corpus.ts
+++ b/packages/gemi/orm/corpus.ts
@@ -71,27 +71,39 @@ export const READS: unknown[] = [
   // and a refused entry covers nothing. `userWithProfile` grew `metadata` for
   // this; see the note there for why that cost nothing.
   //
-  // **Every entry below is refused on exactly one dialect, by name, and that is
-  // the point rather than a nuisance.** `path` is the only argument in this
-  // corpus whose *grammar* is dialect-specific: Postgres takes
-  // `path: ["a", "b"]` and refuses a string, SQLite takes `path: "$.a.b"` and
-  // refuses an array. That is Prisma's own split, measured against a generated
-  // client on both, so writing one spelling would leave the other dialect's
-  // half of the feature unwalked. `plan-key.invariants.test.ts` lists both
-  // messages in `EXPECTED_REFUSALS` and asserts each still fires, so a dialect
-  // that quietly started accepting the other form fails there.
+  // **Every entry in the two *grammar* blocks below is refused on exactly one
+  // dialect, by name, and that is the point rather than a nuisance.** `path` is
+  // the only argument in this corpus whose *grammar* is dialect-specific:
+  // Postgres takes `path: ["a", "b"]` and refuses a string, SQLite takes
+  // `path: "$.a.b"` and refuses an array. That is Prisma's own split, measured
+  // against a generated client on both, so writing one spelling would leave the
+  // other dialect's half of the feature unwalked.
+  // `plan-key.invariants.test.ts` lists both messages in `EXPECTED_REFUSALS` and
+  // asserts each still fires, so a dialect that quietly started accepting the
+  // other form fails there.
+  //
+  // It is **not** true of the refusal block at the end, and the earlier draft of
+  // this note claimed it of the whole section: an empty path, a `path` on a
+  // String column, a malformed segment and `path: 5` are refused on *both*
+  // dialects. Which ones split and which do not is spelled out at the block
+  // itself, because it is exactly the thing that decides whether an entry needs
+  // writing twice.
 
-  // The Postgres grammar. The first four are the shape #301 opened to measure:
-  // three depths and a numeric segment, all compiling to the *same* statement —
-  // `("metadata" #>> $1) = $2` — because `#>` takes the whole path as one
-  // `text[]` parameter. Recording them element-wise minted four cache entries
-  // holding one statement, keyed by a depth that is as request-derived as an
-  // `in` list's length; `path` is in `LIST_KEYS` now, and this is what measures
-  // it.
+  // The Postgres grammar. The first three are the shape #301 opened to measure:
+  // three depths compiling to the *same* statement — `("metadata" #>> $1) = $2`
+  // — because `#>` takes the whole path as one `text[]` parameter. Recording
+  // them element-wise minted three cache entries holding one statement, keyed by
+  // a depth that is as request-derived as an `in` list's length; `path` is in
+  // `LIST_KEYS` now, and this is what measures it.
   { where: { metadata: { path: ["a"], equals: "x" } } },
   { where: { metadata: { path: ["a", "b"], equals: "x" } } },
   { where: { metadata: { path: ["a", "b", "c"], equals: "x" } } },
-  { where: { metadata: { path: ["a", 0], equals: "x" } } },
+  // An array *index*, spelled as a string. `#>` takes a `text[]` and reaches the
+  // same element either way, and the string spelling is the one #380 keeps when
+  // it narrows `JsonPath` to `readonly string[]` — this entry sits among the
+  // reads expected to compile, so a numeric segment here would report itself as
+  // an unexpected refusal the moment that lands.
+  { where: { metadata: { path: ["a", "0"], equals: "x" } } },
   { where: { metadata: { path: ["a"], not: "x" } } },
   { where: { metadata: { path: ["a"], string_contains: "x" } } },
   { where: { metadata: { path: ["a"], string_starts_with: "x" } } },
@@ -121,19 +133,45 @@ export const READS: unknown[] = [
   { where: { metadata: { path: "$.a", gt: 1 } } },
   { where: { metadata: { path: "$.a", array_contains: "x" } } },
 
-  // The refusals #299 owns, each written in **both** grammars so the message
-  // fires on the dialect that would otherwise reject the spelling first — the
-  // grammar check runs before all of them.
+  // The refusals #299 owns. Every one is written in **both** grammars, but for
+  // two different reasons, and the distinction was got wrong here first time —
+  // measured by compiling each entry on both dialects.
+  //
+  // **The first two pairs genuinely need both spellings.** `assertPathShape`
+  // runs before the bare-path and unknown-filter checks, so the array form
+  // reaches them on Postgres and is swallowed by the grammar message on SQLite,
+  // and the string form does the mirror of that. One spelling would leave one
+  // dialect's copy of the message unwalked.
   { where: { metadata: { path: ["a"] } } },
   { where: { metadata: { path: "$.a" } } },
   { where: { metadata: { path: ["a"], nonsense: "x" } } },
   { where: { metadata: { path: "$.a", nonsense: "x" } } },
+  // **The next two pairs do not**, and saying otherwise was the error. The empty
+  // check sits *above* the grammar branch inside `assertPathShape`, and the
+  // `field.type !== "Json"` throw sits above `assertPathShape` altogether, so
+  // each of these fires the same message on both dialects from either spelling.
+  // They are kept because they are the pairs whose redundancy depends on that
+  // ordering — the ordering the four entries above rely on being the other way
+  // round — so if a refactor ever hoists the grammar check to the top of
+  // `compileJsonFilter`, these are what keep both dialects covered instead of
+  // half of each. `[]` and `""` are distinct inputs in their own right besides.
   { where: { metadata: { path: [] } } },
   { where: { metadata: { path: "" } } },
   { where: { name: { path: ["a"], equals: "x" } } },
   { where: { name: { path: "$.a", equals: "x" } } },
-  // Neither grammar, which is its own branch of the shape check.
+  // Neither grammar. Not its own branch — `got` is `"other"`, which falls
+  // through to the same `got !== wanted` throw and the same message as the
+  // wrong-grammar entries above — but a distinct input class, and the one a
+  // caller reaches by forgetting to split a dotted string.
   { where: { metadata: { path: 5, equals: "x" } } },
+  // An array whose *segments* are malformed, which is the case the `LIST_KEYS`
+  // collapse has to keep out of a valid path's cache entry: refused cold on both
+  // dialects — the array grammar's own message on Postgres, the wrong-grammar
+  // one on SQLite — where `["a"]` beside it compiles. `plan-key.invariants` owns
+  // the assertion that the two do not share a key; a refused entry can never
+  // reach the same-key-implies-same-SQL property, so the corpus alone would not
+  // have caught it.
+  { where: { metadata: { path: ["a", null], equals: "x" } } },
 ];
 
 export const WRITES: [string, unknown][] = [
--- a/packages/gemi/orm/corpus.ts
+++ b/packages/gemi/orm/corpus.ts
@@ -62,6 +62,78 @@ export const READS: unknown[] = [
   { take: 5 },
   { skip: 1 },
   { skip: 1, take: 2 },
+
+  // ---------------------------------------------------------------------
+  // JSON path filters (#299), and the reason they arrived late (#301).
+  //
+  // They need a `Json` column, and the schema both invariant suites compile
+  // against did not have one — a `path` on a String column is refused outright,
+  // and a refused entry covers nothing. `userWithProfile` grew `metadata` for
+  // this; see the note there for why that cost nothing.
+  //
+  // **Every entry below is refused on exactly one dialect, by name, and that is
+  // the point rather than a nuisance.** `path` is the only argument in this
+  // corpus whose *grammar* is dialect-specific: Postgres takes
+  // `path: ["a", "b"]` and refuses a string, SQLite takes `path: "$.a.b"` and
+  // refuses an array. That is Prisma's own split, measured against a generated
+  // client on both, so writing one spelling would leave the other dialect's
+  // half of the feature unwalked. `plan-key.invariants.test.ts` lists both
+  // messages in `EXPECTED_REFUSALS` and asserts each still fires, so a dialect
+  // that quietly started accepting the other form fails there.
+
+  // The Postgres grammar. The first four are the shape #301 opened to measure:
+  // three depths and a numeric segment, all compiling to the *same* statement —
+  // `("metadata" #>> $1) = $2` — because `#>` takes the whole path as one
+  // `text[]` parameter. Recording them element-wise minted four cache entries
+  // holding one statement, keyed by a depth that is as request-derived as an
+  // `in` list's length; `path` is in `LIST_KEYS` now, and this is what measures
+  // it.
+  { where: { metadata: { path: ["a"], equals: "x" } } },
+  { where: { metadata: { path: ["a", "b"], equals: "x" } } },
+  { where: { metadata: { path: ["a", "b", "c"], equals: "x" } } },
+  { where: { metadata: { path: ["a", 0], equals: "x" } } },
+  { where: { metadata: { path: ["a"], not: "x" } } },
+  { where: { metadata: { path: ["a"], string_contains: "x" } } },
+  { where: { metadata: { path: ["a"], string_starts_with: "x" } } },
+  { where: { metadata: { path: ["a"], string_ends_with: "x" } } },
+  // The four Postgres-only filters, which SQLite's `jsonFilters` withholds
+  // because Prisma refuses them there with "Unknown argument".
+  { where: { metadata: { path: ["a"], array_contains: "x" } } },
+  { where: { metadata: { path: ["a"], gt: 1 } } },
+  { where: { metadata: { path: ["a"], gte: 1 } } },
+  { where: { metadata: { path: ["a"], lt: 1 } } },
+  { where: { metadata: { path: ["a"], lte: 1 } } },
+  // Two filters on one path — the `and`-grouped branch, and the only shape that
+  // extracts the same path twice in one statement.
+  { where: { metadata: { path: ["a"], gt: 1, lt: 9 } } },
+
+  // The SQLite grammar, same filters minus the ones the dialect declines.
+  { where: { metadata: { path: "$.a", equals: "x" } } },
+  { where: { metadata: { path: "$.a.b", equals: "x" } } },
+  { where: { metadata: { path: "$.a", not: "x" } } },
+  { where: { metadata: { path: "$.a", string_contains: "x" } } },
+  { where: { metadata: { path: "$.a", string_starts_with: "x" } } },
+  { where: { metadata: { path: "$.a", string_ends_with: "x" } } },
+  { where: { metadata: { path: "$.a", equals: "x", not: "y" } } },
+  // Refused on SQLite for the *filter* rather than for the grammar, which is a
+  // different message and the one that would fall silent if `jsonFilters` were
+  // ever widened there to match Postgres.
+  { where: { metadata: { path: "$.a", gt: 1 } } },
+  { where: { metadata: { path: "$.a", array_contains: "x" } } },
+
+  // The refusals #299 owns, each written in **both** grammars so the message
+  // fires on the dialect that would otherwise reject the spelling first — the
+  // grammar check runs before all of them.
+  { where: { metadata: { path: ["a"] } } },
+  { where: { metadata: { path: "$.a" } } },
+  { where: { metadata: { path: ["a"], nonsense: "x" } } },
+  { where: { metadata: { path: "$.a", nonsense: "x" } } },
+  { where: { metadata: { path: [] } } },
+  { where: { metadata: { path: "" } } },
+  { where: { name: { path: ["a"], equals: "x" } } },
+  { where: { name: { path: "$.a", equals: "x" } } },
+  // Neither grammar, which is its own branch of the shape check.
+  { where: { metadata: { path: 5, equals: "x" } } },
 ];
 
 export const WRITES: [string, unknown][] = [
@@ -115,6 +187,12 @@ export const WRITES: [string, unknown][] = [
   ["update", { where: { id: 1 }, data: { profile: { connect: { userId: 1 } } } }],
   ["updateMany", { where: { id: 1 }, data: { name: "x" } }],
   ["updateMany", { data: { name: "x" } }],
+  // A JSON path filter in a *write's* `where`, both grammars. The predicate
+  // compiler is shared, but its entry point is not — `compileWrite` builds its
+  // own `WhereContext` — and a `path` binder that located against the read's
+  // argument tree would only show up from here.
+  ["updateMany", { where: { metadata: { path: ["a"], equals: "x" } }, data: { name: "x" } }],
+  ["updateMany", { where: { metadata: { path: "$.a", equals: "x" } }, data: { name: "x" } }],
   ["delete", { where: { id: 1 } }],
   ["delete", { where: { id: 1 }, select: { id: true } }],
   // The conflict target has to be set by `create`, or `Model.upsert` runs it as
--- a/packages/gemi/orm/plan-key.invariants.test.ts
+++ b/packages/gemi/orm/plan-key.invariants.test.ts
@@ -119,17 +119,52 @@ describe("the plan key", () => {
   const EXPECTED_REFUSALS: Record<string, RegExp[]> = {
     // Postgres has ILIKE; SQLite has no case-insensitive LIKE for non-ASCII,
     // and Prisma refuses `mode` there too.
-    sqlite: [/mode: "insensitive"/],
-    postgres: [],
+    sqlite: [
+      /mode: "insensitive"/,
+      // The **array** path grammar, which is Postgres's. Prisma refuses it here
+      // too — the split is the client's, not this ORM's — so every array-form
+      // corpus entry is refused on this dialect and every JSONPath-string one
+      // is refused on the other. Both lists exist so that a dialect quietly
+      // accepting the form it should not fails here rather than silently
+      // gaining coverage.
+      /a JSON path is a JSONPath string/,
+      // ...and the filters `jsonFilters` withholds on SQLite — `gt` and
+      // `array_contains` in the corpus. Prisma answers "Unknown argument" for
+      // both, so this refusal is parity rather than a gap.
+      /is not available on sqlite/,
+    ],
+    postgres: [
+      // The JSONPath-**string** grammar, which is SQLite's. The mirror of the
+      // entry above.
+      /a JSON path is an array of keys/,
+    ],
   };
+
+  /**
+   * Refusals both dialects produce, kept apart from the per-dialect lists so
+   * that "this is a dialect split" and "this is refused everywhere" do not read
+   * alike.
+   */
+  const SHARED_REFUSALS = [
+    // A per-parent `take` inside a to-many needs the lateral strategy, and
+    // these compile with the default (batched) on both dialects — so it is
+    // refused everywhere here, not only on SQLite.
+    /accounts\.take/,
+    // The four JSON path refusals that are about the *argument* rather than
+    // about the database. Each is in the corpus in both grammars, because the
+    // grammar check runs before all of them — written in one spelling only,
+    // each would fire on one dialect and be swallowed by the grammar message on
+    // the other.
+    /is a String column/,
+    /A JSON path cannot be empty/,
+    /A 'path' needs a filter beside it/,
+    /A JSON path filter takes/,
+  ];
 
   test.each(DIALECTS)("every corpus entry compiles, or is refused by name — %s", (name, dialect) => {
     const { out, refused } = compiled(dialect);
 
-    // A per-parent `take` inside a to-many needs the lateral strategy, and
-    // these compile with the default (batched) on both dialects — so it is
-    // refused everywhere here, not only on SQLite.
-    const expected = [/accounts\.take/, ...EXPECTED_REFUSALS[name]];
+    const expected = [...SHARED_REFUSALS, ...EXPECTED_REFUSALS[name]];
     const unexpected = refused.filter(
       (message) => !expected.some((pattern) => pattern.test(message)),
     );
@@ -240,5 +275,58 @@ describe("the plan key", () => {
     // ...and present-versus-absent still differs, since that changes the text.
     expect(key({ take: 1 })).not.toBe(key({}));
     expect(key({ skip: 1, take: 2 })).not.toBe(key({ take: 2 }));
+  });
+
+  /**
+   * A JSON path's **depth** is bound too, which is the case #301 opened to
+   * measure rather than to reason about.
+   *
+   * Same shape as `take`/`skip` above and the opposite of the intuition: a
+   * deeper path plainly selects something else, so it looks structural. It is
+   * not. Postgres's `#>` takes the whole path as one `text[]` parameter, so
+   * `["a"]` and `["a", "b", "c"]` compile to the same `("metadata" #>> $1)`,
+   * and SQLite's path is a string, where depth was never in the shape at all.
+   *
+   * Left alone, `canonicalShape` records an array element-wise, so on Postgres
+   * every distinct depth minted its own entry holding a statement identical to
+   * its neighbours' — and a path built from a request (`?field=address.city`)
+   * varies with the data the way an `in` list's length does. `path` is in
+   * `LIST_KEYS` for that reason; this is what would notice it leaving.
+   *
+   * The SQL is asserted alongside the key deliberately. "One entry" is only
+   * correct while it is also "one statement" — assert the key on its own and a
+   * collapse that started serving the wrong text would still pass.
+   */
+  const JSON_PATHS: Record<string, unknown[]> = {
+    sqlite: ["$.a", "$.a.b", "$.a.b.c"],
+    postgres: [["a"], ["a", "b"], ["a", "b", "c"], ["a", 0]],
+  };
+
+  test.each(DIALECTS)("a JSON path's depth is bound, not compiled in — %s", (name, dialect) => {
+    const shapes = JSON_PATHS[name].map((path) => ({
+      where: { metadata: { path, equals: "x" } },
+    }));
+
+    const texts = new Set(
+      shapes.map(
+        (args) =>
+          compileRead(userWithProfile, "findMany" as never, args as never, dialect as never).text,
+      ),
+    );
+    const keys = new Set(
+      shapes.map((args) => planKey(dialect as never, "User", "findMany" as never, args)),
+    );
+
+    expect(texts.size, [...texts].join("\n")).toBe(1);
+    expect(keys.size, [...keys].join("\n")).toBe(1);
+
+    // ...and the *filter* beside the path still discriminates, which is the
+    // direction a too-eager collapse would break: these are two statements.
+    const filterKey = (filter: Record<string, unknown>) =>
+      planKey(dialect as never, "User", "findMany" as never, {
+        where: { metadata: { path: JSON_PATHS[name][0], ...filter } },
+      });
+    expect(filterKey({ equals: "x" })).not.toBe(filterKey({ not: "x" }));
+    expect(filterKey({ equals: "x" })).not.toBe(filterKey({ string_contains: "x" }));
   });
 });
--- a/packages/gemi/orm/plan-key.invariants.test.ts
+++ b/packages/gemi/orm/plan-key.invariants.test.ts
@@ -16,7 +16,7 @@ import {
   userWithProfile,
 } from "./fixtures";
 import { AGGREGATES, GROUPS, READS, vary, WRITES } from "./corpus";
-import { planKey } from "./plan";
+import { clearPlanCache, getOrCompile, planCacheStats, planKey } from "./plan";
 import * as registry from "./registry";
 
 const sqlite = new SqliteDialect();
@@ -151,10 +151,18 @@ describe("the plan key", () => {
     // refused everywhere here, not only on SQLite.
     /accounts\.take/,
     // The four JSON path refusals that are about the *argument* rather than
-    // about the database. Each is in the corpus in both grammars, because the
-    // grammar check runs before all of them — written in one spelling only,
-    // each would fire on one dialect and be swallowed by the grammar message on
-    // the other.
+    // about the database.
+    //
+    // Two of them need both grammars in the corpus and two do not, which is
+    // worth stating precisely because the first draft here claimed the first
+    // reason for all four. `assertPathShape` runs before the bare-path and
+    // unknown-filter checks, so those two are reached by the array form on
+    // Postgres and by the string form on SQLite — one spelling would leave one
+    // dialect's copy unwalked. The other two are reached from either spelling on
+    // either dialect: the `field.type !== "Json"` throw is above
+    // `assertPathShape` entirely, and the empty check is above the grammar
+    // branch inside it. Both spellings are in the corpus regardless; see the
+    // note there for why.
     /is a String column/,
     /A JSON path cannot be empty/,
     /A 'path' needs a filter beside it/,
@@ -299,7 +307,10 @@ describe("the plan key", () => {
    */
   const JSON_PATHS: Record<string, unknown[]> = {
     sqlite: ["$.a", "$.a.b", "$.a.b.c"],
-    postgres: [["a"], ["a", "b"], ["a", "b", "c"], ["a", 0]],
+    // The last is an array *index*, spelled as a string — `#>` takes a `text[]`
+    // and reaches the same element either way, and it is the spelling #380
+    // leaves standing when it narrows `JsonPath` to `readonly string[]`.
+    postgres: [["a"], ["a", "b"], ["a", "b", "c"], ["a", "0"]],
   };
 
   test.each(DIALECTS)("a JSON path's depth is bound, not compiled in — %s", (name, dialect) => {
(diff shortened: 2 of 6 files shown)


packages/gemi/orm/plan.ts
```diff
@@ -343,7 +378,14 @@ export function canonicalShape(
...
  * user-sized list actually arrives — is covered.
  */
-const LIST_KEYS = new Set(["in", "notIn", "hasEvery", "hasSome", "push"]);
+const LIST_KEYS = new Set([
+  "in",
+  "notIn",
+  "hasEvery",
+  "hasSome",
+  "push",
+  "path",
```
nstfkc, line 413: **blocking — the collapse erases the segment *types*, and `assertPathShape` refuses on exactly those.**

`collapsedList` maps any non-empty array to `[*]`, so on Postgres every array path shapes alike regardless of what is *in* it. But `assertPathShape` (`compile/where.ts:1249`) refuses an array whose segments are not all `string | number`. That refusal now becomes unreachable on a cache hit — the same "a hit skips compilation, so the refusal never re-runs" failure this file documents for `disconnect: true`/`false` a few lines below.

Measured through `planKey` + `getOrCompile` + `plan.bind` against `userWithProfile` on Postgres:

```
base (main)
  { path: ["a"] }         key …path:[string]
  { path: ["a", null] }   key …path:[string,null]   -> distinct key, refused on every call

this branch
  { path: ["a"] }         key …path:[*]
  { path: ["a", null] }   key …path:[*]             -> same key
```

and with the good path compiled first:

```
clearPlanCache();
getOrCompile(userWithProfile, "findMany", { where: { metadata: { path: ["a"], equals: "x" } } }, pg);

{ path: ["a", null] }             hit:1 compiles:0   bound ["{\"a\",NULL}", "x"]
{ path: ["limits", undefined] }   hit:1 compiles:0   bound ["{\"limits\",NULL}", "x"]
{ path: [["a"]] }                 hit:1 compiles:0   bound ["{\"a\"}", "x"]
{ path: [{ k: 1 }] }              hit:1 compiles:0   bound ["{\"[object Object]\"}", "x"]
```

Every one of those four is refused by a cold `compileRead` with *"On postgres a JSON path is an array of keys"*. On the base branch they are refused through `getOrCompile` too, because their shapes differ. So this is a regression introduced here, not a pre-existing gap.

The third row is the bad one: `[["a"]]` is *flattened* into `{"a"}` and silently answers as if the caller had written `["a"]` — wrong rows, no error. The second is the realistic arrival: a path built from a request, which is the exact motivation stated for the collapse two paragraphs up (`?field=address.city`). `JsonPath` in `types.ts` is `string | readonly (string | number)[]`, but `assertPathShape` exists precisely because paths also arrive "from a `string[]`-typed variable, from `any`, from JSON off the wire" — `types.ts:437`.

Closing it does not need a new flag; the depth still collapses, only the *kinds* stay in the shape:

```ts
if (collapseLists && LIST_KEYS.has(key) && Array.isArray(value)) {
  if (key !== "path" || value.every((s) => typeof s === "string" || typeof s === "number")) {
    return collapsedList(value);
  }
}
```

Worth a corpus entry (`{ where: { metadata: { path: ["a", null], equals: "x" } } }`) plus an assertion that it keys apart from `{ path: ["a"] }` — nothing here would have caught it, since a refused entry never reaches the same-key-implies-same-SQL property.

packages/gemi/orm/plan-key.invariants.test.ts
```diff
@@ -119,17 +119,52 @@ describe("the plan key", () => {
...
+   * alike.
+   */
+  const SHARED_REFUSALS = [
     // A per-parent `take` inside a to-many needs the lateral strategy, and
     // these compile with the default (batched) on both dialects — so it is
     // refused everywhere here, not only on SQLite.
-    const expected = [/accounts\.take/, ...EXPECTED_REFUSALS[name]];
+    /accounts\.take/,
+    // The four JSON path refusals that are about the *argument* rather than
+    // about the database. Each is in the corpus in both grammars, because the
```
nstfkc: **should fix — the reason is right for two of the four and wrong for the other two.**

Compiled every JSON corpus entry on both dialects and tabulated which message each produced:

```
{ metadata: { path: ["a"] } }               pg BARE             lite GRAMMAR
{ metadata: { path: "$.a" } }               pg GRAMMAR          lite BARE
{ metadata: { path: ["a"], nonsense } }     pg UNKNOWN-FILTER   lite GRAMMAR
{ metadata: { path: "$.a", nonsense } }     pg GRAMMAR          lite UNKNOWN-FILTER
{ metadata: { path: [] } }                  pg EMPTY            lite EMPTY
{ metadata: { path: "" } }                  pg EMPTY            lite EMPTY
{ name: { path: ["a"], equals } }           pg STRING-COL       lite STRING-COL
{ name: { path: "$.a", equals } }           pg STRING-COL       lite STRING-COL
```

`/A 'path' needs a filter beside it/` and `/A JSON path filter takes/` do need both spellings — exactly as written. The other two do not:

- `/is a String column/` — the `field.type !== "Json"` throw is *above* `assertPathShape` in `compileJsonFilter`, so it fires on both dialects from either spelling.
- `/A JSON path cannot be empty/` — the empty check is *above* the `got === wanted` check inside `assertPathShape`, so `[]` and `""` both fire it on both dialects.

So two of the four duplicate corpus entries are redundant, and "the grammar check runs before all of them" is not why they are there. Same sentence appears in `corpus.ts:124`.

packages/gemi/orm/corpus.ts
```diff
@@ -62,6 +62,78 @@ export const READS: unknown[] = [
...
+
+  // ---------------------------------------------------------------------
+  // JSON path filters (#299), and the reason they arrived late (#301).
+  //
+  // They need a `Json` column, and the schema both invariant suites compile
+  // against did not have one — a `path` on a String column is refused outright,
+  // and a refused entry covers nothing. `userWithProfile` grew `metadata` for
+  // this; see the note there for why that cost nothing.
+  //
+  // **Every entry below is refused on exactly one dialect, by name, and that is
```
nstfkc: **should fix — "every entry below is refused on exactly one dialect" is false for 11 of the 32 entries.**

It holds for the two grammar blocks (14 array-form + 9 string-form entries), which is where it matters. It does not hold for the refusal block below: `path: []`, `path: ""`, both `{ name: { path } }` spellings and `path: 5` are refused on **both** dialects, and the bare-path / unknown-filter pairs are refused on both too (for different reasons per dialect, which is the actual point). Measured by compiling each entry on both dialects.

Since the paragraph is doing real work — it is the justification for writing every shape twice — it is worth narrowing to the grammar blocks rather than claiming it of the whole section.

packages/gemi/orm/corpus.ts
```diff
@@ -62,6 +62,78 @@ export const READS: unknown[] = [
...
+  { where: { metadata: { path: "$.a", string_starts_with: "x" } } },
+  { where: { metadata: { path: "$.a", string_ends_with: "x" } } },
+  { where: { metadata: { path: "$.a", equals: "x", not: "y" } } },
+  // Refused on SQLite for the *filter* rather than for the grammar, which is a
+  // different message and the one that would fall silent if `jsonFilters` were
+  // ever widened there to match Postgres.
+  { where: { metadata: { path: "$.a", gt: 1 } } },
+  { where: { metadata: { path: "$.a", array_contains: "x" } } },
+
+  // The refusals #299 owns, each written in **both** grammars so the message
```
nstfkc: **should fix** — same correction as on `plan-key.invariants.test.ts:154`. "The grammar check runs before all of them" is true for the bare-path and unknown-filter pairs, and false for the other two: the `String column` throw is above `assertPathShape`, and the empty-path throw is above the grammar branch inside it, so both of those fire identically on both dialects from a single spelling.

packages/gemi/orm/corpus.ts
```diff
@@ -62,6 +62,78 @@ export const READS: unknown[] = [
...
+  // grammar check runs before all of them.
+  { where: { metadata: { path: ["a"] } } },
+  { where: { metadata: { path: "$.a" } } },
+  { where: { metadata: { path: ["a"], nonsense: "x" } } },
+  { where: { metadata: { path: "$.a", nonsense: "x" } } },
+  { where: { metadata: { path: [] } } },
+  { where: { metadata: { path: "" } } },
+  { where: { name: { path: ["a"], equals: "x" } } },
+  { where: { name: { path: "$.a", equals: "x" } } },
+  // Neither grammar, which is its own branch of the shape check.
```
nstfkc: **nit** — it is not its own branch. `path: 5` gives `got === "other"`, which falls through to the *same* `got !== wanted` throw with the *same* message as the wrong-grammar entries above (`GRAMMAR-array` on Postgres, `GRAMMAR-jsonpath` on SQLite — measured). Keeping the entry is fine, it is a genuinely distinct input class, but the comment promises a code path that does not exist.

nstfkc (commented):
**Verdict: one blocking correctness finding, three comment-accuracy fixes; the mechanism, the measurement and the tests are otherwise right.**

The core of this PR holds up under mutation. I reverted both things it claims to cover and the new tests caught both, on the dialects the PR says they would:

| mutation | result here |
| --- | --- |
| `"path"` out of `LIST_KEYS` | `plan-key.invariants.test.ts:321` — `expected 4 to be 1`, printing `path:[string]`, `[string,string]`, `[string,string,string]`, `[string,number]` over one statement |
| same, live Postgres | `differential.test.ts:1761` — `expected 2 to be 1` compiles |
| `const compiledIn = filter.path; const path: Binder = () => compiledIn` | `binding.invariants.test.ts:199` fails on **both** dialects — `reused ["{\"a\"}",978]` vs `fresh ["{\"ZZZZ\"}",978]`, and the `updateMany` corpus entry fails too, so the write entry pulls its weight |

The corpus-wide binding property is the right call and is not padding: without it the mutation above survives everything in the repository. The name-collision measurement reproduces exactly on a scratch schema (Json `path` → `"path" = $1::text::jsonb` for 1 and 2 elements alike; `String[]` `path` → bare array refused in `where`, one `values` placeholder in `data`), so the collapse is correct there rather than merely unreachable, as the docblock says.

## Findings

**1 — blocking.** Collapsing `path` to `[*]` erases the *element types*, and `assertPathShape` refuses a segment that is not a string or a number. So a warm plan compiled from a valid path now serves an invalid one and the refusal never re-runs — the `disconnect: true`/`false` failure class this very docblock describes two paragraphs down. Measured, details inline on `plan.ts`. Cheap to close by gating the `path` collapse on `value.every(s => typeof s === "string" || typeof s === "number")`.

**2 — should fix.** The "each refusal is in the corpus in both grammars, because the grammar check runs before all of them" rationale — written twice, in `corpus.ts` and in `SHARED_REFUSALS` — is wrong for two of the four. The Json-column check runs *before* `assertPathShape`, and the empty check runs *before* the grammar check inside it, so `{ name: { path } }` and `{ metadata: { path: [] } }` / `{ path: "" }` fire the same message on both dialects from either spelling. Those two duplicate entries are redundant and the stated reason does not apply to them.

**3 — should fix.** `corpus.ts`'s "**Every entry below is refused on exactly one dialect**" is false for 11 of the 32 entries.

**4 — nit.** `path: 5` is not "its own branch of the shape check" — it reaches the same `got !== wanted` throw with the same message as the wrong-grammar entries.

## What I ran

Worktree, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi` (confirmed before trusting any template number). `bun run build:bin`, `gemi-orm-generator` linked into both `node_modules/.bin`, both `prisma generate`s.

```
packages/gemi   bun --bun vitest run orm    64 files / 2425 passed
                bun run typecheck            clean
saas-starter    TZ=UTC bun --bun vitest run app/models
                                             20 passed | 3 skipped; 793 passed
  postgres      app/models/postgres.sh -t postgres app/models
                                             14 passed | 9 skipped; 874 passed
```

All three match the PR body's numbers. `gh pr checks 385`: both jobs green.

## What I did not check

- The fixture claim that the template's `User` declares `metadata Json?` — verified by reading `prisma/schema.prisma`; the seed does give exactly one user `{ plan: "pro", limits: { seats: 3 } }`, so the differential pair really is 1 row / 0 rows rather than 0 / 0.
- `tsconfig.generated.json` and `bunx gemi check models` — not run here.
- Nothing about a third dialect. The docblock's stated safety net for a dialect that spells a path as an array but binds per segment is `plan-key.invariants.test.ts`, which only walks whatever is in `DIALECTS` — true, but conditional on that dialect being added to the corpus suites at the same time. Worth a clause at the line, not a change.

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs.

## #380 × #385 — a numeric path segment: one PR refuses it, the other ships four (blocking for whichever lands second)

Neither PR touches a file the other touches in `packages/gemi/orm` (#380: `compile/where.ts`,
`types.ts`; #385: `corpus.ts`, `plan.ts`, `plan-key.invariants.test.ts`, `fixtures.ts`). They merge
with **no textual conflict**, and the merged tree fails. This one was not on anyone's list — it is
not in `orm.md`'s notes and neither per-PR review could have seen it.

**#380** makes `assertPathShape` refuse a non-string path segment and narrows `JsonPath` to
`readonly string[]`. **#385** ships `["a", 0]` in three places, deliberately and with reasoning
attached.

Measured on `main` + #380 + #385 (`bun --bun vitest run` in `packages/gemi`):

```
FAIL  orm/plan-key.invariants.test.ts > the plan key > every corpus read compiles the same way
+ [
+   "findMany {\"where\":{\"metadata\":{\"path\":[\"a\",0],\"equals\":\"x\"}}} — gemi ORM does not
+    support 'metadata.path' yet (User.findMany). A JSON path is an array of strings, and path[1]
+    is a number. …",
+   "findFirst … (same)",
+ ]

FAIL  orm/plan-key.invariants.test.ts > a JSON path's depth is bound, not compiled in — postgres
UnsupportedQueryError: … A JSON path is an array of strings, and path[1] is a number.
  ❯ assertPathShape orm/compile/where.ts:1365
```

**Three sites in #385 to respell**, all to `"0"`:

1. `packages/gemi/orm/corpus.ts` — `{ where: { metadata: { path: ["a", 0], equals: "x" } } }`,
   which sits among the reads expected to *compile*, so #385's own corpus-wide invariant reports it
   as an unexpected refusal.
2. `packages/gemi/orm/plan-key.invariants.test.ts` — `postgres: [["a"], ["a","b"], ["a","b","c"], ["a", 0]]`.
3. `packages/gemi/orm/plan.ts` — the new `LIST_KEYS` docblock states *"Measured over the corpus:
   `["a"]`, `["a","b"]`, `["a","b","c"]` and `["a", 0]` were four cache entries holding one
   statement."* That sentence becomes false: the fourth no longer compiles at all. A prose fix, but
   it is the same defect class #383 is fixing elsewhere in this batch.

`["a", "0"]` is the right respelling and costs the tests nothing — `#>` takes a `text[]`, so it
reaches the same element, which is exactly #380's own argument. I applied it and the package suite
goes **3185 passed / 1 skipped**.

### One thing worth deciding rather than patching

#385 already carries a blocking finding of its own: collapsing `path` into `LIST_KEYS` erases the
array's element types, so `assertPathShape`'s refusal is unreachable on a warm plan and a malformed
path binds where a cold compile refuses it.

**#380 widens exactly that hole.** It *adds* a refusal to `assertPathShape` (numeric segments) —
so after both land there is strictly more that a cold compile rejects and a warm plan-key hit
sails past. `["a", 0]` becomes a concrete instance: refused cold, bound warm off the `["a","b"]`
plan. Whoever addresses #385's blocking finding should know the target got bigger, and the
respelled corpus entry no longer exercises it.

---

**Merge-order note.** The conflict is symmetric — it appears whichever of the two lands second, and
I confirmed the textual merge is clean in both directions. The repair belongs in #385 (it is #385's
corpus and #385's test), so the cheapest order is **#380 before #385**.

Built the full ten-PR integration branch locally with this repair applied: `typecheck` clean,
`packages/gemi` **3185 passed**, template sqlite **845 passed**, `test:types` **117 passed**, live
Postgres via `postgres.sh` **1247 passed** (the only 2 failures being the harness's designed
one-dialect-at-a-time refusal).

packages/gemi/orm/plan.ts
nstfkc: **Addressed** — you are right, and the reproduction lands exactly as written. Confirmed here before touching anything:

```
["a"]           key …path:[*]   compiles
["a","b"]       key …path:[*]   compiles
["a",null]      key …path:[*]   REFUSED cold: "On postgres a JSON path is an array of keys…"
["a",undefined] key …path:[*]   REFUSED cold
[["a"]]         key …path:[*]   REFUSED cold
[{"k":1}]       key …path:[*]   REFUSED cold
[]              key …path:[]    REFUSED cold (empty check, distinct shape)
```

Four shapes sharing a key with a valid path, each refused by a cold compile. On `main` they key element-wise (`[string,null]`, `[[string]]`, …) and are refused on every call, so this is a regression the branch introduced rather than a pre-existing gap — as you said.

**Fixed in 80f99a8**, with one deliberate difference from your patch: the predicate is `typeof segment === "string"`, not `string | number`.

```ts
if (collapseLists && LIST_KEYS.has(key) && Array.isArray(value)) {
  if (key !== "path" || value.every((segment) => typeof segment === "string")) {
    return collapsedList(value);
  }
}
```

The reason is stated at the line: the predicate has to stay a **subset** of what `assertPathShape` accepts, not an equal. A collapsed shape covering an argument the compiler would refuse is a correctness bug; an accepted argument left uncollapsed costs one cache entry and nothing else. `string | number` is exactly today's accepted set, so it is tight — but #380 narrows `JsonPath` to `readonly string[]` and makes `assertPathShape` refuse a numeric segment, at which point `string | number` becomes *wider* than the accepted set and re-opens this same hole for `["a", 0]`. Your cross-PR comment on this PR says precisely that ("#380 widens exactly that hole"). `string` is correct before and after, with no second edit and no dependency on #380 landing.

The cost of the narrowing is that a numeric segment keeps its element-wise shape today. That is a missed collapse on a spelling #380 is about to outlaw, and it is documented at the line.

**Tests.** Rather than only a corpus entry — which, as you note, can never reach the same-key-implies-same-SQL property once it is refused — there is a new property in `plan-key.invariants.test.ts`:

> `a path a cold compile refuses is not served by a warm plan — %s`

over `["a", null]`, `["a", undefined]`, `["a", true]`, `[["a"]]`, `[{ k: 1 }]`, per dialect. It asserts three things: each is refused by a cold `compileRead` (so the test cannot go vacuous if the compiler ever accepts them), each keys apart from the valid path, and — the actual property — after warming `getOrCompile` with `["a"]`, every one of them still throws, with `planCacheStats().hits === 0`. The key comparison says why; the cache run says what.

`{ where: { metadata: { path: ["a", null], equals: "x" } } }` is in the corpus too, with a note saying what it is for and that the corpus alone would not have caught this. It needs no `EXPECTED_REFUSALS` change: it fires the array-grammar message on Postgres and the wrong-grammar one on SQLite, both already listed.

**Mutation check.** Reverting the gate to an unconditional `collapsedList`:

```
FAIL  a path a cold compile refuses is not served by a warm plan — postgres
AssertionError: ["a",null]: expected 'postgres:batched:User:findMany:{where…' not to be '…'
```

and the existing depth test still fails when `"path"` leaves `LIST_KEYS` — now `expected 3 to be 1` rather than 4, see the respelling reply.

packages/gemi/orm/plan-key.invariants.test.ts
nstfkc: **Addressed** in 80f99a8 — your table reproduces exactly. I recompiled every JSON corpus entry on both dialects and got the same split:

```
STRINGCOL  pg ["a"]   "A 'path' filter reads…"      lite ["a"]   "A 'path' filter reads…"
STRINGCOL  pg "$.a"   "A 'path' filter reads…"      lite "$.a"   "A 'path' filter reads…"
EMPTY      pg []      "A JSON path cannot be…"      lite []      "A JSON path cannot be…"
EMPTY      pg ""      "A JSON path cannot be…"      lite ""      "A JSON path cannot be…"
BARE       pg ["a"]   "A 'path' needs a…"           lite ["a"]   GRAMMAR
BARE       pg "$.a"   GRAMMAR                       lite "$.a"   "A 'path' needs a…"
UNKNOWN    pg ["a"]   "A JSON path filter takes…"   lite ["a"]   GRAMMAR
UNKNOWN    pg "$.a"   GRAMMAR                       lite "$.a"   "A JSON path filter takes…"
```

The stated reason is right for the bottom two pairs and wrong for the top two, exactly as you have it. The comment here now says which two need both spellings and which two do not, and why: `assertPathShape` runs before the bare-path and unknown-filter checks, the `field.type !== "Json"` throw runs before `assertPathShape` entirely, and the empty check runs before the grammar branch inside it.

**I kept the two redundant pairs**, and said so at the corpus rather than leaving it implicit. They are the pairs whose redundancy is *contingent on that ordering* — the same ordering the other two pairs depend on being the other way round — so if a refactor ever hoists the grammar check to the top of `compileJsonFilter`, they are what keeps both dialects covered instead of half of each. `[]` and `""` are also distinct inputs in their own right rather than two spellings of one thing. Two lines against the corpus's own "a case costs one line, and both failure modes it guards are silent" seemed the right trade; if you would rather they went, say so and I will drop them.

Same correction applied at `corpus.ts:124`.

packages/gemi/orm/corpus.ts
nstfkc: **Addressed** in 80f99a8 — the claim was overreaching and I confirmed it the same way you did, by compiling each entry on both dialects. `path: []`, `path: ""`, both `{ name: { path } }` spellings, `path: 5` and (new) `path: ["a", null]` are refused on *both*; the bare-path and unknown-filter pairs are refused on both as well, for different reasons per dialect, which is the actual point and is not what the sentence said.

The paragraph is now scoped to the two **grammar** blocks, where it holds and where it is doing the work of justifying two spellings. A second paragraph says plainly that it is *not* true of the refusal block, and points at that block — which now spells out, pair by pair, which entries split by dialect and which do not, since that is exactly the thing that decides whether an entry needs writing twice.

packages/gemi/orm/corpus.ts
nstfkc: **Addressed** in 80f99a8, alongside the matching correction in `plan-key.invariants.test.ts`.

The comment no longer offers one reason for all four pairs. It now separates them:

- **the bare-path and unknown-filter pairs** genuinely need both spellings, because `assertPathShape` runs before both checks — the array form reaches them on Postgres and is swallowed by the grammar message on SQLite, and the string form mirrors that;
- **the empty-path and String-column pairs** do not, because the `field.type !== "Json"` throw is above `assertPathShape` altogether and the empty check is above the grammar branch inside it, so each fires the same message on both dialects from either spelling.

Both spellings stay in the corpus, with the honest reason written down: those two pairs are the ones whose redundancy depends on an ordering that the *other* two pairs depend on being the other way round, so they are what keeps both dialects covered if that ordering ever changes. Verified by compiling every entry on both dialects rather than by reading, same as your table.

packages/gemi/orm/corpus.ts
nstfkc: **Addressed** in 80f99a8 — you are right, and I confirmed the message rather than the code path:

```
FIVE pg    "…On postgres a JSON path is an array of keys — ["a", "b"] — not a JSONPath string…"
FIVE lite  "…On sqlite a JSON path is a JSONPath string — "$.a.b" — not an array of keys…"
```

Byte-identical to what the wrong-grammar entries produce on each dialect, because `got` is `"other"` and falls through to the same `got !== wanted` throw. The comment promised a branch that does not exist.

Reworded to say exactly that, and to say what the entry *is* worth keeping for: a distinct input class, and the one a caller reaches by forgetting to split a dotted string. Entry kept.

nstfkc:
## Review round 1 — 80f99a8

Seven comments: the review summary, five inline, and the **[cross-PR review]**. **All seven addressed, none declined, one item deferred** (the #386 rename, which is not mine to make). Each inline comment has its own reply carrying the measurement behind it; this is the shape of the whole.

### 1 — blocking. The collapse erased the segment *kinds*, not only the depth. Fixed.

You were right, and the reproduction is exact. Before touching anything:

```
["a"]           key …path:[*]   compiles
["a","b"]       key …path:[*]   compiles
["a",null]      key …path:[*]   REFUSED cold
["a",undefined] key …path:[*]   REFUSED cold
[["a"]]         key …path:[*]   REFUSED cold
[{"k":1}]       key …path:[*]   REFUSED cold
```

Four shapes sharing a cache entry with a valid path, every one refused by a cold `compileRead`. `[["a"]]` is the one that matters — it flattens into `{"a"}` and returns `["a"]`'s rows, so the failure is wrong rows rather than a wrong error. On `main` all four key element-wise and are refused on every call, so this is a regression the branch introduced, exactly as the comment says.

`path` now collapses only when every segment is a `string`. **That is narrower than the `string | number` you proposed, deliberately**, and the reasoning is written at the line: the predicate must stay a *subset* of what `assertPathShape` accepts rather than an equal. A collapsed shape covering an argument the compiler refuses is a correctness bug; an accepted argument left uncollapsed costs one cache entry and nothing else. `string | number` is exactly today's accepted set, so it is tight *today* — but #380 makes `assertPathShape` refuse a numeric segment, at which point `string | number` becomes wider than the accepted set and re-opens this hole for `["a", 0]`. Your own cross-PR comment says so ("#380 widens exactly that hole"). `string` is correct on both sides of that merge with no second edit.

Closed by a new property rather than only by a corpus entry, because — as you note — a refused entry can never reach the same-key-implies-same-SQL invariant:

> `a path a cold compile refuses is not served by a warm plan — %s`

Five malformed shapes per dialect. Each asserted refused cold (so it cannot go vacuous if the compiler ever accepts them), each keyed apart from the valid path, and — the property itself — each still throwing after `getOrCompile` has been warmed with `["a"]`, with `planCacheStats().hits === 0`. `{ path: ["a", null] }` is in the corpus too, needing no `EXPECTED_REFUSALS` change: it fires the array-grammar message on Postgres and the wrong-grammar one on SQLite, both already listed.

Mutation-checked. Reverting the gate fails the new test on Postgres (`["a",null]` keys the same as `["a"]`). The two original mutations still fail as before — `"path"` out of `LIST_KEYS` fails the depth test (`expected 3 to be 1` now, see below), and the compiled-in-path binder fails the corpus-wide binding property on both dialects.

### 2, 3, 4 — comment accuracy. All three confirmed and corrected.

I recompiled every JSON corpus entry on both dialects rather than re-reading, and your table reproduces line for line.

- **The "grammar check runs before all of them" rationale is right for two pairs and wrong for two.** `A 'path' needs a filter beside it` and `A JSON path filter takes` do need both spellings; `is a String column` (thrown above `assertPathShape`) and `A JSON path cannot be empty` (thrown above the grammar branch inside it) fire the same message on both dialects from either spelling. Corrected in both places it was written. **The redundant entries are kept** — they are the pairs whose redundancy is contingent on that ordering, which is the same ordering the *other* two pairs rely on being the other way round, so they are what keeps both dialects covered if a refactor ever hoists the grammar check. That is now the stated reason instead of the wrong one. Happy to drop them if you would rather.
- **"Every entry below is refused on exactly one dialect" is false for the refusal block**, true for the two grammar blocks. Scoped to those, with a second paragraph saying plainly that it does not hold below and pointing at the block, which now spells out pair by pair which entries split by dialect and which do not.
- **`path: 5` is not its own branch.** Measured: byte-identical message to the wrong-grammar entries on each dialect. Reworded to say that, and to say what the entry is actually worth keeping for.

### Cross-PR (#380 × #385) — the numeric segment, respelled here

Three sites, all in this branch, all done: `corpus.ts`, `plan-key.invariants.test.ts`'s `JSON_PATHS`, and the `plan.ts` `LIST_KEYS` docblock. `["a", 0]` → `["a", "0"]`; `#>` takes a `text[]` and reaches the same element, which is #380's own argument.

One consequence worth recording rather than hiding: `["a","0"]` shapes to `[string,string]` element-wise, the same as `["a","b"]`, so the *pre-collapse* count in the docblock and the PR body drops from four entries to **three**. Restated as three depths / three entries, re-measured, and the `"path"`-out-of-`LIST_KEYS` mutation now reads `expected 3 to be 1`. The PR body carries the same correction with a note saying why — a stale measurement in a docblock is the defect class #383 is fixing elsewhere in this batch.

On your "the target got bigger" note: that is precisely why the gate is `string` rather than `string | number`. After #380 lands, `["a", 0]` is refused cold *and* keeps its element-wise shape here, so it is refused warm too.

### Also from the review's "what I did not check"

The third-dialect safety net now states its actual precondition: `plan-key.invariants.test.ts` catches a dialect that spells a path as an array but binds per segment **conditional on that dialect being added to `DIALECTS` there** — the same condition every other property in that file already carries. A clause at the line, as suggested, not a change.

### Deferred

**The `childField` → `childFields` rename from #386.** #386 is not on `main`, so nothing on this branch can reference the new names; it is merge-time work for whoever lands #386 last. This PR touches neither identifier in any case. The `connectOrCreate` hit-branch short-circuit is #378's by the cross-PR review's own assignment and is not reached from here.

### Verification, all re-run on 80f99a8

```
packages/gemi   bun --bun vitest run           120 files / 3116 passed | 1 skipped   (base 3110)
                bun --bun vitest run orm        64 files / 2427 passed
                bun run typecheck               clean
                tsc -p tsconfig.generated.json  clean
                oxlint orm --deny-warnings      0 warnings, 0 errors
saas-starter    TZ=UTC vitest run app/models    20 passed | 3 skipped; 793 passed
                bun run test:types              116 passed, no type errors
  postgres      app/models/postgres.sh -t postgres app/models
                                                14 passed | 9 skipped; 874 passed
```

Worktree, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`, confirmed before trusting any template number. `build:bin`, `gemi-orm-generator` linked into both `node_modules/.bin`, both `prisma generate`s. Schema restored to sqlite, `git status` clean apart from the three changed files.

`Closes #301` was already in the body and `closingIssuesReferences` resolves to 301 — checked, no fix needed here.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `80f99a86` for
this PR.

### Conflict 3 (#380 × #385), numeric-segment half — **resolved by your respelling**

`["a", "0"]` is in `corpus.ts`, `plan-key.invariants.test.ts`'s `JSON_PATHS` and the `plan.ts`
`LIST_KEYS` docblock; no numeric segment is left anywhere on the merged tree. Mutation-checked
rather than read: reverting the corpus entry to `["a", 0]` fails
`every corpus entry compiles, or is refused by name — postgres` with #380's numeric-segment message,
and restoring it goes green. Your comment at `corpus.ts:101` — "a numeric segment here would report
itself as an unexpected refusal the moment that lands" — is measurably true on the integrated tree.

### New, from the same pair: your `["a", null]` entry is refused by a message #380 renamed

**This is the one thing on this PR that fails.** On the merged tree:

```
FAIL orm/plan-key.invariants.test.ts > every corpus entry compiles, or is refused by name — postgres
  findMany {"where":{"metadata":{"path":["a",null],"equals":"x"}}} —
  A JSON path is an array of strings, and path[1] is null. Prisma's path is string[] too.
```

The mechanism, measured on both branch versions of `where.ts`. On this branch alone, a malformed
segment falls off the end of `assertPathShape` into the **grammar** throw, so `["a", null]` is
refused on Postgres with *"a JSON path is an array of keys"* — which is already in
`EXPECTED_REFUSALS.postgres`. That is what made your "needs no `EXPECTED_REFUSALS` change" true, and
it was true. #380 gives the segment case a sentence of its own, so the entry stops matching either
list and reports as an unexpected refusal.

Same class as the original #380 × #385: disjoint files, silent merge, a test failure neither PR can
see alone. It is new — it did not exist on the previous pass, because the `["a", null]` corpus entry
is something this review round added.

**Resolution, verified.** One pattern in `EXPECTED_REFUSALS.postgres`, and the exact spelling
matters:

```ts
/A JSON path is an array of strings, and path\[\d+\] is null/,
```

Postgres-only and not by choice — SQLite refuses the array form on the grammar branch *above* the
segment check and never looks inside, so this message only ever fires on one dialect.

Do **not** widen it to `/A JSON path is an array of strings/`. I tried that first and it passes with
your `["a", "0"]` entry reverted to `["a", 0]` — #380 emits one message for every malformed segment
kind, so the broad pattern silently excuses the numeric case and destroys the tripwire the paragraph
above just confirmed. The narrow one fails on that mutation, which is how I caught it.

### Also stale after the merge (comment only)

`corpus.ts:167-173` says the `["a", null]` entry is refused with "the array grammar's own message on
Postgres". Accurate on this branch, wrong once #380 lands — it is the segment message there. Worth
correcting in the same edit, since it is the sentence that explains why the entry needs no list
change.

With the pattern above applied: `typecheck` clean · `packages/gemi` 120 files / 3199 passed / 1
skipped · template SQLite 21 files / 848 passed / 134 skipped · `test:types` 118 passed, no type
errors · `postgres.sh` 1251 passed / 520 skipped (plus the two SQLite-half differential suites the
harness fails by design) · `oxlint orm --deny-warnings` 0/0 · both `prisma generate` runs leave
`git status` clean.

Closing keyword verified: `closingIssuesReferences` resolves to #301.

PR #386, opened 2026-08-09
Close #271: a nested write contributes every joined field

### LINKED ISSUE #271: Multi-field relations in a nested write are refused, and the refusal is pending rather than declined

nstfkc:
Closes #271

The last **pending** entry on `docs/orm.md`'s *Not in scope* list. `@relation(fields: [tenantId, orderId], references: [tenantId, id])` worked everywhere a relation is *read* — `include` under either strategy, a relation filter, `_count`, an `orderBy` through a relation, and #97's Postgres `unnest` form — and was refused wherever one is *written*. `planNestedWrites` narrowed the link through `singleFieldLink`, which raised by name rather than joining on the first field and writing plausible wrong rows.

The refusal was honest and it was the only thing left on that list. It is gone.

## What Prisma actually does, and how that was established

The template's schema grew two models (below), `prisma db push` to a throwaway SQLite file, a generated 6.19.2 client, `log: [{ emit: "event", level: "query" }]`, and the table read back after each call.

**`connect` by the compound key** — `LedgerEntry.create({ data: { amount: 1, ledger: { connect: { tenantId_code: { tenantId: 1, code: "ab" } } } } })`:

```
BEGIN IMMEDIATE
select Ledger.tenantId, Ledger.code    where tenantId = ? and code = ?     [1, "ab"]
insert into LedgerEntry (tenantId, ledgerCode, amount) values (?,?,?)      [1, "ab", 1]
select the row back
COMMIT
```

**`connect` onto an occupied composite one-to-one** — seal 2 holds `(1, "ab")`, seal 1 holds `(1, "a")`, and seal 2 is pointed at `(1, "a")`:

```
BEGIN IMMEDIATE
select Ledger.tenantId, Ledger.code                    resolve the operand
select LedgerSeal.* where id = 2                       the row being written
select Ledger.*     where (tenantId, code) in ((1,"ab"))   what it holds now
select LedgerSeal.id, tenantId, ledgerCode
       where (tenantId, ledgerCode) in ((1,"a"))       the incumbent
update LedgerSeal set tenantId = ?, ledgerCode = ? where id in (1)   [null, null]
update LedgerSeal set tenantId = ?, ledgerCode = ? where id in (2)   [1, "a"]
COMMIT
```

…and the table afterwards:

```
{ id: 1, tenantId: null, ledgerCode: null, seal: "mine"    }   orphaned, not deleted
{ id: 2, tenantId: 1,    ledgerCode: "a",  seal: "sibling" }
```

**Foreign-side `create` onto an occupied one** is the same shape from the other end: the incumbent's `UPDATE` sets **both** columns to null, then the new row is inserted with both set.

Three things follow, and each is a decision in the diff:

| measured | decision |
| --- | --- |
| Prisma issues the resolving `select` for a compound `connect` | the composite `connect` takes the lookup; the zero-query shortcut stays single-field and parity is unchanged |
| the detach writes `set tenantId = null, ledgerCode = null` | `clearLinks` nulls **every** column, and names all of them as ORM-authored for #98 |
| `@@unique([tenantId, ledgerCode, seal])` and `@@unique([ledgerCode, tenantId])` are each **P1012** — *"Either add an `@@unique([tenantId, ledgerCode])` attribute to the model, or change the relation to one-to-many"* | `displaces` requires an index covering **exactly** the relation's fields |

## The diff

`planOne` resolves the full `Link` and no longer narrows. Both planners take it, and forty hand-spelled `{ [childField]: parent[parentField] }` / `select: { [childField]: true }` literals become four helpers:

```
childLink(link, row)   { <childField>: row[<parentField>], … }
keySelect(fields)      { <field>: true, … }
nullKey(fields)        { <field>: null, … }
isLinked(values)       whether the tuple names a row at all
```

Generalising each of the forty separately is how a composite relation ends up correlating on every field in nine operands and on the first field in the tenth — a silent wrong write, which is precisely what the refusal was protecting against. One function per shape means the count of fields is invisible to the callers, and there is no single-field spelling left for a new operand to copy.

`childLink` serves both sides with the same arithmetic. On the foreign side it reads as *"belonging to this parent"*; on the owning side `link.parentFields` are this row's foreign key and `link.childFields` are what they reference, so the identical call yields the far row's own columns.

**`singleFieldLink` survives with exactly one caller**: the implicit many-to-many join table. Prisma builds that from the two models' *primary keys*, one column each side, into a table with no schema — so "one field" there is an invariant of the shape rather than a limit of this compiler, and the function states it rather than letting `planJoinTable` index into `[0]` with nothing checking it. Its message and docblock say that now instead of announcing a gap.

### A null column means "not linked"

`(tenantId, orderId) = (1, NULL)` is `UNKNOWN`, so a partial key joins to no row. Every place that used to test one value for null — *is anything linked here* — goes through `isLinked`:

- `displaceSibling` returns (the composite generalisation of the `value === null` return it already had);
- `clearLinks` refuses to run, because `where: { tenantId: 1, orderId: null }` matches every *unlinked* row in the tenant and the `updateMany` behind it would write null over null across all of them;
- the owning side's `update` reports the row missing rather than filtering on the half that is set;
- `disconnect`'s filter arm treats it as nothing-linked, which is silent, as Prisma is.

Nothing reaches `clearLinks` with a null today, so that one buys no behaviour; it means a caller that one day does cannot silently rewrite the table.

### The two `displaces` predicates

Both generalise, and neither changes shape. The foreign side already read `relation.kind` and now requires **every** child column nullable. The owning side (#363) still takes both halves — the far side's `kind` *and* the index — and the index test moves from `unique.length === 1 && unique[0] === fkField` to set equality against the relation's fields.

Written as set equality even though Prisma requires the order to match (measured above), because the length check is what carries the weight and a positional loop would read as though order were a thing this had decided.

### Where a partial write could live

The owning side folds its columns into the caller's own `INSERT`/`UPDATE` — `contributions` is keyed by field and `insertColumns`/`updateAssignments` fold every entry into one statement — so "some keys written, some not" is not a state it can reach. The hazard is real on the **foreign** side, where each linked row is its own statement; `$exec` opens a transaction for any plan carrying steps, which is the same guarantee that already covers a `connect` of three rows where the third does not exist.

### `assertDisconnectable`

Takes the whole key and names the first required column. Prisma refuses a mixture of optional and required relation fields, so on a schema it accepted the columns agree and any one of them answers — checking all of them costs nothing and means the refusal does not depend on a rule enforced in another program.

## The template

`LedgerEntry.ledger` is **required**, which is the ordinary tenant-scoped shape and the one #67's read surfaces are measured on. Prisma leaves `disconnect` out of a required relation's nested input type entirely, so every operand that *detaches* a composite link was unreachable from this schema — the same "the harness cannot get there" problem `Profile` fixed one shape down (#358).

So two models, and a migration that is verbatim `prisma migrate diff` output:

- **`LedgerNote`** — the same two columns, nullable. Reaches `set` and `disconnect`.
- **`LedgerSeal`** — those columns plus `@@unique([tenantId, ledgerCode])` beside a non-list back-relation, which is the composite one-to-one. Reaches displacement on both sides.

`differential.ts`'s SQLite `clearAll` did not clear the Ledger tables at all — present in the Postgres `TABLES` list, absent from the `deleteMany` branch. That cost nothing while the only composite cases were reads against a temp database that is fresh per file; a write comparison resets *between* the two runs, so it had to be fixed before any of this could pass.

## Tests

**Compile level** (`composite-relations.test.ts`, +26 cases). Nested writes have no SQL to assert for a link, so the planned steps are driven with a recording `RelationExecutor` and the arguments they hand the child's own operations are what is checked. Covers all four shapes: owning/required, owning/one-to-one, foreign/to-many, foreign/optional.

Mutation-checked rather than assumed — each helper crippled to its first field in turn:

| mutation | compile suite | differential suite |
| --- | --- | --- |
| `childLink` first field only | 10 failed | 14 failed |
| `keySelect` first field only | 3 failed | — |
| `nullKey` first field only | 4 failed | — |
| owning-side contributions, first field only | — | 8 failed |

**Differential** (`writes.differential.test.ts`, +30 cases, both dialects). The failure mode here does not raise: a stamp that wrote `tenantId` and left `ledgerCode` alone attaches the row to *no* ledger while returning a plausible payload, and a filter conjoined on one column reaches the sibling ledger's children. Both are green against a client that only checks what came back, which is why every case names both tables and reads them back.

The seed is arranged so a partial join is *wrong* rather than empty: code `"a"` exists in two tenants, tenant 1 also has `"ab"` (so a concatenating stitch confuses `(1,"ab")` with `(1,"a")+"b"`), and each child model has a row attached elsewhere and a row attached to nothing.

`composite-surfaces.test.ts`'s *"a nested write is still refused, by name"* said *"when this starts compiling, this test fails — which is the prompt to move the entry again"*. It compiled; the entry moved; the test now asserts that the parent statement returns **both** key columns and hides both — return one and the repoint writes half a key, which is the first-field join one layer down.

## Verification

- `packages/gemi`: typecheck, `tsc -p tsconfig.generated.json`, 3128 tests, `oxlint orm --deny-warnings` at zero.
- `templates/saas-starter`, SQLite: 822 tests, `test:types`, `gemi check models` (16 registered).
- `templates/saas-starter`, **Postgres 16** in a container, following the CI job exactly — provider flipped, schema re-derived, client regenerated, `db push` — `vitest run app/models --no-file-parallelism -t postgres`: 803 passed, all 30 composite cases among them.

## Not done here

`upsert` on the *owning* side is still refused by name, on both single-field and composite relations, for the reason it always was: an absent far row would have to be created and then written back to a row that has already been inserted. Unchanged and unrelated to the key's width.

---

## Review round 1

**Two relations that share a foreign-key column are now refused by name.** A composite relation joins on *n* columns and nothing stops two of them from sharing one — `prisma validate` accepts `ledger Ledger @relation(fields: [tenantId, ledgerCode], …)` beside `note Ledger @relation(fields: [tenantId, noteCode], …)` on 6.19.2. Before this PR both were refused for their width, so writing through both at once was unreachable; it is reachable now, and both contribute a `tenantId`. `insertColumns` folds contributions through a `Map` keyed by field and `planNestedWrites` sorts its relation keys, so the alphabetically-last relation decided the shared column. Prisma resolves the operands in the opposite direction and lets the *first resolved* win, so the caller's **last** `data` key decides it — a different `Ledger`, no error either side.

Reproducing that means giving up the sorted key order, which the plan cache needs (two argument objects differing only in key order have to be one plan). So `planNestedWrites` records which relation contributed each column and refuses the second, which is what the width refusal was buying: no plausible wrong row.

**The three "every column is nullable" predicates are pinned.** `planOwningSide`'s `displaces`, `planForeignSide`'s `displaces` and `assertDisconnectable` all stayed green when crippled to `fields[0]`, because Prisma makes a composite relation optional or required as a whole and no schema it validates distinguishes them. `ledgerSealMixed` is the hand-built `ModelSchema` that does — `tenantId` nullable, `ledgerCode` required, the nullable one deliberately first — and each of the three goes red under its own mutation now.

**`docs/orm.md`'s operand list was wider than the code.** The owning side refuses `createMany`, `set`, `delete`, `upsert` and the two `*Many` by name, unchanged by this PR, so the paragraph names the two sides separately rather than claiming "both sides" of a single list. The shared-column refusal is documented as a fourth bullet. Also fixed in passing, and pre-existing: *"Both of the last two throw `UnsupportedByDesignError`"* now names `distinct` and `cursor`, which is what `PERMANENTLY_REFUSED` actually holds.

Verification, on top of the above: `packages/gemi` typecheck clean, **3135 passed** / 1 skipped over 120 files, `oxlint orm --deny-warnings` 0/0; template SQLite **822 passed** / 133 skipped, `test:types` 116 with no type errors; both `prisma generate` runs leave `git status` clean. Each new test mutation-checked — the guard disabled fails 2, and the three nullability narrowings fail 1 each.

### DIFF
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -935,17 +935,32 @@ await List.create({
 | `upsert` | Finds the row **among this parent's**, updates it, or creates it. A to-one has no `where` to require — the link already names the row. |
 
 **A relation may join on more than one field.** `@relation(fields: [tenantId, orderId], references:
-[tenantId, id])` is the tenant-scoped composite-key style, and every read surface handles it: an
-`include` under either strategy, a relation filter, a `_count`, and an `orderBy` through the
-relation. The correlation becomes one equality per field; the batched strategy filters its children
-with an `OR` of `AND`s rather than a tuple `in`, because that is one shape both dialects already
-compile.
+[tenantId, id])` is the tenant-scoped composite-key style, and every surface handles it — reads and
+nested writes alike. On the read side: an `include` under either strategy, a relation filter, a
+`_count`, and an `orderBy` through the relation. The correlation becomes one equality per field; the
+batched strategy filters its children with an `OR` of `AND`s rather than a tuple `in`, because that
+is one shape both dialects already compile.
 
-Two consequences worth knowing. A composite join uses one placeholder *per field per parent*, so
-SQLite's parameter ceiling arrives proportionally sooner on a wide `include` — `ParameterLimitError`
-still names it rather than letting the driver fail. And a **nested write** through a composite
-relation is refused: it would have to contribute that many foreign-key columns to the insert, which
-is not implemented. Write the child separately with its keys set.
+**A nested write through one contributes every column**, in every operand that writes a key —
+`connect`, `create`, `connectOrCreate`, `createMany`, `set`, `disconnect`, `update`, `upsert`,
+`delete` and the two `*Many` — and on both sides of the relation. Three details follow from the key
+having more than one column, and they are the ones worth knowing:
+
+- **`connect` names the row in Prisma's compound form**, `connect: { tenantId_code: { tenantId: 1,
+  code: "a" } }`, because that is what a multi-column unique key is called. Unlike the single-field
+  case there is no zero-query shortcut — the operand does not carry the referenced values directly —
+  so it costs one lookup, which is what Prisma issues anyway.
+- **A key with any column `null` links nothing.** That is SQL's rule, not a choice: `(tenantId,
+  orderId) = (1, NULL)` is unknown, so the join finds no row. A `disconnect` clears every column
+  together, and an `update` through a half-written key reports the row as missing rather than
+  matching on the half that is set.
+- **A composite one-to-one displaces**, the same way a single-column one does — the discriminator is
+  a `@@unique` covering *exactly* the relation's own fields. `@@unique([tenantId, orderId, kind])`
+  covers them and something else, so it does not make the relation exclusive and does not displace.
+
+One consequence of the join itself: a composite `include` uses one placeholder *per field per
+parent*, so SQLite's parameter ceiling arrives proportionally sooner on a wide one —
+`ParameterLimitError` still names it rather than letting the driver fail.
 
 Which direction a nested write runs in is decided by **who holds the foreign key**. When this model
 holds it, the far row is resolved or created *first* and collapses into one more column. When the
@@ -2211,12 +2226,6 @@ Stated so you can plan around them rather than discover them:
   an identity map is worse than none.
 - **No lazy loading.** A relation you did not `include` is absent, not a proxy that queries when
   touched.
-- **No multi-field relations in a nested write.** `@relation(fields: [a, b], references: [c, d])`
-  now works everywhere a relation is *read* — `include` under either strategy, a relation filter,
-  `_count`, and an `orderBy` through a relation all correlate on every field. What is still refused
-  is reaching one from a nested write, and the error names the fields on both sides and the
-  operation it came from, rather than joining on the first field and writing plausible wrong rows.
-  Pending rather than declined.
 - **No migrations, no schema DSL.** Prisma owns both, and gemi must not shadow the Prisma CLI.
 - **No `distinct`, and this one is deliberate rather than pending.** Prisma applies it **in
   memory** — its query log shows no `DISTINCT` at all, so `take` neither reduces the rows pulled
--- a/docs/orm.md
+++ b/docs/orm.md
@@ -941,10 +941,15 @@ nested writes alike. On the read side: an `include` under either strategy, a rel
 batched strategy filters its children with an `OR` of `AND`s rather than a tuple `in`, because that
 is one shape both dialects already compile.
 
-**A nested write through one contributes every column**, in every operand that writes a key —
-`connect`, `create`, `connectOrCreate`, `createMany`, `set`, `disconnect`, `update`, `upsert`,
-`delete` and the two `*Many` — and on both sides of the relation. Three details follow from the key
-having more than one column, and they are the ones worth knowing:
+**A nested write through one contributes every column**, on both sides of the relation, in every
+operand that side offers. Which operands those are is unchanged by the width of the key: the
+**foreign** side — where the children hold it — takes all of `connect`, `create`, `connectOrCreate`,
+`createMany`, `set`, `disconnect`, `update`, `upsert`, `delete` and the two `*Many`; the **owning**
+side takes `connect`, `create`, `connectOrCreate`, `disconnect` and `update`, and refuses
+`createMany`, `set`, `delete`, `upsert` and the two `*Many` by name, because this row holds one
+foreign key and there is nothing there to write many of. A composite relation is refused in exactly
+the same places a single-field one is, and nowhere else. Four details follow from the key having
+more than one column, and they are the ones worth knowing:
 
 - **`connect` names the row in Prisma's compound form**, `connect: { tenantId_code: { tenantId: 1,
   code: "a" } }`, because that is what a multi-column unique key is called. Unlike the single-field
@@ -957,6 +962,13 @@ having more than one column, and they are the ones worth knowing:
 - **A composite one-to-one displaces**, the same way a single-column one does — the discriminator is
   a `@@unique` covering *exactly* the relation's own fields. `@@unique([tenantId, orderId, kind])`
   covers them and something else, so it does not make the relation exclusive and does not displace.
+- **Two relations that share a column cannot both be written in one call.** Nothing stops a model
+  from carrying `@relation(fields: [tenantId, orderId], …)` beside `@relation(fields: [tenantId,
+  noteId], …)` — Prisma accepts it — and then a single `data` writing through both says `tenantId`
+  twice. Prisma lets the *last* key in your `data` object win, which makes the row depend on a key
+  order gemi deliberately does not preserve: two argument objects differing only in key order have
+  to compile to one cached plan. So this is refused, naming the column and both relations. Write one
+  of them through the relation and set the other's own columns directly.
 
 One consequence of the join itself: a composite `include` uses one placeholder *per field per
 parent*, so SQLite's parameter ceiling arrives proportionally sooner on a wide one —
@@ -2243,7 +2255,7 @@ Stated so you can plan around them rather than discover them:
   column at schema validation. They are otherwise fully supported — see
   [Scalar lists](#scalar-lists-postgres-only).
 
-Both of the last two throw `UnsupportedByDesignError`, which says *"and this is a decision rather
+`distinct` and `cursor` throw `UnsupportedByDesignError`, which says *"and this is a decision rather
 than a gap"* rather than *"yet"* — so a refusal you can plan around reads differently from one that
 might lift next release.
 
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -244,34 +244,42 @@ function planOne(
 
   const child = relatedSchema(schema, relation);
 
-  // **Narrowed to one field, deliberately.** Reading across a composite
-  // relation works (#67); writing through one would have to contribute that
-  // many foreign-key columns to this insert, which is a different piece of
-  // work — so it is refused here by name rather than silently writing the
-  // first field. The narrowing is a function call rather than an index, so
-  // there is no single-field property on `Link` to reach for by accident.
+  // **Every joined field, not the first one** (#271).
   //
-  // Safe to do before the join-table branch below: an implicit many-to-many
-  // links parent and child by their primary keys, one field each side, so the
-  // narrowing never refuses one — and it carries `join` through untouched.
-  const link = singleFieldLink(
-    resolveLink(schema, child, relation, operation),
-    schema,
-    relation,
-    operation,
-  );
+  // This used to be `singleFieldLink(resolveLink(…))`: reading across
+  // `@relation(fields: [tenantId, orderId], references: [tenantId, id])` was
+  // implemented by #67 on all six read surfaces, and the write surface refused
+  // it by name rather than joining on the first field and writing plausible
+  // wrong rows. The refusal was honest and it was the only thing left on that
+  // list, so the narrowing is gone and what follows contributes *n* foreign-key
+  // columns wherever it used to contribute one.
+  //
+  // Nothing here branches on how many there are. The two sides read the link
+  // through {@link childLink}, {@link keySelect} and {@link nullKey}, which are
+  // the same three shapes for one field as for four — so a relation joining on
+  // one field compiles to exactly the arguments it did before, and there is no
+  // single-field path left to diverge from the composite one.
+  const link = resolveLink(schema, child, relation, operation);
 
   // An implicit many-to-many is *neither* side: the keys live in a third table
   // with no model, so both directions are the same work and the operand set is
   // wider — `disconnect` and `set` are a delete against the join table, which
   // needs no schema to compile.
   if (link.join) {
+    // **The one caller that still cannot take more than one field**, and it is
+    // an invariant rather than a gap: Prisma's implicit many-to-many links
+    // parent and child by their *primary keys*, one column each side, into a
+    // two-column join table there is no schema to widen. `singleFieldLink`
+    // states that rather than letting `planJoinTable` index into `[0]` and be
+    // silently wrong if it ever stopped holding.
+    const pair = singleFieldLink(link, schema, relation, operation);
+
     for (const key of keys) {
       planJoinTable(
         schema,
         relation,
         child,
-        link,
+        pair,
         key,
         (node as Record<string, unknown>)[key],
         operation,
@@ -345,23 +353,55 @@ function planOne(
 
 /**
  * This model holds the foreign key, so the far row must exist — or be created —
- * before the insert can bind, and the whole thing collapses to one extra column.
+ * before the insert can bind, and the whole thing collapses to *n* extra
+ * columns — one, in the ordinary schema; as many as the relation joins on in a
+ * tenant-scoped one (#271).
+ *
+ * **The columns are contributed to the statement the caller already asked for,
+ * which is what makes a composite key here atomic without arranging anything.**
+ * `contributions` is a list keyed by field, and `insertColumns` /
+ * `updateAssignments` fold every entry into the single `INSERT` or `UPDATE`
+ * this operation compiles to — so "some keys written, some not" is not a state
+ * this side can reach: the `before` step either resolves the whole tuple or
+ * throws, and the statement either runs or does not. The partial-write hazard
+ * is real on the *foreign* side, where each linked row is its own statement,
+ * and it is answered there by the transaction every plan carrying steps opens.
  */
 function planOwningSide(
   schema: ModelSchema,
   relation: RelationSchema,
   child: ModelSchema,
-  link: { parentField: string; childField: string },
+  link: Link,
   key: string,
   operand: unknown,
   operation: string,
   at: (args: any) => any,
   out: NestedWritePlanning,
 ): void {
-  // `link.parentField` is the FK on this model; `link.childField` is what it
-  // references on the other one.
-  const fkField = link.parentField;
-  const referenced = link.childField;
+  // `parentFields` are the FK columns on this model; `childFields` are what
+  // they reference on the other one, paired positionally.
+  const fkFields = link.parentFields;
+  const referencedFields = link.childFields;
+
+  /** Assign the resolved key onto the statement — one contribution per column. */
+  const contributeResolved = () => {
+    for (const field of fkFields) {
+      out.contributions.push({
+        field,
+        value: (_args, context) => context.resolved[field],
+      });
+    }
+  };
+
+  /** Record a whole key on the plan's per-call scratch space. */
+  const resolveKey = (
+    context: { resolved: Record<string, unknown> },
+    values: readonly unknown[],
+  ) => {
+    for (let index = 0; index < fkFields.length; index++) {
+      context.resolved[fkFields[index]] = values[index] ?? null;
+    }
+  };
 
   /**
    * **Whether linking through this relation has to displace a sibling** — the
@@ -428,14 +468,38 @@ function planOwningSide(
    * the foreign side: it would refuse the whole operand, including the
    * empty-far-row case that has always worked.
    *
-   * A single-column index only, which is what `singleFieldLink` has already
-   * narrowed this whole function to. A composite `@@unique` covering the key
-   * *and something else* does not make the relation one-to-one — Prisma wants
-   * the relation's own `fields` unique — and reading it as one would clear rows
-   * the schema never said were exclusive. A foreign key that is also the `@id`
-   * would not be found in `uniques`, which records `@unique` and `@@unique`
-   * rather than the primary key; it is out of reach anyway, since `@id` implies
-   * `NOT NULL` and the nullable guard has already excluded it.
+   * **An index over exactly the relation's own fields**, which is what the
+   * single-field test used to say and reads differently now that there may be
+   * several (#271). Prisma's rule is on the relation: *"A one-to-one relation
+   * must use unique fields on the defining side"*, meaning `fields` — so
+   * `@@unique([tenantId, ownerId])` beside `@relation(fields: [tenantId,
+   * ownerId], …)` is the composite one-to-one, and it is the same fact the
+   * single-column form was stating.
+   *
+   * **Both halves of that rule are Prisma's, measured on 6.19.2 rather than
+   * inferred**, because between them they are the whole discriminator:
+   *
+   *     @@unique([tenantId, ledgerCode])         accepted — the one-to-one
+   *     @@unique([tenantId, ledgerCode, seal])   P1012, "Either add an
+   *                                              `@@unique([tenantId,
+   *                                              ledgerCode])` attribute"
+   *     @@unique([ledgerCode, tenantId])         the identical P1012
+   *
+   * So a wider group does *not* make the relation one-to-one — reading it as
+   * one would clear rows the schema never said were exclusive — and the order
+   * has to match the relation's `fields` as well.
+   *
+   * Written as set equality all the same, and the second row is why that is not
+   * laxity: the length check is what carries the weight, and a positional loop
+   * would read as though order were a thing this had decided, when on any
+   * schema Prisma accepted the two lists are already identical. It also keeps a
+   * hand-built `ModelSchema` — a fixture, or a generator that reorders — out of
+   * the branch where the index is not recognised at all.
+   *
+   * A foreign key that is also the `@id` would not be found in `uniques`, which
+   * records `@unique` and `@@unique` rather than the primary key; it is out of
+   * reach anyway, since `@id` implies `NOT NULL` and the nullable guard has
+   * already excluded it.
    */
   const backRelations = Object.values(child.relations).filter(
     (candidate) =>
@@ -447,10 +511,13 @@ function planOwningSide(
       !(child.name === schema.name && candidate.name === relation.name),
   );
 
+  const keyed = new Set(fkFields);
+
   const displaces =
-    schema.fields[fkField]?.nullable === true &&
+    fkFields.every((field) => schema.fields[field]?.nullable === true) &&
     schema.uniques.some(
-      (unique) => unique.length === 1 && unique[0] === fkField,
+      (unique) =>
+        unique.length === keyed.size && unique.every((name) => keyed.has(name)),
     ) &&
     backRelations.length === 1 &&
     backRelations[0].kind !== "many";
@@ -490,19 +557,16 @@ function planOwningSide(
         const created = (await executor.exec(
           relation.model,
           "create",
-          { data: at(args), select: { [referenced]: true } },
+          { data: at(args), select: keySelect(referencedFields) },
           // NOT pre-scoped. Nothing walks `data.<relation>.create`, so the
           // child's own `onCreate` is the only thing that can scope this row.
           false,
         )) as Record<string, unknown> | null;
 
-        context.resolved[fkField] = created?.[referenced] ?? null;
+        resolveKey(context, valuesOf(created, referencedFields));
       },
     });
-    out.contributions.push({
-      field: fkField,
-      value: (_args, context) => context.resolved[fkField],
-    });
+    contributeResolved();
     return;
   }
 
@@ -566,7 +630,7 @@ function planOwningSide(
     // out of the input type entirely when the relation is required, so
     // `disconnect: false` there is *"Unknown argument `disconnect`"* rather
     // than an accepted no-op. Measured on `SocialAccount.user`.
-    assertDisconnectable(schema, relation, fkField, {
+    assertDisconnectable(schema, relation, fkFields, {
       model: schema.name,
       operation,
       argument: `data.${relation.name}.disconnect`,
@@ -591,13 +655,15 @@ function planOwningSide(
     if (filter === null || filter === undefined) return;
 
     /**
-     * `true` — one bound column set to `null`, exactly as `connect`'s direct
-     * form is one bound column set to a value. No step, nothing on the child
-     * read or written, and so no scoping question: the row being changed is
-     * the one the statement already names.
+     * `true` — the link's columns bound to `null`, exactly as `connect`'s
+     * direct form binds them to a value. No step, nothing on the child read or
+     * written, and so no scoping question: the row being changed is the one the
+     * statement already names.
      */
     if (operand === true) {
-      out.contributions.push({ field: fkField, value: () => null });
+      for (const field of fkFields) {
+        out.contributions.push({ field, value: () => null });
+      }
       return;
     }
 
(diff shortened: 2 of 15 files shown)


packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -345,23 +353,55 @@ function planOne(
...
-  // references on the other one.
-  const fkField = link.parentField;
-  const referenced = link.childField;
+  // `parentFields` are the FK columns on this model; `childFields` are what
+  // they reference on the other one, paired positionally.
+  const fkFields = link.parentFields;
+  const referencedFields = link.childFields;
+
+  /** Assign the resolved key onto the statement — one contribution per column. */
+  const contributeResolved = () => {
```
nstfkc, line 437: **should fix** — two composite relations sharing a foreign-key column silently write a different row than Prisma does, and this is the first change that makes that shape reachable.

A composite relation joins on *n* columns, and nothing stops two relations on the same model from sharing one. That is not exotic — it is the tenant-scoped shape this whole PR is about:

```prisma
model Entry {
  id         Int    @id @default(autoincrement())
  tenantId   Int
  ledgerCode String
  noteCode   String
  ledger Ledger @relation("LedgerToEntry", fields: [tenantId, ledgerCode], references: [tenantId, code])
  note   Ledger @relation("NoteToEntry",   fields: [tenantId, noteCode],   references: [tenantId, code])
}
```

`bunx prisma validate` on 6.19.2: *"The schema is valid 🚀"*. Before this PR both relations were refused by name; now both compile, and `contributeResolved` pushes a contribution for `tenantId` **twice**. `insertColumns` (`write.ts:1060`) folds contributions through `new Map(contributions.map(c => [c.field, c]))`, so the later one wins — and `planOne` sorts the operand keys alphabetically, so the winner is whichever relation sorts last, regardless of how the caller wrote the object. `context.resolved[tenantId]` collides the same way, one layer up.

Measured on both sides, same schema, `prisma db push` to a scratch SQLite file with query events on:

| caller's `data` | Prisma 6.19.2 | gemi (this branch) |
| --- | --- | --- |
| `{ ledger: connect(1,"a"), note: connect(2,"b") }` | `tenantId=2, ledgerCode="a", noteCode="b"` | `tenantId=2, ledgerCode="a", noteCode="b"` ✅ |
| `{ note: connect(2,"b"), ledger: connect(1,"a") }` | `tenantId=**1**, ledgerCode="a", noteCode="b"` | `tenantId=**2**, …` ❌ |

Prisma resolves the operands in reverse key order and lets the *first resolved* win, so the **last key in the caller's `data` object** decides the shared column. gemi always lets the alphabetically-last relation decide. The rows differ: gemi's `ledger` link is `(2,"a")` where Prisma's is `(1,"a")` — a different `Ledger`, or a foreign-key violation where that row does not exist. No error either side.

`planOne`'s own comment above the `keys.sort()` says the alphabetical order is *"a coincidence that is currently correct. An operand added here whose order **is** observable cannot rely on the alphabet and has to sequence itself explicitly."* Composite keys are the first thing that makes relation order observable, and it is not an operand being added — it is the same operand on two relations.

Matching Prisma exactly would mean giving up the sorted, cache-canonical key order, which is not worth it. The cheap and honest fix is a refusal: `insertColumns` already raises *"'x' is set both directly and through a nested relation write"* for the caller-vs-relation collision, and the relation-vs-relation collision has no equivalent. Detecting a duplicate `field` across `out.contributions` and refusing by name would keep the property this PR's own summary claims for the refusal it replaced — *cannot write plausible wrong rows*. Whatever the resolution, it wants a line here saying which it is, since this is where the collision is created.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -447,10 +511,13 @@ function planOwningSide(
       !(child.name === schema.name && candidate.name === relation.name),
   );
 
+  const keyed = new Set(fkFields);
+
   const displaces =
-    schema.fields[fkField]?.nullable === true &&
+    fkFields.every((field) => schema.fields[field]?.nullable === true) &&
```
nstfkc, line 567: **nit** — the three "every column" *nullability* generalisations have no test that distinguishes them from the first-field spelling.

Crippled all three to their first field at once —

- this line → `schema.fields[fkFields[0]]?.nullable === true`
- `planForeignSide`'s `childFields.every(…nullable…)` → `child.fields[childFields[0]]`
- `assertDisconnectable`'s `fieldNames.find(…)` → `[fieldNames[0]].find(…)`

— and the package suite is **120 files, 3128 passed, 1 skipped**, i.e. exactly green. Every other generalisation in this diff fails loudly under the same treatment (see the table in the summary), so these three stand out.

That is defensible and the docblocks say why: Prisma makes a composite relation optional or required as a whole, so on a validated schema the columns agree and the first one answers. But `assertDisconnectable`'s docblock makes a *concrete behavioural claim* about the case Prisma cannot produce — *"a hand-built `ModelSchema` with `tenantId` required beside a nullable `orderId` is refused here rather than nulling half a key"* — and nothing pins it. `fixtures.ts` is exactly where a hand-built `ModelSchema` lives; one fixture with a mixed-nullability composite key would pin all three predicates at once and turn three documented intentions into three measured ones.

(Same category, and the PR is already honest about it: `clearLinks`' `isLinked` early return is likewise unreached — crippling `isLinked` fails one test, and it is the owning-side `update` one, not a `clearLinks` one.)

docs/orm.md
```diff
@@ -935,17 +935,32 @@ await List.create({
...
-still names it rather than letting the driver fail. And a **nested write** through a composite
-relation is refused: it would have to contribute that many foreign-key columns to the insert, which
-is not implemented. Write the child separately with its keys set.
+[tenantId, id])` is the tenant-scoped composite-key style, and every surface handles it — reads and
+nested writes alike. On the read side: an `include` under either strategy, a relation filter, a
+`_count`, and an `orderBy` through the relation. The correlation becomes one equality per field; the
+batched strategy filters its children with an `OR` of `AND`s rather than a tuple `in`, because that
+is one shape both dialects already compile.
+
+**A nested write through one contributes every column**, in every operand that writes a key —
```
nstfkc: **nit** — this operand list is wider than the code on the owning side.

*"in every operand that writes a key — `connect`, `create`, `connectOrCreate`, `createMany`, `set`, `disconnect`, `update`, `upsert`, `delete` and the two `*Many` — **and on both sides of the relation**"*.

On the owning side `createMany`, `set`, `updateMany` and `deleteMany` are refused by name (`planOwningSide`'s `createMany` guard, and `assertToOneWriteOperand` / the to-one refusals), and `upsert` is refused too — the PR body says so itself under *Not done here*. None of that changed with this PR, but the old paragraph did not make a claim about it and the new one does, so a reader planning around this page will reach for `ledger: { upsert: … }` and get a refusal the docs just told them exists.

The fix is a clause, not a rewrite: name the operands the *foreign* side adds separately, or drop `upsert`/`set`/`createMany`/the `*Many` pair from the "both sides" sentence.

Unrelated and pre-existing, but it is three screens below this in the same file and the diff deletes a bullet from that very list: *"Both of the last two throw `UnsupportedByDesignError`"* now sits under `Decimal` and *scalar lists are Postgres-only*, neither of which throws it — it means `distinct` and `cursor`. Deleting the multi-field bullet did not move which two are last, so this is not caused here; flagging it only because #374 in this batch was precisely a comment that stopped being true and nobody was standing next to it.

nstfkc (commented):
**Verdict: correct as far as it goes, one should-fix parity gap, two nits. Nothing blocking.**

The mechanism is right and the generalisation is genuinely uniform: `planOne` stops narrowing, both planners take the whole `Link`, and the ~40 hand-spelled `{ [childField]: parent[parentField] }` literals collapse into `childLink` / `keySelect` / `nullKey` / `isLinked`. Reading for a missed site, the only remaining singular access is `planJoinTable`'s, which is `singleFieldLink`'s last caller and is an invariant of Prisma's implicit m-n rather than a gap — the docblock rewrite in `plan-relations.ts` says exactly that and is accurate. Orientation is safe on both sides: `resolveLink` always returns `parentFields` on the model being planned, so `childLink` reading as "belonging to this parent" on the foreign side and "the row this one points at" on the owning side is the same arithmetic, not a coincidence.

### What I ran

Fresh worktree, `bun install` in-tree, `readlink -f templates/saas-starter/node_modules/gemi` → the worktree's `packages/gemi` (so template results are this branch's code), `build:bin`, the `gemi-orm-generator` symlinks, both `prisma generate` runs.

- `packages/gemi`: `bun run typecheck` clean; `bun --bun vitest run` → **120 files, 3128 passed, 1 skipped**; `bunx oxlint orm --deny-warnings` → 0/0.
- `templates/saas-starter` (SQLite): `TZ=UTC bun --bun vitest run app/models` → **822 passed, 133 skipped**; `TZ=UTC bun run test:types` → 116, no type errors.
- **The committed generated artifacts reproduce byte-for-byte**: after both `prisma generate` runs, `git status` was empty. `differential.prisma` and `app/models/generated/*` are not stale.
- Postgres: I did **not** run it locally; I am relying on the green `postgres` CI job. The nested-write path is executor-driven and dialect-independent, so I would not expect divergence, but that is reasoning rather than measurement.

### Mutation checks (the tests bite)

Each helper crippled to its first field in turn, then restored:

| mutation | result |
| --- | --- |
| `childLink` → first field only | **10** fail in `composite-relations.test.ts`, **14** in `writes.differential.test.ts` |
| `keySelect` + `nullKey` → first field only | **7** fail |
| `out.keyFields.push(...parentFields)` → `[0]` only | **2** fail (incl. `composite-surfaces.test.ts`'s `plan.hidden` pin) |
| `contributeResolved` → first column only | **2** fail |
| `displaceSibling`'s tuple compare `.every` → `.some` | **2** fail |
| `isLinked` → first value only | **1** fails |
| `displaces` unique test → back to `unique.length === 1` | **2** fail |

Those all reproduce the author's table. The central test is not passing on unfixed code.

### Scope

On-topic throughout. The one thing that reads as a drive-by — `differential.ts`'s SQLite `clearAll` gaining the four Ledger tables — is load-bearing for the new write comparisons (a write case resets between the two clients' runs), so it belongs here.

### Merge ordering

Restating `orm.md`'s note rather than finding it: this PR adds `LedgerNote` / `LedgerSeal` and migration `20260809000000`, so any other ORM PR rebased onto `main` after it must re-run `bun prisma/differential-schema.ts` and **both** `prisma generate`s or `app-dependencies.test.ts` fails. That makes this the natural last merge of the batch.

### Small process note

The closing keyword is only in the title and the commit subject, not in the PR body — a squash merge will still close #271 (the title becomes the subject), but a merge commit would not.

---

Three comments inline. The first is the only one I would act on before this ships as *"a nested write contributes every joined field"*.

nstfkc:
**[cross-PR review]** — batch #377–#386, looking only for what shows up *across* PRs. Two findings,
plus the answer to the ordering question `orm.md` raised about this PR.

---

## 1. #386 × #378 and #386 × #379 — a clean merge that does not compile

The `#92`/`#102` shape the `ci.yml` docblock exists for, and the sharpest instance in this batch:
**`git merge` reports no conflict, and `tsc` then fails.**

This PR deletes the single-field bindings at the top of `planForeignSide` —

```
```

— while #378 and #379 each add *new* references to the old names, in hunks this PR does not touch.
So they auto-merge, and then:

```

# all ten merged
orm/compile/nested-writes.ts(2147,16), (2245,13), (2247,20)
```

Loud rather than silent, since `bun run typecheck` catches it — but it will stop the batch.

**What the rebase costs.** #378's is mechanical: its `[childField]` ORM-authored column list becomes
`childFields`, matching what this PR already passes at every other `exec` in the function. #379's
needs real generalisation, because `sameKey` stays scalar-only here — its `alreadyLinked` helper has
to take `readonly string[]` / `readonly unknown[]`, select `keySelect(fields)`, and compare
`fields.every((f, i) => sameKey(named[f], values[i]))`. A partially-matching composite key must not
count as already-linked, which lines up with this PR's own `isLinked` rule.

There is a third site: once #379 lands, `connectOrCreate`'s hit branch needs the same short-circuit
or #378's pin fails (detail on #378/#379). On a tree carrying this PR it has to be spelled
compositely too — `childFields.every((f, i) => sameKey(found[f], valuesOf(parent, parentFields)[i]))`.

---

## 2. Should #386 go last? **Yes — but not for the reason `orm.md` gives.**

`orm.md` says: *"Any orm PR rebased onto `main` after #386 merges must re-run
`bun prisma/differential-schema.ts` and both `prisma generate`s, or `app-dependencies.test.ts`
fails."*

**I measured this and it does not hold for this batch.** Two reasons:

- **No other PR touches the template schema or the generated artifacts.** I diffed all ten against
  `main` over `templates/saas-starter/prisma/` and `templates/saas-starter/app/models/generated/`:
  #386 is the only one with any hit. Nothing another PR does can stale them.
- **This PR's committed artifacts already reproduce byte-for-byte.** With all ten merged, I ran both
  `prisma generate`s and `bun prisma/differential-schema.ts`; `git status` showed **no change** to
  `schema.ts`, `models.ts`, `index.ts` or `differential.prisma`. And `app-dependencies.test.ts`
  asserts `differential.prisma == derive(schema.prisma)` — both files are committed here and changed
  together, so a rebase of another PR cannot break that pair.

(Small thing: `app-dependencies.test.ts` lives at `packages/gemi/orm/app-dependencies.test.ts`, not
in the template — the note reads as though it were a template test.)

**The real reason to go last is this PR's `childField` → `childFields` rename**, finding 1 above.
Merging #386 first forces *two* branches (#378 and #379) to absorb it separately; merging it last
means one rebase, by the author who introduced the rename and owns the composite helpers. I
confirmed the conflict *set* is identical in both directions — I ran the full merge sequence with
#386 first and with #386 last, and got the same two textual conflicts either way — so the ordering
buys reconciliation cost, not correctness.

Recommended: **#377, #380, #381, #382, #383, #384, #385, #379, #378, #386.**

---

Full ten-PR integration branch built locally, both textual conflicts resolved and the repairs above
applied:

| check | result |
| --- | --- |
| live Postgres via `postgres.sh` | **1247 passed** (only 2 failures, the harness's designed one-dialect-at-a-time refusal) |

Also checked and found **clean**: this PR against #381, which derives its nested-write statement
types from `SUPPORTED_STATEMENTS` / `EXISTING_ROW_ONLY_STATEMENTS` / `COLLECTION_ONLY_STATEMENTS`.
This PR does not touch any of those tuples, so #381's derived grammars are unchanged by it —
verified by diff and by `test:types` passing on the merged tree. `fixtures.ts` (shared with #385)
and `docs/orm.md` + `docs/llms-full.txt` (shared with #378, #380, #382) all auto-merge, and
`docs-scope.test.ts`'s verbatim-carry assertion passes on the result.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: **Addressed** — refused, as you suggested, in `bddc268`.

You are right that this PR creates the shape, and right that matching Prisma is not worth the sorted key order. `planNestedWrites` now records which relation contributed each column and raises on the second:

```
'tenantId' is written by both the 'ledger' and 'note' relations, which join on it in
common. Prisma lets the last key in 'data' win; gemi plans relations in a fixed order
and will not guess which row you meant. Write one of them through the relation and set
the other's own columns directly.
```

Three details worth flagging, since none of them were in your comment:

**The guard is over the *pair*, not over a duplicate field as such.** Two operands on *one* relation contribute the same columns twice by construction — a `connect` beside a `create` on a to-one has always done that, and `insertColumns`' fold is what handles it. Refusing on a raw duplicate would have broken that. The comparison is `owner !== key`, and there is a test pinning the same-relation case as *not* a collision, so a later tightening has to notice it is tightening.

**The `context.resolved` collision one layer up is covered by the same refusal**, not separately: it can only happen when both relations are planned, and both being planned is now the refusal. A relation on its own still resolves and contributes its whole key — also pinned.

**The advice in the message is the one that actually works.** "Set the shared column directly" does not, because `insertColumns` refuses a column written *both* ways. Connecting through one relation and setting the other's remaining columns directly does, and that is what it says.

`ledgerCrossEntry` in `fixtures.ts` is the schema; four cases in `composite-relations.test.ts` cover the refusal, the refusal being order-independent (which is the point — the two orders are different rows under Prisma and one plan here), a single relation still writing its whole key, and the same-relation non-collision. Mutation-checked: disabling the guard fails the first two.

Documented too — `docs/orm.md` gains it as a fourth bullet under the composite nested-write paragraph, and `llms-full.txt` re-derived.

One thing I did **not** do: no differential case. There is no SQL for a link and the refusal is compile-time, so a differential test would assert a throw on one side and Prisma's measured behaviour on the other — which is the measurement in the comment, not something the harness can compare. Adding a shared-column model to the template would also mean another migration and regenerated artifacts on the PR that already merges last.

`orm.md`'s Notes listed this as an unfiled defect to raise before merging; it is fixed here instead, so there is nothing left to file.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: **Addressed** in `bddc268`, and your diagnosis was exactly right — including the pointer to `fixtures.ts` as the place a hand-built `ModelSchema` lives.

`ledgerSealMixed`: `tenantId` nullable, `ledgerCode` required, `@@unique([tenantId, ledgerCode])`, and `ledgerWithMixed` carries a non-list `sealMixed` back-relation so the *other* halves of both `displaces` predicates are satisfied and nullability is the only thing left deciding — otherwise the cases would pass for the wrong reason.

**The nullable column is deliberately first**, which is the part that makes it bite. `[fields[0]]` is what all three collapse to, and over `(nullable, required)` that answers *"detachable"* where the tuple answers *"not"*. Ordered the other way the fixture would pin nothing.

Three cases, and I ran your three mutations against them:

| mutation | result |
| --- | --- |
| owning `displaces` → `schema.fields[fkFields[0]]?.nullable === true` | **1** fails — *"the owning side does not clear an incumbent it cannot fully detach"* |
| foreign `displaces` → `child.fields[childFields[0]]?.nullable === true` | **1** fails — the foreign-side mirror |
| `assertDisconnectable` → `[fieldNames[0]].find(…)` | **1** fails — the refusal names `LedgerSealMixed.ledgerCode`, the *second* column |

`assertDisconnectable`'s docblock now points at the fixture and the test rather than only stating the claim, and says why the nullable column is first.

On the parenthetical — **`clearLinks`' `isLinked` early return stays unreached and stays undocumented as reached.** Nothing calls it with a partial key today; the PR body says so and the docblock says so. Writing a fixture whose only purpose is to reach a branch no caller can reach would pin the branch and prove nothing about the file, and the branch exists precisely so that a future caller cannot turn `where: { tenantId: 1, orderId: null }` into an `updateMany` over every unlinked row in the tenant. That is the one of the four I am leaving as documented-intent.

docs/orm.md
nstfkc: **Addressed** in `bddc268` — both halves, including the pre-existing one.

You are right and the distinction you draw is the right one: the old paragraph made no claim about which operands exist, the new one did, and a reader planning around it would reach for `ledger: { upsert: … }` and get a refusal the page had just implied was reachable. Checked against the code rather than from memory — `planOwningSide` refuses `createMany`, `delete`, `set`, `updateMany`, `deleteMany` and `upsert`, and supports `connect`, `create`, `connectOrCreate`, `disconnect`, `update`.

Rather than dropping operands from a "both sides" sentence I named the two sides separately, because the useful fact for a reader is that *the width of the key changes nothing about which operands exist*:

> …on both sides of the relation, in every operand that side offers. Which operands those are is unchanged by the width of the key: the **foreign** side — where the children hold it — takes all of `connect`, `create`, `connectOrCreate`, `createMany`, `set`, `disconnect`, `update`, `upsert`, `delete` and the two `*Many`; the **owning** side takes `connect`, `create`, `connectOrCreate`, `disconnect` and `update`, and refuses `createMany`, `set`, `delete`, `upsert` and the two `*Many` by name, because this row holds one foreign key and there is nothing there to write many of. A composite relation is refused in exactly the same places a single-field one is, and nowhere else.

**And I fixed the unrelated one.** *"Both of the last two throw `UnsupportedByDesignError`"* now reads *"`distinct` and `cursor` throw…"*, which is what `PERMANENTLY_REFUSED` in `compile/read.ts` actually holds — verified there, not inferred from the list's order. Two words, in a file this PR already touches, and your #374 argument is the right one: a comment that stopped being true stays untrue until someone standing next to it fixes it. If a reviewer would rather it were separated out, say so and I will pull it back into its own change.

`docs/llms-full.txt` re-derived for both edits; `docs-scope.test.ts`'s byte-for-byte carry passes.

nstfkc:
## Review round 1 — disposition of every point

Pushed as `bddc268`. Five files: `orm/compile/nested-writes.ts`, `orm/fixtures.ts`, `orm/compile/composite-relations.test.ts`, `docs/orm.md`, `docs/llms-full.txt`. No generated artifact moved — both `prisma generate` runs still leave `git status` clean.

### Inline (replies on each)

| # | verdict | disposition |
| --- | --- | --- |
| [1](https://github.com/nstfkc/gemi/pull/386#discussion_r3743307223) — two relations sharing a foreign-key column | should fix | **addressed** — refused by name, with the fixture, four cases and a docs bullet |
| [2](https://github.com/nstfkc/gemi/pull/386#discussion_r3743307226) — the three nullability generalisations are unpinned | nit | **addressed** — `ledgerSealMixed`, three cases, all three mutations now red. The `clearLinks` parenthetical is **declined**, with reasoning |
| [3](https://github.com/nstfkc/gemi/pull/386#discussion_r3743307228) — the docs operand list is wider than the code | nit | **addressed**, both halves, including the pre-existing `UnsupportedByDesignError` sentence |

### From the main review

**"The closing keyword is only in the title."** Fixed — the body now opens `Closes #271`, and `gh pr view 386 --json closingIssuesReferences` returns the issue rather than `[]`. You are right that a squash would have closed it anyway; #267, #263, #269 and #262 are what a merge commit does instead, and `orm.md`'s Notes flag this PR by name.

**Postgres.** Still not re-run for this round. The round-1 evidence stands — 803 passed against Postgres 16 in a container following the CI job's steps, all 30 composite cases named — and nothing in `bddc268` is dialect-sensitive: the collision guard runs in `planNestedWrites` before a dialect is consulted, and the three new nullability cases are compile-level with a recording executor. Saying so rather than claiming a run I did not do.

### From the [cross-PR review]

**Finding 1 — #386 × #378 and #386 × #379 do not compile after a clean merge.** Confirmed and **deferred to merge time, which is mine.** #386 is not on `main`, so #378 and #379 cannot reference `childFields`/`parentFields` yet, and I will not push into their branches. The recommended order puts this PR last precisely so the rename is absorbed once, by me. What I owe at that point, from your finding: #378's `[childField]` ORM-authored list becomes `childFields`; #379's `alreadyLinked` takes `readonly string[]` / `readonly unknown[]`, selects `keySelect(fields)` and compares `fields.every((f, i) => sameKey(named[f], values[i]))` — a partially-matching composite key must not count as already-linked, which is `isLinked`'s rule and now also `displaceSibling`'s tuple compare. Recorded in `.orm-sessions/issue-271.md` so it survives me.

**The third site — `connectOrCreate`'s hit-branch short-circuit — is #378's, not mine.** It is a change to code #378 introduces, on a branch this PR cannot see; doing it here would be reaching into another PR. Your composite spelling of it, `childFields.every((f, i) => sameKey(found[f], valuesOf(parent, parentFields)[i]))`, is in the session file for the merge, since on a tree carrying this PR that is the form it has to take.

**Finding 2 — go last, but for the rename rather than for the artifacts.** **Accepted, and it corrects my own PR body**, which repeated `orm.md`'s stale reason. Your measurement is the stronger claim: no other PR in the batch touches `templates/saas-starter/prisma/` or the generated artifacts, so nothing another PR does can stale them, and the artifacts here reproduce byte-for-byte anyway — re-confirmed on this round. `app-dependencies.test.ts` living in `packages/gemi/orm/` rather than in the template is also right.

**Clean against #381.** Noted; nothing in `bddc268` touches `SUPPORTED_STATEMENTS` / `EXISTING_ROW_ONLY_STATEMENTS` / `COLLECTION_ONLY_STATEMENTS` either, so that stays true.

### Not changed, and why

- **`clearLinks`' `isLinked` early return stays unreached.** No caller can pass a partial key today. The branch exists so a future one cannot turn `where: { tenantId: 1, orderId: null }` into an `updateMany` across every unlinked row in the tenant; a fixture built only to reach it would pin the branch and prove nothing.
- **`displaces` still compares the unique group as a set**, not positionally. Prisma refuses a reordered `@@unique` with P1012, so on any schema it validated the two lists are identical and the comparison cannot distinguish them; the length check is what carries the weight. Measured, in the body.
- **The composite `connect` still takes a lookup.** Prisma issues the same `select` for a compound `connect`, so parity is unchanged; the zero-query direct form would mean reproducing the generator's `_`-joined naming rule inside the planner.
- **Owning-side `upsert` is still refused**, for its existing reason, unrelated to key width.
- **No differential case for the shared-column refusal.** It is compile-time and has no SQL, so the comparison would be a throw on one side and a measured Prisma row on the other — which is the measurement in the review, not something `expectSameWrite` can run. Adding the model would also mean another migration and regenerated artifacts on the PR that merges last.

### Verification for this round

`packages/gemi`: `bun run typecheck` clean, `bun --bun vitest run` → **120 files, 3135 passed, 1 skipped** (was 3128), `bunx oxlint orm --deny-warnings` → 0/0. Template SQLite: `TZ=UTC bun --bun vitest run app/models` → **822 passed, 133 skipped**; `TZ=UTC bun run test:types` → 116, no type errors. Both `prisma generate` runs: no diff. `readlink -f templates/saas-starter/node_modules/gemi` → this worktree's `packages/gemi`.

Mutation checks on everything added, reverted after each:

| mutation | result |
| --- | --- |
| the shared-column guard disabled | **2** fail |
| owning `displaces` → `fields[fkFields[0]]` | **1** fails |
| foreign `displaces` → `fields[childFields[0]]` | **1** fails |
| `assertDisconnectable` → `[fieldNames[0]].find(…)` | **1** fails |

`nested-writes.ts` diffed against its pre-mutation copy afterwards: identical.

nstfkc:
## [cross-PR re-check]

**#377 → #380 → #381 → #382 → #383 → #384 → #385 → #379 → #378 → #386**, at head `bddc268b` for
this PR.

### Conflict 2 (#386 × #378/#379) — **still open, still merge-time-only, and now larger**

Unchanged in kind: `git merge` is clean and the tree does not compile. The count went from 3 to
**5** `TS2552`, because #378's round added two new old-name references in the hunk it rewrote:

```
nested-writes.ts(2182,44): Cannot find name 'childField'.   ← #378, new this round
nested-writes.ts(2182,64): Cannot find name 'parentField'.  ← #378, new this round
nested-writes.ts(2222,16): Cannot find name 'childField'.   ← #378
nested-writes.ts(2335,13): Cannot find name 'childField'.   ← #379
nested-writes.ts(2337,20): Cannot find name 'parentField'.  ← #379
```

Your merge-order argument holds and is if anything stronger now — going last means absorbing this
once. I resolved it to confirm it is genuinely mechanical, and it is, with one exception worth
knowing before you start.

**Four sites are rename-only:**

```ts
// 2182 — #378's connectOrCreate hit-branch short-circuit
if (displaces && childFields.every((f, i) => sameKey(found[f], parent[parentFields[i]]))) continue;

// 2222 — ormAuthored is already `readonly string[]`
childFields,   // was [childField]

// 2335 — #379's connect short-circuit call site
await alreadyLinked(relation.model, childFields, item, valuesOf(parent, parentFields), executor)
```

**The fifth is not a rename.** `alreadyLinked` (`nested-writes.ts:3064` post-merge) takes a scalar
`field`/`value` pair and ends in a bare `sameKey`, which stays scalar-only under this PR — #379
called this out as needing real generalisation, and it was right:

```ts
async function alreadyLinked(
  model: string, fields: readonly string[], where: unknown,
  values: readonly unknown[], executor: RelationExecutor,
): Promise<boolean> {
  const named = await executor.exec(model, "findUnique", { where, select: keySelect(fields) }, false);
  if (named === null || named === undefined) return false;
  return fields.every((f, i) => sameKey(named[f], values[i]));
}
```

`every` rather than "any component matches" is the load-bearing choice, and it is the trap #379
recorded: a **partial** tuple match must not count as already-linked, or the connect becomes the
silent no-op its third test exists to catch, across a composite key. Worth a comment at the line.

With those five applied, `tsc --noEmit` is clean and the whole batch passes — see below. So the
deferral was correct and the work is bounded.

### Your regeneration correction re-verified on all ten

The Task 1 note that this PR would break every other ORM PR on rebase stayed disproved. On the
merged ten-PR tree, `prisma generate` for `schema.prisma` **and** for `differential.prisma` both
leave `git status` completely clean — your committed artifacts reproduce byte-for-byte alongside
everything else in the batch, not just alongside `main`.

### One thing to watch on your rebase, not yours to fix

`writes.differential.test.ts` now carries two unrelated cases both labelled `M15e` (#379's at line
831, #378's at 858) — git took both silently and both pass, so nothing fails. #378 has been asked to
rename its own to `M15f`. You merge last into that file, so if the rename has not happened by then
you will be the one holding it.

--deny-warnings` 0/0.

Closing keyword verified: `closingIssuesReferences` resolves to #271 — the title-only defect is
fixed.

PR #390, opened 2026-08-10
Application commands and `gemi run`

nstfkc:
A one-off piece of application code — a seeder, a backfill, a tenant repair, a report — had nowhere to live. The two places that already boot the app outside a request (`bin/loadApp.ts`, `bin/check-models.ts`) each do it differently and each skip a phase, so the alternatives were a throwaway script that hand-rolls the boot or an HTTP route that shouldn't exist. `orm/context.test.ts` already carries a comment about the hypothetical `gemi db:seed` nobody could write.

Commands are `defineCommand(...)` chains under `app/commands`, run with `gemi run <name>`.

```ts
export default defineCommand("backfill-avatars")
  .describe("Regenerate avatar URLs for users created before the CDN move")
  .arg("since", { required: true, description: "ISO date to backfill from" })
  .option("dryRun", { type: "boolean" })
  .option("limit", { type: "number", default: 500 })
  .handle(async ({ args, options, line }) => {
    //  args.since   -> string   (required, so not `| undefined`)
    // options.limit -> number   (has a default, so not `| undefined`)
    // options.dryRun-> boolean
  });
```

```
$ gemi run                                  # list every command
$ gemi run backfill-avatars 2024-01-01 --dry-run
$ gemi run backfill-avatars --help          # usage, from the same spec the parser reads
```

## Why a chain and not a class

Every other extension point in gemi is a class — `Job`, `CronJob`, `Controller`, `Middleware` — and a class cannot type its own handler. TypeScript does not contextually type an overriding method's parameters from the base-class method it overrides, so `static args` beside `run(ctx)` leaves `ctx` an implicit `any`. So does a configured base class returned by a `Command.define()`, one indirection further away. The author's only way out is restating the type by hand in every command, which is the ceremony the schema was supposed to remove and which nothing enforces.

A **function expression passed as an argument is contextually typed**. That is the whole mechanism, and it is why `.handle(fn)` is the shape rather than sugar.

`.handle()` returns a `Command` subclass, so nothing downstream knows the difference: discovery finds it by prototype chain, and the registry, parser and `--help` read the same statics a hand-written class would carry.

`console/builder.test-d.ts` is the only place that inference is observable — a regression in it would leave every other test in the repo green. It includes an explicit `not.toBeAny()`, because `expectTypeOf<any>().toEqualTypeOf<string>()` passes and without it the suite couldn't tell a working builder from one that had stopped inferring. Verified by breaking the inference on purpose and watching the file go red.

## Why `gemi run` spawns a process

`dist/bin/gemi.js` bundles its own copy of gemi while the app's Kernel resolves `gemi/*` to source — the constraint `app/App.ts:23-32` documents. Discovery decides membership by walking the prototype chain, so a `Command` base imported into the CLI bundle is a *different class object* from the one `app/commands/*.ts` extends: every command would be invisible, silently. `--preload gemi/bun/preload` and the app's `app/preload.ts` can only be installed at process start besides.

So the CLI resolves `gemi/console/run` out of the app's own `node_modules` and spawns it, the way `dev` and `start` already do. That entrypoint goes in the **existing** `Bun.build` call rather than one of its own: `splitting: true` is what keeps `Command` in a chunk shared with `dist/services/index.js`, so the published runner and the `gemi/services` an app imports hold the same class object. Confirmed against a real build — one copy of the class in the whole `dist/`, reached by both entrypoints.

## Notable choices

- **Boots both phases**, so a command sees exactly what a request handler sees. But listing, an unknown name and `--help` all resolve *before* `waitForBoot()` — the three things an operator does most often shouldn't open a database connection to answer a question about a directory.
- **`GEMI_NO_SCHEDULE=1`** stops `ScheduleServiceProvider` starting the schedule while still discovering it, so a four-minute backfill can't fire the app's whole cron schedule in a process nobody is watching. `app(Scheduler).jobs` still answers honestly. An env var rather than a config field because it's a property of *this process* — it's also how a deploy runs one cron dyno beside several web ones.
- **`app/commands` is walked by `gemi run` and not at boot**, unlike `app/jobs` and `app/cron`. A registry no request consults shouldn't cost every dev server and every production start an import of every file under it.
- **Commander needs `enablePositionalOptions()` + `passThroughOptions()`.** Not `allowUnknownOption()`, which was the first thing tried: it collects unknown options into a separate list and concatenates them *after* the operands, so `gemi run job --flag value` arrives as `job value --flag`. Order silently mangled is the worst failure available to an argument forwarder.
- **An unfinished chain is refused by name.** A `defineCommand(...)` that never calls `.handle()` is a plain object, so discovery would drop it alongside every helper a command file exports and the command would vanish from the listing with nothing raised — the silence #322 and #323 exist to remove, reintroduced by a new authoring surface.
- **Two commands with one name is a hard error naming both**, where `QueueManager.useJobs` and `Scheduler.start` keep the first and warn. Those have a server to keep booting; here there's no boot to protect and a person is waiting, and running the wrong one of two `db:seed`s is worse than running neither.
- **The config slice is `command`, not `console`** — `app/config/console.ts` arrives in a Kernel as `import console from "../config/console"`, shadowing the global `console` for that module.
- **Three output helpers and no more.** `line` (stdout), `error` (stderr), `fail`. The load-bearing thing is stream discipline, not colour: a diagnostic on stdout is what breaks `TOTAL=$(gemi run count-users)`, and it stays invisible until someone pipes the command somewhere.

## Two things found on the way

- **`console/` was missing from `tsconfig.json`'s `include`**, which meant `tsc --noEmit` never looked at any of the new code and the type-test file was decorative. Adding it surfaced two real errors, including that this package compiles with `strictNullChecks: false` — under which `if (!parsed.ok)` does **not** narrow a discriminated union while `=== false` does.
- **`runConsole` was resolving the Kernel against `rootDir` but letting the commands directory fall back to `process.cwd()`.** Identical in the spawned process, which is exactly how a parameter quietly stops being one; the fixture tests caught it.

## Testing

- `console/parse.test.ts` (25) — the parser, pure. Aimed at the cases where a wrong answer still looks like an answer: `--limit abc` arriving as `NaN`, a default winning over a supplied value, a flag after `--` read as a flag.
- `console/builder.test-d.ts` (12) — the inference, plus `not.toBeAny()`.
- `console/builder.test.ts` (16) and `CommandRegistry.test.ts` (10) — declaration-time refusals and the duplicate-name message.
- `console/runner.test.ts` (26) — real fixture projects on disk: the config slice's declared-wins rule, exit codes, stream discipline, that cheap paths don't boot, and that cron doesn't start.
- `services/discovery.test.ts` — `discoverCommands` beside the job/cron cases, and the `GEMI_NO_SCHEDULE` behaviour both ways.
- Full suite green (3292), `tsc` clean, `oxlint` 0 errors and 0 warnings in `console/`, full `bun run build` + `build:publish` ("20 exports, all resolved").
- End-to-end by hand in `templates/saas-starter`: listing, `--help`, real ORM read, exit codes including `fail(msg, 3)`, and a DB-touching command exiting in 0.6s rather than hanging.

## One thing to decide

The template ships **no** example command, per an explicit call during planning. Worth knowing what that costs: the cross-copy class-identity failure this whole design is shaped around produces an empty command list, "not found" for everything, and **a fully green unit-test suite** — tmpdir fixtures can't catch it because they import `defineCommand` by absolute path, which makes the identity hold by construction. Only a real project with a real `node_modules` can, and `templates/saas-starter` is the only one here.

The CI step added therefore asserts the empty-directory listing plus a non-zero exit for an unknown name — enough to prove the spawn, both preloads, the Kernel and the config slice lined up, but not the identity. Dropping a command into `templates/saas-starter/app/commands/` would turn it into a real end-to-end assertion; happy to add one if you want it.

https://claude.ai/code/session_01FeYiZuQjPgdSNAqKG3L6CK

### DIFF
--- a/packages/gemi/console/builder.ts
+++ b/packages/gemi/console/builder.ts
@@ -1,0 +1,270 @@
+import {
+  Command,
+  type ArgSpec,
+  type CommandArgument,
+  type CommandClass,
+  type CommandOption,
+  type CommandResult,
+  type OptionSpec,
+} from "./Command";
+import type { CommandContext } from "./context";
+
+/**
+ * Declaring a command, in the one shape that can type its own body.
+ *
+ * ### Why a builder and not a class
+ *
+ * The point of declaring arguments and options at all is that the handler then
+ * knows what it has. A class body cannot deliver that, and the reason is a hard
+ * TypeScript rule rather than a matter of arrangement: **TypeScript does not
+ * contextually type a method's parameters from the base-class method it
+ * overrides.** So both class-shaped designs fail in the same place:
+ *
+ * ```ts
+ * class Backfill extends Command {
+ *   static args = [{ name: "since", required: true }];
+ *   async run(ctx) {}                       // implicit any
+ * }
+ *
+ * class Backfill extends Command.define({ since: { required: true } }) {
+ *   async run(ctx) {}                       // still implicit any — the
+ * }                                         // configured base's signature does
+ *                                           // not flow into the override
+ * ```
+ *
+ * The author's only way out is to restate the type by hand, which is the
+ * ceremony the descriptors existed to remove — and which nothing enforces, so a
+ * wrong restatement compiles.
+ *
+ * A **function expression passed as an argument is contextually typed**. That is
+ * the entire mechanism here, and it is why this is a builder rather than sugar:
+ * `.handle(fn)` carries the accumulated argument and option types into `fn`'s
+ * parameter with nothing written by hand and nothing to keep in step.
+ *
+ * ### What it produces
+ *
+ * A `Command` subclass, so nothing downstream knows the difference: discovery
+ * finds it by prototype chain, and the registry, parser and `--help` read the
+ * same statics they would read off a hand-written class. The builder is an
+ * authoring surface over one runtime shape, not a second mechanism.
+ */
+
+/**
+ * Marks a builder that was never finished. `discoverCommands` looks for this on
+ * the exports it is about to discard — see the note on `handle` below.
+ */
+export const BUILDER_BRAND: unique symbol = Symbol.for("gemi.console.builder");
+
+/** Flattens an accumulated intersection so editor hovers stay readable. */
+type Simplify<T> = { [K in keyof T]: T[K] } & {};
+
+/**
+ * What the handler sees for one argument.
+ *
+ * The order matters: `variadic` first because it wins over everything, then the
+ * two ways an argument is guaranteed present. Anything else can be absent and
+ * says so.
+ */
+export type ArgValue<S> = S extends { variadic: true }
+  ? string[]
+  : S extends { required: true }
+    ? string
+    : S extends { default: string }
+      ? string
+      : string | undefined;
+
+/** What the handler sees for one option. */
+export type OptionValue<S> = S extends { type: "boolean" }
+  ? boolean
+  : S extends { type: "number" }
+    ? S extends { default: number }
+      ? number
+      : S extends { required: true }
+        ? number
+        : number | undefined
+    : S extends { default: string }
+      ? string
+      : S extends { required: true }
+        ? string
+        : string | undefined;
+
+export interface CommandBuilder<
+  A extends Record<string, unknown>,
+  O extends Record<string, unknown>,
+> {
+  /** One line, shown in the listing and at the top of `--help`. */
+  describe(text: string): CommandBuilder<A, O>;
+
+  /**
+   * A positional argument. Declared in the order it is read from the command
+   * line; a `variadic` one must be last.
+   */
+  arg<N extends string, const S extends ArgSpec = {}>(
+    name: N,
+    spec?: S,
+  ): CommandBuilder<A & { [K in N]: ArgValue<S> }, O>;
+
+  /**
+   * A flag. Declared in camelCase and spelled in kebab-case on the command line,
+   * so `.option("dryRun", …)` is `--dry-run` for the operator and
+   * `options.dryRun` in the handler — no string key nobody can autocomplete, and
+   * no CLI flag with a capital letter in it.
+   */
+  option<N extends string, const S extends OptionSpec>(
+    name: N,
+    spec: S,
+  ): CommandBuilder<A, O & { [K in N]: OptionValue<S> }>;
+
+  /**
+   * The body, and the terminal of the chain.
+   *
+   * A builder that never reaches here is not a class, so discovery would skip it
+   * and the command would simply vanish from `gemi run` with nothing raised —
+   * the silence #322 and #323 exist to remove, reintroduced by a new authoring
+   * surface. `discoverCommands` therefore looks for `BUILDER_BRAND` among the
+   * exports it discards and refuses the file by name.
+   */
+  handle(
+    fn: (ctx: CommandContext<Simplify<A>, Simplify<O>>) => CommandResult,
+  ): CommandClass;
+}
+
+/**
+ * Refused early, where the author is looking.
+ *
+ * The whitespace and brace checks are aimed at one specific reader: somebody who
+ * knows Laravel and reaches for `defineCommand("mail:send {user} {--queue}")`.
+ * Without this they get a command registered under a name containing a space,
+ * which can never be typed and never runs, and no explanation.
+ */
+function assertUsableName(name: string): void {
+  if (name.trim() === "") {
+    throw new Error("A command name cannot be empty.");
+  }
+
+  if (/[{}]/.test(name) || /\s/.test(name)) {
+    throw new Error(
+      `Invalid command name ${JSON.stringify(name)}. gemi does not parse a ` +
+        `signature string; the name is the whole of it. Declare arguments ` +
+        `with .arg() and flags with .option():\n\n` +
+        `    defineCommand("mail:send")\n` +
+        `      .arg("user", { required: true })\n` +
+        `      .option("queue", { type: "boolean" })\n` +
+        `      .handle(async ({ args, options }) => { ... })`,
+    );
+  }
+}
+
+export function defineCommand(name: string): CommandBuilder<{}, {}> {
+  assertUsableName(name);
+
+  const args: CommandArgument[] = [];
+  const options: CommandOption[] = [];
+  let description = "";
+
+  const builder = {
+    [BUILDER_BRAND]: name,
+
+    describe(text: string) {
+      description = text;
+      return builder;
+    },
+
+    arg(argName: string, spec: ArgSpec = {}) {
+      // Checked here rather than at parse time because both of these are
+      // mistakes in the declaration, and the declaration is what the author is
+      // looking at right now. At parse time they would surface as a confusing
+      // usage error in front of an operator who cannot fix them.
+      if (spec.required && spec.default !== undefined) {
+        throw new Error(
+          `Argument "${argName}" of command "${name}" is both required and has ` +
+            `a default. An argument that is supplied when absent cannot also be ` +
+            `absent — drop one.`,
+        );
+      }
+
+      const variadic = args.find((existing) => existing.variadic);
+      if (variadic) {
+        throw new Error(
+          `Argument "${argName}" of command "${name}" is declared after the ` +
+            `variadic argument "${variadic.name}", which collects everything ` +
+            `that follows, so "${argName}" could never receive a value. A ` +
+            `variadic argument must be declared last.`,
+        );
+      }
+
+      if (args.some((existing) => existing.name === argName)) {
+        throw new Error(
+          `Command "${name}" declares the argument "${argName}" twice.`,
+        );
+      }
+
+      args.push({ ...spec, name: argName });
+      return builder;
+    },
+
+    option(optionName: string, spec: OptionSpec) {
+      if (options.some((existing) => existing.name === optionName)) {
+        throw new Error(
+          `Command "${name}" declares the option "${optionName}" twice.`,
+        );
+      }
+
+      const alias = "alias" in spec ? spec.alias : undefined;
+      const clash =
+        alias === undefined
+          ? undefined
+          : options.find(
+              (existing) => "alias" in existing && existing.alias === alias,
+            );
+      if (clash) {
+        throw new Error(
+          `Options "${clash.name}" and "${optionName}" of command "${name}" ` +
+            `both use the alias "-${alias}".`,
+        );
+      }
+
+      options.push({ ...spec, name: optionName });
+      return builder;
+    },
+
+    handle(fn: (ctx: any) => CommandResult) {
+      const commandName = name;
+      const commandDescription = description;
+
+      const cls = class extends Command {
+        static commandName = commandName;
+        static description = commandDescription;
+        static args = args;
+        static options = options;
+
+        run(ctx: any) {
+          return fn(ctx);
+        }
+      };
+
+      // An anonymous class expression reports `name === ""`, and the duplicate-
+      // name error's whole job is to say *which two*. Give it the one name it
+      // has.
+      Object.defineProperty(cls, "name", {
+        value: commandName,
+        configurable: true,
+      });
+
+      return cls;
+    },
+  };
+
+  return builder as unknown as CommandBuilder<{}, {}>;
+}
+
+/** True for a builder that never called `.handle()`. */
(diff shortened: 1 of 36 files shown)


packages/gemi/console/run.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+import { runConsole } from "./runner";
+
+const [name, ...rest] = process.argv.slice(2);
+
+// commander forwards a `--` through as a literal operand, so `gemi run x --
+// --verbose` arrives here with it still attached. Dropping a leading one makes
+// the escape hatch mean what it looks like it means; a second `--` is the
+// command's own and `parse.ts` handles it.
+const argv = rest[0] === "--" ? rest.slice(1) : rest;
```
nstfkc: **Stripping the leading `--` destroys the documented escape hatch.**

Commander (with `enablePositionalOptions` + `passThroughOptions`) forwards `gemi run x -- --weird` to this handler as `name="x"`, `rest=["--", "--weird"]`. This line drops the `--`, so `parseArgv` receives `["--weird"]` and reports `Unknown option --weird for command "x".` and exits 1.

`docs/commands.md:174` (added in this PR) promises the opposite: "A `--` is available for a genuinely hostile tail (`gemi run x -- --weird`) but is never required." As written, the only documented way to pass a flag-shaped positional needs `-- --` typed twice.

`runner.test.ts` calls `runConsole` directly and never exercises `run.ts`, so nothing covers this path.

packages/gemi/console/runner.ts
```diff
@@ -0,0 +1,249 @@
...
+    // same split `bin/gemi.ts` already applies to `CheckModelsError`.
+    //
+    // `DiscoveryError` belongs in this list rather than the other because every
+    // way of raising one — a command file that will not import, a builder chain
+    // that was never finished — is a mistake in the application's own source
+    // with the fix named in the message. A stack above it points at this
+    // runner's frames, which is not where anybody needs to look.
+    if (error instanceof CommandFailed) {
+      err(error.message);
+      return error.code;
```
nstfkc: **`CommandFailed.code` reaches `process.exit` unvalidated**, bypassing the 0–255 guard `exitCodeFor` applies to returned values.

`exitCodeFor` exists precisely because `return users.length` would otherwise turn 300 backfilled users into exit status 44 — but `ctx.fail(message, code)` takes the same kind of number and skips the check. A handler writing ``ctx.fail(`${bad.length} rows failed`, bad.length)`` with 256 bad rows returns 256, `run.ts` calls `process.exit(256)`, and the shell sees **status 0** — a failed command reported as success, so a CI step or `&&` chain proceeds. 300 surfaces as 44; 512 is 0 again.

Routing this return through `exitCodeFor` (or clamping in the `CommandFailed` constructor) closes it.

packages/gemi/console/parse.ts
```diff
@@ -0,0 +1,287 @@
...
+          );
+        } else {
+          const value = inline === undefined ? true : inline === "true";
+          raw.set(option.name, negated ? !value : value);
+        }
+        index += 1;
+        continue;
+      }
+
+      const value = inline ?? argv[index + 1];
```
nstfkc: **A value-taking option swallows the next token even when it is a flag or a bare `--`.**

Nothing here checks that the next token isn't option-shaped; line 148 only rejects end-of-argv, and line 155 then skips the swallowed token entirely.

Verified by running the parser: with `.option("msg", { type: "string" })` and two positionals, `["--msg", "--", "x"]` parses to `{ options: { msg: "--" }, args: { a: "x" } }` — the escape-hatch `--` is eaten as the value.

Same line makes `gemi run notify --msg --dry-run` set `msg` to the literal `"--dry-run"` and leave `dryRun` false: the operator's dry run executes for real, with a nonsense message body and no error printed.

packages/gemi/console/parse.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    positionals.push(token);
+    index += 1;
+  }
+
+  const args: Record<string, string | string[] | undefined> = {};
+  let cursor = 0;
+
+  for (const declared of spec.args) {
+    if (declared.variadic) {
```
nstfkc, line 281: **`required` is silently ignored on a variadic argument.**

The variadic branch `continue`s before the `required` check below can run, and the builder doesn't refuse the combination. Verified: `.arg("files", { required: true, variadic: true })` with argv `[]` returns `{ ok: true, args: { files: [] } }`.

Meanwhile `usage.ts:argSlot` prints `<files...>`, telling the operator it's mandatory. So `gemi run import-files` with the arguments forgotten boots the app, runs the handler over an empty array, prints nothing, and exits 0 — a no-op that reads as a successful import instead of the usage error the declaration asked for.

packages/gemi/console/runner.ts
```diff
@@ -0,0 +1,249 @@
...
+      return 1;
+    }
+
+    const spec = {
+      commandName: command.commandName,
+      args: command.args,
+      options: command.options,
+    };
+
+    if (wantsHelp(spec, params.argv)) {
```
nstfkc: **`wantsHelp` runs before `parseArgv`, so it can't tell a `--help` that is an option's *value* from a request for help.**

`wantsHelp` (parse.ts:74-84) is a flat token scan with no notion of option arity. `gemi run notify --message --help` — or `--message "--help"`, which arrives as the same token — prints usage and `return 0` here, before `waitForBoot()`. The notification is never sent and the exit status is 0, so a script or cron wrapper branching on the exit code records a success for work that did not happen.

Parsing first and treating help as a parsed option (or having `wantsHelp` skip tokens in value position) resolves both this and the `-h`-alias issue below.

packages/gemi/console/parse.ts
```diff
@@ -0,0 +1,287 @@
...
+ * Whether this invocation is asking for help rather than running.
+ *
+ * A command that declares its own `help` option or `-h` alias keeps them — the
+ * application's schema wins over the framework's convenience, because the
+ * alternative is a flag that silently does something other than what the command
+ * documents.
+ */
+export function wantsHelp(spec: CommandSpec, argv: readonly string[]): boolean {
+  const { byName, byAlias } = optionIndex(spec.options);
+  if (byName.has("help") || byAlias.has("h")) return false;
```
nstfkc: **A `-h` alias on any option disables `--help` entirely.**

Given `.option("host", { type: "string", alias: "h" })` — no `help` option declared — `wantsHelp` returns false and `parseArgv(["--help"])` returns `Unknown option --help for command "c".` So `gemi run deploy --help` exits 1, and the usage block it then prints lists both `-h, --help  Show this message` (unconditional, `usage.ts:89`) and `-h, --host <string>` — two contradictory `-h` entries, one of which doesn't work.

The comment above justifies yielding a *declared* `help` option or `-h` alias; killing the long `--help` the command never claimed is collateral. Splitting the guard (`byName.has("help")` suppresses `--help`, `byAlias.has("h")` suppresses only `-h`) keeps the intent.

packages/gemi/console/run.ts
```diff
@@ -0,0 +1,34 @@
...
+// package's inherited `module: NodeNext` — the same TS1309 that keeps `bun/` out
+// of tsconfig.json's `include`. Bun runs it as ESM either way; this just avoids
+// buying a nested tsconfig for two lines.
+runConsole({ rootDir: process.cwd(), name, argv }).then((code) => {
+  // Explicit, and not belt-and-braces. `kernel.destroy()` stops the scheduler,
+  // but an open database pool, a Redis client or an unref'd timer will happily
+  // hold the loop open after the work is done — a command that finishes and then
+  // sits there, which is the failure `kernel/exit.test.ts` exists for and the
+  // one `orm/context.test.ts` names by example.
+  process.exit(code);
```
nstfkc: **`process.exit` immediately after `runConsole` truncates piped stdout above the 64 KB pipe buffer.**

`context.ts`'s `processWriters()` writes via `process.stdout.write`, which is async on a pipe. A handler emitting one large chunk — `line(JSON.stringify(rows))`, a CSV body, a report — loses everything past the buffer under `gemi run export-users | gzip > out.gz` or `TOTAL=$(gemi run report)` (the exact idiom `context.ts` says the three-stream design exists to protect).

Reproduced on Bun 1.3.14: `process.stdout.write("x".repeat(2_000_000)); process.exit(0)` delivers exactly 65536 bytes through a pipe — and the same 65536 through `Bun.spawn({stdout:"inherit"})`, which is how `bin/gemi.ts` runs it — while the same script redirected to a file delivers all 2000001. Exit status is 0, so the consumer stores a silently truncated file.

The new CI step `output=$(timeout 60 bunx gemi run)` is on the same pipe path. Awaiting a drain (or setting `process.exitCode` and letting the process end naturally) avoids it.

nstfkc (commented):
Automated high-effort review of the new `gemi run` console-command subsystem. 10 findings below, most severe first; each was independently verified (several by running the parser or reproducing under Bun). The main cluster is argument handling in `console/`, plus two exit-status/output-integrity issues.

Not reported but checked and dismissed: the static console import on the boot path, the `onSkipped` base-class change, `optionIndex` rebuilds, `enablePositionalOptions` placement. Verified but omitted as lower value: the CI grep assertion, two `GEMI_NO_SCHEDULE` robustness concerns, and three duplication cleanups (`editDistance`, `loadKernel`, the declared-vs-discovered rule).

packages/gemi/services/discovery.ts
```diff
@@ -144,3 +151,70 @@ export async function discoverCronJobs(
...
+  warnIfSourceIsMissing(resolved, "commands", "command");
+
+  const unfinished: string[] = [];
+
+  const commands = await discoverClasses(
+    resolved,
+    Command,
+    [],
+    (value, file) => {
+      if (isUnfinishedBuilder(value)) unfinished.push(file);
```
nstfkc, line 181: **The unfinished-builder refusal keys off *any* exported builder object, so one file can abort all of `gemi run`.**

A natural attempt at sharing option declarations —
```ts
export const base = defineCommand("reindex").option("verbose", { type: "boolean" });
export default base.handle(async () => {});
```
— leaves the still-branded builder visible to `onSkipped`. `refuseUnfinishedBuilders` throws, and `runConsole` returns 1 before the registry is built: `gemi run` with no name lists nothing and *every other command in the project* becomes unrunnable, under a message ("exports a command builder that never called `.handle()`, so there is no command to run") that is untrue for that file.

Clearing the brand inside `handle()`, or tracking chains that were never handled rather than exports that are still builders, makes the check say what it means.

packages/gemi/console/runner.ts
```diff
@@ -0,0 +1,249 @@
...
+      err(error.message);
+      return error.code;
+    }
+
+    if (error instanceof ConsoleError || error instanceof DiscoveryError) {
+      err(error.message);
+      return 1;
+    }
+
+    err(String((error as Error)?.stack ?? error));
```
nstfkc, line 193: **Registry refusals print as framework stack traces.**

`CommandRegistry` raises its duplicate-name and `commandName`-unset refusals as plain `Error`s with messages written for the terminal, and `new CommandRegistry(...)` at line 93 is inside this `try`. The catch above only treats `CommandFailed`, `ConsoleError` and `DiscoveryError` as sentences, so those land here and the operator reads the message buried under `at new CommandRegistry (.../CommandRegistry.ts:52)` plus the runner's frames.

That's the outcome the `DiscoveryError` note two lines above argues against, for the identical class of mistake: an error in the application's own source with the fix named in the message. Making these `ConsoleError`s (or adding the registry's error type to the sentence branch) is enough.

packages/gemi/console/builder.ts
```diff
@@ -0,0 +1,270 @@
...
+    },
+
+    handle(fn: (ctx: any) => CommandResult) {
+      const commandName = name;
+      const commandDescription = description;
+
+      const cls = class extends Command {
+        static commandName = commandName;
+        static description = commandDescription;
+        static args = args;
```
nstfkc: **The generated class aliases the builder's live spec arrays instead of copying them.**

`static args = args` / `static options = options` store the same array instances the builder keeps mutating via `args.push(...)` / `options.push(...)`. So:

```ts
const b = defineCommand("x");
const C = b.handle(fn);
b.arg("late");            // retroactively added to C
```

and two `.handle()` calls on one builder yield classes whose schemas stay permanently linked. `--help` and `parseArgv` both read these statics, so printed usage and parse behaviour change under a command that was already "finished".

`static args = [...args]` / `static options = [...options]` closes it.

PR #392, opened 2026-08-10
Close #391: a to-one whose foreign key is on the parent takes `upsert`

### LINKED ISSUE #391: A to-one whose foreign key is on the parent has no supported spelling for `upsert`
`#354` closed the to-one whose foreign key lives on the **child**. The mirror shape — a to-one whose foreign key lives on the **parent**, the owning side — still has no supported spelling for `upsert`, and it is refused during compilation, so on every call rather than only on a missing row.

Found the same way #354 was: porting a real Prisma application onto the ORM, in the first controller that edits the relation.

## The shape

```prisma
model Workflow {
  previewImageId Int?
  previewImage   Image? @relation(fields: [previewImageId], references: [id], onDelete: Cascade)
}
```

Prisma accepts the nested upsert on the parent:

```ts
prisma.workflow.update({
  where: { publicId },
  data: { previewImage: { upsert: { create: { src, width, height }, update: { src, width, height } } } },
  include: { previewImage: true },
})
```

`Workflow.update` with the identical tree throws from `planOwningSide` (`compile/nested-writes.ts`, the `if (key === "upsert")` branch):

> gemi ORM does not support 'data.previewImage.upsert' yet (Workflow.update). 'previewImage' is a to-one, and this row holds its foreign key — so an absent far row would have to be created and then written back to a Workflow that has already been inserted. That is not implemented. Upsert the Image directly, then 'connect' it.

## Why the suggested workaround does not transfer from #354

The message says *"Upsert the `Image` directly, then `connect` it."* On the foreign side that works, because the child carries a unique back-reference to key the upsert on — which is exactly what `#354`'s fix uses:

```ts
Image.upsert({ where: { thumbnailForMediaId: id }, create: {...}, update: {...} })
```

On the owning side there is no such key. `Image` has no column pointing back at `Workflow`; the link is `Workflow.previewImageId`. So there is nothing to put in the `where`, and the caller has to read the parent's foreign key first and branch by hand:

```ts
await Model.transaction(async () => {
  const current = await Workflow.findUnique({ where: { publicId }, select: { previewImageId: true } })
  data.previewImage = current?.previewImageId
    ? { update: { src, width, height } }
    : { create: { src, width, height } }
  return await Workflow.update({ where: { publicId }, data, include: { previewImage: true } })
})
```

That is a correct workaround and it is what the port now does, but it is one extra round trip plus a transaction the Prisma call did not need, and every call site that edits an owned to-one has to rediscover it.

## Suggested semantics

The compiler already knows the parent's foreign key at plan time, and both operands are implemented on this side (`create` and `update` are the branches immediately around the refusal). The read-then-branch above is mechanical:

- FK null → the `create` branch, then set the FK, which is what a nested `create` on this side already does.
- FK non-null → the `update` branch against the far row that FK identifies.

Since the parent row is written in the same statement, the create case is the one the docblock flags as needing a write-back. If that ordering is the blocker, refusing `upsert` only when the FK is null — and serving it when the row exists — would still be a strict improvement over refusing at compile time, because the compiler cannot know which case a given call is.

## Either way, two smaller things

1. **`docs/orm.md`'s *Not in scope* list does not mention this shape** — the same gap `#354` called out for the foreign side. A caller meets it at run time.
2. **The message points at a workaround that does not exist on this side.** "Upsert the `Image` directly, then `connect` it" reads as actionable and is not: there is no unique key on `Image` to upsert against. Naming the read-then-branch instead would save the reader the detour.

Found on `0.51.0-rc.14`. Related: `#354` (foreign side, closed), `#116`, `#360` / `#361` (occupied to-one collides where Prisma detaches).

nstfkc:
`planOwningSide` refused `upsert` **by name, at compile time**, so a nested upsert through the owning end of a to-one failed on every call rather than on a missing row. It is the mirror of #354 — and the more likely of the two to be met, since it is the direction a `previewImageId`-style link runs in.

## The refusal named a write-back this side does not make

> an absent far row would have to be created and then written back to a `<parent>` that has already been inserted

It has not been inserted. The statement the caller asked for is the one this step runs **before**, so a created row's key reaches it as a contribution — exactly what `create` and `connectOrCreate`, the operands either side of this one, already do. What was genuinely missing was deciding the branch, and that is a lookup: the parent's own foreign key.

```
nothing linked                    create it and take the new key
linked, and the far row matches   update it, key written back unchanged
linked, and it does not           RecordNotFoundError, nothing written
```

The suggested workaround did not transfer either — *"upsert the `<child>` directly, then `connect` it"* needs a unique back-reference on the far model, and on this side the link is this row's column. That is the half of the issue a wording fix would not have closed.

## The create branch is for a parent that points at nothing

The third row is the rule the rest of this side already follows: **no owning-side operand repoints a link the caller did not name a new target for.** Reaching the create branch from a *linked* parent is exactly that repoint, and silently — the row it moves off is by construction one the lookup did not return. Three conditions arrive there, and one answer covers them:

- the operand's `where` matched nothing, so a filter written to *guard* the write would have caused a bigger one;
- the linked child is hidden by the child's own policies — soft-deleted, another tenant's — where "mint a second one and repoint" is most expensive and least visible;
- the foreign key is dangling, the one case where creating is arguably right, and unreachable with the constraint enforced.

`RecordNotFoundError` for all three, which is what the `update` operand two branches up already gives for its own filtered miss.

## Measured, on all three owning shapes the fixture has

`Profile.user` (one-to-one), `User.organization` (many-to-one) and the composite `LedgerNote.ledger` / `LedgerEntry.ledger` — they answer alike. 19 differential cases (`O16`–`O16h`, `M17`–`M17e`, five composite), five policy cases, and the compiler pins. Nothing displaces on either branch that writes: the update branch keeps the link it found, the create branch mints a row nothing can already point at.

**The update branch can move the key it writes back.** `update: { code: "renamed" }` on a relation referencing `code` rewrites the far row out from under the contribution. The new value is taken from the payload rather than from a re-read, because on a composite link the referenced columns are routinely the child's `@@id` too — so there is no key left to find the row by afterwards — and the payload's answer is the one `ON UPDATE CASCADE` computes, which is Prisma's whole answer here. It restricts a referenced column to a value that names itself: `{ set }` derives, `{ increment: 1 }` is refused.

## What the operand refuses on this side, all at plan time

- **A second operand on the same relation.** `connect`, `connectOrCreate`, `create`, `disconnect` and `upsert` all write one foreign-key column, and `updateAssignments` folds them into one assignment — so the last to sort won and the rest were dropped without a word. `{ disconnect: true, upsert }` edited the current organization and left the user in it. `disconnect: false` is exempt; it plans nothing.
- **A nested write inside `upsert.update`**, which reached `assertNoNestedWrites` inside the child's `updateMany` and reported a model and operation the caller never wrote — the #85/#101 shape. Refused with the asymmetry stated: the `create` half does take them.
- **An operator on the column the key references**, for the reason above.
- **`create: null` / `update: null`**, which passed both asserts and reached `insertColumns`, where `data === null` is admitted and inserts a row of nothing but defaults.
- **An unknown column in the operand's `where`**, which used to be compiled only on the branch that had a linked row — so the same query succeeded or raised depending on the data. Now `UnknownFieldError` at plan time, the error `compileWhere` raises one layer down, and it covers the `update` operand too.

## Two reads that were deciding branches wrongly

**The parent's own foreign key, read redacted.** `preScoped` skips `applyPolicies` but not `applyRedaction`, and `redactNullable` permits exactly the nullable columns an optional foreign key is made of — so a policy hiding `Note.folderId` made a linked note read as unlinked. `RelationExecutor.exec` gains an `unredacted` flag for reads that are machinery, implemented as `runAsSystem`; it is only ever passed with `preScoped`, where the scope is already in `args`. The three sites that took this read — the `disconnect` filter arm, this branch, and `displaceSibling` — now share one `readOwnKey` instead of three copies each re-arguing its flags.

**The child read under one scope, written under another.** `updateMany` answers a policy that scopes reads and mutations differently with `{ count: 0 }` rather than by raising, so the caller got a successful `update` whose nested payload was never applied. The count is checked.

## One divergence pinned rather than matched

**Prisma emits invalid SQL when a `where` is present and its branch decides to create.** It splices the operand's filter into the *parent's* `UPDATE` — `where ("id" in (?) and "User"."name" = ?)` — and answers P2022 on SQLite, having already inserted the far row inside the transaction it then rolls back. Both clients now write nothing on that call; what differs is the class — `notFound` here against `other` there — so the case sits in `describe("where the two clients still disagree")` with gemi's asserted by class and Prisma's only asserted as a failure, since the statement is invalid in a dialect-specific way.

The `@updatedAt` stamp on the update branch is the other known one, and it is gemi's: the contribution is in the SET list whether or not the branch that changes it ran, because a plan's text cannot depend on a value read at bind time. Same mechanism and same trade as the owning-side `disconnect` filter arm. Say the word if you want an issue filed for it, the way #359/#360 were.

## Also in here

- `docs/orm.md` documents the operand from this end: what the create branch is for, the refusals, the stamp, and Prisma's bug. `UPGRADE.md` gains the stamp's second instance and a section on the newly-refused operand pairs, which is a real behaviour difference rather than only a refusal — Prisma applies both in sequence. `llms-full.txt` re-synced.
- `write.test.ts`'s "a to-one answers the same on both sides" table drops to one refusal (`delete`) and gains the plan-shape assertion — a `before` step, not an `after` one.
- The file header no longer claims seven operands throw.

## Verification

`packages/gemi` (3213 tests) and `templates/saas-starter` (878) both green, plus `test:types`, `typecheck` and `oxlint`. Each new pin was checked to fail with its own fix reverted. **The Postgres job has not been run** — no local daemon.

### DIFF
--- a/packages/gemi/orm/compile/nested-writes.ts
+++ b/packages/gemi/orm/compile/nested-writes.ts
@@ -38,9 +38,17 @@ import { matchUniqueKey, type RefusalOrigin } from "./unique";
  * `createMany` is the second shape and only exists on the foreign side: the same
  * rows as a nested `create`, in one statement rather than one per row.
  *
- * Everything else in Prisma's nested-write grammar — `set`, `disconnect`,
- * `update`, `upsert`, `delete`, `deleteMany`, `updateMany` — throws
- * `UnsupportedQueryError` naming the operation *and the reason*; see `REFUSED`.
+ * The rest of Prisma's nested-write grammar — `set`, `disconnect`, `update`,
+ * `upsert`, `delete`, `deleteMany`, `updateMany` — is implemented too, and this
+ * paragraph used to say that all seven throw. What is left refused is refused
+ * by *shape* rather than by name, at the site that knows why: the four
+ * collection operands on a to-one, which Prisma's to-one input does not carry
+ * either; `delete` through a key on this row, which would remove a row the
+ * statement is not about (#391 left it the only one of the seven still refused
+ * on that side); everything in {@link EXISTING_ROW_ONLY_STATEMENTS} under a
+ * `create`; and the four that reach the far row *through* an implicit
+ * many-to-many's pairs. `REFUSED` — the table this sentence pointed at — is
+ * empty, and its docblock says why that is worth reading.
  *
  * **The line this file draws has two halves: which rows an operand can name,
  * and whose columns it writes.**
@@ -965,25 +973,6 @@ function planOwningSide(
   }
 
   /**
-   * `upsert` through a to-one, refused **by name**.
-   *
-   * The far row is identified by *this* row's foreign key, so an absent one
-   * means creating the far row and then writing back to a parent that has
-   * already been inserted — a shape no other operand here needs, and the
-   * reason this side is not implemented.
-   *
-   * Named rather than left to fall through, which is what it did: with no
-   * branch of its own it reached the `connect` handling below and reported
-   * `'where' yet (Organization.update.organization.connect)` — a different
-   * operand, a different model, and a claim that `{ id: 1 }` is not a unique
-   * field when it is. The real failure was that a `{ where, create, update }`
-   * operand was being read as a connect key.
-   *
-   * That is the shape #85 was filed for and #101 fixed on its own path: a
-   * refusal that misnames its origin sends the reader to a query they did not
-   * write.
-   */
-  /**
    * The list operands, refused by name on a to-one.
    *
    * `set`, `updateMany` and `deleteMany` describe *many* rows, and this side
@@ -1007,17 +996,184 @@ function planOwningSide(
     );
   }
 
+  /**
+   * `upsert` through a to-one whose key is on **this** row — update the far row
+   * this one points at, or mint one and point at it (#391).
+   *
+   * **Its refusal named a write-back that this side does not need**: *"an
+   * absent far row would have to be created and then written back to a
+   * `<parent>` that has already been inserted"*. It has not been inserted. The
+   * statement the caller asked for is the one this step runs *before*, so a
+   * created row's key reaches it as a contribution — which is exactly what the
+   * `create` and `connectOrCreate` branches above already do, and they are the
+   * two operands either side of this one. What was genuinely missing was
+   * deciding the branch, and that is a lookup, the same one `connectOrCreate`
+   * has been deciding since #94. The seventh refusal in this file to describe
+   * machinery one layer down.
+   *
+   * The refusal also pointed at a workaround that does not exist here: *"upsert
+   * the `<child>` directly, then 'connect' it"*. On the **foreign** side that
+   * works, because the child carries the back-reference to key the upsert on —
+   * `Image.upsert({ where: { thumbnailForMediaId: id }, … })`, which is what
+   * #354 shipped. On this side there is no such column: the link is this row's
+   * `previewImageId`, and `Image` has nothing pointing back. So the caller was
+   * being sent to a spelling with nothing to put in its `where`, and the real
+   * workaround — read the foreign key, then branch by hand into `create` or
+   * `update` — was left to be rediscovered per call site. That is the half of
+   * #391 a wording fix would not have closed.
+   *
+   * **A `before` step, and the branch is decided from the parent's own key.**
+   * The far row is identified by a column of this row, which the arguments do
+   * not carry, so the parent is read first — pre-scoped, exactly as the
+   * `disconnect` filter arm's parent read is and for the same reason: `args.where`
+   * has already been through this model's policies. Then:
+   *
+   *     linked, and the far row matches   ->  update it, key written back unchanged
+   *     linked, filter matches nothing    ->  the create branch, as Prisma's does
+   *     nothing linked                    ->  create it and take the new key
+   *
+   * Measured against 6.19.2/SQLite on `Profile.user` (a one-to-one) and
+   * `User.organization` (a many-to-one), which answer identically:
+   *
+   *     no where, nothing linked   insert the far row, update this row's key
+   *     no where, linked           select the far row, update it, **this row is
+   *                                not written at all**
+   *     where matching, linked     the same update, filtered
+   *
+   * **Nothing displaces**, and that is per branch rather than an omission. The
+   * update branch keeps the link it found; the create branch mints a row a line
+   * earlier, so nothing can already be pointing at it — `connectOrCreate`'s miss
+   * branch says the same thing in the same words, and Prisma agrees by
+   * construction: its log for the create branch is the insert and the repoint,
+   * with no incumbent lookup in it.
+   *
+   * **KNOWN DIVERGENCE — a `where` that matches nothing is served here and is a
+   * Prisma bug there.** Prisma splices the operand's filter into the *parent's*
+   * `UPDATE`, so the create branch emits
+   * `update "Profile" set "userId" = ? where ("id" in (?) and "User"."name" = ?)`
+   * — a column of the far table in this row's statement — and answers P2022,
+   * *"The column `main.User.name` does not exist in the current database"*,
+   * having already inserted the far row inside the transaction it then rolls
+   * back. Measured on SQLite on both relation shapes above and on the composite
+   * `LedgerNote.ledger`; the statement is invalid on Postgres too, though as a
+   * missing `FROM`-clause entry rather than an unknown column, so the pin
+   * asserts the failure and not its class. gemi runs the create branch the
+   * operand asks for. It is the same operand the *hit* branch serves correctly
+   * on both clients, so refusing the `where` outright would refuse a query
+   * Prisma answers; the divergence is pinned in `writes.differential.test.ts`
+   * rather than matched.
+   *
+   * **The hit branch writes this row's foreign key back unchanged**, which is
+   * the one thing here that costs anything: the contribution is in the SET list
+   * whether or not the branch that changes it ran, because a statement's text
+   * cannot depend on a value read at bind time. On a model carrying
+   * `@updatedAt` that stamps a row Prisma leaves alone. Pre-existing rather than
+   * introduced — it is the same mechanism, and the same trade, as the owning
+   * side's `disconnect` filter arm, whose measurements are in
+   * `updateAssignments`' docblock in `write.ts`.
+   *
+   * A branch of its own is also what keeps the refusals honest, and that is why
+   * this sits above `connect` rather than after it. With none, the operand fell
+   * through to the `connect` handling and reported *"'where' yet
+   * (Organization.update.organization.connect)"* — a different operand, a
+   * different model, and a claim that `{ id: 1 }` is not a unique field when it
+   * is. That is the shape #85 was filed for and #101 fixed on its own path.
+   */
   if (key === "upsert") {
-    throw new UnsupportedQueryError(
-      `data.${relation.name}.upsert`,
-      schema.name,
-      operation,
-      `'${relation.name}' is a to-one, and this row holds its foreign key — ` +
-        `so an absent far row would have to be created and then written back ` +
-        `to a ${schema.name} that has already been inserted. That is not ` +
-        `implemented. Upsert the ${relation.model} directly, then 'connect' ` +
-        `it.`,
-    );
+    // Both checks, in this order, and neither subsumes the other: the first is
+    // the to-one *shape* — an array, a scalar, a misspelled `where` that would
+    // otherwise be dropped and turn a filtered write into an unconditional one
+    // — and the second is `upsert`'s own grammar, that both payloads are there.
+    // `false` rather than `many`, for the reason its docblock gives: a to-one's
+    // `where` is an optional `WhereInput`, not a unique key.
+    assertToOneWriteOperand(schema, relation, child, key, operand, operation);
+    assertUpsertOperand(schema, relation, child, operand, operation, false);
+
+    out.before.push({
+      relation: relation.name,
+      operation: "upsert",
+      async run(args, context, executor) {
+        // Optional throughout, the way `connectOrCreate`'s reads are: the
+        // plan-time checks above ran on the *shape*, and `canonicalShape`
+        // records a key's presence, so a plan hit implies the same keys — but
+        // nothing in a step should be the first place a missing one is a
+        // `TypeError` rather than an error naming the argument.
+        const record = at(args) as Record<string, unknown> | null | undefined;
+
+        // Pre-scoped: `args.where` is the effective `where` this call already
+        // put through this model's policies, so re-applying them would `AND`
+        // the same predicate twice. One column of one row, inside the
+        // transaction the nested steps already run in.
+        const parent = (await executor.exec(
+          schema.name,
+          "findFirst",
+          { where: args?.where, select: keySelect(fkFields) },
+          true,
+        )) as Record<string, unknown> | null;
+
+        // A composite key with any column null names no row — see
+        // {@link isLinked} — so a half-written link is the *absent* case here
+        // and takes the create branch, as does a parent this statement will go
+        // on to match nothing of. Prisma reaches the second the same way: its
+        // parent read finds nothing, it inserts the far row, and the failure
+        // arrives from the write-back as P2025, rolling the insert back. gemi's
+        // own `update` raises `RecordNotFoundError` for the same call and takes
+        // this step down with it.
+        if (isLinked(valuesOf(parent, fkFields))) {
+          // `AND`, not a spread: a filter naming a referenced column itself
+          // must narrow rather than replace the restriction that keeps this on
+          // the linked row. `childLink` reads as "the row this one points at"
+          // on this side — the far row's own columns bound to the values just
+          // read off this one.
+          const where = conjoin(
+            record?.where,
+            childLink(link, parent as Record<string, unknown>),
+          );
+
+          const hit = (await executor.exec(
+            relation.model,
+            "findFirst",
+            { where, select: keySelect(referencedFields) },
+            // NOT pre-scoped: the child's own policies decide which rows exist,
+            // and a row this caller cannot see reads as absent — which sends it
+            // to the create branch, exactly as a hidden child does on the
+            // foreign side. The conservative answer, and the same one there.
+            false,
+          )) as Record<string, unknown> | null;
+
+          if (hit) {
+            // The key this row already holds, written back unchanged. The
+            // contribution is structural, so "the link does not move" has to be
+            // spelled as an assignment that changes nothing rather than as an
+            // absent one — the rule the `disconnect` filter arm states.
+            resolveKey(context, valuesOf(hit, referencedFields));
+
+            await executor.exec(
+              relation.model,
+              "updateMany",
+              { where, data: record?.update },
+              // NOT pre-scoped: the child's `onUpdate` and its scope-escape
+              // guard judge the payload, as they do for a nested `update`.
+              false,
+            );
+            return;
+          }
+        }
+
+        const created = (await executor.exec(
+          relation.model,
+          "create",
+          { data: record?.create, select: keySelect(referencedFields) },
+          // NOT pre-scoped — the child's own `onCreate` scopes the new row.
+          false,
+        )) as Record<string, unknown> | null;
+
+        resolveKey(context, valuesOf(created, referencedFields));
+      },
+    });
+
+    contributeResolved();
+    return;
   }
 
   if (key === "connectOrCreate") {
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -536,6 +536,14 @@ operand family and does *not*, while its `connect` one operand over does. That i
 Prisma disagreeing with itself, and matching it would mean special-casing one
 operand to reproduce an inconsistency.
 
+An owning-side `upsert` — `data: { organization: { upsert: … } }` — inherits it
+on the branch that **updates**. The foreign key is written back unchanged there,
+because the create branch needs that column in the statement and which branch
+runs is not known until the call runs, so the parent's stamp moves where
+Prisma's does not. The far row is updated identically on both. Worth searching
+for in the same places as the `disconnect` above: a row whose `updatedAt` you
+show, written through a relation rather than through a column.
+
 ---
 
 # Upgrading from 0.42 to 0.43
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3934,7 +3934,7 @@ await List.create({
 | `set` | Replaces the whole set — detaches what is linked now, attaches what you name. |
 | `updateMany` | Writes your columns to this parent's rows matching a filter. |
 | `deleteMany` | Deletes this parent's rows matching a filter — the filter goes directly under the key, not inside a `where`. |
-| `upsert` | Finds the row **among this parent's**, updates it, or creates it. A to-one has no `where` to require — the link already names the row. |
+| `upsert` | Finds the row **among this parent's**, updates it, or creates it. A to-one has no `where` to require — the link already names the row, from either end of the key. |
 
 **A relation may join on more than one field.** `@relation(fields: [tenantId, orderId], references:
 [tenantId, id])` is the tenant-scoped composite-key style, and every surface handles it — reads and
@@ -3947,8 +3947,8 @@ is one shape both dialects already compile.
 operand that side offers. Which operands those are is unchanged by the width of the key: the
 **foreign** side — where the children hold it — takes all of `connect`, `create`, `connectOrCreate`,
 `createMany`, `set`, `disconnect`, `update`, `upsert`, `delete` and the two `*Many`; the **owning**
-side takes `connect`, `create`, `connectOrCreate`, `disconnect` and `update`, and refuses
-`createMany`, `set`, `delete`, `upsert` and the two `*Many` by name, because this row holds one
+side takes `connect`, `create`, `connectOrCreate`, `disconnect`, `update` and `upsert`, and refuses
+`createMany`, `set`, `delete` and the two `*Many` by name, because this row holds one
 foreign key and there is nothing there to write many of. A composite relation is refused in exactly
 the same places a single-field one is, and nowhere else. Four details follow from the key having
 more than one column, and they are the ones worth knowing:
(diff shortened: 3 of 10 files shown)


packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+          { data: record?.create, select: keySelect(referencedFields) },
+          // NOT pre-scoped — the child's own `onCreate` scopes the new row.
+          false,
+        )) as Record<string, unknown> | null;
+
+        resolveKey(context, valuesOf(created, referencedFields));
+      },
+    });
+
+    contributeResolved();
```
nstfkc, line 1175: **`upsert` silently discards a sibling `connect`/`disconnect` on the same relation.**

`planOne` sorts the operand keys (~L350) and pushes one `before` step each; every owning-side operand writes the same `context.resolved[fkField]` slot and pushes an identical contribution, which `updateAssignments` folds through a field-keyed Map. `"connect"` < `"disconnect"` < `"upsert"`, so the upsert step runs last and its `resolveKey` overwrites whatever the earlier one resolved.

```js
User.update({ where: { id: 1 }, data: { organization: { disconnect: true, upsert: { create, update } } } })
```

edits the current organization and writes `organizationId` back unchanged — the user is **not** detached, and nothing is raised. Likewise `{ connect: { id: 5 }, upsert: {…} }` leaves the user in the old organization and edits that one instead of org 5.

Before this PR the same arguments raised `UnsupportedQueryError`, so the caller was told. Now the discarded operand is invisible.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+            // foreign side. The conservative answer, and the same one there.
+            false,
+          )) as Record<string, unknown> | null;
+
+          if (hit) {
+            // The key this row already holds, written back unchanged. The
+            // contribution is structural, so "the link does not move" has to be
+            // spelled as an assignment that changes nothing rather than as an
+            // absent one — the rule the `disconnect` filter arm states.
+            resolveKey(context, valuesOf(hit, referencedFields));
```
nstfkc, line 1149: **The parent FK is resolved from the read taken *before* `updateMany` runs, so an `update` that changes a referenced column binds a key that no longer exists.**

```js
LedgerNote.update({
  where: { id: 1 },
  data: { ledger: { upsert: { create: { tenantId: 9, code: "z", title: "c" }, update: { code: "renamed" } } } },
})
```

on note 1, linked to ledger `(1, "a")`: the step reads `hit = (1, "a")`, calls `resolveKey` with those values, then issues `updateMany` on `Ledger` setting `code = "renamed"`. The parent statement afterwards binds the **stale** contribution and writes `LedgerNote.tenantId = 1, code = "a"` — a ledger row that no longer exists. With FK enforcement on, that is a foreign-key constraint error on a call Prisma answers; with it off, the note is silently orphaned. Prisma issues no parent UPDATE at all here, so the DB's `ON UPDATE CASCADE` carries the new key through.

Same shape on the non-composite relations: `update: { id: 42 }` on `User.organization` writes `organizationId = 1` back over a row that is now id 42.

Secondary effect: the write-back also stamps `@updatedAt` on a parent row that did not change.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+        // records a key's presence, so a plan hit implies the same keys — but
+        // nothing in a step should be the first place a missing one is a
+        // `TypeError` rather than an error naming the argument.
+        const record = at(args) as Record<string, unknown> | null | undefined;
+
+        // Pre-scoped: `args.where` is the effective `where` this call already
+        // put through this model's policies, so re-applying them would `AND`
+        // the same predicate twice. One column of one row, inside the
+        // transaction the nested steps already run in.
+        const parent = (await executor.exec(
```
nstfkc, line 1107: **A policy `redact` on the (nullable) FK column makes a linked row read as unlinked, sending the operand to the destructive create branch.**

`executor.exec(schema.name, "findFirst", …, true)` skips `applyPolicies` but **not** `applyRedaction` — `Model.$exec` builds `policy` even when pre-scoped and calls `applyRedaction(policies, policy, result)` on the shaped row, and `redactNullable` explicitly permits nullable columns, which is exactly what an optional owning-side FK is.

With a policy that redacts `Note.folderId` for non-owning actors, `Note.update({ where: { id: 60 }, data: { folder: { upsert: { create, update } } } })` reads `folderId` back as `null`, `isLinked` answers false, and the create branch inserts a second Folder and repoints the note off the folder it was really on — the existing link destroyed, a duplicate row left behind. Prisma updates the linked folder in place.

The precedent cited for this read, the `disconnect` filter arm at L798, is immune because reading `null` there is a no-op (`if (!isLinked(linked)) return`); the `update` operand at L883 is immune because it reads the key from the statement's own `RETURNING` (`rows[0]`, passed to `runSteps` before redaction runs) rather than through a nested `$exec`. Copying the read carried the mechanism into a branch where the same misread is a silent relink.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+
+          const hit = (await executor.exec(
+            relation.model,
+            "findFirst",
+            { where, select: keySelect(referencedFields) },
+            // NOT pre-scoped: the child's own policies decide which rows exist,
+            // and a row this caller cannot see reads as absent — which sends it
+            // to the create branch, exactly as a hidden child does on the
+            // foreign side. The conservative answer, and the same one there.
+            false,
```
nstfkc, line 1141: **A linked child hidden by the child's policies reads as absent, so the parent is silently repointed away from the row it was linked to.**

The comment above defends `false` by pointing at the foreign side, but the two branches answer a hidden row very differently. Every other owning-side operand leaves the link alone: `disconnect`'s filter arm writes nothing, and the foreign side raises `UniqueConstraintError`. Here the answer is a write.

```js
Note.$exec("update", { where: { id: 60 }, data: { folder: { upsert: { create: { code: "made" }, update: { code: "x" } } } } })
```

on a note whose folder belongs to another tenant (or is soft-deleted, or is filtered by any read policy) mints a second folder and moves the note onto it — severing a link the caller never asked to change and orphaning a row they could not see. An application calling `upsert` in a loop against rows whose children are policy-filtered accumulates duplicate child rows and loses the original associations, with no error raised.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+              { where, data: record?.update },
+              // NOT pre-scoped: the child's `onUpdate` and its scope-escape
+              // guard judge the payload, as they do for a nested `update`.
+              false,
+            );
+            return;
+          }
+        }
+
+        const created = (await executor.exec(
```
nstfkc, line 1163: **Parent linked + operand `where` matching nothing falls through to create: the parent is repointed at a brand-new row and the `update` payload is discarded.**

```js
Profile.update({
  where: { id: 1 },
  data: { user: { upsert: { where: { name: "Nobody" }, create: { email: "created@example.dev" }, update: { name: "Updated" } } } },
})
```

with profile 1 linked to user 1: `isLinked` is true, `hit` is null because the filter misses, control falls out of the `if` block to here, a new `User` is minted and its id contributed. `Profile.userId` moves off user 1, user 1 is left holding no profile, and the `update` payload is never applied.

The operand's `where` is documented as narrowing *which connected row is written* — here a filter that misses converts a filtered update into an unconditional relink, i.e. the caller's guard causes the write it was meant to prevent. The PR pins this as a divergence (Prisma errors P2022 and rolls back, writing nothing), so a ported call that relied on the filter failing now silently moves the link.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+            // The key this row already holds, written back unchanged. The
+            // contribution is structural, so "the link does not move" has to be
+            // spelled as an assignment that changes nothing rather than as an
+            // absent one — the rule the `disconnect` filter arm states.
+            resolveKey(context, valuesOf(hit, referencedFields));
+
+            await executor.exec(
+              relation.model,
+              "updateMany",
+              { where, data: record?.update },
```
nstfkc, line 1154: **Routing `upsert.update` through the child's `updateMany` refuses nested writes with an error naming a model and operation the caller never wrote.**

```js
User.update({
  where: { id: 1 },
  data: { organization: { upsert: { create: { name: "n" }, update: { name: "n", users: { connect: { id: 2 } } } } } },
})
```

— a call Prisma answers — reaches `executor.exec("Organization", "updateMany", { where, data: { name, users } })`, whose `assertNoNestedWrites` throws `UnsupportedQueryError` with `argument = "data.users"`, `model = "Organization"`, `operation = "updateMany"`: *"updateMany cannot contain nested writes — it has no single row to attach them to. Use update per row."*

The caller wrote a `User.update` with an `upsert`, so the refusal points at a query they did not write — the exact misnamed-origin defect (#85/#101) that this branch's own docblock cites as the reason it needed a branch of its own.

Note the create branch does *not* have the limitation (`compileCreate` plans nested writes), so the two halves of one operand accept different payloads.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+              false,
+            );
+            return;
+          }
+        }
+
+        const created = (await executor.exec(
+          relation.model,
+          "create",
+          { data: record?.create, select: keySelect(referencedFields) },
```
nstfkc, line 1166: **`create: null` passes both plan-time asserts and reaches here as `data: null`, inserting a defaults-only row.**

`assertToOneWriteOperand` type-checks only `where`, and `assertUpsertOperand`'s loop (~L2976) tests `record[name] === undefined` — so `null` survives both.

```js
Profile.update({ where: { id: 2 }, data: { user: { upsert: { create: null, update: { name: "x" } } } } })
```

Profile 2 is unlinked, so this line calls `executor.exec("User", "create", { data: null })`; `insertColumns` explicitly admits `data === null` (`if (data !== undefined && data !== null)`), producing an insert of nothing but defaults. On a child whose columns are all nullable or defaulted this commits a blank row and repoints the parent at it; on one with a required column it surfaces as a raw NOT NULL violation from inside a nested step rather than an argument refusal. Prisma rejects `create: null` at validation with nothing written. Same for `update: null`.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+        // arrives from the write-back as P2025, rolling the insert back. gemi's
+        // own `update` raises `RecordNotFoundError` for the same call and takes
+        // this step down with it.
+        if (isLinked(valuesOf(parent, fkFields))) {
+          // `AND`, not a spread: a filter naming a referenced column itself
+          // must narrow rather than replace the restriction that keeps this on
+          // the linked row. `childLink` reads as "the row this one points at"
+          // on this side — the far row's own columns bound to the values just
+          // read off this one.
+          const where = conjoin(
```
nstfkc, line 1128: **An invalid `where` column is only validated when a row happens to be linked — validation that depends on the row's data.**

`conjoin(record?.where, …)` sits inside `if (isLinked(...))`; the create path at L1163 never touches `record.where`. `assertToOneWriteOperand` only checks that `where` is an object, not that its keys are columns of the child, and the foreign-side `upsert` compiles its `where` on every call so an unknown column always errors there.

On this side:

```js
User.update({ where: { id: 2 }, data: { organization: { upsert: { where: { naem: "Acme" }, create: {…}, update: {…} } } } })
```

against a user with **no** organization succeeds and creates the organization with the typo silently dropped, while the identical call against a user who **has** one raises an unknown-column error from `Organization.findFirst`. Prisma refuses the unknown argument in both cases.

The result is a guarded upsert that runs unguarded on exactly the rows where the guard was cheapest to get wrong, and a validation failure that only appears once production data changes shape.

packages/gemi/orm/compile/nested-writes.ts
```diff
@@ -1007,17 +996,184 @@ function planOwningSide(
...
+            // absent one — the rule the `disconnect` filter arm states.
+            resolveKey(context, valuesOf(hit, referencedFields));
+
+            await executor.exec(
+              relation.model,
+              "updateMany",
+              { where, data: record?.update },
+              // NOT pre-scoped: the child's `onUpdate` and its scope-escape
+              // guard judge the payload, as they do for a nested `update`.
+              false,
```
nstfkc, line 1157: **(PLAUSIBLE — data/policy dependent.) Reading the child as `findFirst` but writing it as `updateMany` makes the whole operand a silent no-op when a policy scopes reads and mutations differently.**

The hit lookup at L1133 runs under the child's read scope; the write at L1151 runs under its `updateMany` scope (`MUTATING` in `policy.ts:325` gives `context.operation === "updateMany"`). A policy that scopes reads to the tenant but mutations to rows the actor owns lets the `findFirst` hit and the `updateMany` match zero rows. `updateMany` returns `{ count: 0 }` rather than raising, the step returns, and the parent statement commits having written its FK back unchanged.

The caller gets a successful `update` whose nested payload was never applied — no error, no count surfaced — where Prisma either applies the update or raises P2025. The create branch cannot cover for it either, since the read hit already sent the call down the update branch.

packages/gemi/orm/compile/nested-writes.ts
nstfkc, line 1107: _cleanup_ — **Third verbatim copy of the pre-scoped "read this row's own foreign key" block.**

`executor.exec(schema.name, "findFirst", { where: args?.where, select: keySelect(fkFields) }, true)` now exists identically at L817 (`disconnect`'s filter arm), here, and L3349 (`displaceSibling`) — each with its own paragraph re-arguing why `preScoped` is `true`.

Any change to how the owning row's key is read — the pre-scoping flag, a `where` that can match more than one row, adding a `take: 1`, or a different failure answer when the parent is missing — has to be found and applied in three places, and missing one gives a single operand a different scoping rule from its neighbours. That is exactly the divergence this file's own `childLink`/`keySelect`/`isLinked` docblock says the shared shapes exist to prevent (*"there is no single-field spelling left for a new operand to copy"*). Worth extracting one closure.

nstfkc (commented):
Automated code review (high effort). 10 findings, each independently verified; one further candidate was refuted and dropped.

Nearly everything lands in the new owning-side `upsert` branch (~L1085-1180), which replaces a hard refusal with a path that writes. The dominant theme: **three separate conditions fall through to the create branch and silently repoint the parent off a row it was linked to** — a policy-redacted nullable FK, a policy-hidden child, and a `where` that matches nothing. Each mints a duplicate far row and orphans the original, with no error raised. Two more are structural: the branch shares `context.resolved` with sibling operands and sorts last, so a `connect`/`disconnect` on the same relation is swallowed; and the hit branch resolves the parent FK from a *pre-update* read, so an `update` touching a referenced column binds a stale key.

One finding is marked PLAUSIBLE rather than CONFIRMED — flagged inline.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31 — refused at plan time, and for the whole family rather than just the pair `upsert` created.

`connect`, `connectOrCreate`, `create`, `disconnect` and `upsert` are now `RESOLVES_THE_KEY`, and `planOne` raises when two of them appear on one owning-side relation, naming both. You are right that `upsert` is what made it worth doing — those arguments raised until this PR, so the caller went from told to unaware — but `{ connect, disconnect }` and either beside `create` had the identical defect and were always accepted, so they go with it rather than being left as the quiet case.

`disconnect: false` is exempt. Its branch returns before it contributes, which is a value-at-plan-time test `toOneOperand` already makes, so it can neither swallow nor be swallowed — `{ disconnect: cond, upsert }` keeps working when `cond` is false.

Pinned in `refusals.test.ts` for both pairs, and `UPGRADE.md` gains a section, since Prisma applies two operands in sequence and a ported call spelling both is a behaviour difference rather than only a refusal.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31, and the fix is not the one I first reached for.

Re-reading the child by its primary key after the write does not generalise: on a composite link the referenced columns are routinely the child's `@@id` as well — `Ledger` is exactly that — so once `code` moves there is no key left to find the row by. The new value comes off the payload instead, which is also what the `ON UPDATE CASCADE` computes, and the cascade is Prisma's entire answer here since it issues no parent `UPDATE` at all.

That restricts what may be written to a referenced column to something that *names* the resulting value: a literal, or `{ set }`. `{ increment: 1 }` and its siblings leave the result to the database, so they are refused at plan time rather than guessed at.

Read at bind time, not plan time — `canonicalShape` records `42` and `{ increment: 1 }` as different shapes, which is what makes the compile-time check sound, but not `42` from `43`. `M17d`, `M17e` and a composite rename pin the behaviour differentially; a fourth case runs the same shape twice without a reset, so a step closing over the plan-time value fails it.

The secondary effect you note stays: the write-back stamps `@updatedAt` on the parent. Same mechanism as the `disconnect` filter arm, documented in `UPGRADE.md`.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Confirmed and fixed in ff7fa31 — and thank you for tracing why the two precedents are immune, because that is what made the shape of the fix obvious.

The read is machinery: one column of one row, used to build a `where` and a contribution whose value was already in that column. `RelationExecutor.exec` gains an `unredacted` flag, implemented in `Model.ts` as `runAsSystem` because that is the only switch `$exec` has for the row transform. It is only ever passed alongside `preScoped`, where the scope is already in `args` and there is no include tree, so the suspension removes the redaction and nothing else.

This and your last comment landed together: all three copies now go through one `readOwnKey`, so `disconnect` and `displaceSibling` stop misreading a redacted key too — a silent no-op in both, where here it was the destructive branch. Pinned by a policy case with a `redact` over `Note.folderId`, which fails with the flag reverted.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31, together with the `where`-misses finding below it — one condition, one answer.

A linked parent no longer reaches the create branch at all. `RecordNotFoundError`, which is what the `update` operand two branches up already gives for its own filtered miss, and what this file's vocabulary means by "the row this points at was not found". The create branch is for a parent that points at nothing.

The rule stated plainly in the docblock: no owning-side operand repoints a link the caller did not name a new target for, and reaching create from a linked parent is exactly that repoint — silently, since the row it moves off is by construction one the lookup did not return.

The policy case that pinned the old behaviour now asserts the refusal, and that nothing was minted and the note still points where it did. `docs/orm.md` says so from the caller's end.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31 — same fix as the hidden-child finding, since it is the same condition: linked parent, lookup returns nothing.

Your framing is the one that went into the docblock — a filter written to *guard* the write causing a bigger one. It now raises `RecordNotFoundError` and writes nothing.

The divergence pin moves rather than disappearing. Both clients now write nothing on this call; what differs is the failure's class — `notFound` here against Prisma's `other` for its own invalid SQL — so the case stays in "where the two clients still disagree" with gemi's class asserted and Prisma's only asserted as a failure, for the dialect reason the original note gives.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31 — refused at plan time, as `data.<relation>.upsert.update.<key>` on the model and operation the caller wrote.

The message states the asymmetry rather than hiding it: the update branch reaches the far row through a filter rather than a unique key, so there is no single row to attach a nested write to, while the `create` half goes through the child's `create` and does plan them. It points at writing the far row through its own `update` instead.

Not served, because serving it means naming the far row by a unique key, and the operand's `where` is a `WhereInput` by design on this side. Refusing at plan time is at least the same failure the caller would get from Prisma's validator in spirit, with an argument path that exists in their query.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31. `assertUpsertOperand` asked whether the key was present and stopped there; it now also requires an object, for `create` and `update` on both sides of a relation since the guard is shared. `where` keeps its own checks — `matchUniqueKey` on a to-many, `assertToOneWriteOperand` on a to-one, where it is optional and a filter rather than a key.

`InvalidArgumentError` naming `data.<relation>.upsert.create`, which classifies the way Prisma's validation refusal does. Both spellings pinned in `refusals.test.ts`.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31 — checked at plan time, so it no longer depends on whether a row happens to be linked.

`UnknownFieldError` with the child's own field list, deliberately: it is what `compileWhere` raises one layer down, so the two paths are now indistinguishable rather than merely both-failing. The walk mirrors that function's dispatch — combinator, relation, field, or Prisma's `a_b` spelling of a composite `@@unique` — and descends into `AND`/`OR`/`NOT` groups but stops at the top level of each, since everything below is an operator the field's own filter owns.

It covers the `update` operand too, which conjoins its `where` on the same branch and had the same data-dependence. Three cases in `refusals.test.ts`, plus one asserting a real column, a relation filter and a composite key still compile.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Fixed in ff7fa31, marked PLAUSIBLE or not — it costs one comparison and turns a silent no-op into an error.

`updateMany`'s `{ count: 0 }` after a read that hit now raises `RecordNotFoundError`. Only where the payload names a column: `update: {}` asks for no write, so a zero count there is the answer rather than a mismatch.

Pinned with a `Folder` policy whose `scope` narrows on `context.operation === "updateMany"`, which is the shape you describe. Without the check the call resolves and the folder is untouched.

packages/gemi/orm/compile/nested-writes.ts
nstfkc: Done in ff7fa31 — `readOwnKey`, and it turned out to be load-bearing rather than tidying.

The three copies did not agree, and one of them was wrong: the flags were `(preScoped: true)` everywhere, and `preScoped` skips `applyPolicies` but not `applyRedaction`. That is the redacted-foreign-key finding above, and extracting the closure is what fixed it in all three places at once — `disconnect` and `displaceSibling` had the same misread, benign in both because an absent link means nothing to do.

One function, one paragraph arguing both flags, and the three sites now point at it instead of restating it.

PR #396, opened 2026-08-12
Add `gemi/testing`: a `<Page>` that mounts a view with its inputs

### LINKED ISSUE #395: No test harness for views: route params, translations and `useQuery` data can't be seeded, so component tests can only assert the empty state
Mounting a view in a unit test works — but every input a view reads is unseedable from the public API, so the only state a component test can assert is the empty/loading one.

## What already works

Worth stating first, because the gap is narrower than "views don't mount". Rendering a component that uses `gemi/client` needs no providers at all — the contexts have defaults and nothing throws:

| | result |
|---|---|
| `useParams()` | renders → `{}` |
| `<Link href="/app/abc/chat">` | renders → `<a data-active="false" data-pending="false" href="/app/abc/chat">go</a>` ✅ |
| `useTranslator()` | renders → the key back, `Test:greeting` |
| `useQuery("/api/health")` | renders → full shape (`data`, `loading`, `error`, `mutate`, `trigger`, `prefetch`, `refetch`, …), `data` empty |

`Link` is genuinely correct. The other three render a shape with nothing in it.

## The gap: no way to seed the inputs

**Route params.** `useParams()` returns `{}` and there is no public provider to change that. `ClientRouterProvider` is not exported from `gemi/client` — the only provider on the public entry is `QueryManagerProvider`. Even reaching past the entry point, `ClientRouterProviderProps` wants 13 props, three of which are build-time artifacts (`routeManifest`, `cssManifest`, `modulePreloadManifest`). A test cannot reasonably construct that, so any view reading `:orgId` is untestable above the empty case.

**Translations.** `useTranslator()` returns the key and logs:

```
Unresolved translation Component:undefined key:Test:greeting
```

Note `Component:undefined` — outside a view tree the translator does not know which dictionary it belongs to. There is no exported way to mount a component with a dictionary, so a test can never assert rendered copy, only key strings.

**Query data.** `useQuery` is the important one. `QueryManagerContextValue` has exactly the hook a test needs:

```ts
hydrate: (prefetchedData?: PrefetchedData | null) => void;
```

…but `QueryManagerProvider` accepts only `{ queryConfig }`. `prefetchedData` is reachable on the context value and not on the provider, so there is no supported way to render a view with query data already present. Tests are stuck asserting the loading/suspended branch of every data-driven view.

## Why this matters

The runtime side is in good shape — under `bun test` a component suite needs no framework-specific config at all (React 19 lifecycles, `@testing-library/user-event`, and MSW fetch interception all work against `gemi/client` components with just a happy-dom preload). The blocker is not the runner, it is that the framework offers no seam for supplying a view's inputs.

## Proposal

A `gemi/testing` entry with one helper that builds the provider tree:

```ts
import { renderView } from "gemi/testing";

const { container } = renderView(<OrgChat />, {
  params: { orgId: "abc" },
  searchParams: "?tab=recent",
  pathname: "/app/abc/chat",
  locale: "en-US",
  dictionaries: [ChatDictionary],
  queryData: { "/api/organizations/abc/messages": { data: [...] } },
  user: { id: 1, globalRole: 0 },
});
```

Renderer-agnostic (return the element tree, or expose a `<TestProviders>` wrapper) so it composes with Testing Library rather than depending on it. The minimum useful subset is `params` + `queryData` + `dictionaries`; the router internals can stay private behind it.

Two smaller alternatives, if a full harness is too much scope:

1. Export a `prefetchedData` prop on `QueryManagerProvider` (the context method already exists — this is a prop passthrough).
2. Export a params/locale provider that does not require the three build manifests.

Either one unblocks most component tests on its own.

## Environment

gemi 0.52.0, React 19.2.8, `bun test` 1.3.14. Reproduced by rendering the components above with `renderToString` and no providers.

Sibling of #348 — same class of gap on the server side (`AuthController` has no harness, so `onUserCreated` ordering is prose-only). Between them, gemi's contracts are documented but not assertable from an app.

nstfkc:
Closes #395.

Mounting a view in a unit test already worked — the contexts have defaults and nothing throws — but every input a view reads was unseedable from the public API, so the only state a component test could assert was the empty one. `useParams()` returned `{}` with no exported provider to change it; `useTranslator()` returned the key back because outside a view tree it does not know which dictionary it belongs to; `QueryManagerContext` carried the `hydrate` a test needs while `QueryManagerProvider` accepted only `queryConfig`.

## The API

`gemi/testing` exports one component. It assembles the tree `ClientRouter` builds at runtime and seeds it:

```tsx
import { render, screen } from "@testing-library/react";
import { Page } from "gemi/testing";

render(
  <Page
    pathname="/app/:orgId/chat"          // a template, resolved against params
    params={{ orgId: "abc" }}
    searchParams="?tab=recent"           // or { tab: "recent" }
    locale="tr-TR" defaultLocale="en-US" // links then carry the /tr-TR prefix
    dictionaries={[dictionaries.Chat]}   // the app's real dictionaries
    queryData={{ "/orgs/:orgId/messages": [{ id: 1, body: "hi" }] }}
    user={{ id: 1 }}
  >
    <OrgChat />
  </Page>,
);

expect(screen.getByText("hi")).toBeDefined();
```

Also seedable: `breadcrumbs`, `queryConfig`, `hash`, `supportedLocales`, a `fallback` (the `Suspense` boundary standing in for a view's `Loading` export), an `errorFallback` (its `Error` export) and `onNavigate`. It is a plain component with no runner dependency, so it composes with Testing Library, `renderToString`, or anything else that renders React.

## Three choices worth reviewing

- **`queryData` keys are `useQuery` paths, without `/api`.** The prefix is added when the client fetches, so keying by it (as the issue's example does) seeds a path no query reads — silent, and looks like the harness not working. A key starting with `/api/` warns. `:params` resolve against the page's params, and a `?query` targets one search variant, matching how the cache is keyed.
- **Seeded at mount only.** Re-rendering with a new `queryData` would clobber whatever the component's own `mutate()`/`refetch()` had put in the cache, so it deliberately does not re-hydrate. To assert a second state, render a second `<Page>`.
- **No routing.** There is no view tree or route manifest behind it, so a navigation is reported through `onNavigate` rather than resolved; `useLocation()` keeps reporting the pathname it was given.

`useUser()` resolves from `user` (default `null`, an anonymous visitor) without reaching `/auth/me`, so a layout that greets the current user does not have to be mocked out of a test that is about something else.

## Build

`client` and `testing` are now two entries in **one** vite pass. Built separately, each bundle would carry its own copy of every `createContext` result — two distinct context objects — and a seeded `<Page>` would provide values the app's `gemi/client` components never read. The failure would be silent seeding, in the published package only. Verified the emitted bundles share one chunk.

## Two adjacent fixes, both load-bearing for the above

- **`Dictionary` imported the `../facades` barrel**, which drags in `Redis` and its `import { RedisClient } from "bun"`. A dictionary is the one server-side artifact a component test has to import, and under a browser-targeted runner that specifier does not resolve — so importing the app's own `app/i18n` failed to collect before rendering anything. Narrowed to `../facades/Lang`.
- **`useIsNavigationPending` handed unbound `Subject` methods to `useSyncExternalStore`**, which calls them as plain functions: `this.value` threw on the first render of any component calling the hook, in the browser as much as in a test. `useFormData` already binds its own; this now does too.

## Verification

- 18 new tests in `packages/gemi/testing/Page.test.tsx`, asserting through the hooks an application actually calls (`useParams`, `useSearchParams`, `useTranslator`, `useQuery`, `useUser`, `useBreadcrumbs`, `Link`, `Form`), including that a seeded query issues no request and an unseeded one still suspends and fetches.
- Full package suite green (3371 passed, 1 skipped), `typecheck` clean, `oxlint` unchanged at 79 warnings / 0 errors.
- `build:client` and `build:client-types` emit `dist/testing/index.{js,d.ts}`; the publish-shape check in `build-publish.ts` covers the new export.
- Checked end-to-end against a real app: `templates/saas-starter`'s `About.tsx`, imported through the package's `exports` map, rendered its seeded query variant and its real dictionary (`t.jsx` interpolation included) with zero fetches. Done on a scratch branch and reverted — landing it would mean adding `jsdom` and `@testing-library/react` to the scaffold every new app is generated from.

Docs: `docs/testing.md`, indexed in `README.md`, `llms.txt`, `llms-full.txt` and `index.html`.

## Not done, deliberately

- **An example test in the template.** It would also give CI a consumer-side check of `gemi/testing`, which nothing covers today — the "Template tests" step runs `app/models` only. Costs two devDependencies in the scaffold.
- **Simulating navigation** so `useLocation()` updates after a `Link` click. Without a route manifest the new route's params cannot be re-derived, and carrying stale ones forward would be a lie.

### DIFF
--- a/packages/gemi/client/useIsNavigationPending.ts
+++ b/packages/gemi/client/useIsNavigationPending.ts
@@ -3,10 +3,15 @@ import { ClientRouterContext } from "./ClientRouterContext";
 
 export function useIsNavigationPending() {
   const { isNavigatingSubject } = useContext(ClientRouterContext);
+  // Bound, like `useFormData` binds its own subject's: `Subject`'s methods
+  // read `this`, and `useSyncExternalStore` calls whatever it is handed as a
+  // plain function — so the unbound references this used to pass threw
+  // `undefined is not an object (evaluating 'this.value')` on the first render
+  // of any component that called this hook.
   const isNavigating = useSyncExternalStore(
-    isNavigatingSubject.subscribe,
-    isNavigatingSubject.getValue,
-    isNavigatingSubject.getValue,
+    isNavigatingSubject.subscribe.bind(isNavigatingSubject),
+    isNavigatingSubject.getValue.bind(isNavigatingSubject),
+    isNavigatingSubject.getValue.bind(isNavigatingSubject),
   );
 
   return isNavigating;
--- a/packages/gemi/client/useIsNavigationPending.ts
+++ b/packages/gemi/client/useIsNavigationPending.ts
@@ -3,15 +3,16 @@ import { ClientRouterContext } from "./ClientRouterContext";
 
 export function useIsNavigationPending() {
   const { isNavigatingSubject } = useContext(ClientRouterContext);
-  // Bound, like `useFormData` binds its own subject's: `Subject`'s methods
-  // read `this`, and `useSyncExternalStore` calls whatever it is handed as a
-  // plain function — so the unbound references this used to pass threw
-  // `undefined is not an object (evaluating 'this.value')` on the first render
-  // of any component that called this hook.
+  // These read `this`, and `useSyncExternalStore` calls whatever it is handed
+  // as a plain function — which used to throw `undefined is not an object
+  // (evaluating 'this.value')` on the first render of any component calling
+  // this hook. `Subject` now binds in its constructor, so the references are
+  // both correct and stable across renders; binding here instead would hand
+  // uSES a new `subscribe` every render and churn the subscription.
   const isNavigating = useSyncExternalStore(
-    isNavigatingSubject.subscribe.bind(isNavigatingSubject),
-    isNavigatingSubject.getValue.bind(isNavigatingSubject),
-    isNavigatingSubject.getValue.bind(isNavigatingSubject),
+    isNavigatingSubject.subscribe,
+    isNavigatingSubject.getValue,
+    isNavigatingSubject.getValue,
   );
 
   return isNavigating;
--- a/packages/gemi/testing/Page.tsx
+++ b/packages/gemi/testing/Page.tsx
@@ -1,6 +1,7 @@
 import {
   Suspense,
-  useEffect,
+  useContext,
+  useMemo,
   useRef,
   useState,
   type ComponentType,
@@ -15,6 +16,7 @@ import { ClientRouterContext } from "../client/ClientRouterContext";
 import { I18nProvider } from "../client/I18nContext";
 import { ProgressManager } from "../client/ProgressManager";
 import {
+  QueryManagerContext,
   QueryManagerProvider,
   type QueryConfig,
 } from "../client/QueryManagerContext";
@@ -28,10 +30,12 @@ import {
   ServerDataContext,
   type ServerDataContextValue,
 } from "../client/ServerDataProvider";
+import { ThemeProvider } from "../client/ThemeProvider";
 import type { Breadcrumb } from "../client/useBreadcrumbs";
 import { WebSocketContext } from "../client/WebsocketContext";
 import { applyParams } from "../utils/applyParams";
 import { Subject } from "../utils/Subject";
+import { toVariantKey } from "../utils/variantKey";
 
 /**
  * A dictionary as `Dictionary.create` produces it: a name and a
@@ -100,8 +104,16 @@ export interface PageProps {
   /**
    * Data `useQuery` finds already cached, keyed by API path — the path passed
    * to `useQuery`, without the `/api` prefix the client adds when it fetches.
-   * Keys may carry a query string (`"/lists?page=2"`) to seed one search
-   * variant, and may use `:params`, which resolve against `params`.
+   * A key may carry a query string (`"/lists?page=2"`) to seed one search
+   * variant.
+   *
+   * The cache is keyed by the *resolved* path, so a key's `:params` are
+   * resolved against the page's `params` — and a key naming one the page does
+   * not carry is an error, naming the key. That case is common enough to be
+   * worth the loud failure: a query often takes its params from the call site
+   * rather than the route (`useQuery("/orgs/:orgId/lists", { params: { orgId } })`
+   * with `orgId` in component state), and the right seed for it is the resolved
+   * path, `"/orgs/abc/lists"`.
    *
    * Seeded at mount only: a component that mutates or refetches its data owns
    * the cache from then on.
@@ -116,6 +128,12 @@ export interface PageProps {
   user?: Partial<User> | null;
   /** What `useBreadcrumbs()` returns, in order. */
   breadcrumbs?: Breadcrumb[];
+  /**
+   * The theme `useTheme()` starts on. The app reads a visitor's stored choice
+   * out of `localStorage`; a test has no session to have made one in, so this
+   * seeds it without writing to storage. `setTheme` works from either.
+   */
+  theme?: "light" | "dark" | "system";
   /**
    * Fallback for the `Suspense` boundary wrapped around the children — the
    * stand-in for the view's own `Loading` export, which the real router
@@ -150,11 +168,61 @@ function toSearchString(search: PageProps["searchParams"]): string {
   return query ? `?${query}` : "";
 }
 
-/** The cache's variant key: sorted search params, empty when there are none. */
-function toVariantKey(search: string): string {
-  const searchParams = new URLSearchParams(search);
-  searchParams.sort();
-  return searchParams.toString();
+/**
+ * The `:params` a seed key needs supplied. Wildcards (`:rest*`) are excluded —
+ * `applyParams` drops those when missing, which is a resolution rather than a
+ * failure.
+ *
+ * Optional params (`:id?`) cannot reach here at all: a key is split at its
+ * first `?` to separate the search string, so `"/lists/:folderId?"` arrives as
+ * the path `/lists/:folderId`. Seed the resolved path for those.
+ */
+function requiredParamsOf(path: string): string[] {
+  return Array.from(path.matchAll(/:([^/]+)/g), ([, name]) => name).filter(
+    (name) => !name.endsWith("*"),
+  );
+}
+
+/**
+ * Fails a `queryData` key whose `:params` the page does not carry, saying so
+ * in terms of the key.
+ *
+ * Without this the two documented runners disagree, and neither is legible.
+ * `applyParams` throws `Missing parameter: orgId` under `import.meta.env.DEV`
+ * (vitest) — a router-utility stack trace out of `<Page>`'s render, before
+ * anything mounts, naming nothing that appears in the test. Under `bun test`
+ * that flag is falsy, so it logs and seeds the literal `/orgs/undefined/lists`
+ * instead: the seed misses, the query goes to the network, and the assertion
+ * runs against a loading state.
+ *
+ * The case is common because a query's params often come from the call site
+ * rather than the route — `useQuery("/orgs/:orgId/lists", { params: { orgId } })`
+ * with `orgId` in component state. The cache is keyed by the *resolved* path,
+ * so the way out is to seed that path directly.
+ */
+function resolveSeedPath(
+  key: string,
+  rawPath: string,
+  params: Record<string, string>,
+): string {
+  const missing = requiredParamsOf(rawPath).filter(
+    (name) => params[name] === undefined,
+  );
+  if (missing.length > 0) {
+    throw new Error(
+      `[gemi] <Page> cannot resolve the queryData key "${key}": no value for ` +
+        `${missing.map((name) => `:${name}`).join(", ")}. The query cache is ` +
+        `keyed by the resolved path, so either add ${missing
+          .map((name) => `\`${name}\``)
+          .join(", ")} to the page's \`params\`, or seed the resolved path ` +
+        `(e.g. "${applyParams(
+          rawPath,
+          Object.fromEntries(missing.map((name) => [name, "123"])),
+        )}") — which is what a query taking its params from component state ` +
+        `reads under.`,
+    );
+  }
+  return applyParams(rawPath, params);
 }
 
 /**
@@ -164,18 +232,22 @@ function toVariantKey(search: string): string {
 function toPrefetchedData(
   queryData: Record<string, unknown>,
   params: Record<string, string>,
+  warned: Set<string>,
 ) {
   const prefetchedData: Record<string, Record<string, unknown>> = {};
   for (const [key, data] of Object.entries(queryData)) {
     const [rawPath, rawSearch = ""] = key.split("?");
-    if (rawPath.startsWith("/api/")) {
+    // Once per key per page, not once per render: this runs again on every
+    // re-render, and a duplicated warning reads as a second mistake.
+    if (rawPath.startsWith("/api/") && !warned.has(key)) {
+      warned.add(key);
       console.warn(
         `[gemi] queryData key "${key}" starts with /api. useQuery is keyed by ` +
           `the path you pass it — the /api prefix is added when it fetches — ` +
           `so this entry seeds a path no query reads.`,
       );
     }
-    const path = applyParams(rawPath, params);
+    const path = resolveSeedPath(key, rawPath, params);
     prefetchedData[path] = {
       ...prefetchedData[path],
       [toVariantKey(rawSearch)]: data,
@@ -248,11 +320,22 @@ function toClientDictionary(
  * ```
  *
  * It is a plain component, so it composes with any renderer — Testing
- * Library's `render`, `renderToString`, `react-test-renderer`.
+ * Library's `render`, `react-test-renderer`, or `renderToString`.
  *
- * What it does not do is route. There is no view tree and no route manifest
- * behind it, so a navigation is reported through `onNavigate` rather than
- * resolved; to assert the page *after* a navigation, render a second `<Page>`.
+ * Two things it does not do.
+ *
+ * It does not route: there is no view tree and no route manifest behind it, so
+ * a navigation is reported through `onNavigate` rather than resolved. To
+ * assert the page *after* one, render a second `<Page>`.
+ *
+ * And under `renderToString` it renders the server's own no-data branch for an
+ * *unseeded* query rather than suspending — deliberately, because suspending
+ * there is the streaming server's job: `createRoot` threads the request's
+ * `ServerQueryStore` through `ServerQueryContext`, and that store is what runs
+ * the route handler. A harness has no request and no handler to run, so there
+ * is nothing to suspend on. Seeded queries render their data on the server
+ * exactly as they do in the browser; unseeded ones behave as a route with no
+ * `Query.prefetch` does, framework warning included.
  */
 export const Page = (props: PropsWithChildren<PageProps>) => {
   const {
@@ -269,41 +352,61 @@ export const Page = (props: PropsWithChildren<PageProps>) => {
     queryConfig,
     user = null,
     breadcrumbs = [],
+    theme,
     fallback = null,
     errorFallback,
     onNavigate,
   } = props;
 
-  const supportedLocales = Array.from(
-    new Set([defaultLocale, locale, ...(props.supportedLocales ?? [])]),
+  // The caller's list, in the caller's order — a language switcher maps over
+  // it, so hoisting the page's own locale to the front would reorder the
+  // rendered options. The defaults are appended, only to guarantee presence.
+  const supportedLocales = useMemo(
+    () =>
+      Array.from(
+        new Set([...(props.supportedLocales ?? []), defaultLocale, locale]),
+      ),
+    [props.supportedLocales, defaultLocale, locale],
   );
   const pathname = applyParams(routePath, params) || "/";
   const search = toSearchString(searchParams);
   // The URL's locale segment, which the default locale does not get.
   const urlLocaleSegment = locale === defaultLocale ? null : locale;
 
-  const dictionary = toClientDictionary(
-    dictionaries,
-    supportedLocales,
-    locale,
-    translations,
+  const dictionary = useMemo(
+    () =>
+      toClientDictionary(dictionaries, supportedLocales, locale, translations),
+    [dictionaries, supportedLocales, locale, translations],
   );
-  const prefetchedData = toPrefetchedData(queryData, params);
-  // `useUser` reads `/auth/me` like any other query. Seeding it — with `null`
-  // for the anonymous default — is what keeps a component test off the network
-  // for a hook nothing in the test asked for.
-  prefetchedData["/auth/me"] ??= { "": user ?? null };
+
+  // Keys that have already drawn a warning, so a re-render does not repeat it.
+  const warnedRef = useRef<Set<string>>(new Set());
+  const prefetchedData = useMemo(() => {
+    const seeded = toPrefetchedData(queryData, params, warnedRef.current);
+    // `useUser` reads `/auth/me` like any other query. Seeding it — with `null`
+    // for the anonymous default — is what keeps a component test off the
+    // network for a hook nothing in the test asked for.
+    seeded["/auth/me"] ??= { "": user ?? null };
+    return seeded;
+  }, [queryData, params, user]);
 
   // Keyed by `${view}:${pathname}` in the router's cache; the view names are
   // ours to choose here, and only their order reaches `useBreadcrumbs`.
-  const breadcrumbViews = breadcrumbs.map((_, index) => `breadcrumb-${index}`);
-  const breadcrumbsCache = new Map<string, Breadcrumb>(
-    breadcrumbs.map((breadcrumb, index) => [
-      `${breadcrumbViews[index]}:${pathname}`,
-      breadcrumb,
-    ]),
-  );
-  const breadcrumbsRecord = Object.fromEntries(breadcrumbsCache);
+  const { breadcrumbViews, breadcrumbsCache, breadcrumbsRecord } =
+    useMemo(() => {
+      const views = breadcrumbs.map((_, index) => `breadcrumb-${index}`);
+      const cache = new Map<string, Breadcrumb>(
+        breadcrumbs.map((breadcrumb, index) => [
+          `${views[index]}:${pathname}`,
+          breadcrumb,
+        ]),
+      );
+      return {
+        breadcrumbViews: views,
+        breadcrumbsCache: cache,
+        breadcrumbsRecord: Object.fromEntries(cache),
+      };
+    }, [breadcrumbs, pathname]);
 
   const [isNavigatingSubject] = useState(() => new Subject<boolean>(false));
   const [progressManager] = useState(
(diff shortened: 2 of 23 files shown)


nstfkc:
Follow-up on a review question — *dictionaries are server-side, so what happens in a client test?*

Two halves, and only one of them was a real gap.

**The data was never the problem.** The browser never sees a `Dictionary`: the server resolves the route's dictionaries for the active locale and serializes them onto the page, and `useTranslator` reads that payload. `<Page>` mirrors the payload, not the class — it reads `name` and `dictionary` as plain data, its prop type is structural (`{ name, dictionary }`, so a literal is accepted), and `gemi/testing` does not import `gemi/i18n` at all. The genuinely server-only methods, `render` and `reference`, throw once `window` is defined; nothing here calls them, and a test that calls one gets the same error a browser would.

**The import was.** `app/i18n/index.ts` imports `gemi/i18n`, so a test file reading the app's dictionaries pulls the container and `node:async_hooks` into its module graph. Nothing there executes, so it loads under any runner with Node builtins — vitest, `bun test`, which is what the smoke test in the PR description exercised — and not under one that renders in a real browser (vitest browser mode, Playwright component tests).

So `5c5a1de` adds the payload shape as a first-class prop:

```tsx
<Page translations={{ About: { title: "About {{version:[hi]}}" } }}>
  <About />
</Page>
```

`{ dictionary name: { key: string } }` for the current locale — exactly what the server puts on the page, importing nothing. Merged **over** `dictionaries`, so the two compose: seed from the app's real dictionary and override the one key a test is about.

`dictionaries` stays, because using the app's real source of truth is what most tests want and it works on both supported runners. The docs now say plainly which one to reach for and why. Two tests added (20 in the file, 3373 in the suite); typecheck and lint unchanged.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+  for (const [key, data] of Object.entries(queryData)) {
+    const [rawPath, rawSearch = ""] = key.split("?");
+    if (rawPath.startsWith("/api/")) {
+      console.warn(
+        `[gemi] queryData key "${key}" starts with /api. useQuery is keyed by ` +
+          `the path you pass it — the /api prefix is added when it fetches — ` +
+          `so this entry seeds a path no query reads.`,
+      );
+    }
+    const path = applyParams(rawPath, params);
```
nstfkc: **`queryData` `:params` resolve against the page's params only, and a miss hard-throws.**

`toPrefetchedData` runs every key through `applyParams(rawPath, params)` with only the `<Page params>` prop. But a `useQuery` path's params often come from the call site, not the route — `useQuery("/orgs/:orgId/lists", { params: { orgId } })` where `orgId` is component state.

Seed that key without a matching `params` entry (which the docs' "a key may carry `:params`" invites) and `applyParams` hits its `import.meta.env.DEV` branch — true under vitest — and throws `Missing parameter: orgId` out of `<Page>`'s render, before anything mounts. The failure is a router-utility stack trace with no hint that the seed key was the problem.

Under `bun test` (the other documented runner) `import.meta.env.DEV` is falsy, so it logs and seeds the literal `/orgs/undefined/lists` instead: the query misses the cache, the test hits the network, and asserts against a loading state. Two different wrong behaviours for the same mistake.

Worth either catching the throw and re-reporting it as a `queryData`-key error, or letting a seed key opt out of param resolution.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+    breadcrumbs: breadcrumbsRecord,
+    appId: "test",
+  };
+
+  const [routerSubject] = useState(() => new Subject<RouteState>(routeState));
+
+  const onNavigateRef = useRef(onNavigate);
+  onNavigateRef.current = onNavigate;
+  useEffect(() => {
+    return history.listen(({ location, action }) => {
```
nstfkc: **`onNavigate` misses any navigation issued from a child's mount effect.**

The `history.listen` registration lives in `<Page>`'s own `useEffect`, and React flushes child effects *before* parent effects. `client/Redirect.tsx` navigates from a mount effect, so:

```tsx
render(<Page pathname="/app" onNavigate={fn}><Redirect href="/login" action="replace" /></Page>)
```

leaves `fn.mock.calls` empty — the listener wasn't registered when the push landed. A test written straight from docs/testing.md's navigation section fails with "Number of calls: 0", and the natural read is that the view's redirect is broken.

That makes auth-guard and onboarding-redirect views — the most common users of mount-time navigation — untestable through this prop. Registering during render (`useState(() => history.listen(...))`, cleaned up in an effect) closes the window.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+    <ServerDataContext.Provider value={serverData}>
+      <I18nProvider>
+        <WebSocketContext.Provider value={websocket}>
+          <QueryManagerProvider queryConfig={queryConfig}>
+            <ClientRouterContext.Provider value={routerContext}>
+              <RouteTransitionProvider
+                isPending={false}
+                isFetching={false}
+                transitionPath={["", pathname]}
+              >
```
nstfkc: **The `errorFallback` boundary is missing the `onReset={clearErrors}` wiring the real router has.**

`ClientRouter.tsx:139-143` builds its boundary with `resetKeys` and `onReset={clearErrors}`; this one is bare. So a test that mirrors a view's real `Error` export — a fallback whose "Try again" button calls `resetErrorBoundary()` — resets the boundary, re-renders the child, and `useQuery` reads the still-stored `state.error` off the `QueryResource` and throws it again. Nothing ever called `QueryManagerContext.clearErrors`.

The fallback stays on screen forever, so the retry path that works in the app can't be tested here, and the failure looks like the component's bug rather than the harness's. `QueryManagerContext` already exposes `clearErrors` — passing it as `onReset` is the same one-line wiring the router does.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+ *   >
+ *     <OrgChat />
+ *   </Page>,
+ * );
+ *
+ * expect(screen.getByText("hi")).toBeDefined();
+ * ```
+ *
+ * It is a plain component, so it composes with any renderer — Testing
+ * Library's `render`, `renderToString`, `react-test-renderer`.
```
nstfkc: **`renderToString` is advertised here and in docs/testing.md:31, but `<Page>` provides no `ServerQueryContext`.**

`useQuery`'s suspend branch is `if (typeof window === "undefined") { if (serverQueries) … }` — with no `window` and a null `ServerQueryContext` (which `createRoot.tsx` provides and this doesn't), `readPromise` stays undefined and the hook falls through to `data: undefined` instead of suspending.

Under `@vitest-environment node`, `renderToString(<Page fallback={<div>loading</div>}><View/></Page>)` with an unseeded `useQuery("/todos")` renders the no-data branch — `fallback` never appears — and prints the framework's "rendered on the server without data … add `Query.prefetch("/todos")` to the route's view handler" warning, sending the author to fix a route handler unrelated to the test. A view doing `data.map(...)` just crashes.

So both this JSDoc's "composes with any renderer" and the docs' "a query with no seed suspends into `fallback`" are false on the SSR path. Either provide `ServerQueryContext` or scope the claim to a DOM renderer.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+  };
+
+  const websocket = {
+    subscribe: async () => {},
+    unsubscribe: async () => {},
+    broadcast: () => {},
+  };
+
+  const ErrorFallback = errorFallback;
+  const content = <Suspense fallback={fallback}>{children}</Suspense>;
```
nstfkc, line 616: **`ThemeProvider` is the one provider from `ClientRouter`'s stack that isn't reassembled here.**

`ClientRouter.tsx:422` wraps the whole tree in `<ThemeProvider>`; this stack starts at `ServerDataContext.Provider`. `useTheme()` is a public `gemi/client` export, and its guard only throws on a *null* context — `ThemeProvider.tsx:12` gives the context a non-null default `{ theme: "light", setTheme: () => {} }`.

So a theme toggle under test renders, clicking it calls a no-op, `theme` stays `"light"`, no class lands on `document.documentElement`, and the assertion fails with no error and no missing-provider message. Adding `ThemeProvider` (with an optional `theme` prop to seed it) keeps the harness's "same tree `ClientRouter` builds" promise intact.

packages/gemi/client/useIsNavigationPending.ts
```diff
@@ -3,10 +3,15 @@ import { ClientRouterContext } from "./ClientRouterContext";
...
+  // Bound, like `useFormData` binds its own subject's: `Subject`'s methods
+  // read `this`, and `useSyncExternalStore` calls whatever it is handed as a
+  // plain function — so the unbound references this used to pass threw
+  // `undefined is not an object (evaluating 'this.value')` on the first render
+  // of any component that called this hook.
   const isNavigating = useSyncExternalStore(
-    isNavigatingSubject.subscribe,
-    isNavigatingSubject.getValue,
-    isNavigatingSubject.getValue,
+    isNavigatingSubject.subscribe.bind(isNavigatingSubject),
```
nstfkc: **The bind is right, but doing it in the render body churns the subscription.**

`isNavigatingSubject.subscribe` was a prototype method — the same function object on every render. `.bind(...)` allocates a fresh closure each render, and `useSyncExternalStore` tears down and re-creates the subscription whenever its `subscribe` argument changes identity. Every component rendering a navigation spinner now deletes and re-adds its entry in `Subject.subscribers` on each render.

That's mostly waste, but `Subject.next` notifies via `this.subscribers.forEach(...)` and `Set.forEach` visits entries inserted during iteration, so a `progressManager`-driven `next(true)` landing mid-resubscribe can notify both the outgoing and incoming subscriber.

Binding once restores the stable reference the deleted lines had — either `useMemo` here, or (better, since `useFormData` and `Mutation.tsx` each work around the same thing) binding in `Subject`'s constructor so every call site gets it for free.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+    queryData = {},
+    queryConfig,
+    user = null,
+    breadcrumbs = [],
+    fallback = null,
+    errorFallback,
+    onNavigate,
+  } = props;
+
+  const supportedLocales = Array.from(
```
nstfkc: **`supportedLocales` doesn't preserve the caller's order.**

`new Set([defaultLocale, locale, ...props.supportedLocales])` hoists the page's locale to the front. With `<Page locale="de-DE" supportedLocales={["en-US","tr-TR","de-DE"]}>` a language switcher mapping over the list renders `de-DE, en-US, tr-TR` instead of the configured order, so an assertion on the first option — or a snapshot of the switcher — fails even though the app renders it correctly.

The docs describe the prop as seeding "`Lang`-style locale lists", which implies the app's own list verbatim. Appending the defaults instead of prepending them (`new Set([...(props.supportedLocales ?? []), defaultLocale, locale])`) keeps the caller's order while still guaranteeing both are present.

packages/gemi/vite.client.config.mts
```diff
@@ -4,13 +4,27 @@ import { defineConfig } from "vite";
...
       formats: ["es"],
-      fileName: () => `index.js`,
+      fileName: (_format, entryName) => `${entryName}.js`,
     },
     minify: false,
-    outDir: "dist/client",
+    outDir: "dist",
+    // `dist/` already holds the output of `build:core`, `build:bin` and
+    // `build:types` by the time this runs — this build adds to it.
+    emptyOutDir: false,
```
nstfkc, line 32: **`emptyOutDir: false` drops the implicit clean the old `outDir: "dist/client"` had.**

Vite empties an `outDir` inside the project root by default, so the previous config swept `dist/client` on every run. Writing to `dist` with emptying disabled is correct for coexisting with `build:core`/`build:bin`/`build:types` — but this build now also emits content-hashed chunks (`chunks/[name]-[hash].js`).

Run `bun run build:client` on its own — the normal incremental loop, or any pipeline step reusing a warm `dist/` rather than going through `bun run build` (only `build:core` does `rm -rf dist`) — and every earlier run's `dist/chunks/*-<oldhash>.js` stays on disk. `scripts/build-publish.ts` then does `cp("dist", ".publish/dist", { recursive: true })` with no filtering, and `files: ["dist/**/*"]` ships it, so a tarball staged from such a tree carries dead chunks no export references.

A targeted `rm -rf dist/chunks dist/client dist/testing` in the `build:client` script would keep the coexistence and the sweep.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+  // The URL's locale segment, which the default locale does not get.
+  const urlLocaleSegment = locale === defaultLocale ? null : locale;
+
+  const dictionary = toClientDictionary(
+    dictionaries,
+    supportedLocales,
+    locale,
+    translations,
+  );
+  const prefetchedData = toPrefetchedData(queryData, params);
```
nstfkc: **Every derived value is rebuilt on each render, including the `/api/` warning.**

`dictionary`, `prefetchedData`, the breadcrumb `Map`/`Record`, `routeState`, `serverData` and `routerContext` are all computed in the render body — in contrast to `isNavigatingSubject`, `progressManager` and `history` right below, which are held in `useState`.

Rendering `<Page queryData={{"/api/todos": []}}>` and calling `rerender` twice logs the "starts with /api" warning three times, so a test that re-renders floods its output with duplicates. Meanwhile `toClientDictionary`/`toPrefetchedData` re-loop over every key, and `ServerDataContext`/`ClientRouterContext`/`RouteStateContext` each get a fresh value object, re-rendering every consumer under the page on each pass.

`useMemo` keyed on the props each derives from does the same job once.

packages/gemi/testing/Page.tsx
```diff
@@ -0,0 +1,435 @@
...
+    typeof search === "string"
+      ? search.replace(/^\?/, "")
+      : new URLSearchParams(
+          Object.entries(search).map(([key, value]) => [key, String(value)]),
+        ).toString();
+  return query ? `?${query}` : "";
+}
+
+/** The cache's variant key: sorted search params, empty when there are none. */
+function toVariantKey(search: string): string {
```
nstfkc: **`toVariantKey` is a fourth copy of the cache-variant key.**

The same "`new URLSearchParams` → `sort()` → `toString()`" lives at `client/useQuery.ts:216-218`, `client/useMutate.ts:39-40` and `client/QueryManager.ts:65-66`. `useQuery` additionally runs `omitNullishValues` before building its key.

That divergence is the risk: if the normalization or the sort ever changes on the read side, this copy keeps producing the old key and every `queryData={{"/lists?page=2": …}}` seed silently misses — the test then fetches over the network and asserts an empty state, with nothing pointing at the key mismatch. One `variantKey(search)` helper in `utils/`, called from all four sites, removes the drift.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255`. Confirmed both halves — vitest throws `Missing parameter: orgId` out of `applyParams`, `bun test` seeds `/orgs/undefined/lists`.

`resolveSeedPath` now checks a key's required params against the page's before resolving, and fails with the key in the message:

```
[gemi] <Page> cannot resolve the queryData key "/orgs/:orgId/lists": no value for
:orgId. The query cache is keyed by the resolved path, so either add `orgId` to the
page's `params`, or seed the resolved path (e.g. "/orgs/123/lists") — which is what a
query taking its params from component state reads under.
```

Same on both runners, since it no longer depends on `import.meta.env.DEV`. I went with failing rather than opting out of resolution: a key that silently seeds nothing is the failure mode this whole harness exists to remove, and pointing at the resolved path is the actual answer for the call-site-params case you describe.

One thing your example surfaced that I could not preserve: an optional `:folderId?` cannot be written in a key at all, because the key is split at its first `?` to separate the search string. `requiredParamsOf` filters `*` wildcards only, and the collision is documented on the prop.

Test: `a key whose :param the page cannot fill fails naming the key`.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255`. Reproduced exactly as written — `<Redirect>` in a `<Page>` left `onNavigate` uncalled.

The listener is now registered during render, ref-guarded so React's double-invoke registers one:

```tsx
const listeningRef = useRef(false);
if (!listeningRef.current) {
  listeningRef.current = true;
  history.listen(({ location, action }) => { … });
}
```

I took `useRef` over `useState(() => history.listen(…))` deliberately: a `useState` initializer is re-invoked under StrictMode, and cleaning it up in an effect would unsubscribe on StrictMode's mount→unmount→mount cycle without a render in between to re-subscribe. Nothing unsubscribes at all now, which is safe here because the memory history is created per `<Page>` and reachable from nowhere else — it is collected with the page.

Test: `catches a navigation issued from a child's mount effect`, using `<Redirect>` itself rather than a stand-in.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255`, with the same pair the router uses — `resetKeys={[pathname]}` and `onReset={clearErrors}`.

It needed a small inner component (`Boundary`) rather than an inline change, because `clearErrors` comes from `QueryManagerContext` and the boundary sits inside `QueryManagerProvider`. That also gave the `Suspense` boundary a home, so the two now live together the way `Route` has them.

Test: `resetErrorBoundary clears the stored query error, so a retry recovers` — a 500 puts the fallback on screen, the next fetch succeeds, clicking "Try again" renders the data. It fails without the `onReset` wiring, which is what I wanted to pin.

packages/gemi/testing/Page.tsx
nstfkc: Correct, and I scoped the claim rather than providing the context — `e861255`.

Providing a `ServerQueryContext` would mean inventing a store, and the store is exactly the part a harness cannot have: `createRoot` threads the *request's* `ServerQueryStore` through, and what makes it able to suspend is that it runs the route handler. There is no request and no handler here, so anything I put behind that context would be a fake that resolves to nothing — which is a worse lie than the current behaviour.

What is actually true is now stated in both places: seeded queries render their data under `renderToString` exactly as in the browser; an unseeded one renders the server's no-data branch and logs the framework warning — the same thing that happens to a real route whose view handler declares no `Query.prefetch`. The docs gained a `## Server rendering` section saying to assert seeded output under `renderToString` and use a DOM renderer for the suspend-then-resolve path, and the JSDoc's "composes with any renderer" now carries the caveat.

Worth noting the warning text is arguably wrong for a test author either way — it sends you to a route handler that has nothing to do with the test. I left it alone as it is correct for its main audience, but flagging it in case you want it to mention seeding.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255` — `<ThemeProvider>` now wraps the stack, and I took the optional prop you suggested.

`ThemeProvider` gained `theme?: Theme`, used ahead of the stored value:

```tsx
const [theme, setTheme] = useState(() => {
  if (props.theme) return props.theme;
  if (typeof window === "undefined") return "light";
  return localStorage.getItem("theme") || "light";
});
```

Backward compatible — the app passes nothing and still reads `localStorage`. `<Page theme="dark">` forwards it, so a test can render a component in a theme without writing to storage, and `setTheme` works from there.

Test: `seeds the theme, and setTheme still works from there`. The silent-no-op you describe is exactly what it caught first — the click needed `fireEvent` to flush, which the bare `.click()` in the earlier navigation test did not need because that one asserts a mock call rather than a re-render.

packages/gemi/client/useIsNavigationPending.ts
nstfkc: Taken the second option — `Subject` binds in its constructor (`e861255`), so every call site gets a stable, correct reference and the hook goes back to the plain property reads it had:

```ts
constructor(initialValue: T) {
  this.value = initialValue;
  this.subscribe = this.subscribe.bind(this);
  this.next = this.next.bind(this);
  this.getValue = this.getValue.bind(this);
}
```

`useFormData` and `Mutation.tsx` drop their `.bind` calls with it — they had the same per-render churn, just without the crash to make it visible.

`utils/Subject.test.ts` pins both properties, since neither is obvious from the class: methods work detached from the instance, and `subject.subscribe === subject.subscribe` across reads. The `Set.forEach`-visits-entries-inserted-during-iteration point is what convinced me the churn was worth fixing rather than tolerating — thanks for spelling it out.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255`, exactly as suggested — the defaults are appended now:

```ts
Array.from(new Set([...(props.supportedLocales ?? []), defaultLocale, locale]))
```

Presence is still guaranteed, order is the caller's. The prop's docs row said "`Lang`-style locale lists", which was vague about the thing that matters; it now says the app's list verbatim, with the defaults appended if absent.

Test: `keeps the caller's supportedLocales order` — `locale="de-DE"` with `["en-US","tr-TR","de-DE"]` renders in that order.

packages/gemi/vite.client.config.mts
nstfkc: Fixed in `e861255`, with the targeted sweep you propose:

```json
"build:client": "rm -rf dist/client dist/testing dist/chunks && vite build -c vite.client.config.mts"
```

Exactly the three directories this config writes, so the coexistence with `build:core`/`build:bin`/`build:types` is unaffected and the hashed chunks cannot accumulate. `emptyOutDir: false` now carries a comment explaining what it gives up and where that is made good, so the next person to read it does not restore the default.

Verified by running `build:client` twice in a row: `dist/chunks/` holds one chunk, not one per run. The path from stale chunk to shipped tarball you traced — `cp("dist", …)` with no filtering, plus `files: ["dist/**/*"]` — is what made me fix it in the script rather than just documenting it.

packages/gemi/testing/Page.tsx
nstfkc: Fixed in `e861255`. `dictionary`, `prefetchedData`, the breadcrumb `Map`/`Record`, `routeState`, `serverData` and `routerContext` are all `useMemo`d on the props they derive from, and the two inline objects that were also rebuilt per render (`viewEntriesSubject`, the websocket stub) moved into `useState`.

The duplicated warning needed its own fix rather than falling out of the memo: a test that passes `queryData` as an inline literal gets a new object identity every render, so the memo misses and the warning would fire again. It is now guarded by a per-`<Page>` ref of keys already warned about, which makes it once-per-key regardless of how the props are written.

Test: `the /api-prefix warning fires once per key, not once per render` — renders and re-renders twice, expects one call.

packages/gemi/testing/Page.tsx
nstfkc: Extracted in `e861255` — `utils/variantKey.ts`, one `toVariantKey(search)` taking either the object form the hooks hold or the query-string form a seed key carries, with `omitNullishValues` folded in so the read and write sides cannot disagree about it either.

`useQuery`, `useMutate` and `<Page>` all call it. Named `toVariantKey` rather than `variantKey` because both `useQuery` and `useMutate` hold a local `const variantKey`, and in `useQuery` the local is declared *after* the call site — an import of that name would have been a TDZ error rather than a shadowing warning.

The fourth copy, `client/QueryManager.ts`, I left alone: nothing imports it, so it is dead rather than drifting (it is a pre-`QueryResource` implementation). Happy to delete it in a follow-up if you agree it is dead — it seemed the wrong thing to fold into this PR.

`utils/variantKey.test.ts` pins the properties the copies had to agree on: sort-independence, nullish omission, and that the string and object forms produce the same key.

Issue #286, opened 2026-08-01
Crawler UAs still get an unsettled document: awaiting allReady does not inline boundaries over progressiveChunkSize


nstfkc:
Found while migrating a real app to `0.49.0-rc.1`. The bot path in `ViewRouter.render()` awaits `allReady` before responding, but that is not sufficient to produce a settled document: React Fizz splits a boundary out of the shell once its content exceeds `progressiveChunkSize` (default ~12,800 bytes), and that decision is made *during* render, so awaiting afterwards cannot undo it.

The result is that large pages ship their entire body inside `<div hidden id="S:0">…</div><script>$RC("B:0","S:0")</script>` even for crawler UAs. A fetcher that does not execute JS sees a body containing nothing but `<title>`.

This matters most for exactly the non-JS agents already in gemi's own regex — `whatsapp`, `telegram`, `discord`, `facebookexternalhit`, `twitterbot`, `linkedinbot` — which power link previews. It is a regression against `0.48.1`, which had no streaming and always emitted complete HTML.

## Measured

Production build (`gemi build` + `gemi start`), requested with `-A facebookexternalhit/1.1`. "deferred" = visible text reachable only by running the `$RC` scripts.

| route | text in shell | deferred (needs JS) |
|---|---|---|
| `/` (landing) | **39** — just the `<title>` | **3,666** |
| `/terms-and-conditions` | **39** | **7,417** |
| `/about-us` | 4,190 | 0 |
| `/pricing` | 930 | 0 |
| `/support-and-faq` | 3,262 | 0 |

The split tracks content size, not route shape or query use — the two failing routes are simply over the chunk budget. `/` defers 67 KB of HTML this way.

## Cause

`isBotUserAgent` fires correctly and `await stream.allReady` runs; neither is the problem. Isolated repro against React 19 (no gemi involved):

```jsx
const app = <html><body><h1>shell</h1><Suspense fallback={null}><Slow /></Suspense></body></html>;
const stream = await renderToReadableStream(app, opts);
await stream.allReady;
const html = await new Response(stream).text();
```

```
rows=10   opts={}                                   deferred($RC)=false
rows=100  opts={}                                   deferred($RC)=false
rows=400  opts={}                                   deferred($RC)=true    <-- over ~12.8 KB
rows=400  opts={progressiveChunkSize: MAX_SAFE}     deferred($RC)=false
```

[...]

PR #397, opened 2026-08-12
Per-component dictionaries, split by locale at build time

nstfkc:
## Why

i18n did not scale. A dictionary was declared with `Dictionary.create(name, {...})` in `app/i18n/`, re-exported from `app/i18n/index.ts`, and listed against a route in a hand-maintained `prefetch` map. Three costs followed:

- **Files grew far from the components they serve.** folio's `app/i18n/` is ~14k lines across 7 locales.
- **The `prefetch` map drifted.** Nothing checked it, and the only symptom of getting it wrong was a missing string at runtime.
- **Everything was resident.** `Translator`'s constructor materialized the full locale × route × dictionary × key cross-product at boot and held it for the life of the process.

## What

A dictionary is declared next to the component that reads it, and a visitor downloads only the locale they are reading in:

```ts
// app/views/Home.i18n.ts
export const homeDict = defineDictionary({
  title: { "en-US": "Welcome {{name}}", "tr-TR": "Hoş geldin {{name}}" },
});
```
```tsx
// app/views/Home.tsx
const t = useDictionary(homeDict);
t("title", { name });
```

No name string, no central index, no `prefetch` entry, no `gemi.d.ts` — keys and interpolation params are inferred from the literal. Interpolation syntax (`{{p}}`, `{{p:number}}`, `{{p:[content]}}`, `t.jsx`) is unchanged, so strings migrate verbatim.

### Locale splitting

`packages/gemi/vite/dictionaryPlugin.ts` rewrites the call at build time into one lazy import per locale, so the strings leave the component's chunk entirely:

```js
const homeDict = __gemi_dict__("d_f3cc5a4461dfa4b2", {
  "en-US": () => import("./en-US-BSyWTQsL.js"),
  "tr-TR": () => import("./tr-TR-o52Tzo3M.js"),
});
```

The plugin is **purely an optimization** — untransformed (in `bun test`, or in server code that never meets the bundler) `defineDictionary` holds every locale and resolves synchronously. That equivalence is load-bearing and is why the source locale is a per-dictionary `sourceLocale` rather than a build-time global.

Because the literal is read at build time it must be statically analysable, and `defineDictionary` must be imported by name; anything the plugin cannot trace warns at build time rather than silently shipping every locale.

### Delivery

Lazy `import()` + Suspense, but as the fallback rather than the normal path.

- **SSR** preloads whenever dictionaries are registered — independent of any `translation` config, since the documented minimal setup has none — then streams the strings it used through the existing `streamQueryInjection` queue, which already guarantees a payload lands ahead of the segment that used it.
- **The client router** warms a view's dictionaries inside `loadViewModule`, folding the fetch into the loading state the route already shows.
- **Suspense** covers what neither can see: a dictionary inside a `lazy()` subtree, and the first render after a locale switch.

## Design notes worth reviewing

**Dictionary ids are content hashes, not paths.** The id the plugin bakes into a chunk and the id the runtime computes from the same literal must agree by construction, or the SSR payload and the client registry silently miss each other.

**The registry lives on `globalThis`.** Verified against a staged `.publish/`: the registry module is compiled into two separate bundles — one from the Bun build behind `gemi/dictionary`, one from the Vite build behind `gemi/client`/`gemi/testing`. Module scope would give an installed app two registries and an empty hydration payload, while passing every test in the package.

**Handles re-assert their registration on read.** A dictionary is declared at module scope, so its handle outlives anything that clears the registry under it — a test's `afterEach`, an HMR boundary.

## Migration

`Dictionary.create`, `useTranslator` and the `prefetch` config are deprecated but **untouched at runtime**, so an app migrates one dictionary at a time. The template migrates `Home` and deliberately leaves `About` on the old API so both paths stay exercised. `docs/i18n.md` leads with the new API and carries a migration section.

## Testing

Works with `gemi/testing`'s `<Page>` with no changes to it. The seeding story inverts: a `useTranslator` component needs its strings handed in, a `useDictionary` component brings its own, so a test seeds `locale` and nothing else. Both APIs compose on one page. `docs/testing.md` covers it.

## Verification

- 3449 tests passing, typecheck clean, 28 type tests, lint clean.
- Every fix from review has a test that fails without it; for the ordering-sensitive ones I reverted the change to confirm the test catches it rather than passing regardless.
- **Build:** the view chunk carries zero translations; each locale chunk carries exactly one locale. A `lazy()`-only component gets its own per-locale chunks, so splitting does not depend on the view statically importing the dictionary.
- **Runtime**, both locales, prod and dev: correct locale renders, no cross-locale leakage in markup or payload, legacy `/about` still renders from `Dictionary.create`.
- **The SSR preload was measured, not assumed:** with the documented minimal setup, restoring the old gate produces 2 reveal segments where the fix produces 1 — matching a control route that has no dictionary at all. A dictionary now adds zero reveal chunks.

## Review

Nine findings were raised and all nine were valid; see the [review reply](https://github.com/nstfkc/gemi/pull/397#issuecomment-5265956035). One is answered only in part: `sourceLocale` removes the order-dependence of the source language but does not make `defaultLocale` authoritative. Doing that without breaking plugin/runtime equivalence needs a boot-time default the runtime also reads — a follow-up, not a blocker.


packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -851,6 +867,16 @@ export class ViewRouteDispatcher {
...
+
+          // Resolve every `defineDictionary` dictionary this process knows
+          // about for the locale, before rendering. The view modules were
+          // imported (and so registered their dictionaries) before the request
+          // arrived, and already-resolved ones return from cache, so this costs
+          // one pass over freshly imported modules. It buys a render in which
+          // `useDictionary` never suspends — which matters more here than on
+          // the client: a suspend splits the segment out of the shell, and a
+          // page would otherwise fragment into one reveal chunk per dictionary.
+          await preloadDictionaries(locale);
```
nstfkc: **SSR preload is gated behind `isI18nEnabled`, which the documented minimal setup never satisfies.**

`isI18nEnabled` is `hasLegacyDictionaries || translator.supportedLocales.length > 0`, and `supportedLocales` defaults to `[]` (`i18n/config.ts` → `translationConfigDefaults`). But `docs/i18n.md` presents `defineDictionary` + `useDictionary` as "the whole setup" and explicitly says dictionaries are *not* listed in config — so an app that adopts the new API without ever writing a `translation` config slice takes the `isI18nEnabled === false` branch and `preloadDictionaries(locale)` never runs.

Every component calling `useDictionary` then suspends during SSR on its locale chunk. On a normal streamed (non-bot, non-`no-stream`) route that means each component's markup is emitted inside `<div hidden>` with a `$RC()` reveal script instead of inline in the shell — exactly the outcome the surrounding `#286`/`#289`/`#294` comments call out. A JS-disabled visitor or a script-less fetcher sees the Suspense fallback (`Loading`, or blank when the view has no `Loading` export) where the page text should be, and every request pays a per-dictionary reveal chunk.

The preload probably needs to run whenever any dictionary is registered, independent of the legacy translator config.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -824,7 +834,14 @@ export class ViewRouteDispatcher {
...
         const translator = app(Translator);
-        const isI18nEnabled = translator.isEnabled;
+        // Two ways to be an i18n app now. `isEnabled` means the legacy
+        // `prefetch` config has dictionaries to serve; configured
+        // `supportedLocales` alone means the app uses `defineDictionary` and
+        // still needs its locale detected, or every `useDictionary` would fall
+        // back to its source language.
+        const hasLegacyDictionaries = translator.isEnabled;
+        const isI18nEnabled =
+          hasLegacyDictionaries || translator.supportedLocales.length > 0;
```
nstfkc: **The locale-prefix redirect gate (line 710) was not widened alongside this one.**

Line 710 still reads `if (translator.isEnabled && !isOgRequest)`, while the render gate here is the broader `hasLegacyDictionaries || translator.supportedLocales.length > 0`. The two now disagree.

Concretely: an app migrated to `defineDictionary` sets `supportedLocales: ["en-US", "tr-TR"]`, `defaultLocale: "en-US"` and an empty `prefetch`, so `translator.isEnabled` is `false` but `isI18nEnabled` here is `true`. A Turkish visitor requests `/`:

- line 710 skips the redirect block, so no `302` to `/tr-TR/` is issued;
- line 855's `detectLocale` resolves `tr-TR` and the page renders fully in Turkish at `/`.

The response carries no `Vary` on locale, so any CDN or reverse proxy caching `/` then serves that Turkish document to every subsequent visitor, and there is no locale-prefixed URL to bookmark or share. Under the legacy `Dictionary.create` config the same app would have redirected.

packages/gemi/client/useDictionary.ts
```diff
@@ -0,0 +1,88 @@
...
+  const sink = useContext(DictionarySinkContext);
+  // An app with i18n switched off still renders dictionaries — fall back to the
+  // source language rather than blowing up on an empty `i18n` payload.
+  const locale = i18n?.currentLocale ?? dictionary.locales[0];
+  // Tells the view loader which locale's chunks to warm on the next
+  // navigation. `__GEMI_DATA__` only knows the locale the document loaded in.
+  setActiveLocale(locale);
+
+  const pending = loadDictionary(dictionary.id, locale);
+  const strings = pending instanceof Promise ? use(pending) : pending;
```
nstfkc, line 58: **A failed dictionary load now escapes the render phase instead of degrading to the raw key.**

The deprecated `useTranslator` wrapped every lookup in `try { ... } catch { console.error(...); return String(key) }`, so any i18n failure degraded to rendering the key. `useDictionary` re-establishes that guard only for the "key absent from the resolved strings" case (line 70) — not for a *failed load*.

`loadDictionary` returns a promise that rejects (it deletes the `inFlight` entry and rethrows), and `use(pending)` here rethrows that rejection during render:

- **Client:** after a rolling deploy, a browser holding stale HTML requests a locale chunk whose hashed filename no longer exists. The dynamic `import()` rejects with `Failed to fetch dynamically imported module`, and instead of the page rendering with untranslated keys, the whole route unmounts into `Route`'s ErrorBoundary (`ClientRouter.tsx` ~139).
- **Server:** if the throw happens before the shell is ready, `renderToReadableStream` rejects into `ViewRouteDispatcher`'s outer `catch` and the visitor gets the render-error document.

Nothing is cached on failure, so every retry re-throws. `loadViewModule.test.ts` ("a failing dictionary does not fail the view") only asserts that `loadViewModule` resolves — it never renders the component afterwards, so this path is untested.

packages/gemi/client/ComponentContext.tsx
```diff
@@ -27,19 +32,48 @@ export function loadViewModule(name: string): Promise<any> {
...
 }
 
+function currentLocale(): string {
+  // `getActiveLocale` is whatever the last render used, which survives a locale
+  // switch; `__GEMI_DATA__` covers the first navigation, before any
+  // `useDictionary` has run.
+  return (
+    getActiveLocale() ??
+    (typeof window !== "undefined"
+      ? (window.__GEMI_DATA__?.i18n?.currentLocale ?? "")
```
nstfkc, line 72: **`currentLocale()` falls back to `""` while `useDictionary` falls back to `dictionary.locales[0]` — the preload warms a key no render ever reads.**

`useDictionary` resolves its locale as `i18n?.currentLocale ?? dictionary.locales[0]` (`useDictionary.ts:47`), but `currentLocale()` here resolves the same thing as `getActiveLocale() ?? window.__GEMI_DATA__?.i18n?.currentLocale ?? ""`. The two disagree whenever `i18n.currentLocale` is absent — which is exactly the case for any app without `supportedLocales` configured (the dispatcher leaves `i18n = {}`), the setup `docs/i18n.md` documents as supported.

On the first navigation of such an app:

1. `preloadDictionaries("", mark)` calls `loadDictionary(id, "")`;
2. `__gemi_dict__` falls through to `loaders[locales[0]]` and `putResolved(id, "", strings)` caches under `""`;
3. the subsequent render asks for `dictionary.locales[0]` (`"en-US"`), misses the cache entirely, and issues a second `loadDictionary`.

So the view suspends into its `Loading`/`null` fallback on exactly the navigation the preload exists to keep flash-free, and the registry accumulates a duplicate `""` entry per dictionary that is never read again. Both call sites should share one locale-resolution helper.

packages/gemi/vite/dictionaryPlugin.ts
```diff
@@ -0,0 +1,317 @@
...
+        for (const spec of node.specifiers ?? []) {
+          if (
+            spec.type === "ImportSpecifier" &&
+            spec.imported?.name === "defineDictionary"
+          ) {
+            localNames.add(spec.local.name);
+          }
+        }
+      }
+      if (localNames.size === 0) {
```
nstfkc: **This is the fail-open path the file's own contract says must not exist.**

`readObjectLiteral` hard-fails the build on any non-analysable literal precisely because "a silent fallback to 'leave this one un-split' would ship every locale to the browser and nobody would notice." But the import scan gating it only collects `ImportSpecifier`s named `defineDictionary` whose `source.value` is literally `gemi/dictionary`, `gemi/client` or `gemi/i18n`, and returns `null` here otherwise — no error, no warning.

Forms that silently slip through:

- namespace import — `import * as gemi from "gemi/client"; gemi.defineDictionary({...})`
- aliased re-export through an app-local barrel — `import { defineDictionary } from "@/app/lib/gemi"`
- a dictionary declared in a linked workspace package whose id contains `/node_modules/`

All yield `localNames.size === 0` → `return null`. The dictionary stays inline, so every locale's strings ship in the component's client chunk to every visitor, and the team gets no build signal that the locale-splitting the docs promise silently stopped applying to that file.

At minimum this deserves a warning when a file contains a `defineDictionary(` call expression that the import scan did not resolve.

packages/gemi/i18n/translate.ts
```diff
@@ -0,0 +1,41 @@
...
+  U extends keyof T[K],
+  R = Prettify<ParseTranslationParamsServer<T[K][U] & string>>,
+>(
+  dictionary: DictionaryHandle<T>,
+  key: K,
+  ...args: R extends Record<string, never>
+    ? [args?: { locale?: U | (string & {}) }]
+    : [args: { locale?: U | (string & {}); params: R }]
+): Promise<string> {
+  const { locale = Lang.locale(), params } = { params: {}, ...args[0] };
```
nstfkc, line 30: **`translate()` re-detects the locale instead of using the one the request is rendering in.**

`Lang.locale()` (`facades/Lang.ts:18-26`) never consults `RequestContext`'s locale field — it calls `translator.detectLocale(requestStore.req)`, which reads the *incoming* `i18n-locale` cookie and `accept-language`. So it ignores both `ctx.locale` and the URL locale segment.

Failure case: a returning visitor whose `i18n-locale` cookie says `en-US` opens `/tr/about`. `handleViewRequest` takes the URL-locale branch, calls `ctx.setLocale("tr-TR")` and renders the page in Turkish. A controller or email template on that same request calling `await translate(dict, "subject")` hits `Lang.locale()` → reads the old cookie (since `Lang.setLocale` only queues a *response* cookie) → returns `en-US`.

The visitor gets a Turkish page containing an English breadcrumb/subject/flash message, or an order-confirmation email in the wrong language, with no error anywhere. The default should come from the request context's active locale.

packages/gemi/i18n/defineDictionary.ts
```diff
@@ -0,0 +1,101 @@
...
+  registerDictionary({ id, locales, load });
+
+  return {
+    id,
+    locales,
+    // Routed through the registry rather than calling `load` directly so
+    // concurrent readers share one in-flight import and results are cached.
+    load: (locale: string) => loadDictionary(id, locale),
+    get: (locale: string) => {
+      const strings = loadDictionary(id, locale);
```
nstfkc, line 136: **Dropping this promise leaves an unhandled rejection that can take the server down.**

`loadDictionary` builds `promise = result.then(ok, err => { inFlight.delete(key); throw err; })`, stores it in `inFlight` and returns it. `get` sees `strings instanceof Promise`, returns `undefined` and drops the reference — nothing else awaits it (`preloadDictionaries` only sees ids it slices from `order`, and `useDictionary` may never run for this dictionary).

So when server or worker code calls `dict.get("tr-TR")` on a bundled dictionary whose locale chunk 404s after a deploy — a common transient during a rolling release — the rejection escapes as an `unhandledRejection`. Under Node's default handler that terminates the process: the whole server goes down for one missing translation chunk.

Attaching a no-op `.catch()` before discarding the promise is enough to contain it.

packages/gemi/i18n/dictionaryShape.ts
```diff
@@ -0,0 +1,102 @@
...
+export type LocaleStrings = Record<string, string>;
+
+/**
+ * Locales the dictionary declares, in first-seen order.
+ *
+ * The order is meaningful: the first locale is the source language, and a key
+ * missing from some other locale falls back along this list. Write the locale
+ * you author in first.
+ */
+export function dictionaryLocales(
```
nstfkc, line 28: **The source language ends up decided by key iteration order rather than the configured `defaultLocale`.**

`dictionaryLocales` walks `Object.values(translations)` and returns locales in first-seen order; `localeStrings` then falls back along that list, and `useDictionary` uses `locales[0]` as its no-i18n default.

That makes the source language an emergent property of literal ordering: adding a new key at the *top* of a dictionary that is so far only translated in `tr-TR` flips the source language for the whole dictionary. Every key missing an `en-US` string, and every render in an app whose i18n is off, then resolves to Turkish for English readers — from a diff that only added one key.

gemi already has an authoritative answer for this in `TranslationConfig.defaultLocale` / `translator.defaultLocale`; this path invents a second, order-dependent one.

packages/gemi/i18n/dictionaryRegistry.ts
nstfkc, line 1: **`cacheKey` (line 83) embeds a raw NUL byte, which makes git classify this whole file as binary.**

There is a literal `U+0000` at byte offset 3372, inside the template string on line 83 — `` return `${id}\x00${locale}`; `` was written with an actual NUL rather than the two-character `\0` escape.

Git's binary heuristic therefore treats `dictionaryRegistry.ts` as a blob:

- `git diff` and the GitHub PR view print only *"Binary files differ"* / *"Binary file not shown"* — for this PR's most central new module (`git diff --stat` shows it as `Bin 0 -> 9216 bytes`);
- `git grep` reports *"Binary file matches"* without the matching line;
- `git blame` and patch-based tooling degrade the same way.

So reviewers see 273 lines of registry logic — the `globalThis` state, the watermark, the `__GEMI_DICT__` adoption shim — as an unreadable diff unless they know to pass `--text`. (This comment is file-level for exactly that reason: GitHub has no diff lines here to anchor to.)

Writing the delimiter as the escape `\0` produces an identical runtime key and keeps the file textual.

---

**Second issue in this file — `preloadDictionaries` advances the watermark before its loads settle (line ~194).**

```
const from = mark ?? warmed.get(locale) ?? 0;
const ids = order.slice(from);
if (mark === undefined) { warmed.set(locale, order.length); }   // <- synchronous
...
await Promise.all(...)
```

The watermark moves before the imports resolve, so a concurrent request warms nothing:

1. Cold-start burst — two requests for `/` in `en-US` arrive ~1ms apart.
2. Request 1 sets `warmed["en-US"] = order.length` and awaits the locale chunk imports.
3. Request 2 computes `from = warmed.get("en-US")`, gets an empty `order.slice(from)`, hits the `ids.length === 0` early return, and starts rendering while request 1's imports are still in flight.
4. `useDictionary` finds only the `inFlight` promise and calls `use(pending)` — so request 2 suspends and the visitor gets a streamed Suspense skeleton (or an extra reveal chunk per dictionary) instead of the full server-rendered page. That is exactly the fragmentation this preload exists to prevent.

The same line also moves the watermark past dictionaries whose load *rejected* and was swallowed by the `catch` below, which contradicts the doc comment's claim that a failed id "stays below the watermark" — those dictionaries are never retried by a later preload.

Setting the watermark after `await Promise.all(...)`, and to the count actually resolved, fixes both halves.

nstfkc:
All nine findings were real and all nine are fixed in `b213902`. Nothing here needed pushing back on. Grouped by what each one breaks, with the verification where it exists.

### The file nobody could review

`cacheKey` embedded a literal `U+0000` instead of the `\0` escape, so git classified `dictionaryRegistry.ts` as binary — the change's most central module rendered as *"Binary file not shown"*. Confirmed exactly as described (`file` → `data`, `Bin 0 -> 9216 bytes`); my own `grep` for `cacheKey` silently returning nothing was the same symptom. Written as the escape it produces an identical runtime key, and `git diff main...HEAD` now shows all 366 lines.

### Failures that escalated instead of degrading

- **`useDictionary`** — a rejected locale chunk rethrew out of `use()` during render, unmounting the route into its error boundary or failing the server render. `loadDictionaryForRender` restores the `useTranslator` behaviour of degrading to the raw key. The degraded promise is cached because `use()` needs a stable reference, and it is only consulted after `getResolved` misses, so a later successful load supersedes it without eviction.
- **`handle.get`** — now attaches a no-op `catch` before discarding. As described, this was an `unhandledRejection` that ends the process over one missing translation file.

### The preload not doing its job

- **Gating.** Moved out of the `isI18nEnabled` branch entirely. This was the most consequential finding, and I measured it rather than assuming: with the documented minimal setup (`defineDictionary`, no `translation` config), restoring the gate produces **2 reveal segments** where the fix produces **1** — the extra one being the dictionary suspending. Against a control route with no dictionary at all (`/pricing`, also 1), a dictionary now adds zero reveal chunks.
- **Watermark.** Correct on both halves you raised — it now advances after `Promise.all` settles and only across the leading run that resolved, so "below the watermark" and "resolved" are the same statement, and a swallowed failure is retried by the next preload. This one was mine from `2a31e23`, introduced by the fix for the O(all) rescan.

### Locales resolved two different ways

- **`currentLocale()` vs the hook** — preloading now resolves per dictionary the way `useDictionary` does, so an empty locale warms each dictionary under its own source language rather than caching under `""`.
- **The redirect gate** — both it and the render gate now read a single `Translator.isLocaleAware`. The cache-poisoning path you described was the deciding detail; two predicates that must agree should not be written twice.
- **`Lang.locale()`** — now prefers the request context's locale, falling back to detection. `Lang.setLocale` also records it there, since otherwise setting the locale did not change what the rest of the request rendered in. Worth noting this fixes `Dictionary.text`/`Dictionary.render` on the legacy path too — `translate()` had inherited the flaw rather than introduced it.

### Two silent fail-open paths

- **Source language** — `defineDictionary` now takes an optional `sourceLocale`, read at build time as well as at runtime so the bundled and unbundled paths cannot disagree. **This is a partial answer to your point:** it removes the emergence for anyone who sets it, but it does not make `defaultLocale` authoritative, because the plugin bakes fallback order into each per-locale chunk at build time and has no access to `app/config/translation.ts`. Making the configured default win would mean teaching the plugin to read that config. Happy to do it as a follow-up if you want it — flagging rather than quietly redefining the finding as fixed.
- **The plugin's import scan** — now warns on any `defineDictionary(` call expression it cannot trace to a gemi import, covering namespace imports and app-local barrels, while staying quiet when a module declares its own `defineDictionary`. Documented in `docs/i18n.md` alongside the static-analysability constraint.

### Verification

Every fix has a test that fails without it, including the ones that only appear under concurrency or after a rejected load. For the two ordering-sensitive fixes I reverted the change to confirm the test actually catches it rather than passing regardless.

Separately, `0262e58` covers dictionaries inside **dynamically imported components** — the one path no preload can reach, and one nothing exercised before (the template has no `lazy()`, and every existing test registered its dictionary before rendering). Against a real build, a `lazy()`-only component gets its own per-locale chunks distinct from the view's, its own chunk carries no translations, and server-rendering `/tr-TR` emits both dictionaries' payloads in Turkish only with the panel's payload at byte 1543 against its markup at 48033.

### Still not verified

Client-side navigation has not been driven through a real browser — there is no Playwright or Puppeteer in this repo. The `loadViewModule` path it runs is covered by jsdom tests with real promise ordering, and the built artifacts are correct, but "no flash on navigation" remains reasoned rather than observed.

nstfkc:
Ready to merge. Since the review: `main` merged in (0.53.0 + `gemi/testing`), the two open questions answered, description brought up to date.

**On defining the default locale in `gemi.config.ts` — I'd recommend against it.**

The plugin could read it there, since it already calls `loadGemiConfig`. But that only fixes the bundled path: server code, jobs, emails and tests never see `gemi.config.ts`, so they would keep falling back in declaration order. That breaks the property the design leans on and states in the code — *the plugin is purely an optimization; semantics are identical with or without it* — and the drift would only surface in half-translated dictionaries read from the server, which is the case nothing tests. It would also be a second home for a fact that already lives in `app/config/translation.ts`.

The per-dictionary `sourceLocale` is one line, and only matters for dictionaries that are not uniformly translated. If the repetition becomes annoying, the version that does not break equivalence is a boot-time default the runtime also reads, fed from `translation.defaultLocale` — happy to do that separately.

**`gemi/testing` integration** — `<Page>` needed no changes; it already seeds `i18n.currentLocale`, which is the input `useDictionary` reads. Checking it found a real defect in this branch, now fixed: a handle is created once at module scope and was permanently orphaned by anything that cleared the registry under it (a test's `afterEach`, an HMR boundary), throwing `Unknown dictionary` for a dictionary the caller was holding. Handles now re-assert their registration on read.

Worth knowing for anyone writing tests: the seeding story inverts. `useTranslator` needs its strings handed to `<Page>`; `useDictionary` brings its own, so a test seeds `locale` and nothing else — and it works under a real-browser runner, since a `defineDictionary` module imports `gemi/client` rather than the server entry. Both APIs compose on one page. Covered in `docs/testing.md`.

**One more thing the publish check turned up.** Staging `.publish/` showed the registry module compiled into two separate bundles — `chunk-jr6vv0f2.js` from the Bun build behind `gemi/dictionary`, and `chunks/ThemeProvider-*.js` from the Vite build behind `gemi/client`/`gemi/testing`. They only share state through the global key. Module scope would give an installed app one registry for the view chunks and another for `useDictionary`, so the hydration payload would come out empty — passing every test in the package and failing only once installed. Added a test that pins it.

**Final state:** 3449 tests, typecheck clean, 28 type tests, lint clean, `build:publish` stages 22 exports all resolved. Rebuilt and re-ran the template end to end after the merge — both locales render with no cross-locale leakage, and the legacy `/about` still renders from `Dictionary.create`.

nstfkc:
The source-language finding is now closed properly rather than left to a `sourceLocale` nobody remembers to write (`d7b9906`).

**The build refuses to guess.** The source language is the first locale of the first entry, which is invariant under reordering exactly when every entry lists the same locale first. Where two disagree, one key added at the top re-points what the whole dictionary falls back to — a change in what language a reader sees, from a diff that reads as an addition. That now fails the build:

```
defineDictionary() cannot infer a source language. Key "onlyTurkish" lists
"tr-TR" first but "title" lists "en-US" first, so the locale every key falls
back to depends on the order they are written in — moving one key would change
what language a reader sees. Pass { sourceLocale: "<locale>" } as the second
argument to pin it.
```

It fires only on that shape. A uniformly translated dictionary is unaffected, and so is a half-translated but consistently authored one — `{ done: {en,tr}, pending: {en} }` still builds, because no ordering of those keys can move the fallback. The rule is drawn at "keys disagree" rather than the coarser "not uniformly translated" for exactly that reason, and there is a test asserting the order-independence directly rather than only asserting the error fires.

Verified end to end, not just in unit tests: adding an unsourced key to the template's dictionary fails `gemi build` with the file named, and the template builds clean once reverted.

**Still not sourced from `translation.defaultLocale`,** and `docs/i18n.md` now explains why instead of leaving it implicit. It answers a different question — which locale a visitor gets when detection fails, which one has no URL prefix — and a file's source language can legitimately differ. The deciding reason is mechanical: the plugin bakes fallback order into each locale chunk at build time, while server code, jobs and tests read the same literal without meeting the bundler. A build-time-only default would make those two disagree, and `defineDictionary` is meant to behave identically either way.

3455 tests, typecheck clean, 28 type tests, lint clean, template builds and runs.

PR #398, opened 2026-08-13
fix(i18n): stop gating the locale switch on an awaited cookie write

nstfkc:
## The bug

`useLocale().setLocale` gated the navigation on the cookie write:

```ts
setCookie(locale).then(() => {
  replace(pathname, { locale, ... });
});
```

If that promise never settles, `.then()` never runs, `replace()` never happens, and changing the language is a **silent no-op** — no exception, so nothing reaches an error reporter. This matches the reported iOS Safari symptom: Safari ships the Cookie Store API from 18.4, so `globalThis.cookieStore.set()` is reached rather than throwing into the `fetch` fallback (pre-18.4 it throws synchronously and the fallback works).

The same shape breaks from the other end too, on any browser: the fallback's `return await fetch(...)` rethrows out of `setCookie`, and there is no `.catch()` on the chain — an offline user gets no navigation plus an unhandled rejection. And on the happy path the switch waits on a full round trip with no timeout.

## The fix

Write the cookie synchronously with `document.cookie`. It can't hang and can't fail asynchronously, so nothing is left to gate the navigation on.

The write stays **before** the navigation deliberately. Switching *to* the default locale produces a URL with no locale segment, and `ViewRouteDispatcher.handleViewRequest` resolves those through `detectLocale` and 302s back to the cookie's locale — `.json` route-data requests included (only `.og` is exempt). So "navigate anyway and let the URL win" is only true for non-default locales; the cookie has to be correct before the route-data request goes out.

The `/set-locale` request is kept, now fire-and-forget, so the documented `onLocaleChange` server hook fires on **every** switch instead of only in browsers without `cookieStore`.

## Tests

`packages/gemi/client/useLocale.test.tsx` — all three fail against the old implementation, pass with the fix:

- navigates even though the cookie request never settles (the iOS shape)
- navigates even though the cookie request fails (offline)
- writes the locale cookie before navigating (default-locale switch)

https://claude.ai/code/session_0126d5QL4Z7DzQgfYSpmUvVw

### DIFF
--- a/packages/gemi/client/useLocale.ts
+++ b/packages/gemi/client/useLocale.ts
@@ -3,15 +3,40 @@ import { useNavigate } from "./useNavigate";
 import { useParams } from "./useParams";
 import { useRouteData } from "./useRouteData";
 
-const setCookie = async (locale: string) => {
-  try {
-    return await globalThis.cookieStore.set("i18n-locale", locale);
-  } catch (err) {
-    return await fetch(`/api/__gemi__/services/i18n/set-locale/${locale}`);
-    // TODO: show unsuported browser error
-    // console.log(err);
+const LOCALE_COOKIE = "i18n-locale";
+const LOCALE_COOKIE_MAX_AGE = 60 * 60 * 24 * 365;
+
+/**
+ * The active locale lives in two places: the URL segment `setLocale` navigates
+ * to, and the `i18n-locale` cookie the server falls back to when the URL has no
+ * segment — which is exactly the case for the default locale. Both have to be in
+ * place *before* the navigation asks for route data, or `Translator.detectLocale`
+ * answers the request with a redirect back to the locale the cookie still names.
+ *
+ * Hence a synchronous `document.cookie` write rather than an awaited one. The
+ * previous implementation awaited `cookieStore.set()` and hung the navigation off
+ * its `.then()`; that promise is reported to never settle on iOS Safari, so the
+ * `.then()` never ran and changing the language became a silent no-op — no
+ * rejection, nothing for an error reporter to pick up. A synchronous write cannot
+ * hang and cannot fail asynchronously, so nothing is left to gate navigation on.
+ */
+function persistLocale(locale: string) {
+  if (typeof document === "undefined") {
+    return;
   }
-};
+  try {
+    document.cookie = [
+      `${LOCALE_COOKIE}=${locale}`,
+      `Max-Age=${LOCALE_COOKIE_MAX_AGE}`,
+      "SameSite=Strict",
+      "Path=/",
+    ].join("; ");
+  } catch {
+    // A sandboxed iframe with no storage access throws here. The navigation
+    // still carries the locale in its URL, so it is better to switch without
+    // remembering the choice than to not switch at all.
+  }
+}
 
 export function useLocale() {
   const { i18n } = useRouteData();
@@ -20,16 +45,24 @@ export function useLocale() {
   const params = useParams();
 
   const setLocale = async (locale: string) => {
+    persistLocale(locale);
+
+    // `onLocaleChange` is a server-side hook, so it needs a request to fire —
+    // but the language switch must not wait on a round trip, nor be cancelled
+    // by one that fails. Fire and forget: the route re-sets the same cookie,
+    // which is already correct locally either way.
+    void fetch(`/api/__gemi__/services/i18n/set-locale/${locale}`).catch(
+      () => {},
+    );
+
     const urlSearchParams = new URLSearchParams(search);
-    setCookie(locale).then(() => {
-      replace(pathname, {
-        locale,
-        // TODO: fix: this conversion is wrong, because there can be multiple
-        // search params with the same name
-        search: Object.fromEntries(urlSearchParams.entries()),
-        params,
-      } as any);
-    });
+    await replace(pathname, {
+      locale,
+      // TODO: fix: this conversion is wrong, because there can be multiple
+      // search params with the same name
+      search: Object.fromEntries(urlSearchParams.entries()),
+      params,
+    } as any);
   };
 
   return [i18n.currentLocale, setLocale] as const;
--- a/packages/gemi/client/useLocale.test.tsx
+++ b/packages/gemi/client/useLocale.test.tsx
@@ -1,0 +1,124 @@
+/** @vitest-environment jsdom */
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { createElement } from "react";
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+
+import { ClientRouterContext } from "./ClientRouterContext";
+import { I18nContext } from "./I18nContext";
+import { RouteStateProvider, type PageData, type RouteState } from "./RouteStateContext";
+import { useLocale } from "./useLocale";
+
+/**
+ * The switch has to survive a cookie write that never answers — that is the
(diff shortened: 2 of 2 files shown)


packages/gemi/client/useLocale.ts
```diff
@@ -20,16 +45,24 @@ export function useLocale() {
   const params = useParams();
 
   const setLocale = async (locale: string) => {
+    persistLocale(locale);
+
+    // `onLocaleChange` is a server-side hook, so it needs a request to fire —
+    // but the language switch must not wait on a round trip, nor be cancelled
+    // by one that fails. Fire and forget: the route re-sets the same cookie,
+    // which is already correct locally either way.
+    void fetch(`/api/__gemi__/services/i18n/set-locale/${locale}`).catch(
```
nstfkc, line 54: **medium — the new `Max-Age` is undone seconds later by this very request.**

`/set-locale/:locale` does `req.ctx().setCookie("i18n-locale", locale)` with no `maxAge`/`expires` (packages/gemi/i18n/I18nRouter.ts:12), and `createCookie` only emits `Max-Age`/`Expires` when passed (packages/gemi/http/requestContext.ts:26). Same name and `Path=/`, so the response's `Set-Cookie` replaces the persistent cookie wholesale — expiry included — with a session cookie.

Scenario: user picks Türkçe, quits the browser, relaunches → no cookie, `detectLocale` falls back to `accept-language`/default. Nothing restores the year in practice: the only route that sets an expiry is `/translations/:locale` (I18nRouter.ts:27), and `fetchTranslations` has no callers outside its own definition (packages/gemi/client/I18nContext.tsx:89). Before this PR the request only fired on browsers without `cookieStore`; now every switch takes the downgrade.

Fix server-side: give `/set-locale` the same `expires` as `Lang.setLocale` (packages/gemi/facades/Lang.ts:35).

nstfkc, line 54: **low — rapid switching can leave the cookie naming a stale locale.**

Two `/set-locale` requests are now in flight per double-click and can respond out of order; the losing response's `Set-Cookie` wins.

Scenario: click TR then EN quickly — locally the cookie and URL settle on `en-US` (default → no URL segment), then TR's response lands and rewrites the cookie to `tr-TR`. The next route-data request or reload of `/pricing` 302s to `/tr-TR/pricing`, silently reverting the switch.

Cheapest fix: drop the `setCookie` from the `/set-locale` route (packages/gemi/i18n/I18nRouter.ts:12) — the client owns the cookie now — or abort the previous request.

nstfkc (commented):
The core fix looks right, and I verified it: checked the branch out and ran `bun --bun vitest run client/useLocale.test.tsx` — 3/3 pass here, and swapping in `main`'s `useLocale.ts` makes all 3 fail, so the tests genuinely pin the behaviour. The default-locale argument in the description holds up against the code: `ViewRouteDispatcher.handleViewRequest` (packages/gemi/services/router/ViewRouteDispatcher.ts:700) 302s on `urlLocale === null && locale !== defaultLocale`, and `.json` route-data requests are not exempt — so the cookie does have to be correct before the navigation.

Four follow-ups, all about the now-unconditional fire-and-forget request and the raw cookie write rather than the fix itself. Only the first is worth blocking on.

packages/gemi/client/useLocale.ts
nstfkc, line 54: **low — a synchronous throw here still skips the navigation.**

`fetch` is called before `replace` and outside any `try`, so if `globalThis.fetch` is missing (old WebViews, some jsdom/test setups) the call throws synchronously, `setLocale` rejects, and `replace` never runs — exactly the silent no-op this PR fixes, via a different door. The new tests stub `fetch`, so they can't catch it.

Fix: move the call after `await replace(...)`, or wrap it in `try {}`.

packages/gemi/client/useLocale.ts
```diff
@@ -3,15 +3,40 @@ import { useNavigate } from "./useNavigate";
...
+    return;
+  }
   try {
-    return await globalThis.cookieStore.set("i18n-locale", locale);
-  } catch (err) {
-    return await fetch(`/api/__gemi__/services/i18n/set-locale/${locale}`);
-    // TODO: show unsuported browser error
-    // console.log(err);
+    document.cookie = [
+      `${LOCALE_COOKIE}=${locale}`,
```
nstfkc, line 29: **low — `locale` is interpolated raw into `document.cookie`.**

`cookieStore.set()` used to reject malformed values; a raw string write does not.

Scenario: an app wires the switcher to untrusted input (`setLocale(searchParams.get("lang") ?? "en-US")`) and a value like `en-US; Domain=.example.com` or `x; Path=/admin` injects cookie attributes, while a value containing a space or `=` writes a malformed cookie. Neither end validates — `/set-locale` also passes `req.params.locale` straight through (packages/gemi/i18n/I18nRouter.ts:11).

Fix: `encodeURIComponent(locale)`, or check membership in `i18n.supportedLocales` before writing.

packages/gemi/client/useLocale.ts
nstfkc: Fixed in 6290ad4, though by removing the write rather than giving it an expiry.

You are right about the downgrade and I confirmed the dead end you describe — `fetchTranslations` really has no callers, so nothing ever restores the year. But once the client has to write the cookie before it navigates (which the default-locale redirect forces), a second write on the response can only ever race the first. Adding `expires` would fix the persistence and leave the ordering problem below intact; dropping the `setCookie` fixes both, and makes it unambiguous who owns the cookie.

`Lang.setLocale` still sets it with a year for server-side flows, and it validates against `supportedLocales` on the way — that path is unchanged.

nstfkc: Took the cheapest fix — the `setCookie` is gone from the route in 6290ad4, so the double-click race has nothing left to lose. Aborting would not have closed it anyway: a response already in flight applies its `Set-Cookie` whether or not the client is still listening.

Worth noting the route was named for the thing it no longer does. I left the path alone to keep the diff to the bug, but `/locale-changed` would describe it better now that `onLocaleChange` is its whole job — happy to rename if you want it in this PR.

nstfkc: Fixed in 6290ad4. Wrapped rather than moved: the call now sits in a `notifyLocaleChange` helper with its own `try {}`, so a missing `fetch` cannot reach the navigation. Moving it after `await replace(...)` would have left the hook hostage to the navigation instead, which is the same coupling in the other direction.

Test added — `navigates on a runtime with no fetch at all` stubs the global to `undefined`, and it fails against 4bcd600.

packages/gemi/client/useLocale.ts
nstfkc: Fixed in 6290ad4 with `encodeURIComponent`, on both the cookie value and the URL path (a locale carrying `/` or `?` would reshape the request just as readily). Chose encoding over a `supportedLocales` membership check because the check would silently swallow a `setLocale` call rather than fail visibly, and gemi has no channel here to complain through.

The `/set-locale` end you flagged is moot now that the route no longer writes a cookie — `req.params.locale` only reaches `onLocaleChange`, where it is the app own hook receiving the app own input. `Lang.setLocale` remains the validating path.

Test added: `keeps a malformed locale inside the cookie value` round-trips `en-US; Domain=.example.com` through the cookie and fails against 4bcd600, where the value was truncated at the `;`.

PR #400, opened 2026-08-13
feat(useQuery): revalidate on window focus

nstfkc:
Adds `revalidateOnFocus` to `useQuery` — when the tab comes back to the
foreground, the query revalidates.

```tsx
const { data } = useQuery("/todos", {}, { revalidateOnFocus: true });
```

It is also an app-wide default, threaded like the other `queryConfig` keys:

```tsx
createRoot(RootLayout, { queryConfig: { revalidateOnFocus: true } });
```

## Semantics

- **Off by default** — nothing changes for existing apps.
- **`staleTime` gates it.** The listener calls the same `getVariant` read the
  mount effect performs, so a quick tab-out-and-back costs nothing and only
  data older than its freshness window goes back to the wire. `staleTime:
  Infinity` disables it, `staleTime: 0` makes every focus refetch.
- **Silent.** The refetch never flips `loading` or suspends — the data on
  screen keeps rendering until the new payload lands.
- **Two events.** `focus` (window switch, document stays visible) and
  `visibilitychange` (tab switch, device unlock) neither of which covers the
  other; both land on the same `getVariant` call, so the overlap is a no-op —
  the second read finds the fetch already in flight.
- A `focus` fired while the document is hidden, and a `lazy` query that was
  never triggered, are both left alone.

## Tests

New `packages/gemi/client/useQuery.revalidateOnFocus.test.tsx` (8 tests): off
by default, focus revalidates without dropping the rendered rows,
`visibilitychange` in both directions, the `staleTime` gate, untriggered
`lazy`, the provider-level default, a per-call `false` beating it, and
listener teardown on unmount. Full client suite (18 files / 178 tests),
`tsc --noEmit` and `oxlint` all pass.

Docs: `docs/data-fetching.md` config table + a paragraph on the semantics.

### DIFF
--- a/packages/gemi/client/useQuery.ts
+++ b/packages/gemi/client/useQuery.ts
@@ -42,6 +42,14 @@ interface Config<T> {
    * back to the wire — silently, keeping what's on screen rendered.
    */
   revalidateOnFocus?: boolean;
+  /**
+   * Minimum gap between two `revalidateOnFocus` revalidations of the same
+   * query, in ms (default 5000). `focus` fires on far more than tab returns —
+   * dismissing a file picker, closing a popup, clicking in and out of an
+   * embedded iframe — and `staleTime: 0` would otherwise turn each of those
+   * into a request per mounted query.
+   */
+  focusThrottleInterval?: number;
   /** How long cached data stays fresh before a read revalidates it, in ms. */
   staleTime?: number;
   debug?: boolean;
@@ -66,6 +74,7 @@ const defaultConfig: Config<any> = {
   retryIntervalOnError: 10000,
   refreshInterval: 999999,
   revalidateOnFocus: false,
+  focusThrottleInterval: 5000,
   staleTime: DEFAULT_STALE_TIME,
   debug: false,
   lazy: false,
@@ -275,6 +284,18 @@ export function useFrameworkQuery<T extends keyof GetRPC>(
   );
   const refetchUntilDurationRef = useRef(0);
   const prefetchedRef = useRef(false);
+  // Focus revalidation eligibility, deliberately *not* `fetchedRef`: that one
+  // is flipped by `prefetch()` too, and a hover prefetch is documented as
+  // "fetch once" for data the user may never look at — enrolling it in focus
+  // revalidation would keep re-requesting it for the life of the component.
+  // Only the reads that put data on screen (`trigger`, `refetch`, `mutate`,
+  // or simply not being lazy) count.
+  const focusEligibleRef = useRef(!lazy);
+  // The away latch: `focus` fires for things that never left the page —
+  // dismissing an `alert`, returning from devtools — so a revalidation is only
+  // owed when the window actually lost focus or the tab was hidden first.
+  const wasAwayRef = useRef(false);
+  const lastFocusRevalidationRef = useRef(0);
 
   const subscribe = useCallback(
     (onStoreChange: () => void) => resource.store.subscribe(onStoreChange),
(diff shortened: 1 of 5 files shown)


packages/gemi/client/useQuery.ts
```diff
@@ -446,6 +454,35 @@ export function useFrameworkQuery<T extends keyof GetRPC>(
...
+  // the overlap a no-op — the second read finds the fetch in flight.
+  useEffect(() => {
+    if (!config.revalidateOnFocus) return;
+    const revalidate = () => {
+      // A lazy query that was never triggered stays untouched, and a
+      // `focus` fired while the document is hidden isn't a return to the
+      // foreground.
+      if (!fetchedRef.current) return;
+      if (document.visibilityState === "hidden") return;
+      resource.getVariant(variantKey, configRef.current.staleTime);
```
nstfkc: The focus refetch is not always silent. `getVariant` only takes the silent path (`resolveVariant(vk, true)`) when the variant already `hasData`; when the cached variant exists without data — the state a failed fetch leaves behind under `suspense: false` — it falls into `if (!variant.hasData) this.resolveVariant(variantKey)`, i.e. `silent = false`, which writes `loading: true` into the cache.

Repro (verified against this branch): a `suspense: false` query whose first fetch returns 500 renders `data:none loading:false err:yes`; tab away, come back past `staleTime`, and any re-render while the focus refetch is in flight renders `data:none loading:true err:yes`.

The paragraph this PR adds to `docs/data-fetching.md` promises the opposite ("The refetch is silent ... no `loading` flip"). Either narrow the doc claim to the has-data case, or make the handler read through a path that stays silent.

packages/gemi/client/useQuery.ts
```diff
@@ -446,6 +454,35 @@ export function useFrameworkQuery<T extends keyof GetRPC>(
...
+  // switching tabs (or unlocking the device) fires `visibilitychange` without
+  // necessarily firing `focus`. Landing on the same `getVariant` call makes
+  // the overlap a no-op — the second read finds the fetch in flight.
+  useEffect(() => {
+    if (!config.revalidateOnFocus) return;
+    const revalidate = () => {
+      // A lazy query that was never triggered stays untouched, and a
+      // `focus` fired while the document is hidden isn't a return to the
+      // foreground.
+      if (!fetchedRef.current) return;
```
nstfkc: `fetchedRef` is not only set by `trigger()` — `prefetch()`, `refetch()` and `mutate()` all set it too. So a `lazy` query that was merely *prefetched* (the documented hover case: "Fetch once, eagerly, without subscribing to loading state") becomes eligible for focus revalidation forever after, even though the user never opened it and its data is not on screen.

Repro (verified against this branch): `useQuery("/menu", {}, { lazy: true, revalidateOnFocus: true })`, call `prefetch()` once, let it resolve, advance past `staleTime`, dispatch `focus` → a second request goes out.

That contradicts both the comment right here ("A lazy query that was never triggered stays untouched") and the `prefetch()` row in `docs/data-fetching.md` ("Fetch **once**"). If the intent is trigger-only, this needs a separate ref that `prefetch` does not flip.

packages/gemi/client/useQuery.ts
```diff
@@ -446,6 +454,35 @@ export function useFrameworkQuery<T extends keyof GetRPC>(
...
+  // `staleTime`, so this is the same read the mount effect performs: data
+  // still inside its freshness window is left alone, and anything older
+  // refetches *silently*, so what's on screen never flashes a loading state.
+  //
+  // Both events are listened for because neither covers the other: switching
+  // between windows fires `focus` while the document stays visible, and
+  // switching tabs (or unlocking the device) fires `visibilitychange` without
+  // necessarily firing `focus`. Landing on the same `getVariant` call makes
+  // the overlap a no-op — the second read finds the fetch in flight.
+  useEffect(() => {
```
nstfkc, line 499: `staleTime` is the only throttle here, and `window`'s `focus` fires on far more than tab returns: dismissing a native `alert`/`confirm`/file picker, closing an OAuth or payment popup, clicking into an embedded iframe (Stripe, YouTube) and back out, or returning from devtools. Every one of those runs this handler for every mounted query.

With the combination the docs recommend two paragraphs above — `staleTime: 0` ("always revalidate") plus app-wide `revalidateOnFocus: true` — a dashboard with a dozen queries fires a dozen `/api` requests on each of those interactions, several times per checkout flow. SWR keeps a separate `focusThrottleInterval` (5s) precisely because `dedupingInterval`/staleness is not enough of a guard.

Worth either a minimum interval between focus revalidations or a latch that only revalidates when the document was actually hidden/blurred since the last one.

packages/gemi/client/useQuery.ts
nstfkc: Correct, and the doc claim was the one I wanted to keep — fixed in 79d9c9e by adding `QueryResource.revalidate`: `getVariant`'s staleness gate, but `resolveVariant(vk, true)` on every path, so a variant sitting on a failed fetch is retried without `loading: true` ever entering the cache. Focus revalidation is the only caller.

Test: `silent even for a variant sitting on a failed fetch`. It forces a re-render while the retry is in flight, because a non-silent write mutates the map without calling `store.next()` — my first attempt at this test passed against `getVariant` for exactly that reason and was not actually pinning anything.

packages/gemi/client/useQuery.ts
nstfkc: Right — trigger-only was the intent. 79d9c9e adds `focusEligibleRef`, initialised to `!lazy` and flipped by `trigger()`, `refetch()` and `mutate()` but deliberately not by `prefetch()`, so "fetch once" stays once. `fetchedRef` keeps its existing meaning for the mount and refresh-interval effects.

Test: `prefetch() does not enrol a lazy query; trigger() does` — prefetch, resolve, age past `staleTime`, return → no request; then trigger and repeat → one.

packages/gemi/client/useQuery.ts
nstfkc: Took both suggestions in 79d9c9e, since neither covers the other:

- **Away latch** — `blur` and `visibilitychange`-to-hidden arm it; `revalidate` bails unless it is armed. That kills the cases where the page never lost focus at all (dismissed `alert`, closed devtools).
- **`focusThrottleInterval`**, default 5000ms like SWR's, per call or app-wide. This is what bounds the iframe case, where the blur/focus pair is genuine and the latch would let every one through. When a return is throttled the latch stays armed, so the next one past the interval still revalidates rather than being swallowed.

Tests: `a focus with nothing to return from is ignored` and `focusThrottleInterval spaces out repeated returns` (the latter under `staleTime: 0`, so the throttle is the only guard left). Documented in the `data-fetching.md` paragraph, which now lists all three guards.

PR #402, opened 2026-08-14
Features: attribution in code, one switch in the database

nstfkc:
Features are **declared in code and switched on from the database**. `app/features` owns each feature's key, its targeting and how far it has rolled out; a `FeatureFlag` row owns one thing — whether it is switched on at all. Shipping a feature is a deploy; turning it on is not.

```typescript
// app/features/index.ts
import { defineFeature } from "gemi/services";

export default {
  "new-checkout": defineFeature({ describe: "Rebuilt checkout flow" }),

  "pricing-redesign": defineFeature({
    rollout: 50,
    when: (ctx) => {
      if (ctx.user?.email.endsWith("@acme.com")) return true; // staff, always
      if (ctx.user?.plan === "free") return false;            // never
      // return nothing → the rollout decides
    },
  }),
};
```

```tsx
useFeature("pricing-redesign"); // boolean
```

```typescript
if (await Features.enabled("new-checkout")) { ... }
```

```prisma
model FeatureFlag {
  id       Int    @id @default(autoincrement())
  publicId String @unique @default(cuid())
  key      String @unique
  active   Boolean @default(false)
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
}
```

That is the whole schema. Nothing per-subject is stored.

## What changed since the first draft of this PR

The original design put the targeting in the database — a `rules` JSON column with sixteen condition operators, segments, weighted variants, per-row seeds and `bucketBy` — and required a normalization layer because that column is the one part of the system nobody reviews before it reaches production.

This revision moves the targeting into code and leaves the database a single boolean. A closure is expressive by construction and cannot be malformed, so the entire apparatus that existed to make a JSON column safely expressive is gone: `conditions.ts`, `normalize.ts`, the operator set, segments, seeds, `bucketBy`, and `FeatureRouter` itself.

**Net: +1,500 / −3,366 across 47 files.**

### `FeatureRouter` is gone

It was a router that routed nothing — no request, no dispatch, no handler. It wore a router's shape because `CreateFeatures<T>` needed a tree to walk, and the nesting cost more than it bought:

- Keys were assembled at runtime from a prefix and a leaf, so `"billing/new-invoices"` appeared **nowhere in the source**. The most common thing anyone does with a feature — asking whether it is still referenced, so it can be deleted — stopped being a grep.
- The `"a/" + "b"` vs `"a/b"` collision class existed only because nesting existed. A plain object makes it a compile error.
- `this.boolean()` never touched `this`; the factories lived on the class purely to mimic `this.view()`.
- The "never annotate `features`" footgun could only be documented, never enforced. `defineFeature`'s signature removes it.

`CreateFeatures<T>` collapses to `{ [K in keyof T]: boolean }` — no recursive walk, no `__brand`, no `KeyAndValue`, and no `InstanceType` in the augmentation.

### Every feature is a boolean

`FlagValue`, `variant()`, weighted variants and `FeatureValueOf` are removed.

A feature is a thing you turn on. A config value with three settings wants a different lifecycle — versioned, reviewed, deployed — than something you flip at 2am, and a three-arm experiment built from two booleans that can both be true is worse than not having one. If real experimentation is wanted later, it comes back additively.

### Rollouts are computed, not stored

`sha1(salt:subject)` reduced to a bucket in `[0, 10000)`, compared against the threshold. Nothing is written anywhere, which buys:

- **Stability** across devices, processes and deploys, with no row and no cookie holding assignments.
- **Monotonicity** — `bucket < threshold`, so 10% → 25% only ever *adds* people.
- **Independence** — salted by the feature key, so two 20% rollouts pick two different slices.
- **No growth** — one row per feature, not one per feature per user.

SHA-1 rather than `Bun.hash`, which is seeded and not a documented stable contract across versions: a runtime upgrade would silently re-bucket every subject mid-rollout, and nobody would find out from a stack trace. `bucket.test.ts` pins exact digests for exactly this.

## Resolution order

1. Store never loaded → **off** (fails closed)
2. No row, or `active = false` → **off** — `when` and `rollout` are not consulted
3. `when(ctx)` returned a boolean → that
4. No `rollout` → **on**, crawlers included — nothing to sample, so every visit already resolves the same way
5. A crawler, on a feature that *has* a rollout → **off**
6. Inside the bucket → on, else off

**`active` short-circuits everything.** A kill switch application code could defeat is not one.

**`when` outranks `rollout`.** A rollout is a statement about strangers; `when` is a statement about someone you can name. Staff overrides have to beat the dice. Returning nothing abstains.

**A throwing `when` reads off, logged.** `ctx.user.plan` without the `?.` would otherwise white-screen every anonymous page load. Off rather than abstain, because falling through to the rollout would ship the feature to a slice of users nobody approved.

**An unreachable database that has never loaded fails closed.** Previously it served declared defaults. Turning features *on* at the moment the operator has least ability to turn them off is the wrong direction to fail in. The store still keeps its last good snapshot across a failed refresh.

## Fixes `session_id` being `SameSite=Strict`

This one was a real bug, and it broke the exact population the cookie exists for.

`Strict` withholds the cookie on cross-site **top-level navigation**. So a visitor arriving from a search result, a shared link or a newsletter arrived without it, was minted a new id, and overwrote the old one — re-bucketing on every external entry, which is how most anonymous traffic arrives.

`Lax` sends on top-level GET navigations. Safe here because this is an opaque bucketing id, not a credential, and `Lax` still blocks cross-site POST. `access_token` stays `Strict`.

Two related operational notes are now documented:

- **A shared cache must never store a response carrying `Set-Cookie: session_id`.** One visitor's id handed to everybody collapses the whole anonymous population into a single bucket.
- **Crawlers are pinned off** for any feature with a rollout. Bots discard cookies, so each crawl would otherwise land in a fresh bucket and index a different branch than the last one. Uses the existing `isBotUserAgent`.

## Still true from the original design

**Eager, server-side evaluation.** Every feature resolves once per request and the key→boolean map rides the SSR payload, refreshed on each navigation envelope. `useFeature` is a context read: no fetch, no suspense, no flash of the wrong branch. Lazy isn't available — the payload is written as `bootstrapScriptContent` before any component renders.

**Only booleans reach the browser.** Never the targeting, never the reason a feature resolved the way it did — `reason` distinguishes "targeted by name" from "landed in the rollout", which is a fact about the viewer. `serverOnly: true` keeps a feature out of the payload entirely, for when the feature's *name* is the secret.

**Route gating** renders a real 404 rather than a 403, because a 403 confirms the route exists. See the caveat above for what may safely gate a route.

**The three prerequisite fixes** still ride along: the SSR payload XSS hole (`JSON.stringify` doesn't escape `<`), `session_id` being minted too late to bucket with, and `.middleware("cache:private")` silently iterating a bare string one character at a time.

## Verification

- **3,488 framework tests + typecheck + lint (0 errors)** pass.
- **Template:** 335 tests pass; type tests 123/123, no type errors. Three `differential` suites can't run in this worktree — `@prisma/client` isn't installed — and they don't touch `FeatureFlag`.
- **Typing verified end to end** with a throwaway probe: `useFeature("new-checkout")` resolves to `boolean`, and `useFeature("pricing-redesgin")` / `Features.enabled("nope")` both fail to compile.
- `bucket.test.ts` asserts monotonicity across 300 subjects at seven rollout levels, distribution within 1.5% at seven percentages, and salt independence.

## Review round (`2c69db9`)

Four real defects, two documentation errors, one caveat that stays a caveat.

**A configured `source` was thrown away.** The provider compared the *merged* `model` against the source's model name — and `model` always has a value, defaulting to `"FeatureFlag"`. An app configuring only `source: new DatabaseFeatureFlagSource("Flags")` had that source replaced by one pointed at a table it never mentioned, and every feature read off with nothing to explain why. The rebuild now keys on what the application actually wrote: `model` alone repoints the default source, an explicit `source` always wins, disagreeing values warn.

**A gated route rendered its real view on a client-side navigation.** The gated path keeps its manifest entry, so `is404` scheduled `views: ["404"]` and then fell through to a second `setRouteState` whose `...routerState` spread put the gated route's own views back — same batch, so the 404 never reached the screen. `.feature()` 404'd on a hard load and mounted the real view on a `<Link>`. Now returns. Both 404 paths also reset `x-gemi-from` to `""`, since after a 404 commit the mounted tree is `["404"]` and carries none of the segments a stale value would have the server skip.

**`onEvaluate` fired twice per key on every request that rendered.** The memo lived in `explain()`, above the evaluation, so `forClient()` and `passesFeatureGates` both went around it — two exposures for one viewer, against a documented contract of once per key per request. Memo and notification moved down into `evaluateIn`, gated on the context being the ambient request's own by identity, so `Features.for(...)` still shares nothing with the current viewer.

**`useFeature` accepted `serverOnly` keys**, then warned "declare it in `app/features`" — which the app had — and read `false` forever. `Feature` now carries `serverOnly` in its type, and `ClientFeatureKey` filters those keys out of `useFeature` while the facade keeps accepting them. Still one augmentation line in `gemi.d.ts`.

**A gated 404 shipped the gated route's i18n namespace.** The gate cleared `currentPathName` but not `httpRequest.routePath`, and it is the latter that selects the dictionary. For an unannounced feature the translation keys are usually the most descriptive thing about it.

Documentation: the resolution order above had the crawler check before the no-rollout check, inverting the implementation; and `.feature()`'s jsdoc pointed at a `.serverOnly()` method that does not exist.

**A rollout is not access control**, and this is documented rather than fixed. The bucketing subject for an anonymous visitor is their own `session_id` cookie, unvalidated, and bucketing is a published pure function of it. Signing the cookie would not close it: a visitor can discard it and collect fresh ones until one lands inside — a couple of dozen requests for a 5% rollout. Nothing that hands out identities on demand can restrict anything to a fraction of them. So: gate on `active`, or on a `when` that reads something the visitor cannot choose.

Two of these have no regression test, because neither code path has a harness: the `ClientRouter` navigation callback and the dispatcher's request pipeline. The other five do — 9 new tests, including the `enabled` + `forClient` double-exposure the review found, and a type test that can only live in the template, since `packages/gemi` sees an empty `Features` where every key degrades to `string` and the assertions pass vacuously.

## Merged with main at 0.57.0 (`f0bdfef`)

The branch was cut at 0.52.0 and thirty-six commits behind. Three files
conflicted; one of them mattered.

**`gemi.d.ts` moved into the package** (#414), and with it every `@/app/*`
import became optional — a program with no application in it compiles that
file too, and an unresolved import is `any`, which `CreateRPC<any>` does not
survive. `Features` now follows the same shape as `RPC`, `ViewRPC` and
`I18nDictionary`: an `AppFeatureMap<T>` alias whose conditional wraps the
instantiation rather than guarding at the use site. `app/features` is the one
of the four an app can simply not have, and it degrades the same way — the
interface stays empty and `FeatureKey` falls back to `string`.

That also settles a follow-up listed below as open: `gemi.d.ts` ships in the
tarball as of 0.56, so the feature typing reaches a real `npm install gemi`
along with everything else. **It is struck from the list.**

**`gemi/testing` needed wiring, not just resolving.** `<Page>` landed on main
in 0.53 and builds `RouteState & PageData` and `ServerDataContextValue` by
hand — both of which this branch gives a required `features` field. A compile
error rather than a conflict, and the narrow fix (an empty record) would have
left every view under test reading every feature as off with no way to say
otherwise. So `<Page features={{ "new-checkout": true }}>` seeds the outcome.
Not the declaration: re-deriving one there would mean assembling a request
context — a user, a `session_id`, a bot verdict — to assert a branch that only
cares which way the feature came out.

`kernel/providers.ts` was a rewrap with a stale count in it — seventeen
providers for eighteen services now. `docs/llms-full.txt` had a page appended
on both sides; Features sits where `llms.txt` and `docs/README.md` list it,
between Services and Facades. The `docs/README.md` entry still described the
*first* draft — `FeatureRouter` declarations, database-controlled rules — and
now matches.

### Verification after the merge

3,900 framework tests across 160 files. Typecheck clean, including the
generated ORM artifact. `oxlint` 0 errors, `orm/` still at zero warnings.
Packaging test passes, so the augmentation survives declaration emit and an
installed app typechecks against the tarball.

Template: 895 tests across 24 files — **the three `differential` suites
included** this time, by generating a Prisma client and pushing the schema
rather than skipping them, since this branch adds a model. 131 type tests, no
type errors. `prisma generate` produces a zero-line diff against the committed
`app/models/generated`, so that artifact matches main's generator rather than
0.52's. `gemi check models` reports 17 registered models.

### CI is red, and not for a reason in this branch

`app/models/raw-sql.test.ts` fails two tests on the sqlite job:

```
TypeError: Binding expected string, TypedArray, boolean, number, bigint or null
```

**`main` fails them identically** — run
[32487089167](https://github.com/nstfkc/gemi/actions/runs/32487089167), a
`workflow_dispatch` on `2051b82` with nothing from this branch in it. The
cause is Bun 1.4.0, released since main last ran green on the 18th, and the
workflow pins `bun-version: latest`. Bun's SQLite driver now **throws** on an
object binding where it used to bind it:

| bun | binding a plain object into a tagged-template query |
| --- | --- |
| 1.3.14 | binds, matches nothing, 0 rows |
| 1.4.0 | `TypeError` |

The guard those two tests cover still holds — a forged fragment is bound, not
spliced — and 1.4.0 arguably holds it harder, by refusing rather than binding
something meaningless. It is the assertion that is stale, and it belongs on
`main` rather than in a feature branch.

## Follow-ups, deliberately not here

- **Ramping a rollout needs a deploy**, since `rollout` lives in code. If that bites, the fix is one nullable `rollout` column overriding the declared value; nothing else in the design moves.
- ~~**`packages/gemi/gemi.d.ts` isn't published.**~~ Fixed on main by #414; see the merge notes above.
- Route gating gates the response, not discoverability: a gated route's path stays in the client route manifest.
- No cross-instance cache invalidation; propagation is TTL-bounded. Redis pub/sub is the natural follow-up.
- Signing up moves a visitor across rollout boundaries, since the subject changes from the cookie to the account. Inherent — there is no id both stable per person and known before they have one.

https://claude.ai/code/session_01PLnMZn4cLsKBNoGR5Rq3FC


packages/gemi/services/features/FeaturesServiceProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+        featuresConfigDefaults(),
+        this.app.config.get<FeaturesConfig>("features", {}),
+      );
+
+      // The default source is constructed by `featuresConfigDefaults()` with the
+      // default model name, so a `model` override would otherwise be ignored
+      // unless the app also replaced the source. Rebuild it here instead of
+      // making every app that renames the table also know about the source.
+      if (
+        config.source instanceof DatabaseFeatureFlagSource &&
```
nstfkc: **A custom `source` model name is silently overwritten by the default.**

`config.model` defaults to `"FeatureFlag"`, so for an app that configures only the source — `defineFeaturesConfig({ source: new DatabaseFeatureFlagSource("Flags") })` — this comparison is `"FeatureFlag" !== "Flags"` and the app's source is thrown away and replaced by one pointed at `FeatureFlag`. The app then reads the wrong table (or raises `FeatureModelMissingError`) and every feature stays off, with no indication why.

The rebuild should only fire when the app actually set `model`, e.g. by comparing against `featuresConfigDefaults().model` / checking whether the raw config carried a `model` key, rather than against the defaulted value.

packages/gemi/client/ClientRouter.tsx
```diff
@@ -360,6 +363,10 @@ const Routes = (props: { componentTree: ComponentTree }) => {
             ...routerState,
             appId,
             i18n,
+            // `?? state.features` rather than a bare assignment: an error-path
```
nstfkc, line 392: **A feature-gated route does not 404 on a client-side navigation — it renders the gated view.**

The gated path stays in the client route manifest (as the PR notes), so `routerState.views` is non-empty and the early `views.length === 0` return above doesn't apply. The server answers the `.json` navigation with `is404: true`, which schedules `setRouteState({...state, views: ["404"]})` a few lines up — but that updater is immediately followed by this second `setRouteState`, whose `...routerState` spread reinstates the gated route's own `views`. Both run in the same batch, so the `["404"]` is overwritten before anything renders.

Net effect: `this.view("Beta").feature("beta-access")` 404s correctly on a hard document load, but a `<Link>` navigation to `/beta` mounts the real `Beta` view with empty page data — the opposite of what `.feature()` promises. (The `is404` block pre-dates this PR, but `.feature()` is the first thing that makes `is404 === true` reachable with a non-empty client `views`.) It needs an early `return` after the 404 transition, or `views` has to be forced in the merged state.

packages/gemi/services/features/FeatureManager.ts
```diff
@@ -0,0 +1,224 @@
...
+  async evaluateAllIn(
+    ctx: FeatureContext,
+    buckets: Map<string, number>,
+    options: { clientOnly: boolean },
+  ): Promise<Record<string, boolean>> {
+    const values: Record<string, boolean> = {};
+
+    for (const [key, feature] of Object.entries(this.declared)) {
+      if (options.clientOnly && feature.serverOnly) continue;
+      values[key] = (await this.evaluateIn(key, ctx, buckets)).value;
```
nstfkc, line 170: **`onEvaluate` fires twice per key on the real request path.**

`evaluateIn` is called directly here, bypassing the `store.featureEvaluations` memo that `explain()` maintains — so `notify()` runs again for keys already evaluated earlier in the same request. Two concrete cases, both on every request:

- a handler does `await Features.enabled("x")` (cached), then the dispatcher calls `forClient()` at `ViewRouteDispatcher.ts:933` → `onEvaluate("x", …)` fires a second time;
- any route with `.feature("x")` — `passesFeatureGates` evaluates and caches `x`, then `forClient()` re-evaluates it.

The documented contract (`config.ts`, `docs/feature-flags.md`) is "once per key per request", and an exposure pipeline wired to this will double-count. `FeatureManager.test.ts` only covers `enabled`/`explain` in combination, never `enabled` + `forClient`, so it passes. Routing `evaluateAllIn` through the same cache (when the ctx is the ambient one) would fix it.

packages/gemi/client/useFeature.ts
```diff
@@ -0,0 +1,58 @@
...
+ * `RouteStateContext` is what each navigation replaces — the same reason
+ * `useLocale` reads it. Reading `ServerDataContext` instead would pin every
+ * feature to the values from the initial document for the life of the session.
+ */
+export function useFeature(key: FeatureKey): boolean {
+  // `= {}` because several existing tests provide a bare object as page data,
+  // and an error-path envelope may carry no features at all.
+  const { features = {} } = useContext(RouteStateContext);
+
+  if (process.env.NODE_ENV !== "production" && !(key in features)) {
```
nstfkc, line 42: **`serverOnly` features are typed as valid `useFeature` keys but always warn and always read off.**

`CreateFeatures<T>` maps *every* declared key, including ones marked `serverOnly: true`, so `useFeature("project-nightingale")` compiles. `forClient()` deliberately omits those keys from the payload, so this branch fires and logs `Unknown feature "project-nightingale". Declare it in app/features, or check the key.` on every render — advice that is wrong, since the feature *is* declared. The value is then `false` forever, silently.

Either exclude `serverOnly` keys from `CreateFeatures` (they are statically known from the declaration object) or distinguish the two cases in the message.

docs/feature-flags.md
```diff
@@ -0,0 +1,259 @@
...
+The reason is grep. The most common thing anyone ever does with a feature is ask "is this still referenced, can I delete it?", and that question should be answerable with a text search. A nested registry breaks it: the key you search for exists nowhere in the source, only as a concatenation at runtime.
+
+## How a feature resolves
+
+In order, stopping at the first that applies:
+
+1. **The store has never loaded** → off. An unreachable database fails closed.
+2. **No row, or `active = false`** → off. `when` and `rollout` are not consulted.
+3. **`when(ctx)` returned `true` or `false`** → that.
+4. **A crawler** → off.
```
nstfkc: **The numbered order contradicts both the implementation and line 104 of this same file.**

`evaluateFeature` checks `feature.rollout === undefined` *before* `ctx.isBot` (`evaluate.ts:95-104`), so a crawler hitting a switched-on feature with no rollout gets `true`, not `false`. Line 104 below states this correctly ("A feature with no rollout is served to bots normally"), and the PR description agrees — it is steps 4 and 5 here (and the identical list in `evaluate.ts`'s doc block) that are inverted.

As written, a reader implementing against this spec would expect every feature to be off for bots. Swap 4 and 5.

packages/gemi/http/ViewRouter.ts
```diff
@@ -91,8 +94,27 @@ export class RedirectRoute<Input, Output, Params> {
...
+   * Hides this route behind a feature flag.
+   *
+   * When the flag is off the route behaves as though it does not exist — the
+   * application's own `404` view, with a 404 status — rather than returning an
+   * error page that confirms something is there.
+   *
+   * Returns `this`, so the route's type is unchanged and the path keeps its
+   * entry in `ViewPaths`: `<Link>` and `Redirect.to` stay typed. That also means
+   * the path is still in the route manifest the browser receives, so this gates
+   * the *response*, not the route's existence. Use `.serverOnly()` on the flag
```
nstfkc: **`.serverOnly()` is not an API that exists.** (Same text is repeated on `FileRoute`, `ViewRoute` and `LayoutRoute` — lines 231, 293, 383.)

The flag-level option is `defineFeature({ serverOnly: true })`; there is no method on a `Feature`, and features aren't chainable objects. A reader following this advice gets a `TypeError`. Should read `serverOnly: true` in the declaration.

packages/gemi/http/requestContext.ts
```diff
@@ -101,10 +127,80 @@ class Store {
...
+ */
+export function sessionId(): string | null {
+  const store = RequestContext.getStore();
+  if (!store) {
+    return null;
+  }
+  if (store.sessionId) {
+    return store.sessionId;
+  }
+  const sent = store.req?.cookies?.get(SESSION_ID_COOKIE);
```
nstfkc, line 168: **The bucketing subject for an anonymous visitor is entirely client-supplied.**

`sessionId()` returns whatever `session_id` the browser sent, unvalidated (no signature, no format check) — and that value is the rollout subject in `subjectFor`. Since `bucketOf` is a published pure function of `salt:subject`, anyone can compute a `session_id` that lands inside any rollout and set the cookie themselves (`httpOnly` stops JS from reading it, not curl from sending it).

That's acceptable for measurement, but `.feature()` route gating now builds an access decision on top of it: a route gated on a partially rolled-out feature can be reached by an anonymous visitor who picks a cookie value, which defeats the "unannounced feature stays hidden" case the gate is documented for. Worth either signing the id (HMAC, like `csrf_token`) or documenting explicitly that a rollout must never be the only thing gating a route.

packages/gemi/services/router/ViewRouteDispatcher.ts
```diff
@@ -820,9 +845,33 @@ export class ViewRouteDispatcher {
...
+        // about a route can be made.
+        //
+        // Clearing the match rather than throwing is what produces the real 404:
+        // the framework renders the application's `404` view exactly when
+        // `currentPathName` is null, so a gated route becomes indistinguishable
+        // from one that was never defined — status, view and component tree
+        // included. A `RequestBreakerError` here would instead return a bare
+        // text body, which both looks broken and confirms the route exists.
+        if (featureGates.length > 0 && !(await this.passesFeatureGates(featureGates))) {
+          currentPathName = null;
```
nstfkc, line 895: **A gated route still ships the gated route's i18n namespace with its 404.**

`currentPathName` is cleared here, but `httpRequest.routePath` is not — and it is `httpRequest.routePath` that is passed to `translator.getPageTranslations(locale, ...)` a few lines below (line 892). So the 404 response for `/beta` carries the translation dictionary registered for `/beta`, while an actually-unmatched path carries none. That's an observable difference between "gated" and "never existed", which is the distinction this block exists to erase — and for an unannounced feature the translation keys are usually the most descriptive thing about it.

Clearing `httpRequest.routePath` (or passing `currentPathName` to `getPageTranslations`) closes it.

packages/gemi/services/features/FeaturesServiceProvider.ts
nstfkc: Fixed in `2c69db9`. Confirmed — `model` is always set after merging, so the comparison fired for an app that had configured only a source.

Keyed on the raw config now rather than the merged values: `model` alone repoints the default source, an explicit `source` always wins (it is the more specific statement, and the only one of the two that can name something other than the local table), and setting both to disagreeing values warns rather than resolving silently. `FeaturesServiceProvider.test.ts` covers all four combinations.

packages/gemi/client/ClientRouter.tsx
nstfkc: Fixed in `2c69db9`. Both updaters ran in the same batch and the second one won, exactly as described.

The `is404` branch now returns. It also resets `renderedRouteRef` to `""` — that is the `x-gemi-from` value telling the server which segments are already mounted so it can skip their handlers, and after this commit the mounted tree is `["404"]`, carrying neither the target route's layouts nor the previous route's. `resolvePartialRender` reads the empty string as "carry nothing". The `views.length === 0` guard above had the same stale-`from` problem and got the same one-line fix.

No regression test: there is no harness for the `ClientRouter` navigation callback, and standing one up (routerSubject, `loadRoutePayload`, the module registry) is more than this fix warrants. Flagged in the PR description rather than left implied.

packages/gemi/services/features/FeatureManager.ts
nstfkc: Fixed in `2c69db9`. The memo was one level too high — it lived in `explain()`, so everything that did not go through it evaluated afresh and notified again.

Both the memo and `notify` moved down into `evaluateIn`, which is now the single place a feature is evaluated. Gated on the context being the ambient request's own **by identity**, so `Features.for({ user })` inside a request still neither reads from nor writes to the current viewer's memo — an admin previewing a customer would otherwise poison every subsequent read on that request.

Two tests, both of which failed before the change: `enabled` followed by `forClient` notifies once, and an interleaved `Features.for(...)` neither serves from nor pollutes the memo.

packages/gemi/client/useFeature.ts
nstfkc: Fixed in `2c69db9`, both ways — the exclusion, and the message for the caller who can still reach it untyped.

Excluding the keys needed `serverOnly` to survive into the type, which it did not: `defineFeature` returned a plain `Feature` and the flag was value-only. It is a type parameter now, set by an overload pair rather than a `const` type param, so `defineFeature({ serverOnly: true })` is `Feature<true>`.

`Features` keeps every key — the facade evaluating server-only features is the whole point of the option — but their value type is a marker rather than `boolean`, and `ClientFeatureKey` filters on that. So it stays one augmentation line in `gemi.d.ts` rather than two interfaces the app has to maintain.

The type test lives in the template, because it can only live there: `packages/gemi` sees an empty `Features` where every key degrades to `string` and every assertion passes vacuously. Added a `serverOnly` feature to the starter to have something real to assert against.

docs/feature-flags.md
nstfkc: Fixed in `2c69db9`. Swapped in all three places — this file, the `evaluate.ts` doc block, and the PR description, which had the same inverted order (line 104 was the one that was right).

Both now read: no `rollout` → on, crawlers included, there being nothing to sample; a crawler on a feature that *has* a rollout → off.

packages/gemi/http/ViewRouter.ts
nstfkc: Fixed in `2c69db9`, all four copies. Reads `defineFeature({ serverOnly: true })` now. Left over from an earlier draft where features were chainable.

The same jsdoc gained the rollout-is-not-access-control note — see the reply on `requestContext.ts`.

packages/gemi/http/requestContext.ts
nstfkc: Documented in `2c69db9`, not signed — and I think signing would be the wrong call rather than merely a bigger one.

An HMAC stops a visitor *choosing* a bucket, but not landing in one. They discard the cookie, get a fresh signed id, check whether the gated route answers, and repeat: a 5% rollout falls in a couple of dozen requests, all of them well-formed. The subject is self-assigned whether or not it is signed, because the server hands out identities on demand. Nothing that does that can restrict anything to a fraction of them.

So the honest fix is the second one you offered, stated where somebody hits it — `.feature()`'s jsdoc, `sessionId()`'s, and a section in the guide. Gate on `active` (a feature nobody has switched on is unreachable, which is the case the gate is really for) or on a `when` reading something the visitor cannot choose. A rollout answers "how many people should see this yet"; `when` answers "who is allowed to".

Worth noting the signed-in case is already fine: the subject is the user id, not the cookie.

packages/gemi/services/router/ViewRouteDispatcher.ts
nstfkc: Fixed in `2c69db9`. Clearing `httpRequest.routePath` alongside `currentPathName`, rather than changing the `getPageTranslations` call — the block's job is to erase the route, and `routePath` is the other half of what says which route this is.

Safe at that point: the feature context is built and memoized by the gate check immediately above, so a `when` reading `ctx.request.routePath` has already seen the real value. The only other server-side reader is `RateLimitMiddleware`, which runs earlier.

PR #405, opened 2026-08-14
fix(ssr): splice stream payloads at markup boundaries (#404)

### LINKED ISSUE #404: SSR stream transform splices a queued data-push <script> into a mid-tag / raw-text boundary
## Summary

The SSR streaming transform that injects `__GEMI_STREAM__` / `__GEMI_DICT__` data-push `<script>` chunks flushes them at React stream chunk boundaries. `renderToReadableStream` chunk boundaries are not aligned with HTML tag boundaries, so when a `useQuery` settles while the previous React chunk ended mid-markup, the queued `<script>…</script>` is emitted **inside** an open tag or a raw-text element, producing invalid HTML.

## Where

The transform wired into the streamed `Response` (deminified `R4` in the server chunk). Simplified:

```js
const flush = (ctrl) => { while (queue.length) ctrl.enqueue(encode(queue.shift())); };
// ...
async pull(ctrl) {
  const { done, value } = await reader.read();
  if (done) { flush(ctrl); ctrl.close(); return; }
  if (!wroteShell) { wroteShell = true; write(ctrl, value); onShell(); flush(ctrl); return; }
  flush(ctrl);          // <-- flushes the queue regardless of where the previous chunk ended
  write(ctrl, value);
}
```

`onSettle` pushes a complete `<script>(self.__GEMI_STREAM__=…).push([...])</script>` onto the queue. The script itself is well-formed, but `flush` runs at an arbitrary byte offset, so it can land in the middle of a tag or inside raw-text.

## Reproduction

Serve any page that streams a `useQuery` whose data settles after the shell. Two observed failure modes:

1. **Mid-tag splice** — a `<link rel="modulepreload" href="…"/>` gets split across two reads and the script is inserted between them:

   ```
   <link rel="modulepreload"<script>(self.__GEMI_STREAM__=…).push([…])</script> href="/assets/x.js"/>
   ```

   The parser stays in the tag's attribute state, the swallowed markup renders as visible text at the top of the page, the push script never executes (it is consumed as attributes), and the modulepreload is lost.

2. **Raw-text splice** — the script lands right after a `<style type="text/css">` open tag (or inside a React bootstrap `<script>`), i.e. inside a raw-text element, so it is parsed as CSS/script text and never executes.

Both are deterministic on a page with a settling query; which one occurs depends only on where the chunk boundary falls.

## Suggested fix

Track HTML position across the emitted React bytes with a small state machine (in-tag / comment / declaration / raw-text aware; `<style>` `<script>` `<textarea>` `<title>` are raw-text) and flush the queue only when the emitted output is in the DATA state, so a push always lands between complete elements. Defer, never drop — the stream-end path can still flush unconditionally. A byte-level scan is enough since `<` / `>` are ASCII and never appear as UTF-8 continuation bytes.

## Version

`gemi@0.54.2` (current npm `latest`).

nstfkc:
Fixes #404.

## The bug

React's SSR stream copies into a fixed 2048-byte view and enqueues it when full, so its chunk boundaries fall at arbitrary byte offsets — routinely inside a tag. `injectQueryPayloads` flushed its queue at those boundaries, producing markup like:

```
<link rel="modulepre<script>(self.__GEMI_STREAM__=…).push([…])</script>load" href="/assets/dist-BrOFgeeG2.js"/>
```

Parsed for real (via jsdom), the damage is threefold: the push script is swallowed as `<link>` attributes so the payload never reaches the client, `load" href="…"/>` leaks as text in `<head>` — which implicitly opens `<body>`, so it renders as visible garbage at the top of the page — and the modulepreload is lost. The same splice into a `<style>`/`<script>` body lands the payload in raw text, where it never executes.

`__GEMI_DICT__` dictionary payloads ride the same queue and had the identical exposure.

### `no-stream` routes break deterministically

A settled document (`no-stream`, or a crawler UA) has its entire backlog queued before the pump starts, because `ServerQueryStore.onSettle` replays on subscription. The first-chunk branch forwards chunk 1 and then flushes, so the splice always landed at byte 2048 — not occasionally, every time. `progressiveChunkSize: MAX_SAFE_INTEGER` doesn't help; it controls suspense chunking, not the byte view size.

Confirmed on a production `no-stream` page (`cache-control: public`, no `Vary: User-Agent`), byte-identical for browser and Googlebot UAs, spliced at exactly offset 2048.

## The fix

`htmlInsertionPoint.ts` adds a byte-level scanner tracking the emitted stream's parse state — DATA, in-tag (quoted attribute values included), comment, declaration/bogus comment, and raw text for `script`/`style`/`textarea`/`title`/`xmp`/`iframe`/`noembed`/`noframes`/`noscript`. State persists across chunks, so a token split at any byte is handled, and scanning bytes rather than decoded text is sound because every byte it keys on is ASCII, which can never appear inside a multi-byte UTF-8 sequence.

The injector then splices at the first DATA offset **inside** the chunk, splitting the chunk around it, rather than at the chunk boundary. When the previous chunk ended cleanly that offset is 0 — the old behaviour exactly.

Splicing within the chunk rather than deferring to the next one is what preserves the payload-before-reveal ordering documented in the file header: the earliest safe offset is at worst the end of the single tag straddling the boundary, still ahead of anything else that chunk carries. The "nothing before the shell's first chunk" guarantee also holds — the first chunk is never spliced into; its queue goes out behind it, and only if it ended in DATA. A payload with nowhere legal to go waits for the next chunk, or for the unconditional flush at stream end.

The scanner is deliberately conservative: calling a legal splice point unsafe only defers a payload, while the converse is this bug. Ambiguous constructs therefore resolve to unsafe.

## Verification

- **Replayed the real production bytes.** Stripped the payload out of the affected page to recover React's document, re-chunked at 2048, re-injected. Before: offset 2048, mid-`<link rel="modulepre`. After: offset 2088, immediately after that `<link>` closes. Document otherwise byte-identical.
- `htmlInsertionPoint.test.ts` — 33 tests: every unsafe state, the safe states, one document scanned at every possible split offset, and a chunk boundary inside a multi-byte UTF-8 sequence.
- `streamQueryInjection.splice.test.tsx` — 7 regression tests, including a real `renderToReadableStream` render parsed with jsdom to assert the payload survives as a script element, all 60 modulepreloads survive as elements, and nothing leaked into the body. **All 7 fail against the pre-fix injector** (verified by reverting the source).
- Full gemi suite: 3513 passed, 1 skipped. `tsc --noEmit` clean, `oxlint` 0 errors, no new warnings.

### DIFF
--- a/packages/gemi/services/router/htmlInsertionPoint.ts
+++ b/packages/gemi/services/router/htmlInsertionPoint.ts
@@ -1,0 +1,272 @@
+/**
+ * Tracks where an HTML byte stream *is* — mid-tag, inside a comment, inside a
+ * `<script>` body, or between elements — so the stream injector only splices a
+ * payload script in at a point the parser will read it as markup (#404).
+ *
+ * React's SSR stream copies into a fixed 2048-byte view and enqueues it when
+ * full, so its chunk boundaries fall at arbitrary byte offsets — routinely
+ * inside a tag. Splicing at one produced
+ *
+ *   <link rel="modulepre<script>(self.__GEMI_STREAM__=…)…</script>load" href="…"/>
+ *
+ * where the push script is swallowed as `<link>` attributes (so the payload
+ * never reaches the client), the tail leaks as text, and the modulepreload is
+ * lost.
+ *
+ * This is a byte scanner, not an HTML parser: it answers one question, "is the
+ * output currently in the DATA state", and it only has to be conservative in
+ * one direction. Reporting "not safe" when a splice would in fact have been
+ * legal just defers the payload to the next opportunity; reporting "safe" when
+ * it isn't is the bug. Ambiguous constructs therefore resolve to *unsafe*.
+ *
+ * Scanning bytes rather than decoded text is sound because every byte it keys
+ * on (`<`, `>`, `"`, `'`, `/`, `!`, `-`) is ASCII, and no ASCII byte can appear
+ * inside a multi-byte UTF-8 sequence. A chunk that splits a character mid-way
+ * is therefore harmless.
+ */
+
+const TAB = 9;
+const LF = 10;
+const FF = 12;
+const CR = 13;
+const SPACE = 32;
+const BANG = 33;
+const DQUOTE = 34;
+const SQUOTE = 39;
+const DASH = 45;
+const SLASH = 47;
+const LT = 60;
+const GT = 62;
+const QUESTION = 63;
+
+/**
+ * Elements whose content is raw text or RCDATA — everything until the matching
+ * end tag is text, so a `<script>` spliced inside one is never executed.
+ * `<noscript>` is raw text exactly when scripting is enabled, which is the case
+ * that matters here. `<plaintext>` is deliberately absent: it has no end tag,
+ * so treating it as raw text would stall every later payload, and React never
+ * emits it.
+ */
+const RAW_TEXT_ELEMENTS = new Set([
+  "script",
+  "style",
+  "textarea",
+  "title",
+  "xmp",
+  "iframe",
+  "noembed",
+  "noframes",
+  "noscript",
+]);
+
+type State =
+  | "data"
+  | "tagOpen"
+  | "endTagOpen"
+  | "tagName"
+  | "inTag"
+  | "attrDouble"
+  | "attrSingle"
+  | "bang"
+  | "bogus"
+  | "comment"
+  | "rawText";
+
+function isWhitespace(byte: number): boolean {
+  return (
+    byte === SPACE || byte === TAB || byte === LF || byte === FF || byte === CR
+  );
+}
+
+function isAsciiAlpha(byte: number): boolean {
+  return (byte >= 65 && byte <= 90) || (byte >= 97 && byte <= 122);
+}
+
+/** ASCII lower-case; tag names are compared case-insensitively. */
+function toLower(byte: number): number {
+  return byte >= 65 && byte <= 90 ? byte + 32 : byte;
+}
+
+export interface HtmlInsertionScanner {
+  /**
+   * Consume `chunk`, stopping at the first byte offset where the output is in
+   * the DATA state, and return that offset. Returns `chunk.length` if the whole
+   * chunk was consumed without reaching DATA — the caller then forwards it
+   * whole and asks again with the next one.
+   *
+   * The scanner's state always reflects exactly the bytes it consumed, so a
+   * caller that stops short resumes by calling again with the remainder.
+   */
+  scanToInsertionPoint(chunk: Uint8Array): number;
+  /** Consume `chunk` whole, for bytes forwarded without looking for a splice. */
+  write(chunk: Uint8Array): void;
+  /** Is the output, as of every byte consumed so far, between elements? */
+  isSafe(): boolean;
+}
+
+export function createHtmlInsertionScanner(): HtmlInsertionScanner {
+  let state: State = "data";
+  /** Accumulated name of the tag currently being read. */
+  let tagName = "";
+  let isEndTag = false;
+  /** `"</" + name` for the raw-text element being scanned past. */
+  let closeSequence = "";
+  /** How much of `closeSequence` the raw-text scan has matched so far. */
+  let closeMatched = 0;
+  /** Consecutive `-` bytes: two after `<!` open a comment, two before `>` close one. */
+  let dashes = 0;
+
+  const finishTag = () => {
+    if (!isEndTag && RAW_TEXT_ELEMENTS.has(tagName)) {
+      // A raw-text element's content is text, so nothing may be spliced into
+      // it. `<script/>` does not self-close in HTML — browsers open a script
+      // element either way — so the end tag is what gets us out.
+      state = "rawText";
+      closeSequence = `</${tagName}`;
+      closeMatched = 0;
+    } else {
+      state = "data";
+    }
+    tagName = "";
+  };
+
+  const step = (byte: number) => {
+    switch (state) {
+      case "data":
+        if (byte === LT) state = "tagOpen";
+        return;
+
+      case "tagOpen":
+        if (byte === BANG) {
+          state = "bang";
+          dashes = 0;
+        } else if (byte === SLASH) {
+          state = "endTagOpen";
+        } else if (isAsciiAlpha(byte)) {
+          state = "tagName";
+          isEndTag = false;
+          tagName = String.fromCharCode(toLower(byte));
+        } else if (byte === QUESTION) {
+          // A bogus comment: consumed to the next `>`.
+          state = "bogus";
+        } else if (byte === LT) {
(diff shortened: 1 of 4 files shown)


packages/gemi/services/router/htmlInsertionPoint.ts
```diff
@@ -0,0 +1,272 @@
...
+  let tagName = "";
+  let isEndTag = false;
+  /** `"</" + name` for the raw-text element being scanned past. */
+  let closeSequence = "";
+  /** How much of `closeSequence` the raw-text scan has matched so far. */
+  let closeMatched = 0;
+  /** Consecutive `-` bytes: two after `<!` open a comment, two before `>` close one. */
+  let dashes = 0;
+
+  const finishTag = () => {
```
nstfkc, line 154: **`<template>` content is DATA to the tokenizer but inert to the parser — the payload still gets lost.**

The scanner is a tokenizer-state tracker, but "the parser will execute a `<script>` here" is a *tree-construction* question. Inside `<template>`, the tokenizer is in DATA, so `isSafe()` returns true — but the parser puts the content into the template's `DocumentFragment`, where a script never executes.

This is reachable on the exact code path this PR fixes. React 19 emits `<template id="B:n"></template>` for every Suspense boundary (verified against `react-dom@19.2.3`):

```
<!--$?--><template id="B:0"></template><div>skeleton</div><!--/$-->
```

If a 2048-byte view boundary falls anywhere inside that `<template id="B:0">` start tag (~19 bytes) while the queue is non-empty, `scanToInsertionPoint` returns the offset just past its `>` — which is inside the template. The payload is emitted there, parsed into the fragment, and `self.__GEMI_STREAM__` never receives the entry: same user-visible outcome as #404 (client refetches / hydrates without data), just a narrower window.

Reproduced end-to-end through `injectQueryPayloads` (parse5 as the oracle, counting only non-template scripts):

```js
source.write('<!doctype html><html><head><title>t</title></head><body><main><!--$?--><template id="B');
source.write(':0"></template><div>skeleton</div><!--/$--></main>');
source.write('</body></html>');
// expected 1 executable __GEMI_STREAM__ script, got 0
```

Fix: track template nesting depth alongside the tokenizer state and report unsafe while it is non-zero — increment in `finishTag` when `tagName === "template"` and the tag is not self-closed-in-foreign-content, decrement on `</template>`, and make `isSafe()` require `depth === 0`. This stays on the conservative side the file already commits to.

Worth a regression test too — the fuzz that found this (every scanner-safe offset of 1500 generated documents, spliced and re-parsed) is clean on everything else, so the tokenizer half of the scanner looks solid.

packages/gemi/services/router/htmlInsertionPoint.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Elements whose content is raw text or RCDATA — everything until the matching
+ * end tag is text, so a `<script>` spliced inside one is never executed.
+ * `<noscript>` is raw text exactly when scripting is enabled, which is the case
+ * that matters here. `<plaintext>` is deliberately absent: it has no end tag,
+ * so treating it as raw text would stall every later payload, and React never
+ * emits it.
+ */
+const RAW_TEXT_ELEMENTS = new Set([
```
nstfkc, line 67: **Foreign content (`<math>`) is another tokenizer-DATA-but-not-executable region.**

Same root cause as the `<template>` case, lower blast radius. Inside `<math>…</math>` the tokenizer is in DATA, so the scanner calls those offsets safe — but the parser creates the spliced `<script>` in the MathML namespace, where it is an unknown MathML element and never runs. The payload is silently dropped for that entry.

Concrete scenario: a page renders inline MathML (or any `<math>` block), a query settles while React is streaming, and a chunk boundary lands inside the `<math>` subtree. `injectQueryPayloads` splices the payload there and the client never sees it.

Confirmed by differential fuzzing against parse5 — e.g. splicing at the offset after `<math>` in `…<div hidden id="S:0">seg</div><math><mi>x</mi></math>…` yields a script with `namespaceURI = http://www.w3.org/1998/Math/MathML`.

`<svg>` shows the same namespace divergence but is benign in practice: an SVG `<script>` is an `SVGScriptElement` and browsers do execute it. MathML is the one that actually loses the payload.

A foreign-content depth counter (enter on `<svg>`/`<math>` that are not self-closing, exit on the matching end tag) alongside the `<template>` fix would close both.

packages/gemi/services/router/htmlInsertionPoint.ts
nstfkc: Fixed in f71a5b2 — you're right, and the `<template>` window is live: `react-dom@19.2.3` does emit `<template id="B:n">` per streamed boundary (grepped the shipped `cjs` build to confirm).

The scanner now counts `<template>` and foreign-content (`<svg>`/`<math>`) nesting alongside the tokenizer state, and `isSafe()` requires both depths at zero. Details:

- Foreign roots honour `/>` — `<svg/>` opens nothing — while `<template/>` does not, matching the parser. `/ >` correctly does *not* self-close, so the subtree still opens.
- Depth counters clamp at zero, so a stray `</template>` can't drive one negative and re-open the hole for every later template.
- I sat out `<svg>` too, despite your note that an `SVGScriptElement` does execute. Deferring costs nothing and it keeps the rule one sentence rather than a namespace-by-namespace exception list.

Ordering is unaffected: the scan still takes the first safe offset *inside* the chunk, and a template or foreign subtree always closes before the reveal script that follows it.

Tests: your exact repro is now an end-to-end case in `streamQueryInjection.splice.test.tsx`, counting only scripts jsdom exposes via `querySelectorAll` (which doesn't descend into template content), plus the MathML one asserting `namespaceURI` is XHTML. Both fail against the previous scanner with `expected [] to have a length of 1` — i.e. zero executable scripts, exactly what you found. 20 new scanner cases cover the depth tracking.

Re-ran the property check against the production document from #404: 1305 offsets now reported safe, down from 1405. The 100 dropped are all inside that page's 20 `<svg>` subtrees, and zero of the remaining ones disagree with a spec parser.

packages/gemi/services/router/htmlInsertionPoint.ts
nstfkc: Fixed in f71a5b2, closed together with the `<template>` case by one depth counter each — see the reply on that thread for the mechanics.

Both `<svg>` and `<math>` open the foreign-content counter. I took the conservative option on `<svg>` even though you're right that an `SVGScriptElement` executes: the deferral is free, and "foreign content is not an insertion point" is easier to keep true than a namespace-by-namespace carve-out.

Worth noting your lower-blast-radius call was right for the page in #404 specifically, but not by much — it has 20 `<svg>` subtrees and no `<math>`, and re-running the offset property check shows 100 of its 1405 previously-safe offsets sat inside them. Those were all benign under the old scanner; they just aren't offsets worth arguing about now.

Regression test asserts the injected script's `namespaceURI` is XHTML and that `<math>` contains no script.

PR #406, opened 2026-08-15
Load heavy SDKs lazily so the barrels stay cheap to import

### LINKED ISSUE #403: Importing `CronJob` or `DB` loads `sharp`, `resend`, `@aws-sdk/client-s3` and `satori`, because the barrels are the only export path
Split out of #401, where it turned up as the cause of a mock rather than as a harness gap. It stands on its own: it costs every app's bundle and cold start, not just its tests.

## The graph

`exports` in `packages/gemi/package.json` publishes `"./facades"` and `"./services"` and no subpaths, so a barrel is the **only** door. Neither is tree-shakeable at import time in a test runner, and the package declares no `sideEffects` field, so a bundler has to assume every module in the graph has them.

**`gemi/services`** — five eager re-exports, each a value import of a heavy external:

| `services/index.ts` | pulls |
| --- | --- |
| `:5` `export { S3Driver }` | `@aws-sdk/client-s3`, `@aws-sdk/s3-request-presigner` |
| `:46` `export { ResendDriver }` | `resend` |
| `:56` `export { ViewRouteDispatcher }` | `satori`, `sharp` |
| `:81` `export { Sharp }` | `sharp` |
| `:133` `export { RedisManager }` | `bun` — `import { RedisClient } from "bun"`, a **value** import |

**`gemi/facades`** — two:

| | pulls |
| --- | --- |
| `facades/Storage.ts:2` | `import sharp from "sharp"` |
| `facades/Redis.ts` → `services/redis/RedisManager.ts:1` | `bun`'s `RedisClient` |

`sharp` is a `peerDependency` and is not marked optional, so an app that never touches an image still has to install a native binary to write `import { Lang } from "gemi/facades"`.

## This is the path the template teaches

Not an unusual import — the scaffold every new app is generated from does exactly this:

```ts
templates/saas-starter/app/cron/TestCron.ts:1     import { CronJob } from "gemi/services"
templates/saas-starter/app/jobs/TestJob.ts:1      import { Job } from "gemi/services"
templates/saas-starter/app/http/routes/view.ts:1  import { Cookie, Lang, Meta, Query, Auth } from "gemi/facades"
templates/saas-starter/app/models/transactions.test.ts:6   import { DB } from "gemi/facades"
```

`services/cron/CronJob.ts` is a dependency-free 90-line file. Reaching it costs the AWS SDK, Resend, satori and sharp, because it shares a barrel with them. `app/models/transactions.test.ts` is a database test that loads `sharp`.

## How it shows up

**In tests, acutely.** A runner evaluates every module in the barrel — there is no tree-shaking to rescue it. From #401, a real app's cron test:

```ts
mock.module("gemi/services", () => ({
  CronJob: class { static exp(e: string) { return e } },
}));
```

An app re-implemented the framework's base class, `static exp()` included, because importing it was not viable. The same file does it to the other barrel:

```ts
mock.module("gemi/facades", () => ({ DB: { connection } }));
```

— which silently sets `Auth`, `Redirect`, `Lang`, `Storage`, `Query`, `Broadcast`, `Url`, `Log`, `Meta`, `Cookie`, `Redis` and `RateLimiter` to `undefined` for the module under test. Nothing reports it. The day someone adds a `Log.info()` to that job, the failure is `undefined is not a function` in a test about newsletter segments.

**Under a browser-targeted runner it does not resolve at all.** `import { RedisClient } from "bun"` is the specifier #396 already hit one level down, where `Dictionary` imported the `../facades` barrel and dragged in `Redis`; the fix there was to narrow the import to `../facades/Lang`. That fix is not available to an application, which only has the barrel.

**In production**, an app pays the module graph on every cold start, and ships it in any bundle a conservative tree-shaker can't prove safe to drop.

## Proposed fix

**Lazy-load the heavy drivers, following the precedent already in this repo.** `AzureBlobDriver.ts:107` does exactly this for its optional peer:

```ts
// Memoized so concurrent requests share one client, and one import.
this.serviceClientPromise ??= (async () => {
  let sdk: any;
  try {
    sdk = await import("@azure/storage-blob");
  } catch {
    throw new Error("AzureBlobDriver requires the '@azure/storage-blob' package. …");
  }
```

Applied to the five, this needs **no change to any application's imports** and no new export paths:

- `S3Driver`, `ResendDriver`, `Sharp` — move the SDK import into the first call that uses it.
- `facades/Storage.ts` — `sharp` is a value import at module top used by exactly one method (`metadata()`, `:31`) and one type alias (`:17`). The alias becomes `import type`; the call becomes a lazy import.
- `ViewRouteDispatcher` — `satori` and `sharp` are on the OG-image path only.
- `RedisManager` — the client already connects lazily ("nothing connects until something actually issues a command"); it is the *import* that is eager.

**Add `"sideEffects": false`** to `package.json` so bundlers can drop what an app doesn't reference. Free, and orthogonal to the above — it does nothing for test runners, which is why the lazy imports are the substantive half.

**Not proposed: subpath exports** (`gemi/services/cron`). It would work, but it splits the documented import path in two, invalidates what the template teaches, and leaves the barrel just as expensive for everyone who keeps using it. The lazy imports fix the cost where it is, for imports that already exist.

## Verification this would need

- The publish-shape check in `scripts/build-publish.ts` still passes, and `build:core`'s `splitting: true` still puts the shared modules in one chunk.
- A test that importing `gemi/services` does not evaluate `resend`, `@aws-sdk/client-s3`, `satori` or `sharp` — the guard that keeps this from regressing the next time something is added to a barrel. `AzureBlobDriver.test.ts:412` already asserts the absent-package path, so there is a shape to copy.
- The error an app gets when a driver's package really is missing stays as legible as `AzureBlobDriver`'s.

Related: #401 (server-side test harness), which this blocks the cleanest version of — a `createApp()` an app cannot import is not much of a harness.

nstfkc:
Fixes #403.

## What was wrong

`exports` publishes `./services` and `./facades` and no subpaths, so a barrel is the **only** door an application has to `CronJob`, `Job`, `Command`, `DB` or `Lang`. Five drivers behind those barrels imported their SDK at module scope, so the line the saas-starter template teaches —

```ts
import { CronJob } from "gemi/services";
```

— evaluated `@aws-sdk/client-s3`, `resend`, `satori` and `sharp`, and `import { DB } from "gemi/facades"` loaded a native image binary for a database test. A bundler can sometimes shake that out; a test runner never can, which is why apps end up re-implementing the framework's base classes behind `mock.module`.

## What changed

Each import moves into the first call that needs it. **No application import changes, and no new export paths.**

| | before | after |
| --- | --- | --- |
| `S3Driver` | `@aws-sdk/client-s3` at module scope, client in the constructor | `connect()` imports the SDK and builds the client on first use, memoized with `??=` |
| `ResendDriver` | `resend` at module scope | imported on first `send()` |
| `Sharp`, `Storage.metadata()`, OG images | `sharp` at module scope in three files | one `support/sharp.ts` loader, which names the caller when the peer dependency is genuinely absent |
| `ViewRouteDispatcher` | `satori` at module scope | loaded on the OG-image path only, beside the `react-dom/server` deferral already there |
| `RedisManager` | `import { RedisClient } from "bun"` — a **value** import | reaches `Bun.RedisClient` through the global, so the barrel no longer carries a specifier a browser-targeted runner cannot resolve (#396) |

`Storage.metadata()` loads sharp *outside* its `try`: the empty object it returns means "not an image sharp can read", and a missing install must not be able to hide behind that.

## Measured

On the saas-starter template, `import "gemi/services"` (three cold runs each):

| | wall | RSS |
| --- | --- | --- |
| before | 184 / 192 / 298 ms | ~120 MB |
| after | 42 / 43 / 63 ms | ~60 MB |

## The regression guard

`packages/gemi/barrel-imports.test.ts` walks the **static** import graph of both barrels and fails if any of the four packages becomes eager again, naming the file that reintroduced it.

It reads imports rather than instrumenting a run because Bun offers nothing that observes the second: a runtime plugin's `onResolve` fires for `await import(x)` and **not** for `import x from "…"`, so an instrumented child process reports a clean graph even with the eager import right there — verified before settling on the walk. Two self-checks come with it: one that the walk reached the whole graph (an unresolvable relative specifier would prune a subtree and pass vacuously), and one that it really does count a static import and not a dynamic or type-only one.

Negative control: reverting `ResendDriver.ts` alone makes it fail with `resend <- services/email/drivers/ResendDriver.ts`.

## `"sideEffects"` is deliberately **not** added

The issue proposed it as free. It is not — it breaks our own build.

Under Bun 1.3.14 with `splitting: true`, declaring `sideEffects` (as `false`, or as any list that leaves gemi's own modules out) makes `scripts/build.ts` emit a `dist/services/index.js` that still re-exports `filesystemConfigDefaults` after the chunk dropped the binding:

```
SyntaxError: Exported binding 'Mk' needs to refer to a top-level declared variable.
```

`SharpDriver`'s class body disappears the same way. `["**/*"]` — "everything has side effects", i.e. the same as saying nothing — builds fine, which is what identifies this as elimination rather than a stale artifact. The reproduction is recorded at the top of `scripts/build.ts`. The lazy imports are the substantive half anyway, and unlike `sideEffects` they help test runners too.

## Still eager, deliberately out of scope

- **`twitter-api-v2`**, via `XOAuthProvider` — it builds its `TwitterApi` in the constructor and reads it from a *synchronous* `getRedirectUrl()`, so deferring the import means changing the `OAuthProvider` contract rather than moving a line.
- **`bun`**, via `database/Connection.ts`'s value import of `SQL` — so `gemi/facades` is still Bun-only. The `RedisManager` half of #396 is fixed here; the database half is its own change.

Both are noted in the guard test's scope comment so a passing run is not read as more than it is.

## Verification

- `bun run typecheck`, `bun run lint` (78 warnings, all pre-existing), `bun --bun vitest run` — 140 files, 3539 passed.
- `bun scripts/build-publish.ts` — 22 exports, all resolved; the staged `dist/services/index.js` loads and still hands back `CronJob`, `Sharp` and `filesystemConfigDefaults`.
- saas-starter template suite on the branch's code: 24 files, 883 passed.

https://claude.ai/code/session_01LJQBFgWKiZZrp7sXP8zma4

### DIFF
--- a/packages/gemi/barrel-imports.test.ts
+++ b/packages/gemi/barrel-imports.test.ts
@@ -1,5 +1,15 @@
-import { existsSync, readFileSync, statSync } from "node:fs";
-import { dirname, relative, resolve } from "node:path";
+import { spawnSync } from "node:child_process";
+import {
+  existsSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  statSync,
+  writeFileSync,
+} from "node:fs";
+import { createRequire } from "node:module";
+import { tmpdir } from "node:os";
+import { dirname, join, relative, resolve } from "node:path";
 import { describe, expect, test } from "vitest";
 
 /**
@@ -15,16 +25,21 @@ import { describe, expect, test } from "vitest";
  * a database test (#403). A bundler can sometimes shake that out; a test runner
  * never can.
  *
- * The check is a walk of the *static* import graph rather than an instrumented
- * run, because Bun offers nothing that observes the second: a runtime plugin's
- * `onResolve` fires for `await import(x)` and not for `import x from "…"`, so
- * an instrumented child process reports a clean graph even when the eager
- * import is right there. Reading the imports is the thing that can tell the two
- * apart, and it names the file that introduced one.
+ * Two checks, because neither alone is enough:
+ *
+ *   1. `loadedBy()` really imports each barrel in a child Bun process, with a
+ *      plugin whose `onLoad` watches every module under the heavy packages. It
+ *      answers the question that matters — *was this evaluated* — so it catches
+ *      any eager shape, including a module-scope `await import("sharp")`, which
+ *      is awaited during the importer's own evaluation and is every bit as
+ *      eager as a static import.
+ *   2. `walk()` reads the static import graph. It cannot see that
+ *      top-level-await shape, but when something does go eager it names the
+ *      file that did it, and it runs without the heavy packages installed.
  *
  * Scope, so a passing run is not read as more than it is:
  *
- *   - It watches the packages listed in `HEAVY`. A *new* heavy dependency added
+ *   - Both watch the packages listed in `HEAVY`. A *new* heavy dependency added
  *     to a barrel is not caught until someone adds it here.
  *   - `bun` is not on the list, because it cannot be yet: `database/Connection.ts`
  *     still imports `SQL` as a value, so `gemi/facades` remains Bun-only. The
@@ -45,16 +60,103 @@ const HEAVY = [
 
 const ROOT = import.meta.dirname;
 
+/** Matches any file inside one of the `HEAVY` packages, on either separator. */
+const HEAVY_PATH_SOURCE = `[\\\\/]node_modules[\\\\/](${HEAVY.map((name) =>
+  name.replace("/", "[\\\\/]"),
+).join("|")})[\\\\/]`;
+
+// ---------------------------------------------------------------------------
+// 1. Did importing the barrel actually evaluate any of them?
+// ---------------------------------------------------------------------------
+
+/**
+ * Imports `target` in a child Bun process and returns the `HEAVY` packages that
+ * were loaded while it did.
+ *
+ * `onLoad` rather than `onResolve`: a runtime plugin's `onResolve` is not
+ * consulted for a bare specifier that Bun's module loader resolves on its own,
+ * so it never sees `import sharp from "sharp"`. `onLoad` sees every module that
+ * is actually read, which is the question being asked.
+ *
+ * The hook has to return something, so it returns an empty module. Anything it
+ * matched has already failed the test, and stubbing keeps the first heavy
+ * package from dragging in the rest before the child reports what it found.
+ */
+function loadedBy(target: string): string[] {
+  const dir = mkdtempSync(join(tmpdir(), "gemi-barrel-"));
+
+  try {
+    writeFileSync(
+      join(dir, "probe.ts"),
+      [
+        `import { plugin } from "bun";`,
+        `const HEAVY = /${HEAVY_PATH_SOURCE}/;`,
+        `const seen = new Set();`,
+        `plugin({`,
+        `  name: "gemi-barrel-probe",`,
+        `  setup(build) {`,
+        `    build.onLoad({ filter: HEAVY }, (args) => {`,
+        `      const name = args.path.match(HEAVY)[1].replace(/\\\\/g, "/");`,
+        `      if (!seen.has(name)) {`,
+        `        seen.add(name);`,
+        `        console.log("gemi-loaded:" + name);`,
+        `      }`,
+        `      return { contents: "", loader: "js" };`,
+        `    });`,
+        `  },`,
+        `});`,
+        ``,
+      ].join("\n"),
+    );
+
+    writeFileSync(
+      join(dir, "entry.ts"),
+      `await import(${JSON.stringify(target)});\nconsole.log("gemi-barrel-ok");\n`,
+    );
+
+    // `process.execPath` is the Bun binary the suite already runs under
+    // (`bun --bun vitest`), so this does not depend on `bun` being on PATH.
+    const result = spawnSync(
+      process.execPath,
+      ["--preload", join(dir, "probe.ts"), join(dir, "entry.ts")],
+      { cwd: ROOT, encoding: "utf8" },
+    );
+
+    const stdout = result.stdout ?? "";
+    const loaded = [...stdout.matchAll(/^gemi-loaded:(.+)$/gm)].map((m) => m[1]);
+
+    // A child that died for some *other* reason would otherwise look like a
+    // clean graph. Loading a heavy package is allowed to kill it — the stub
+    // above guarantees that — so this only applies when nothing was found.
+    if (loaded.length === 0 && !stdout.includes("gemi-barrel-ok")) {
+      throw new Error(
+        `Importing ${target} failed in the child process.\n` +
+          `${stdout}\n${result.stderr ?? ""}`,
+      );
+    }
+
+    return loaded.sort();
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+}
+
+// ---------------------------------------------------------------------------
+// 2. Which file introduced it?
+// ---------------------------------------------------------------------------
+
 const transpilers = {
   ts: new Bun.Transpiler({ loader: "ts" }),
   tsx: new Bun.Transpiler({ loader: "tsx" }),
 };
 
 /**
- * `import x from "…"` and `require("…")` both evaluate the target as the
- * importer loads. `import("…")` does not, which is the whole distinction this
- * file exists to hold: every fix for #403 moved a specifier from the first
- * category to the second.
+ * `import x from "…"` and `require("…")` evaluate the target as the importer
+ * loads. `import("…")` *inside a function* does not, which is the distinction
+ * every fix for #403 turns on.
+ *
+ * A module-scope `await import("…")` is eager despite landing in the same
+ * bucket here as the lazy one. `loadedBy()` above is what covers that gap.
  *
  * `import type` never appears here at all — the transpiler elides it — so a
  * type-only reference to `sharp` is correctly not a cost.
@@ -79,7 +181,7 @@ function packageOf(specifier: string): string {
 }
 
 interface Graph {
-  /** Heavy package -> the files that import it eagerly, nearest first. */
+  /** Heavy package -> the files that import it eagerly. */
   heavy: Map<string, string[]>;
   /** Relative specifiers that resolved to no file, which would hide a subtree. */
   unresolved: string[];
@@ -128,29 +230,42 @@ describe.each([
   ["gemi/services", "services/index.ts"],
   ["gemi/facades", "facades/index.ts"],
 ])("%s", (_subpath, entry) => {
-  const graph = walk(entry);
-
-  test("evaluates no heavy SDK when it is imported", () => {
+  test("loads no heavy SDK when it is imported", () => {
     // `CronJob` is a 90-line dependency-free file, and `DB.connection()` is a
     // database call. Reaching either through its barrel must not cost the AWS
     // SDK, Resend, satori, or a native image binary.
-    expect(describeHeavy(graph)).toEqual([]);
+    expect(loadedBy(join(ROOT, entry))).toEqual([]);
+  }, 60_000);
+
+  test("holds no eager import of one, so none is a hoist away", () => {
+    // Redundant with the check above on today's code, and reported separately
+    // because this is the one that can say *which file*.
+    expect(describeHeavy(walk(entry))).toEqual([]);
   });
 
-  test("the walk reached the whole graph", () => {
+  test("is walked in full, so the check above cannot pass vacuously", () => {
+    const graph = walk(entry);
     // A relative specifier that resolves to nothing silently prunes everything
-    // beneath it, and the assertion above would pass because of the hole.
+    // beneath it, and `describeHeavy` would come back empty because of the hole.
     expect(graph.unresolved).toEqual([]);
-    // And a walk that stopped at the barrel itself would also pass. Pinned well
-    // under the ~110 files each entry currently reaches.
+    // And a walk that stopped at the barrel itself would also look clean. Pinned
+    // well under the ≈150 files `services` reaches today, and the ≈105 of
+    // `facades`.
     expect(graph.visited).toBeGreaterThan(50);
   });
 });
 
-describe("the import walk", () => {
-  test("counts a static import and not a dynamic one", () => {
-    // The two assertions above are only worth anything if `walk` can tell an
-    // eager import from a lazy one — every fix for #403 is exactly that edit.
+describe("the guards themselves", () => {
+  test("the child process reports a package that really is loaded", () => {
+    // Without this the assertions above pass just as happily when the plugin
+    // stops matching — a guard that cannot fail is not one. `resend` stands in
+    // for all five: one filter watches them together.
+    const resend = createRequire(import.meta.url).resolve("resend");
+
+    expect(loadedBy(resend)).toEqual(["resend"]);
+  }, 60_000);
+
+  test("the walk counts a static import and not a deferred one", () => {
     const scan = (code: string) =>
       transpilers.ts.scanImports(code).filter((i) => EAGER.has(i.kind));
 
--- a/packages/gemi/barrel-imports.test.ts
+++ b/packages/gemi/barrel-imports.test.ts
@@ -1,0 +1,162 @@
+import { existsSync, readFileSync, statSync } from "node:fs";
+import { dirname, relative, resolve } from "node:path";
+import { describe, expect, test } from "vitest";
+
+/**
+ * `gemi/services` and `gemi/facades` are barrels, and `exports` publishes no
+ * subpaths beneath them — so they are the *only* door an application has to
+ * `CronJob`, `Job`, `Command`, `DB` and `Lang`. Anything a barrel re-exports is
+ * therefore paid for by every app and every test that reaches any of those.
+ *
+ * They used to re-export five drivers that imported their SDK at module scope,
+ * so `import { CronJob } from "gemi/services"` — the line the saas-starter
+ * template teaches — evaluated `@aws-sdk/client-s3`, `resend`, `satori` and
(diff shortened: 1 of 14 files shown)


nstfkc:
## Automated review

_Run with a multi-agent workflow at the author's request: four fresh reviewers with no context from the authoring session, over separate dimensions; every finding then put to two independent verifiers, one of them tasked with refuting it. Nine were filed, nine survived._

**Reviewed by four independent reviewers over separate dimensions (behaviour change, comment/doc accuracy, the new regression guard, the build/packaging story). Every finding was then put to two verifiers, one of them tasked with refuting it; what follows is what survived.**

---

### 1. The new guard cannot see a module-scope `await import("sharp")` — the exact regression shape this PR creates

`packages/gemi/barrel-imports.test.ts:62`

`EAGER = new Set(["import-statement", "require-call"])` classifies purely by `scanImports` kind, so every `import("…")` is treated as deferred. A top-level `await import("…")` is awaited during the importer's evaluation and is exactly as eager as a static import. The comment above `EAGER` states the opposite as an unqualified fact ("`import(\"…\")` does not [evaluate the target as the importer loads]").

This matters because all five fixes in this PR have the form `(await import("sharp")).default`, and hoisting one out of its function is a one-line edit:

- Prepend `const eagerSharp = (await import("sharp")).default;` to `packages/gemi/services/image-optimization/drivers/SharpDriver.ts`.
- A child process importing the barrel then loads `node_modules/.bun/sharp@0.34.5/node_modules/sharp/lib/{index,constructor,sharp,libvips}.js`.
- `bun --bun vitest run barrel-imports.test.ts` still reports **5 passed**.

The static form is caught correctly (2 of 5 fail, naming `sharp <- services/image-optimization/drivers/SharpDriver.ts`), so the guard is not broken in general — it has a hole shaped exactly like its own fix pattern.

Honest caveat, from verification: today `bun run typecheck` does reject that mutation with `TS1309: The current file is a CommonJS module and cannot use 'await' at the top level` (the `services/` tree is covered by the inherited CommonJS `module` setting), and CI runs typecheck before the tests. So the silent-regression path is currently backstopped by an unrelated tsconfig detail rather than by the guard, and it opens the day that setting changes.

Suggested fix: after the `scanImports` walk, add a cheap second pass over the source text for a heavy specifier reached by a dynamic import that is not inside a function body (a `/^\s*(?:const|let|var)?[^\n]*\bawait import\(\s*["'](…)["']/m` scan is enough for the shapes this repo actually writes), and reword the comment to "a dynamic import *inside a function* does not".

---

### 2. The comment justifying the static-walk design misdescribes what Bun can observe at runtime

`packages/gemi/barrel-imports.test.ts:20`

The header argues a runtime check is impossible: "a runtime plugin's `onResolve` fires for `await import(x)` and not for `import x from \"…\"`, so an instrumented child process reports a clean graph even when the eager import is right there."

On Bun 1.3.14 the discriminator is relative-vs-bare, not static-vs-dynamic: a `Bun.plugin` `onResolve({filter:/.*/})` registered via `--preload` **does** fire for a static relative import from a non-entry module. (The bare-specifier half was the one point where the two verifiers got different results and I would not lean on it either way.)

The load-bearing part is the conclusion, and that reproduces cleanly: `onLoad({filter:/node_modules\/(sharp|resend|satori)\//})` printed nothing when loading this branch's `services/index.ts`, and printed `HEAVY LOADED: …/sharp/lib/index.js` when `origin/main`'s eager `SharpDriver.ts` was restored. A five-line probe observes precisely the thing the comment says Bun offers nothing for. (One verifier hit `TypeError: onLoad() expects an object returned` when the callback returns `undefined`; returning a trivial `{ contents, loader }` passthrough or letting the probe throw after logging avoids that.)

The static walk remains a defensible primary check. Suggested fix: rewrite the paragraph to state the real reason (a static walk needs no native binaries and fails with a readable import chain), and drop the "Bun offers nothing that observes the second" claim — or add the `onLoad` probe as a complementary test, since it catches finding #1's hole for free.

---

### 3. The `sideEffects` comment's conclusion is right but every causal detail in it is wrong

`packages/gemi/scripts/build.ts:7`

Do-not-add-`sideEffects` is correct and reproduces. The explanation does not. The comment blames "Under Bun 1.3.14 with `splitting: true`" and says "the barrel keeps re-exporting `filesystemConfigDefaults` while the chunk that declared it drops the binding". Setting `"sideEffects": false` and rebuilding:

- `splitting: true` → `dist/services/index.js` is 2,097 bytes, literally `import"../chunk-….js";export{…84 names}` with **no named specifiers imported**; the chunk is 642 bytes of Bun interop helpers and zero gemi declarations. Import throws `SyntaxError: Exported binding 'S' needs to refer to a top-level declared variable`.
- `splitting: false` → still broken: 2,635 bytes, throws `Exported binding 'pj' needs to refer to a top-level declared variable`.
- Control, `splitting: false` with no `sideEffects` field → 1.8 MB, imports fine.
- `"sideEffects": ["**/*"]` with splitting on → 109 KB, imports fine (this half of the comment is accurate).

So splitting is irrelevant to the failure, nothing is "dropped at a chunk boundary" — Bun eliminates the entire entry graph while retaining the export list — and the quoted binding `Mk` matches neither observed error (the mangled name varies by engine and entrypoint).

Minor, since the actionable instruction is correct. Suggested fix: replace the causal sentence with "Bun applies the package's own `sideEffects` to the modules it is bundling and eliminates the whole entry graph while keeping the `export {}` list, so the emitted file re-exports names nothing declares — reproduced with `splitting` both on and off," and drop the specific mangled name.

---

### 4. `RedisManager.client` became a readonly getter; assigning to it now throws

`packages/gemi/services/redis/RedisManager.ts:38`

`public client: RedisClient` was a writable own instance property set in the constructor; it is now a prototype accessor with a getter and no setter, on a class `packages/gemi/services/index.ts:133` re-exports and `docs/project-structure.md:178` documents as the `redis` token's manager. Verified under Bun 1.3.14 against both revisions: on `origin/main` `Object.keys(m)` is `["config","client"]` and the assignment succeeds; on this branch it is `["config","instance"]`, `JSON.stringify(m)` emits `{"config":{},"instance":null}`, and both `(m as any).client = fake` and `Object.assign(m, { client })` throw `TypeError: Attempted to assign to readonly property.`

Minor, and the pushback is fair: nothing in `packages/`, `templates/`, or `docs/` writes to `.client`, and rebinding the token is the documented substitution seam. But the parallel S3 change in this same PR deliberately preserved the identical stubbing seam (`this.client ??= …` in `connect()`, relied on by `S3Driver.test.ts:28`'s `(driver as any).client = { send }`), and `RedisManager` gets no equivalent.

Suggested fix: `set client(c: RedisClient) { this.instance = c; }` — one line, no cost to the lazy-loading goal, restores the old contract and makes the new error message's advice reachable without replacing the whole binding.

---

### 5. Two neighbouring comments about `RedisManager` were made false by this PR

`packages/gemi/services/rate-limiter/drivers/RedisRateLimiterDriver.ts:16-19,178` and `packages/gemi/database/DatabaseManager.ts:16`

Both files are untouched by the diff, and both describe a `RedisManager` that no longer exists:

- The rate-limiter driver says "that module imports Bun's `RedisClient` as a value, and importing it here would make this driver — and anything that re-exports it — unloadable outside the Bun runtime." `RedisManager.ts:1` is now `import type { RedisClient, RedisOptions } from "bun"`. Confirmed empirically: `bun build --target=browser` on a probe importing `origin/main`'s `RedisManager` fails with `Browser build cannot import Bun builtin: "bun"`; on this branch it succeeds with no `bun` specifier in the output. The dead rationale is what justifies the `{ token: "redis" } as unknown as ServiceToken<{…}>` cast at line 21, which bypasses container type safety.
- `DatabaseManager.ts:16` reads "Like `RedisManager`, this is safe to build eagerly" — `RedisManager` now builds in a memoizing getter, verified with a `globalThis.Bun.RedisClient` construction counter (0 after the constructor, 1 after the first `.client` access). The reasoning about `DatabaseManager` itself is still sound; only the comparison is wrong, and this is the comment someone will be reading while doing the `database/Connection.ts` half of #396 that `barrel-imports.test.ts:30-34` flags as follow-up.

Both minor and doc-only, but they are self-introduced by a PR whose subject is exactly this class of staleness. Suggested fix: update both sentences; and since the constraint is gone, replace `REDIS_TOKEN` with a direct `import { RedisManager }` plus `container.bound/make(RedisManager)` — I verified that variant passes `bun run typecheck`, `barrel-imports.test.ts`, and the rate-limiter suite with no cast. (Note the barrel is still Bun-only via `database/Connection.ts:1`, so this is about the cast and the comment, not about actually shipping a browser build.)

---

### 6. `S3Driver`'s comment points at a method that does not exist

`packages/gemi/services/file-storage/drivers/S3Driver.ts:3`

"The SDK is loaded on first use instead — see `sdk()`." There is no `sdk()` member; the loader is `connect()` at line 36, and `sdk` is only a destructured local and the `S3Sdk` type alias. `grep -rn "sdk()" --include="*.ts" packages/gemi` returns exactly one hit: the comment itself. Trivial, and the `connect()` JSDoc 21 lines below does the real explaining — just change the reference to `connect()`.

---

### 7. S3 client construction errors move from boot to the first storage operation

`packages/gemi/services/file-storage/drivers/S3Driver.ts:36`

Construction moved from the constructor into `connect()`. The AWS SDK v3 constructor is mostly lazy, but not entirely: under Bun, `new S3Client({ region: "" })` throws `Region is missing` synchronously, while `{}`, no-args, `{region: undefined}`, a bad endpoint, empty credentials, a bogus region string, and a numeric region all construct silently. Running the two revisions side by side with `{region: ""}`, `origin/main` threw at construction and this branch only threw from `await size()`. That construction is genuinely on the boot path — `templates/saas-starter/app/kernel/Kernel.ts:5` statically imports `app/config/filesystem.ts`, which does `new S3Driver()` at module scope.

Minor by a wide margin: it is one config shape out of seven, the error is unchanged and still propagates, nothing asserts the old timing, and the boot-time validation was an accident of the SDK rather than a contract. Flagging it only because it is the single respect in which the driver does not behave identically — a deploy that used to fail fast now fails as a 500 on a user's first upload. If you want the old behaviour, a two-line region check in the constructor buys it back; otherwise this is fine as documented.

---

### 8. "~110 files each entry currently reaches" is wrong for `services`

`packages/gemi/barrel-imports.test.ts:145`

Running the test file's own `walk()` as a script: `services/index.ts` visits **151**, `facades/index.ts` visits **105**. The `visited > 50` assertion is unaffected. Trivial, but it is the number a future reader uses to judge whether the floor is still meaningful — write "≈105 (facades) and ≈150 (services)" instead.

nstfkc:
## Review addressed — [`7743bd0`](https://github.com/nstfkc/gemi/commit/7743bd0)

All eight, in the order they mattered.

**1 + 2 — the guard's blind spot, and the false claim about what Bun can observe.** Both were right, and the second is the fix for the first. `onLoad` *does* see a bare-specifier load, so `barrel-imports.test.ts` now leads with `loadedBy()`: a child Bun process that really imports the barrel with a plugin watching every module under the heavy packages. The graph walk stays as the check that can name the offending file, with its comment rewritten to say plainly that a module-scope `await import()` is eager and that `loadedBy()` is what covers it.

Both regression shapes now fail, verified:

| mutation | `loads no heavy SDK` | `holds no eager import` |
| --- | --- | --- |
| hoisted `const eagerSharp = (await import("sharp")).default` | ✗ fails (services + facades) | passes — blind, as documented |
| `origin/main`'s static `ResendDriver` | ✗ fails | ✗ fails, `resend <- services/email/drivers/ResendDriver.ts` |

**3 — the `sideEffects` note.** Reproduced your `splitting: false` result (2,635 bytes, `Exported binding 'pj'`). Comment rewritten: whole-entry-graph elimination with the `export {}` list retained, splitting irrelevant, mangled name not to be grepped for.

Also acted on the scenario in it — that `bun run build` and `build-publish.ts` both report success. `build-publish.ts` now imports `gemi/services` and `gemi/facades` out of the staged tree before staging, since the dangling-export check only asks whether a file *exists* and this failure mode is a file that exists. With `"sideEffects": false` reintroduced it now says:

```
Refusing to stage gemi@0.54.3: `import "gemi/services"` throws against the staged build.

SyntaxError: Exported binding 'S' needs to refer to a top-level declared variable.
```

Only those two subpaths: `./client` and `./vite` want a DOM and a Vite config, and a smoke test that needs a fixture is one that gets deleted the first time it is inconvenient.

**4 — `RedisManager.client`.** Setter added, exactly as suggested. New `RedisManager.test.ts` pins the whole contract: nothing constructed until asked, memoised after, injectable, and legible off Bun.

**5 — the two stale comments.** Both fixed, and I took the cleanup you verified: `RedisRateLimiterDriver` imports `RedisManager` directly and the `{ token: "redis" } as unknown as ServiceToken<…>` cast is gone. That moves the resolution onto a path nothing covered, so it comes with two tests — one resolving through a bound `RedisManager`, one asserting the "register `RedisServiceProvider`" message when nothing is bound. `DatabaseManager`'s comment now says what changed and why, rather than citing the counterexample.

**6 — `sdk()`.** Now points at `connect()`.

**7 — S3 boot-vs-first-call timing.** Taken as documented rather than reverted: buying the old timing back means constructing the client eagerly, which means importing the SDK eagerly. The `connect()` JSDoc now records which config shapes the SDK actually validates (an empty-string `region`, and nothing else out of seven), that an app building an `S3Driver` at module scope in `app/config/filesystem.ts` sees it on the first upload instead of at boot, and that the error itself is unchanged. Say the word if you would rather have the two-line region check.

**8 — the file count.** ≈150 for `services`, ≈105 for `facades`.

Green: typecheck clean, lint at the pre-existing 78 warnings, 141 files / 3548 tests passing, `build-publish` stages 22 resolved exports with both barrels imported clean.

PR #411, opened 2026-08-16
Implement `_count: true`, the count-every-relation shorthand

### LINKED ISSUE #394: ORM: `include: { _count: true }` (Prisma count-all-relations shorthand) throws UnsupportedQueryError at runtime
### Summary

The ORM rejects Prisma's count-all-relations shorthand `include: { _count: true }` with an `UnsupportedQueryError` instead of counting every relation. Only the named form `_count: { select: { <relation>: true } }` is accepted. Since the ORM markets Prisma-shaped queries, code ported from Prisma that uses the bare `true` form compiles fine but throws at runtime on every call.

### Version

gemi 0.52.0 (Postgres dialect).

### Repro

Any `findMany`/`findUnique` (including a nested `include`) that carries the bare `_count`:

```ts
await Product.findMany({
  include: {
    variants: {
      include: {
        _count: true, // Prisma: count all of variant's relations
      },
    },
  },
});
```

### Actual

```
UnsupportedQueryError: gemi ORM does not support '_count' yet (Product.findMany).
'_count: true' is not implemented. Name the relations: _count: { select: { <relation>: true } }.
```

Thrown from `planRelationCounts` (`dist/orm/compile/relation-count.ts`, the `X === true` branch). The error is `UnsupportedQueryError` (a "gap, not a decision"), and the `.d.ts` comment confirms the shorthand was "listed as out of scope and never rescheduled".

### Expected

Parity with Prisma: `include: { _count: true }` counts every relation of the model and returns them under `_count`, i.e. it desugars to `_count: { select: { <every relation>: true } }`. At minimum, if the shorthand stays unsupported by design, it would help to make this a compile-time type error rather than a runtime throw, so a Prisma-to-gemi migration surfaces it before deploy instead of 500ing in production.

### Impact / context

This bit us in production. Moving a set of controllers from Prisma onto the ORM left a `variants: { include: { _count: true } }` in a customer-facing list-detail endpoint. It type-checked and passed review, then every request to that endpoint threw the above and returned a 500. The runtime-only failure is the sharp edge: the bare form is valid Prisma and valid TypeScript, so nothing catches it until traffic hits the path.

Workaround on our side is to name the relations (or drop the count where it was unused), which is straightforward. Filing this because the silent-until-runtime behavior is a Prisma-compat trap others porting code will hit.

nstfkc:
Closes #394.

## The report

`include: { _count: true }` threw `UnsupportedQueryError` at runtime, pointing at the explicit `_count: { select: { … } }` form. #394 is an external report of what that costs in practice — a set of controllers ported from Prisma kept a `variants: { include: { _count: true } }`, type-checked, passed review, and then 500'd on every request to a customer-facing endpoint.

The sharp edge is not the refusal, it is that the refusal was invisible until traffic. `RelationCountPayload` returned `{ [K in keyof Relations<M>]: number }` for the shorthand, so the type system actively vouched for a query that always threw. The reporter offers the cheaper fix themselves — make it a compile-time error — and that would close the trap. It closes it by telling everyone porting from Prisma to rewrite the query, on a shorthand whose semantics are unambiguous and whose machinery already exists, so this answers it instead.

## What the expansion is, and why it lives where it does

`countableRelations(schema)` — every to-many relation, sorted — is the single derivation, and **both** the compiler and the policy walk read from it. That sharing is the entire difficulty of the change.

The original refusal gave two reasons. The first, *"it names no relation, so a policy has nowhere to attach"*, was an argument about **where** the expansion happens rather than whether it can happen at all. Expanded in `compile/relation-count.ts` alone, the policy walk still sees `true`, `scopeCounts` hands it back untouched, and every count is over rows the caller cannot read. So `policy.ts` expands it too, through the same helper, and the set that reaches SQL is exactly the set that carried a scope.

That failure mode is worth naming because of how it presents: what leaks is a **number**, not a row. Nothing in the response body looks wrong.

The compiler keeps its own copy of the expansion rather than depending on the walk having run, for the same reason `include: { accounts: true }` compiles when the compiler is called directly — scoping is `$exec`'s job, uniformly, and the compiler compiles what it is handed.

The policy walk expands **only when a policy actually applies**. Rewriting unconditionally would move the plan-cache key for every `_count: true` on an unpolicied model — a cache invalidation and a canonical-shape change in exchange for nothing, since the compiler expands it either way and the SQL is identical. This is the same trade the `node === true` branch in `applyNestedPolicies` already declines a few lines up.

## Semantics

**To-many only.** The explicit form refuses a to-one *by name*, because counting one answers 0 or 1 — which its nullability already says. The shorthand named nothing, so there is no argument to report back and skipping it is the whole of what Prisma's shorthand means.

`RelationCountPayload` now filters to `kind extends "many"`. It previously said `keyof Relations<M>`, which was unreachable and wrong in the same breath: a to-one would have been typed as a key that is never on the row, the moment the refusal lifted.

**A model with no to-many relations still raises**, rather than answering `{}`. The reason is mechanical: `resolveSelection` counts `_count` as *something selected*, so an empty expansion clears its "at least one field" check and then projects no columns — `select  from "Membership"`.

**What it returns changes when the schema grows a to-many relation.** This was the refusal's second reason. It is true, it is Prisma's semantics rather than a defect this could fix, and it is now documented at the shorthand — with a note to prefer naming the relations in code you intend to keep — instead of being answered with a throw.

## Tests

- **Two differential cases** (`include` and inside a `select`) run the shorthand against Prisma on the real starter schema. This is the only place the semantics are *checked* rather than reasoned about: `countableRelations` derives "every to-many" from gemi's schema and Prisma derives it from its own, so a relation missing from one side, or a to-one counted by either, shows up as a key difference in the row.
- **A corpus entry**, so the shorthand and the explicit form keep compiling identically on a one-to-many schema — if the expansion ever picked up the to-one, the discrimination suite says so. Corpus count 48 → 49, and the docs figure moves with it.
- **Policy tests** covering the four cases that matter: the shorthand is expanded then scoped; an unpolicied shorthand is returned *identically* (`toBe`, pinning the plan-key trade above); a mixed model keeps its unpolicied relations as `true` rather than dropping them; `asSystem` suspends it.
- **Type tests** asserting the payload with `toEqualTypeOf` on the whole object rather than key-by-key — the assertion that matters is the one a per-key check cannot make, that `organization` and `profile` are *absent*.

## Verification

| Suite | Result |
| --- | --- |
| `packages/gemi` ORM (64 files) | 2545 passed |
| saas-starter differential (vs Prisma) | 265 passed |
| saas-starter type tests | 27 passed, no type errors |

`bun run typecheck` in `packages/gemi` reports one error — `vite/dictionaryPlugin.ts` cannot find `magic-string`. It reproduces on clean `main` with this branch stashed, so it is pre-existing and unrelated.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -4376,9 +4376,20 @@ await User.findMany({ include: { _count: true } })
 ```
 
 To-many only. The explicit form refuses a to-one by name, because counting one answers 0 or 1; the
-shorthand named nothing, so there is no argument to refuse and the to-one is simply skipped. On a
-model with no to-many relations at all it raises rather than returning `{}` — there is nothing for
-it to mean, and inside a bare `select: { _count: true }` it would otherwise select no columns.
+shorthand named nothing, so there is no argument to refuse and the to-one is simply skipped.
+
+On a model with **no** to-many relations the shorthand is a compile error, not a runtime one. There
+is nothing for it to mean, and a query that type-checks and then throws on every call is the whole
+of what #394 reported — closing that for models which do have a to-many while leaving it open on
+models which do not would move the trap rather than remove it. `_count: false` stays legal
+everywhere, because the compiler short-circuits it before reaching any of this: a count switched off
+by a flag is a thing you can write on any model.
+
+A `_count` that ends up counting *nothing* is refused too, but only when it is the sole thing in a
+`select` — `select: { _count: { select: {} } }`, or every relation in it switched off. `_count`
+counts as something selected, so an empty one satisfies the "at least one field" check and then
+projects no columns at all. Under an `include` the model's own columns keep the statement valid, so
+counting nothing is simply nothing.
 
 The counted relations carry their policies exactly as the explicit form's do. That is worth stating
 because it is the part that is easy to get wrong: the shorthand names no relation, so the scope has
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3923,7 +3923,7 @@ back to batching for that node alone. It declines:
 A declined node still runs under this strategy one level down, so a decline costs one statement, not
 the subtree. A mixed include tree therefore uses both, which is fine — the results are the same
 either way, and there are tests asserting exactly that against Prisma across every relation shape
-in the differential corpus — **48** of them today, and the suite fails if this number stops
+in the differential corpus — **49** of them today, and the suite fails if this number stops
 matching the corpus.
 
 Override per call when you need to:
@@ -4367,9 +4367,29 @@ times inside that report before its own sentences were derived rather than writt
 The starter template indexes every foreign key for this reason, so a new application starts on the
 right side of it.
 
-`_count: true` (Prisma's "count every to-many relation") is **not** implemented. Name the relations
-instead — the shorthand names none, so a policy has nowhere to attach, and what it returns changes
-silently when the schema grows a relation.
+`_count: true` — Prisma's "count every to-many relation" — is the shorthand for naming them all:
+
+```ts
+await User.findMany({ include: { _count: true } })
+// _count carries one key per to-many relation on User:
+// { accounts: 3, passwordResetToken: 0, session: 1, socialAccount: 2 }
+```
+
+To-many only. The explicit form refuses a to-one by name, because counting one answers 0 or 1; the
+shorthand named nothing, so there is no argument to refuse and the to-one is simply skipped. On a
+model with no to-many relations at all it raises rather than returning `{}` — there is nothing for
+it to mean, and inside a bare `select: { _count: true }` it would otherwise select no columns.
+
+The counted relations carry their policies exactly as the explicit form's do. That is worth stating
+because it is the part that is easy to get wrong: the shorthand names no relation, so the scope has
+nowhere to attach until the shorthand is expanded, and expanding it in the compiler alone would
+count every relation unscoped. The expansion is shared with the policy walk instead, so the set that
+reaches SQL is the set that carried a scope.
+
+**What it returns changes when the schema grows a to-many relation.** That is Prisma's semantics
+rather than a wrinkle here, and it is the reason to prefer naming the relations in code you intend to
+keep: a response body that silently gains a key is a change no test asked for. The shorthand is for
+the exploratory case.
 
 ### Saying where nulls go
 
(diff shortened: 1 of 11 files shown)


nstfkc:
## Review addressed

Ran a four-lens adversarial review over this diff (policy leakage, derivation drift between the two expansions, type/runtime agreement, edge cases and test honesty), each finding then handed to skeptics prompted to refute it. Five survived. Three belong to this PR and are fixed in 6e33b6a; two were pre-existing defects this diff surfaced but did not cause, and one of those is serious enough that it has its own PR.

### Fixed here

**1. The shorthand type-checked on models it always throws on.** `readCountSelection` raises when a model has no to-many relation, but `IncludeInput`/`SelectInput` still offered `_count?: boolean` unconditionally — so on the **nine of the template's fourteen** models whose relations are all to-one (`Account`, `Session`, `Profile`, `PasswordResetToken`, `SocialAccount`, …), `include: { _count: true }` compiled clean under `--strict` and threw on every call.

That is verbatim the failure #394 reports, so closing it only for models that *do* have a to-many would have relocated the trap, not removed it — including in the nested shape the issue actually used, `include: { variants: { include: { _count: true } } }`. Found independently by two of the four lenses.

`CountShorthand<M>` now gates the key the way Prisma does structurally, reusing the branded-uninhabitable device from `ToOneRelationsCannotBeCounted` directly above it so the error names the reason. `_count: false` stays legal everywhere — `planRelationCounts` short-circuits it before reaching the throw, and refusing it would be the same divergence pointing the other way.

**2. A malformed operand was counted, unscoped.** `scopeCounts` reads `_count.select.<rel>` through `typeof value !== "object"` and `continue`s past anything that fails it, so `{ accounts: 1 }` carried no policy scope — while `countPlan` read `node?.where` as `undefined` and emitted a perfectly well-formed **unscoped** `count(*)`. Two walks, one argument, opposite readings: precisely the disagreement the shared `countableRelations` exists to prevent one level up. Now refused in the compiler, where the identical malformation on a relation node (`include: { accounts: 1 }`) is already refused — which is what makes the same `continue` in `applyNestedPolicies` safe.

**3. `select  from "User"` was still reachable.** The `countable.length === 0` guard this PR added covered one route into the empty-projection state; `select: { _count: { select: {} } }` — or every relation in it switched off by a flag — reached it through the explicit form and handed the caller a raw driver syntax error. Guarded on the *plan list* being empty rather than the object being, and only under `select`, since an `include` carries the model's own columns.

### Split out

**A cross-tenant leak in `$execute`, unrelated to `_count` — [#412](https://github.com/nstfkc/gemi/pull/412).** `applyPolicies` was passed the original `args` instead of `effective`, discarding the entire `applyNestedPolicies` rewrite whenever the **root** model carries any policy. Under that shape `include: { rel: true }` returns the other tenant's rows and `_count` returns the other tenant's total. It predates this PR, but this PR's shorthand routes through the discarded rewrite, so #412 should land first. Verified empirically and covered by three regression tests that fail on the previous line.

### Not changed

`countableRelations`' `.sort()` was flagged twice as inert — the fixture's relation names happen to be in sorted order already, so the test's docblock claims more than the test proves. Both skeptics refuted it as a defect: the sort is still load-bearing for any schema whose declaration order differs, which is the case it exists for. Left as is.

### Verification

| Suite | Result |
| --- | --- |
| `packages/gemi` ORM | 2548 passed |
| saas-starter `app/models/` | 871 passed, 134 skipped |
| saas-starter type tests | 124 passed, no type errors |

One review agent died mid-run (machine slept); its lens was the one that had already reported, and its findings were carried by the parallel lens covering the same ground.

PR #413, opened 2026-08-16
Go to definition on a route path

nstfkc:
`useQuery("/reports")` names its handler as precisely as a call names a function — the string is checked against the router at compile time, and autocomplete offers only paths that exist. But the connection is made by conditional types rather than by a symbol, so go-to-definition has nothing to follow and stops at the literal. Reading one call site meant opening `api.ts`, finding the entry, and following it to the controller by hand.

This adds a TypeScript language service plugin that closes the gap, enabled with one line in an app's `tsconfig.json`:

```json
"plugins": [{ "name": "gemi/ide/typescript-plugin" }]
```

It answers on any typed path — `useQuery`, `useMutation`, `useMutate`, `Form`'s `action`, `Link`'s `href`, and wrappers an app writes over them — resolving to the controller method, the inline callback, the `resource()` method for that verb, or for a view both the component and its handler. Hover names the route and where its handler lives. Everything else is delegated untouched, and nothing here can throw: a surprise degrades to "not a route", because a broken jump beats an editor with no language features.

Verified against the `saas-starter` template through a live `tsserver`:

```
useQuery("/suspense-demo/metrics")   → SuspenseDemoController.metrics()
useQuery("/test")                    → HomeController.index()
<Link href="/about">                 → app/views/About.tsx + its handler
<Form action="/home">                → HomeController.post()      (POST default)
<Form action="/auth/sign-up">        → AuthController.signUp()    (framework route)
```

## Three decisions worth reviewing

**Routers are discovered by reading the `RPC` / `ViewRPC` interfaces, not the app's route config.** The interfaces are what the call site is typed against, and their heritage clauses carry the router *and* its prefix — including the framework's own `CreateRPC<AuthApiRouter, "/auth">`, so `/auth/me` is as jumpable as any app route. Config-driven discovery would have found the app's routers and missed it. Only the app's own files are scanned to find one declaration; the checker supplies the rest of the merge group, so `node_modules` is never walked.

**The verb comes from the parameter's constraint, with no hook name hardcoded anywhere.** `useQuery(url: T)` declares `T extends keyof GetRPC`, so the constraint of the parameter an argument fills *is* the GET route set; testing that set against each verb's paths identifies the verb. A path carrying both a GET and a POST resolves correctly from either hook, and an app's own wrapper works for free. `useMutation`'s `K extends keyof Methods[M]` does not resolve to literals (probed, not assumed), so that case falls back to reading a verb from the preceding argument — keyed off shape, not callee name — and JSX falls back to the attribute name.

**`rpcParity.test.ts` is the load-bearing test.** The walker's key set has to be character-identical to `keyof RPC`, and nothing else enforces that: one side is a template literal type, the other a `.slice()`. So it asks the checker for `keyof RPC` over the same fixture the walker reads and requires set equality. It immediately caught three misreadings of the types:

- `ParsePrefixAndKey<"/org", "/">` is `/org`, not `/org/`
- `/(app)/` joined to `/(admin)/users` really does produce `///users`
- `RemoveTrailingId<"/:id">` is `""`, not the `"X"` the type appears to promise

`routePath.ts` reproduces all three rather than correcting them, because matching the types is what makes a lookup possible at all.

## Packaging note

tsserver ignores the `exports` map and resolves a plugin by Node directory rules, so `ide/typescript-plugin/package.json` exists only to point `main` at the built file. `build-publish.ts` stages it explicitly and fails the publish if it dangles. See `WHY-PACKAGE-JSON.md`.

In-repo the plugin needs `bun run build:ts-plugin` before an editor can load it from the template; the published tarball always has it.

## Also

Fixes a doc error next door: `data-fetching.md` claimed `gemi ide:generate-api-manifest` regenerates `gemi.d.ts`. It does not — it builds the manifest behind the Emacs integration, and `gemi.d.ts` needs no regeneration.

## Checks

`bun --bun vitest run ide/typescript-plugin` → 55 passing. Full package suite green (3,477). `bun run typecheck` covers the new directory through its own `tsconfig.json`. Lint and `oxfmt` clean.

## Known follow-ups (not in this PR)

- `VIEW_EXTENSIONS` tries `.tsx/.jsx/.ts/.js`, but the runtime only ever resolves `app/views/<path>.tsx` — both `createRollupInput` and `httpProd`'s manifest lookup are `.tsx`-only. One-line fix.
- Go-to-definition and completions for the view path *inside* the router (`this.view("partial/Reports")`), which is untyped today — a typo compiles and renders the 404 view silently. Discussed but deliberately not implemented here.
- `bin/ide/generateApiManifest.ts` now overlaps with this walk and has gaps it does not (it reads a verb-map object as nested paths, and hardcodes `@/app` resolution). Left alone so the Emacs integration keeps working.

### DIFF
--- a/docs/data-fetching.md
+++ b/docs/data-fetching.md
@@ -3,8 +3,8 @@
 gemi ships a small set of typed hooks for talking to your API routes from the
 client, plus a server-side facade for priming that data during SSR. Every hook is
 fully type-safe: the endpoint path, its params, its search input, and its response
-shape are all inferred from your API routes through the generated `gemi.d.ts` network
-layer. You never write a raw `fetch` or hand-annotate a response type.
+shape are all inferred from your API routes through the type augmentation `gemi/client`
+ships. You never write a raw `fetch` or hand-annotate a response type.
 
 All hooks and components below come from `gemi/client`:
 
@@ -436,10 +436,15 @@ response types from your `api.ts` / `view.ts` routers and applies them to `useQu
 the mutation hooks, `Form`, and `ViewProps` / `LayoutProps` automatically. You get
 autocomplete for valid paths and params and typed response data, with no manual wiring.
 
-This is backed by a `gemi.d.ts` that augments the `RPC` and `ViewRPC` interfaces with
-`CreateRPC<Api>` and `CreateViewRPC<View>`. It is wired up for you in the template and
-needs no regeneration — the types are derived from your routers by the compiler, so they
-follow a route the moment you add one.
+This is backed by an augmentation of the `RPC` and `ViewRPC` interfaces that reads your
+`api.ts` and `view.ts` routers. Since 0.56 it ships inside the package and is referenced
+by `gemi/client` and `gemi/facades`, so importing from either is the whole of the wiring
+— there is nothing to install and nothing to regenerate. The types are derived from your
+routers by the compiler, so they follow a route the moment you add one.
+
+> Upgrading from 0.55 or earlier? Delete your app's root `gemi.d.ts` and its
+> `"./node_modules/gemi/gemi.d.ts"` entry in `tsconfig.json`'s `types` — leaving them is
+> not a no-op. See [UPGRADE.md](../UPGRADE.md).
 
 ### Jumping from a path to its handler
 
--- a/docs/data-fetching.md
+++ b/docs/data-fetching.md
@@ -436,9 +436,37 @@ response types from your `api.ts` / `view.ts` routers and applies them to `useQu
 the mutation hooks, `Form`, and `ViewProps` / `LayoutProps` automatically. You get
 autocomplete for valid paths and params and typed response data, with no manual wiring.
 
-> **Note:** This is backed by a generated `gemi.d.ts` at your app root — don't edit it by
-> hand. Regenerate it after changing API routes with `gemi ide:generate-api-manifest`
-> (see the [CLI reference](./cli.md)).
+This is backed by a `gemi.d.ts` that augments the `RPC` and `ViewRPC` interfaces with
+`CreateRPC<Api>` and `CreateViewRPC<View>`. It is wired up for you in the template and
+needs no regeneration — the types are derived from your routers by the compiler, so they
+follow a route the moment you add one.
+
+### Jumping from a path to its handler
+
+The path in `useQuery("/reports")` names a handler as precisely as a call names a
+function, but the connection is made by conditional types rather than by a symbol, so
+go-to-definition has nothing to follow and stops at the string. gemi ships a TypeScript
+language service plugin that closes that gap. Add it to your `tsconfig.json`:
+
+```json
+{
+  "compilerOptions": {
+    "plugins": [{ "name": "gemi/ide/typescript-plugin" }]
+  }
+}
+```
+
+Go to definition on a route path then jumps to the code behind it — the controller
+method for `this.get(HomeController, "index")`, the callback for an inline handler, the
+right method of a `resource()` for the verb you are using, and for a view path both the
+component and the handler feeding it. Hovering a path names the route and its handler.
+It works on any typed path: `useQuery`, `useMutation`, `useMutate`, `Form`'s `action`,
+`Link`'s `href`, and on wrappers you write over them.
+
+> **VS Code** ships its own copy of TypeScript and ignores `plugins` unless told to use
+> the workspace's: run **TypeScript: Select TypeScript Version → Use Workspace Version**
+> once per project. Editors that drive `tsserver` over LSP — Neovim, Emacs, Helix,
+> JetBrains — read `tsconfig.json` directly and need nothing extra.
 
 ## Related
 
@@ -446,4 +474,3 @@ autocomplete for valid paths and params and typed response data, with no manual 
 - [Views and Layouts](./views-and-layouts.md) — server props vs. client queries.
 - [Controllers](./controllers.md) — writing the endpoints these hooks call.
 - [File Storage](./file-storage.md) — handling `useUpload` on the server.
-- [CLI](./cli.md) — regenerating `gemi.d.ts`.
--- a/packages/gemi/ide/typescript-plugin/callSite.ts
+++ b/packages/gemi/ide/typescript-plugin/callSite.ts
@@ -1,0 +1,206 @@
+import type ts from "typescript";
+
+import { getStringValue, type TS } from "./ast";
+import type { HttpVerb } from "./types";
+
+const HTTP_VERBS = new Set<string>(["GET", "POST", "PUT", "PATCH", "DELETE"]);
+
+/** Attributes that name a page to navigate to rather than an endpoint to call. */
+const VIEW_ATTRIBUTES = new Set(["href", "to"]);
+/** Attributes that name an endpoint to submit to. `<Form action=… method=…>`. */
+const ACTION_ATTRIBUTES = new Set(["action"]);
+
+export interface RouteQuery {
+  node: ts.StringLiteralLike;
+  /** The literal's text — the route path as the user typed it. */
+  path: string;
+  /**
+   * The paths this position is allowed to hold, when the signature says so.
+   *
+   * `useQuery(url: T)` declares `T extends keyof GetRPC`, so the constraint of
+   * the parameter this argument fills *is* the GET route set — which is how a
+   * `useQuery` call is told apart from a `useMutation` one without either name
+   * appearing here. Any wrapper an app writes over these hooks carries the same
+   * constraint through, and is understood for free.
+   */
+  allowedPaths?: ReadonlySet<string>;
+  /** A verb the call site names outright, e.g. `useMutation("PUT", …)`. */
+  verb?: HttpVerb;
+  /** Whether the surrounding syntax means a page or an endpoint. */
+  prefer?: "api" | "view";
+}
+
+/**
+ * Decides whether the string literal at `position` is naming a route, and
+ * gathers whatever the surrounding code says about *which* route.
+ *
+ * Nothing here confirms the path exists — that is the route table's job. This
+ * only answers "could this be a route reference, and what narrows it".
+ */
+export function findRouteQuery(
+  ts: TS,
+  checker: ts.TypeChecker,
+  sourceFile: ts.SourceFile,
+  position: number,
+): RouteQuery | undefined {
+  const node = findStringLiteral(ts, sourceFile, position);
+  if (!node) return undefined;
+  if (!isRouteReferencePosition(ts, node)) return undefined;
+
+  return {
+    node,
+    path: node.text,
+    allowedPaths: allowedPathsAt(ts, checker, node),
+    ...contextHints(ts, checker, node),
+  };
+}
+
+function findStringLiteral(
+  ts: TS,
+  sourceFile: ts.SourceFile,
+  position: number,
+): ts.StringLiteralLike | undefined {
+  let found: ts.StringLiteralLike | undefined;
+  const visit = (node: ts.Node): void => {
+    if (position < node.getFullStart() || position > node.getEnd()) return;
+    if (ts.isStringLiteralLike(node)) {
+      // Inside the quotes only: at the closing quote the request belongs to
+      // whatever follows, and hijacking it there would be surprising.
+      const start = node.getStart(sourceFile);
+      if (position > start && position < node.getEnd()) found = node;
+    }
+    ts.forEachChild(node, visit);
+  };
+  ts.forEachChild(sourceFile, visit);
+  return found;
+}
+
+/**
+ * Rejects the string positions that already mean something else, so the plugin
+ * never shadows a definition TypeScript can answer better itself.
+ */
+function isRouteReferencePosition(ts: TS, node: ts.StringLiteralLike): boolean {
+  const parent = node.parent;
+  if (!parent) return false;
+
+  // Module specifiers already go to definition — the module.
+  if (ts.isImportDeclaration(parent) || ts.isExportDeclaration(parent)) return false;
+  if (ts.isImportTypeNode(parent) || ts.isExternalModuleReference(parent)) return false;
+  if (ts.isImportSpecifier(parent) || ts.isExportSpecifier(parent)) return false;
+
+  // A key, not a value.
+  if (
+    (ts.isPropertyAssignment(parent) ||
+      ts.isPropertySignature(parent) ||
+      ts.isMethodDeclaration(parent) ||
+      ts.isPropertyDeclaration(parent) ||
+      ts.isEnumMember(parent)) &&
+    parent.name === node
+  ) {
+    return false;
+  }
+
+  // A type, where the definition is the type declaration.
+  if (ts.isLiteralTypeNode(parent)) return false;
+
+  // `config["/home"]` — TypeScript resolves this to the property, and that is
+  // the better answer.
+  if (ts.isElementAccessExpression(parent) && parent.argumentExpression === node) return false;
+
+  return true;
+}
+
+function allowedPathsAt(
+  ts: TS,
+  checker: ts.TypeChecker,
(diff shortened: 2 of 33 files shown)


packages/gemi/ide/typescript-plugin/entryPoints.ts
```diff
@@ -0,0 +1,300 @@
...
+function readMounts(
+  ts: TS,
+  checker: ts.TypeChecker,
+  declaration: ts.InterfaceDeclaration,
+  helperName: string,
+): RouterMount[] {
+  const mounts: RouterMount[] = [];
+  for (const clause of declaration.heritageClauses ?? []) {
+    for (const type of clause.types) {
+      if (!ts.isIdentifier(type.expression) || type.expression.text !== helperName) continue;
```
nstfkc: **Router discovery matches the heritage clause by the literal name `CreateRPC`/`CreateViewRPC`, which the augmentation gemi now ships does not use.**

This PR branches off `1bafb99` (0.54.3). On current `main`, #414 moved the augmentation into the package, and `packages/gemi/gemi.d.ts` now reads:

```ts
declare module "gemi/client" {
  export interface RPC extends AppRPC<Api> {}
  export interface ViewRPC extends AppViewRPC<View> {}
}
```

`AppRPC` / `AppViewRPC` are the `IsAny`-guarding aliases that wrap `CreateRPC`. `type.expression.text !== helperName` rejects them, so `readMounts` returns no mount for the app's own routers.

Verified by running `createFixture({ "gemi.d.ts": <the 0.56 shape> })` through `tableFor`:

```
api paths:  [ '/auth/sign-in', … '/auth/sign-in-with-pin-v2' ]   // framework only
view paths: []
diagnostics: []
```

Against the current fixture (`CreateRPC<Api>` spelled directly) the same call yields 26 api paths and 5 view paths. So after this merges, go-to-definition works on `/auth/*` and on nothing else in the application — the entire feature.

There is a second, independent blocker on an installed app: `findInterfaceDeclarations` skips every file whose path contains `/node_modules/`, and post-#414 the only file declaring `RPC` is `node_modules/gemi/dist/gemi.d.ts`. In-repo the workspace symlink hides this (realpath lands in `packages/gemi`), so it would not show up in template testing either.

`fixture.ts`'s `GEMI_ENV` pins the pre-0.56 shape, which is why `rpcParity.test.ts` — the test written specifically to catch drift between the walker and the types — cannot see this.

nstfkc (commented):
Code review of the TypeScript language service plugin. Six findings, each verified by running the plugin's own harness against modified fixtures. The first two are the significant ones: as of #414 (already on `main`) this plugin discovers **zero** application routes.

packages/gemi/ide/typescript-plugin/entryPoints.ts
```diff
@@ -0,0 +1,300 @@
...
+    for (const declaration of findInterfaceDeclarations(ts, checker, program, interfaceName)) {
+      result.dependencies.push(declaration.getSourceFile().fileName);
+      for (const mount of readMounts(ts, checker, declaration, helperName)) {
+        result[kind].push(mount);
+        result.dependencies.push(mount.declaration.getSourceFile().fileName);
+      }
+    }
+  }
+
+  if (result.api.length > 0 || result.view.length > 0) return result;
```
nstfkc: **This early return makes the fallbacks and the diagnostic below unreachable in any real gemi app, so total discovery failure is silent.**

`client/rpc.ts` unconditionally declares `export interface RPC extends CreateRPC<AuthApiRouter, "/auth">`, and it is always in the merge group. So `result.api.length` is > 0 in every gemi project regardless of whether the app's own routers were found — the route-config fallback, the `app/http/routes/api` convention fallback, and the `"no RPC / ViewRPC augmentation found"` diagnostic never run.

Concretely: in the probe above (0.56 augmentation shape) `table.diagnostics` came back `[]` while the table held only the twelve `/auth/*` routes. Nothing in the tsserver log says the app's routers were missed, and the fallback that would have found `app/http/routes/api.ts` by convention — and rescued this case entirely — is skipped.

Suggest gating on *app* mounts rather than any mount, e.g. track whether a mount came from a declaration outside the framework's own `client/rpc.ts` and fall through if none did.

packages/gemi/ide/typescript-plugin/lookup.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (query.prefer) {
+    candidates = narrow(candidates, (candidate) => candidate.kind === query.prefer);
+  }
+  if (query.verb) {
+    candidates = narrow(
+      candidates,
+      (candidate) => candidate.kind === "api" && candidate.entry.verb === query.verb,
+    );
+  }
+  if (query.allowedPaths) {
```
nstfkc, line 76: **Verb narrowing is all-or-nothing across the whole project: one route the walker cannot read but the types can silently disables it for every call site.**

`isSubset(allowed, pathsByVerb.get(verb))` requires the walker's set for a verb to be a *superset* of `keyof GetRPC`. Any route that reaches the RPC types but not the walk breaks that containment for every verb at once, `narrow` then keeps all candidates, and disambiguation stops working everywhere — not just at the route that was missed.

A spread in `routes` is enough, and it typechecks fine (`RouteParser` reads the declared type of `routes`, so the spread's members *are* in `keyof RPC`; `walkApiRouter` skips them because they are not `PropertyAssignment`s):

```ts
export default class Root extends ApiRouter {
  routes = {
    "/thing": { get: this.get(HomeController, "index"),
                post: this.post(HomeController, "archive") },
    ...{ "/spread": this.get(HomeController, "index") },
  };
}
```

`getDefinitionAndBoundSpan` on `useQuery("/thing")`:

```
without the spread:  [ api.ts index   [HomeController] ]          // GET, correct
with the spread:     [ api.ts index   [HomeController],
                       api.ts archive [HomeController] ]          // GET + POST
```

The blast radius is what makes this worth fixing rather than accepting as graceful degradation — the failure is remote from its cause, and there is no diagnostic. Two cheap mitigations: log a diagnostic when a `routes` object contains a `SpreadAssignment`/`ShorthandPropertyAssignment` the walk drops, and/or fall back to "does the table contain this path under this verb" when the subset test fails wholesale.

packages/gemi/ide/typescript-plugin/callSite.ts
```diff
@@ -0,0 +1,206 @@
...
+  return true;
+}
+
+function allowedPathsAt(
+  ts: TS,
+  checker: ts.TypeChecker,
+  node: ts.StringLiteralLike,
+): ReadonlySet<string> | undefined {
+  const call = node.parent;
+  if (!call || (!ts.isCallExpression(call) && !ts.isNewExpression(call))) return undefined;
```
nstfkc: **`useMutate` gets no narrowing at all, and the test that claims to cover it uses a signature the hook does not have.**

The real API is `mutate({ path, params?, search? }, fn?)` (`client/useMutate.ts:22`, and `docs/data-fetching.md:425`). The path is an object property, so `node.parent` is a `PropertyAssignment`, this guard bails, and `allowedPathsAt` returns `undefined` — even though `mutate`'s `path` is declared `T extends keyof GetRPC`, which is exactly the constraint the rest of the design reads.

Measured on the fixture with the real signature:

```
mutate({ path: "/thing" })      => [ index [HomeController], archive [HomeController] ]
mutate({ path: "/org/:orgId" }) => [ index [HomeController], OrgOverview [view],
                                     OrgLayout [view], overview [PageController] ]
```

The first should be the GET handler alone; the second should be the GET handler alone (the constraint is `keyof GetRPC`, so it cannot be a view route) but offers four targets including two view components.

`plugin.test.ts` asserts on `useMutate()("/reports")` — a call shape `useMutate` does not accept — so the claim in the PR description that `useMutate` is supported is untested. Reading the constraint through a property assignment's contextual type (`checker.getContextualType`) would cover this and any other options-object hook.

packages/gemi/ide/typescript-plugin/routeTable.ts
```diff
@@ -0,0 +1,467 @@
...
+    prefix: string,
+    stack: ts.ClassLikeDeclaration[],
+  ): void {
+    const ts = this.ts;
+    const routes = getRoutesObjectLiteral(ts, classDeclaration);
+    if (!routes) return;
+    this.dependencies.add(classDeclaration.getSourceFile().fileName);
+
+    for (const property of routes.properties) {
+      if (!ts.isPropertyAssignment(property)) continue;
```
nstfkc: **`routes` entries that are not `PropertyAssignment`s are dropped without a diagnostic, though the RPC types keep them.**

Same line at 226 for the view walk. A spread (`...sharedRoutes`) or a shorthand entry is valid TypeScript, and `RouteParser`/`RoutesParser` operate on the *declared type* of `routes`, so those paths do appear in `keyof RPC` and autocomplete. Here they vanish: `useQuery("/spread-route")` gets no definition, and — see the `lookup.ts` comment — the resulting hole also disables verb narrowing project-wide.

The docstring on `getRoutesObjectLiteral` argues that shapes invisible to the walk are also invisible to `CreateRPC`. That is true of a `routes` built in a constructor, but not of a spread inside the initializer. At minimum this deserves a `this.diagnostics.push(...)` so the tsserver log names the route that was skipped.

docs/data-fetching.md
```diff
@@ -436,14 +436,41 @@ response types from your `api.ts` / `view.ts` routers and applies them to `useQu
 the mutation hooks, `Form`, and `ViewProps` / `LayoutProps` automatically. You get
 autocomplete for valid paths and params and typed response data, with no manual wiring.
 
-> **Note:** This is backed by a generated `gemi.d.ts` at your app root — don't edit it by
-> hand. Regenerate it after changing API routes with `gemi ide:generate-api-manifest`
-> (see the [CLI reference](./cli.md)).
+This is backed by a `gemi.d.ts` that augments the `RPC` and `ViewRPC` interfaces with
```
nstfkc: This rewrite is correct against 0.54.3 but wrong against `main`: #414 removed the app-root `gemi.d.ts` and `916c007` documents deleting it on upgrade. The template no longer has one and no longer names it in `tsconfig.json`'s `types` — the augmentation now ships inside the package and is pulled in by importing `gemi/client` or `gemi/facades`.

Related merge hazard: the `templates/saas-starter/tsconfig.json` hunk in this PR carries `"types": ["vite/client", "bun", "./node_modules/gemi/gemi.d.ts"]` as context, and that line changed on `main` to `"types": ["vite/client", "bun"]`. Re-check it after rebasing so the stale `types` entry is not reintroduced — an unresolvable `types` entry makes `tsc` report TS2688 and nothing else.

packages/gemi/ide/typescript-plugin/entryPoints.ts
nstfkc: Fixed — router discovery no longer reads the helper's name at all.

`readMounts` scans the heritage clause's *type arguments*: whichever resolves to a class is the router, a string literal beside it is the prefix. `AppRPC<Api>`, `CreateRPC<AuthApiRouter, \"/auth\">` and any wrapper an app writes all read the same way.

The node_modules blocker is fixed separately: `findInterfaceDeclarations` now asks the checker for `gemi/client`'s exports instead of walking files. Reaching the module symbol costs one scan of the top-level statements of the app's own files, looking for an import of it or a `declare module` for it; the checker supplies the whole merge group from there, including declarations inside `node_modules`.

And you were right that `fixture.ts` was the reason none of this was visible. The fixture now has **no `gemi.d.ts` of its own** — with `gemi/*` mapped at the package source, that is exactly a 0.56 app, reached through `packages/gemi/gemi.d.ts`. Every existing test now runs against that shape; only the dependency assertion in `rpcParity.test.ts` needed changing.

`entryPoints.test.ts` adds the shapes around it, including one where every declaration of `RPC` is under `node_modules` and the router sits where the convention fallback does not look. It fails without the module-symbol lookup.

packages/gemi/ide/typescript-plugin/entryPoints.ts
nstfkc: Fixed, along the lines you suggested — the gate is now on *application* mounts.

`isApplicationFile` is "under `projectRoot` and not in `node_modules`", so `client/rpc.ts`'s `/auth/*` no longer counts as evidence that anything was found. The per-kind convention fallbacks moved to the same test (they were guarded on `result.api.length === 0`, which had the same problem), and the diagnostic now reads *"the RPC / ViewRPC augmentation names none of this project's routers"*.

Covered in `entryPoints.test.ts` by an app whose `@/app/*` does not resolve — the case `AppRPC`'s `IsAny` guard exists for. The framework mounts still arrive, the diagnostic fires, and the convention fallback rescues the app's routers. It passes on the old gate only because the old gate returns early; restoring `result.api.length > 0` fails it.

packages/gemi/ide/typescript-plugin/lookup.ts
nstfkc: Fixed, and thank you for the blast-radius framing — that is what decided the shape of the fix.

Rather than falling back when the subset test fails wholesale, `allowedPaths` is intersected with the set of paths the walk actually found before the comparison. A path the walk never saw has no row to jump to either way, so dropping it from the comparison costs nothing and keeps the failure at the route that caused it. In the healthy case the intersection is a no-op.

Your spread repro is the test (`entryPoints.test.ts`), with `allowedPaths` read from the checker as `keyof RPC` over the same fixture rather than hand-written — `/spread` is in it, and `useQuery(\"/thing\")` still resolves to GET alone. Reverting the intersection fails it.

The diagnostic you asked for is in too — see the reply on `routeTable.ts`.

packages/gemi/ide/typescript-plugin/callSite.ts
nstfkc: Fixed. `allowedPathsAt` now locates the argument through an `argumentSite` helper that handles both spellings: the argument itself, and a property of an options object that is an argument.

One correction to the suggestion — `checker.getContextualType` does not work here. I measured it: by the time it is asked, inference has run and the contextual type of `\"/reports\"` is `\"/reports\"`, for `useQuery` exactly as for `mutate`. What does work is the same route the argument case already took — the *declared* type of the parameter, then `getProperty(\"path\")` on it, then the base constraint. That yields the 10-member `keyof GetRPC` union, and covers any options-object hook.

`plugin.test.ts` now calls the hook the way it is actually declared. Both of your measurements are assertions: `mutate({ path: \"/thing\" })` resolves to the GET handler alone, and `mutate({ path: \"/org/:orgId\" })` no longer offers the two view components.

packages/gemi/ide/typescript-plugin/routeTable.ts
nstfkc: Fixed — `noteUnreadableEntry` pushes a diagnostic naming the router, the prefix, and the entry (`the spread \`...sharedRoutes\``, or a shorthand entry by name), on both the API and the view walk.

The docstring on `getRoutesObjectLiteral` is corrected too. You are right that the argument holds for a `routes` built in a constructor and not for a spread inside the initializer, and it now says so and points at the diagnostic.

The hole no longer costs verb narrowing project-wide either — see the reply on `lookup.ts`.

docs/data-fetching.md
nstfkc: Fixed. Rebased onto `main`; the `types` conflict resolved to `[\"vite/client\", \"bun\"]`, so the stale entry is not reintroduced — the template's `tsconfig.json` hunk is now the `plugins` line and its comment only.

The prose is rewritten for 0.56: the augmentation ships inside the package, `gemi/client` and `gemi/facades` reference it, and there is a pointer to UPGRADE.md for deleting the old root file. While there I also corrected the older claims elsewhere in `docs/` that the same file is generated, must not be hand-edited, and sits at the app root (`docs/README.md`, `getting-started.md`, `project-structure.md`, `index.html`, both `llms*.txt`) — that last one is the file UPGRADE.md tells you to delete, so leaving it seemed worse than the wider diff. Say the word if you would rather have those split out.

PR #417, opened 2026-08-17
Events iteration 1: sync dispatch

nstfkc:
Iteration 1 of `plans/events` — `Event`, `Listener`, `EventManager`, discovery, sync fan-out, and the zero-listener warning. The plan files land with this PR under `plans/events/`.

```typescript
// app/events/UserRegistered.ts
export class UserRegistered extends Event {
  static name = "UserRegistered";
  constructor(public userId: number, public email: string) { super(); }
}

// app/listeners/SendWelcomeEmail.ts
export class SendWelcomeEmail extends Listener {
  static name = "SendWelcomeEmail";
  static event = UserRegistered;
  async handle(event: UserRegistered) { await sendWelcome(event.email); }
}

// the controller stops knowing what happens next
UserRegistered.dispatch(user.id, user.email);
```

## What it delivers

- **`services/events/Event.ts`** — `static name`, `dispatch` (void) and `dispatchAndWait` (resolves when every sync listener has settled), both typed through `ConstructorParameters` the way `Job.dispatch` is typed through `Parameters<T["run"]>`.
- **`services/events/Listener.ts`** — `static name`, `static event`, abstract `handle`. No generic parameter: the base annotates `handle(event: Event)` and method-parameter bivariance is what lets a subclass narrow it.
- **`services/events/EventManager.ts`** — token `"events"`, `Record<string, ListenerClass[]>`, `useListeners` / `registeredListeners` shaped after `QueueManager`.
- **`services/events/config.ts`** — the `events` slice, carrying the queue's "declared wins, and `[]` is declared" rule. No `eventsDir`: event classes are never discovered.
- **`services/events/EventServiceProvider.ts`** — binds in `register()`, discovers in `boot()` off the **raw** slice, registered after `QueueServiceProvider`.
- **`services/discovery.ts`** — `discoverListeners`, plus the two name checks and the refusal described below.
- Barrel exports, `docs/events.md`, and entries in `docs/README.md`, `docs/index.html`, `docs/llms.txt` and `docs/llms-full.txt`.

## The three decisions worth reviewing

**The registry is keyed by the declared `static name`, never by class identity.** A class-keyed map is wrong in production only: `gemi build` bundles and minifies the controller and every event class it imports to dispatch, while discovery imports `app/listeners/*.ts` — and their `app/events/*.ts` — from source. Two module graphs, two class objects, and a lookup that finds nothing. `warnIfEventNameWillNotSurviveTheBuild` reads `Object.getOwnPropertyDescriptor(event, "name")?.writable`, because in development a declared name and an implicit one read the same string.

**A dispatch nothing handles warns once per event name, in development only.** That single line is the only symptom of four different mistakes, and zero listeners is also a legal steady state.

**One listener's failure never stops another.** Each listener runs in its own `try`; the throw is logged with both names and the loop continues. Neither `dispatch` nor `dispatchAndWait` rejects — a rejection would make listener 1's failure the caller's problem while listener 5's silently was not, depending only on where in a filesystem walk the throw landed.

## Deviations from the plan

- **`queued` is not shipped.** The deliverable sketch in §2 shows `queued = false`, but `queued` is also on the iteration's own out-of-scope list, and shipping the field before the queue routing exists would mean `queued = true` runs the listener synchronously inside the request while the author believes it does not — the silent no-op the invariants exist to prevent. Iteration 2 adds it with the behaviour.
- **`Event` carries a `declare protected readonly __isEvent: true` marker.** The plan asks the `.test-d.ts` to make `static event = SomethingThatIsNotAnEvent` an error, and an event has no instance members, so `new (...args: any[]) => Event` was structurally satisfied by every class in existence. `declare` emits no field and no subclass initialises it; without it that half of the type test is vacuous.
- **`EventClass` and `ListenerClass` are exported from the barrel** alongside the six names the plan lists, since `EventConfig.listeners` is written in terms of one of them.
- **A listener with no `static event` is refused in two places.** `discoverListeners` throws a `DiscoveryError` naming the classes (the plan's "refused here rather than warned about"), and `EventManager.useListeners` logs and skips (the plan's test, "refused, by name, at `useListeners`") so an explicit `listeners: [...]` slice does not take the boot down. Note the interaction with the documented residual: an abstract base in `listenersDir` **with** a `static event` is still discovered and registered like a listener, as the plan says; one **without** now stops the boot with a message telling the author to move it out of the directory.

## Verification

From `packages/gemi`, on the worktree:

```
(no output)

$ bunx oxlint
Found 80 warnings and 0 errors.
Finished in 47ms on 527 files with 93 rules using 8 threads.
   # 80/0 is also what `main` reports; nothing under services/events warns.

$ bun --bun vitest run services/events services/discovery.test.ts docs-imports.test.ts barrel-imports.test.ts
 Test Files  4 passed (4)
      Tests  68 passed (68)

$ bunx vitest run --typecheck.only
 Test Files  4 passed (4)
      Tests  32 passed (32)
Type Errors  no errors

$ bun --bun vitest run services support foundation kernel container console
 Test Files  38 passed (38)
      Tests  509 passed (509)
```

The last one is the regression check on adding a sixteenth framework provider: every suite that boots a kernel is in it.

Formatting: the new files under `services/events/` are prettier-clean. `services/index.ts`, `services/discovery.ts` and `services/discovery.test.ts` were **not** run through a formatter — the first two already fail `oxfmt --check` at `HEAD` and the third already fails `prettier --check`, so formatting them would have buried a small edit in unrelated churn. Their diffs are hand-matched to the surrounding style.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,6 +60,7 @@ See **[Getting Started](./getting-started.md)** for requirements and a first rou
 - **[Jobs & Queues](./jobs-and-queues.md)** — defining and dispatching background `Job`s.
 - **[Cron](./cron.md)** — scheduling recurring `CronJob`s.
 - **[Commands](./commands.md)** — one-off application commands run with `gemi run`.
+- **[Events & Listeners](./events.md)** — `Event.dispatch(...)` fanning out to the listeners under `app/listeners`.
 - **[Broadcasting](./broadcasting.md)** — websocket channels, the `Broadcast` facade, `useSubscription` / `useBroadcast`.
 - **[Internationalization](./i18n.md)** — component-scoped dictionaries, `useTranslator`, `useLocale`, locale detection.
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -104,6 +104,8 @@ await UserRegistered.dispatchAndWait(user.id, user.email);
 
 It resolves once every listener has settled. Two methods rather than one that is sometimes worth awaiting, because "should I await this?" is a question the call site should not have to guess at.
 
+Listeners run **one at a time**, each awaited before the next one starts, so registration order is the order they run in and not merely the order they start. The cost is worth knowing: a listener that *hangs* — an un-timed-out request to a host that never answers — holds up every listener after it for as long as it hangs, and under `dispatch` the caller has already moved on, so the row a later listener writes is simply not there yet. A listener that can block indefinitely wants its own timeout, or a [job](./jobs-and-queues.md). A listener that *throws* is not this case — see below.
+
 Neither method ever rejects. See below.
 
 ## When a listener throws
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,0 +1,242 @@
+# Events & Listeners
+
+An event is something your application did, announced to nobody in particular. A **listener** is one side effect of it, in its own file. One dispatch fans out to every listener bound to that event, and the code that dispatched it does not hold the list.
+
+```typescript
+// The controller stops knowing what happens next.
+const user = await User.create(input);
+UserRegistered.dispatch(user.id, user.email);
+```
+
+That is the whole of what this buys: adding a fourth side effect to a registration becomes **adding a file**, rather than editing the controller that already has three.
+
+Listeners run **in-process and synchronously** inside the dispatching request — see [When to reach for a job instead](#when-to-reach-for-a-job-instead).
+
+## Defining an event
+
+An event is a class extending `Event` (from `gemi/services`) with a **static `name`** and a constructor holding the payload. Events live wherever you like; `app/events/` is the convention.
+
+```typescript
+// app/events/UserRegistered.ts
+import { Event } from "gemi/services";
+
+export class UserRegistered extends Event {
+  static name = "UserRegistered";
+
+  constructor(
+    public userId: number,
+    public email: string,
+  ) {
+    super();
+  }
+}
+```
+
+An event is the payload plus a name. It carries no framework state — no timestamp, no id, no "stop propagation" — because each of those is a field you can add yourself if you want one, and a listener that could cancel the listeners after it would make registration order matter (see [When a listener throws](#when-a-listener-throws)).
+
+> **Note:** The static `name` is **required**, and omitting it does not fall back harmlessly to the class name. The listener registry is keyed by this string. `gemi build` minifies the server entry, and the app code reachable from it — your controller, and every event class it imports in order to dispatch — is bundled and minified with it, which renames the class binding; a class's implicit `.name` *is* that binding, so `UserRegistered` becomes something like `D`. Discovery, meanwhile, imports `app/listeners/*.ts` from source at runtime, and those files import `app/events/*.ts` from source too. Two module graphs, two class objects, two different names — and the dispatch reaches nobody, in production and nowhere else. A declared `static name` is a string literal, which survives minification intact. Discovery warns at boot about any event a listener binds to that leaves it out.
+
+### Never `instanceof` an event
+
+The same two class objects are why this is unsafe **in your own code**:
+
+```typescript
+// Don't. `true` in development and in every test, `false` in a production build.
+if (event instanceof UserRegistered) { /* ... */ }
+```
+
+Your listener's `UserRegistered` came from source; the instance it was handed was built from the bundled one. Nothing the framework does can fix that, so the framework's job is to never need it — which is why a listener binds to exactly one event and always knows what it received.
+
+## Writing a listener
+
+A listener is a class extending `Listener` (from `gemi/services`) under `app/listeners/`, with a static `name`, a static `event`, and a `handle` method.
+
+```typescript
+// app/listeners/SendWelcomeEmail.ts
+import { Listener } from "gemi/services";
+import { UserRegistered } from "@/app/events/UserRegistered";
+
+export class SendWelcomeEmail extends Listener {
+  static name = "SendWelcomeEmail";
+  static event = UserRegistered;
+
+  async handle(event: UserRegistered) {
+    await sendWelcome(event.email);
+  }
+}
+```
+
+The directory is read at boot, so writing the file is all it takes — there is no list to keep alongside it.
+
+**The binding lives on the listener, not on the event.** The listener is the thing with an opinion about what it cares about; an event has no business knowing who is watching. An event listing its listeners would also put the property back the way it was: adding a side effect would mean editing an existing file.
+
+**Annotate `handle` with the event you bound to.** Narrowing the base class's `Event` parameter is legal and is how the payload gets typed without a generic to carry around. What the compiler *cannot* check is that the annotation and `static event` agree — a static and an instance member cannot reference each other's types. Copy a listener, change the static, forget the annotation, and TypeScript is satisfied while the listener receives something else. The failure is local and immediate (that one listener reads a field that is not there, in its own stack), which is why it is an accepted seam rather than a reason to complicate the API.
+
+### One event per listener
+
+There is no `static events = [A, B]`, deliberately. A listener bound to two events has to discriminate inside `handle`, and the natural way to write that is `event instanceof UserRegistered` — which is exactly the line that is `true` in every test and `false` in production. Two events wanting the same side effect are two small listeners calling one shared function.
+
+### Configurable fields
+
+| Field | Type | Default | Description |
+| --- | --- | --- | --- |
+| `static name` | `string` | `"unset"` | Unique listener identifier. Required. |
+| `static event` | `Event` subclass | — | The one event this listener handles. Required. |
+| `handle` | method | — | The side effect. May be `async`. Required. |
+
+## Dispatching
+
+Call `dispatch` with exactly the arguments your event's constructor takes — the call is typed against it, the same way `Job.dispatch` is typed against `run`.
+
+```typescript
+import { UserRegistered } from "@/app/events/UserRegistered";
+
+UserRegistered.dispatch(user.id, user.email);
+```
+
+`dispatch` returns `void`. It fires the listeners and moves on; nothing they do can be reported back to the call site.
+
+When the caller genuinely needs the side effects to have happened first — a request that renders what a listener wrote, or a test asserting on the result — use `dispatchAndWait`:
+
+```typescript
+await UserRegistered.dispatchAndWait(user.id, user.email);
+```
+
+It resolves once every listener has settled. Two methods rather than one that is sometimes worth awaiting, because "should I await this?" is a question the call site should not have to guess at.
+
+Neither method ever rejects. See below.
+
+## When a listener throws
+
+Each listener runs inside its own `try`. A throw is logged with the event name, the listener name and the error, and **the next listener still runs**.
+
+That is not politeness, it is the point. Listeners are independent side effects, and their order comes from a filesystem walk rather than from anything you chose — so letting listener 2 cancel 3 through 5 would make that walk load-bearing, which is the exact coupling events exist to remove. For the same reason neither `dispatch` nor `dispatchAndWait` rejects: a rejection would make listener 1's failure the dispatcher's problem while listener 5's silently was not, depending only on where in the walk the throw landed.
+
+A listener that must not fail silently should handle its own errors — retry, log to your own sink, or dispatch a [job](./jobs-and-queues.md) that can be retried and dead-lettered.
+
+## When nothing is listening
+
+A dispatch with no listeners is legal and does nothing. In development it warns **once per event name**:
+
+```
+[gemi] UserRegistered was dispatched and nothing is listening for it. ...
+```
+
+That single line is the whole early-warning system here, because four different mistakes produce the same symptom: an event that never declared `static name`, a typo, a listener that was never discovered, and a listener directory that was never walked. It fires once per name so that a dispatch in a loop does not bury the terminal, and never in production, where "no listeners" is an ordinary answer.
+
+## Registering listeners — `app/listeners/`
+
+Every class under `app/listeners` that extends `Listener` is registered when the kernel boots.
+
+Event classes are **not** discovered, and there is no `eventsDir`. Each event is already in the module graph — imported by the listener that binds to it, and by the code that dispatches it — so a walk over them would only be a boot cost.
+
+### Two listeners, one name
+
+A listener's `static name` identifies it, and two classes cannot claim the same one. The first is registered, the second is refused with a line on stderr. A directory walk is what makes that ordinary: `auth/NotifyAdmins.ts` beside `billing/NotifyAdmins.ts` is a natural thing to write, and nothing forces the import alias a hand-written list would have demanded. Rename one.
+
+Both still appear in `registeredListeners`, which reports what the manager was handed rather than what it accepted, so a test walking it sees the clash.
+
+### A listener that binds to nothing stops the boot
+
+A `Listener` subclass with no `static event` cannot be registered under any name, so it would be a file you wrote that never runs. Discovery refuses it by name rather than warning, because there is no reading of it under which you meant it.
+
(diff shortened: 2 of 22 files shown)


nstfkc:
## Review — invariants

Checked the five invariants in `plans/events/README.md`, plus the raw-slice rule. All five hold. Detail, so the next reader does not have to re-derive it:

1. **Keyed by the declared name string.** `EventManager.ts:109` keys registration off `listener.event.name` and `:165` keys dispatch off `(event.constructor).name` — both plain strings into a `Record<string, ListenerClass[]>` (`:38`). No `Map`/`Set`/`WeakMap` keyed by a class, no `instanceof`, no `===` between class references anywhere in `services/events/` (the only `instanceof` occurrences in the subsystem are inside doc comments telling apps not to). The one identity-keyed collection in the diff, `warned = new Set<EventClass>()` in `discovery.ts:warnIfEventNameWillNotSurviveTheBuild`, is a warning de-duplicator confined to a single source-side walk, so both sides of the `has`/`add` are the same module graph; the worst it can do is print a warning twice.
2. **Zero-listener warning.** `EventManager.ts:186-197`: production checked first and returns without recording, then the per-name `warnedFor` set, which is per-manager (`:44`). Covered by four tests in `EventManager.test.ts:199-246`, including the second-event and fresh-manager cases.
3. **Listener isolation.** `EventManager.ts:173-183` is a `for…of` with the `await` and the `new Listener()` both inside a per-iteration `try`. No `Promise.all`/`allSettled` over the listener list, so a throw (sync or async, from `handle` or from the constructor) is logged with both names and the loop continues. `dispatch` floats the promise but `dispatchAndWait` has nothing left to reject with.
4. **Sync/queued context.** `queued` is not declared in this iteration, and nothing in `docs/events.md` or the doc comments mentions it — so there is no declaration site to be wrong at. The context statement that does exist is on `Listener.handle` (`Listener.ts:91-95`) and in `docs/events.md:222`, and both describe it as a context boundary rather than a speed knob.
5. **No framework dispatch.** `EventManager` is referenced only from `Event.ts`, `EventServiceProvider.ts`, the barrel and the tests; nothing under `packages/gemi` extends `Event` or calls `dispatch`.

**Raw-slice rule.** `EventServiceProvider.boot` reads `this.app.config.get<EventConfig>("events", {})` and branches on `slice.listeners !== undefined` (`EventServiceProvider.ts:40-41`). `Repository.get` returns the stored object untouched, so that really is the pre-defaults slice, and `withDefaults` is only applied afterwards to pull `listenersDir` out. Absent key → `undefined` → discovers; `listeners: undefined` → discovers; `listeners: []` → returns and the directory is never read. Also confirmed the manager cannot corrupt the check: `withDefaults` copies, so `useListeners`'s `this.config.listeners = listeners` mutates the copy, not the slice `boot` later re-reads. Three provider tests pin all three cases.

Two nits, neither blocking:

- **`services/events/Event.ts:130-139`** — the doc comment on `refuseUnnamed` says the throw is reachable from "the base itself, or an anonymous class expression that never got a binding to take a name from". The second half is not true. An anonymous class expression gets an own, non-writable `name` of `""`, not the inherited `"unset"`: `Object.getOwnPropertyDescriptor(class extends Base {}, "name")` is `{ value: "", writable: false }`. So `(class extends Event {}).dispatch()` does not throw — it dispatches under the key `""` and surfaces as the generic "nothing is listening for it" warning instead of the message that names the actual fix. Exotic input (any binding, including `export default`, gives the class a name), so the floor is no weaker in practice than the comment claims — but a maintainer reading that line would believe unnamed dispatch is refused when it is only renamed.
- **`services/events/EventManager.ts:173`** — sequential `await` means a listener that *hangs* rather than throws does block the ones after it. A listener doing `await fetch(unreachableHost)` with no timeout delays every later listener by the socket timeout, and under `dispatch()` the caller has already returned, so the audit-row listener registered after it silently does not write its row for the duration. Invariant 3 is written about failures, and a throw is handled correctly, so this is not a violation — but it is the one way listener independence is not total, and it is the difference `Promise.allSettled` would make. Worth a sentence in the `dispatchAndWait` doc comment if the sequential choice is deliberate, which the handoff says it is.

**Verdict:** no blockers — all five invariants hold as implemented, and the declared-wins/`[]`-is-declared distinction is read from the genuinely raw slice.

nstfkc:
## Review — plan-conformance

Walked every numbered deliverable in `plans/events/01-sync-dispatch.md` §1–§7 and every bullet in its **Tests** section against the diff. All seven deliverables are present with the sketched shapes and signatures (`Event.dispatch`/`dispatchAndWait` through `ConstructorParameters`, `EventManager` with `static token = "events"`, `Record<string, ListenerClass[]>`, `warnedFor`, the `registeredListeners` getter reporting what was *handed* in, `discoverListeners` with both descriptor-based name checks, the `events` config slice with no `eventsDir`, `EventServiceProvider` reading the **raw** slice in `boot()` and registered immediately after `QueueServiceProvider` in `kernel/providers.ts`, and the six §7 barrel exports).

All eleven Tests bullets exist, including the two easiest to skip: `discovery.test.ts:706` (`"the two spellings are told apart exactly, not guessed at"`) pins the descriptor-vs-string distinction, and `EventManager.test.ts:145` covers the `dispatch` half of the isolation bullet via an `unhandledRejection` listener rather than only the `dispatchAndWait` half. Both `.test-d.ts` cases are there and non-vacuous — the `@ts-expect-error` on `static event = NotAnEvent` only fires because of the `__isEvent` marker, so that undeclared addition is load-bearing rather than decorative.

**Out of scope respected.** `grep` over `packages/gemi/services/events/` and `docs/events.md` finds no `afterCommit`, no `fake`, no wildcard, no priority, and no `queued` field. No template changes. No unrequested runtime surface.

Two nits, both plan-internal contradictions the author resolved and disclosed:

- **nit — `packages/gemi/services/events/Listener.ts:53`.** §2s sketch reads `queued = false;` and the shipped class has no such field. The same documents **Out of scope** opens with "`queued`", so the plan disagrees with itself; `02-queued-listeners.md`s **Prerequisite state** does not list the field among what it inherits ("Every dispatch goes through `EventManager`, listeners are discovered from `app/listeners`, and `Event` declares `static name`"), so nothing downstream is left dangling. Shipping an inert `queued = true` would have been the silent no-op invariant 4 exists to prevent. Agreed as-is; noted only so the omission is not read as an oversight.
- **nit — `packages/gemi/services/discovery.ts:279` (`refuseListenersWithNoEvent`).** §4 says a listener whose `static event` is missing "is refused here rather than warned about", which this implements as a boot-stopping `DiscoveryError`. That is narrower than READMEs **Decisions, open**: "*What happens to an abstract listener base that lives in `app/listeners`?* It gets discovered and registered". Concretely: an app with `app/listeners/BaseNotifier.ts` — abstract, no `static event`, each subclass declaring its own — boots per the README paragraph and fails to boot here. §4 is the more specific clause, the throw names the classes and the fix, and `docs/events.md:139-143` documents the new behaviour, so this reads as the right resolution; it is worth a line because the README paragraph it narrows is now stale.

Nothing was silently dropped, narrowed, or swapped for something easier, and no unrequested scope beyond the three additions the handoff declares.

**Verdict:** conforms to the plan; two nits, no blockers.

nstfkc:
## Review — integration

Checked the diff against the rest of the repo rather than against itself. The wiring is clean: `EventManager.token = "events"` collides with none of the 22 tokens already bound (`grep 'static token'`); the new `// Events` block in `services/index.ts` publishes every name `docs/events.md` imports (`Event`, `Listener`, `defineEventConfig`, `EventManager`, `discoverListeners`), so `docs-imports.test.ts`'s "every named import exists on the entrypoint" pass holds, and its pinned module set is unchanged; `barrel-imports.test.ts` is unaffected — `Listener.ts` imports `Event.ts` type-only, so `discovery.ts` does not drag `foundation/app` or `EventManager` in, and nothing new reaches a `HEAVY` package; the `providers.ts` count is right (16 providers, 17 services, `RouteServiceProvider` still owning two); the `llms-full.txt` copy of `events.md` is byte-identical to the page modulo the H1→H2 demotion the file already uses. No reformatted files, no `templates/` edits, no `bun.lock`.

Three things.

**1. `minor` — `packages/gemi/services/events/Event.ts:93`, `:120`, `:138`: the class-side dispatch API has no runtime test at all.**
Every test in the PR goes through `EventManager` directly — `manager.dispatchAndWait(new UserRegistered(...))` in `EventManager.test.ts`, and `application.make(EventManager).dispatchAndWait(new UserRegistered(...))` in the one end-to-end case at `services/discovery.test.ts:909`. Nothing calls `UserRegistered.dispatch(...)`, and `grep 'Cannot dispatch an event'` matches only `Event.ts` itself. Concretely: replace both static bodies with `{}` / `Promise.resolve()`, or point them at `app(QueueManager)`, or drop the spread in `new this(...args)` — the suite stays green. That leaves the exact line the plan's Goal is written around (`UserRegistered.dispatch(user.id, user.email)`) and the "cheap half of invariant 1" (`refuseUnnamed`) untested, including the container resolution, which is the one part of the subsystem that can break from a change made elsewhere. `http/RateLimitMiddleware.test.ts` and `support/Service.test.ts` both show the `kernelContext.run` / `setInstance` pattern that makes this reachable, and `discovery.test.ts` already builds a real `Application`. (`Job.dispatch` is untested too, so this is precedent rather than a repo rule — but it is precedent from before the plan documents that this file's comments spend paragraphs on.)

**2. `minor` — `docs/project-structure.md:88` (and its copy at `docs/llms-full.txt:337`): "the three discovered directories" is now four.**
That heading and the paragraph under it enumerate exactly which directories are read at boot, and the next paragraph is the warning that matters: "a file in one of these directories that does work when imported does that work when the directory is read". `app/listeners` is now walked in `EventServiceProvider.boot()` on every start, in dev and production, and the canonical page that tells a reader which directories cost that says it is not. `docs/events.md` repeats the warning locally, but a reader auditing boot cost goes to `project-structure.md`. The `### The slice reference` table at `:262` is also missing an `events` / `defineEventConfig` row — weaker, since `command` / `defineCommandConfig` is already absent from it.

**3. `nit` — `docs/llms-full.txt:63`:** the hunk that inserts the Events entry sits in the embedded copy of the docs map that is still missing its `Commands` line (present in `docs/README.md:62`, `docs/llms.txt:48`, `docs/index.html:131`). Noted in the handoff as pre-existing and out of scope; it is one line, in a list this PR is already editing, and leaving it means the next person to touch that map re-discovers it.

Verdict: no blockers — the integration surface (barrel, docs imports, token, provider count, scope) is correct; the untested class-side `dispatch` is the one gap worth closing before merge.

nstfkc:
## Review addressed

Pushed as `9c4e8c8`. Six of the seven findings led to a change; one was declined, and one of the changes went slightly further than asked — both called out below.

### Changed

**`minor` — `Event.ts:93`: the class-side dispatch API had no runtime test.** Correct, and the mutants named in the review really did survive. New `packages/gemi/services/events/Event.test.ts` drives the API an application actually calls, through a real container:

- `UserRegistered.dispatch(7, "ada@example.com")` inside `kernelContext.run(application, …)`, asserting the listener received an instance with `userId === 7` — dropping the spread in `new this(...args)` now fails two tests (verified by making that edit).
- Two applications, two contexts, one event class: the manager is resolved from the context the call was made in rather than captured at module scope, and a `dispatch` pointed at any other token fails to resolve at all.
- `dispatchAndWait` blocked on a deferred, asserting the caller settles after the listener does.
- `refuseUnnamed` for both spellings that reach it (below).

**`nit` — `Event.ts:133`: the `refuseUnnamed` comment overstated its coverage.** Confirmed exactly as reported — `Object.getOwnPropertyDescriptor(class extends Base {}, "name")` is `{ value: "", writable: false }`, so the inherited `"unset"` was never consulted and the dispatch went out under the key `""`.

I fixed the code as well as the comment rather than only the comment: the check is now `event.name !== "unset" && event.name !== ""`, because `""` is no more routable than `"unset"` and the behaviour the comment described is the one worth having. `""` reaching a real registry key was the failure — the generic zero-listener warning instead of the message naming the fix. The comment now says what each spelling produces, and that this is still only the floor, with `discoverListeners`' descriptor read the real check.

**`nit` — `EventManager.ts:173`: sequential execution was undocumented.** Kept sequential — parallel would make "registration order", the thing `registeredListeners` reports and the discovery tests assert on, a claim about start order only. But the review is right that the cost was nowhere on the page. `dispatchAndWait` gains a `### One at a time, deliberately` section stating the hazard in the review's own terms (a listener that *hangs* rather than throws delays every listener after it; under `dispatch` the caller has already moved on), and distinguishing it from a throw, which is caught. `docs/events.md` gains the same paragraph under **Dispatching**.

**`minor` — `docs/project-structure.md:88`: "the three discovered directories" is now four.** Heading, enumeration, the "differs in **when** it is read" paragraph and the See-also line all now include `app/listeners`, plus one sentence noting it is the one directory the scaffold does not create — absent until you write the first listener, which is not an error (the tree above that section describes the `saas-starter` template, and templates are out of scope for this PR, so the tree itself is unchanged and still accurate).

The `### The slice reference` table gains an `events` / `defineEventConfig` row — **and a `command` / `defineCommandConfig` row**, which the review flagged in the same breath as already absent. That second row is outside this PR's subject; I added it because it is one line in a table I was already editing and the alternative is the next person re-deriving it. Say the word and I will drop it.

Mirrored into `docs/llms-full.txt` (both the embedded `project-structure.md` copy and the embedded `events.md` copy).

**`nit` — `docs/llms-full.txt:63`: the embedded docs map is missing `Commands`.** Added. It was also missing **`Services`**, which nobody flagged — diffing the map against `docs/README.md:26-67` now comes back clean apart from the trailing heading, so a partial fix would have left the same defect for the next person. Both pages' full text is still absent from `llms-full.txt`; that is a much larger pre-existing gap and I have not touched it.

**`nit` — `discovery.ts:279` narrows the README's documented residual.** Agreed on the resolution: §4's "refused here rather than warned about" is the more specific clause and the code stays as shipped. What I changed is the stale paragraph, in both places it appears — `plans/events/README.md`'s **Decisions, open** entry and §4's **Known residual** in `01-sync-dispatch.md` now split the case by whether the base declares `static event`: without one the boot stops with a message naming the class and the fix; with one it is registered and run like any other listener, which is the residual jobs and cron share.

**This is an edit to the plan, in the PR that implements it, so flagging it loudly.** The normal handling would be a separate correction, but both files are *added* by this PR, so "separately" would mean a follow-up editing a file this commit introduces with a known contradiction against its own §4. No behaviour, test or doc changed with it. Happy to revert the two paragraphs and take it to a plan-only PR if that is preferred.

### Declined

**`nit` — `Listener.ts:53`: no `queued = false` field.** Declining, as the review itself concludes. §2's sketch and the **Out of scope** list contradict each other, `02-queued-listeners.md`'s **Prerequisite state** does not list the field among what it inherits, and shipping an inert `queued = true` — a flag that reads as a context boundary and silently runs the listener in the request anyway — is precisely the silent no-op invariant 4 exists to prevent. It arrives in iteration 2 with the routing that makes it mean something.

### Verification

From `packages/gemi`:

- `bun --bun vitest run services/events/ services/discovery.test.ts docs-imports.test.ts barrel-imports.test.ts exports-map.test.ts` — 77 passed (19 in `services/events/`, six of them new).
- `bun --bun vitest run --typecheck.only services/events/Listener.test-d.ts` — 4 passed, no type errors.
- `bunx tsc --noEmit -p tsconfig.json` — clean.
- `bunx oxlint` — 80 warnings, 0 errors; identical to the count with the three changed source files stashed, so the new test file adds none.
- `oxfmt --check` on `services/events/Event.ts` fails at the **base** commit too (one signature it wants on a single line), so it was left alone per the standing rule; the new test file matches the 80-column style of `EventManager.test.ts` beside it.

Not re-measured: still nothing exercised against a real minified `gemi build`, so the production-only hazard the design turns on rests on the descriptor test and the existing `Bun.build({ minify: true })` finding recorded in `discovery.test.ts`.

PR #418, opened 2026-08-17
Events iteration 2: queued listeners

nstfkc:
**Stacked on #417** (`feat/events-01-sync-dispatch`) — review that one first; this PR's base is its branch, so the diff here is only iteration 2.

Implements [`plans/events/02-queued-listeners.md`](../blob/feat/events-02-queued-listeners/plans/events/02-queued-listeners.md).

## What it delivers

`queued = true` on a listener moves it off the request and onto the existing queue. Retries, `maxAttempts`, dead-lettering, `concurrency` and worker threads come from `QueueManager` rather than from anything written here — a queued listener is registered as a synthetic `Job` named `listener:<name>`, and a dispatch pushes `["UserRegistered", [7, "ada@example.com"]]` to it.

```typescript
export class SendWelcomeEmail extends Listener {
  static name = "SendWelcomeEmail";
  static event = UserRegistered;

  queued = true;
  maxAttempts = 5;

  async handle(event: UserRegistered) { await sendWelcome(event.email); }
}
```

- **`QueueManager.registerJob(job)`** — a single-job entry point beside `useJobs`, sharing its first-claim-wins collision handling and appending to `config.jobs` so `registeredJobs` still reports everything the manager was handed. `useJobs` replaces the registry wholesale, so `EventServiceProvider.boot()` (which runs after the queue's) could not use it without discarding every job the app wrote.
- **`Listener` gains `queued`, `maxAttempts`, `worker`** — the last two forwarded to the synthetic job, and documented as ignored when `queued` is false on the fields themselves, because a `maxAttempts = 5` on a sync listener is a line an author wrote that does nothing.
- **`static name` becomes required in earnest for a queued listener** — an implicit class name is refused at registration (a throw), not warned about. A sync listener still only warns: its name never leaves the source-side walk, while a queued one's is the queue's key on both ends, and those ends can be two different builds.
- **`EventManager.rehydrate(eventName, args)`** — a second registry (event name → class, from each listener's `static event`) and `new EventClass(...args)`. An unknown name throws with the name in the message; the payload is already off the queue by then and the alternative is `event.email` failing several frames later.
- **`EventManager.dispatch/dispatchAndWait` take the constructor arguments**, which `Event.dispatch` forwards. An instance cannot be asked what it was built with, and the arguments are what is serialized.
- **Invariant 4 documented** — in `Listener.queued`'s doc comment and in `docs/events.md` as its own subsection, plus the durability note (`a dispatch is not a durability guarantee`) and the `listener:` prefix showing up in `registeredJobs`.

## Verification

Run from `packages/gemi` in the worktree:

```
$ bun --bun vitest run services docs-imports.test.ts barrel-imports.test.ts
 Test Files  28 passed (28)
      Tests  355 passed (355)

$ bun --bun vitest run --typecheck.only services/events
 Test Files  1 passed (1)
      Tests  4 passed (4)
 Type Errors  no errors

(no output)

$ bunx oxlint
Found 80 warnings and 0 errors.   # unchanged from the base branch; none in the files touched
```

New tests: `services/events/queuedListeners.test.ts` (12) covering sync + queued in one dispatch, `dispatchAndWait` not waiting, retry-to-dead-letter on the listener's own `maxAttempts`, the rehydration round trip, the worker's `dispatchJob` entry point, the synthetic job's writable `name`, `worker` forwarding, the implicit-name refusal, two queued listeners claiming one name, a payload JSON cannot carry, and a manager-level dispatch that cannot queue. Plus `rehydrate` in `EventManager.test.ts` and three `registerJob` cases in `QueueManager.test.ts`.

## Deviations from the plan

1. **The constructor arguments are passed to the manager, not recovered from the event.** The plan writes `push(job, JSON.stringify([event.constructor.name, ctorArgs]))` without saying where `ctorArgs` comes from — an event instance does not remember them, and the base class cannot capture them (`super()` runs before the subclass assigns its fields). So `dispatch(event, args?)` takes them and `Event.dispatch` forwards them. A queued listener reached without them is skipped with a line on stderr naming both, rather than queueing `[]` and rehydrating an event whose every field is `undefined`.
2. **The name escalation in §3 landed as a refusal for *queued* listeners only**, in `jobForListener`, rather than as a change to `warnIfListenerNameWillNotSurviveTheBuild`. Discovery cannot tell a queued listener from a sync one without constructing every class it found; the registry constructs them anyway, and the plan's own sentence for the escalation ("a queued listener whose name is implicit is a side effect that stops happening in production only") is about queued ones. Sync listeners keep the iteration-1 warning.
3. **The synthetic job's `run` reads two payload shapes.** `QueueManager.run` spreads the parsed arguments while `QueueManager.dispatchJob` — the worker thread's entry point — passes the parsed array whole. That asymmetry is pre-existing and affects every job with arguments, not just these; it is worked around in `readPayload` rather than fixed, because fixing it changes what every existing `worker = true` job receives. Worth its own issue.
4. **The context boundary is documented as "not promised" rather than "gone".** gemi's queue is in-process and `push` starts the drain itself, so a queued listener frequently runs inside the dispatching request's `kernelContext` — `app()` there resolves the request's application. Claiming isolation the runtime does not provide would be the more dangerous doc, so both the field comment and `docs/events.md` say the context may be there and may not, and that nothing may rely on it.

## Also worth a reviewer's eye

- `EventServiceProvider.boot()` no longer returns early when `listeners` is declared in config — it skips discovery but still hands the queued listeners' jobs to the queue, otherwise `queued = true` would silently do nothing for an app that lists its listeners.
- The queue push is wrapped in its own `try`: an argument JSON cannot serialize, or an application with no queue bound, would otherwise reject `dispatchAndWait`, which is documented never to.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,7 +60,7 @@ See **[Getting Started](./getting-started.md)** for requirements and a first rou
 - **[Jobs & Queues](./jobs-and-queues.md)** — defining and dispatching background `Job`s.
 - **[Cron](./cron.md)** — scheduling recurring `CronJob`s.
 - **[Commands](./commands.md)** — one-off application commands run with `gemi run`.
-- **[Events & Listeners](./events.md)** — `Event.dispatch(...)` fanning out to the listeners under `app/listeners`.
+- **[Events & Listeners](./events.md)** — `Event.dispatch(...)` fanning out to the listeners under `app/listeners`, inline or on the queue.
 - **[Broadcasting](./broadcasting.md)** — websocket channels, the `Broadcast` facade, `useSubscription` / `useBroadcast`.
 - **[Internationalization](./i18n.md)** — component-scoped dictionaries, `useTranslator`, `useLocale`, locale detection.
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -177,6 +177,8 @@ If the name on the payload resolves to no registered event, the job throws with 
 
 These are the queue's, exactly as [Jobs & Queues](./jobs-and-queues.md) documents them: when `handle` throws, the listener is re-queued until it has been attempted `maxAttempts` times, and the last failure dead-letters it. A sync listener has none of this — its throw is logged once and the next listener runs.
 
+Every attempt that throws is written to stderr with the listener's name, the event it is bound to and the error, and the dead-letter is written separately, because "it failed again" and "it will not be tried again" are different things to know. By then `dispatchAndWait` has long since resolved and no caller is left to tell, so those lines are the whole of what says a side effect stopped happening.
+
 > **Note:** The queue is **in-memory and in-process**. A queued listener that has not run when the process exits does not run. That is true of every job today, but an event system invites "dispatch it and it will happen", so it is worth saying plainly: a dispatch is not a durability guarantee. Work that must survive a restart wants a row somewhere that a later pass can pick up.
 
 `dispatchAndWait` does **not** wait for a queued listener. When it resolves, every sync listener has settled and the queued ones have been handed to the queue — which may mean not started, or already failed once and re-queued.
--- a/docs/events.md
+++ b/docs/events.md
@@ -10,7 +10,7 @@ UserRegistered.dispatch(user.id, user.email);
 
 That is the whole of what this buys: adding a fourth side effect to a registration becomes **adding a file**, rather than editing the controller that already has three.
 
-Listeners run **in-process and synchronously** inside the dispatching request — see [When to reach for a job instead](#when-to-reach-for-a-job-instead).
+Listeners run **in-process and synchronously** inside the dispatching request by default. One line moves a listener onto the queue instead — see [Queued listeners](#queued-listeners), and read it before you write it: `queued = true` changes what the listener can see, not only when it runs.
 
 ## Defining an event
 
@@ -83,6 +83,9 @@ There is no `static events = [A, B]`, deliberately. A listener bound to two even
 | `static name` | `string` | `"unset"` | Unique listener identifier. Required. |
 | `static event` | `Event` subclass | — | The one event this listener handles. Required. |
 | `handle` | method | — | The side effect. May be `async`. Required. |
+| `queued` | `boolean` | `false` | Run on the queue instead of inline. A [context boundary](#queued-listeners), not a performance dial. |
+| `maxAttempts` | `number` | `3` | Attempts before dead-lettering, counting the first. Ignored unless `queued`. |
+| `worker` | `boolean` | `false` | Run `handle` in a Worker thread with its own cloned app. Ignored unless `queued`. |
 
 ## Dispatching
 
@@ -102,9 +105,9 @@ When the caller genuinely needs the side effects to have happened first — a re
 await UserRegistered.dispatchAndWait(user.id, user.email);
 ```
 
-It resolves once every listener has settled. Two methods rather than one that is sometimes worth awaiting, because "should I await this?" is a question the call site should not have to guess at.
+It resolves once every **sync** listener has settled — not the [queued](#queued-listeners) ones, whatever the name suggests. Two methods rather than one that is sometimes worth awaiting, because "should I await this?" is a question the call site should not have to guess at.
 
-Listeners run **one at a time**, each awaited before the next one starts, so registration order is the order they run in and not merely the order they start. The cost is worth knowing: a listener that *hangs* — an un-timed-out request to a host that never answers — holds up every listener after it for as long as it hangs, and under `dispatch` the caller has already moved on, so the row a later listener writes is simply not there yet. A listener that can block indefinitely wants its own timeout, or a [job](./jobs-and-queues.md). A listener that *throws* is not this case — see below.
+Sync listeners run **one at a time**, each awaited before the next one starts, so registration order is the order they run in and not merely the order they start. A [queued](#queued-listeners) listener is pushed at the place it occupies in that order and not waited for, so its *dispatch* is ordered with the rest and its execution is the queue's business. The cost is worth knowing: a listener that *hangs* — an un-timed-out request to a host that never answers — holds up every listener after it for as long as it hangs, and under `dispatch` the caller has already moved on, so the row a later listener writes is simply not there yet. A listener that can block indefinitely wants its own timeout, or a [job](./jobs-and-queues.md). A listener that *throws* is not this case — see below.
 
 Neither method ever rejects. See below.
 
@@ -116,6 +119,8 @@ That is not politeness, it is the point. Listeners are independent side effects,
 
 A listener that must not fail silently should handle its own errors — retry, log to your own sink, or dispatch a [job](./jobs-and-queues.md) that can be retried and dead-lettered.
 
+All of this is the **sync** path. A [queued](#queued-listeners) listener's throw never reaches the dispatcher at all: it is the queue's, and it is retried up to `maxAttempts` and then dead-lettered.
+
 ## When nothing is listening
 
 A dispatch with no listeners is legal and does nothing. In development it warns **once per event name**:
@@ -125,6 +130,70 @@ A dispatch with no listeners is legal and does nothing. In development it warns 
 ```
 
 That single line is the whole early-warning system here, because four different mistakes produce the same symptom: an event that never declared `static name`, a typo, a listener that was never discovered, and a listener directory that was never walked. It fires once per name so that a dispatch in a loop does not bury the terminal, and never in production, where "no listeners" is an ordinary answer.
+
+## Queued listeners
+
+`queued = true` moves a listener off the request and onto the [queue](./jobs-and-queues.md), with retries, `maxAttempts`, dead-lettering and worker threads coming from there rather than from anything events add:
+
+```typescript
+// app/listeners/SendWelcomeEmail.ts
+export class SendWelcomeEmail extends Listener {
+  static name = "SendWelcomeEmail";
+  static event = UserRegistered;
+
+  queued = true;
+  maxAttempts = 5;
+
+  async handle(event: UserRegistered) {
+    await sendWelcome(event.email);
+  }
+}
+```
+
+That is the whole of the opt-in. Behind it, the listener is registered with the `QueueManager` as a job named `listener:SendWelcomeEmail`, and a dispatch pushes `["UserRegistered", [7, "ada@example.com"]]` onto the queue instead of calling `handle`.
+
+`static name` stops being a label here and becomes the queue's key, so a queued listener that does not declare one is **refused at boot** rather than warned about — see [Two listeners, one name](#two-listeners-one-name) for what the name is doing, and the note under [Defining an event](#defining-an-event) for why an implicit class name is not the same string after a production build.
+
+### What `queued = true` changes — read this part
+
+It is a **context boundary spelled as one line**, not a latency setting.
+
+- A **sync** listener runs inside the dispatching request: it has that request's `app()`, its authenticated user through `currentActor()`, and it joins the ambient transaction.
+- A **queued** listener is handed to the queue and run from a drain, on the queue's terms. With `worker = true` that is a different thread with a cloned application. Only the event's name and constructor arguments cross; nothing else does.
+
+The part that catches people is that the request's context is not reliably *gone* either. gemi's queue is in-process, so a drain that happens to start from the push is still standing inside the dispatching request, and `app()` there resolves the request's application. **None of that is promised.** A queued listener that reads the current actor, or writes expecting to join the transaction the controller opened, works right up until the day the queue was already busy — and then reads different rows, or commits separately, with no error either way.
+
+So the rule is simple: a listener that needs anything from the request stays sync; a listener that needs only the payload is free to queue.
+
+### What crosses the queue
+
+The **constructor arguments** are what is serialized, not the instance — the same bargain [`Job.dispatch`](./jobs-and-queues.md) makes. The payload is JSON, so pass plain, serializable data: numbers, strings, plain objects. A model instance, a `Date`, a function or anything holding a database handle does not survive it.
+
+On the far side the event is rebuilt with `new UserRegistered(...args)`, which means **an event constructor that does work does it again**, in whatever process runs the listener and without the request around it. Keep event constructors to assigning fields.
+
+If the name on the payload resolves to no registered event, the job throws with that name in the message rather than handing the listener `undefined` — the failure surfaces where it can still be explained, and takes the queue's ordinary retry path.
+
+### Retries, dead-lettering and durability
+
+These are the queue's, exactly as [Jobs & Queues](./jobs-and-queues.md) documents them: when `handle` throws, the listener is re-queued until it has been attempted `maxAttempts` times, and the last failure dead-letters it. A sync listener has none of this — its throw is logged once and the next listener runs.
+
+> **Note:** The queue is **in-memory and in-process**. A queued listener that has not run when the process exits does not run. That is true of every job today, but an event system invites "dispatch it and it will happen", so it is worth saying plainly: a dispatch is not a durability guarantee. Work that must survive a restart wants a row somewhere that a later pass can pick up.
+
+`dispatchAndWait` does **not** wait for a queued listener. When it resolves, every sync listener has settled and the queued ones have been handed to the queue — which may mean not started, or already failed once and re-queued.
+
+### Queued listeners in `registeredJobs`
+
+They are visible, under the prefix:
+
+```typescript
+import { app } from "gemi/foundation";
+import { QueueManager } from "gemi/services";
+
+app(QueueManager).registeredJobs.map((job) => job.name);
+// ["SendInvoiceEmail", "listener:SendWelcomeEmail"]
+```
+
+`registeredJobs` reports what the queue will run, and a queued listener genuinely is something it will run — hiding them would make a queue introspection tool lie. The `listener:` prefix is what keeps you from going to look for a job file you never wrote.
 
 ## Registering listeners — `app/listeners/`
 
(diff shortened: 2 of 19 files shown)


nstfkc:
## Review — plan-conformance

Walked all six numbered deliverables and all eight **Tests** bullets in `plans/events/02-queued-listeners.md`. Deliverables 1, 2, 4 and 6 land as written (`registerJob` + shared `claim()` + `config.jobs` append; `maxAttempts`/`worker` at `3`/`false` with the "ignored unless queued" doc comment on each; the second `events` registry populated at `useListeners` time with the named throw; the in-memory-queue durability note). All eight test bullets exist. Nothing from iteration 3 (`afterCommit`, `Event.fake()`) leaked in, `listener:` entries are shipped visible as the **Open** section decided, and the provider order the plan depends on is still `QueueServiceProvider` → `EventServiceProvider` (`packages/gemi/kernel/providers.ts:42`).

Three things, all small.

**1. minor — deliverable 3 is implemented for queued listeners only.**
`packages/gemi/services/discovery.ts:334`

The plan's heading is "**`static name` on `Listener` becomes required**", and the body asks to "upgrade the iteration-1 `warnIfListenerNameWillNotSurviveTheBuild` from advisory to something an author cannot miss". README's invariant 1 says the same: "required on `Event`, and required on `Listener` from iteration 2".

`warnIfListenerNameWillNotSurviveTheBuild` is unchanged apart from its doc comment; the refusal moved to `refuseImplicitName` in `listenerJob.ts`, which only runs for a listener whose `queued` is true. Concretely: a listener under `app/listeners` with `static event = UserRegistered`, no `static name`, and no `queued` still boots on a `console.warn` and registers under its minifiable class binding — "required" is not enforced for it. This is a declared deviation and the plan's own justification sentence is about queued listeners, so it may well be the right call; flagging it because the deliverable heading says otherwise and the plan text should be amended if this is what ships.

**2. nit — the "unknown event name on the queue" test never touches the queue.**
`packages/gemi/services/events/EventManager.test.ts:316`

Plan: "**An unknown event name on the queue throws**, with the name in the message." The test calls `manager.rehydrate("OrderPaid", [])` directly. The message and the throw are covered; the path the plan names — a payload already off the queue reaching the synthetic job's `run` — is not, and `queuedListeners.test.ts` has no equivalent. The rest of the file drives a real `Application` + `QueueManager`, so a `dispatchJob("listener:X", JSON.stringify(["OrderPaid", []]))` case would cost one test and cover the boundary the plan is actually pointing at.

**3. nit — the docs contradict invariant 4's stated model (deliberately).**
`docs/events.md:164`, mirrored in `Listener.queued` and `EventManager`'s class doc.

Plan deliverable 5: "A queued listener runs later, from `QueueManager.next()`, **outside all of it**"; README invariant 4: "A queued listener runs in a cloned app with none of that." The docs say the opposite — "the request's context is not reliably *gone* either… `app()` there resolves the request's application" — because the in-process drain frequently starts from `push`. The normative half of invariant 4 ("must not be documented as a performance toggle") is honored everywhere, and the softening is the truthful description of what the runtime does, so this is not a defect in the code. It does mean README invariant 4 is now wrong about its own subsystem and should be reworded, or iteration 3 should clear the context on drain and the docs tightened back.

**Verdict:** the plan is implemented; one numbered deliverable (3) is narrower than its heading and two test/doc details drift from the plan's wording — nothing here blocks the merge.

nstfkc:
## Review — invariants

All five hold. Traced them rather than reading the comments:

1. **Keyed by declared name.** `EventManager.useListeners` keys on `listener.event.name` (`EventManager.ts:161-165`) and the queue on `job.name` (`QueueManager.push`), both strings. No `Map`/`Set` keyed by a class object, no `instanceof`, no `===` between class references anywhere in `services/events/*.ts` — the only class references held are the direct closure captures in `jobForListener`, which are same-graph by construction. `refuseImplicitName` (`listenerJob.ts:101`) reads the `name` property descriptor, which is the right discriminator, and correctly fires only for queued listeners.
2. **Zero-listener warning.** `warnNothingIsListening` (`EventManager.ts:383`) is dev-only, once per event name, per manager. Unchanged by this iteration.
3. **One listener's failure never stops another.** Sequential `for` loop, `new Listener()` *inside* the try (`EventManager.ts:293-302`); the queued branch's `push` has its own try/catch (`EventManager.ts:371-380`). No shared `Promise.all`. Nothing that escapes into the loop: `QueueManager.next`/`run` are `async`, so a listener's throw during the synchronous drain that `push` kicks off becomes a rejection inside `run`'s own catch, not a sync throw back through `enqueue`.
4. **Context difference documented at the declaration.** `Listener.queued` (`Listener.ts:99-126`) leads with "a context boundary, not a performance dial", and `maxAttempts`/`worker` each carry the "ignored unless queued" line.
5. **Nothing in the framework dispatches.** `EventManager`/`dispatchAndWait` appear outside `services/events/` only in `services/index.ts` and tests.

The **"declared wins, and `[] `is declared"** rule is correct and reads the raw slice: `EventServiceProvider.boot():58-63` re-reads `config.get("events", {})` and tests `slice.listeners === undefined`, and `withDefaults` builds a new object rather than mutating the slice — so the `make(EventManager)` on line 57, which happens *before* the slice is read, does not stamp a defaulted `listeners: []` onto it. `[]` is honoured verbatim; absent and explicit `undefined` both discover; matches `QueueServiceProvider` exactly.

### Findings

**1. major — `packages/gemi/services/events/listenerJob.ts:58` — the `rehydrate` throw reaches nobody, so a queued listener that cannot be rebuilt fails with zero output.**

`rehydrate`'s doc (`EventManager.ts:314-320`) and `docs/events.md:174` both say the throw is the informative failure — "the failure surfaces where it can still be explained". Trace it: the synthetic job overrides `run` only, and `Job.onFail` / `Job.onDeadletter` are empty bodies (`services/queue/Job.ts:11-13`). `QueueManager.run:256-263` catches the rejection, calls `onFail(err)` (no-op), re-pushes until `retries >= maxAttempts - 1`, then calls `onDeadletter(err)` (no-op). Nothing is written to stderr, and `dispatchAndWait` resolved long before.

Concrete: a queued `SendWelcomeEmail` whose event name is not in the running app's `events` registry — the `worker = true` case, where the thread boots `dist/server/bootstrap.mjs` and discovery there resolves a different (or empty) `app/listeners` — dispatches, is attempted three times, dead-letters, and the process prints nothing at all. The welcome email stops being sent in production with no signal. That is the failure shape the plan built the throw to avoid, and the one your own test only catches because it spies on `Job.prototype.onDeadletter` (`queuedListeners.test.ts:224`); `EventManager.test.ts:325` asserts the throw by calling `rehydrate` directly, so it passes regardless.

The plan's letter is met (it does throw rather than warn, and the listener is never handed `undefined`), which is why this is not a blocker. But the doc claim is false as written. One `onFail`/`onDeadletter` override on the synthetic job, logging the listener name and the event name, would make it true — and would also give every queued listener the "logged with the event and listener names" that invariant 3 gives the sync ones.

**2. nit — `packages/gemi/kernel/providers.ts:27-30` — the ordering comment still says the dependency does not exist.**

"`EventServiceProvider` sits after `QueueServiceProvider` for a dependency it does not have yet … its `boot()` *will* want the queue's registry already populated." As of this PR it does have it. A future reader who trusts that comment and reorders the list gets `QueueServiceProvider.boot()`'s `useJobs` resetting `this.jobs = {}` after every `listener:` entry was registered, i.e. every queued listener stops running (visible only as `next()`'s "Dropped a queued job" line, after the dispatch returned). Worth updating to state the dependency as real, since it is now load-bearing.

**Verdict:** invariants 1–5 and the raw-slice rule all hold; one doc/observability gap worth closing before merge, nothing that makes the PR wrong as merged.

nstfkc:
## Review — integration

Lens: the diff against the rest of the repo. Entrypoint exports, barrel weight, provider wiring, token uniqueness, unrelated churn, whether the tests bite, comment style.

**Clean on most of it.** `docs/events.md` and `docs/jobs-and-queues.md` only import `app` from `gemi/foundation` and `QueueManager` from `gemi/services`, both already exported, so `docs-imports.test.ts` has nothing new to resolve (`jobForListener` / `listenerJobName` are internal and never named in docs). `barrel-imports.test.ts` is unaffected: `EventManager.ts`'s new module-scope imports are `foundation/app` (already reached via `Event.ts`) and `queue/QueueManager` (already re-exported by the barrel), and the `EventManager ⇄ listenerJob` cycle touches nothing at module scope. No new service token, so the kernel's "already bound" guard is not in play. No `templates/`, no `bun.lock`, no reformatting — the only stray hunk is one blank line dropped from `docs/llms-full.txt`, which is hand-maintained. And the tests are not tautologies: `queuedListeners.test.ts:262` exercises the real `dispatchJob` array shape (a job reading only the spread shape fails it), the `isRunning = true` tests genuinely observe queued-but-not-run, and the dead-letter test would not pass against an empty implementation.

Three findings.

### 1. `worker = true` on a listener cannot work — major
`packages/gemi/services/events/Listener.ts:148`, `docs/events.md:88`

`QueueManager.run` sends a `worker` job through `runInWorker`, whose worker script (`QueueManager.ts:19-28`) is:

```js
result = clone.dispatchJob(event.data.jobName, event.data.args)  // no await
clone.destroy()
self.postMessage({result})
```

`App.dispatchJob` returns `kernelContext.run(app, () => queue.dispatchJob(...))`, and `QueueManager.dispatchJob` returns `job.run(...)` — so `result` is whatever `run` returns. The synthetic job's `run` is `async` (`listenerJob.ts:38`, it awaits `rehydrate` then `handle`), so `result` is always a Promise. `postMessage` structured-clones its argument, a Promise is not cloneable, so `onmessage` throws `DataCloneError` before any message is posted; `runInWorker`'s promise never settles, `QueueManager.run` never returns, `activeRunningJobsCount` is never decremented, and neither `onSuccess` nor the retry/dead-letter path ever fires. `clone.destroy()` has torn the kernel down under the still-running `handle` in the meantime.

Concretely: `class ResizeAvatar extends Listener { queued = true; worker = true; ... }` in a real app never completes, and after `concurrency` such dispatches the queue sits in the 1-second poll loop in `next()` forever.

The root cause is pre-existing, but it is not symmetric: a job author can write a synchronous `run` and have `worker = true` work today, whereas a listener's generated `run` is unavoidably async, so the field is broken for 100% of listeners. This PR is what introduces `Listener.worker` and puts it in the fields table as a supported option. The one test that touches the flag (`queuedListeners.test.ts:361`) only asserts the boolean is copied onto the job, and the "worker's entry point" test drives `dispatchJob` in-process, which is not the worker. Either drop `worker` from `Listener` for this iteration, or say in its doc comment that it is inert until `createWorker` awaits.

### 2. `kernel/providers.ts:27` still says the dependency is hypothetical — minor

> `EventServiceProvider` sits after `QueueServiceProvider` for a dependency it does not have yet … Placing it here now saves reordering this list from a patch that is about something else.

This is that patch; the dependency is now real, and the sentence two lines above it ("no provider here depends on an earlier one having run") reads as licence to reorder. Someone who does — moving `EventServiceProvider` up for an unrelated reason — gets `EventServiceProvider.boot()` calling `registerJob` first, then `QueueServiceProvider.boot()` running `useJobs(await discoverJobs(...))`, which does `this.jobs = {}` and drops every `listener:*` entry. Nothing errors at boot; every queued dispatch then lands on `next()`'s "Dropped a queued job: nothing is registered under the name `listener:SendWelcomeEmail`" long after `dispatch` returned. `EventServiceProvider.boot()`'s own doc comment states the constraint correctly — `providers.ts` is where the person editing the order is looking, and it is the file the PR left stale.

### 3. `listenerJob.ts:72` explains a failure mode that cannot occur — minor

> And `writable: true` is what `warnIfNameWillNotSurviveTheBuild` reads to decide a name was *declared*: a synthetic class whose `name` was a non-writable own property would be reported to the author at every boot as a job they never wrote, under a fix they cannot apply.

`warnIfNameWillNotSurviveTheBuild` is module-private to `discovery.ts` and has exactly one caller, `discoverJobs()` (`discovery.ts:96`), which walks `app/jobs` for `Job` subclasses on disk. A synthetic listener job is built in memory and handed to `QueueManager.registerJob`; it never passes through discovery, so no warning could ever fire on it and `writable: true` prevents nothing. The other half of the comment ("the queue keys its registry off `job.name`") is the half that is load-bearing.

`queuedListeners.test.ts:290-295` repeats the same reasoning in a comment, so the test locks in a rationale rather than a behaviour. The plan asserts it too, so this is inherited rather than invented — but the house style in `QueueManager.ts` is a comment naming the failure it prevents, and this one names a failure that is unreachable. Either keep the assertion with the real reason (`Object.defineProperty` without `writable` leaves the class's non-writable `name`, and a future reader who removes the flag should know why it is there) or drop the claim.

**Verdict:** approve on integration once #1 is either fixed or documented as inert; #2 and #3 are comment corrections.

nstfkc:
## Review addressed

Both `major`s fixed, both `minor` comment corrections fixed, two of three `nit`s fixed, one `minor` declined with the plan amended to match. `39d25b7`.

### Fixed

**`listenerJob.ts:58` — the synthetic job's failures reached nobody (major).**
Correct, and the doc claim was false as written. The job now overrides `onFail` and `onDeadletter`. `onFail` fires per attempt and carries the error; `onDeadletter` fires once and says no attempt is left, because "it failed again" and "it will not be tried again" are different things to know and the last attempt legitimately produces both. Each line names the listener and the event it binds to — the queued half of what invariant 3 gives a sync listener, and what makes `rehydrate`'s throw the informative failure `EventManager.ts` and `docs/events.md` claim it is. The event is named from `static event` rather than parsed out of the payload: a throw out of either hook escapes `QueueManager.run`'s `catch` into a `run()` nobody awaits.

Knock-on, and the reviewer's own point made real: `vi.spyOn(Job.prototype, "onDeadletter")` no longer fires, because the override shadows it. That test now asserts stderr, which is the only thing an operator gets, and it would not pass over an implementation that writes nothing.

**`Listener.ts:148` — `worker = true` can never complete (major).**
Confirmed, exactly as traced. Fixed at the root rather than documented as inert: `createWorker`'s script now awaits `clone.dispatchJob(...)`. `postMessage` structured-clones its argument, so the un-awaited call posted a pending Promise, threw `DataCloneError` before any message went out, and left `runInWorker` unsettled — `run` never returning, `activeRunningJobsCount` never decremented, no retry, no dead-letter, and after `concurrency` such jobs a queue stuck in `next()`'s one-second poll forever.

Dropping the field was the other option and the plan asks for it (deliverable 2), so the defect is in the framework rather than the plan. The asymmetry argument is what decided it: a `Job` author can write a synchronous `run`, a listener author cannot. The await earns two more things — a rejecting job now reaches the `catch` and takes the queue's ordinary retry path instead of being lost with the promise, and `clone.destroy()` now runs after the job rather than under it. **Not covered by a test**: `runInWorker` needs a real bootstrap entry point, and nothing in the repo drives it today.

**`kernel/providers.ts:27` — the ordering comment (minor, raised twice).**
Rewritten to state the dependency as real and load-bearing, with the failure spelled out: `QueueServiceProvider.boot()`'s `useJobs` does `this.jobs = {}`, so reordering registers every `listener:*` entry and then discards it, boot succeeds, and each dispatch lands on `next()`'s "Dropped a queued job" long after `dispatch` returned. It also no longer reads as licence to reorder — "no provider here depends on an earlier one having run" now carries its exception.

**`listenerJob.ts:72` — the `writable: true` rationale (minor).**
Right, `warnIfNameWillNotSurviveTheBuild` is module-private and only ever walks `app/jobs`; a synthetic class cannot reach it. The comment now says nothing reads the flag today, and gives the reason to keep it: `Object.defineProperty` keeps the attributes it is not given and a class's own `name` is non-writable, so dropping the flag silently produces a descriptor that every "was this name declared?" check in the framework — they all compare descriptors, not strings — reads as an implicit binding. One such check pointed at the queue's registry rather than a directory would report these as jobs the author never wrote. The test comment says the same thing, and the plan's version of the claim is corrected too.

**`EventManager.test.ts:316` — the unknown-name test never touched the queue (nit).**
Added to `queuedListeners.test.ts`: a payload naming `OrderPaid` is pushed onto the real queue and reaches the synthetic job's `run` in a process whose event registry does not hold it, then dead-letters with the name in the message. That is the `worker = true` shape the bullet points at. The direct `rehydrate` test stays — it pins the message.

**`docs/events.md:164` — invariant 4 (nit).**
As diagnosed: the docs are truthful and README invariant 4 was not. Reworded there rather than in the docs — "may rely on none of it", with the reason an intermittently present context is the *worse* hazard, not the milder one, and why the invariant is normative rather than descriptive. Deliverable 5's "outside all of it" bullet carries the same correction. If iteration 3 clears the context on drain, both go back the other way.

### Declined

**`discovery.ts:334` — deliverable 3 enforced for queued listeners only (minor).**

Declined as a code change; the plan is amended instead.

The plan is internally split here: the heading says "`static name` on `Listener` becomes required", while the sentence under it and invariant 1's own parenthetical both scope the requirement to the queue's JSON boundary — "where the listener name crosses" it. Only a queued listener has one. A sync listener's name is read for de-duplication at registration and for error messages, both from the same source-side walk that discovered the class, so a minified build and that walk cannot produce two strings for it. Refusing it would stop a boot over a failure that cannot occur, which is a worse trade than the warning it replaces.

What did change: the warning message now says why it is a warning and what would change it — that nothing reads the name across a build boundary while the listener is sync, and that adding `queued = true` is refused at boot for the same reason. An author who later flips the flag meets the refusal at the moment the hazard becomes real, with the fix in the message.

Plan amended in both places: deliverable 3's heading is now "`static name` on a **queued** `Listener` becomes required" with a note on the narrowing, and invariant 1 says the same and points at where the refusal lives.

### Verification

`bunx tsc --noEmit`, `bunx oxlint` (80 pre-existing warnings, unchanged), `prettier --check` on every touched TS file, and `vitest run services/ kernel/ foundation/` plus the docs tests — 366 + 135 passing. No `oxfmt` run: `listenerJob.ts`, `QueueManager.ts`, `discovery.ts` and both test files already fail it at HEAD.

nstfkc:
Superseded by #420 — same branch, rebased onto `main`. GitHub auto-closed this PR when its base branch `feat/events-01-sync-dispatch` was deleted on merge of #417, and refuses to reopen it because the head branch was force-pushed afterwards. The three lens reviews and the author's response above remain the review record for iteration 2.

PR #419, opened 2026-08-17
Events iteration 3: after-commit dispatch and testing

nstfkc:
Third and last iteration of the events subsystem, per
[`plans/events/03-after-commit-and-testing.md`](../blob/feat/events-03-after-commit-and-testing/plans/events/03-after-commit-and-testing.md).

> **Stacked PR.** Base is `feat/events-02-queued-listeners` (#418), not `main`.
> Review that one first; this diff is only iteration 3.

## Part one — `static afterCommit`

An event declaring `static afterCommit = true` has its **whole** fan-out — sync
listeners and queued ones alike — held until the surrounding transaction
commits, and dropped if it rolls back. Outside a transaction it changes nothing.

The failure it exists for is silent: a welcome email sent from inside a
transaction that then rolls back leaves a user who does not exist and has been
welcomed. The email succeeded and the transaction failed, in two subsystems that
never meet, so nothing anywhere reports it.

`withTransaction` in `orm/context.ts` is the single implementation behind both
`Model.transaction` and `DB.transaction`, so there is one place to hook:

- The **outermost** scope carries the callback list. A savepoint's scope is a
  spread of its parent's, so it shares the list by reference — which is wanted:
  an outer rollback discards everything, savepoint entries included, because
  nothing ever drains it.
- The case sharing gets wrong is a savepoint that **rolls back** under a
  transaction that commits. The savepoint branch marks the list length on the
  way in and truncates on failure. **An index, not a per-entry depth tag**: two
  sibling savepoints leave entries at the same depth, adjacent in one list, so
  "drop everything at or below the failing depth" also discards the committed
  sibling's work. That implementation passes every other test in the file, hence
  the test named for it.
- Draining runs after `begin` resolves, on the fulfilled path only, inside an
  explicit `ormContext.run` scope carrying no `tx` — a listener that runs after
  commit must not join a transaction that has closed, and Bun's handle stays
  callable afterwards, so the wrong answer there is silent rather than an error.
- A throwing callback is caught and logged; it does not turn a committed
  transaction into a rejected one.
- `dispatchAndWait` on a deferred event resolves immediately having run nothing.
  That is the sharp edge the plan flags, and it warns once per event name in
  development.

`Event.dispatch` reaches this through `deferUntilCommit(callback): boolean`,
whose return value is the contract: `false` means nothing will ever drain, so
run it now. Consulting `currentTransaction()` instead would queue onto a
hand-built scope's non-existent list and lose the dispatch.

## Part two — `Event.fake()`

`Event.fake()` swaps the container's `EventManager` for a `FakeEventManager`
that records and runs nothing — no listener constructed, no `handle`, nothing
pushed to the `QueueManager`, nothing queued on an open transaction. It returns
the same recorder on a second call, and `restore()` puts back the exact instance
the application had (not a rebuilt one with an empty registry).

`assertDispatched`, `assertNotDispatched`, `assertDispatchedTimes` and
`assertNothingDispatched`, all matching by declared `static name` rather than
class identity, all taking an optional predicate typed off the class passed in.
Failures name what *was* dispatched:

```
Expected OrderPaid to have been dispatched. Dispatched: UserRegistered(2, "b@c.d").
```

## Deviations from the plan

1. **`gemi/testing` publishes the recorder's type, not the class.** The plan
   says to export it from `gemi/testing` rather than `gemi/services`. That
   entrypoint is bundled *for the browser* — `vite.client.config.mts` builds
   `client/index.ts` and `testing/index.ts` in one pass so `<Page>` and the
   components under it share one copy of every React context — so a runtime
   export would drag `EventManager`, `node:async_hooks`, the container and Bun's
   `SQL` into a component-test bundle. The class therefore lives beside the real
   manager in `services/events/`, `gemi/testing` re-exports it with
   `export type`, and the only runtime door is `Event.fake()` on the already-
   exported `Event` — which is the API the plan's own sketch uses. No new name
   is added to `gemi/services`.
2. **Savepoint truncation is by list index, not by a depth recorded per
   callback.** The plan says "record the depth alongside each callback and
   truncate on savepoint failure"; depth cannot separate two siblings (see
   above), and the index is exact because the ORM issues one statement at a time
   on a transaction handle. Both behaviours are pinned by tests.
3. **`AfterCommitCallback` is not exported.** `orm/architecture.test.ts` fails an
   export nothing outside its own file uses, and callers pass a lambda.

Nothing in the out-of-scope list was built: no `fakeExcept`, no snapshotting, no
assertion about listeners having run.

## Verification

From `packages/gemi`, on this branch:

```
$ bun --bun vitest run
 Test Files  151 passed (151)
      Tests  3763 passed | 1 skipped (3764)

$ bun --bun vitest run --typecheck.only
 Test Files  5 passed (5)
      Tests  36 passed (36)
 Type Errors  no errors

$ bunx tsc --noEmit -p tsconfig.json          # clean
$ bunx tsc --noEmit -p tsconfig.browser.json  # clean, incl. the new type re-export
$ bunx oxlint
Found 80 warnings and 0 errors.               # 80 on the base branch too
```

New tests: 21 in `orm/context.test.ts` (commit/rollback, savepoints in both
directions, the committed sibling, the drain's scope, a throwing callback, two
concurrent transactions, and the two `false` cases), 11 in
`services/events/afterCommit.test.ts` (the three-case table, the queue, the
`dispatchAndWait` warning), 17 in `services/events/fake.test.ts`, and 4 type
tests for the predicate's inference.

Formatting: `orm/context.ts` already fails `prettier --check` at HEAD on one
pre-existing line, which is left alone; every added line is prettier-clean, and
the four new files pass `prettier --check` whole.

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -163,7 +163,7 @@ That is the whole of the opt-in. Behind it, the listener is registered with the 
 
 It is a **context boundary spelled as one line**, not a latency setting.
 
-- A **sync** listener runs inside the dispatching request: it has that request's `app()`, its authenticated user through `currentActor()`, and it joins the ambient transaction.
+- A **sync** listener runs inside the dispatching request: it has that request's `app()`, its authenticated user through `currentActor()`, and it joins the ambient transaction — unless the event declares [`static afterCommit`](#deferring-until-the-transaction-commits), which releases the listeners after the commit and so outside it.
 - A **queued** listener is handed to the queue and run from a drain, on the queue's terms. With `worker = true` that is a different thread with a cloned application. Only the event's name and constructor arguments cross; nothing else does.
 
 The part that catches people is that the request's context is not reliably *gone* either. gemi's queue is in-process, so a drain that happens to start from the push is still standing inside the dispatching request, and `app()` there resolves the request's application. **None of that is promised.** A queued listener that reads the current actor, or writes expecting to join the transaction the controller opened, works right up until the day the queue was already busy — and then reads different rows, or commits separately, with no error either way.
--- a/docs/events.md
+++ b/docs/events.md
@@ -36,6 +36,11 @@ An event is the payload plus a name. It carries no framework state — no timest
 
 > **Note:** The static `name` is **required**, and omitting it does not fall back harmlessly to the class name. The listener registry is keyed by this string. `gemi build` minifies the server entry, and the app code reachable from it — your controller, and every event class it imports in order to dispatch — is bundled and minified with it, which renames the class binding; a class's implicit `.name` *is* that binding, so `UserRegistered` becomes something like `D`. Discovery, meanwhile, imports `app/listeners/*.ts` from source at runtime, and those files import `app/events/*.ts` from source too. Two module graphs, two class objects, two different names — and the dispatch reaches nobody, in production and nowhere else. A declared `static name` is a string literal, which survives minification intact. Discovery warns at boot about any event a listener binds to that leaves it out.
 
+| Field | Type | Default | Description |
+| --- | --- | --- | --- |
+| `static name` | `string` | `"unset"` | The name the registry is keyed by. Required. |
+| `static afterCommit` | `boolean` | `false` | Hold the listeners until the surrounding transaction commits, and drop them if it rolls back. See [Deferring until the transaction commits](#deferring-until-the-transaction-commits). |
+
 ### Never `instanceof` an event
 
 The same two class objects are why this is unsafe **in your own code**:
@@ -197,6 +202,67 @@ app(QueueManager).registeredJobs.map((job) => job.name);
 
 `registeredJobs` reports what the queue will run, and a queued listener genuinely is something it will run — hiding them would make a queue introspection tool lie. The `listener:` prefix is what keeps you from going to look for a job file you never wrote.
 
+## Deferring until the transaction commits
+
+A dispatch inside a transaction happens immediately. The transaction may not:
+
+```typescript
+await DB.transaction(async () => {
+  const user = await User.create(input);
+  UserRegistered.dispatch(user.id, user.email);   // welcome email sent
+  await Billing.provision(user);                  // throws
+});                                               // rolled back
+```
+
+The user does not exist and has been welcomed. Nothing errored — the email succeeded and the transaction failed, in two subsystems that never meet — so nothing anywhere reports it.
+
+`static afterCommit = true` on the **event** moves its listeners to the commit:
+
+```typescript
+// app/events/UserRegistered.ts
+export class UserRegistered extends Event {
+  static name = "UserRegistered";
+  static afterCommit = true;
+
+  constructor(
+    public userId: number,
+    public email: string,
+  ) {
+    super();
+  }
+}
+```
+
+| `afterCommit` | Inside a transaction | What happens |
+| --- | --- | --- |
+| `false` (default) | either | The listeners run now. |
+| `true` | no | The listeners run now — there is nothing to wait for. |
+| `true` | yes | The listeners run when that transaction commits, and not at all if it rolls back. |
+
+It is a property of the **event**, not of the listener: whether a side effect is allowed to happen before the write is durable is a fact about what happened, and every listener bound to it agrees. Both kinds are deferred together — a sync listener runs after the commit, and a queued one is *pushed* after the commit. Pushing early would be running early, because the queue drains in-process and often starts from the push itself.
+
+The listeners run **outside** the transaction that deferred them. A listener that opens its own transaction gets a real one rather than a savepoint on a handle that has already closed — which matters more than it looks: Bun's transaction handle stays callable after the commit and simply runs on the pool, so the wrong answer here would be silent.
+
+### The two things to know before using it
+
+**`dispatchAndWait` on a deferred event resolves immediately, having run nothing.** The listeners are queued on a transaction the caller has not finished, so the `await` buys nothing and code on the next line reading what a listener wrote finds it missing. Development warns once per event name; there is no other symptom. Await the transaction instead.
+
+**The transaction does not resolve until the deferred listeners have.** `await DB.transaction(...)` covers the commit *and* the listeners it released. That is deliberate — the alternative is work floating after the transaction returns, which a short-lived process (a command, a seed) exits out from under, and a failure nobody is left to report.
+
+A listener that throws after the commit is logged and the next one still runs, exactly as on the inline path. It does **not** fail the transaction: the database kept the rows, and reporting a rejection would have the caller roll back or retry work that is already durable.
+
+### Savepoints
+
+Nesting `DB.transaction` takes a savepoint, and deferred dispatches follow the transaction rather than the block:
+
+- A dispatch inside a savepoint that commits runs when the **outer** transaction commits — not when the savepoint closes.
+- A dispatch inside a savepoint that **rolls back** does not run, even if the outer transaction commits.
+- An outer rollback discards everything, savepoint dispatches included.
+
+### Why it is opt-in
+
+Defaulting it on would mean a `dispatch` inside a transaction silently has not happened yet — a fire-and-forget call turned into a deferred one by ambient state the call site cannot see. That is a surprise of the same size as the one being fixed, and the harder one to debug, because nothing went wrong. Turn it on for the events where a rollback must take the side effect with it: a welcome email, a webhook, an invoice.
+
 ## Registering listeners — `app/listeners/`
 
 Every class under `app/listeners` that extends `Listener` is registered when the kernel boots.
@@ -290,6 +356,74 @@ import { discoverListeners } from "gemi/services";
 const listeners = await discoverListeners(); // every Listener under app/listeners
 ```
 
+## Testing with `Event.fake()`
+
+`Event.fake()` swaps the container's `EventManager` for one that records dispatches and runs nothing:
+
+```typescript
+import { Event } from "gemi/services";
+import { UserRegistered } from "@/app/events/UserRegistered";
+
+test("registering announces the new user", async () => {
+  const events = Event.fake();
+
+  await post("/register", { email: "ada@example.com" });
+
+  events.assertDispatched(UserRegistered);
+  events.assertDispatched(UserRegistered, (e) => e.email === "ada@example.com");
+});
+```
+
+The second assertion is the point of the whole subsystem, in a test. A controller test that asserts *the event fired* is testing the controller. One that asserts *a welcome email was sent* is testing the controller, the listener and the mailer at once, and it breaks the day a fourth listener is added to an event it was never about.
+
+While a fake is installed, no listener is constructed, no `handle` runs, nothing is pushed to the queue, and an [`afterCommit`](#deferring-until-the-transaction-commits) event is recorded at the dispatch rather than held for a commit — so a controller test does not have to commit a transaction to see what the controller dispatched.
+
+### Restore it
+
+```typescript
+let events: FakeEventManager;
+
+beforeEach(() => { events = Event.fake(); });
+afterEach(() => { events.restore(); });
+```
+
+**There is no automatic teardown.** A fake that outlives its test is a later test whose listeners silently never run — and nothing catches that, because a fake legitimately has no listeners and so never trips the "nothing is listening" warning.
+
+Calling `Event.fake()` twice returns the **same** recorder, so a fake set up in a helper and one in the body of a test do not disagree; `restore()` twice is harmless. The type is importable for a signature:
+
+```typescript
+import type { FakeEventManager } from "gemi/testing";
+```
+
+### The assertions
+
+| Assertion | Passes when |
+| --- | --- |
+| `assertDispatched(Event, predicate?)` | The event was dispatched at least once — matching `predicate`, if given. |
+| `assertNotDispatched(Event, predicate?)` | It was never dispatched; with a predicate, no dispatch of it matched. |
+| `assertDispatchedTimes(Event, times, predicate?)` | It was dispatched exactly `times` times. |
+| `assertNothingDispatched()` | Nothing at all was dispatched. |
+
+The predicate takes the event instance and is typed off the class you passed, so `(e) => e.email` is checked without a cast. Matching is by the event's declared `static name`, never by class identity — the same rule the registry follows, and for the same reason.
+
+A failure names **what was dispatched**:
+
+```
+Expected OrderPaid to have been dispatched. Dispatched: UserRegistered(2, "b@c.d").
+```
+
+The common failure is not "nothing fired" but "that fired with a different payload", and printing the recorded dispatches usually ends the investigation on the spot.
+
+### What there is no assertion for
+
+**That a listener ran.** A test asserting that is a test *of the listener*, and it can be written with no framework around it at all:
+
+```typescript
+await new SendWelcomeEmail().handle(new UserRegistered(7, "ada@example.com"));
+```
+
+That is the separation the fake exists to make possible, so re-coupling the two through an assertion would give it back.
+
 ## When to reach for a job instead
 
 A **sync** listener runs inside the dispatching request, in the same process and the same context. That makes it right for fast in-request side effects — writing an audit row, warming a cache, updating a counter — and wrong for slow ones, because the request waits for `dispatchAndWait` and the process does the work either way.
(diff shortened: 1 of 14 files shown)


nstfkc:
## Review — plan-conformance

Walked every numbered deliverable in `plans/events/03-after-commit-and-testing.md`, both **Tests** lists, and the **Out of scope** list against the diff.

**Part one — all four deliverables present.** `OrmScope.afterCommit` created by the outermost `withTransaction` and shared by spread into savepoints (`orm/context.ts:94`, `orm/context.ts:600`); drain only on the fulfilled path of the outermost branch, after `.finally(stopWatching)`, inside a rebuilt `ormContext.run` with `tx`/`afterCommit` cleared, per-callback catch-and-log (`orm/context.ts:441`, `orm/context.ts:646`); `static afterCommit = false` on `Event` with the sharp-edge doc comment (`Event.ts:110`); all three rows of the plan's table implemented in `EventManager.deferToCommit` (`EventManager.ts:265`), and the `dispatchAndWait` + `afterCommit` pairing warns once per event name in development (`EventManager.ts:479`).

**Part one tests — all seven present**, in `orm/context.test.ts` and `services/events/afterCommit.test.ts`: commit runs / rollback runs nothing; savepoint defers to the outer commit; savepoint rollback under a committing transaction drops only its own entries (plus a sibling case); `currentTransaction()` undefined inside a callback; throwing callback neither rejects nor stops the next; two concurrent transactions keep their own lists; `afterCommit` with no transaction dispatches immediately.

**Part two — all six deliverables present**, plus "the interaction worth testing" (fake + queued listener asserts `queue.queue.size === 0` and that `listener:SendWelcomeEmail` is still registered, so the test proves the push did not happen rather than that there was nothing to push to).

**Out of scope respected.** No `fakeExcept`, no snapshotting, no listener-ran assertion — and `docs/events.md` declines the last one explicitly with the direct-`handle()` alternative. Nothing outside `docs/`, `orm/context*`, `services/events/*` and `testing/index.ts` was touched.

### Findings

**minor — `packages/gemi/services/events/Event.ts:3`** — the deviation on the `gemi/testing` export moves the cost the plan's clause was about, rather than removing it.

> "Exported from `gemi/testing`, not `gemi/services` — it is test-only, and `barrel-imports.test.ts` is about not making every app pay for things it does not use."

`testing/index.ts:21` publishes only the type, which is defensible for the reason given (the client vite build). But the runtime door is `Event.fake()`, and `Event.ts` reaches it through a **value** import of `./FakeEventManager` at module scope. `Event` is on the `gemi/services` barrel (`services/index.ts:78`), so `import { Event } from "gemi/services"` — the line every app and every template writes — now evaluates `FakeEventManager.ts` and its assertion/formatting helpers, and they land in every `gemi build` server bundle. That is the exact direction the quoted clause rules out: the test-only class is now paid for by `gemi/services` and not by `gemi/testing`.

Small in absolute terms (one module, no new heavy dependency — I confirmed the class pulls in nothing beyond `EventManager`, which the barrel already exports, so `barrel-imports.test.ts` itself still passes), which is why this is minor and not major. A lazy `await import("./FakeEventManager")` inside `fake()` would close it, at the cost of making `Event.fake()` async.

### Checked, and not findings

- **Depth tag → mark-and-truncate** (`orm/context.ts:557`). The plan says "Record the depth alongside each callback and truncate on savepoint failure", and the diff does neither. Accepted: the sibling case the author names is real — a committed savepoint and a rolled-back one leave entries at the same depth, adjacent in one shared list — so the literal instruction discards work it must keep. The replacement is pinned by "a sibling savepoint that committed keeps its callbacks", which a depth implementation fails, so the behaviour the deliverable actually specifies is stronger here, not weaker.
- **`deferUntilCommit()` instead of `currentTransaction()`** in deliverable 3. Strictly more conservative: a hand-built scope carrying a handle answers `false` and dispatches now, rather than queueing onto a list nothing drains. Pinned by "a hand-built scope carrying a handle defers nothing either".
- **The drain is awaited.** The plan is silent; the choice is documented on `drainAfterCommit`, in `docs/events.md`, and pinned by "the transaction does not resolve until the callbacks have".
- **Docs.** Iteration 3's deliverable list does not name them, but the plan asks for `restore()` shown in `afterEach` — `docs/events.md` does exactly that, so this is in scope rather than added scope.
- **The browser type build.** The new `export type` in `testing/index.ts` pulls `services/events/*` into `tsconfig.browser.json`'s program for the first time. I ran both emits into scratch dirs: `services/events/*.d.ts` is byte-identical between `build:types` and `build:client-types`, so the last-writer-wins overwrite in `dist/` changes nothing.

**Verdict:** the plan is implemented — every deliverable and every listed test is present, the three deviations are declared and each is defensible, and the one finding is a minor bundle-cost inversion in `Event.ts:3`, not a blocker.

nstfkc:
## Review — invariants

Traced all five plus the config rule against the code rather than the comments. One minor finding.

**1. Registry keyed by declared `static name`** — holds. `useListeners` still keys off `listener.event.name` (`EventManager.ts:177`), and both new read sites read the string structurally off the constructor: `deferToCommit` (`EventManager.ts:285`) and `FakeEventManager.matching` → `nameOf(record.event.constructor)` (`FakeEventManager.ts:240,272`). No map keyed by a class object, no `instanceof`, no `===` between class references on the dispatch or the assertion path — `assertDispatched(UserRegistered)` compares `UserRegistered.name` against the recorded constructor's `.name`, which is the right side of the two-module-graph boundary. `install`'s `instanceof FakeEventManager` (`FakeEventManager.ts:79`) is an instance-vs-class test whose both sides can only come from `install` itself, and the test runner drives source directly (`bun --bun vitest`), so there is no second copy to make it false. Not a violation.

**2. Zero-listener warning** — holds. `warnNothingIsListening` (`EventManager.ts:497`) is still `NODE_ENV === "production"`-gated and once-per-name, and it is reached on the deferred path too, since the drained callback is `runListeners` itself. The new `warnedAboutWaiting` set (`EventManager.ts:98`) is genuinely separate, so a legitimately-unlistened event cannot silence the `dispatchAndWait`+`afterCommit` warning or vice versa.

**3. One listener cannot stop another** — holds, and I traced the new path rather than trusting the comment. `runListeners` (`EventManager.ts:371-386`) still awaits each listener inside its own `try`, with `new Listener()` inside the `try` as well; `enqueue` catches its own two failure modes. The deferred path adds a second layer that also cannot break it: each deferred event is its own callback, and `drainAfterCommit` (`orm/context.ts:449-460`) wraps each callback in a `try` and continues. No `Promise.all`/`allSettled` over listeners or over callbacks anywhere.

**4. Sync/queued context documented at the declaration** — see the finding below; `queued`'s doc is still correct about `app()` and the actor, but the transaction half of the sentence is now conditionally false.

**5. Nothing in the framework dispatches an event** — holds. Outside `services/events/`, the only files mentioning `Event`/`EventManager` are the two barrels and `discovery.test.ts`. The ORM side takes an opaque `() => void | Promise<void>` and does not import the events subsystem, so the new hook adds no framework-side dispatch.

**"Declared wins, and `[]` is declared"** — verified against the raw slice, not the defaulted config. `Repository.get` returns the stored object by reference with no defaults applied, `withDefaults` builds a fresh object rather than mutating its input, and `useListeners`'s `this.config.listeners = listeners` (`EventManager.ts:131`) writes to that copy. So even though `boot()` calls `make(EventManager)` before reading the slice, `slice.listeners` at `EventServiceProvider.ts:60` is still exactly what the app wrote: absent and `undefined` both discover, `[]` is honoured verbatim.

### Findings

**minor — `packages/gemi/services/events/Listener.ts:104` (and the same sentence at `:153`)**
`queued`'s doc says a sync listener "joins the ambient transaction", and `handle`'s says it has "its ambient transaction". With `afterCommit` that is now false for a whole class of listeners: the drain re-enters a scope with `tx: undefined` (`orm/context.ts:441-447`), so a sync listener bound to an `afterCommit` event runs with `currentTransaction()` undefined. The concrete failure is a change made in a different file from the one that is wrong: a listener is kept sync *because* the doc promises it joins the transaction — an audit or ledger row that is meant to roll back with the write it describes — and then someone adds `static afterCommit = true` to the event, which is exactly what `docs/events.md` recommends for that shape. The listener now commits separately, and there is no error and no warning on that path. `Event.afterCommit` documents it, but it is documented on the event, and the author who has to know is reading the listener. One clause on `queued` ("unless the event declares `static afterCommit`, in which case a sync listener runs after the commit, outside the transaction") closes it. `docs/events.md:166` has the same sentence, though it at least sits a section away from the afterCommit text that corrects it.

**Verdict:** all five invariants hold and the declared-wins rule really distinguishes absent from `undefined` from `[]`; nothing blocking, one doc clause on `Listener.queued` that `afterCommit` has made conditionally untrue.

nstfkc:
## Review — integration

Checked the diff against the rest of the repo rather than against itself. Everything the lens names came back clean; details below so the checks are not repeated.

### Finding

**nit — `packages/gemi/testing/index.ts:21`: `DispatchedEvent` is not published beside `FakeEventManager`, so the type of its one public field is unnameable.**

`FakeEventManager.dispatched` is declared `readonly dispatched: DispatchedEvent[]`, and `DispatchedEvent` is exported from `services/events/FakeEventManager.ts` but not re-exported from the testing entrypoint. Concretely: `import type { DispatchedEvent } from "gemi/testing"` does not resolve, so a shared test helper written as

```typescript
function paidOrders(records: DispatchedEvent[]) { … }
```

cannot be typed without `FakeEventManager["dispatched"][number]`. One more `export type` line on the same statement, if you want the escape hatch to be nameable; entirely optional, since the docs present `dispatched` as a last resort and the workaround exists.

### Verified clean

- **`docs-imports.test.ts`.** `docs/events.md` adds `import { Event } from "gemi/services"` (already exported, `services/index.ts:78`) and `import type { FakeEventManager } from "gemi/testing"`. `exportsOf()`s `export\s+(?:type\s+)?\{…\}` regex picks the new `export type { FakeEventManager } from …` line up, so the name resolves; the entrypoint set the sweep pins is unchanged (`gemi/testing` was already in it) and the import count only goes up. The `llms-full.txt` copy round-trips exactly against `events.md` with the H1 demoted — checked programmatically, byte-for-byte.
- **`barrel-imports.test.ts`.** No new heavy package. `EventManager.ts` -> `orm/context.ts` -> `database/Connection.ts` -> `import { SQL } from "bun"` is a new *edge*, but not a new *reachability*: walking `services/index.ts` with that edge removed still lands on `database/Connection.ts` via `services/router/ViewRouteDispatcher.ts -> auth/routes.ts -> auth/AuthController.ts -> facades/index.ts -> facades/DB.ts -> orm/context.ts`. `gemi/services` was already Bun-eager, so the barrel guard is not weakened. Graph size 158 files, well over the `> 50` floor.
- **The `gemi/testing` type-only deviation.** Ran `vite build -c vite.client.config.mts` to a scratch outDir: `testing/index.js`, `client/index.js` and the shared chunk contain zero occurrences of `EventManager` or `async_hooks`. The `export type` is erased as intended.
- **Declaration emit.** `testing/index.ts` now drags `services/events/{Event,EventManager,FakeEventManager,Listener,listenerJob}.ts` into the `tsconfig.browser.json` program, which is the last emit in the `build` chain and so overwrites what `build:types` wrote. Emitted both configs to scratch dirs and diffed all five: byte-identical, so `build:client-types` does not downgrade the shipped types with its `strict: false` settings. `tsc -p tsconfig.browser.json` and `tsc --noEmit -p tsconfig.json` both exit 0.
- **`kernel/providers.ts`.** No provider or service added, so the "Sixteen providers for seventeen services" count is still right — correct not to touch it.
- **Service token.** `FakeEventManager` inherits `token = "events"` deliberately and is installed with `application.instance()`, which overwrites rather than throws. The `Kernel.ts:153` duplicate-token guard only inspects app-declared `kernel.services` at boot, so a fake installed after boot never reaches it. `restore()`s `instance`/`forget` split matches `Container.forget` leaving the lazy singleton factory in place — the "leaves an unresolved binding unresolved" test is a real assertion against that, not a tautology.
- **Tests assert their names.** Spot-checked the ones most at risk: "its listener runs outside the transaction that deferred it" fails against a no-op `afterCommit` (the handle would be defined); "a sibling savepoint that committed keeps its callbacks" is exactly the case a depth-tagged implementation loses; "is pushed at the commit, not at the dispatch" pins `ran` empty inside the block, so an immediate push cannot slip through; the fake/`afterCommit` test would throw at `assertDispatched` if the fake did not short-circuit ahead of `deferToCommit`.
- **Churn.** Thirteen files, all in scope. No `templates/`, no `bun.lock`, no reformatting — `oxfmt --check` flags the touched files, but it flags untouched ones (`services/queue/QueueManager.ts`, `orm/model.ts`) the same way, so that is the existing baseline and not this PR. `oxlint --deny-warnings` over `services/events`, `orm/context.ts` and `testing/index.ts`: 0/0.
- **Comment style.** The all-caps section header, the failure-mode-first docblocks and the "why not the obvious alternative" paragraphs match `withTransaction`s existing `THE SINGLE-CONNECTION ASSUMPTION…` block and `QueueManager.ts`. I did not find a comment that only narrates.

**Verdict:** no integration blockers — one optional nit about publishing `DispatchedEvent`.

nstfkc:
## Review addressed

Two fixed, one declined with a measurement. `470c527`.

### Fixed — invariant 4: `Listener.queued` promised a transaction `afterCommit` takes away

`packages/gemi/services/events/Listener.ts:104` (and `handle` at `:159`), plus the matching bullet in `docs/events.md` and its `llms-full.txt` copy.

The finding is right and it is the one that could bite. `queued`'s doc said a sync listener "joins the ambient transaction" flatly; with `static afterCommit` on the event the drain re-enters a scope with `tx` cleared, so the listener runs with `currentTransaction()` undefined and commits on its own. The correcting text was on `Event.afterCommit` — a different file from the one whose author has to decide, and the flag is what `docs/events.md` recommends for exactly the listener shape (an audit or ledger row) that must not commit separately. Nothing on the listener side reports it: no error, no warning, and a `queued = false` that still reads as "inside the transaction". That is now said where `queued` is declared.

### Fixed — `DispatchedEvent` published beside `FakeEventManager`

`packages/gemi/testing/index.ts`. Type-only, on the same statement, so the browser bundle is unchanged (`tsc -p tsconfig.browser.json` still exits 0). A field that is public but whose type has no public name is not really an escape hatch.

### Declined — the `FakeEventManager` value import in `Event.ts`

I measured it rather than argue it. `bun build ./services/index.ts --minify --splitting --target=bun` with the same externals as `scripts/build.ts`, run with and without the import:

| | `dist/services/index.js` |
| --- | --- |
| with `FakeEventManager` | 308,860 bytes |
| without | 306,864 bytes |

**1,996 bytes**, against 1.8 MB for the build as a whole, and no new dependency: the class imports `EventManager`, which the barrel already exports, and the assertion helpers are local functions. Evaluating it does nothing — a class declaration and some function declarations, no I/O, no module-scope work — so the startup cost is not measurable either.

That is not the cost the quoted clause is about. `barrel-imports.test.ts` says so in its own doc comment: the failure it exists for is `import { CronJob } from "gemi/services"` **evaluating** `@aws-sdk/client-s3`, `resend`, `satori` and `sharp` (#403), which a test runner can never shake out and which costs real seconds. `HEAVY` is a list of third-party packages, and 2 KB of first-party code with no dependencies is not on the same axis.

The fix on offer costs more than it buys. An `await import()` makes `Event.fake()` async, and the plan's own sketch is `const events = Event.fake();` — unawaited, on the first line of a test. A missing `await` there gives you a `Promise` whose `assertDispatched` is `undefined`, which is a worse failure than 2 KB. The other shape, a real runtime export on a new `gemi/testing/events` subpath, is the one I would take — but it needs `exports`, `scripts/build.ts` and `build-publish.ts`'s dangling-export check to move together, and iteration 3 touches none of those.

So I read this as the plan clause being slightly over-stated rather than the code being wrong: **the class is not exported from `gemi/services`** — it is not on the barrel, and `Event.fake()` is the only door, exactly as specified. What the clause additionally implies (that no byte of it reaches the services graph) is not achievable alongside a synchronous `Event.fake()`, and is worth correcting in `plans/events/03-after-commit-and-testing.md` when the server-side testing subpath is worth doing on its own.

### Verification

`bunx tsc --noEmit -p tsconfig.json`, `tsc --noEmit -p tsconfig.browser.json`, `bunx oxlint` (80 warnings, 0 errors — the HEAD baseline), and `bun --bun vitest run docs-imports.test.ts services/events orm/context.test.ts barrel-imports.test.ts orm/docs-*.test.ts`: 257 tests, all passing.

PR #422, opened 2026-08-22
Resolve the initial route's views before hydrating

nstfkc:
## The symptom

A cold load paints its server HTML, blanks the content between the layout for ~300 ms, then paints it again. The layout is unaffected — only the page content flashes.

Measured on a production build of `templates/saas-starter` (CPU throttled 6×, `MutationObserver` on `documentElement`), three consecutive runs, before:

```
175ms REMOVED  "brain.coAboutPricingSign InTR-TREN-US"
175ms REMOVED  "Welcome to Gemi 0.52A simple and fast framework for building"
175ms REMOVED  "brain.coAboutPricingSign In2025 brain.co"
188ms added    "brain.coAboutPricingSign InTR-TREN-US"
188ms added    "brain.coAboutPricingSign In2025 brain.co"
488ms added    "Welcome to Gemi 0.52A simple and fast framework for building"
```

After: `no large removals`, on every run.

It reproduces on a real deployment too — a gemi 0.56 app's landing page goes content-present at 573 ms → empty at 671 ms → back at 973 ms.

## The cause

1. The server renders each route segment through a real component, so the shell ships **complete** `<Suspense>` boundaries with content inside.
2. The browser renders the same segments through `lazy()`, so on the first client render every one of them suspends. The `modulepreload` hint gets the chunk into cache, but `import()` still resolves asynchronously.
3. React tolerates that by itself — a boundary that *merely* suspends during hydration keeps its server HTML on screen. What it does not tolerate is an **update arriving while the boundary is in that state**: the boundary is client-rendered from scratch, and a view with no `Loading` export has nothing but a `null` fallback to show.
4. `loadViewModule` guarantees such an update. It announces the module to `Route`'s `useSyncExternalStore` the moment the chunk lands, which is one dictionary `await` *before* the `lazy()` promise it feeds settles.

The layout's chunk resolves first, so the layout snaps back within ~13 ms and only the content is left empty for the remaining ~300 ms. That asymmetry is the whole visible symptom.

## The fix

`initialViewModulesReady` resolves the views the document was rendered with — `routeManifest[pathname]`, the same lookup `ClientRouterProvider` seeds its route state from — and `init` waits on it before `hydrateRoot`. `lazy()` then returns those components synchronously, the first render never suspends, and no boundary is left half-hydrated for a later update to blank.

It costs nothing on the wire: the shell already emits a `modulepreload` for each of those chunks. It never rejects, so a chunk that fails to load still hydrates and surfaces through the route's error boundary.

## Tests

`hydrationBlank.test.tsx` pins the four React 19.2.3 behaviors this rests on. One of them forecloses the obvious alternative: **wrapping the registry notification in `startTransition` does not help**. React's advice to do that applies to a boundary the server left *pending*; this one was served complete, so the transition commits the fallback just the same. That case is a test rather than a code comment because it looked like it worked when measured against a warm dev server.

Full suite green on this branch: 162 files, 3907 tests. `typecheck` and `oxlint` clean.

https://claude.ai/code/session_01A1vkcLhp3M1b8AxxSCPBUp

### DIFF
--- a/packages/gemi/client/ComponentContext.tsx
+++ b/packages/gemi/client/ComponentContext.tsx
@@ -103,6 +103,17 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "test") {
 }
 
 /**
+ * How long hydration will wait for those modules before giving up on them.
+ *
+ * Only a load that never *settles* reaches this — a stalled HTTP/2 stream, a
+ * service worker that never answers. A rejection is already handled. Without
+ * a deadline that case costs the whole document: nothing hydrates, and
+ * nothing is raised. Expiring it just puts the page back on the pre-#352 path
+ * for one route, which is a bad outcome and not a dead one.
+ */
+const INITIAL_VIEW_MODULES_DEADLINE_MS = 2000;
+
+/**
  * Resolves once the views the document was server-rendered with are in the
  * registry — which is what `init` waits for before it hydrates.
  *
@@ -117,21 +128,48 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "test") {
  * `Loading` export is `null`. The page then sits with its layout and no
  * content until the chunk's `import()` and dictionaries resolve.
  *
- * Awaiting the imports first costs nothing on the wire (the shell already
- * emits a `modulepreload` for each of them) and makes `lazy()` return the
- * component synchronously, so the initial render never suspends and there is
- * no half-hydrated boundary for a later update to blank.
+ * What this buys is NOT a synchronous `lazy()`. `lazy` suspends on its first
+ * render no matter what: the initializer calls the ctor, attaches `.then`,
+ * and re-reads a status its callback only sets a microtask later, so even an
+ * already-settled promise goes to `Pending` and throws. The render still
+ * suspends here — it just suspends *untouched*. `loadViewModule` notifies its
+ * listeners only on a view's FIRST registration, so once the module is in
+ * `viewModules` the call `lazy` makes during hydration is a silent one, the
+ * boundary sees no update, and React keeps the server HTML until it settles.
  *
- * Never rejects: a chunk that fails to load must still leave the app to
- * hydrate and surface the error through the route's error boundary.
+ * The invariant to preserve, then, is "no first registration for a mounted
+ * view happens during hydration" — not anything about how `lazy` resolves.
+ * Notifying on every call, or memoizing the promise so the `isNew` guard
+ * stops being the discriminator, would reintroduce the flash while leaving
+ * this preload apparently intact.
+ *
+ * It is not free. The view chunks themselves are already `modulepreload`ed by
+ * the shell, but `loadViewModule` also awaits `preloadDictionaries`, and
+ * dictionary locale chunks are dynamic imports, which `collectModulePreloads`
+ * deliberately leaves out of those hints. So hydration is gated on a
+ * serially-discovered round trip — view chunk, parse, dictionary chunk — and
+ * the layout's own handlers stay dead across it. That wait is deliberate
+ * rather than incidental: a view whose dictionary is still in flight suspends
+ * on `use()` inside this same boundary, which is the same blank arriving by a
+ * different route. Warming the current locale's dictionaries from the shell
+ * would buy the interactivity back; awaiting less here would not.
+ *
+ * Never rejects and never waits forever: a chunk that fails or hangs must
+ * still leave the app to hydrate and surface the error through the route's
+ * error boundary.
  */
 export const initialViewModulesReady: Promise<unknown> =
   typeof window !== "undefined" && process.env.NODE_ENV !== "test"
-    ? Promise.all(
-        initialViewNames(window.__GEMI_DATA__).map((name) =>
-          loadViewModule(name).catch(() => null),
+    ? Promise.race([
+        Promise.all(
+          initialViewNames(window.__GEMI_DATA__).map((name) =>
+            loadViewModule(name).catch(() => null),
+          ),
         ),
-      )
+        new Promise((resolve) =>
+          setTimeout(resolve, INITIAL_VIEW_MODULES_DEADLINE_MS),
+        ),
+      ])
     : Promise.resolve();
 
 /**
--- a/packages/gemi/client/ComponentContext.tsx
+++ b/packages/gemi/client/ComponentContext.tsx
@@ -47,6 +47,13 @@ export function loadViewModule(name: string): Promise<any> {
     // Before the dictionary await, not after: this exists to surface the view's
     // `Loading` export the moment it lands, and holding it behind a network
     // fetch would put back the very `null` flash it removes.
+    //
+    // On a cold load this fires at the worst possible moment — the boundary it
+    // wakes is still suspended on the very `lazy()` this module is resolving,
+    // and an update there costs the boundary its server HTML. Wrapping it in a
+    // transition does not help (`hydrationBlank.test.tsx` pins that); what
+    // makes it harmless is `initialViewModulesReady`, which puts the initial
+    // route's modules in the registry before anything subscribes to it.
     if (isNew) {
       for (const listener of viewModuleListeners) listener();
     }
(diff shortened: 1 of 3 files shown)


nstfkc (commented):
The mechanism checks out, and I verified the parts that could have silently made it a no-op:

- `window.loaders` is assigned by the inline bootstrap script *before* the client entry imports (`ViewRouteDispatcher.ts:452`), so the module-scope read in `initialViewModulesReady` is safe on both the `clientEntry` and `bootstrapModules` paths.
- `initialViewNames` matches `ClientRouterProvider`'s seeding exactly. `router.pathname` is the route *pattern* (`currentPathName`), not the URL, so `routeManifest[pathname]` is the same lookup the server used for `currentViews` — param routes resolve correctly.
- The missing `is500` branch is dead code rather than a gap: `ClientRouter.tsx:437` hardcodes `is500={false}`.
- `"404"` is a real key in `window.loaders` on both dev and prod.

Four comments below. The two error-handling ones are the ones I'd fix before merging — both take a failure that used to cost one route segment and turn it into a permanently unhydrated document with nothing reported.

packages/gemi/client/init.tsx
```diff
@@ -39,38 +40,45 @@ export function init(
...
-              const overlay = new ErrorOverlay({
-                message: (error as any).message,
-                stack: (error as any).stack || "",
-              });
-              document.body.appendChild(overlay);
-            }
+    // Held until the current route's view chunks are in the registry, so the
+    // first client render resolves them synchronously instead of suspending
+    // every route segment mid-hydration. See `initialViewModulesReady`.
+    void initialViewModulesReady.then(() => hydrate(RootLayout, options));
```
nstfkc: **This re-opens the hole the entry-import `.catch` exists to close.**

`ViewRouteDispatcher.ts:507-514` wraps the entry `import()` in `.catch(function(e){setTimeout(function(){throw e})})`, with the comment explaining why: an `unhandledrejection` "is not seen by `window.onerror`-based reporters. The page would then sit permanently unhydrated with nothing raised."

That net only covered hydration because `init()` used to hydrate *synchronously*, inside the entry's evaluation — a throw out of `hydrateRoot` propagated into that `.catch`. Now `init()` returns immediately, the entry import resolves clean, and a synchronous throw from `hydrate` (a `RootLayout` module-level failure, a root already attached) lands in a `.then` callback with no rejection handler. That is exactly the silent, unhydrated page the comment forbids.

```suggestion
    void initialViewModulesReady
      .then(() => hydrate(RootLayout, options))
      .catch((e) => {
        setTimeout(() => {
          throw e;
        });
      });
```

packages/gemi/client/ComponentContext.tsx
```diff
@@ -95,6 +102,49 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "test") {
...
+ *
+ * Awaiting the imports first costs nothing on the wire (the shell already
+ * emits a `modulepreload` for each of them) and makes `lazy()` return the
+ * component synchronously, so the initial render never suspends and there is
+ * no half-hydrated boundary for a later update to blank.
+ *
+ * Never rejects: a chunk that fails to load must still leave the app to
+ * hydrate and surface the error through the route's error boundary.
+ */
+export const initialViewModulesReady: Promise<unknown> =
```
nstfkc, line 161: **No deadline on the wait.**

`.catch(() => null)` covers *rejection*, but not a request that never settles — a stalled HTTP/2 stream, a service worker that never responds, a proxy holding the connection open. In that case `loadViewModule`'s promise never settles, the `.catch` never fires, and `hydrate()` is never called.

The blast radius changed: before this PR a stuck view chunk cost you that one segment (its server HTML stayed on screen, the rest of the document was interactive). Now nothing anywhere hydrates, and nothing is raised.

A race keeps the win and caps the downside — hydrating without the module just puts you back on the old path for that one route:

```js
? Promise.race([
    Promise.all(initialViewNames(window.__GEMI_DATA__).map((name) =>
      loadViewModule(name).catch(() => null),
    )),
    new Promise((r) => setTimeout(r, 2000)),
  ])
: Promise.resolve();
```

packages/gemi/client/ComponentContext.tsx
```diff
@@ -95,6 +102,49 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "test") {
...
+ * segments through `lazy()`, so on the first client render every one of them
+ * suspends. React keeps the server HTML up while a boundary is merely
+ * suspended, but the moment any update reaches one — the module registry
+ * announcing the chunk, an effect, a settling query — it gives up on that
+ * boundary and shows the fallback instead, which for a view without a
+ * `Loading` export is `null`. The page then sits with its layout and no
+ * content until the chunk's `import()` and dictionaries resolve.
+ *
+ * Awaiting the imports first costs nothing on the wire (the shell already
+ * emits a `modulepreload` for each of them) and makes `lazy()` return the
```
nstfkc: **"costs nothing on the wire" isn't true of what's actually awaited.**

`loadViewModule` awaits `preloadDictionaries(currentLocale(), mark)` before it resolves (line 66), and dictionary locale chunks are reached through the per-locale `import()` calls emitted by `dictionaryPlugin.ts:246-254` — i.e. `dynamicImports`, which `collectModulePreloads` *deliberately* leaves out of the shell head (`modulePreloads.ts:47`: "`dynamicImports` are deliberately left out").

So whole-document hydration is now gated on a serially-discovered, un-warmed round trip: view chunk fetch → parse → dictionary chunk fetch. The `modulepreload` claim holds for the view chunks only.

Concretely on `templates/saas-starter`, which ships dictionaries and a locale switcher: the header's `<button>` handlers — theme toggle, mobile menu, the locale switcher itself — stay dead for that extra round trip on a cold load, where previously they were live as soon as the entry ran. Probably still the right trade against a 300 ms content flash, but the comment shouldn't say it's free.

Two ways out if you want it back: preload the current locale's dictionary chunks in the shell, or await only the module import here and let the dictionary warm-up continue in the background.

packages/gemi/client/ComponentContext.tsx
```diff
@@ -95,6 +102,49 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "test") {
...
+ * component synchronously, so the initial render never suspends and there is
```
nstfkc: **The stated mechanism isn't the one doing the work.**

`lazy()` does not return the component synchronously and the initial render *does* still suspend:

- `loadViewModule` doesn't memoize — it builds a fresh `Promise.resolve(loader()).then(...)` on every call — so `lazy(() => loadViewModule(name))` (line 101) receives a **pending** promise on the hydration render, not a resolved one.
- Even if it did receive a resolved promise, React's `lazy` can't read it synchronously: it attaches `.then`, re-checks `payload._status`, finds it still `Uninitialized` (the callback runs a microtask later), marks it `Pending` and throws.

What actually prevents the blank is the `isNew` guard on line 60. The second call finds the module already in `viewModules`, so it **doesn't notify listeners** — the boundary suspends briefly and resolves with no update ever reaching it.

That matters because it changes what a future refactor must preserve. The real invariant is *"no first registration for a mounted view happens during hydration"*, and it's breakable while still satisfying the comment as written — e.g. notifying on every call rather than only the first, or memoizing the promise so `isNew` stops being the discriminator.

The fourth test doesn't catch this: `hydrationBlank.test.tsx:150` hands `lazy` one already-resolved promise reused across renders, which isn't the production shape (fresh promise per ctor call). It passes for a different reason than the code works. Worth having that test drive `lazy` through a `loadViewModule`-shaped loader so it pins the dedupe rather than the resolved-promise timing.

packages/gemi/client/init.tsx
nstfkc: Agreed, fixed in 2f73b51 with your exact suggestion.

Confirmed the chain you describe: `ViewRouteDispatcher.ts:483` is the `.catch(function(e){setTimeout(function(){throw e})})`, and the comment above it at 478-480 names this failure precisely. It only ever covered `hydrateRoot` by accident of `init()` being synchronous, and I broke that without noticing.

Worth stating that the deadline in the sibling comment does not overlap with this: that one covers a promise that never settles, this one covers a throw out of `hydrate` itself. Different failure, different net.

packages/gemi/client/ComponentContext.tsx
nstfkc: Agreed, fixed in 2f73b51 — `Promise.race` with a named `INITIAL_VIEW_MODULES_DEADLINE_MS = 2000`.

Your framing of it as blast radius is what makes it worth doing. `.catch` was written against "the chunk 404s after a deploy", and a hang is not that. The pre-PR cost of a stuck chunk was one segment keeping its server HTML while the rest of the document stayed interactive; I turned that into a silently dead page, which is strictly worse than the flash I set out to remove.

Noted in the constant’s doc comment that expiring the deadline lands you back on the pre-#352 path for that one route, so a future reader knows the timeout is a floor on badness rather than a tuning knob.

packages/gemi/client/ComponentContext.tsx
nstfkc: Correct, and the comment now says so — but I am keeping the await, so this is a partial pushback.

Verified both halves of your claim: `modulePreloads.ts:47-49` says `dynamicImports` are deliberately left out, and `dictionaryPlugin.ts:246-254` emits the per-locale loaders as exactly those. So the round trip is real and un-warmed, and "costs nothing on the wire" was wrong.

Where I disagree is the second of your two ways out. Awaiting only the module import does not move the wait, it relocates the blank: a view whose dictionary is still in flight suspends on `use()` inside the same boundary this PR is protecting, so the content goes empty again for the same reason, just later and only on dictionary-carrying views. The dictionary await is load-bearing for the fix, not incidental to it.

Your first way out is the right one and I have not done it — warming the current locale’s dictionary chunks from the shell buys the interactivity back without giving up the guarantee. That is a change to `collectModulePreloads` and the shell, so I would rather it be its own PR than ride along here.

The comment now states the cost plainly, names the layout handlers as what goes dead, and records why awaiting less is not the way out.

packages/gemi/client/ComponentContext.tsx
nstfkc: You are right, and this was the most useful comment of the four — the comment described a mechanism that does not exist.

I checked it rather than taking it on faith. Client-rendering `lazy(() => settled)` where `settled` was awaited before the render still commits the fallback first:

```
{ sync: "FB", after: "content" }
```

So the first render suspends regardless, exactly as you describe — `.then` is attached and the status is re-read before the callback has run.

Rewritten in 2f73b51 to say what actually happens: the render suspends either way, and what makes it survive is that `loadViewModule` notifies only on first registration, so with the module already in `viewModules` the call `lazy` makes during hydration is silent. The comment now names the invariant as "no first registration for a mounted view happens during hydration", and calls out that notifying on every call or memoizing the promise would break it while leaving the preload apparently intact.

The test is rebuilt on a `loadViewModule`-shaped loader — fresh promise per ctor call, notify guarded on first registration — so the last case pins the dedupe rather than the resolved-promise timing. Added a separate case asserting `lazy` does *not* resolve synchronously, so the mechanism you ruled out stays ruled out.

PR #423, opened 2026-09-01
Two dev-server crashes, plus the pending client and events work

nstfkc:
Everything that was sitting uncommitted in the working tree, split by topic. The two dev-server fixes at the top are new; the other three were already in the tree.

## Fixes

**`Log.*` threw on an undefined writer** (`79a2d3a`) — `Server.start` awaits `waitForBoot()`, where `LogServiceProvider.boot()` creates the storage directory, and only *then* imports the http layer that sets `ROOT_DIR`. `createStorage()` therefore ran with the variable undefined, hit its guard and bailed, while `boot()` still set `isReady`: no writer, so the first `Log.info()` threw `undefined is not an object (evaluating 'this.writer.write')`. It now resolves the directory through `projectRoot()` — the rule `httpProd` computes `ROOT_DIR` from a moment later — and a failed storage setup degrades to no file sink instead of rejecting on every call.

**"Invalid hook call" after a dev reload of server-only code** (`49d264e`) — editing `app/config/*`, a feature declaration or a controller made the next render throw a null `dispatcher.useContext` from `useRouteData`, inside a view nobody touched. `bun --hot` re-evaluates its whole module graph including node_modules, so after a reload `react`, `react-dom/server` and the Bun-loaded `gemi` are fresh instances — while the Vite server, kept on `globalThis` across reloads, still holds every view module it has evaluated, bound through the externalized `gemi/*` imports to the previous React. Editing a *view* never showed it: Vite's watcher invalidates that module. The fix drops the SSR runner's evaluated modules once per reload.

Both are dev/boot-time only; production builds are unaffected. Folio is on published `gemi@0.58.1`, so both need a release to land there.

## Already in the tree

- **`c38fd3e`** — hydration is held until the initial route's view modules are registered, so a first registration cannot land on a boundary that is mid-hydration and blank its own server HTML. Brings `hydrationBlank.test.tsx` with it.
- **`7c88f06`** — `Event.fake()` says it needs a booted application, instead of surfacing `app()`'s "Boot a Kernel before resolving services" to a caller who never asked to resolve a service.
- **`db9788a`** — the `ai` rfc sketch continues: `useChat` takes an `Agent`, `sendMessage` takes a payload, and `ai` joins the tsconfig include list.

## Verification

Reproduced the reload crash in `templates/saas-starter` (same stack as the report) and confirmed the fix over three consecutive server-file edits: 0 errors, against 3/3 before. Rendered HTML matches the clean baseline apart from the per-boot `appId`; view edits still re-evaluate; a throwing view still reports a source-mapped location; a syntax error still gives the 500 page and recovers. The LogManager crash reproduces on the pre-fix file with `ROOT_DIR` unset and does not after.

`hydrationBlank.test.tsx` passes under vitest (5/5). `packages/gemi/services/events/` has 12 failures, identical on pristine `main` — pre-existing, not from this branch.

https://claude.ai/code/session_014yZRsSsvsHv2n6isjQuBhT

### DIFF
--- a/packages/gemi/server/httpDev.ts
+++ b/packages/gemi/server/httpDev.ts
@@ -217,30 +217,6 @@ export async function httpDev(app: App, instrumentation: Instrumentation) {
     },
   }));
 
-  // `bun --hot` re-evaluates its *whole* module graph on a server-code change —
-  // node_modules included — so after a reload `react`, `react-dom/server` and the
-  // Bun-loaded `gemi` are all fresh instances. This Vite server is not reloaded
-  // with them (it is deliberately kept on `globalThis` above), and its SSR runner
-  // still holds every view module it has evaluated — each bound, through the
-  // externalized `gemi/*` imports, to the *previous* `gemi` and therefore the
-  // previous `react`. The next render then sets the dispatcher on the new React
-  // while a cached view calls hooks on the old one: "Invalid hook call", a null
-  // `dispatcher.useContext` thrown from `useRouteData` inside a view nobody
-  // touched. Editing a *view* never showed it — Vite's own watcher invalidates
-  // that module, so it re-evaluates against the new instances. Editing a
-  // server-only file (`app/config/*`, a feature declaration, a controller) is
-  // the case with nothing to invalidate the views, so drop the evaluated
-  // modules here, once per reload.
-  //
-  // Only the *evaluated* modules go: the server-side transform cache is
-  // untouched, so this costs a re-evaluation and not a re-transform. Clearing
-  // the module graph as well would not have been enough anyway — externalized
-  // deps (`gemi`, and React through it) are cached in the runner by URL and
-  // carry no invalidation flag, which is exactly where the stale React hides.
-  if (isReload) {
-    ssrRunner(vite).clearCache();
-  }
-
   process.env.ROOT_DIR = rootDir;
   process.env.APP_DIR = appDir;
 
@@ -350,6 +326,39 @@ export async function httpDev(app: App, instrumentation: Instrumentation) {
       return await instrumentation(req, requestHandler);
     },
   });
+
+  // `bun --hot` re-evaluates its *whole* module graph on a server-code change —
+  // node_modules included — so after a reload `react`, `react-dom/server` and the
+  // Bun-loaded `gemi` are all fresh instances. This Vite server is not reloaded
+  // with them (it is deliberately kept on `globalThis` above), and its SSR runner
+  // still holds every view module it has evaluated — each bound, through the
+  // externalized `gemi/*` imports, to the *previous* `gemi` and therefore the
+  // previous `react`. The next render then sets the dispatcher on the new React
+  // while a cached view calls hooks on the old one: "Invalid hook call", a null
+  // `dispatcher.useContext` thrown from `useRouteData` inside a view nobody
+  // touched. Editing a *view* never showed it — Vite's own watcher invalidates
+  // that module, so it re-evaluates against the new instances. Editing a
+  // server-only file (`app/config/*`, a feature declaration, a controller) is
+  // the case with nothing to invalidate the views, so drop the evaluated
+  // modules here, once per reload.
+  //
+  // Here and not before `Bun.serve`, with nothing awaited in between, because
+  // both sides of the swap have to see a consistent pair. Until that call
+  // returns the *previous* fetch handler is still serving — old dispatcher, old
+  // `react-dom/server` — and a request landing there after an early clear would
+  // re-evaluate its views against the *new* `gemi`: the same mismatch, reversed.
+  // After it, the new handler is live and must not be given the stale ones. The
+  // two statements are synchronous neighbours, so no request can be dispatched
+  // between them.
+  //
+  // Only the *evaluated* modules go: the server-side transform cache is
+  // untouched, so this costs a re-evaluation and not a re-transform. Clearing
+  // the module graph as well would not have been enough anyway — externalized
+  // deps (`gemi`, and React through it) are cached in the runner by URL and
+  // carry no invalidation flag, which is exactly where the stale React hides.
+  if (isReload) {
+    ssrRunner(vite).clearCache();
+  }
 
   // On a reload `app` has already been rebuilt with the new code and Bun.serve
   // above has swapped in the new fetch handler, so the updated controllers are
--- a/packages/gemi/server/httpDev.ts
+++ b/packages/gemi/server/httpDev.ts
@@ -2,7 +2,7 @@ import { IncomingMessage, ServerResponse } from "node:http";
 import { Socket } from "node:net";
 import { join } from "node:path";
 
-import { createServer, type ViteDevServer } from "vite";
+import { createServer, type RunnableDevEnvironment, type ViteDevServer } from "vite";
 
 import { App } from "../app";
 import gemiVite from "../vite";
@@ -141,6 +141,26 @@ if (!globalThis.__gemiErrorHooked) {
   });
 }
 
+/**
+ * The SSR environment's module runner — what `vite.ssrLoadModule` is a thin
+ * (and, in Vite 8, deprecated) wrapper over. Going through it directly is what
+ * gives `httpDev` a handle on the evaluated-module cache it has to drop after a
+ * `bun --hot` reload; `ssrLoadModule` keeps its own runner private.
+ *
+ * Duck-typed rather than `isRunnableDevEnvironment(environment)`: this server
+ * outlives the module instance that created it, and after a reload the freshly
+ * evaluated `vite` brings a *new* `RunnableDevEnvironment` class object, so the
+ * `instanceof` that predicate performs is false for a perfectly runnable
+ * environment. (The same one-copy-per-reload rule this whole dance is about.)
+ */
+function ssrRunner(vite: ViteDevServer) {
+  const environment = vite.environments.ssr as RunnableDevEnvironment;
+  if (typeof environment?.runner?.import !== "function") {
+    throw new Error("gemi dev requires a runnable Vite SSR environment.");
+  }
+  return environment.runner;
+}
+
 export async function httpDev(app: App, instrumentation: Instrumentation) {
   // `bun --hot` re-runs this module on every server-code change, so keep a
   // single Vite server on `globalThis` instead of spawning a new one (and a new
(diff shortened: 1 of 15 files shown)


packages/gemi/tsconfig.json
```diff
@@ -59,7 +59,8 @@
     "bin/orm",
     "auth",
     "config",
-    "internal"
+    "internal",
+    "ai"
```
nstfkc, line 63: **Adding `"ai"` to `include` breaks `bun run typecheck` and `bun run build`.**

`tsc --noEmit -p tsconfig.json` now exits 2:

```
ai/Agent.ts(76,5): error TS2322: Type 'Agent<AgentTool<any, any>[]>' is not assignable to type 'Agent<T>'.
  Type 'AgentTool<any, any>[]' is not assignable to type 'T'.
```

`Agent.create<const T>` does `const agent = new Agent()`, which instantiates the *default* type argument rather than `T`, then returns it as `Agent<T>`. I confirmed this is new to the branch: swapping in `main`'s `include` list and re-running the same command exits 0.

This is not just the lint job — `build:types` is a bare `tsc` on this config, so `bun run build` fails at that step too, and the PR body says both dev fixes need a release to reach Folio.

Fix is either `new Agent<T>()` in `ai/Agent.ts:74`, or leaving `ai` out of the include list until the sketch typechecks.

packages/gemi/services/logging/LogManager.ts
```diff
@@ -138,14 +156,22 @@ export class LogManager {
     } catch (err) {
       console.log("Error parsing log object", err);
     }
+    // Broadcast.channel("/logs/live", {}).publish(JSON.stringify(logObject));
+    this.config.onLogCreated(logObject);
```
nstfkc: **`onLogCreated` was moved out of the `try`/`catch` that used to contain it.**

Before this commit the call sat inside the writer `try`:

```ts
try {
  this.writer.write(log);
  this.writer.write("\n");
  // Broadcast.channel(...)
  this.config.onLogCreated(logObject);
  this.tryFlush();
} catch (err) {
  console.error("Error writing log", err);
}
```

Now it runs ahead of the writer guard with nothing around it. `onLogCreated` is an app-supplied hook (`services/logging/config.ts`), `log()` is `async`, and every facade method calls it without awaiting (`facades/Log.ts` — `this.getFacadeRoot().log("info", ...)`, result discarded).

So an app whose `onLogCreated` throws synchronously — cyclic metadata hitting a `JSON.stringify` in the hook, a transport that isn't connected yet — turns a floating `Log.info()` into an unhandled rejection, which Bun treats as fatal. That is the same failure shape the rest of this commit is fixing (`Log.*` must never become the caller's error).

Putting the call back inside a `try`/`catch` (its own, if it should stay ahead of the writer guard) restores the old containment.

packages/gemi/services/logging/LogManager.ts
```diff
@@ -36,19 +49,24 @@ export class LogManager {
...
     return this.bootPromise;
   }
 
   private async createStorage() {
-    if (!process.env.ROOT_DIR) {
-      return;
-    }
-    if (!(await exists(`${process.env.ROOT_DIR}/storage`))) {
-      await mkdir(`${process.env.ROOT_DIR}/storage`);
+    const storageDirPath = `${projectRoot()}/storage`;
```
nstfkc, line 67: **`createStorage` lost its "no project root, do nothing" escape.**

The old body opened with:

```ts
if (!process.env.ROOT_DIR) {
  return;
}
```

`ROOT_DIR` is only ever assigned in `server/httpDev.ts:244` and `server/httpProd.ts:31`, so that early return covered every *non-server* kernel boot: `gemi` CLI commands, migrations, and any test that boots a Kernel with `LogServiceProvider` registered. `projectRoot()` always returns a path, so the guard is now unreachable and `LogServiceProvider.boot()` does real file IO in all of them.

Concretely, each such invocation now `mkdir`s `<cwd>/storage/logs`, `readdir`s it, and — via `newLogFile()` — reads the most recent log file's entire text into memory and rewrites it into a fresh `FileSink` (`this.writer.write(await _file.text())`), up to `maxFileSize`, 10 MB by default. In whatever directory the command happened to be run from.

The `projectRoot()` change itself is right for the server path; it's the removal of the "am I actually in a server context" gate that widens it. Either restore a guard, or make the eager `boot()` in `LogServiceProvider` conditional on a server context.

nstfkc (commented):
Correctness review of the diff. Four findings — one release-blocking, one that reintroduces the failure shape the branch is fixing, two narrower.

Cleared while reading: `initialViewNames` matches `ClientRouterProvider`'s `routeManifest[pathname] ?? ["404"]` exactly (`router.pathname` is the route *pattern*, and `is500` is hard-coded `false` at `ClientRouter.tsx:438`, so the missing `is500` branch is unreachable); `window.loaders` is assigned in the inline `bootstrapScriptContent` before any module script runs, so the new module-scope preload cannot race it; `initialViewModulesReady` can neither reject nor hang; `Event.fake()`'s `try`/`catch` around `app()` can only catch the one no-token throw; the `ssrLoadModule` → `environment.runner` swap is correct (`ssrLoadModule` really did keep a private `_ssrCompatModuleRunner`, and the runner's `sourcemapInterceptor` is on by default, so dropping `fixStacktrace: true` costs no source-mapped stacks). `vitest run client/` 187/187 including the new `hydrationBlank.test.tsx` 5/5; `bun test services/events/fake.test.ts` 18/18; `tsconfig.browser.json` typechecks clean; `oxlint` 0 errors.

packages/gemi/server/httpDev.ts
```diff
@@ -197,6 +217,30 @@ export async function httpDev(app: App, instrumentation: Instrumentation) {
...
+  // the case with nothing to invalidate the views, so drop the evaluated
+  // modules here, once per reload.
+  //
+  // Only the *evaluated* modules go: the server-side transform cache is
+  // untouched, so this costs a re-evaluation and not a re-transform. Clearing
+  // the module graph as well would not have been enough anyway — externalized
+  // deps (`gemi`, and React through it) are cached in the runner by URL and
+  // carry no invalidation flag, which is exactly where the stale React hides.
+  if (isReload) {
+    ssrRunner(vite).clearCache();
```
nstfkc: **`clearCache()` here leaves a window where the old handler imports through the cleared runner — the same mismatch, reversed.**

On a `bun --hot` reload the previous `Bun.serve` fetch handler stays live until the new `Bun.serve(...)` call below returns. A request landing in that window is served by the *old* handler — old `ViewRouteDispatcher`, old Bun-loaded `react-dom/server` — but now imports its views through a runner whose evaluated modules were just dropped, so those views re-evaluate and bind to the *new* externalized `gemi` and therefore the *new* `react`.

That is exactly the "Invalid hook call" this hunk fixes, with the two sides swapped. Narrow and dev-only, but it is a new failure mode: before this change that in-flight request got the consistent (stale) pair and rendered fine.

Moving the `clearCache()` to after `Bun.serve` returns — e.g. into the `if (isReload)` block at the bottom of `httpDev`, which already runs post-swap and comments that "Bun.serve above has swapped in the new fetch handler" — would close it.

packages/gemi/tsconfig.json
nstfkc: Fixed in 94ca936 — `new Agent<T>()`, which was the first of your two options. Confirmed the same way you found it: `tsc --noEmit -p tsconfig.json` now exits 0, as does `ide/typescript-plugin/tsconfig.json`, and the sqlite job that was failing at Typecheck is green.

Kept `ai` in the include list rather than dropping it: the sketch is meant to be compiled, and this is exactly the error that not compiling it hid.

packages/gemi/services/logging/LogManager.ts
nstfkc: Right, and it is the failure shape the commit is about — thanks. Fixed in 1f24dbf: the call has its own `try`/`catch` ahead of the writer guard.

One thing beyond restoring the old containment: an `async` hook that rejects is the same unhandled rejection arriving a tick later, and the old code did not catch that either (`onLogCreated` is typed `void | Promise<void>`), so a returned promise gets a `.catch` too. Not awaited — that would put the app's transport in front of the file write.

Verified with a floating `log()` and a `process.on('unhandledRejection')` that exits 1: synchronous throw and async rejection both leave the caller alive and the entry on disk.

packages/gemi/services/logging/LogManager.ts
nstfkc: Agreed — that guard was doing two jobs and I only saw the path one. Fixed in 1f24dbf, taking your second option, with the marker moved to where it is actually true.

`Server.start` now sets `ROOT_DIR` from `projectRoot()` *before* `waitForBoot()`. That is the same ordering bug the original fix was about, one level up: `httpDev`/`httpProd` set the same value a moment later, but only after every provider has booted. `LogServiceProvider.boot()` then prepares storage only when it is set, so `runConsole`, migrations and Kernel tests do no file IO at all.

They stay lazy rather than silent, though: `LogManager.log()` still boots on first use, so a command that actually logs gets its file — which is more than it got before this branch, when that call threw on the undefined writer.

Verified A/B in a scratch cwd: provider boot with `ROOT_DIR` unset → no `storage/` created; with it set → `storage/logs` created; unset plus an explicit `Log` call → file created and the entry written. The template dev server still prepares its log file at boot (existing file reopened, mtime moves at startup).

packages/gemi/server/httpDev.ts
nstfkc: Good catch — the reversed pair is real. Fixed in bc54671, though not quite where you suggested, because the `if (isReload)` block at the bottom would trade one window for the other: by then the *new* handler is live, and anything that yields before the clear lets a request render with the new `react-dom/server` over the stale cached views — the original bug.

So the clear sits immediately after `Bun.serve` returns, with nothing awaited in between. The old handler keeps the modules it was rendering with right up to the swap; the new one gets fresh ones; and because the two statements are synchronous neighbours, no request can be dispatched between them. The comment says that, so the next person does not "tidy" it into the block below.

Re-verified in `templates/saas-starter`: three consecutive server-file edits, 0 "Invalid hook call"; view edits still re-evaluate (probe string changes across edits); a throwing view still reports a source-mapped location and a syntax error still gives the 500 page and recovers.

packages/gemi/tsconfig.json
nstfkc: Follow-up: the sketch is now `@ts-nocheck`'d instead (6c1340e), which is closer to your second option without taking `ai` back out of `include`. Nothing exports `gemi/ai` and nothing in the package imports it, so `tsc` is its only reader — per-file keeps the files in the project and puts the exemption where the person editing them will see it.

`new Agent<T>()` stays; it is right whether or not anything is checking. Verified the directive actually covers the folder: a deliberate `const x: number = "str"` in `ai/Agent.ts` produces no error, and both tsconfigs are clean.

PR #424, opened 2026-09-01
List the features, and see your own switch flip

nstfkc:
An admin screen for features needs two things the facade did not have: the set of features that exist, and a way to make a write to the switch table visible without waiting out the TTL.

## `Features.list()`

One descriptor per declaration — `key`, `describe`, `rollout`, `targeted`, `serverOnly`, `salt`, `active`.

```ts
public async index() {
  return { features: await Features.list() };
}
```

**The declarations are what exists.** A feature is on the list because `app/features` declares it, never because a row was inserted — a row for an undeclared key is litter, and the store already says so.

**`active: undefined` means no row**, which is deliberately not the same as a row saying `false`. One feature was never switched on; the other was switched off by somebody. A toggle UI that renders those the same cannot tell "never shipped" from "killed".

**`when` is not on the descriptor and cannot be.** It is a function over the viewer, so the only honest answer to "who does this target" is to run it. `targeted` reports that targeting exists; `Features.for(subject).explain(key)` answers it one subject at a time.

Server-side only — `rollout` and `targeted` describe who is in an experiment. With `enabled: false` the declarations still list, `active` is `undefined` throughout, and no query is issued.

## `Features.invalidate()`

```ts
await FeatureFlag.update({ where: { key }, data: { active } });
await Features.invalidate();
return { features: await Features.list() };
```

The obvious call here is `refresh()`, and it is wrong twice.

It **joins whatever load is already in flight** — the single-flight that keeps a cold start under load down to one query — and that load issued its query at some earlier moment, possibly before the `UPDATE` committed. So the admin flips a switch, the screen reloads, and it says the switch is still off.

It also **leaves the request's evaluation memo in place**, so a handler that read the feature before writing to it is answered from before its own write.

`invalidate()` waits the in-flight load out, starts a fresh one behind it, and clears the memo.

Still **process-local**. Every other instance is serving its own snapshot and converges within `ttl` — this is not cross-instance invalidation and does not pretend to be. What it closes is the one window that reads as a bug: the operator watching their own change fail to happen.

## Checks

`bun run typecheck` clean. Full suite: 3920 passed, 1 skipped, 162 files — including 20 new tests in `FeatureManager.test.ts` and `FeatureFlagStore.test.ts`. Two of them pin the ordering the design turns on: a load that started before the write must not be what `invalidate()` settles on, and a feature the request already read must be re-evaluated after one.

`docs/feature-flags.md` gains an "An admin screen" section; `docs/llms-full.txt` is regenerated from it.

`oxfmt --check` flags `packages/gemi/services/index.ts`, but flags it identically on `main` — pre-existing, untouched here beyond one added type export.

https://claude.ai/code/session_01U7ZxzSCxAu6nUoFKZrnerQ

### DIFF
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -153,7 +153,7 @@ export default defineFeaturesConfig({
 });
 ```
 
-`ttl` is the propagation delay in seconds — how long a loaded snapshot is reused before the next refresh. It is how long switching something off takes to reach every instance, since there is no cross-instance invalidation. Lower it if thirty seconds is too long during an incident; the cost is one query per instance per window.
+`ttl` is the propagation delay in seconds — how long a loaded snapshot is reused before the next refresh. It is how long switching something off takes to reach every instance, since there is no cross-instance invalidation. `Features.invalidate()` collapses that window for the process that made the write — see [An admin screen](#an-admin-screen) — but the others still wait out the TTL. Lower it if thirty seconds is too long during an incident; the cost is one query per instance per window.
 
 ## Reading a feature
 
@@ -173,12 +173,52 @@ await Features.enabled("new-checkout");
 await Features.explain("new-checkout"); // { value, reason } — server only
 await Features.all(); // the client-visible map
 await Features.for({ user }).enabled("digest-v2");
+await Features.list(); // every declared feature and its switch
 await Features.refresh(); // reload this process now
+await Features.invalidate(); // reload after writing to the table
 ```
 
 Everything is async. After the boot-time warm-up each call settles in a microtask with no I/O, but the promise is kept because the first call in a cold process does hit the database — and a sync variant would have to answer that window with "off", which is a feature silently vanishing while a deploy rolls.
 
 `explain` is **server-only**. `reason` distinguishes "targeted by name" from "landed in the rollout", which is a fact about the viewer; never serialize it into a response.
+
+## An admin screen
+
+`Features.list()` is the read side. It returns one descriptor per feature the code declares — key, `describe`, `rollout`, `serverOnly`, whether the declaration carries targeting, and the switch:
+
+```typescript
+[
+  { key: "new-checkout", describe: "Rebuilt checkout", rollout: 20, targeted: false, serverOnly: false, salt: undefined, active: true },
+  { key: "digest-v2", describe: undefined, rollout: undefined, targeted: true, serverOnly: false, salt: undefined, active: undefined },
+]
+```
+
+The declarations are what exists. A feature is on the list because `app/features` declares it, never because a row was inserted — a row for an undeclared key is litter and does not appear. `active: undefined` is **no row**: a feature deployed but never switched on, which is not the same as a row that says `false`, and is the distinction the screen has to show. One is untouched, the other is somebody's decision.
+
+The `when` function is not on the descriptor and cannot be. It is a function over the viewer, so the only honest answer to "who does this target" is to run it: `targeted` reports that targeting exists, and `Features.for({ user }).explain("new-checkout")` answers it for one subject.
+
+**Server-side only.** `rollout` and `targeted` describe who is in an experiment.
+
+### Writing the switch
+
+The write is an ordinary `UPDATE` on your own model — the framework owns no mutation path, because the table is yours. What it does own is the cache in front of it:
+
+```typescript
+public async update(request: HttpRequest<{ active: boolean }>) {
+  await FeatureFlag.update({
+    where: { key: request.params.key },
+    data: { active: request.input.get("active") },
+  });
+
+  await Features.invalidate();
+
+  return { features: await Features.list() };
+}
+```
+
+`invalidate()` rather than `refresh()`, for two reasons that both amount to "the admin must see their own write". `refresh()` joins whatever load is already in flight, and that load may have queried before the update committed — so the screen could come back saying the switch is still off a moment after flipping it. And `invalidate()` clears the request's evaluation memo, so a feature this request already read is re-evaluated instead of answered from before the write.
+
+It is **process-local**, like `refresh()`. Every other instance is still serving its own snapshot and converges within `ttl`. What this fixes is the one window that reads as a bug — the operator who flips a switch, reloads, and is told for the next thirty seconds that nothing happened.
 
 ## Hiding a feature's existence
 
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -187,36 +187,66 @@ Everything is async. After the boot-time warm-up each call settles in a microtas
 `Features.list()` is the read side. It returns one descriptor per feature the code declares — key, `describe`, `rollout`, `serverOnly`, whether the declaration carries targeting, and the switch:
 
 ```typescript
-[
-  { key: "new-checkout", describe: "Rebuilt checkout", rollout: 20, targeted: false, serverOnly: false, salt: undefined, active: true },
-  { key: "digest-v2", describe: undefined, rollout: undefined, targeted: true, serverOnly: false, salt: undefined, active: undefined },
-]
+{
+  unavailable: false,
+  features: [
+    { key: "new-checkout", describe: "Rebuilt checkout", rollout: 20, targeted: false, serverOnly: false, salt: undefined, active: true },
+    { key: "digest-v2", describe: undefined, rollout: undefined, targeted: true, serverOnly: false, salt: undefined, active: undefined },
+  ],
+}
 ```
 
 The declarations are what exists. A feature is on the list because `app/features` declares it, never because a row was inserted — a row for an undeclared key is litter and does not appear. `active: undefined` is **no row**: a feature deployed but never switched on, which is not the same as a row that says `false`, and is the distinction the screen has to show. One is untouched, the other is somebody's decision.
 
+`unavailable` is why the list is not a bare array. It means no snapshot has ever loaded — an unreachable database at boot — so every `active` is _unknown_ rather than absent. Collapsing the two would tell an operator that no feature has ever been switched on, about features that are switched on in the table, at the moment they are most likely to act on it. Render it as an error, and do not offer to flip switches whose current state you do not know.
+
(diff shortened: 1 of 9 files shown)


docs/feature-flags.md
```diff
@@ -173,13 +173,53 @@ await Features.enabled("new-checkout");
...
+
+### Writing the switch
+
+The write is an ordinary `UPDATE` on your own model — the framework owns no mutation path, because the table is yours. What it does own is the cache in front of it:
+
+```typescript
+public async update(request: HttpRequest<{ active: boolean }>) {
+  await FeatureFlag.update({
+    where: { key: request.params.key },
+    data: { active: request.input.get("active") },
```
nstfkc: **medium** — `request.input.get("active")` treats the async `input()` **method** as a property.

`HttpRequest.input()` is `async input(): Promise<Input<T>>` (`packages/gemi/http/HttpRequest.ts:341`), so anyone copying this admin-controller example gets `TypeError: request.input.get is not a function`, or — if the throw is swallowed upstream — writes `undefined` into the column. It should be:

```ts
const input = await request.input();
// ... data: { active: input.get("active") }
```

Same line, second issue: `request.params.key` on the line above needs the second generic — `HttpRequest<{ active: boolean }, { key: string }>`. `Params` defaults to `Record<string, never>` (`HttpRequest.ts:167-177`), so `params.key` is `never`, which is assignable to anything and therefore type-checks silently into `where: { key: never }` rather than erroring. Every other doc that reads params declares it (`docs/controllers.md:38-39`, `docs/routing.md:54`).

The identical block in `docs/llms-full.txt` needs the same two fixes.

packages/gemi/services/features/FeatureManager.ts
```diff
@@ -173,6 +216,20 @@ export class FeatureManager {
...
 
+  /**
+   * The whole switch map in one load, for `list()`.
+   *
+   * Separate from `switchFor` because listing asks about every key at once and
+   * has no use for `unavailable` — a store that has never loaded and a table
+   * with no rows both mean "nothing is switched on", which `list()` already
+   * renders as `active: undefined` per feature.
+   */
+  private async switches(): Promise<Map<string, boolean>> {
```
nstfkc: **medium** — `switches()` discards `snapshot.unavailable`, so `list()` renders "the store has never loaded" as "never switched on".

Concrete scenario: a process boots while the database is unreachable. `FeatureFlagStore.handleFailure` installs `{ active: new Map(), loadedAt: now, unavailable: true }`, `get()` returns it for a full TTL, and every descriptor comes back `active: undefined`. The admin screen then states that no feature has ever been switched on — for features that are in fact switched on in the table — which is exactly the distinction `FeatureDescriptor.active` is documented to carry (`undefined` = no row, a deliberate decision-vs-untouched signal).

The doc comment's premise ("a store that has never loaded and a table with no rows both mean 'nothing is switched on'") is true for *evaluation*, where `unavailable` fails closed, but false for a *display*: the operator is shown a confidently wrong picture at the moment they are most likely to act on it (flip switches that are already on). `FeatureDescriptor` currently has no way to say "unknown" — either propagate `unavailable` into the descriptor/return shape, or have `list()` refuse rather than report `undefined` for a store that never loaded.

packages/gemi/facades/Features.ts
```diff
@@ -67,8 +67,57 @@ export class Features extends Facade {
...
+   *
+   * ```ts
+   * public async update(request: HttpRequest<{ active: boolean }>) {
+   *   await FeatureFlag.update({ where: { key }, data: { active } });
+   *   await Features.invalidate();
+   *   return { features: await Features.list() };
+   * }
+   * ```
+   *
+   * Unlike `refresh()` this never settles on data older than the moment it was
```
nstfkc: **medium** — "never settles on data older than the moment it was called" is false when the reload fails.

`store.invalidate()` → `refresh()` → `load()`, and `load()` swallows the error into `handleFailure` (`FeatureFlagStore.ts:150-160`), which keeps the pre-write snapshot **and** moves `loadedAt` forward to `now`. So on a transient read failure right after the `UPDATE`: `invalidate()` resolves successfully, the `list()` on the next line returns the pre-write switches, and the stale snapshot is now pinned for a *further full TTL* from the failure — strictly worse than not calling `invalidate()` at all, and precisely the outcome the method exists to prevent.

`invalidate(): Promise<void>` gives the caller no signal to detect this. Note the store's own comment at `FeatureFlagStore.ts:92-94` reasons about a rejection leaving the stale snapshot in place — `load()`'s internal catch produces that same end state without ever rejecting. Either surface the failure (return the snapshot / throw / expose `unavailable`) or qualify the guarantee in both this JSDoc and `docs/feature-flags.md`.

packages/gemi/services/features/FeatureManager.ts
```diff
@@ -73,6 +93,29 @@ export class FeatureManager {
...
+   * they flipped it on, and is told for the next thirty seconds that nothing
+   * happened.
+   */
+  async invalidate(): Promise<void> {
+    // The per-request memo is why this is not simply `refresh()`. A handler that
+    // toggles a row and then reads the feature back in the same request would
+    // otherwise be answered from the evaluation it memoised before the write.
+    // Cleared unconditionally: it costs nothing, and the early returns below are
+    // about the database, which the memo is not.
+    RequestContext.getStore()?.featureEvaluations?.clear();
```
nstfkc: **low** — the per-request memo is cleared *before* `await this.store.invalidate()` rather than after.

Anything in the same request that evaluates a feature during that await — a `Promise.all` in the handler, an SSR render already in flight — goes through `evaluateIn`, reads the still-stale snapshot via `store.get()`, and re-memoises the pre-write value into `store.featureEvaluations`. That entry then survives for the rest of the request, which is the exact failure this `clear()` exists to prevent.

Moving the clear to after `await this.store.invalidate()` closes the window at no cost; the early returns for `enabled: false` / no declarations would just need the clear duplicated or restructured.

docs/feature-flags.md
```diff
@@ -173,13 +173,53 @@ await Features.enabled("new-checkout");
...
+  });
+
+  await Features.invalidate();
+
+  return { features: await Features.list() };
```
nstfkc: **low** — this example hands `Features.list()` straight to the browser with no admin gate mentioned.

`list()` includes `serverOnly` features (asserted by the new test "lists server-only features, which `forClient` omits") and every descriptor carries `rollout` and `targeted`. The facade JSDoc on `Features.list()` (`packages/gemi/facades/Features.ts:89`, not quotable inline from this hunk) says "**Server-side only.** `rollout` and `targeted` describe who is in an experiment, which is not a fact to hand the browser" — and then the documented usage returns exactly that in a response body.

Given that this same document has a "Hiding a feature's existence" section whose whole point is that a key like `project-nightingale` must never reach the HTML, the example should show the route behind admin middleware, or state explicitly that the response must be restricted. As written, a reader who copies it onto an ungated route leaks the unannounced keys the `serverOnly` flag exists to protect.

nstfkc (commented):
Review of the feature-flags admin surface. The `FeatureFlagStore.invalidate()` single-flight ordering holds up as far as I can trace it — the `.finally` that nulls `inflight` is chained ahead of every awaiter, so the post-await `refresh()` can only ever join a load created after the call — and the new tests are coherent. Five issues, inline.

docs/feature-flags.md
nstfkc: Fixed in 5dc5787c. `const input = await request.input()`, and `HttpRequest<{ active: boolean }, { key: string }>` so `params.key` is `string` rather than `never`. Applied in the page, in `llms-full.txt`, and in the `Features.invalidate()` JSDoc, which had the same block with bare `{ key }` / `{ active }` identifiers.

packages/gemi/services/features/FeatureManager.ts
nstfkc: Agreed, and the framing is the fix: `unavailable` is a fact about the load, not about any feature, so putting it on descriptors would invite reading it off one of them. `list()` now returns `{ features, unavailable }`, and `switches()` carries the bit out with the map instead of discarding it. The doc comment you quoted is rewritten to say exactly what you said — true for evaluation, false for a display.

`enabled: false` deliberately reports `unavailable: false`: the application turned features off on purpose, which is not the store failing to answer. Covered by two new tests, one for the never-loaded store and one asserting a kept snapshot after a failed reload is still available.

packages/gemi/facades/Features.ts
nstfkc: Surfaced rather than qualified — "strictly worse than not calling it at all" is a defect, not a caveat.

`invalidate()` throws `FeatureReloadError` when its reload failed, detected with a generation counter incremented on successful loads and read *after* the in-flight wait (a load that succeeded before the call is not this call's success; there is a test pinning that).

I also fixed the second half you identified: `handleFailure` expressed the retry backoff by moving the kept snapshot's `loadedAt` forward, which is what bought the failure an extra TTL and made stale switches claim to be freshly loaded. The clock and the timestamp are now two fields — `retryAfter` gates the background refresh in `get()`, `loadedAt` says when the switches were actually read. Tests cover the throw and that `loadedAt` no longer moves.

packages/gemi/services/features/FeatureManager.ts
nstfkc: Right — moved into a `finally` after `await this.store.invalidate()`, with the early returns restructured into the guard so the clear runs on every path. `finally` rather than after the await, because a failed reload leaves the memo just as suspect as a successful one.

New test drives exactly your scenario: a read lands mid-reload, memoises the pre-write `false`, and the read after the await still comes back `true`.

docs/feature-flags.md
nstfkc: Added a "Gate the route" subsection before the controller examples: the route now goes through `.middleware(["admin"])`, with a sentence spelling out that the listing carries `serverOnly` keys and every rollout, cross-linked to "Hiding a feature's existence". The facade JSDoc says the same thing where it previously only said "server-side only" and then showed an ungated handler.

PR #480, opened 2026-09-03
A body that is not application/json is refused before it is read

### LINKED ISSUE #448: ai: readJsonBody parses any body as JSON regardless of Content-Type

`AgentController#readJsonBody` deliberately reads `raw.text()` and `JSON.parse`s it whatever the `Content-Type` says (to accept `application/json; charset=utf-8`). The side effect is that a cross-site `<form method=POST enctype=text/plain>` whose field name/value concatenate into valid JSON is also accepted, which removes the usual "JSON-only endpoints are not simple-request CSRF targets" defense.

Today this is mitigated because the session cookie is `SameSite=Lax` (`http/requestContext.ts`), and `CSRFMiddleware` is opt-in (and only verifies the cookie, which the browser sends automatically — it is not a double-submit check). If anyone relaxes SameSite for a cross-domain frontend, the agent route becomes CSRF-able.

**Fix**: accept only `Content-Type` starting with `application/json` (prefix match, so charset params still work) and answer 415 otherwise. Consider making `CSRFMiddleware` compare the cookie against a header/body token as well.

nstfkc:
## The problem

`AgentController#readJsonBody` read `raw.text()` and `JSON.parse`d it regardless of `Content-Type`. That was deliberate — `req.input()` matches the type exactly, so `application/json; charset=utf-8` produced an empty body and a lost turn — but ignoring the type altogether gives up the property that makes a JSON-only endpoint safe from a plain cross-site form. A `<form method="POST" enctype="text/plain">` whose single field is named `{"text":"` and valued `"}` posts the body `{"text":"="}`, the browser sends it without a preflight, and the parser turned it into a turn on a logged-in user's conversation.

This is closed today by the `SameSite=Lax` session cookie, but the route should not depend on that; relaxing SameSite for a cross-domain frontend would have made `/chat`, `/chat/attach` and `/chat/stop` CSRF targets.

## The fix

`readJsonBody` now accepts only a `Content-Type` whose value starts with `application/json` and answers **415** otherwise, in the controller's usual `{ error: { code, message } }` shape with `code: "unsupported_media_type"`. `stream`, `attach` and `stop` all go through it, so all three are covered.

## Decisions

- **Media type equality, parameters stripped, case-insensitive.** The header value is cut at the first `;`, trimmed and lowercased, and must equal `application/json`. So `; charset=utf-8` keeps working (the reason the exact match was abandoned in the first place) while `application/json-seq`, `application/jsonp` and `application/json-patch+json` are refused; lowercased because media types compare case-insensitively and some proxies rewrite them. The first commit used a prefix match; the review pointed out it was looser than the comment described, and the second commit narrowed it. The CSRF property only needs the three simple-request types (`text/plain`, `application/x-www-form-urlencoded`, `multipart/form-data`) to be refused, and they are.
- **Checked before the body is read, not after.** A refused body is not consumed, and the refusal does not depend on what the bytes were — a `text/plain` body that happens to be valid JSON is refused the same way as one that is not.
- **Only a request that carries a body is held to it.** A bodiless POST (`raw.body === null`) has no type to check and still reads as `{}`, so the existing "a turn that just lets the model continue" behaviour is unchanged. A body with *no* `Content-Type` at all is a 415: it is what a hand-written `fetch` with a string body and no header sends (as `text/plain;charset=UTF-8` in a browser, untyped in Bun), and there is no safe reading of it.
- **The status rides on `ParsedBody` rather than a second return type.** `ParsedBody` gained optional `status` and `code` fields and `invalidRequest` takes the parsed result, defaulting to the 400 / `invalid_request` it always sent. This keeps the "one shape with an optional `error`" design that the type's comment explains (the package compiles without `strictNullChecks`, so a discriminated union would not narrow) and leaves the three call sites as a one-line change each.

## Callers checked

- `useChat.tsx` `post()` sends `Content-Type: application/json` on every JSON request (`stream`, `attach`, `stop`). The `/files` upload deliberately sends no header (multipart boundary) and goes through `upload()`, which does not use `readJsonBody`. No client change needed.
- `AgentController.test.ts`: both `jsonRequest` and `rawRequest` already default to `application/json`, and the one test that posts without a header posts without a body, which is the case that stays a 200. No existing test needed loosening.
- `http/ApiRouter.agent.test.ts`, `ai/entrypoints.test.ts`: do not post bodies.

## Verification

From `packages/gemi`:

- `bun --bun vitest run ai` — 19 files, **458 passed, 20 skipped** (the 20 are `ai/live` skipping itself without API keys; pre-existing)
- `bun run test:types` — 8 files, **84 passed**, no type errors

New tests, in `describe("the request body")`: the charset case now asserts the 200; `Application/JSON; charset=UTF-8` passes; `text/plain` with a valid-JSON body, `application/x-www-form-urlencoded` and `multipart/form-data` all 415 with no run started; a body with no type at all 415s; `attach` and `stop` return the same 415 and the run is not stopped.

## Adjacent, not fixed here

The issue also suggests making `CSRFMiddleware` compare the cookie against a header or body token. Today it only verifies the cookie's signature, which the browser sends on its own, so it is not a double-submit check. That is an `http/` change with its own surface (where the client gets the token, how forms carry it) and is left for a separate PR.

Fixes #448

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -2,7 +2,8 @@ process.env.SECRET ??= "agent-test-secret";
 
 import { describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
-import type { AgentProvider, ProviderEvent, ProviderStreamParams } from "./AgentProvider";
+import type { AgentProvider, ProviderEvent } from "./AgentProvider";
+import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
 import type {
@@ -76,53 +77,6 @@ const usage = (inputTokens: number, outputTokens: number): Usage => ({
   outputTokens,
   totalTokens: inputTokens + outputTokens,
 });
-
-/**
- * Scripted `ProviderEvent`s, one script per model call.
- *
- * The real provider is written against the same interface elsewhere; depending
- * on it here would make these tests a test of two things at once, and would
- * need a network.
- */
-class FakeProvider {
-  readonly model = "fake";
-  readonly capabilities = {
-    reasoning: true,
-    structuredOutput: true,
-    fileInput: true,
-    parallelToolCalls: true,
-    toolSearch: true,
-  };
-  readonly calls: ProviderStreamParams[] = [];
-
-  constructor(private scripts: ProviderEvent[][]) {}
-
-  stream(params: ProviderStreamParams) {
-    this.calls.push(params);
-    const script = this.scripts[this.calls.length - 1] ?? [
-      { type: "finish", reason: "stop", usage: usage(0, 0) },
-    ];
-    return (async function* () {
-      for (const event of script) yield event;
-    })();
-  }
-
-  async upload() {
-    return "file_1";
-  }
-
-  normalizeError(error: unknown) {
-    return {
-      code: "provider_error" as const,
-      message: error instanceof Error ? error.message : String(error),
-      retryable: false,
-    };
-  }
-}
-
-function fakeProvider(...scripts: ProviderEvent[][]) {
-  return new FakeProvider(scripts) as unknown as AgentProvider & FakeProvider;
-}
 
 function deferred<T = void>() {
   let resolve!: (value: T) => void;
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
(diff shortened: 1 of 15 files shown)


nstfkc (commented):
Fresh review of #480 at d7ede78 (base `ai/14-attach-miss` @ 2ff5cbb).

**Findings**

F1 [nit] packages/gemi/ai/AgentController.ts:580 — `startsWith("application/json")` is looser than the comment describes

`value.trim().toLowerCase().startsWith("application/json")` also accepts `application/jsonp`, `application/json-seq`, `application/json-patch+json` and any other type that merely begins with those bytes, not only `application/json` plus parameters. None of those is a simple-request type, so the CSRF property the PR is after is unaffected and this is purely tidiness. If you want the check to say exactly what the doc comment says (bare type or type followed by parameters), `type === "application/json" || type.startsWith("application/json;")` on the lowercased, trimmed value does it at the same cost. Fine to leave as is.

**What I checked and found no problem with**

- The failure scenario in #448 is reproduced by the new `text/plain` test: without the fix, `{"text":"hi"}` under `text/plain` parses, `calls` has one entry and the response is a 200, so the test would fail on the base branch. Same for the no-header test. `application/x-www-form-urlencoded` and `multipart/form-data` are covered, and the refusal happens before `raw.text()`, so it does not depend on the bytes.
- Bodiless POSTs. The fix relies on `raw.body === null` to skip the type check. I confirmed on a real `Bun.serve` (Bun 1.3.14) that a POST with no body, an empty-string body, an explicit `Content-Length: 0`, and a raw-socket `POST` with neither header all arrive with `req.body === null`, so a header-less bodiless POST still reads as `{}` on a live server, not only with `new Request()` in tests.
- Callers: `stream`, `attach`, `stop` are the only users of `readJsonBody`/`invalidRequest`; `upload` reads `formData()` directly and is untouched. `useChat.post()` sends `Content-Type: application/json` on every JSON request; the `/files` upload deliberately sends none and does not go through `readJsonBody`. `http/ApiRouter.agent.test.ts` and `ai/entrypoints.test.ts` post no bodies. No docs show a hand-written `fetch` to the agent route without the header.
- `stop` on a 415 returns a `Response` and leaves the run running (asserted by `run.stopped === false`), consistent with the existing 400 handling and its comment about not answering `{ stopped: false }`.
- `ParsedBody` gaining optional `status`/`code` keeps the non-narrowing shape the type's comment explains; `invalidRequest` defaults preserve the 400/`invalid_request` contract at all three call sites. The `stop` doc comment was updated to match ("the union is the error").
- Minimality: 2 files, no reformatting, no unrelated hunks, no TODOs, no secrets. Comments and the commit message are in the module's explanatory voice.
- From the worktree's `packages/gemi`: `bun --bun vitest run ai` — 19 files, 458 passed, 20 skipped (the `ai/live` self-skip); `bunx tsc --noEmit -p tsconfig.json` — clean; `bun run test:types` — 8 files, 84 passed, no type errors.

**Verdict:** approve — fixes #448 as stated; the one finding is a nit that does not need to block.

nstfkc:
Thanks for the review. One finding, addressed:

- **F1** (nit, `startsWith("application/json")` looser than the doc comment) — fixed in 71d39b14. `isJsonContentType` now cuts the value at the first `;`, trims and lowercases what is left, and compares it for equality with `application/json`, so `Application/JSON; charset=utf-8` still passes while `application/json-seq`, `application/jsonp` and `application/json-patch+json` no longer do. The doc comment was reworded to say exactly that, and the 415 refusal test gained `application/json-seq` so the narrowing is pinned.

Verified from `packages/gemi` in the worktree: `bun --bun vitest run ai` (19 files, 458 passed, 20 skipped — the live suite skipping itself), `bunx tsc --noEmit -p tsconfig.json` (clean), `bun run test:types` (84 passed, no type errors). Both touched files are still oxfmt-clean.

PR #481, opened 2026-09-03
A message resolved on a later turn replaces its earlier self in the store

### LINKED ISSUE #449: ai: default store duplicates the amended message after a threaded approval

**Verified with a script.**

When a turn resolves a pending call, `Agent.amend` clones the earlier assistant message, attaches the result, and pushes the clone onto `produced`. `AgentController.persistRun` then calls `store.appendMessages(threadId, result.messages)`, and `MemoryAgentStore.appendMessages` **pushes**. The thread now holds the same message id twice — the original without the result and the clone with it.

```
messagesInThread: 4
duplicateMessageIds: ["msg_a4e8…"]
next request items: message | function_call:call_1 | function_call:call_1 | function_call_output:call_1 | message
```

I checked live: OpenAI accepts the duplicated `function_call`, so this is token waste and a transcript the model sees twice rather than a 400. But a real store keyed by message id will either throw on the primary key or upsert, and the docs on `AgentStreamParams.onMessage` already say "a store keyed by id should upsert". The built-in store doesn't, and `reconcileToolPairs` in `providers/request.ts` dedupes repeated *outputs* but not repeated *calls*.

**Fix**
- `MemoryAgentStore.appendMessages`: upsert by `id` (replace in place, append if new). Add a test that a threaded approval leaves one message per id.
- `reconcileToolPairs`: drop a second `function_call` with an already-seen `call_id` as a belt-and-braces guard.
- Document `AgentStore.appendMessages` as upsert semantics.

nstfkc:
## Problem

When a turn resolves a pending call, `Agent.amend` clones the earlier assistant message, attaches the result, and pushes the clone onto `produced` under the id the message already had. `AgentController.persistRun` hands `result.messages` to `store.appendMessages`, and `MemoryAgentStore.appendMessages` pushed. After a threaded approval the thread held the same message id twice — the original without the result and the clone with it — and every following request carried `function_call:call_1 | function_call:call_1 | function_call_output:call_1`.

OpenAI accepts the duplicate, so on the default store the cost was tokens and a transcript the model reads twice rather than a 400. A store backed by a table keyed on message id would have thrown on the primary key instead. `AgentStreamParams.onMessage` already documented that a store keyed by id should upsert; the built-in store did not, and the interface did not say so.

## Fix

- `MemoryAgentStore.appendMessages` upserts by message `id`: replace in place when the id is held, append otherwise. The amended message keeps the position the conversation gave it, which is what the next request needs.
- `AgentStore.appendMessages` on the interface now documents upsert semantics, since the name suggests plain append and an app writing its own store reads the interface rather than the memory implementation.
- `reconcileToolPairs` in `providers/request.ts` drops a second `function_call` with an already-seen `call_id`, mirroring the existing guard for a second output. Belt and braces for a store that did not follow the contract.

`Agent.amend` is untouched: the store contract is the fix, per the issue.

## Decisions

- **Upsert lives in the store, not in `persistRun`.** The controller could dedupe before calling the store, but then a custom store would still see duplicates through `onMessage`, and the interface would still be lying about append. Putting it on the contract means every store gets it, and a table keyed by id gets it for free.
- **In-place replacement rather than delete-and-append.** The amended message is the one holding the tool call; if it moved to the end of the thread, the next request would place the call after messages that came later in the conversation. The controller test checks the index.
- **The call guard is silent, like the output guard.** Dropping the second call is not a lossy operation — it is the same call — so there is nothing to warn about, and the comment on `reconcileToolPairs` says where duplicates come from.
- **The controller test uses a real `Agent` behind the controller** rather than `StubAgentRun`. What the store ends up holding depends on what `Agent.amend` reports, and a stub run reports whatever the test hands it, so a stub-based test would only have asserted the store's own behaviour, which the store unit test already does. The test sets `process.env.SECRET` the way `Agent.test.ts` does, because a pending call has to be signed. The scripted `FakeProvider` that `Agent.test.ts` used to define locally now lives in `providers/fakeProvider.ts`, shared by both test files (source tree rather than a test file, following `store/stubAgentRun.ts`), and `Agent.test.ts` drops its own copy.
- **The upsert test mints its thread with `store.createThread`** rather than a hand-picked id, because the PR below this one in the stack answers 404 for a thread the store never created.

## Verification

From `packages/gemi`:

- `bun --bun vitest run ai` — 19 files, 457 passed, 20 skipped

Each of the three new tests was also run with its fix temporarily reverted and failed as expected.

## Adjacent, not changed here

`persistRun` calls `store.appendMessages` once per run with every message the run produced; a store whose `appendMessages` is not atomic could persist the user turn and not the assistant turn if it throws halfway. That is a store-implementation concern and out of scope for this issue.

Fixes #449

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -2,7 +2,8 @@ process.env.SECRET ??= "agent-test-secret";
 
 import { describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
-import type { AgentProvider, ProviderEvent, ProviderStreamParams } from "./AgentProvider";
+import type { AgentProvider, ProviderEvent } from "./AgentProvider";
+import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
 import type {
@@ -76,53 +77,6 @@ const usage = (inputTokens: number, outputTokens: number): Usage => ({
   outputTokens,
   totalTokens: inputTokens + outputTokens,
 });
-
-/**
- * Scripted `ProviderEvent`s, one script per model call.
- *
- * The real provider is written against the same interface elsewhere; depending
- * on it here would make these tests a test of two things at once, and would
- * need a network.
- */
-class FakeProvider {
-  readonly model = "fake";
-  readonly capabilities = {
-    reasoning: true,
-    structuredOutput: true,
-    fileInput: true,
-    parallelToolCalls: true,
-    toolSearch: true,
-  };
-  readonly calls: ProviderStreamParams[] = [];
-
-  constructor(private scripts: ProviderEvent[][]) {}
-
-  stream(params: ProviderStreamParams) {
-    this.calls.push(params);
-    const script = this.scripts[this.calls.length - 1] ?? [
-      { type: "finish", reason: "stop", usage: usage(0, 0) },
-    ];
-    return (async function* () {
-      for (const event of script) yield event;
-    })();
-  }
-
-  async upload() {
-    return "file_1";
-  }
-
-  normalizeError(error: unknown) {
-    return {
-      code: "provider_error" as const,
-      message: error instanceof Error ? error.message : String(error),
-      retryable: false,
-    };
-  }
-}
-
-function fakeProvider(...scripts: ProviderEvent[][]) {
-  return new FakeProvider(scripts) as unknown as AgentProvider & FakeProvider;
-}
 
 function deferred<T = void>() {
   let resolve!: (value: T) => void;
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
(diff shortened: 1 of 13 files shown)


nstfkc (commented):
Fresh review of #481 against issue #449, on branch `ai/18-store-upsert` at 6386862 stacked on `ai/14-attach-miss` (2ff5cbb).

**What I checked**

- Re-derived the failure from the issue: `Agent.amend` clones the earlier assistant message, pushes the clone onto `produced` under the same id, `persistRun` hands `result.messages` to `store.appendMessages`, and the memory store pushed. Read `Agent.ts` (`amend`, `attachToHistory`, `reportAmended`, `mergeMessages`, the `onMessage` contract), `AgentController.ts` (`persistRun`, the single call site at `stream()`; `attach()` does not persist again), `MemoryAgentStore.ts`, `providers/request.ts`, and all four test files in full.
- Ran the three new tests against the base versions of the files they cover, in a scratch copy of the package (not the worktree): the store test fails with `['1','2','3','2','4']`, the controller test fails with `expected 3 to be 4` (four ids, three distinct), and the request test fails with both `function_call`s kept. None of them passes without its fix.
- `bun --bun vitest run ai`: 19 files, 457 passed, 20 skipped. `bunx tsc --noEmit -p tsconfig.json`: clean. `bun run test:types`: 8 files, 84 passed, no type errors.
- `oxfmt --check` on every changed file: clean at both head and base, so the diff carries no formatting drift.
- No callers of `AgentStore.appendMessages` outside the controller and the tests; `example.ts` and `index.ts` only construct/export the store. The interface change is doc-only, so no custom store breaks. `useChat`/the reducer are untouched and do not depend on the store.
- The in-place replacement keeps the amended message at its original index, which is what `reconcileToolPairs`' positional check needs (the output must follow its call). Replacing the object rather than mutating it means the copy handed out by `loadThread` on an earlier turn is unaffected. A batch that carries the same id twice collapses to the last one, which matches `mergeMessages` in `Agent.ts`.
- The `reconcileToolPairs` guard keeps the first `function_call` and drops later ones under the same id; in the append-only shape the two calls are identical, so nothing is lost, and the existing output guard already dropped the second output. Partial calls never reach it.
- `process.env.SECRET ??=` sits above the imports but ESM hoists them; `signing.ts` reads `process.env.SECRET` at call time (line 91), so this works the same way it does in `Agent.test.ts`. The value is a test placeholder, not a secret.

**Findings**

F1 [should-fix] packages/gemi/ai/AgentController.ts:43 — "A table keyed by message id gets this for free" overstates what a table gives you.

A table with a primary key on message id does not upsert on its own; a plain `INSERT` of the amended message throws on the key, which is exactly the failure the issue describes for real stores. What the table gives for free is the *lookup*; the write still has to be insert-or-update (`ON CONFLICT (id) DO UPDATE`, or a merge). Since this comment is the contract an app author reads when writing a store, suggest wording it so it cannot be read as "insert is enough", e.g. "A table keyed by message id has the lookup already and needs an insert-or-update rather than a plain insert; a store that is a list has to look before it pushes."

F2 [nit] packages/gemi/ai/AgentController.test.ts:94 — `scriptedProvider` is a second copy of `FakeProvider` from `Agent.test.ts`.

Same capabilities block, same `upload`, same `normalizeError`, same per-call script. The PR body explains why the controller test needs a real `Agent`, and that reasoning holds; the duplication is the only cost. A shared fixture (there is already `providers/__fixtures__`) would let both files use one. Optional; fine to leave for a later pass.

**Verdict**: the fix is correct, minimal, and does what #449 asks; F1 is a one-line doc precision on the new contract and F2 is optional. Good to ship once F1 is reworded.

nstfkc:
Review findings addressed in b6e25b20.

- **F1** (should-fix, `AgentController.ts` contract comment) — fixed. The comment now says a table keyed by message id has the lookup already but still needs an insert-or-update rather than a plain insert, which would fail on the key; a list-backed store has to look before it pushes.
- **F2** (nit, duplicated fake provider) — fixed. `FakeProvider` / `fakeProvider` moved out of `Agent.test.ts` into `ai/providers/fakeProvider.ts` (in source rather than a test file, for the reason `store/stubAgentRun.ts` gives: a test file importing another test file gets its suites collected twice). `AgentController.test.ts` drops its `scriptedProvider` copy and imports the shared one; the reason that test runs a real `Agent` moved onto the test's own doc comment.

Verified from the worktree's `packages/gemi`: `bun --bun vitest run ai` (19 files, 457 passed, 20 skipped), `bunx tsc --noEmit -p tsconfig.json` (clean), `bun run test:types` (84 passed, no type errors).

PR #482, opened 2026-09-03
A thread the store does not have is a 404, not an empty conversation

### LINKED ISSUE #465: ai: an unknown or expired threadId silently starts a fresh conversation

`MemoryAgentStore.loadThread` returns `[]` for a thread it does not have, and `appendMessages` recreates it. So a thread that expired (24 h TTL), a mistyped id, or a thread from another instance after a scale-in, all look like an empty conversation: the user's history is gone with no signal, and the new turn is persisted under the old id.

**Fix**: `loadThread` returns `null` for unknown; the controller answers 404 `thread_not_found` before running anything. `useChat` surfaces it as `error` and an app can fall back to stateless or create a new thread. Keep "append creates" only for the documented client-owned-id case, behind an explicit flag.

nstfkc:
Fixes #465

## The problem

`MemoryAgentStore.loadThread` answered `[]` for a thread it did not have, and `appendMessages` created one on the spot. Three unrelated ways of losing a thread — it expired past the 24 h TTL, the client mistyped the id, it lived on an instance that was scaled in — therefore all looked like a brand-new conversation: the history vanished with no signal, and the next turn was persisted under the old id as though it were the first.

## The fix

- **`AgentStore.loadThread` returns `null` for a thread the store does not have.** The interface's doc comment now spells out that `[]` and `null` are different answers: an empty array is a conversation with nothing in it yet and a turn on it runs; `null` is a 404. `MemoryAgentStore` is the only implementation in the tree.
- **`AgentController.stream` answers 404 `thread_not_found`** as soon as the store says `null` — before `instructions()`, before `agent.stream`, before `liveRuns.register`. Nothing is charged for and there is nothing for a client to attach to.
- **`MemoryAgentStore.appendMessages` refuses an unknown thread** (throws, naming the id) rather than minting it. The controller never reaches this because it checked first; a store-level caller that does gets told.
- **"Append creates" survives only behind `new MemoryAgentStore({ clientOwnedIds: true })`.** In that mode `loadThread` answers `[]` for an unknown id and `appendMessages` creates it — the setup the old comment called "stateless-with-a-thread-id", where the client mints the id and an unknown one really is a conversation starting. `createThread` is unchanged.
- **`useChat` carries the code through.** `httpError` reads `error.code` off the JSON body and passes `thread_not_found` through instead of folding it into `unknown`; `"thread_not_found"` is added to `AgentErrorCode`. The app sees `status === "error"` with `error.code === "thread_not_found"`, `retryable: false`, and the turn that failed still in `messages`, so it can drop the id and remount on a new thread or stateless.

## Decisions

- **Constructor option, not a per-call flag.** The store is what knows where its ids come from; the controller does not, and threading a flag through `appendMessages` would have meant the controller guessing. `clientOwnedIds` also has to affect `loadThread` (unknown must be `[]` there or the first turn 404s), which a per-append flag could not express.
- **`attach` does not consult the store.** It answers whether a run is in flight in this process, and any run it finds was started on a thread `stream` had already loaded. A miss is `no_live_run` whether or not the thread exists: a client behind a round-robin balancer hits that routinely, `useChat` treats it as "nothing to resume" and fires `onAttachMiss`, whose documented remedy is to re-read the thread — which is where such an app learns the thread is gone. Adding a store read to every attach would duplicate that read for no new information. A comment in `attach` records this.
- **`stop` does not consult the store.** `threadId` there is one of three handles for finding a run; a missing thread means no run, and `{ stopped: false }` is already that answer. **`upload`** never sees a `threadId`.
- **Only `thread_not_found` is passed through `httpError`,** not every server code. `store/sse.ts` notes that `AgentErrorCode` names what the client can act on rather than what the server can say; a lost thread is the one HTTP failure that calls for a different next request (new thread or stateless), so it earns a name. Auth and validation failures stay `unknown` with the server's message.
- **`example.ts` gains a `/support/threads` route** that mints a thread with `store.createThread` and returns its id, because with this change an app has to obtain a thread id from the server before its first threaded turn, and the example is where the module shows how.
- **Tests that hand-pick thread ids** (`"t1"` throughout the attach and stop suites) now run on a `clientOwnedIds` store rather than being rewritten around `createThread`; a comment in `attachable()` says why. Tests about the store itself mint ids with `createThread`.

## Adjacent, not fixed here

- `useChat` keeps its `threadId` after a `thread_not_found`; a second `sendMessage` will 404 again until the app remounts without the id. That is deliberate — the app chooses between a new thread and stateless — but a `clearThread()` on the hook, or an `onThreadNotFound` handler, might be worth its own issue.
- `AgentStore` has no `deleteThread`; `MemoryAgentStore.delete` is a test seam. An app that wants to drop a conversation on request has no interface-level way to.

## Verification

From `packages/gemi` in the worktree:

- `bun --bun vitest run ai` — 19 files, 459 passed, 20 skipped

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -2,7 +2,8 @@ process.env.SECRET ??= "agent-test-secret";
 
 import { describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
-import type { AgentProvider, ProviderEvent, ProviderStreamParams } from "./AgentProvider";
+import type { AgentProvider, ProviderEvent } from "./AgentProvider";
+import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
 import type {
@@ -76,53 +77,6 @@ const usage = (inputTokens: number, outputTokens: number): Usage => ({
   outputTokens,
   totalTokens: inputTokens + outputTokens,
 });
-
-/**
- * Scripted `ProviderEvent`s, one script per model call.
- *
- * The real provider is written against the same interface elsewhere; depending
- * on it here would make these tests a test of two things at once, and would
- * need a network.
- */
-class FakeProvider {
-  readonly model = "fake";
-  readonly capabilities = {
-    reasoning: true,
-    structuredOutput: true,
-    fileInput: true,
-    parallelToolCalls: true,
-    toolSearch: true,
-  };
-  readonly calls: ProviderStreamParams[] = [];
-
-  constructor(private scripts: ProviderEvent[][]) {}
-
-  stream(params: ProviderStreamParams) {
-    this.calls.push(params);
-    const script = this.scripts[this.calls.length - 1] ?? [
-      { type: "finish", reason: "stop", usage: usage(0, 0) },
-    ];
-    return (async function* () {
-      for (const event of script) yield event;
-    })();
-  }
-
-  async upload() {
-    return "file_1";
-  }
-
-  normalizeError(error: unknown) {
-    return {
-      code: "provider_error" as const,
-      message: error instanceof Error ? error.message : String(error),
-      retryable: false,
-    };
-  }
-}
-
-function fakeProvider(...scripts: ProviderEvent[][]) {
-  return new FakeProvider(scripts) as unknown as AgentProvider & FakeProvider;
-}
 
 function deferred<T = void>() {
   let resolve!: (value: T) => void;
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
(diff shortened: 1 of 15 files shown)


nstfkc (commented):
Fresh review of #482 at 966ed37e (base `ai/14-attach-miss` @ 2ff5cbba).

**Checked**
- Re-derived the failure from #465 against the diff: without the fix `loadThread` answers `[]`, so the two new controller tests ("answers 404 for a thread the store does not have" and "a thread that expired is the same 404") would get a 200 with `calls.length === 1` and the store test would get `[]` where it now expects `null`; the `useChat` test would get `code: "unknown"`. All four fail without the change and pass with it. The expired case is exercised for real (`ttlMs: 10` + `sweep(now + 90s)`), not just the never-existed case.
- Ordering in `stream`: the store read is the first thing after body parsing; the 404 returns before `instructions()`, `agent.stream`, `liveRuns.register` and `persistRun`, so nothing is registered, charged or persisted. The test pins `instructed === false`, `calls.length === 0` and `liveRuns.find(...) === null`.
- `persistRun` -> `appendMessages`: the only in-tree caller of the now-throwing path. The throw lands in the existing `try/catch` -> `reportHookFailure`, so a thread deleted or expired mid-run reports rather than re-minting under the dead id, and the run still completes. `touchedAt` is bumped by `loadThread`, so with the 24h default a thread cannot expire under an in-flight run.
- `attach` and `stop` still do not read the store; a stale thread on attach is `no_live_run` -> `onAttachMiss`, unchanged. `/stop` and `/files` never emit `thread_not_found`, so the wider `httpError` only matters on `sendMessage`.
- `useChat` after the 404: `status === "error"`, `error.retryable === false` (404 is neither 429 nor >= 500), user message retained in `messages`; `threadId` retained (documented as deliberate in the PR).
- Callers of the changed contract: `AgentController.ts` is the only reader of `loadThread`/`appendMessages`; `MemoryAgentStore` the only implementation; no docs under the repo mention `loadThread`. `AgentErrorCode` has no runtime allow-list that needed the new member.
- `strict: false` in the package's base tsconfig means `AgentMessage[] | null` erases inside the package; the `.d.ts` still carries `| null` for consumers and the doc comment carries the contract. `if (!history)` also 404s on `undefined`, which is the safe direction for a third-party store.
- Minimality: 7 files, all in `packages/gemi/ai`, no reformatting, no TODOs, no secrets. Comments and messages match the module's explanatory voice.
- `bun --bun vitest run ai`: 19 files, 459 passed, 20 skipped. `bunx tsc --noEmit -p tsconfig.json`: clean. `bun run test:types`: 8 files, 84 passed, no type errors.

**Findings**

F1 [should-fix] packages/gemi/ai/useChat.tsx:47 — nothing tells an app where a `threadId` now has to come from.
With the default store an id that did not come out of `store.createThread` is a 404 on the first turn, and that is the new contract. But `UseChatParams.threadId` still reads "Continues a server-side thread", `example.ts` takes `threadId` as a prop from nowhere, and `ApiRouter.agent()` mounts no route that mints one — so the only way to learn that a client-picked id no longer works is the 404 message itself. The old "stateless-with-a-thread-id" comment that used to say the client may mint the id has been removed without a replacement saying who does. One sentence on `UseChatParams.threadId` (and on `AgentStore.createThread`) — the id has to come from `createThread` in an app route unless the store is `clientOwnedIds` — would close it; the example could show the route. Survivable, since the 404 message names the situation, but it is the one thing an existing app on the default store hits first.

F2 [nit] packages/gemi/ai/AgentController.ts:169 — the existence check now runs before the hook the comment above names for ownership checks.
The comment at 160–166 points an app at `instructions()` to check that `req.user` owns the thread; the 404 now fires before that, so a caller learns whether a uuid it holds is live on this instance without the app's check running. With unguessable ids this is not an exploitable oracle and the ordering is the right one (no charge before the check), but the comment at 171 could say so in one clause, so the next reader does not "fix" it by moving the load below `instructions()`.

**Verdict**: fixes #465 as stated, tests exercise the real failure, no regressions found; ship with F1 addressed in this PR or a follow-up.

nstfkc:
Both findings addressed in 01362b94.

- **F1** (should-fix, `useChat.tsx:47`) — fixed. `UseChatParams.threadId` now says the id is the store's: an app route calls `store.createThread` and hands the result to the hook, and an id the store never minted (or has since expired) is a `thread_not_found` on the first turn, unless the store is `clientOwnedIds`. `AgentStore.createThread` gained a doc comment saying the same from the store's side and why `ApiRouter.agent()` mounts no such route (minting is where the app records ownership, which a store holding no user cannot do for it). `example.ts` now shows that route: `"/support/threads": this.post(...)` behind `"auth"`, calling `supportStore.createThread({ userId })`, with a comment tying it to the `threadId` prop `Chat` takes.
- **F2** (nit, `AgentController.ts:169`) — fixed. The comment inside the `!history` branch now says the ordering is deliberate and what it costs: the ownership check that the comment above points at `instructions()` for has not run yet, so a caller with a uuid learns whether it is live here without the app's say; `/attach` already answers a question of that shape without calling `instructions()`, the id is unguessable, and moving the load below `instructions()` would spend that work on every dead id.

Verified from `packages/gemi` in the worktree: `bun --bun vitest run ai` (19 files, 459 passed, 20 skipped), `bunx tsc --noEmit -p tsconfig.json` (clean), `bun run test:types` (8 files, 84 passed, no type errors).

Issue #454, opened 2026-09-03
ai: two SSE encoders and two frame buffers — serve `stream` through LiveRuns too


nstfkc:
_Found in a review of the `ai` module at the tip of the stacked PRs (#425–#440, branch `ai/14-attach-miss`). Paths are under `packages/gemi/ai/`._

Today:
- `AgentController.stream` returns `run.toResponse()`: the run's own encoder, push-based inside `start()` with no `desiredSize` check (no backpressure), header `Cache-Control: no-cache`.
- `AgentController.attach` returns `sseResponse(liveRuns.replay(...))`: pull-based, header `no-store`.
- `AgentRunImpl.buffer` keeps every frame for the life of the run, and `MemoryLiveRuns` keeps a second window of pointers to the same frames (measured earlier: ~47 KB/run of duplication at the buffer level).

Two encoders drift (they already differ in headers) and any fix — keepalive, backpressure, frame-size trimming — has to be made twice.

**Fix**: register the run first (already the case), then have `stream` return `sseResponse(this.liveRuns.replay(run.runId))`. Then `AgentRunImpl.frames()` has exactly one consumer (the pump) and its buffer can go; `LiveRuns` owns the only window. One encoder, one buffer, one place for keepalive and backpressure. Delete `toResponse` or make it delegate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Bnj9jSx45LoYp5tiv1W8Nw

PR #483, opened 2026-09-03
A quiet run is not an idle connection

### LINKED ISSUE #452: ai: SSE stream has no keepalive — Azure's idle timeout kills long tool calls and awaiting-input

Neither SSE encoder (`AgentRunImpl.toResponse` or `store/sse.ts#sseResponse`) ever writes a comment line. A run emits nothing while a slow tool runs, while a sub-agent thinks, or while it sits in `awaiting-input` after `run-end` (the connection is closed then, but a long tool is the real case). Azure App Service's front end drops idle connections at ~230 s. The client sees the stream end with no `error` and cannot tell it from a normal close.

Both client parsers already skip `:` comment lines.

**Fix**: emit `: keepalive\n\n` every 20–30 s while no frame has been written (a timer reset on every enqueue). Belongs in the single shared encoder — see the "one SSE path" issue.

nstfkc:
## Problem

Neither SSE encoder — `AgentRunImpl.toResponse` in `ai/Agent.ts` or `sseResponse` in `ai/store/sse.ts` — ever wrote anything between frames. While a slow tool runs, a sub-agent thinks, or a model takes its time before its first token, the connection carries nothing, and a proxy that drops idle connections closes it (Azure App Service's front end does, at about 230 seconds). The client then sees the stream end with no `error` frame, which is indistinguishable from a run that finished.

## Fix

Both encoders write `: keepalive\n\n` every 5 seconds of silence. The timer is:

- armed when the stream is created, because the wait for the first frame is silence too;
- reset on every frame, so it measures silence rather than elapsed time;
- cleared when the stream closes, errors, or the reader cancels, so a finished response holds no timer.

The interval (`SSE_KEEPALIVE_INTERVAL_MS`) and the comment text (`SSE_KEEPALIVE`) are one constant in `ai/store/sse.ts`, and a small `sseKeepalive(controller)` helper next to them owns the arm/reset/clear logic. Both encoders import it, so the two paths cannot drift.

Both parsers on our side already skip comment lines: `ai/client/sse.ts` (`parse` drops any line starting with `:`) and `ai/providers/stream.ts` (`handleLine` does the same). Nothing on the receiving end changes.

The interval is 5 seconds because the nearest ceiling is gemi's own server, not a proxy: Bun's `idleTimeout` counts socket silence, a streaming body included, and `server/httpProd.ts` runs at the 10-second default unless `SERVER_IDLE_TIMEOUT` is set. Measured on Bun 1.3.14, a comment every 25 seconds loses the connection at 12 s (`ECONNRESET`); every 5 seconds keeps it open. Azure's 230 s is the far ceiling. A thousand quiet streams cost 200 thirteen-byte writes a second.

## Decisions

- **A shared helper rather than only a shared constant.** The issue asks for the interval to live in one place. Sharing only the number would still have left the reset-on-enqueue / clear-on-close logic written twice, and that logic is where the two paths would actually diverge. The helper is ~30 lines and does not merge the encoders — that is #454.
- **`setTimeout` re-armed after each fire, not `setInterval`.** "Reset on every enqueue" is a single `clearTimeout` + `setTimeout`; a re-armed timeout also makes "no timer outlives the stream" a one-line invariant (`timer === null` after `stop()`), which the tests check with `vi.getTimerCount()`.
- **The keepalive write is guarded.** A cancel can land between the timer firing and the enqueue, and a closed controller throws. The catch stops the timer and does nothing else: there is no reader to tell. `arm` also refuses after `stop()`, so a late `touch()` cannot leave a finished stream holding a timer.
- **No option on `sseResponse`/`toResponse` for the interval.** Nothing needs to vary it; the tests use the exported constant with fake timers.

## Adjacent, not fixed here

- A `Response` body that is created and never read or cancelled keeps its keepalive timer ticking and enqueues a 13-byte comment every 5 s into a queue nobody drains. That stream was already holding a parked frame generator before this change, so it is the same abandoned-response leak with one more thing attached, and a caller-side bug rather than something the encoder can detect. Noting it so the #454 encoder can decide whether to stop the timer when `desiredSize` drops.

## Verification

From `packages/gemi`:

- `bun --bun vitest run ai` — 20 files passed, 459 tests passed, 20 skipped. New: `ai/store/sse.test.ts` (3 tests) and a `toResponse keepalive` describe in `ai/Agent.test.ts` (2 tests), all with `vi.useFakeTimers()`: a keepalive goes out after `SSE_KEEPALIVE_INTERVAL_MS` of silence, a second one after continued silence, a frame pushes the next tick back, `vi.getTimerCount()` is 0 after both close and cancel, and a `touch()` after `stop()` arms nothing.
- `bunx tsc --noEmit -p tsconfig.json` — clean.
- `bun run test:types` — 8 files, 84 tests passed, no type errors.

Fixes #452

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
@@ -950,6 +951,91 @@ describe("a run outliving its request", () => {
     expect(body.startsWith("id: 2\ndata: {")).toBe(true);
     expect(body).toContain('"type":"text-delta"');
     expect(body.endsWith("\n\n")).toBe(true);
+  });
+});
+
+describe("toResponse keepalive", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  /**
+   * Reads until a read stays pending, and hands that pending read back.
+   *
+   * Boxed rather than returned bare: an async function that returns a promise
+   * adopts it, and the caller would then be waiting for the very chunk the
+   * test has not produced yet.
+   */
+  async function untilSilent(reader: ReadableStreamDefaultReader<Uint8Array>) {
+    while (true) {
+      let settled = false;
+      const next = reader.read().then((chunk) => {
+        settled = true;
+        return chunk;
+      });
+      await vi.advanceTimersByTimeAsync(0);
+      if (!settled) return { next };
+    }
+  }
+
+  test("a comment line goes out while a tool is running, and no timer outlives the stream", async () => {
+    vi.useFakeTimers();
+    const started = deferred();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => {
+        started.resolve();
+        return release.promise;
+      },
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const reader = run.toResponse().body!.getReader();
+    const bytes = new TextDecoder();
+    await started.promise;
+
+    // The tool is running and nothing has been written for a while.
+    const { next } = await untilSilent(reader);
+    expect(vi.getTimerCount()).toBe(1);
+    await vi.advanceTimersByTimeAsync(SSE_KEEPALIVE_INTERVAL_MS);
+    expect(bytes.decode((await next).value)).toBe(SSE_KEEPALIVE);
+
+    release.resolve({ ok: true });
+    let last = "";
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      last = bytes.decode(value);
+    }
+    expect(last).toContain('"type":"run-end"');
+    expect(vi.getTimerCount()).toBe(0);
+  });
+
+  test("a reader that cancels leaves no timer behind either", async () => {
+    vi.useFakeTimers();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => release.promise,
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const response = run.toResponse();
+    expect(vi.getTimerCount()).toBe(1);
+    await response.body!.cancel();
+    expect(vi.getTimerCount()).toBe(0);
+
+    release.resolve({ ok: true });
+    expect((await run.result()).finishReason).toBe("stop");
   });
 });
 
--- a/packages/gemi/ai/Agent.ts
+++ b/packages/gemi/ai/Agent.ts
@@ -14,6 +14,7 @@ import {
   verifyNestedRun,
   verifyPendingCall,
 } from "./signing";
+import { sseKeepalive } from "./store/sse";
 import type {
   AgentError,
   AgentMessage,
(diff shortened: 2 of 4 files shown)


nstfkc (commented):
Fresh review of #483 against `ai/14-attach-miss` @ 2ff5cbba. Read `ai/store/sse.ts`, the `toResponse` region of `ai/Agent.ts`, both new test files, the callers (`AgentController.ts:188,244`), the client parser (`ai/client/sse.ts`), `stubAgentRun.ts`, and `server/httpProd.ts` / `server/httpDev.ts`.

**F1 [blocking] packages/gemi/ai/store/sse.ts:26 — a 25 s keepalive never fires before gemi's own Bun server drops the idle stream**

`server/httpProd.ts:245` runs `Bun.serve({ idleTimeout: Number(process.env.SERVER_IDLE_TIMEOUT ?? 10) })`, and `httpDev.ts` passes no `idleTimeout`, so it takes Bun's default of 10 s. Bun's idle timeout counts silence on the socket, not request lifetime, and it applies to a streaming response body. Measured on Bun 1.3.14 with a `Bun.serve` returning a `ReadableStream` SSE body that writes `: keepalive\n\n` on an interval:

- `idleTimeout: 10`, write every 25 s: connection closed at **12.0 s** (in-process `fetch`: "The socket connection was closed unexpectedly"; separate-process `curl -N`: exit 18 at 12.0 s). 1 chunk delivered.
- `idleTimeout: 10`, write every 8 s: still open at 40 s. Write every 5 s: still open at 40 s.
- `idleTimeout: 6`, write every 30 s: closed at 8.0 s. `idleTimeout: 0`: stays open.

So a stream write does reset Bun's timer, and a keepalive under the idle timeout works — but 25 s is 13 s too late. On a gemi deployment (dev or prod with the default) a tool that runs longer than ~12 s still ends the stream with no `error` frame, exactly the failure #452 describes, before and after this PR. Azure's 230 s front-end limit only becomes the binding constraint once the origin stops closing the connection first. The fake-timer tests cannot see this, and `ai/live` would only if a tool were slower than 12 s.

Suggested fix, in one place: make `SSE_KEEPALIVE_INTERVAL_MS` comfortably below the server's idle timeout (5 s is under every value in play and 1,000 idle streams at 5 s is 200 fourteen-byte writes a second, which is nothing), and say in the constant's comment that Bun's own `idleTimeout` is the nearer ceiling, not Azure's. The alternative — `server.timeout(req, 0)` for SSE responses or a larger `SERVER_IDLE_TIMEOUT` — needs the request handler to know a response is a stream, and is a different PR.

**F2 [nit] packages/gemi/ai/store/sse.ts:71 — `touch()` after `stop()` re-arms a timer for one interval**

`arm` does not check `stopped`, so a `touch()` after `stop()` schedules a timer that fires once, sees `stopped`, and does nothing; the stream would hold a timer for one more interval. No current caller does this (`toResponse` checks `cancelled` before enqueue+touch; `sseResponse` only touches after a successful enqueue), but the helper's doc promises "`stop()` … keeps a finished stream from holding a timer", and a `if (stopped) return;` at the top of `arm` makes that true for any caller.

**Checked and fine**

- Both new test files fail against the base sources (verified in a throwaway worktree with the fix reverted and the constants kept: the `getTimerCount` assertions fail and the `sseResponse` silence test times out), so they do exercise the fix.
- `keepalive` is assigned synchronously inside `start` in both encoders before any await, so `cancel` cannot see it unset. `stop()` is idempotent, the guarded enqueue in the timer handles a cancel landing between fire and write, and no touch happens after a stop on any existing path.
- Both encoders stop on close, error (after the last `error` frame in `sseResponse`), and cancel. `stubAgentRun.toResponse` returns a null body and needs nothing.
- `ai/client/sse.ts` `parse` drops `:` lines and a block with no `data:` yields no frame, so a bare `: keepalive\n\n` produces nothing downstream; `cursor` is untouched.
- No other timers in `Agent.ts`'s execute path, so the fake-timer counts in `Agent.test.ts` are the keepalive's alone; `afterEach(vi.useRealTimers)` is scoped to the new describe.
- Diff is limited to the four files; comments are in the module's voice; no TODOs, no secrets.
- From `packages/gemi`: `bun --bun vitest run ai` — 20 files, 458 passed, 20 skipped. `bunx tsc --noEmit -p tsconfig.json` — clean.

**Verdict:** changes requested — F1 means the fix does not hold on gemi's own server with the default idle timeout; the interval constant is the one-line change.

nstfkc:
Both findings addressed in b8c00730b7ed65b1ce54c211057e277ccf3a2632.

- **F1** — fixed. Reproduced first on Bun 1.3.14 with a `Bun.serve` SSE body: `idleTimeout: 10` with a comment every 25 s ends in `ECONNRESET`; every 5 s stays open past 20 s (52 bytes read, so the comment is 13 bytes, not 14). `SSE_KEEPALIVE_INTERVAL_MS` is now `5_000`, and the constant's comment names Bun's `idleTimeout` (10 s default, `SERVER_IDLE_TIMEOUT` in `server/httpProd.ts`) as the ceiling that matters, with the measurement, and Azure's 230 s as the far one. `server.timeout(req, 0)` for streaming responses is left for a separate PR as you suggested. Tests use the exported constant, so they moved with it.
- **F2** — fixed. `arm` returns early when `stopped`, and a new test in `ai/store/sse.test.ts` calls `touch()` after `stop()` and asserts `vi.getTimerCount()` is 0 and nothing is enqueued after an interval.

PR body updated to say 5 s and 13 bytes. From `packages/gemi`: `bun --bun vitest run ai` 20 files, 459 passed, 20 skipped; `bunx tsc --noEmit -p tsconfig.json` clean; `bun run test:types` 84 passed, no type errors.

PR #484, opened 2026-09-03
A thread holds one run at a time, and a new turn ends the old one first

### LINKED ISSUE #450: ai: two runs on one thread race — sending while streaming does not stop the previous run

**Verified with a script.**

`useChat.send` aborts the in-flight `fetch` when the user sends again, but a disconnect no longer stops a run (by design). The old run keeps going server-side; the new turn's `loadThread` reads history that does not yet contain the old run's output; both runs `appendMessages` when they finish. Result:

```
user: second
assistant: answer#2 (saw 1 user msgs)
user: first
assistant: answer#1 (saw 1 user msgs)
```

Run 1 never saw turn 2, the thread is out of order, and `LiveRuns.byThread` now points at run 2 while run 1 is still live and unstoppable via `threadId`. Stateless mode is unaffected (the client history is authoritative).

**Fix**
- Server: in `AgentController.stream`, when `threadId` names a thread with a live run, either `stop()` it first (and wait for its transcript to persist) or answer 409 `run_in_progress`. A per-thread lock is the durable version of this.
- Client: `useChat.send` should post `/stop` for the superseded run (it already has `clientRunId`), not just abort the fetch.

nstfkc:
## The problem

`useChat.send` aborted the in-flight `fetch` when the user sent again, but a dropped connection no longer stops a run (by design, so `/attach` works). The old run kept going server-side and never saw the new turn; the new turn's `loadThread` read a history that did not yet contain the old run's answer; both runs `appendMessages` when they finished, in whichever order the model returned them. With the first run slower the thread read `user2, assistant2, user1, assistant1`, and `LiveRuns.byThread` pointed at run 2 while run 1 was still live and unstoppable by `threadId`.

## The fix

**Server.** `AgentController.stream` now runs its threaded setup — find the live run, stop it, wait for its transcript to be stored, load the thread, start and register the new run — under a per-thread lock (`withThread`). The wait is on the old run's `persistRun` promise (kept in a `WeakMap` keyed by run), not on `run.result()`: the result settles when the transcript is final, and `appendMessages` runs *after* that, so a `loadThread` in the gap would read a history the old answer is missing from — the original bug by a shorter route. The lock is what makes a third turn wait for the second rather than for the first; without it, two turns arriving together both see the same live run, both stop it, both wait for it, and both start.

**Client.** `useChat.send` posts `/stop` for the superseded run in addition to aborting the fetch. The posting code is shared with `stop()` (`postStop`) so a stop that fails is reported the same way. It names the run by `runId` and `clientRunId` only — never by `threadId`, because the post is not awaited and by the time it lands the thread's live run may be the one this send just started. A superseded run this client neither started nor saw start (an attach from a cursor past `run-start`) has no such handle and posts nothing; the server ends it when the new turn reaches the thread. Stateless mode has no server-side thread to hold, so the client post is what ends the old run there.

## Decisions

- **Stop-and-wait rather than 409 `run_in_progress`.** Sending again is what `useChat.send` means by a stop — the UI already shows the new turn and the aborted old answer. A 409 would need the client to post `/stop`, then poll until the run is really gone (`/stop` returns on abort, not on unwind), then post the turn; a worse client for no better thread. The cost of stopping is that the new turn waits for the old run to unwind, which can be as long as its slowest tool in flight — and that wait is exactly what puts `assistant1` before `user2`.
- **Module-level lock and persist map**, keyed by `threadId` / run, rather than fields on the controller (constructed per request, so nothing on it survives to the next request) or on `MemoryLiveRuns` (which owns frames, not the controller's protocol). Per process, like `LiveRuns`, and for the same reason: the run being guarded lives here. The lock entry is deleted when its last queued setup releases; the persist map is weak.
- **A finished run still within `ttlMs`** is `find`-able; stopping it is a no-op on `AgentRunImpl` and waiting on its promise is at most the append it may still be doing. That closes a smaller pre-existing race too: a turn sent right at `run-end` could previously load the thread before the append landed.
- **Only the superseding send sends `{ runId, clientRunId }`.** `stop()` itself still sends `threadId` as a third fallback and is unchanged.

## Tests

`AgentController.test.ts`, `one live run per thread`:
- the issue's interleaving — second turn while the first is live, first finishes last; asserts the second was held (no `agent.stream` call, no `loadThread` without the first answer), that the first was stopped, that the second run loaded `[u1, a1]`, and that the store reads `u1, a1, u2, a2`;
- three overlapping turns — the third stops the second, not the first, and the thread stays in order;
- a finished run within its ttl holds nothing up;
- stateless turns are not serialized.

`useChat.test.tsx`, `lifecycle`:
- a second send posts `/stop` with the first run's `runId` and `clientRunId`, and nothing else;
- an attached run with no handle posts no stop;
- the two existing supersede tests now route their fetch mocks by URL, since `/stop` sits between the two turn POSTs.

## Adjacent, not fixed here

- `stop()` in `useChat` sends `runId`, `threadId` *and* `clientRunId`, and the route resolves `runId` first. `runId` is cleared on `run-end` so it is not stale in practice, but the `threadId` fallback could name a newer run if a stop is ever posted late; the superseding send avoids that by construction, `stop()` does not.
- A hung provider that never settles `run.result()` after `stop()` would hold the next turn on that thread indefinitely. A timeout on the wait, falling back to a 409, would be the follow-up if that turns out to happen.

## Verification

From `packages/gemi`:
- `bun --bun vitest run ai` — 19 files, 460 passed, 20 skipped (6 new tests)

Fixes #450

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
@@ -950,6 +951,91 @@ describe("a run outliving its request", () => {
     expect(body.startsWith("id: 2\ndata: {")).toBe(true);
     expect(body).toContain('"type":"text-delta"');
     expect(body.endsWith("\n\n")).toBe(true);
+  });
+});
+
+describe("toResponse keepalive", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  /**
+   * Reads until a read stays pending, and hands that pending read back.
+   *
+   * Boxed rather than returned bare: an async function that returns a promise
+   * adopts it, and the caller would then be waiting for the very chunk the
+   * test has not produced yet.
+   */
+  async function untilSilent(reader: ReadableStreamDefaultReader<Uint8Array>) {
+    while (true) {
+      let settled = false;
+      const next = reader.read().then((chunk) => {
+        settled = true;
+        return chunk;
+      });
+      await vi.advanceTimersByTimeAsync(0);
+      if (!settled) return { next };
+    }
+  }
+
+  test("a comment line goes out while a tool is running, and no timer outlives the stream", async () => {
+    vi.useFakeTimers();
+    const started = deferred();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => {
+        started.resolve();
+        return release.promise;
+      },
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const reader = run.toResponse().body!.getReader();
+    const bytes = new TextDecoder();
+    await started.promise;
+
+    // The tool is running and nothing has been written for a while.
+    const { next } = await untilSilent(reader);
+    expect(vi.getTimerCount()).toBe(1);
+    await vi.advanceTimersByTimeAsync(SSE_KEEPALIVE_INTERVAL_MS);
+    expect(bytes.decode((await next).value)).toBe(SSE_KEEPALIVE);
+
+    release.resolve({ ok: true });
+    let last = "";
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      last = bytes.decode(value);
+    }
+    expect(last).toContain('"type":"run-end"');
+    expect(vi.getTimerCount()).toBe(0);
+  });
+
+  test("a reader that cancels leaves no timer behind either", async () => {
+    vi.useFakeTimers();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => release.promise,
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const response = run.toResponse();
+    expect(vi.getTimerCount()).toBe(1);
+    await response.body!.cancel();
+    expect(vi.getTimerCount()).toBe(0);
+
+    release.resolve({ ok: true });
+    expect((await run.result()).finishReason).toBe("stop");
   });
 });
 
--- a/packages/gemi/ai/Agent.ts
+++ b/packages/gemi/ai/Agent.ts
@@ -14,6 +14,7 @@ import {
   verifyNestedRun,
   verifyPendingCall,
 } from "./signing";
+import { sseKeepalive } from "./store/sse";
 import type {
   AgentError,
   AgentMessage,
@@ -1284,8 +1285,14 @@ class AgentRunImpl implements AgentRun<ToolShapes, unknown> {
     const encoder = new TextEncoder();
     let cancelled = false;
 
+    let keepalive!: ReturnType<typeof sseKeepalive>;
+
     const body = new ReadableStream<Uint8Array>({
       start: async (controller) => {
+        // A slow tool or a thinking sub-agent can leave the connection silent
+        // for longer than a proxy's idle timeout, and a proxy that closes it
+        // looks to the client exactly like a run that finished.
+        keepalive = sseKeepalive(controller);
         try {
           for await (const frame of frames) {
             if (cancelled) break;
@@ -1294,11 +1301,13 @@ class AgentRunImpl implements AgentRun<ToolShapes, unknown> {
             controller.enqueue(
               encoder.encode(`id: ${frame.seq}\ndata: ${JSON.stringify(frame.event)}\n\n`),
             );
+            keepalive.touch();
           }
         } catch {
           // A stream that cannot be written to is a dead reader, not a dead
           // run. Nothing to report and nothing to stop.
         }
+        keepalive.stop();
         try {
           controller.close();
         } catch {
@@ -1311,6 +1320,7 @@ class AgentRunImpl implements AgentRun<ToolShapes, unknown> {
         // tool loop is here and a closed tab has not stopped step four from
         // charging a card.
         cancelled = true;
+        keepalive.stop();
       },
     });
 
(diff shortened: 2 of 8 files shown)


nstfkc (commented):
Fresh review of #484 against issue #450. I re-derived the interleaving from the issue, read all four changed files in full plus `LiveRuns.ts`, `AgentRunImpl` (`stop`/`result`/`execute`/`finalizeAborted`), `MemoryAgentStore.appendMessages` and `stubAgentRun.ts`, and ran two scratch tests against the branch for the scenarios below.

**Does it fix the issue?** Yes. `withThread` stops the live run, waits on its `persistRun` promise (not `result()`), and only then loads the thread and registers the new run under the per-thread lock. The new `AgentController` tests genuinely exercise the failure: without the fix `first.stopped` stays `false`, `started` is `true` before the first transcript lands, and `seen[1].messages` would not contain `a1`. The `useChat` test likewise fails without the fix (`/stop` would be absent from the call list). The lock chain (`previous.then(() => held)`, release in `finally`, delete-if-tail) is correct and cannot wedge on a throwing `fn`; the `threadLocks` entry is cleaned up when its last waiter releases; `persisted` is weak. Stateless is untouched. No unrelated changes, no whole-file reformatting, no TODOs, no secrets; comments and commit message match the module's voice and carry the required trailers.

Three findings, none blocking.

---

**F1 [should-fix] packages/gemi/ai/AgentController.ts:261 — the wait covers the hook chain, not just the append, so a slow or hung `onMessage`/`onStreamComplete` holds the thread.**

`persisted.get(run)` is the whole `persistRun` promise. After `appendMessages` (the thing the comment at L107 and the `withThread` doc say is being waited for), `persistRun` goes on to `await this.safely(() => this.onMessage(...))` per message (L542) and `await this.safely(() => this.onStreamComplete(...))` (L545). Those are app hooks — `onMessage` is documented as "the intended persistence point", `onAwaitingInput`-style hooks are documented as network I/O — and `safely` catches throws but cannot catch a promise that never settles. `LiveRuns.pump` already fixed exactly this class of bug for retention ("a `fetch` with no timeout never rejects, it just never settles… retention is this class's job and belongs on its own clock"); this PR reintroduces it for the thread lock.

Verified with a scratch test: with `onStreamComplete` returning a never-settling promise, turn 1's transcript is in the store (`loadThread` → `[u1, a1]`) and turn 2 is still held 300 ms later (`agent.stream` never called). The lock is held across the await, so every later turn on that thread queues behind it for the life of the process, and each queued request is an open HTTP connection. Short of a hang, every superseding send now pays the full latency of every hook for every message of the old run, which the PR's "cost" paragraph does not mention.

Suggested change: split `persistRun` so the promise stored in `persisted` resolves right after `appendMessages` (or after `result()` when there is no thread), and let the hook chain continue on its own promise as before. That makes the wait exactly what the comments already claim it is.

---

**F2 [should-fix] packages/gemi/ai/AgentController.ts:255 / packages/gemi/ai/useChat.tsx:568 — a Stop pressed while the new turn is queued misses it; the turn then runs to completion unwatched and unstoppable from that client.**

Flow: user sends (run 1 live), sends again (client posts `/stop {runId: run_1, clientRunId: c1}` and `POST` turn 2; server enters `withThread`, stops run 1, and is now awaiting `persisted.get(run_1)`), then presses Stop. `useChat.stop()` has no `runId` for turn 2 yet (its `run-start` has not arrived) so it posts `{ threadId, clientRunId: c2 }`. On the server `findByClientRunId("c2")` is null — turn 2 is not registered until `start()` runs — so `stop` falls through to `find({ threadId })` (L383), resolves run 1, no-ops, and answers `{ stopped: true }`. When run 1 unwinds, `start()` runs turn 2 anyway. The client aborted its fetch, so `run-start` for run 2 never reaches it; `abortRef` is null and `state.runId` still says `run_1`, so a second press of Stop posts a stop for run 1 again. Run 2 bills, appends to the thread, and the client shows turn 2 as aborted while the store has a full answer for it.

Verified with a scratch test (turn 2 queued → `stop({threadId, clientRunId:"c2"})` → `{stopped:true}` → run 1 finishes → `seen.length === 2`, `run_2.stopped === false`).

Before this PR the window between the turn's arrival and its registration was one `loadThread`; the PR widens it to the old run's whole unwind ("as long as its slowest tool in flight") plus F1's hook chain, and does so for exactly the send-send-stop sequence the PR is about. Suggested change: record the queued turn's `clientRunId` before `await previous` (a module-level `pending: Map<clientRunId, { cancelled: boolean }>` next to `threadLocks`), have `stop` mark it cancelled when `findByClientRunId` misses, and have `start()` return a stopped/empty response instead of calling `agent.stream` when its entry was cancelled. The client side then needs no change: `stop()` already sends `clientRunId`.

---

**F3 [should-fix] packages/gemi/ai/useChat.tsx:568 — a failed unawaited `/stop` lands as an error on the *new* turn, after `send` has already cleared `error`.**

`postStop` calls `fail()` on a non-ok response or a network error, which is right for `stop()`: the user pressed Stop and the run may still be billing. For the superseding send the post is fire-and-forget and its response lands after `commit({... error: null})` at L593, so a 5xx/502 from `/stop` (a gateway hiccup under load is not exotic at 1,000 users) sets `state.error` and fires `onError` while the new answer is streaming, and once `send`'s `finally` puts `phase` back to `idle` the derived `status` is `"error"` for a turn that succeeded. In threaded mode this is a false alarm by construction — the server ends the old run when the new turn reaches the thread, which is the point of the PR — so the only mode in which the report is actionable is stateless.

Suggested change: pass a flag (or a second callback) so the superseding path logs/ignores a failed stop when `stateRef.current.threadId` is set, and reports it only in stateless mode; or at minimum do not commit `error` onto a state whose in-flight request is a different `clientRunId` than the one being stopped.

---

Verification on the branch (`packages/gemi`): `bun --bun vitest run ai` — 19 files, 460 passed, 20 skipped; `bunx tsc --noEmit -p tsconfig.json` — clean; `bun run test:types` — 8 files, 84 passed, no type errors.

**Verdict:** changes requested — the fix is correct for the issue as stated and the tests are real, but F1 and F2 are ordering/cancellation gaps the PR introduces or widens on the same code path, and should be closed before this goes out.

nstfkc:
All three findings addressed in 6d364086.

- **F1** — fixed (6d364086). `persistRun` now settles right after `appendMessages` (or after `result()` when there is no thread), and the hook chain (`onMessage` per message, `onStreamComplete`, and the `onError` on a failed `result()`) continues on its own promise via a new `notifyRun`. `withThread` therefore waits for exactly the append. Test: "a hook that never settles does not hold the thread" — `onStreamComplete` returns a never-settling promise, the second turn still starts on `[u1, a1]`.
- **F2** — fixed (6d364086). A module-level `pendingTurns: Map<clientRunId, { cancelled }>` records every turn from body parse until `register` (covers the `instructions()` and `loadThread` windows too). `/stop` checks it when `findByClientRunId` misses and, if it names a pending turn, marks it cancelled and answers `{ stopped: true }` without falling through to `threadId`. `start()` checks the flag after its last `await` and answers `409 { code: "stopped" }` instead of calling `agent.stream`; the entry is removed in `stream`'s `finally`. Test: "a stop that names a turn still waiting its place ends it before it starts" — send/send/stop by `{ threadId, clientRunId }`, old run finishes, queued turn gets 409, `agent.stream` was called once, and a third turn then starts normally.
- **F3** — fixed (6d364086). `postStop` now takes a `report` callback: `stop()` passes `fail` as before; the superseding `send` passes a no-op when `threadId` is set (the server ends the old run when the new turn reaches the thread, so the failed stop is not actionable) and `handlers.current.onError` only when stateless, never committing `state.error`. Tests: "a supersede stop that fails is not the new turn's error" (threaded: `/stop` 502, status `idle`, `error` null, `onError` not called) and "without a thread, a supersede stop that fails is the app's to hear, not the turn's" (stateless: `onError` once, status `idle`, `error` null).

Verification from the worktree's `packages/gemi`: `bun --bun vitest run ai` — 19 files, 464 passed | 20 skipped (4 tests added); `bunx tsc --noEmit -p tsconfig.json` — clean; `bun run test:types` — 84 passed, no type errors. oxfmt status of the four files is unchanged from base (only `useChat.test.tsx` flagged, and the added regions format clean).

Issue #468, opened 2026-09-03
ai: nested-run errors are only visible inside `nested-event` — the parent never reports them


nstfkc:
_Found in a review of the `ai` module at the tip of the stacked PRs (#425–#440, branch `ai/14-attach-miss`). Paths are under `packages/gemi/ai/`._

When a sub-run rejects an answer (`invalid_tool_result`), errors mid-stream, or is content-filtered, the `error` event is wrapped as `nested-event` and applied to the nested transcript. The parent stream ends with its own `run-end` and no `error`, `useChat.error` stays null, and `onError` never fires. Seen in the re-entry verification: "signature rejected? false" at the top level even though the sub-run rejected it.

**Fix**: in `runNested`'s forwarding loop, mirror nested `error` events with `code` in `{invalid_tool_result, content_filtered, provider_error}` as a top-level `error` carrying `toolCallId` (and `runId` of the sub-run), so the UI can show it. The reducer's `error` case clears `pending`, which is correct here too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Bnj9jSx45LoYp5tiv1W8Nw

PR #485, opened 2026-09-03
A re-entered tool runs on a signed record and a checked input

### LINKED ISSUE #441: ai: stateless re-entry executes a server tool with unvalidated, client-authored input

**Verified with a script.**

In stateless mode the whole history is client-carried, including `ToolCallPart.nested`. `Agent.ingestTurn` gates re-entry on `parkedBelow`, which reads that client-supplied `nested` record, and `reenter` then calls `executeTool(resolved, call, call.input, 0, { answers })` with `call.input` taken straight from the posted history — **no `inputSchema.safeParse`**, unlike `runTools`. The signature on the answer is only checked later, inside the sub-run, and its rejection is emitted as a nested `error` wrapped in `nested-event`, so the parent stream ends `stop` with no top-level error.

Repro (fake provider, `SECRET` set): post `messages` containing an assistant message with a fabricated `tool-call` for any server tool whose `execute` calls `ctx.runAgent` (no `requiresApproval` needed), `input: { path: 123, extra: "not-in-schema" }` against `s.object({ path: s.string() })`, and `nested: [{ finishReason: "awaiting-input", messages: [... an open "ask" call q1 ...] }]`; send `toolResults: [{ toolCallId: "q1", path: ["tc1"], signature: "garbage", output: {...} }]`.

Observed:
```
tool executed: [{"input":{"path":123,"extra":"not-in-schema"},"resumed":true}]
signature rejected? false   (parent stream)
run finished: stop | sub-agent took a model step on the fabricated transcript: true
```

Consequences: (1) a tool's typed input contract is void in stateless mode — anything reading `input.path` as a string is wrong; (2) side effects before the first `runAgent` run on attacker-chosen input; (3) the sub-agent takes a real, billed model step on an attacker-authored transcript.

Threaded mode is not affected (history comes from the store and was validated on the way in).

**Fix**
- `reenter`: `safeParse` the input before `executeTool`, exactly as `runTools` does; on failure attach an `invalid_tool_input` result and do not run.
- Stop trusting `parkedBelow` in stateless mode: either refuse re-entry when the history is client-carried, or sign the `NestedRun` record (HMAC over `signingRunId`, `toolCallId`, `runId`, `finishReason`, open call ids) the way pending calls are signed, and verify before re-entering.
- Surface a nested `invalid_tool_result` at the top level (see the separate issue on nested errors).

nstfkc:
## Problem

In stateless mode the whole history is client-carried, including `ToolCallPart.nested`. `Agent.ingestTurn` gates re-entry on `parkedBelow`, which reads that client-supplied record, and `reenter` then calls `executeTool` with `call.input` taken straight from the posted history — no `inputSchema.safeParse`, unlike `runTools`. The answer's own signature is only checked later, inside the sub-run, after the tool body has already run from the top. So a turn carrying a fabricated tool call with out-of-schema input and a forged "awaiting-input" record executed a server tool on attacker-chosen input, and the sub-agent took a billed model step on a transcript the client wrote (issue #441, verified with a script).

## Fix

1. **`reenter` validates the input** with the tool's `inputSchema.safeParse`, exactly as `runTools` does. On failure it attaches an `invalid_tool_input` result and does not execute. The parsed value replaces the input on the amended clone, as in `runTools`, so the transcript carries what the tool was given. This also covers the legitimate case of a schema that changed between the parking turn and the answering turn.

2. **A parked `NestedRun` is signed, and verified before re-entry.** `signing.ts` gains `signNestedRun` / `verifyNestedRun`, a second token kind (`agn1.…`) on the same primitives (`mac`, `payload`, `canonicalize`, one shared token reader), over: the root run's `signingRunId` (carried in the clear and covered by the MAC, like a pending call's), the full path of tool-call ids down to the call the record hangs off, the sub-run's `runId`, and the sorted set of the sub-run's open call ids. `runNested` mints it when the sub-run ends `awaiting-input`; `parkedBelow` refuses any record that has no signature, or whose signature does not verify against what the record now says. A record without a valid signature is not a parked run, so the turn is rejected with the existing top-level `invalid_tool_result` error and the call is closed as refused.

3. **The signature reaches a stateless client** by re-sending the `tool-call` frame with the record on it at the moment the escalation is raised. The client reducer already treats a re-sent `nested` on a `tool-call` frame as authoritative ("a server that persists the nested transcript and re-sends it is more authoritative than what this client accumulated"), so no client change is needed and `useChat.forWire` hands the field back untouched. The re-sent frame carries `resent: true` (a new optional field on the `tool-call` stream event, documented in `types.ts`), and `AgentController.dispatchEvent` skips it the way it skips `partial`, so `onToolCall` still fires once per call rather than again on every park.

4. **The signature covers the tool's input.** `NestedRunClaims` includes the canonicalized `call.input`; `runNested` signs over it and `parkedBelow` verifies it against the history's input before the schema parse, so a genuine record cannot be replayed with a different, schema-valid input.

## Decisions

- **Sign rather than refuse.** The task allowed falling back to refusing re-entry from client-carried history. That would have removed stateless nested re-entry entirely, which the rest of this stack built (the reducer's re-enterable carve-out, `forWire` keeping `nested`). Signing came out at about a hundred lines including comments, so it is the fix.
- **Only parked records are signed.** A finished record is replayed out of its transcript and executes nothing; forging one feeds the tool a made-up sub-agent answer, which a client-carried history already permits everywhere. Signing all records would have added nothing the gate uses.
- **The transcript content is not signed.** The claim the gate needs is "the server parked *here*, on *these* calls." A sub-run resumes from client-carried messages on the same terms the parent run does in stateless mode; the issue's proposed claim set is the same.
- **`finishReason` is not in the claims** (the issue listed it) because it is always `awaiting-input` for a signed record; the signature's presence is the claim, and the gate only consults `awaiting-input` records.
- **The verifying run reads the minting run's id out of the token** rather than the caller passing it, because the parent has no other source for it and no other use for it. Pending calls do it the other way round only because `resolveAnswer` wants the malformed case named separately before verifying.
- **The record's nonce is spent on re-entry.** `consumeNestedRun` shares the registry (and a factored `spend`) with `consumePendingCall` and is called from `ingestTurn` once per record per turn, before the tool body runs. The first version left the nonce unspent on the theory that the answer's nonce stops a replay; the review showed the tool body runs before that nonce is checked, so a replay of the same pre-result history with the same answer would execute the tool a second time. Now it is refused with `invalid_tool_result` ("already been re-entered on that record. Ask again."). The token keeps the five-part shape so one reader serves both kinds, and a token of one kind presented as the other is `malformed` before the MAC is compared.
- **`parkedBelow` reports why it refused.** It returns `{ ok: true, signature }` or `{ ok: false, reason: "unparked" | "unsigned" | "expired" }`, and the rejection message names the reason; missing, malformed and forged collapse into `unsigned`, as `resolveAnswer` collapses forged.
- **Transport by re-sent `tool-call` frame, not a new event.** The alternative — an optional `signature` on the `nested-event` wrapper, copied by `applyNested` — needs a reducer change and a reducer test, and would leave a client on the previous reducer unable to re-enter. The re-sent frame does cost re-sending the sub-run transcript once per park; it is the same order as the forwarded events themselves.
- **The `resolveAnswer` approve path was not given the same parse**: there the executed `call.input` is the value the approval's MAC covers, so a rewrite already fails verification.

## Out of scope / noticed

- Surfacing a *nested* `invalid_tool_result` at the top level (the issue's third bullet) is #468.
- `tool-call` frames are emitted holding the part by reference, so a reader that serialises late can see `nested` on the first frame too. Pre-existing; harmless, since the re-sent frame carries the authoritative copy and comes later.
- After a successful resume the client's copy of the record keeps the stale `signature` next to `finishReason: "stop"` (nested events update the record in place). Harmless: only `awaiting-input` records are ever verified, and a re-park re-sends a fresh one.

## Verification

From `packages/gemi`:

- `bun --bun vitest run ai` — 19 files, **464 passed**, 20 skipped (was 454 passed at the base; 10 new tests)

New tests: `signing.test.ts` "a signed parked-run record" (6), `Agent.test.ts` "a client-carried history that says a tool parked" (4: the issue's forged script; the record re-sent on the stream with its signature; a genuine record widened to an unasked question; a genuine record with a rewritten input), plus one assertion that a finished record is unsigned. The changed files were checked with oxfmt on copies; they introduce no hunks beyond those already present at HEAD.

Fixes #441

### DIFF
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -2,7 +2,8 @@ process.env.SECRET ??= "agent-test-secret";
 
 import { describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
-import type { AgentProvider, ProviderEvent, ProviderStreamParams } from "./AgentProvider";
+import type { AgentProvider, ProviderEvent } from "./AgentProvider";
+import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
 import type {
@@ -76,53 +77,6 @@ const usage = (inputTokens: number, outputTokens: number): Usage => ({
   outputTokens,
   totalTokens: inputTokens + outputTokens,
 });
-
-/**
- * Scripted `ProviderEvent`s, one script per model call.
- *
- * The real provider is written against the same interface elsewhere; depending
- * on it here would make these tests a test of two things at once, and would
- * need a network.
- */
-class FakeProvider {
-  readonly model = "fake";
-  readonly capabilities = {
-    reasoning: true,
-    structuredOutput: true,
-    fileInput: true,
-    parallelToolCalls: true,
-    toolSearch: true,
-  };
-  readonly calls: ProviderStreamParams[] = [];
-
-  constructor(private scripts: ProviderEvent[][]) {}
-
-  stream(params: ProviderStreamParams) {
-    this.calls.push(params);
-    const script = this.scripts[this.calls.length - 1] ?? [
-      { type: "finish", reason: "stop", usage: usage(0, 0) },
-    ];
-    return (async function* () {
-      for (const event of script) yield event;
-    })();
-  }
-
-  async upload() {
-    return "file_1";
-  }
-
-  normalizeError(error: unknown) {
-    return {
-      code: "provider_error" as const,
-      message: error instanceof Error ? error.message : String(error),
-      retryable: false,
-    };
-  }
-}
-
-function fakeProvider(...scripts: ProviderEvent[][]) {
-  return new FakeProvider(scripts) as unknown as AgentProvider & FakeProvider;
-}
 
 function deferred<T = void>() {
   let resolve!: (value: T) => void;
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1,11 +1,12 @@
 process.env.SECRET ??= "agent-test-secret";
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 import { Agent, AgentTool, Skill, ToolNamespace } from "./Agent";
 import type { AgentProvider, ProviderEvent } from "./AgentProvider";
 import { fakeProvider } from "./providers/fakeProvider";
 import type { Schema } from "./Schema";
 import { readSignature, verifyPendingCall } from "./signing";
+import { SSE_KEEPALIVE, SSE_KEEPALIVE_INTERVAL_MS } from "./store/sse";
 import type {
   AgentMessage,
   AgentStreamEvent,
@@ -950,6 +951,91 @@ describe("a run outliving its request", () => {
     expect(body.startsWith("id: 2\ndata: {")).toBe(true);
     expect(body).toContain('"type":"text-delta"');
     expect(body.endsWith("\n\n")).toBe(true);
+  });
+});
+
+describe("toResponse keepalive", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  /**
+   * Reads until a read stays pending, and hands that pending read back.
+   *
+   * Boxed rather than returned bare: an async function that returns a promise
+   * adopts it, and the caller would then be waiting for the very chunk the
+   * test has not produced yet.
+   */
+  async function untilSilent(reader: ReadableStreamDefaultReader<Uint8Array>) {
+    while (true) {
+      let settled = false;
+      const next = reader.read().then((chunk) => {
+        settled = true;
+        return chunk;
+      });
+      await vi.advanceTimersByTimeAsync(0);
+      if (!settled) return { next };
+    }
+  }
+
+  test("a comment line goes out while a tool is running, and no timer outlives the stream", async () => {
+    vi.useFakeTimers();
+    const started = deferred();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => {
+        started.resolve();
+        return release.promise;
+      },
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const reader = run.toResponse().body!.getReader();
+    const bytes = new TextDecoder();
+    await started.promise;
+
+    // The tool is running and nothing has been written for a while.
+    const { next } = await untilSilent(reader);
+    expect(vi.getTimerCount()).toBe(1);
+    await vi.advanceTimersByTimeAsync(SSE_KEEPALIVE_INTERVAL_MS);
+    expect(bytes.decode((await next).value)).toBe(SSE_KEEPALIVE);
+
+    release.resolve({ ok: true });
+    let last = "";
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      last = bytes.decode(value);
+    }
+    expect(last).toContain('"type":"run-end"');
+    expect(vi.getTimerCount()).toBe(0);
+  });
+
+  test("a reader that cancels leaves no timer behind either", async () => {
+    vi.useFakeTimers();
+    const release = deferred<any>();
+    const slow = AgentTool.create({
+      name: "slow",
+      description: "Finishes eventually",
+      inputSchema: anything(),
+      execute: () => release.promise,
+    });
+    const provider = fakeProvider([toolCall("c1", "slow", {}), finish()], [finish()]);
+    const agent = Agent.create({ name: "patient", provider, tools: [slow] });
+    const run = agent.stream({ messages: [], req });
+
+    const response = run.toResponse();
+    expect(vi.getTimerCount()).toBe(1);
+    await response.body!.cancel();
+    expect(vi.getTimerCount()).toBe(0);
+
+    release.resolve({ ok: true });
+    expect((await run.result()).finishReason).toBe("stop");
   });
 });
 
--- a/packages/gemi/ai/Agent.test.ts
+++ b/packages/gemi/ai/Agent.test.ts
@@ -1227,6 +1227,8 @@ describe("a sub-agent that runs to completion", () => {
       "user",
       "assistant",
     ]);
+    // Unsigned: nothing is executed on the strength of a finished record.
+    expect("signature" in part.nested[0]).toBe(false);
     expect(partsOf(result.messages, "tool-result")[0]).toMatchObject({
       status: "ok",
       output: { heard: "eleven", from: "researcher" },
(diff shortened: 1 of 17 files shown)


nstfkc (commented):
Review of 7990d99d (`ai/15-reenter-validate` on `ai/14-attach-miss`). Read every changed file in full plus the reducer, `useChat.forWire`, `AgentController.dispatchEvent`, the stores and the controller's body intake. Ran the ai suite, tsc and the type tests; reproduced every finding below with a scratch test in a separate worktree at the PR head.

**F1 [should-fix] packages/gemi/ai/Agent.ts:2051 — the record's signature does not cover the tool's input, so a genuine parked record still runs the tool body on any schema-valid input the client chooses.**
`NestedRunClaims` binds path, sub-run id and open call ids, but not `call.input`. In stateless mode a client that holds a real parked record and a real answer can rewrite the call's `input` to any value the schema accepts, and `reenter` runs the body from the top on it. Reproduced: turn one parked `readFile({ path: "notes.md" })`; on turn two the posted history had `input: { path: "/etc/secrets.md" }` and the sub-run's seed user message rewritten to match (the transcript is unsigned by design, so `replayMismatch` is not a control here). Result: `bodies[1].input.path === "/etc/secrets.md"`, the sub-agent resumed and took a model step, the tool result was `status: "ok"`, and there was no error at any level. The issue's consequence (2) — "side effects before the first `runAgent` run on attacker-chosen input" — is narrowed to schema-valid input, not closed. The PR body's own argument for skipping the parse on the approve path ("the executed `call.input` is the value the approval's MAC covers") is the standard this path does not meet. Suggested change: add `canonicalize(call.input)` to the claims in `signNestedRun` (the part already carries the parsed value, as `runTools` writes it) and verify in `parkedBelow` against `entry.call.input` before parsing — the same "the history carries the value the signature covers" bargain the pending-call token makes. One more claim field and one more test.

**F2 [should-fix] packages/gemi/ai/Agent.ts:2104 (and packages/gemi/ai/AgentController.ts:396) — the re-sent `tool-call` frame makes `AgentController.onToolCall` fire twice for every call that parks.**
`dispatchEvent` calls `onToolCall` on every `tool-call` frame whose part is not `partial`. Before this PR exactly one such frame existed per call (the streaming ones go out as `tool-call-delta` with `partial: true`). The re-send is a second non-partial frame with the same `toolCallId` and input, so a hook that meters, audits or notifies per tool call now sees each parking call twice, and again on every re-park in a later turn — where it also fires for a call made by a previous run. The PR's new test asserts the two frames but nothing exercises the hook count (the controller test at `AgentController.test.ts:192` checks that it fires, not how often). Suggested change: dedupe in `dispatchEvent` by `toolCallId` per run, or mark the re-sent frame so the hook can tell it apart; alternatively carry the signature on the `nested-event` wrapper as the PR considered, which has no hook cost.

**F3 [should-fix] packages/gemi/ai/signing.ts:314 — "a second re-entry on the same record is stopped by the nonce on the *answer* it is delivering" is not what happens.**
The tool body runs before the sub-run ever looks at the answer's nonce. Reproduced: after a successful turn two, posting the same pre-result history with the same (now spent) answer ran the body a third time, the sub-agent took a third billed model step (it was resumed, refused the spent answer as a nested `invalid_tool_result`, and continued), and the parent ended `stop` with the tool result `status: "ok"` and no top-level error. The re-execution and the billed step are pre-existing (they were worse before this PR), but the "Nonce is not spent" decision in the PR body rests on this sentence and the comment now states a guarantee the code does not provide. At minimum the comment should say what the answer nonce actually stops (the answer being applied, not the re-entry). Spending the record nonce in `parkedBelow` would be consistent with the design: every park mints a fresh record, so a legitimate re-park is unaffected and a replay is refused before the body runs.

**F4 [nit] packages/gemi/ai/Agent.ts:2316 — `parkedBelow` discards the verify reason.**
`malformed`, `forged` and `expired` all surface as "has no sub-agent run waiting on that question". `signing.ts` says collapsing those "loses the only signal that says someone is probing", and the expired case is now a normal one for a threaded conversation: a sub-agent's question answered after the 24h TTL used to fail inside the sub-run as "has expired. Ask again." and is now rejected at the top level with a message that says the run never parked. Worth surfacing at least the expired case.

**Checked and found sound**
- The new tests exercise the fix: with `parkedBelow` reverted to the base version (parse kept), "a forged record" and "cannot be widened" fail; with the `reenter` parse removed, "a rewritten input is not executed" fails.
- Stateless resumption end to end through the real reducer: two questions in a row from one sub-agent (the second park re-signs, the client's rebuilt record verifies on turn three) and a grandchild's question two levels down, both from a JSON round trip of `state.messages`. The re-sent frame lands on the parked message from a previous run because `withMessage` has no finished-message guard, and `forWire` keeps `signature`.
- `readJsonBody` passes `messages` through untouched, so the signature survives the controller in stateless mode; `MemoryAgentStore` keeps references, so it survives threaded mode.
- Token kinds are separated by tag before the MAC; `open` is compared as a sorted set; the path is the server's own (`pathPrefix` + call id), not the client's.
- `bun --bun vitest run ai`: 19 files, 464 passed, 20 skipped. `bunx tsc --noEmit -p tsconfig.json`: clean. `bun run test:types`: 8 files, 84 passed, no type errors.
- Formatting: oxfmt on copies of the five changed files produces the same hunk counts at HEAD as at the base (14/0/0/15/1), so no new formatting drift. No TODOs, no secrets, comments in the module's voice.

**Verdict:** changes requested — F1 leaves the issue's stated consequence (2) open for schema-valid input, and F2 is a behaviour change for every `onToolCall` consumer.

nstfkc:
All four findings addressed in 18da0780.

- **F1 — fixed (18da0780).** `NestedRunClaims` gains `input`, canonicalized into the MAC the same way a pending call's is; `runNested` signs over `call.input` (the parsed value on every path that reaches it) and `parkedBelow` verifies against the history's `call.input` before anything else looks at it. The existing "rewritten input" Agent test now runs your script — `{ path: "/etc/secrets.md" }` with the sub-run's seed rewritten to match — and asserts the body does not run, the sub provider is not called again, the top-level error says the server did not sign the record, and `c1` closes as refused. The schema parse in `reenter` stays for the case it actually covers now, a schema that changed between turns, with its own test (a `readFile` whose field was renamed → `invalid_tool_input`, body not run). Plus a `signing.test.ts` case for the rewritten input and for key order surviving a client's serializer.
- **F2 — fixed (18da0780).** The re-sent frame is marked: `AgentStreamEvent`'s `tool-call` member gains `resent?: true`, `runNested` sets it, and `AgentController.dispatchEvent` skips it alongside `partial`. Marking beat dedupe-by-toolCallId because dedupe would still fire on a re-park for a call a previous run made, and beat the `nested-event` wrapper route for the reason in the PR body (reducer change, older clients cannot re-enter). The reducer merges the frame exactly as before. Tests: the Agent stream test asserts `resent` is absent on the model's frame and `true` on the record's; the controller test emits a `resent` frame and asserts `onToolCall` still fired once.
- **F3 — fixed (18da0780).** You are right that the sentence promised what the code did not do, and I have reversed decision 5 from the PR body rather than reworded it: the record's nonce is spent. `consumeNestedRun` shares the registry and `spend` with `consumePendingCall`; `ingestTurn` spends it once per record per turn, before the group for that host is created, so a sub-run that asked two things at once is re-entered once and a replay is refused with `invalid_tool_result` ("already been re-entered on that record. Ask again.") before the body runs. Test: your reproduction — a successful turn two, then the same pre-result history with the same answer — now leaves `bodies` at 2 and the sub provider at 2 calls, with `c1` closed as refused. The `signNestedRun` comment now says what the nonce is for.
- **F4 — fixed (18da0780).** `parkedBelow` returns `{ ok: true, signature } | { ok: false, reason: "unparked" | "unsigned" | "expired" }` and `ingestTurn` says which: "has no sub-agent run waiting on that question", "carries a record of a parked sub-agent run the server did not sign", or "parked a sub-agent run that has since expired. Ask again." Missing, malformed and forged collapse into the middle one, as `resolveAnswer` collapses forged into "does not match". Test: park, then answer with `Date.now` 25h later → the top-level error for `s1` says expired and the body did not run.

Verification from the worktree's `packages/gemi`: `bun --bun vitest run ai` → 19 files, 469 passed / 20 skipped (was 464 at the review head); `bunx tsc --noEmit -p tsconfig.json` → clean; `bun run test:types` → 8 files, 84 passed, no type errors. oxfmt on copies of the seven touched files produces the same hunk count as on HEAD's copies for each, so no formatting change rides along.

PR #486, opened 2026-09-03
React Compiler via oxc, plus three CLI commands and the agent skill

nstfkc:
Two independent pieces of work that meet at one line of the skill. Happy to split them if you'd rather review separately.

## React Compiler through oxc

`@vitejs/plugin-react` 6.1 runs the React Compiler through `oxc-transform-react` — the Rust port — instead of Babel, so auto-memoization costs a native transform rather than a second parse per module. The template now passes `react({ compiler: true })`.

gemi already delegates the React plugin to the app's `gemi.config.ts`, so **nothing in `packages/gemi` needed to change** for this.

Verified on a real `gemi build`:

| | `compiler-runtime` references |
|---|---|
| `dist/client` | 80 |
| `dist/server` | 0 |

That split is the point: the plugin only memoizes where `consumer !== "server"`, and gemi's SSR view build is a single pass, so memo caches there would be pure allocation. In dev, `GET /app/views/Home.tsx` comes back with both `const _c = …compilerRuntime["c"]` and five `$RefreshReg$` registrations — memoization and Fast Refresh coexist.

⚠️ `oxc-transform-react` is pinned to `^0.145.0` because that is the optional-peer range `@vitejs/plugin-react@6.1.1` declares, and `^0.145.0` on a `0.x` excludes the current npm latest (`0.148.0`). This needs a bump when upstream widens the range.

Upstream calls the integration experimental. It is build-time only — reverting is deleting `{ compiler: true }`.

## `gemi dev` — update notice

```
Starting dev server...

  Update available: gemi 0.58.0 → 0.59.0
  Run `gemi upgrade` to update.
  This crosses a minor version — read UPGRADE.md before you do:
  https://github.com/nstfkc/gemi/blob/main/UPGRADE.md
```

- **It cannot delay or fail startup.** Started *after* the server process is spawned, 2s timeout, every failure swallowed. Verified against a dead port: dev starts normally, no notice, no error.
- **It compares against your own channel.** An app on `0.60.0-rc.1` asks for the `rc` dist-tag. Resolving `latest` from an rc would tell users to "update" backwards onto the older stable release.
- Off via `GEMI_NO_UPDATE_CHECK=1`, and automatically when `CI` is set. `GEMI_REGISTRY` points it at a mirror.

The semver comparison is hand-written — gemi has no `semver` dependency and adding one to compare two strings would be the CLI's only startup dependency. It implements precedence only. The two cases that justify it: `0.60.0-rc.2 < 0.60.0-rc.10` (string order puts `.10` first, hiding a real release) and a prerelease sorting below its release.

## `gemi upgrade`

```bash
gemi upgrade            # newest on your channel
gemi upgrade --dry-run
gemi upgrade 0.58.0     # pin or roll back — no network needed
gemi upgrade rc
```

It resolves the version and hands the install to your package manager rather than editing `package.json` itself; a manifest written while the lockfile goes stale is a dependency graph nothing can reproduce. Lockfile detection picks bun/pnpm/yarn/npm (bun by default and on ties). A gemi declared in `devDependencies` stays there — `bun add` would otherwise silently promote it, which is an unexplained manifest diff in review.

## `gemi install-skill` + the shipped skill

Copies the `gemi-react-best-practices` skill (44 rules) that ships with the installed gemi into `.agents/skills/`. The skill documents the framework, so its correct version *is* the framework's version — which is why this is a command and not a folder you copy once. `gemi upgrade` re-runs it, but only where it is already installed; an upgrade is not the moment to create a directory nobody asked for.

Two behaviours worth reviewing closely:

- **It will not discard local edits.** `.agents/.skill-lock.json` stores a content hash; a mismatch refuses and names `--force`. This matters *because* `upgrade` runs it unattended. File paths are part of the digest, so a rule rename registers as a change — a hash over concatenated contents would call it identical.
- **It replaces, never merges.** A rule retracted upstream disappears locally, rather than being read by an agent as current guidance.

The lock follows the `vercel-labs/skills` shape and preserves unrelated entries and top-level keys, so gemi's skill sits beside skills installed from elsewhere.

The skill is authored markdown with no build output, so `build-publish.ts` stages it explicitly — the staging dir is built from explicit copies, and a `files` entry alone would have put it in no tarball at all. Confirmed present in an actual `npm pack --dry-run`.

**Genericized before shipping.** The skill was written against a specific app; every `@folio/*`, `folio-ai-concept` path and "in-house" reference is gone (31 of 46 rule files touched). `testing-new-tests-are-bun-tests` was a Vitest→bun migration note and is now `testing-match-the-suite`, keeping the two genuinely gemi facts in it.

**The rule these two halves meet at:** ground rule 5 asserted "this app runs no React Compiler". That is now a per-app `gemi.config.ts` decision with both answers live in the wild, so the rule says to check the config rather than asserting either.

## Checks

- Full suite: **166 files, 4016 passed, 1 skipped, 0 failed** (212 of those are new, across `bin/`).
- `typecheck` clean; `oxlint` 0 warnings on every file touched.
- Packaging suite passes; `npm pack --dry-run` confirms `skills/**` in the tarball.
- `oxfmt` clean on all new files. `bin/gemi.ts` and `scripts/build-publish.ts` were left unformatted **on purpose** — both already fail oxfmt on `main`, and reformatting would bury the real change.
- Smoke-tested end to end against the real npm registry, a stub returning `99.0.0`, and an unreachable one.

## Note

`.claude/skills/gemi-react-best-practices` is a symlink into `packages/gemi/skills/` so there is one copy. It reads correctly through the link, but I could not verify that Claude Code resolves a symlinked skill directory — skills load at session start. If it does not appear, make it a real directory, or install to `.agents/` and symlink the other way (documented in `docs/cli.md`).

https://claude.ai/code/session_01H8M4cjazvKLeXDgspyoroF


nstfkc:
Added  (64fb4a4).

The template was passing a literal `true`, so switching the compiler off to test whether it was the cause of something meant editing a checked-in `gemi.config.ts` and remembering to revert it. It now passes `reactCompiler()` from `gemi/config`:

```bash
GEMI_REACT_COMPILER=off bun dev   # still broken? not the compiler
```

A named opt-out rather than a boolean, matching `GEMI_COMPRESSION` — `0`, `false` and `no` all read as "off" to a human and only one of them could be the one that works. Case-insensitive; unset means **on**, because a variable nobody set must never read as "off" or an unoptimized client build ships to everyone whose environment does not mention the flag.

It lives in `gemi/config` rather than as a `process.env` read in each app's config so the flag means the same thing in every project — but the switch is still in the app's `gemi.config.ts`, so replacing it with `true`/`false`/your own condition is one line. Options pass through it: `reactCompiler({ compilationMode: "annotation" })`.

Verified on real builds, not just unit tests:

| | `compiler-runtime` refs in `dist/client` |
|---|---|
| unset | 80 |
| `off` from the shell | 0 |
| `off` from `.env` | 0 |
| `OFF` from `.env` | 0 |

All 40 view chunks still emitted in the disabled case. Suite is 167 files / 4023 passed.

packages/gemi/bin/version.ts
```diff
@@ -0,0 +1,136 @@
...
+ * `0.60.0-rc.2` has to be compared against `rc` — compared against `latest` it
+ * would be told to "upgrade" to the older stable release it deliberately went
+ * ahead of.
+ */
+export function channelFor(version: string): string {
+  const parsed = parseVersion(version);
+  const first = parsed?.prerelease[0];
+  // `0.60.0-1` is a prerelease with a numeric identifier and no channel name;
+  // there is no dist-tag to infer from it, so fall back to `latest`.
+  return first && !NUMERIC_RE.test(first) ? first : "latest";
```
nstfkc, line 95: `channelFor` pins the check to the tag the installed version came from. That fixes the backwards case, but it opens a forwards one: once a channel's dist-tag stops moving, everyone sitting on it goes quiet permanently.

That is today's state, not a hypothetical:

```
$ npm view gemi dist-tags
{ next: '0.49.0-rc.4', rc: '0.51.0-rc.14', latest: '0.59.0' }
```

An app on `0.51.0-rc.14` is eight minors behind `latest`, and with this change it gets no notice at all — `checkForUpdate` resolves `rc`, gets back the version it is already running, and `formatUpdateNotice` returns `null`. `gemi upgrade` with no argument prints *"gemi 0.51.0-rc.14 is already the newest release on `rc`"*. Both are true sentences that leave the user stranded on precisely the release the feature exists to move them off. `next` is worse — `0.49.0-rc.4` has been frozen for ten minors.

`compareVersions` already gets this right (`compareVersions("0.60.0", "0.60.0-rc.2") > 0`), so only the tag choice is in the way.

Suggestion: when `channelFor(installed) !== "latest"`, resolve both that tag and `latest`, and report whichever is newer. Two advisory requests instead of one, both already swallowing every failure, and an rc user hears about the stable release that superseded their branch instead of silence.

packages/gemi/bin/install-skill.ts
```diff
@@ -0,0 +1,239 @@
...
+    // Nothing to do, and saying so beats a silent success that looks like a
+    // no-op failure.
+    log(`${SKILL_NAME} is already up to date${version ? ` (gemi ${version})` : ""}.`);
+    return 0;
+  }
+
+  // Replace rather than merge: a rule deleted or renamed upstream has to
+  // disappear here too, and `cpSync` over the top would leave the old file
+  // behind for an agent to keep reading.
+  rmSync(target, { recursive: true, force: true });
```
nstfkc: `installSkill`'s docstring says *"Every failure is a sentence, never a stack"*, but nothing from `hashDirectory` down to the end of the function is guarded. A target that is not a directory throws straight out — run against a fixture where `.agents/skills/gemi-react-best-practices` is a regular file:

```
THREW: ENOTDIR: not a directory, scandir '.../.agents/skills/gemi-react-best-practices'
```

The `gemi upgrade` path is the more consequential one. `rmSync` deletes the installed copy *before* `cpSync` recreates it, so an IO failure in between (EACCES on a rule file, ENOSPC, a read-only `.agents/`) leaves the skill directory destroyed **and** the lock still holding the pre-delete hash — so the next run reports *"has local edits. Re-run with --force"* about a directory this command itself emptied.

And because `runUpgrade` calls `installSkillImpl` with no `try`, that throw rejects `runUpgrade`'s promise and commander turns it into an unhandled rejection: `gemi upgrade` exits with a stack trace *after* the package install already succeeded. That is the exact outcome the comment at the call site is written to prevent — "a skill copy that was declined … is a note, not a failed upgrade".

Wrapping the body in try/catch and returning 1 with the message covers both cases. Copying to a sibling temp dir and renaming over the target would additionally make the replace atomic, if that is worth the extra lines.

packages/gemi/bin/install-skill.ts
```diff
@@ -0,0 +1,239 @@
...
+  try {
+    const parsed = JSON.parse(readFileSync(file, "utf8"));
+    if (!parsed || typeof parsed !== "object") return empty;
+    return {
+      ...parsed,
+      version: typeof parsed.version === "number" ? parsed.version : 3,
+      skills: parsed.skills && typeof parsed.skills === "object" ? parsed.skills : {},
+    };
+  } catch {
+    return empty;
```
nstfkc, line 123: This is the one path where the invariant this file and `docs/cli.md` both promise does not hold.

Returning `empty` for a damaged lock is right — a bad lock must not block an install. But `writeLock` then serialises that empty object over the file, so every other tool's entry and every unknown top-level key goes with it. The neighbouring test says exactly what is at stake: *"Clobbering that would uninstall someone else's skill."*

The two tests bracket the gap without covering it — `preserves unrelated keys and other skills in the lock` uses a valid lock, and `a damaged lock does not block an install` asserts only the exit code and the copied file. A test asserting `find-skills` survives a truncated lock would fail today.

Cheapest fix that keeps both properties: on a parse failure, rename the file to `.skill-lock.json.corrupt` before writing the new one. The install still succeeds, and the other tool's state is recoverable rather than gone.

packages/gemi/bin/upgrade.ts
```diff
@@ -0,0 +1,216 @@
...
+  const installed = readInstalledVersion(rootDir);
+  if (!installed) {
+    error(
+      `Could not find an installed gemi under ${rootDir}. Run this from the ` +
+        `root of a gemi project (one with \`gemi\` in its node_modules).`,
+    );
+    return 1;
+  }
+
+  const kind = dependencyKind(rootDir) ?? "dependencies";
```
nstfkc: `readInstalledVersion` walks *up* from `rootDir` to find `node_modules/gemi`, but everything downstream stays at `rootDir` — `dependencyKind` here, `detectPackageManager` below, and `spawn(cmd, rootDir)` at the end. In a workspace those are different directories, and then the command reads a version from one manifest and writes to another.

Run from a monorepo root where gemi is declared in `apps/web/package.json` and hoisted: `dependencyKind` finds no gemi, falls through this `?? "dependencies"`, and `bun add gemi@0.59.0` adds a direct dependency to the root manifest that was never there — while `apps/web` keeps its old range and nothing about the actual install changes. Same shape from any subdirectory without a `package.json`.

`findGemiPackageDir` already knows the answer: `path.dirname(path.dirname(packageDir))` is the directory whose `node_modules` holds gemi. Either run the install there, or refuse when it differs from `rootDir` and name the directory to run from. The `?? "dependencies"` default is fine once gemi is known to be declared *somewhere* — it is the silent add to an unrelated manifest that is the problem.

docs/configuration.md
```diff
@@ -44,6 +44,69 @@ interface GemiConfig {
...
+});
+```
+
+`reactCompiler()` from `gemi/config` is the value to pass, with an environment switch in front of it. Pass a plain `true` / `false` instead to pin the choice.
+
+Two packages are required, both already in the scaffolded template:
+
+```bash
+bun add -d "@vitejs/plugin-react@^6.1.1" "oxc-transform-react@^0.145.0"
+```
```
nstfkc, line 68: *"the plugin will refuse to load it"* is not what happens. `@vitejs/plugin-react@6.1.1` does no version check at all:

```js
// dist/index.js:204
return compiler = await import("oxc-transform-react");
```

That is the whole of it — the surrounding `catch` fires only when the import fails outright, i.e. the package is absent. A `0.148.0` install loads and runs. What enforces `^0.145.0` is your package manager's peer resolution: a warning under bun/npm, an install failure under pnpm with strict peers.

Worth correcting because it changes what a reader does when they hit that warning. As written it reads as "the build will break"; in fact the build works, and the real risk is an unchecked API mismatch between the plugin and a transform version it never declared support for.

(The quoted error message in the next sentence *is* verbatim-correct — I checked it against the same file.)

templates/saas-starter/package.json
```diff
@@ -28,8 +28,9 @@
     "@types/bun": "^1.1.6",
     "@types/react": "19.0.2",
     "@types/react-dom": "19.0.2",
-    "@vitejs/plugin-react": "^6.0.0",
+    "@vitejs/plugin-react": "^6.1.1",
     "autoprefixer": "^10.4.19",
+    "oxc-transform-react": "^0.145.0",
```
nstfkc, line 33: Confirmed as described: `@vitejs/plugin-react@6.1.1` declares `oxc-transform-react: ^0.145.0` as an optional peer, and npm latest is already `0.148.0`, outside it.

What worries me is where the warning lives. It is in the PR description and in prose in `docs/configuration.md`, and nothing in the repo fails when it goes stale. The template pins `@vitejs/plugin-react` at `^6.1.1`, so a `6.2.0` that widens the peer range arrives on the next `bun install` — with a peer warning that every scaffolded app sees and nobody here does.

A ~10-line test would hold it: read `oxc-transform-react` from the template's `devDependencies`, read the optional-peer range from the installed `@vitejs/plugin-react` manifest, assert they agree. It fails the moment upstream moves, which is the moment you want to know.

nstfkc (commented):
Reviewed both halves. The CLI work is unusually well-tested — I ran the four new files and got **81 pass / 0 fail**, and the tests cover the cases that actually matter (`rc.2` vs `rc.10`, a prerelease below its release, the skill lock's `installedAt` surviving an update, `--dry-run` not touching the skill). The React Compiler claims all check out against the plugin source: the `consumer !== "server"` gate is real (`dist/index.js:225`), and the *"React Compiler requires the optional `oxc-transform-react` package"* error you quote is verbatim.

Six things below, three of which I think are worth acting on before merge.

**The one I would not ship as-is** is `channelFor` stranding prerelease users (comment on `version.ts`). `npm view gemi dist-tags` right now returns `rc: 0.51.0-rc.14` against `latest: 0.59.0` — so an rc user gets *no notice*, and `gemi upgrade` tells them they are current, eight minors behind. The reasoning in the description is sound for the backwards case; it just needs the forwards case too, and `compareVersions` already supports it.

The other two: `installSkill` throws rather than returning a sentence, which in the `upgrade` path turns a *successful* package install into an unhandled rejection; and a damaged `.skill-lock.json` silently takes every other tool's entry with it, which is the one case where the invariant the docstring and `docs/cli.md` both advertise does not hold.

The remaining three are smaller — a wrong claim in `docs/configuration.md` about how the peer range is enforced, the workspace `rootDir` mismatch in `upgrade`, and the absence of any in-repo guard on the `oxc-transform-react` drift you flagged.

---

On the two open questions in your description:

**Splitting the PR** — I would not bother. The halves are genuinely independent, but the only coupling is ground rule 5, and reviewing them together is what let me check that rule's claim against the plugin source in the same pass.

**The `.claude/skills/...` symlink** — separate from whether Claude Code resolves it, a committed symlink is a portability cost: a Windows checkout without `core.symlinks` materialises it as a *text file* containing `../../packages/gemi/skills/gemi-react-best-practices`, and any tool reading that directory gets a one-line file where a skill should be. Given `docs/cli.md` already documents the `.agents/` → `.claude/` direction as the user-facing recipe, having the repo commit the arrangement in reverse is worth a second look — a real directory, or nothing checked in at all, both avoid the question.

nstfkc:
All six addressed in be10596, plus two things you did not see because they were not in the diff you were reading.

### The frozen dist-tag — you were right, and it was worse than the PR admitted

Verified against the live registry: `{ next: '0.49.0-rc.4', rc: '0.51.0-rc.14', latest: '0.59.0' }`. `fetchNewestPublished` now asks a prerelease channel's tag **and** `latest` and takes whichever is newer. A stable install still makes one request — nothing on a prerelease tag is an upgrade from it, so the second answer would only be fetched to discard it. Three tests: the frozen-`rc` case, a channel still legitimately ahead of `latest`, and the one-request assertion.

### `installSkill` throwing

Reproduced exactly as you described (`ENOTDIR`). Body is wrapped now. I also took the atomic suggestion: the copy stages into a sibling and renames over the target, so a failure part-way cannot leave the directory destroyed **and** the lock holding the pre-delete hash — the state where the next run blames the user for edits the command itself made.

### The damaged lock

Moved to `.skill-lock.json.corrupt` before the fresh one is written. Your read of the two bracketing tests was right; there is now one that truncates a lock containing `find-skills` and asserts it survives.

### The workspace `rootDir`

Went with "run the install where gemi is declared" over "refuse". `findDeclaringDir` walks up for the manifest that actually declares it, and `detectPackageManager` now walks up too (it was reporting bun's default for a pnpm monorepo). It refuses only when *no* manifest anywhere declares gemi — no silent add. Four tests on a three-directory workspace fixture.

### The docs claim

Correct, and I had checked the same file and still wrote the wrong sentence. Rewritten to say the plugin does no version check, peer resolution is what enforces the range, and the real risk is an unchecked API mismatch rather than a broken build.

### The pin guard

`react-compiler-peer.test.ts` holds the template's range against the plugin's declared peer. Skips when the template has no `node_modules`, fails when upstream moves.

### The symlink

Removed, on the Windows point rather than the discovery one — `core.symlinks` off turns it into a text file holding the target path, which is a worse failure than not having it. `docs/cli.md` documents the local link and says to keep it uncommitted.

---

### Two things the review could not have caught

The author flagged that the skill teaches APIs gemi does not have. It did, in three places:

- **`TranslationService.transform`** — an entire rule, 50 lines with a worked example, built on a class that exists in one private application and nowhere in this package. Deleted; gemi's server-side translation is `translate(dictionary, key, …)`, a per-key read with no "transform a structure" analogue, so there was nothing to rewrite it into.
- **`useDebounceValue`** — listed in a table of gemi replacements, actually from `usehooks-ts`.
- **`orm-analytics-connection`** — described an app's four-pool topology, its `mapWithConcurrency` helper, its config file, two issue numbers and an incident date as the framework's. Rewritten around what gemi has: `Model.on(name)`, `DB.connection(name)`, per query not per model.

`skills.test.ts` is the guard — every `gemi/*` import, every identifier presented as a gemi hook, every facade-position call, checked against the real export lists. It caught a leftover `useDebounceValue` while I was writing it. It cannot check prose and says so in its header.

### Project structure

Added on request, scoped to gemi's canonical `app/` layout: an annotated tree and a where-does-this-go table in `SKILL.md`, plus `structure-discovered-vs-registered` and `structure-do-not-reinvent-the-framework`.

---

**169 files / 4040 passed.** Also reverted 14 files that `oxfmt bin/` swept up — they had no changes of mine in them, and 52 lines of reformat churn is not something to bury this diff under.

